PR #5980, opened 2020-07-14
Added List & details page for SnapshotContent

cloudbehl:
Screenshot 2020-07-14 at 5 51 16 PM

Screenshot 2020-07-14 at 5 51 04 PM

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
@@ -1,0 +1,107 @@
+import * as React from 'react';
+import {
+  SectionHeading,
+  ResourceSummary,
+  ResourceLink,
+  navFactory,
+  Kebab,
+  humanizeBinaryBytes,
+} from '@console/internal/components/utils';
+import { VolumeSnapshotClassModel, VolumeSnapshotModel } from '@console/internal/models';
+import { referenceForModel, VolumeSnapshotContentKind } from '@console/internal/module/k8s';
+import { DetailsPage, DetailsPageProps } from '@console/internal/components/factory';
+import { Status } from '@console/shared';
+import { ResourceEventStream } from '@console/internal/components/events';
+import { volumeSnapshotStatus } from '../../status';
+
+const { editYaml, events } = navFactory;
+
+const Details: React.FC<DetailsProps> = ({ obj }) => {
+  const { deletionPolicy, driver } = obj?.spec;
+  const { volumeHandle, snapshotHandle } = obj?.spec?.source || {};
+  const { name: snapshotName, namespace: snapshotNamespace } = obj?.spec?.volumeSnapshotRef || {};
+  const size = obj.status?.restoreSize;
+  const sizeMetrics = size ? humanizeBinaryBytes(size).string : '-';
+
+  return (
+    <div className="co-m-pane__body">
+      <SectionHeading text="Volume Snapshot Details" />
+      <div className="row">
+        <div className="col-md-6 col-xs-12">
+          <ResourceSummary resource={obj}>
+            <dt>Status</dt>
+            <dd>
+              <Status status={volumeSnapshotStatus(obj)} />
+            </dd>
+          </ResourceSummary>
+        </div>
+        <div className="col-md-6">
+          <dl className="co-m-pane__details">
+            {size && (
+              <>
+                <dt>Size</dt>
+                <dd>{sizeMetrics}</dd>
+              </>
+            )}
+            <dt>Volume Snapshot</dt>
+            <dd>
+              <ResourceLink
+                kind={referenceForModel(VolumeSnapshotModel)}
+                name={snapshotName}
+                namespace={snapshotNamespace}
+              />
+            </dd>
+            <dt>Snapshot Class</dt>
+            <dd>
+              <ResourceLink
+                kind={referenceForModel(VolumeSnapshotClassModel)}
+                name={obj?.spec?.volumeSnapshotClassName}
+              />
+            </dd>
+            <dt>Deletion Policy</dt>
+            <dd>{deletionPolicy}</dd>
+            <dt>Driver</dt>
+            <dd>{driver}</dd>
+            {volumeHandle && (
+              <>
+                <dt>Volume Handle</dt>
+                <dd>{volumeHandle}</dd>
+              </>
+            )}
+            {snapshotHandle && (
+              <>
+                <dt>Snapshot Handle</dt>
+                <dd>{snapshotHandle}</dd>
+              </>
+            )}
+          </dl>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const pages = [
+  {
+    href: '',
+    name: 'Details',
+    component: Details,
+  },
+  editYaml(),
+  events(ResourceEventStream),
+];
+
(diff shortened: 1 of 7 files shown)


frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { DetailsPage, DetailsPageProps } from '@console/internal/components/factory';
+import { Status } from '@console/shared';
+import { volumeSnapshotStatus } from '../../status';
+import { ResourceEventStream } from '@console/internal/components/events';
+
+const { editYaml, events } = navFactory;
+
+const Details: React.FC<DetailsProps> = ({ obj }) => {
+  const { deletionPolicy, driver } = obj?.spec;
+  const { volumeHandle, snapshotHandle } = obj?.spec?.source;
```
bipuladh, line 21:
```suggestion
  const { volumeHandle, snapshotHandle } = obj?.spec?.source || {};
```
Whenever you're using optional chaining with destructuring if the property doesn't exist then it will try to destructure `undefined` and fail.

frontend/public/module/k8s/types.ts
```diff
@@ -981,8 +981,34 @@ export type VolumeSnapshotKind = K8sResourceCommon & {
...
+export type VolumeSnapshotContentKind = K8sResourceCommon & {
+  status: {
+    readyToUse: boolean;
+    restoreSize?: number;
+    snapshotHandle?: string;
+    error?: {
+      message: string;
+      time: string;
+    };
+  };
```
bipuladh, line 1000: Let's create a common volume snapshot status type and use that. It will make the type declaration common between them.

frontend/packages/console-app/src/plugin.tsx
```diff
@@ -59,7 +63,10 @@ type ConsumedExtensions =
   | DashboardsOverviewHealthPrometheusSubsystem
   | DashboardsOverviewInventoryItem
   | DashboardsOverviewHealthOperator<ClusterOperator>
-  | ReduxReducer;
+  | ReduxReducer
```
bipuladh, line 66: I would prefer it if we do not pass it via plugins. We already have Nodes, Volume Snapshot, and Volume Snapshot Class on the resource page. Let's not mix these two. Or let's do it the other way and move the aforementioned resources here.
Let's be consistent.

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
```diff
@@ -0,0 +1,108 @@
...
+  ResourceLink,
+  navFactory,
+  Kebab,
+  humanizeBinaryBytes,
+} from '@console/internal/components/utils';
+import { VolumeSnapshotClassModel, VolumeSnapshotModel } from '@console/internal/models';
+import { referenceForModel, VolumeSnapshotContentKind } from '@console/internal/module/k8s';
+import { DetailsPage, DetailsPageProps } from '@console/internal/components/factory';
+import { Status } from '@console/shared';
+import { volumeSnapshotStatus } from '../../status';
```
bipuladh, line 14: Move it to the end

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const { editYaml, events } = navFactory;
+
+const Details: React.FC<DetailsProps> = ({ obj }) => {
+  const { deletionPolicy, driver } = obj?.spec;
+  const { volumeHandle, snapshotHandle } = obj?.spec?.source;
+  const size = obj.status?.restoreSize;
+  const sizeMetrics = size ? humanizeBinaryBytes(size).string : '-';
+  const snapshotName = obj?.spec?.volumeSnapshotRef?.name;
+  const snapshotNamespace = obj?.spec?.volumeSnapshotRef?.namespace;
```
bipuladh, line 25:
```suggestion
  const {name: snapshotName, namespace: snapshotNamespace} = obj?.spec?.volumeSnapshotRef || {};
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content-details.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </dd>
+            <dt>Snapshot Class</dt>
+            <dd>
+              <ResourceLink
+                kind={referenceForModel(VolumeSnapshotClassModel)}
+                name={obj?.spec?.volumeSnapshotClassName}
+              />
+            </dd>
+            <dt>Deletion Policy</dt>
+            <dd>{deletionPolicy || '-'}</dd>
```
bipuladh, line 63: Can the deletion policy be `undefined`?

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-content.tsx
```diff
@@ -0,0 +1,144 @@
...
+    props: { className: tableColumnClasses[6] },
+  },
+];
+
+const Row: RowFunction<VolumeSnapshotContentKind> = ({ key, obj, style, index }) => {
+  const { name, creationTimestamp } = obj?.metadata || {};
+  const size = obj.status?.restoreSize;
+  const sizeMetrics = size ? humanizeBinaryBytes(size).string : '-';
+  const snapshotName: string = obj?.spec?.volumeSnapshotRef?.name;
+  const snapshotNamespace: string = obj?.spec?.volumeSnapshotRef?.namespace;
```
bipuladh, line 85:
```suggestion
  const {name: snapshotName, namespace: snapshotNamespace } = obj?.spec?.volumeSnapshotRef || {};
 ```

File: frontend/packages/console-app/src/status/snapshot.ts [hunk 8]
```diff
@@ -1,7 +1,24 @@
-import { VolumeSnapshotKind } from '@console/internal/module/k8s';
+import { VolumeSnapshotKind, VolumeSnapshotContentKind } from '@console/internal/module/k8s';
 
-export const volumeSnapshotStatus = ({ status }: { status: VolumeSnapshotKind['status'] }) => {
+export const volumeSnapshotStatus = ({
+  status,
+}: {
+  status: VolumeSnapshotKind['status'] | VolumeSnapshotContentKind['status'];
```
bipuladh, line 6:
```suggestion
  status: VolumeSnapshotKind['status'];
```
They are the same. We could also refactor the status field from the type and make a common `VolumeSnapshotStatus` and use it in the two Kinds.

frontend/public/models/index.ts
```diff
@@ -980,7 +980,7 @@ export const VolumeSnapshotContentModel: K8sKind = {
   abbr: 'VSC',
   namespaced: false,
   kind: 'VolumeSnapshotContent',
-  id: 'volumesnapshotcontents',
-  labelPlural: 'Volume Snaphot Contents',
+  id: 'volumesnapshotcontent',
```
bipuladh, line 983: Good catch. Can you make changes in Line 971 too?

frontend/packages/console-app/src/plugin.tsx
rawagner: I prefer using the extension points, I find it easier to maintain.

PR #5981, opened 2020-07-14
Disable error overlay of react-refresh

glekner:
image

----------

Disable react-refresh error overlay, it hurts the developing experience, and requires an extra click to close it

### DIFF
--- a/frontend/webpack.config.ts
+++ b/frontend/webpack.config.ts
@@ -184,9 +184,7 @@ const config: Configuration = {
     ...(IS_WDS
       ? [
           new ReactRefreshWebpackPlugin({
-            overlay: {
-              sockPort: WDS_PORT,
-            },
+            overlay: false,
           }),
         ]
       : []),


spadgett:
@glekner Are these legitimate errors we should be fixing? This shows for uncaught exceptions and unhandled failed promises, right?

glekner:
@spadgett  It shows for failed promises even if you handle them. We already see these errors in the browser's console or in your UI component if it knows to handle errors. this overlay is showing what we were already seeing, basically annoying.

spadgett:
> @spadgett It shows for failed promises even if you handle them. We already see these errors in the browser's console or in a UI component if it knows to handle errors. this overlay is showing what we were already seeing, basically annoying.

Are you sure it shows if you handle the failed promise? I just tested an invalid name in YAML editor, and I didn't get the screen. That should be a failed promise.

I have a feeling these are errors we shouldn't be ignoring.

/hold

glekner:
@spadgett here is an example

ezgif com-video-to-gif

What do you mean by ignoring? we were exposed to these errors before react-refresh was merged

spadgett:
> What do you mean by ignoring? we were exposed to these errors before react-refresh was merged

Yes, and it's good react-refresh tells us there might be a problem! Instead of disabling the overlay, let's investigate and fix the cause instead. In general, we should avoid unhandled promise rejection.

It appears to be something in our error handling in dialogs:

Screen Shot 2020-07-15 at 9 09 15 AM

glekner:
@spadgett yes, the `handlePromise` function is handling the promise but it also returns a promise, so usages who don't implement a `catch` will get this error.
I guess the solution is either handling `catch` in all dialogs which we don't do atm, we mostly use eslint:
`    // eslint-disable-next-line promise/catch-or-return`

Or, we don't return a Promise, and we pass a success function as an argument to `handlePromise`, so it will fire it with the promise's response. 
So Instead of `handlePromise(promise).then(closeModal)` we will use `handlePromise(promise, closeModal)`

thoughts? @rawagner @suomiy

suomiy:
After analyzing the code base it is clear this is a pain point for others as well.

There are some legitimate catch use cases like this one.

```
 .catch((errorMessage) => {
      actions.setSubmitting(false);
      actions.setStatus({ submitError: errorMessage });
)
``` 

but most of the people opt for one of these "solutions".


```
// eslint-disable-line promise/catch-or-return
```

```
.catch((error) => {
    throw error;
});
```

```
.catch((e) => {
    console.error(e);
});
```
    
```
.catch(() => null);
```

```
.catch(_.noop);
```


So we have to decide if we want to enforce strict behaviour, eg:

 ``` 
handlePromise(promise, onFulfilled = null, onRejected = null)
```

and forbidding `then.catch` semantic and resolving catch by default if onRejected param is missing.

or if we want to keep the promise semantic and replacing all misused places with the following snippet (or an alternative) in the codebase:

```
.catch((e) => {
    console.error(e);
});
```

glekner:
I think devs got used to the fact that `handlePromise` returns a promise.
I added missing catches and removed all `// eslint-disable-next-line promise/catch-or-return` from the console

https://github.com/openshift/console/pull/6048

PR #5984, opened 2020-07-14
Bug 1856803: Fix incorrect monitoring doc link

spadgett:
The doc link in the OperatorHub install page for non-Red Hat operators that request monitoring was incorrect.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1856803

/assign @rhamilto


jerolimov:
Hope its ok, could verify this as well

/lgtm

PR #5985, opened 2020-07-14
migrate to patternfly/react-topology

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4102

**Analysis / Root cause**: 
Topology package has been migrated to `patternfly/react-topology`.

**Solution Description**: 
Remove topology package from console.
Update all imports.
Update to align with API changes.
 - Use `VisualizationProvider` to set controller.
 - Update `useSelection` which now accepts options object instead of multiple params.
 - Update `EventingPubSubLink.tsx` to fix drawing of the trigger filter decoration because `getPathAlongTheLine` is not in pf-topology. Chose to not move code to a new util because there was issues with how the decorator was rendered.

Before:
image
After:
image

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [x] Safari
- [x] Edge

/assign @jeff-phillips-18 

_Requires updating to latest patternfly_


frontend/packages/dev-console/src/components/topology/components/nodes/WorkloadNode.tsx
```diff
@@ -73,7 +73,9 @@ const ObservedWorkloadNode: React.FC<WorkloadNodeProps> = ({
         content={tipContent}
         trigger="manual"
         isVisible={dropTarget && canDrop}
-        tippyProps={{ duration: 0, delay: 0 }}
+        animationDuration={0}
+        entryDelay={0}
+        exitDelay={0}
```
dtaylor113, line 78: entryDelay & exitDelay default to 0, so you can remove these.

christianvogt:
waiting on https://github.com/openshift/console/pull/5987

PR #5986, opened 2020-07-14
Use DetailsItem component on operand details page

TheRealJon:
- Convert SpecDescriptor and StatusDescriptor components to use DetailsItem. 
- Provide JSONSchema to details items so that description can be displayed in the popover. 
- DetailsItem now shows the popover interaction if either a description or a path is provided. If no description is present, the popover will still be shown with just the heading and path.

Story https://issues.redhat.com/browse/CONSOLE-2283
Sub-task https://issues.redhat.com/browse/CONSOLE-2311

**Details Item With Description**
details item w description

**Details Item Without Description**
details item wo description

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/index.spec.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/index.spec.tsx
@@ -1,156 +1,113 @@
 import * as React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
-import { ResourceLink, Selector } from '@console/internal/components/utils';
+import { Button } from '@patternfly/react-core';
+import { Provider } from 'react-redux';
+import { Router } from 'react-router';
+import { mount, ReactWrapper } from 'enzyme';
+import store from '@console/internal/redux';
+import { ResourceLink, Selector, history } from '@console/internal/components/utils';
 import { DescriptorProps, SpecCapability, Descriptor } from '../types';
 import { testResourceInstance, testModel } from '../../../../mocks';
-import { EndpointList, Endpoint } from './endpoint';
+import { EndpointList } from './endpoint';
 import { ResourceRequirementsModalLink } from './resource-requirements';
 import * as configureSize from './configure-size';
 import { SpecDescriptor } from '.';
-import { Button } from '@patternfly/react-core';
+
+const OBJ = {
+  ...testResourceInstance,
+  spec: {
+    ...testResourceInstance.spec,
+    pods: 3,
+    endpoints: [{ targetPort: 80, scheme: 'TCP' }],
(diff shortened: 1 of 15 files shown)


frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/index.spec.tsx
```diff
@@ -1,242 +1,162 @@
...
 describe(SpecDescriptor.name, () => {
-  let wrapper: ShallowWrapper<DescriptorProps>;
+  let wrapper: ReactWrapper<DescriptorProps>;
   let descriptor: Descriptor;
 
   beforeEach(() => {
     descriptor = {
       path: '',
       displayName: 'Some Spec Control',
-      description: 'This is a description',
```
TheRealJon, line 20: Removed the description because Popover component causes issues with enzyme.

TheRealJon:
cc @spadgett

TheRealJon:
/hold

Just found an issue. The path in each popover is excluding 'status'/'spec'

TheRealJon:
/assign @spadgett 

cc @tlwu2013 @itsptk

tlwu2013:
Hi @TheRealJon , 

Per discussed on the Slack, could you help make sure on the "popover":

```
IF `description` from CSV exist
            Show `description` on the popover

ELSE IF `description` from schema exist 
            Show `description` on the popover

ELSE `description` from schema exist 
            No `description` on the popover,
            but still has the field `displayName` (title on the top) + the `path` (breadcrumbs at the bottom)
```

Thanks!

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
```diff
@@ -445,141 +451,139 @@ const OperandDetailsSection: React.FC = ({ children }) => (
...
+    </div>
+  );
+});
 
 const ResourcesTab = (resourceProps) => (
   <Resources {...resourceProps} clusterServiceVersion={resourceProps.csv} />
 );
 
-export const OperandDetailsPage = connectToPlural((props: OperandDetailsPageProps) => {
+// FIXME (jon) useK8sModel hook instead of connectToPlural
```
spadgett, line 572: It looks like you've addressed the fixme?

frontend/public/kinds.ts
```diff
@@ -50,7 +50,7 @@ export const connectToPlural: ConnectToPlural = connect(
       match: match<{ plural: GroupVersionKind | string }>;
     },
   ) => {
-    const plural = props.plural || _.get(props, 'match.params.plural');
+    const plural = props.plural ?? props.match?.params?.plural;
```
spadgett, line 53: I'm not sure if this could be an issue, but this does change the behavior if `props.plural` is the empty string.

PR #5988, opened 2020-07-14
[release-4.5] Bug 1856867: Do not prune empty values from sample on create operand page

TheRealJon:
Manual cherry pick of #5966


TheRealJon:
@spadgett Took a look at the conflicts. They just came from some util functions that were moved around since 4.5 release.

PR #5989, opened 2020-07-14
Add PVC Extensions

glekner:
Added 3 Extensions to `persistent-volume-claim.jsx`

`PVCAlert`
`PVCStatus`
`PVCCreateProp`

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/pvc.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/pvc.ts
@@ -1,0 +1,49 @@
+import { K8sResourceCommon } from '@console/internal/module/k8s';
+import { Extension, LazyLoader } from './base';
+
+namespace ExtensionProperties {
+  export interface PVCAlert<T> {
+    /** Loader for the corresponding Alert component. */
+    loader: LazyLoader<T>;
+  }
+
+  export interface PVCStatus<T> {
+    /** Loader for the corresponding Status component. */
+    loader: LazyLoader<T>;
+    /** predicate that tells if to render the status or not */
+    predicate: (pvc: K8sResourceCommon) => boolean;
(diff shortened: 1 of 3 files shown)


frontend/public/components/persistent-volume-claim.jsx
```diff
@@ -34,9 +36,26 @@ const menuActions = [
...
-);
+const PVCStatus = ({ pvc }) => {
+  const pvcStatusExtensions = useExtensions(isPVCStatus);
+  if (pvcStatusExtensions.length > 0) {
+    const priorityStatus = pvcStatusExtensions.reduce((prev, current) =>
+      prev.properties.priority > current.properties.priority ? prev : current,
+    );
+    const predicate = priorityStatus.properties.predicate;
+
+    const priorityStatusComponent = predicate(pvc) && (
```
rawagner, line 47: you need to pick the extension status which returns true and has highest priority. 

Doing it this way, if the extension status with highest priority returns false, you will just ignore the others and fallback to default status which is not correct.

frontend/packages/console-plugin-sdk/src/typings/pvc.ts
```diff
@@ -0,0 +1,51 @@
...
+    loader: LazyLoader<T>;
+    /** predicate that tells if to render the status or not */
+    predicate: (pvc: K8sResourceCommon) => boolean;
+    /** priority for the corresponding Status component. bigger is prioritized */
+    priority: number;
+  }
+
+  export interface PVCCreateProp {
+    /** key for the create prop */
+    key: string;
```
rawagner, line 21: do you actually need to specify `key` ? What if there are two extensions with same key ? I think you should just generate the `key` in code, using index is fine I think.

PR #5993, opened 2020-07-15
Fix critical alerts not able to close in notification drawer

jcaianirh:
The usePrevious hook has changed in the last week to add dependency. adding a dependency to the use of usePrevious in the notification drawer code allowed the state to update correctly and open and close the critical alerts category

### DIFF
--- a/frontend/public/components/notification-drawer.tsx
+++ b/frontend/public/components/notification-drawer.tsx
@@ -231,7 +231,7 @@ export const ConnectedNotificationDrawer_: React.FC<ConnectedNotificationDrawerP
     true,
   );
   const [isClusterUpdateExpanded, toggleClusterUpdateExpanded] = React.useState<boolean>(true);
-  const prevDrawerToggleState = usePrevious(isDrawerExpanded);
+  const prevDrawerToggleState = usePrevious(isDrawerExpanded, [isAlertExpanded]);
 
   const hasCriticalAlerts = criticalAlertList.length > 0;
   const hasNonCriticalAlerts = otherAlertList.length > 0;


frontend/public/components/notification-drawer.tsx
```diff
@@ -231,7 +231,7 @@ export const ConnectedNotificationDrawer_: React.FC<ConnectedNotificationDrawerP
     true,
   );
   const [isClusterUpdateExpanded, toggleClusterUpdateExpanded] = React.useState<boolean>(true);
-  const prevDrawerToggleState = usePrevious(isDrawerExpanded);
+  const prevDrawerToggleState = usePrevious(isDrawerExpanded, [isAlertExpanded]);
```
spadgett, line 234: @jcaianirh I'm a little unclear how the drawer expanded state depends on the alert expanded state. Can you help me understand?

jcaianirh: @spadgett this may not be the fix i'm looking for...I've put this on hold. Prior to this fix, usePrevious took no dependencies. Since the change the in usePrev the prevToggleState isn't being updated after the first state change (thinking its like passing an empty dependency array to useEffect). isAlertExpanded is not the right dependency here after thinking more about it. I thought I just needed something to trigger the update of prevstate, but this is updating too often.

spadgett: Looks like this might be fixed by #6002? Can you check?

cc @rawagner

jcaianirh:
closing this as @rawagner has a pr to fix usePrevious: https://github.com/openshift/console/pull/6002

PR #5995, opened 2020-07-15
update namespace for sources created through yaml if edited

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4227

**Analysis / Root cause**: 
If user omits namespace from yaml and tries to create it gives 404 as these sources are namespaced

**Solution Description**: 
Update namespace if not present with current namespace user is in

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified and works as expected
/lgtm

PR #5996, opened 2020-07-15
Fix queries and constants organization in Noobaa Plugin

bipuladh:


frontend/packages/noobaa-storage-plugin/src/constants/strings.ts
```diff
@@ -0,0 +1,17 @@
...
+export const BUCKET_CLASS = 'Bucket Class';
+export const OCS_NS = 'openshift-storage';
+export const NB_PROVISIONER = 'noobaa.io/obc';
+
+export const CHART_LABELS = {
+  [BY_LOGICAL_USAGE]: 'Logical used capacity per account',
+  [BY_PHYSICAL_VS_LOGICAL_USAGE]: 'Physical vs. Logical used capacity',
+  [BY_EGRESS]: 'Egress Per Provider',
+  [BY_IOPS]: 'I/O Operations count',
+};
```
afreen23, line 17: This should be in a utils folder for data consumption card

frontend/packages/noobaa-storage-plugin/src/queries.ts
```diff
@@ -2,21 +2,26 @@ import { ProjectModel } from '@console/internal/models';
 import { PROJECTS, BUCKET_CLASS } from './constants';
```
cloudbehl, line 2: can we call it `dashboard-queries.ts` to be more clear?

afreen23: Its possible that other noobaa components would use same queries in future, so we can keep it as it is for now.

frontend/packages/noobaa-storage-plugin/src/constants/data-consumption.ts
```diff
@@ -0,0 +1,17 @@
+export const ACCOUNTS = 'Accounts';
+export const PROVIDERS = 'Providers';
+export const BY_IOPS = 'I/O Operations';
+export const BY_LOGICAL_USAGE = 'Logical Used Capacity';
+export const BY_PHYSICAL_VS_LOGICAL_USAGE = 'Physical Vs Logical Usage';
+export const BY_EGRESS = 'Egress';
+export const PROJECTS = 'Projects';
+export const BUCKET_CLASS = 'Bucket Class';
+export const OCS_NS = 'openshift-storage';
+export const NB_PROVISIONER = 'noobaa.io/obc';
```
afreen23, line 10: This should be not be in data-consumption, rather in a generic file e.g `common.ts`

frontend/packages/noobaa-storage-plugin/src/constants/data-consumption.ts
```diff
@@ -0,0 +1,15 @@
+export const ACCOUNTS = 'Accounts';
+export const PROVIDERS = 'Providers';
+export const BY_IOPS = 'I/O Operations';
+export const BY_LOGICAL_USAGE = 'Logical Used Capacity';
+export const BY_PHYSICAL_VS_LOGICAL_USAGE = 'Physical Vs Logical Usage';
+export const BY_EGRESS = 'Egress';
+export const PROJECTS = 'Projects';
+export const BUCKET_CLASS = 'Bucket Class';
```
afreen23, line 8: These two we can later move to breakdown file.

PR #5998, opened 2020-07-15
add tool-tip to remove icon

debsmita1:
**Jira Stroy:**
https://issues.redhat.com/browse/ODC-4247

**Solution Description:**
This PR adds "Remove" tooltip to `MinusCircleIcon` & "Add" tooltip to `PlusCircleIcon` button

**Screenshots:**
Screenshot from 2020-07-15 13-49-09
Screenshot from 2020-07-15 13-33-34
Screenshot from 2020-07-15 13-33-27
Screenshot from 2020-07-15 13-30-57
Screenshot from 2020-07-15 13-30-47
Screenshot from 2020-07-15 13-30-40
Screenshot from 2020-07-15 13-58-43
Screenshot from 2020-07-15 13-57-59
Screenshot from 2020-07-15 15-17-54
Screenshot from 2020-07-15 15-58-45
Screenshot from 2020-07-15 22-48-53

### DIFF
--- a/frontend/packages/dev-console/src/components/health-checks/RequestTypeForms.tsx
+++ b/frontend/packages/dev-console/src/components/health-checks/RequestTypeForms.tsx
@@ -71,7 +71,6 @@ export const HTTPRequestTypeForm: React.FC<RequestTypeFormProps> = ({ probeType 
           nameValuePairs={nameValue}
           valueString="Value"
           nameString="Header Name"
-          addString="Add Header"
           readOnly={false}
           allowSorting={false}
           updateParentData={handleNameValuePairs}
@@ -105,7 +104,6 @@ export const CommandRequestTypeForm: React.FC<RequestTypeFormProps> = ({ probeTy
     <TextColumnField
       name={`healthChecks.${probeType}.data.exec.command`}
       label="Command"
-      addLabel="Add command"
       placeholder="argument"
       helpText="The command to run inside the container."
       required
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineParameters.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineParameters.tsx
@@ -9,13 +9,12 @@ type PipelineParametersProps = {
 };
 
 const PipelineParameters: React.FC<PipelineParametersProps> = (props) => {
-  const { addLabel = 'Add Pipeline Parameter', fieldName, isReadOnly = false } = props;
+  const { fieldName, isReadOnly = false } = props;
   const emptyMessage = 'No parameters are associated with this pipeline.';
 
   return (
     <MultiColumnField
       name={fieldName}
-      addLabel={addLabel}
       headers={['Name', 'Description', 'Default Value']}
       emptyValues={{ name: '', description: '', default: '' }}
       emptyMessage={emptyMessage}
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineResources.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineResources.tsx
@@ -10,12 +10,11 @@ type PipelineResourcesParam = {
 };
 
 const PipelineResources: React.FC<PipelineResourcesParam> = (props) => {
-  const { addLabel = 'Add Pipeline Resource', fieldName, isReadOnly = false } = props;
+  const { fieldName, isReadOnly = false } = props;
   const emptyMessage = 'No resources are associated with this pipeline.';
   return (
     <MultiColumnField
       name={fieldName}
-      addLabel={addLabel}
       headers={['Name', 'Resource Type']}
       emptyValues={{ name: '', type: '' }}
       emptyMessage={emptyMessage}
--- a/frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
+++ b/frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
@@ -23,10 +23,8 @@ const ProjectAccessForm: React.FC<FormikProps<FormikValues>> = ({
     <div className="co-m-pane__form">
       <MultiColumnField
         name="projectAccess"
-        addLabel="Add Access"
         headers={['Name', 'Role']}
         emptyValues={{ user: '', role: '' }}
-        toolTip="Remove Access"
       >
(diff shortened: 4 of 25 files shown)


debsmita1:
/cc @beaumorley @serenamarie125 ^
Here https://issues.redhat.com/browse/ODC-4248 I have added a list which contains the areas that I have fixed

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineParameters.tsx
```diff
@@ -9,13 +9,12 @@ type PipelineParametersProps = {
...
 
 const PipelineParameters: React.FC<PipelineParametersProps> = (props) => {
-  const { addLabel = 'Add Pipeline Parameter', fieldName, isReadOnly = false } = props;
+  const { fieldName, isReadOnly = false } = props;
   const emptyMessage = 'No parameters are associated with this pipeline.';
 
   return (
     <MultiColumnField
       name={fieldName}
-      addLabel={addLabel}
```
andrewballantyne, line 18: Hmm, was this needed to be removed?

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineResources.tsx
```diff
@@ -10,12 +10,11 @@ type PipelineResourcesParam = {
 };
 
 const PipelineResources: React.FC<PipelineResourcesParam> = (props) => {
-  const { addLabel = 'Add Pipeline Resource', fieldName, isReadOnly = false } = props;
+  const { fieldName, isReadOnly = false } = props;
   const emptyMessage = 'No resources are associated with this pipeline.';
   return (
     <MultiColumnField
       name={fieldName}
-      addLabel={addLabel}
```
andrewballantyne, line 18: Same here; was this needed?

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -23,10 +23,8 @@ const ProjectAccessForm: React.FC<FormikProps<FormikValues>> = ({
     <div className="co-m-pane__form">
       <MultiColumnField
         name="projectAccess"
-        addLabel="Add Access"
```
andrewballantyne, line 26: Not sure we want to drop this... it looks odd imo:
image

andrewballantyne:
Yeah, not a fan of this:
image

I think in this case the `Add` should have more description as there are two types of "add"s for this item.

debsmita1:
@andrewballantyne - @beaumorley & I conversed on whether to have simply `Add` or `Add <context>`. The conversation is here https://issues.redhat.com/browse/ODC-4248. I have added a list of all the places I have made the changes in the mentioned subtask. We can collaborate and decide on which areas we want to have the context mentioned with Add/Remove & which areas we want it simply Add/Remove.

Add/Remove can be the default addString/tooltip & if user wants a customized one it can be passed as props.

frontend/packages/dev-console/src/components/health-checks/RequestTypeForms.tsx
```diff
@@ -71,7 +71,6 @@ export const HTTPRequestTypeForm: React.FC<RequestTypeFormProps> = ({ probeType
           nameValuePairs={nameValue}
           valueString="Value"
           nameString="Header Name"
-          addString="Add Header"
```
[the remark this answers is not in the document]
debsmita1: I referred to this doc https://docs.google.com/document/d/1i_mPwL5ZNwrkSHIYAorQjIZDJDLvD6ikffNkq52eTFA/edit?ts=5e94bacb (page 3) while working on the health checks feature

andrewballantyne:
> @andrewballantyne - @beaumorley & I conversed on whether to have simply `Add` or `Add <context>`. The conversation is here https://issues.redhat.com/browse/ODC-4248. I have added a list of all the places I have made the changes in the mentioned subtask. We can collaborate and decide on which areas we want to have the context mentioned with Add/Remove & which areas we want it simply Add/Remove.
> 
> Add/Remove can be the default addString/tooltip & if user wants a customized one it can be passed as props.

I think on paper it looked good... imo it does not look good in practice. Remove has context of what it is... but Add does not always have context. I think we should leave "Add" alone and not truncate it down. 

If there are no rows, we are making a big leap to the user reading the block of text that is above it; my experience in what users do / do not do... is read text lol. I think the Add "button" being more descriptive is better imo.

@beaumorley Definitely will need you to weigh in here.

beaumorley:
@andrewballantyne I think you are saying that sometimes "Add" alone is not clear enough. If there is nothing there and they see "Add" they may not know what they are adding.  I like to keep the verbiage down when we can but I see what you are saying. If we want to add specific Add <xxx> (which sounds like a good call) we would like to help create the copy. I think we could do this after HLD? Could we use the list that was generated in the subtask to do that or do you have a lookup table somewhere?

I would say we should offer this option to include context to both Add and Remove. Then we can decide if it is needed.

Thanks! @debsmita1

andrewballantyne:
> @andrewballantyne I think you are saying that sometimes "Add" alone is not clear enough. If there is nothing there and they see "Add" they may not know what they are adding. I like to keep the verbiage down when we can but I see what you are saying. If we want to add specific Add (which sounds like a good call) we would like to help create the copy. I think we could do this after HLD? Could we use the list that was generated in the subtask to do that or do you have a lookup table somewhere?
> 
> I would say we should offer this option to include context to both Add and Remove. Then we can decide if it is needed.
> 
> Thanks! @debsmita1

Well, it's definitely your call @beaumorley. My preference is the Story was to focus on the tooltip for Remove and we expanded it to truncate the "Add", which imo is not a good idea.

The PR looks fine from a technical point of view, I just don't believe we needed to touch the "Add" _context_ text that helped give the add button definitive meaning. 

/lgtm

We can address it after the fact for sure as one-off bugs or whatever as the real issues arise.

andrewballantyne:
/assign @christianvogt 

Need your approval here.

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
christianvogt: @beaumorley the above screen from Andrew looks odd.

In other places we follow this pattern where there is secondary text saying there are no _list items_:
image


If we want to follow the same, we should log an issue for it.

PR #2184, opened 2019-07-26
Update Monaco editor colors

spadgett:
https://jira.coreos.com/browse/CONSOLE-1621

/assign @rhamilto 
cc @bmignano 

Screen Shot 2019-07-26 at 4 44 05 PM

PR #6001, opened 2020-07-15
Bug 1857310: Give current line number more contrast in YAML editor

jhadvig:
Didn't wanted to give the same white as background.
/assign @spadgett 

Screen:
Screenshot 2020-07-15 at 18 46 53

### DIFF
--- a/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
+++ b/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
@@ -6,11 +6,7 @@ import {
 import { getLanguageService, TextDocument } from 'yaml-language-server';
 import { openAPItoJSONSchema } from '@console/internal/module/k8s/openapi-to-json-schema';
 import { getStoredSwagger } from '@console/internal/module/k8s/swagger';
-import {
-  global_BackgroundColor_100 as lineNumberActiveForeground,
-  global_BackgroundColor_200 as lineNumberForeground,
-  global_BackgroundColor_dark_100 as editorBackground,
-} from '@patternfly/react-tokens';
+import { global_BackgroundColor_dark_100 as editorBackground } from '@patternfly/react-tokens';
 
 window.monaco.editor.defineTheme('console', {
   base: 'vs-dark',
@@ -25,8 +21,8 @@ window.monaco.editor.defineTheme('console', {
   colors: {
     'editor.background': editorBackground.value,
     'editorGutter.background': '#292e34', // no pf token defined
-    'editorLineNumber.activeForeground': lineNumberActiveForeground.value,
-    'editorLineNumber.foreground': lineNumberForeground.value,
+    'editorLineNumber.activeForeground': '#fff',
+    'editorLineNumber.foreground': '#f0f0f0',
   },
 });
 


frontend/public/components/_edit-yaml.scss
```diff
@@ -51,5 +51,5 @@
 
 // This can't be customized in a theme, so use CSS.
 .yaml-editor .monaco-editor .margin .current-line {
-  background-color: var(--pf-global--BackgroundColor--dark-200);
+  background-color: var(--pf-global--BackgroundColor--light-300);
```
spadgett, line 54: @jhadvig Can you check if this was a problem prior to the PF upgrade? If not, we should try to use the same colors as before.

spadgett:
It looks like the number should be white instead. See screenshots from #2184

cc @bmignano

frontend/public/components/_edit-yaml.scss
jhadvig: Bgcolor + color of the font prior of the PF upgrade:
Screenshot 2020-07-16 at 11 25 04

```css
// Taken from _edit-yaml.scss:48
.yaml-editor .monaco-editor .margin .current-line {
    background-color: var(--pf-global--BackgroundColor--dark-200);
}
// Taken from <style>
.monaco-editor .current-line ~ .line-numbers {
    color: #ffffff;
}
```

Styling in the 4.5 (styling is the same):
Screenshot 2020-07-16 at 11 31 37

jhadvig:
@spadgett the font color is set to #151515 in the `monaco-colors` stylesheet

spadgett:
@jhadvig To be clear, I'm seeing black text on a dark gray background in current master. The fix should be go to back to white on dark gray as in the screenshots you posted above.

Screen Shot 2020-07-16 at 9 16 18 AM

jhadvig:
@spadgett yeah, I understand that. Will add the additional styling to the `_edit-yaml.scss`

jhadvig:
@spadgett found out that the line numbers are set programmatically and this was not caused by PF upgrade.
PTAL

frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
```diff
@@ -7,8 +7,8 @@ import { getLanguageService, TextDocument } from 'yaml-language-server';
 import { openAPItoJSONSchema } from '@console/internal/module/k8s/openapi-to-json-schema';
 import { getStoredSwagger } from '@console/internal/module/k8s/swagger';
 import {
-  global_BackgroundColor_100 as lineNumberActiveForeground,
```
spadgett, line 10: Hm, I would expect `global_BackgroundColor_100` to still be white based on https://patternfly-pr-3139.surge.sh/documentation/overview/global-css-variables. Is this a Patternfly bug?

@jcaianirh @dtaylor113

dtaylor113: I'm seeing it defined in: `frontend/node_modules/@patternfly/react-tokens/dist/js/global_BackgroundColor_100.d.ts`
```
export const global_BackgroundColor_100: {
  "name": "--pf-global--BackgroundColor--100",
  "value": "#151515",
  "var": "var(--pf-global--BackgroundColor--100)"
};
export default global_BackgroundColor_100
```
#151515 is a black color

dtaylor113: In `https://patternfly-pr-3139.surge.sh/documentation/overview/global-css-variables#global-css-variables` it is defined as `--pf-global--BackgroundColor--dark-100`.  I do see a 'light' theme, ie:  `--pf-global--BackgroundColor--light-100`...`...200` etc..

spadgett: I'm still confused because `--pf-global--BackgroundColor--100` is defined as `#fff` on https://patternfly-pr-3139.surge.sh/documentation/overview/global-css-variables

Screen Shot 2020-07-20 at 9 56 39 AM

redallen: [Core is the source of truth](https://github.com/patternfly/patternfly/blob/master/src/patternfly/sass-utilities/scss-variables.scss#L26) for this which defines `$pf-global--BackgroundColor--100: $pf-color-white !default;` where `$pf-color-white: #fff !default;`. Our documentation site shows the correct value.


Our react-token `global_BackgroundColor_100.js` is wrong (and has been wrong) for reasons @jschuler can explain. We recommend using `@patternfly/react-tokens/dist/js/patternfly_variables.js` which has the correct value `#fff`.

jschuler: Unfortunately some of the generated tokens have wrong values, i've filed a bug here https://github.com/patternfly/patternfly-react/issues/4583.
The reason for the bug is that variables are sometimes redefined in components, and the token generation script wrongly keeps the last value in finds in the patternfly scss files. 
The `patternfly_variables` file has the correct values but it contains all the global vars so is larger.

bmignano:
Do we still need to make an update here? The last screenshot @jhadvig posted looks right to me! If it's helpful, I think the CSS PF variable for white text is `--pf-global--Color--light-100`.

spadgett:
> Do we still need to make an update here? The last screenshot @jhadvig posted looks right to me!

Yeah, the last screenshot doesn't match what you see when you install 4.6. We still need a fix.

frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
spadgett: I'd be OK switching to `BackgroundColor--light-100` as well. I just wanted to check whether there was a PF bug here so we can file it upstream. Thanks @redallen and @jschuler !

frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
```diff
@@ -25,8 +21,8 @@ window.monaco.editor.defineTheme('console', {
   colors: {
     'editor.background': editorBackground.value,
     'editorGutter.background': '#292e34', // no pf token defined
-    'editorLineNumber.activeForeground': lineNumberActiveForeground.value,
-    'editorLineNumber.foreground': lineNumberForeground.value,
+    'editorLineNumber.activeForeground': '#fff',
+    'editorLineNumber.foreground': '#fff',
```
spadgett, line 25: This was previously a different value (`#f0f0f0` i believe)

jhadvig:
@spadgett comment addressed

jcaianirh:
FYI: https://github.com/patternfly/patternfly-react/issues/4583

PR #6002, opened 2020-07-15
use value as default dependency array for usePreviousHook

rawagner:


frontend/packages/console-shared/src/hooks/previous.ts
```diff
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
 
-export const usePrevious = <P = any>(value: P, deps: any[] = []): P =>
+export const usePrevious = <P = any>(value: P, deps: any[] = [value]): P =>
```
spadgett, line 3: Maybe I'm misunderstaning, but won't it always return the current value now instead of the previous?

rawagner: yes, this actually does not make sense. Sorry i guess its too late for me today :)

PR #6009, opened 2020-07-16
Add capability to collapse all grouping in topology view

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4316

**Description**: 
As a developer, I want to view all my “Grouped Resources” (Application Groups, Helm Releases, Knative Services, Operator Backed Services)as collapsed groupings in the Topology Graph View

image

image

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

/kind feature

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/filters/FilterDropdown.tsx
+++ b/frontend/packages/dev-console/src/components/topology/filters/FilterDropdown.tsx
@@ -1,16 +1,25 @@
 import * as React from 'react';
-import { Select, SelectGroup, SelectOption, SelectVariant } from '@patternfly/react-core';
+import { Select, SelectGroup, SelectOption, SelectVariant, Switch } from '@patternfly/react-core';
 import { TopologyDisplayFilterType, DisplayFilters } from '../topology-types';
+import { EXPAND_GROUPS_FILTER_ID } from './const';
+
+import './FilterDropdown.scss';
 
 type FilterDropdownProps = {
   filters: DisplayFilters;
   supportedFilters: string[];
   onChange: (filter: DisplayFilters) => void;
+  opened?: boolean; // Use only for testing
 };
 
-const FilterDropdown: React.FC<FilterDropdownProps> = ({ filters, supportedFilters, onChange }) => {
-  const [isOpen, setIsOpen] = React.useState(false);
-  const selected = filters.filter((f) => f.value).map((f) => f.id);
+const FilterDropdown: React.FC<FilterDropdownProps> = ({
+  filters,
+  supportedFilters,
+  onChange,
+  opened = false,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(opened);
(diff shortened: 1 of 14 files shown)


christianvogt:
@beaumorley is there a design for this feature somewhere?

I don't like how the filters become disabled. Furthermore there was discussion in the past about adding individual expand / collapses buttons on the nodes themselves which would conflict with what's shown here.

Giving the user the ability to expand or collapse all and then modify from that state would make sense.

jeff-phillips-18:
@christianvogt https://docs.google.com/document/d/1xVBWoaBDDAqzgD18SrCt4GZZiSvznSkTmzGsEpILmJM

frontend/packages/dev-console/src/components/topology/filters/FilterDropdown.tsx
```diff
@@ -1,76 +1,101 @@
...
+  const showFilters = filters
+    .filter((f) => f.type === TopologyDisplayFilterType.show && supportedFilters.includes(f.id))
+    .sort((a, b) => a.priority - b.priority);
+
+  const selectContent = (
+    <div className="odc-topology-filter-dropdown">
+      {expandFilters.length && (
+        <div className="odc-topology-filter-dropdown__group">
+          <span
+            key="expand-groups-switch"
```
christianvogt, line 49: key isn't required here.

frontend/packages/dev-console/src/components/topology/filters/FilterDropdown.tsx
```diff
@@ -1,76 +1,101 @@
...
+            className="odc-topology-filter-dropdown__expand-groups-switcher"
+          >
+            <span className="pf-c-select__menu-group-title">Expand</span>
+            <Switch
+              id="groups-collapse-switch"
```
christianvogt, line 54: Remove `id` attribute.

frontend/packages/dev-console/src/components/topology/topology-types.ts
```diff
@@ -42,7 +42,11 @@ export type TopologyDisplayOption = {
 export type DisplayFilters = TopologyDisplayOption[];
 
 // Applies the filters on the model and returns the ids of filters that were relevant
-export type TopologyApplyDisplayOptions = (model: Model, filters: DisplayFilters) => string[];
+export type TopologyApplyDisplayOptions = (
+  model: Model,
+  filters: DisplayFilters,
+  groupsExpanded?: boolean,
```
christianvogt, line 48: Why isn't `groupsExpanded` just another type of display filter?
I was thinking that any display options could be handled the same way.

Then hide the usage behind a single hook which can fetch the expand display filter and only if the expand filter is `true` does it fetch the individual expand filter.
```
const isExpanded = useExpand(MY_DISPLAYFILTER);
```

PR #6010, opened 2020-07-16
Add Prometheus hook to listen for multiple queries

bipuladh:


frontend/packages/console-shared/src/components/dashboard/utilization-card/prometheus-hook.ts
```diff
@@ -35,4 +35,39 @@ export const usePrometheusQuery: UsePrometheusQuery = (query, humanize) => {
...
+export const usePrometheusQueries: UsePrometheusQueries = (queries, humanize) => {
+  const dispatch = useDispatch();
+  React.useEffect(() => {
+    queries.forEach((query) => dispatch(watchPrometheusQuery(query)));
+    return () => {
+      queries.forEach((query) => dispatch(stopWatchPrometheusQuery(query)));
+    };
+  }, [dispatch, queries]);
+
+  const queryResults = useSelector<RootState, ImmutableMap<string, any>>(({ dashboards }) => {
```
kyoto, line 47: It looks to me like this will cause unnecessary re-rendering in components.

This `useSelector` will return a new object each time it runs even if no data has changed. That in turn will trigger a new `results` object to be created and returned from the hook, which will likely cause the component to re-render.

Also, whenever a single query runs, a new object will be returned, likely causing a full re-render of all values (perhaps even if no value actually changed).

If a component has several queries, it's probably easier for that component to manage its performance with several separate `usePrometheusQuery` calls rather than a single `usePrometheusQueries` call.

Maybe this is acceptable? Or maybe I'm missing something?

bipuladh: Yeah, I realized this after using it in one of my components. I have made a few changes around the mechanism of how we are selecting PTAL.

frontend/packages/console-shared/src/components/dashboard/utilization-card/prometheus-hook.ts
```diff
@@ -35,4 +35,39 @@ export const usePrometheusQuery: UsePrometheusQuery = (query, humanize) => {
...
+
+  const results = React.useMemo<UsePrometheusQueriesResult>(() => {
+    if (!queryResults) {
+      return queries.reduce(
+        (acc, query) => (acc[query] = [{}, null, null]),
+        {},
+      ) as UsePrometheusQueriesResult;
+    }
+    return Object.keys(queryResults).reduce((acc, curr) => {
+      const value = getInstantVectorStats(queryResults?.[curr]?.get('data'))[0]?.y;
```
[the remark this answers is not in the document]
bipuladh: That would be a solution as well, but looking into a lot of components in Ceph and Noobaa plugin they need the whole DataPoint so changing this implementation.

frontend/packages/console-shared/src/components/dashboard/utilization-card/prometheus-hook.ts
```diff
@@ -35,4 +37,41 @@ export const usePrometheusQuery: UsePrometheusQuery = (query, humanize) => {
...
+  );
+
+  const querySelector = React.useMemo(() => customSelectorCreator(selectors, (...data) => data), [
+    selectors,
+  ]);
+
+  const queryResults = useSelector<RootState, ImmutableMap<string, any>>(querySelector);
+
+  const results = React.useMemo<UsePrometheusQueriesResult>(() => {
+    if (_.isEmpty(queryResults?.[0])) {
```
rawagner, line 63: I think we should return `[null, null]` for not yet loaded queries. The current implementation would return something like `[[], null]`

Maybe like this ?
```javascript
if (_.isEmpty(queryResults)) {
  return queries.map(() => [null, null]) as UsePrometheusQueriesResult;
}

return queryResults.reduce((acc, curr) => {
  const data = curr.get('data');
  const value = data ? getInstantVectorStats(data, metric) : null;
  return [...acc, [value, curr.get('loadError')]];
}, []);
```

bipuladh: as disucssed making changes to align with useWatchK8sResource `[data, loading, loadError][]`

frontend/packages/console-shared/src/components/dashboard/utilization-card/prometheus-hook.ts
```diff
@@ -35,4 +38,41 @@ export const usePrometheusQuery: UsePrometheusQuery = (query, humanize) => {
...
+      return queries.map(() => [null, true, null]) as UsePrometheusQueriesResult;
+    }
+    return queryResults.reduce((acc, curr) => {
+      const value = getInstantVectorStats(curr.get('data'), metric);
+      return [...acc, [value, false, curr.get('loadError')]];
```
rawagner, line 69: you need to check if the data are set, if not then the query is still pending
```suggestion
      return [...acc, [value, !curr.get('data'), curr.get('loadError')]];
```

rawagner:
/approve

I want to give a chance to @kyoto to look at the PR again and give the final lgtm

frontend/packages/console-shared/src/components/dashboard/utilization-card/prometheus-hook.ts
```diff
@@ -35,4 +38,42 @@ export const usePrometheusQuery: UsePrometheusQuery = (query, humanize) => {
...
+
+  const queryResults = useSelector<RootState, ImmutableMap<string, any>>(querySelector);
+
+  const results = React.useMemo<UsePrometheusQueriesResult>(() => {
+    if (_.isEmpty(queryResults?.[0])) {
+      return queries.map(() => [null, true, null]) as UsePrometheusQueriesResult;
+    }
+    return queryResults.reduce((acc, curr) => {
+      const data = curr.get('data');
+      const value = getInstantVectorStats(curr.get('data'), metric);
```
kyoto, line 69: nit: You can use `data` instead of `curr.get('data')` here

kyoto:
Thanks @bipuladh, looks good!
/lgtm

PR #6014, opened 2020-07-17
[feat]:Add disk inventory to Nodes View

afreen23:
- LSO operator allows discovery of the disks present on the node.
- The discovered devices comes in the result for localvolumediscoveryresult CR
- Adds a Disk nav tab to the exiting nodes view

Implements https://issues.redhat.com/browse/RHSTOR-1195

cc @yuvalgalanti 
Screenshot from 2020-07-17 05-33-17

### DIFF
--- a/frontend/public/components/factory/table-filters.ts
+++ b/frontend/public/components/factory/table-filters.ts
@@ -265,6 +265,7 @@ export const tableFilters: TableFilterMap = {
     const status = volumeSnapshotStatus(snapshot);
     return statuses.selected.has(status) || !_.includes(statuses.all, status);
   },
+  'node-disk-name': (name, disks) => fuzzyCaseInsensitive(name, disks?.path),
 };
 
 export interface TableFilterGroups {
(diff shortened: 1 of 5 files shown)


frontend/public/components/factory/table-filters.ts
```diff
@@ -29,7 +29,7 @@ export const fuzzyCaseInsensitive = (a: string, b: string): boolean =>
 
 // TODO: Table filters are undocumented, stringly-typed, and non-obvious. We can change that.
 export const tableFilters: TableFilterMap = {
-  name: (filter, obj) => fuzzyCaseInsensitive(filter, obj.metadata.name),
+  name: (filter, obj) => fuzzyCaseInsensitive(filter, obj.metadata?.name || obj.path),
```
afreen23, line 32: The data source for this list is a not a K8sResource, hence this is failing the filter by name

frontend/public/components/factory/table-filters.ts
```diff
@@ -265,6 +265,7 @@ export const tableFilters: TableFilterMap = {
     const status = volumeSnapshotStatus(snapshot);
     return statuses.selected.has(status) || !_.includes(statuses.all, status);
   },
+  'node-disk-name': (name, disks) => fuzzyCaseInsensitive(name, disks?.path),
```
afreen23, line 268: This filter is needed on lso and ocs pages as well, hence added in all filters.
Also found no other place to contribute.  May be passing a filter function in table or list component would help as well.
cc @rawagner

PR #6015, opened 2020-07-17
Bug 1857809: Resource Quota title is duplicated

jhadvig:
/assign @rhamilto

### DIFF
--- a/frontend/public/components/resource-quota.jsx
+++ b/frontend/public/components/resource-quota.jsx
@@ -384,53 +384,56 @@ export const quotaType = (quota) => {
 // Split each resource quota into one row per subject
 export const flatten = (resources) => _.flatMap(resources, (resource) => _.compact(resource.data));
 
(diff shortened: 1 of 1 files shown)


rhamilto:
This change results in the standalone Resource Quotas page losing its title.  
Screen Shot 2020-07-17 at 7 23 44 AM

frontend/public/components/resource-quota.jsx
```diff
@@ -428,6 +428,7 @@ export const ResourceQuotasPage = connectToFlags(FLAGS.OPENSHIFT)(({ namespace,
       title="Resource Quotas"
       rowFilters={rowFilters}
       mock={mock}
+      showTitle={false}
```
rhamilto, line 431:
```suggestion
      showTitle={showTitle}
```

Then change 387 as follows:
```
export const ResourceQuotasPage = connectToFlags(FLAGS.OPENSHIFT)(
  ({ namespace, flags, mock, showTitle = true }) => {
```

jhadvig:
@rhamilto comment addressed. Thanks!

PR #6016, opened 2020-07-17
Node CSR approval

rawagner:
Client CSR approval
Screenshot from 2020-07-23 09-59-49

Server CSR approval (different node than client csr)
Screenshot from 2020-07-23 09-59-57

Filters
Screenshot from 2020-07-23 09-59-29

### DIFF
--- a/frontend/packages/console-app/src/plugin.tsx
+++ b/frontend/packages/console-app/src/plugin.tsx
@@ -30,11 +30,6 @@ import {
 } from '@console/internal/models';
 import { referenceForModel, ClusterOperator } from '@console/internal/module/k8s';
 import {
-  getNodeStatusGroups,
-  getPodStatusGroups,
-  getPVCStatusGroups,
-} from '@console/shared/src/components/dashboard/inventory-card/utils';
-import {
   fetchK8sHealth,
   getK8sHealthState,
   getControlPlaneHealth,
@@ -150,14 +145,20 @@ const plugin: Plugin<ConsumedExtensions> = [
     type: 'Dashboards/Overview/Inventory/Item',
     properties: {
       model: NodeModel,
-      mapper: getNodeStatusGroups,
+      mapper: () =>
+        import(
+          '@console/shared/src/components/dashboard/inventory-card/utils' /* webpackChunkName: "console-app" */
+        ).then((m) => m.getNodeStatusGroups),
     },
   },
   {
     type: 'Dashboards/Overview/Inventory/Item',
     properties: {
       model: PodModel,
-      mapper: getPodStatusGroups,
+      mapper: () =>
+        import(
+          '@console/shared/src/components/dashboard/inventory-card/utils' /* webpackChunkName: "console-app" */
+        ).then((m) => m.getPodStatusGroups),
     },
   },
   {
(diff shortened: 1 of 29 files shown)


rawagner:
@andybraren I've updated the screenshots to match the latest designs. Is the Server CSR approval popup ok ? Do we want to improve the description ? Here's a list of what does not work when server csr is not approved
```
any communication to kubelets
metrics
pod logs
pod exec
service proxy
pod port-forward
I think I remember a build thingy using it
must-gather
```

rawagner:
Another question for filtering. Do we want to have a different filter category for Discovered and Server CSR pending nodes ? Or we just push them into one ie `Certificate pending` category ?

rawagner:
> Looking good!
> 
> The designs include `Approval required` substatus text below `Discovered` and `Ready` to more clearly indicate that an action is required. Do you think that substatus text would be valuable to include?
> 

I forgot about it :) Substatus added.

> 
> > Is the Server CSR approval popup ok ?
> 
> I've suggested a slight tweak below, but I wonder if it's accurate. Networking stuff seems related to most of the features you mentioned. Have you found any good documentation on Server CSRs that I could read to help inform this text? It seems like there's so little out there... 
> 

Im not sure, I did not find any good docs :/ I've updated the popup text. 

> > Another question for filtering. Do we want to have a different filter category for Discovered and Server CSR pending nodes ? Or we just push them into one ie Certificate pending category ?
> 
> Good question. One category seems natural since they're so related, and if we add substatus text a filter of `Approval Required` would match the text that appears in the Status column. This is probably the first time that we'd have a filter for a substatus, but this approach seems to make the most sense to me. WDYT @spadgett?
> 

+1

frontend/packages/console-shared/src/components/status/StatusIconAndText.tsx
```diff
@@ -16,14 +19,15 @@ const StatusIconAndText: React.FC<StatusIconAndTextProps> = ({
   iconOnly,
   noTooltip,
   className,
+  description,
```
jtomasek, line 22: I think it would be better to use `<SecondaryStatus>` to add the description to the status. It is a concept which is already used and available (see NodeStatus.tsx for example). Your change is introducing another implementation of the same concept which I think we should avoid.

Is your reasoning for doing this to have the icon vertically aligned evenly among the status and description? I'd rather avoid this as it is not used anywhere else. Current usage in console is:
```
 ----------------
| <icon><status> |
| <substatus>    |
 ----------------
```

rawagner: Thats exactly the reason. Overview pages also use vertically aligned icon and we have the same in UXD mockups. So at some point we should start aligning to just one style.

rawagner:
Removed the vertically aligned icon after discussion with @jtomasek  - this aligns with other statuses on list pages
Screenshot from 2020-07-24 14-26-28

@andybraren we should figure out one style for statuses in List and Overview pages and align them in next release.

andybraren:
> @andybraren we should figure out one style for statuses in List and Overview pages and align them in next release.

FYI @abichen, in the new Overview design we may want to consider repositioning substatus text to start below the status icon rather than below the primary status text. That seems to be what we do in List pages at the moment, provides a bit more room for substatuses to be longer, and would allow code to be reused.

rawagner:
/hold

main bundle is too big

rawagner:
/hold cancel

bundle got too big because of new dependencies. Solved by lazy loading the inventory status group mappers.

PR #6020, opened 2020-07-17
Hide internal operators

jhadvig:
Based on https://issues.redhat.com/browse/CONSOLE-2293

Hiding **non-standalone** operators from the Operator Hub and Installed Operators pages.

For the Operator Hub we should be checking for the `currentCSVDesc` annotations for the `operators.operatorframework.io/operator-type: 'non-standalone'`, which indicates a non-standalone operator.
For the Installed Operators list page we are hiding:
- **CSV** that contain the provided annotation and are in *Failed* phase
- **Subscriptions**, for which are we matching a packageManifest and checking its `currentCSVDesc` annotations, and the Subscription is in the *UpgradeFailed* state.

/assign @spadgett

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
@@ -78,6 +78,7 @@ import {
   PackageManifestKind,
   SubscriptionKind,
 } from '../types';
+import { operatorTypeAnnotation, nonStandardAnnotationValue } from '../const';
 import { subscriptionForCSV, getSubscriptionStatus } from '../status/csv-status';
 import { getInternalObjects, isInternalObject } from '../utils';
 import { ProvidedAPIsPage, ProvidedAPIPage } from './operand';
@@ -517,19 +518,33 @@ export const NamespacedClusterServiceVersionList: React.SFC<ClusterServiceVersio
   );
   const NoDataEmptyMsg = () => <MsgBox title="No Operators Found" detail={noDataDetail} />;
 
-  const isCopiedCSV = (source, kind) => {
+  const isCopiedCSV = (source: ClusterServiceVersionKind, kind: string) => {
     return (
       referenceForModel(ClusterServiceVersionModel) === kind && source.status?.reason === 'Copied'
     );
   };
 
-  const removeCopiedCSVs = (operators: (ClusterServiceVersionKind | SubscriptionKind)[]) => {
+  const isStandardCSV = (operator: ClusterServiceVersionKind) => {
+    return (
+      operator.metadata.annotations?.[operatorTypeAnnotation] !== nonStandardAnnotationValue ||
+      operator.status?.phase === ClusterServiceVersionPhase.CSVPhaseFailed
+    );
+  };
+
+  const filterOperators = (
+    operators: (ClusterServiceVersionKind | SubscriptionKind)[],
+    allNamespaceActive: boolean,
+  ): (ClusterServiceVersionKind | SubscriptionKind)[] => {
     return operators.filter((source) => {
       const kind = referenceFor(source);
-      if (isSubscription(kind)) {
+      if (isSubscription(source)) {
         return true;
       }
-      return !isCopiedCSV(source, kind);
+      const csv = source as ClusterServiceVersionKind;
+      if (allNamespaceActive) {
+        return !isCopiedCSV(csv, kind) || isStandardCSV(csv);
+      }
+      return isStandardCSV(csv);
     });
   };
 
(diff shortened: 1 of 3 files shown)


frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -674,6 +684,11 @@ export const ClusterServiceVersionsPage: React.FC<ClusterServiceVersionsPageProp
             prop: 'catalogSources',
             optional: true,
           },
+          {
+            kind: referenceForModel(PackageManifestModel),
```
spadgett, line 688: Package manifests are expensive to watch. It's a fair amount of data and could noticeably slow down page load.

Is this simply to hide Subscriptions when there isn't a CSV yet? We should discuss if it's worth the extra complexity. It's a transient state (unless there is a problem creating the CSV, in which case you probably want to see the Subscription).

We might step back and look at this generally (as a separate PR). The experience is a little weird when installing an operator since the Subscription appears briefly and disappears. In some cases, there are briefly two rows for the same operator. @tlwu2013 can probably give more background on this. I assume showing the Subscription is for troubleshooting when the CSV can't be created.

tlwu2013: > Is this simply to hide Subscriptions when there isn't a CSV yet? 

The intention here is not for "hiding Subscriptions when there isn't a CSV yet" (but we do have another bug tracking that issue, but that one probably doesn't need to check `PackageManifest`object. And agree that PR could be a generic fix like you suggested).

The idea is "if the Operator has been marked as 'non-standalone' in its CSV file, then UI **doesn't show** Subscription object **unless** in error". Basically allowing `Subscription` object to mimic the "hidden" effect for the `CSV`object as well, so users won't see the those 'non-standalone' Operator showed up for certain period of time during the installation process (because of the `Subscription` object) and then be hidden afterward.

@spadgett in this case, is there another suggestion instead of like we thought here using the status stanza of `PackageManifest` object to get the annotation in the CSV file?

jhadvig: @spadgett @tlwu2013 so in term of the PR we want (in the Installed Operators page) to hide all the Subscriptions and only show those that are in one of [these](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/github.com/operator-framework/api/pkg/operators/subscription_types.go#L17-L20) states. Right? That way we can remove listing on Package Manifests. Am I missing something ?

jhadvig:
@spadgett I've updated the PR so the filtering only is for CSV.

cc'ing @tlwu2013

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -517,19 +520,32 @@ export const NamespacedClusterServiceVersionList: React.SFC<ClusterServiceVersio
...
       const kind = referenceFor(source);
-      if (isSubscription(kind)) {
+      if (isSubscription(source)) {
         return true;
       }
-      return !isCopiedCSV(source, kind);
+      if (allNamespaceActive) {
+        return !(isCopiedCSV(source, kind) || isFailedNonStandardCSV(source));
+      }
+      return !isFailedNonStandardCSV(source);
```
spadgett, line 548: Are only filtering out non-standard CSVs that are failed? The logic should be to filter non-standard CSVs that are *not* failed.

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx
```diff
@@ -50,16 +51,20 @@ export const OperatorHubList: React.SFC<OperatorHubListProps> = (props) => {
...
-        const { defaultChannel } = getPackageStatus(pkg);
         if (!defaultChannel) {
           // eslint-disable-next-line no-console
           console.warn(
             `PackageManifest ${pkg.metadata.name} has no status.defaultChannel and has been excluded`,
           );
           return false;
         }
-        return true;
+        const { currentCSVDesc } = _.find(channels || [], { name: defaultChannel } as any);
```
spadgett, line 63: Can you use `Array.find` to avoid needing `as`?

jhadvig:
@spadgett comments addressed. PTAL

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -517,19 +518,32 @@ export const NamespacedClusterServiceVersionList: React.SFC<ClusterServiceVersio
...
     return (
       referenceForModel(ClusterServiceVersionModel) === kind && source.status?.reason === 'Copied'
     );
   };
 
-  const removeCopiedCSVs = (operators: (ClusterServiceVersionKind | SubscriptionKind)[]) => {
+  const isStandardCSV = (operator: ClusterServiceVersionKind | SubscriptionKind) => {
+    return (
+      operator.metadata.annotations?.[operatorTypeAnnotation] !== nonStandardAnnotationValue &&
+      operator.status?.phase !== ClusterServiceVersionPhase.CSVPhaseFailed
```
spadgett, line 530: Won't this hide any failed operator from the list? We want to hide CSVs with the annotation unless they're failed.

```suggestion
      operator.metadata.annotations?.[operatorTypeAnnotation] !== nonStandardAnnotationValue ||
      operator.status?.phase === ClusterServiceVersionPhase.CSVPhaseFailed
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -517,19 +518,32 @@ export const NamespacedClusterServiceVersionList: React.SFC<ClusterServiceVersio
...
 
-  const isCopiedCSV = (source, kind) => {
+  const isCopiedCSV = (source: ClusterServiceVersionKind | SubscriptionKind, kind: string) => {
     return (
       referenceForModel(ClusterServiceVersionModel) === kind && source.status?.reason === 'Copied'
     );
   };
 
-  const removeCopiedCSVs = (operators: (ClusterServiceVersionKind | SubscriptionKind)[]) => {
+  const isStandardCSV = (operator: ClusterServiceVersionKind | SubscriptionKind) => {
```
spadgett, line 527: This should always be a CSV and not a Subscription, right?

```suggestion
  const isStandardCSV = (operator: ClusterServiceVersionKind) => {
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -559,37 +573,14 @@ export const NamespacedClusterServiceVersionList: React.SFC<ClusterServiceVersio
     const olmOperatorNamespace = obj.metadata?.annotations?.['olm.operatorNamespace'];
     return olmOperatorNamespace ?? getNamespace(obj);
   };
+  const allNamespaceActive = activeNamespace === ALL_NAMESPACES_KEY;
 
-  return activeNamespace === ALL_NAMESPACES_KEY ? (
```
spadgett, line 563: Thank you for cleaning this up 👍

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
jhadvig: is should.. but then the func needs to be call with `isStandardCSV(source as ClusterServiceVersionKind)` since the the source is from `(ClusterServiceVersionKind | SubscriptionKind)[]`

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
jhadvig: So I was in the impression that we want to hide only

jhadvig:
@spadgett thanks for the logic adjustments! PR updated

PR #6021, opened 2020-07-17
Bug 1857821: Fix for pod ring exception

rebeccaalpert:
This a manual backport for a fix used in 4.6 to correct a pod ring exception. This was caused by a lack of handling for a bad pods parameter in getFailedPods.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1857821.

### DIFF
--- a/frontend/packages/console-shared/src/utils/pod-ring-utils.ts
+++ b/frontend/packages/console-shared/src/utils/pod-ring-utils.ts
@@ -76,6 +76,10 @@ const isPendingPods = (
   (!currentPodCount && !!desiredPodCount);
 
 export const getFailedPods = (pods: ExtPodKind[]): number => {
+  if (!pods?.length) {
+    return 0;
+  }
+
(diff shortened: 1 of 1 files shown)


frontend/packages/console-shared/src/utils/pod-ring-utils.ts
```diff
@@ -76,6 +76,10 @@ const isPendingPods = (
   (!currentPodCount && !!desiredPodCount);
 
 export const getFailedPods = (pods: ExtPodKind[]): number => {
+  if (!pods.length) {
```
spadgett, line 79: This will have the same runtime error as the original bug because `pods` is undefined or null.

```suggestion
  if (!pods?.length) {
```

rebeccaalpert: Thanks, I can't read apparently. I'll fix it.

PR #5784, opened 2020-06-18
Label resources owned by operators

rebeccaalpert:
Added web socket for keeping track of CSVs and resource label described in design document. I moved some logic in the ResourceLink component so that URLs are consistently generated for the owner of a given resource. Per @tlwu2013, the Owner link should match the label link. The link in the Owner section of the details page (if present) will now always have a link that matches the label. This should also affect anything using the OwnerReferences component (list views, etc.).

Desktop:
Screen Shot 2020-07-15 at 2 24 40 PM

Mobile:
Screen Shot 2020-07-15 at 2 24 49 PM

Fixes https://issues.redhat.com/browse/CONSOLE-2288.

Heads up @itsptk and @openshift/team-ux-review.

PR #6023, opened 2020-07-17
Add warnings for managed resources

rebeccaalpert:
I added a warning modal for when users save a managed resource. I also edited the delete modal so it has an alert if the resource is managed. I have a separate PR for the managed resources label, but this PR shares code (first commit) in order to get the modals to work. The other PR is https://github.com/openshift/console/pull/5784. I'll drop that commit once the prior PR is merged.

Screen Shot 2020-07-17 at 2 09 56 PM
Screen Shot 2020-07-17 at 2 10 16 PM

Fixes https://issues.redhat.com/projects/CONSOLE/issues/CONSOLE-2163.

Heads up @openshift/team-ux-review and @itsptk.

### DIFF
--- a/frontend/public/components/edit-yaml.jsx
+++ b/frontend/public/components/edit-yaml.jsx
@@ -12,7 +12,7 @@ import YAMLEditor from '@console/shared/src/components/editor/YAMLEditor';
 import { isYAMLTemplate, withExtensions } from '@console/plugin-sdk';
 
 import { connectToFlags } from '../reducers/features';
-import { errorModal } from './modals';
+import { errorModal, managedResourceSaveModal } from './modals';
 import { Firehose, checkAccess, history, Loading, resourceObjPath } from './utils';
 import {
   referenceForModel,
@@ -25,7 +25,9 @@ import { ConsoleYAMLSampleModel } from '../models';
 import { getResourceSidebarSamples } from './sidebars/resource-sidebar-samples';
 import { ResourceSidebar } from './sidebars/resource-sidebar';
 import { getYAMLTemplates } from '../models/yaml-templates';
-
+import { findOwner } from '../module/k8s/managed-by';
+import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager/src/models';
+import { k8sList } from '../module/k8s/resource';
 import { definitionFor } from '../module/k8s/swagger';
 
 const generateObjToLoad = (templateExtensions, kind, id, yaml, namespace = 'default') => {
@@ -61,6 +63,7 @@ export const EditYAML_ = connect(stateToProps)(
           sampleObj: props.sampleObj,
           fileUpload: props.fileUpload,
           showSidebar: !!props.create,
+          owner: null,
         };
         this.monacoRef = React.createRef();
         // k8s uses strings for resource versions
@@ -68,6 +71,8 @@ export const EditYAML_ = connect(stateToProps)(
         // Default cancel action is browser back navigation
         this.onCancel = 'onCancel' in props ? props.onCancel : history.goBack;
         this.downloadSampleYaml_ = this.downloadSampleYaml_.bind(this);
+        this.updateYAML = this.updateYAML.bind(this);
+        this.loadCSVs = this.loadCSVs.bind(this);
 
         if (this.props.error) {
           this.handleError(this.props.error);
@@ -86,8 +91,26 @@ export const EditYAML_ = connect(stateToProps)(
         this.setState({ error, success: null });
       }
 
+      loadCSVs() {
+        const { obj, create } = this.props;
+        const namespace = obj?.metadata?.namespace;
+        if (create || !namespace || !obj?.metadata?.ownerReferences?.length) {
+          return;
+        }
+        k8sList(ClusterServiceVersionModel, { ns: namespace })
+          .then((csvList) => {
+            const owner = findOwner(obj, csvList);
+            this.setState({ owner });
+          })
+          .catch((e) => {
+            // eslint-disable-next-line no-console
+            console.error('Could not fetch CSVs', e);
+          });
+      }
+
       componentDidMount() {
         this.loadYaml();
+        this.loadCSVs();
       }
 
       UNSAFE_componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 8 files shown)


rebeccaalpert:
```
error: some steps failed:
  * could not run steps: step e2e-gcp-console failed: template pod "e2e-gcp-console" failed: the pod ci-op-mqjlc765/e2e-gcp-console failed after 59m6s (failed containers: test): ContainerFailed one or more containers exited```

/retest

File: frontend/public/components/modals/_modals.scss [hunk 1]
```diff
@@ -51,6 +51,21 @@
...
+
+  .pf-c-alert {
+    margin-top: var(--pf-global--spacer--md);
+  }
+
+  // fixes title misalignment with icon
+  .pf-c-alert__title {
+    display: flex;
+    align-items: center;
+  }
```
spadgett, line 63: We definitely don't want to be overriding the `pf-c-alert__title` class generally. We have other modals that have an icon. Can we reuse the style from those?

frontend/public/components/modals/_modals.scss
```diff
@@ -51,6 +51,21 @@
...
+
+  // setting resource item to inline via react adds padding
+  .co-resource-item {
```
spadgett, line 66: We can't override this globally. This could be bad effects in other places.

frontend/public/components/modals/delete-modal.jsx
```diff
@@ -45,8 +53,35 @@ class DeleteModal extends PromiseComponent {
...
+    const { data } = this.state;
+    const ownerReferences = resource.metadata.ownerReferences || [];
+    const owner = data
+      ? ownerReferences.find((o) => data.some((csv) => isOwnedByOperator(csv, o)))
+      : undefined;
+    return owner;
+  }
+
+  componentDidMount() {
+    const { namespace } = this.props;
```
spadgett, line 67: You should skip listing CSVs if there is no namespace.

frontend/public/components/edit-yaml.jsx
```diff
@@ -86,8 +93,24 @@ export const EditYAML_ = connect(stateToProps)(
         this.setState({ error, success: null });
       }
 
+      loadCSVs() {
+        const { namespace } = this.props.obj.metadata;
```
spadgett, line 97: You should skip listing CSVs if there is no namespace.

This is problematic because the namespace can be edited in the YAML editor, and we won't get the updates here.

frontend/public/components/modals/delete-modal.jsx
```diff
@@ -76,6 +111,20 @@ class DeleteModal extends PromiseComponent {
...
             )}
+            {owner && (
+              <Alert variant="warning" title="Managed resource">
+                This resource is managed by{' '}
+                <ResourceLink
+                  key={owner.uid}
+                  kind={referenceForOwnerRef(owner)}
+                  name={owner.name}
+                  namespace={resource.metadata.namespace}
+                  owner={owner}
```
spadgett, line 122: Does `ResourceLink` take an owner prop?

frontend/public/components/modals/delete-modal.jsx
```diff
@@ -76,6 +111,20 @@ class DeleteModal extends PromiseComponent {
                 </label>
               </div>
             )}
+            {owner && (
+              <Alert variant="warning" title="Managed resource">
```
spadgett, line 115: Shouldn't this be an inline alert?

frontend/public/components/edit-yaml.jsx
spadgett: The namespace can only change when creating a new resource. I think we'd want to skip listing CSVs for the creation editor. (If you're creating something, it's never owned by an operator.)

frontend/public/components/modals/delete-modal.jsx
rebeccaalpert: I used it because of the styling in the design, but I just checked with @itsptk and it should be inline. Will fix.

frontend/public/components/modals/_modals.scss
```diff
@@ -51,6 +51,10 @@
   .input-group-addon {
     background-color: rgba(227, 227, 227, 0.5);
   }
+
+  .pf-c-alert {
```
spadgett, line 55: This can't be a global style change. You must scope it to this specific dialog.

frontend/public/components/utils/resource-link.tsx
```diff
@@ -80,14 +80,16 @@ export const ResourceLink = connectToModel(
     hideIcon,
     title,
     children,
+    noPadding = false,
```
spadgett, line 83: Don't do it this way. Pass in a `className` and write some CSS that targets the dialog specifically.

frontend/public/components/edit-yaml.jsx
```diff
@@ -61,13 +63,17 @@ export const EditYAML_ = connect(stateToProps)(
           sampleObj: props.sampleObj,
           fileUpload: props.fileUpload,
           showSidebar: !!props.create,
+          data: undefined,
```
spadgett, line 66: It's generally better to default to `null` when setting it explicitly (and that's what's used just above).

frontend/public/components/edit-yaml.jsx
```diff
@@ -86,8 +92,27 @@ export const EditYAML_ = connect(stateToProps)(
         this.setState({ error, success: null });
       }
 
+      loadCSVs() {
+        const { namespace } = this.props.obj.metadata;
+        if (!namespace) {
```
spadgett, line 97: Can you check if this is a creation editor as well? We should skip this if creating a new resource.

frontend/public/components/modals/_modals.scss
rebeccaalpert: Looking into this.

frontend/public/components/utils/managed-by.tsx
```diff
@@ -28,6 +28,14 @@ const matchOwnerAndCSV = (owner: OwnerReference, csvs: K8sResourceCommon[] = [])
   return csvs.find((csv) => isOwnedByOperator(csv, owner));
 };
 
+export const findOwner = (obj: K8sResourceCommon, data: ClusterServiceVersionKind[]) => {
```
spadgett, line 31: There's no reason to lazy load the dialog if we use this helper from this module. The module won't be lazy-loaded.

Can you move this into a different file under `public/module/k8s`?

spadgett:
Build errors:

```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/modals/managed-resource-save-modal.tsx
(13,20): Expected 0 arguments, but got 4.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/modals/managed-resource-save-modal.tsx
(51,10): Cannot find name 'K8sKind'.
```

rebeccaalpert:
Yup, it takes a good 20 minutes to build so I haven't even been able to confirm this works with all the feedback. I don't know that this will make the deadline.

spadgett:
> Yup, it takes a good 20 minutes to build so I haven't even been able to confirm this works with all the feedback. I don't know that this will make the deadline.

Are you running `yarn run dev`? It should take less than a minute on changes.

serenamarie125:
FYI @jeff-phillips-18 @andrewballantyne @christianvogt will be interesting to see how this affects us in topology.  I think we disallow these commands, but probably worth a look

PR #6028, opened 2020-07-20
(feat): Support multiple storage class selection during storage cluster expansion

afreen23:
- Adds the required device set request implementation for updating storage cluster on expansion
- Includes minor refactors/cleanups

https://issues.redhat.com/browse/KNIP-1459

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import * as _ from 'lodash';
 import * as classNames from 'classnames';
 import { FieldLevelHelp, humanizeBinaryBytes } from '@console/internal/components/utils/index';
 import {
@@ -10,40 +9,46 @@ import {
 } from '@console/internal/components/factory';
 import { usePrometheusPoll } from '@console/internal/components/graphs/prometheus-poll-hook';
 import { k8sPatch, StorageClassResourceKind } from '@console/internal/module/k8s';
-import { PrometheusEndpoint } from '@console/internal/components/graphs/helpers';
-import { getName } from '@console/shared';
+import { getName, getRequestedPVCSize } from '@console/shared';
 import { OCSServiceModel } from '../../../models';
 import { OSD_CAPACITY_SIZES } from '../../../utils/osd-size-dropdown';
-import { CEPH_STORAGE_NAMESPACE, NO_PROVISIONER } from '../../../constants';
-import { labelTooltip, storageClassTooltip } from '../../../constants/ocs-install';
-import { CAPACITY_USAGE_QUERIES, StorageDashboardQuery } from '../../../constants/queries';
+import { NO_PROVISIONER, OCS_DEVICE_SET_REPLICA } from '../../../constants';
+import {
+  labelTooltip,
+  storageClassTooltip,
+  defaultRequestSize,
+} from '../../../constants/ocs-install';
 import { OCSStorageClassDropdown } from '../storage-class-dropdown';
 import { PVsAvailableCapacity } from '../../ocs-install/pvs-available-capacity';
+import { createDeviceSet, DeviceSet } from '../../ocs-install/ocs-request-data';
+import { cephCapacityResource } from '../../../constants/resources';
 import './_add-capacity-modal.scss';
 
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
 
+const getCurrentDeviceSet = (deviceSets: DeviceSet[], selectedSCName: string): number =>
+  deviceSets.findIndex((ds) => ds.dataPVCTemplate.spec.storageClassName === selectedSCName);
+
 export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const { ocsConfig, close, cancel } = props;
+  const deviceSets: DeviceSet[] = ocsConfig?.spec.storageDeviceSets || [];
 
-  const [response, loadError, loading] = usePrometheusPoll({
-    endpoint: PrometheusEndpoint.QUERY,
-    namespace: CEPH_STORAGE_NAMESPACE,
-    query: CAPACITY_USAGE_QUERIES[StorageDashboardQuery.CEPH_CAPACITY_USED],
-  });
+  const [response, loadError, loading] = usePrometheusPoll(cephCapacityResource);
   const [storageClass, setStorageClass] = React.useState<StorageClassResourceKind>(null);
+  /* TBD(Afreen): Show installation storage class as preselected 
+                  Change state metadata
+  */
   const [inProgress, setProgress] = React.useState(false);
   const [errorMessage, setError] = React.useState('');
 
-  const osdSizeWithUnit = _.get(
-    ocsConfig,
-    'spec.storageDeviceSets[0].dataPVCTemplate.spec.resources.requests.storage',
-  );
-  const presentCount = _.get(ocsConfig, 'spec.storageDeviceSets[0].count');
-  const capacity = _.get(response, 'data.result[0].value[1]');
-
-  const osdSizeWithoutUnit = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
+  const cephCapacity: string = response?.data?.result?.[0]?.value[1];
+  const osdSizeWithUnit = getRequestedPVCSize(deviceSets[0].dataPVCTemplate);
+  const osdSizeWithoutUnit: number = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
   const provisionedCapacity = getProvisionedCapacity(osdSizeWithoutUnit);
+  const isNoProvionerSC: boolean = storageClass?.provisioner === NO_PROVISIONER;
+  const selectedSCName: string = getName(storageClass);
+  const deviceSetIndex: number = getCurrentDeviceSet(deviceSets, selectedSCName);
+  const replica = OCS_DEVICE_SET_REPLICA;
 
   let currentCapacity: React.ReactNode;
 
@@ -51,33 +56,45 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
     currentCapacity = (
       <div className="skeleton-text ceph-add-capacity__current-capacity--loading" />
     );
-  } else if (loadError || !capacity || !presentCount || !osdSizeWithoutUnit) {
+  } else if (loadError || !cephCapacity || !osdSizeWithoutUnit || deviceSetIndex === -1) {
     currentCapacity = <div className="text-muted">Not available</div>;
   } else {
     currentCapacity = (
       <div className="text-muted">
-        <strong>{`${humanizeBinaryBytes(capacity / 3).string} / ${presentCount *
-          osdSizeWithoutUnit} TiB`}</strong>
+        <strong>{`${humanizeBinaryBytes(Number(cephCapacity) / replica).string} / ${deviceSets[
+          deviceSetIndex
+        ].count * osdSizeWithoutUnit} TiB`}</strong>
       </div>
     );
   }
 
+  const onChange = (sc: StorageClassResourceKind) => setStorageClass(sc);
+
   const submit = (event: React.FormEvent<EventTarget>) => {
     event.preventDefault();
     setProgress(true);
-    const scName = getName(storageClass);
-    const patch = [
-      {
-        op: 'replace',
-        path: `/spec/storageDeviceSets/0/count`,
-        value: presentCount + 1,
-      },
-    ];
-    if (!scName) {
+    const patch = {
+      op: '',
+      path: '',
+      value: null,
+    };
+    const portable = !isNoProvionerSC;
+    const osdSize = isNoProvionerSC ? defaultRequestSize.BAREMETAL : osdSizeWithUnit;
+    if (deviceSetIndex === -1) {
+      patch.op = 'add';
+      patch.path = `/spec/storageDeviceSets/-`;
+      patch.value = createDeviceSet(selectedSCName, osdSize, portable);
+    } else {
+      patch.op = 'replace';
+      patch.path = `/spec/storageDeviceSets/${deviceSetIndex}/count`;
+      patch.value = deviceSets[deviceSetIndex].count + 1;
+    }
+
+    if (!selectedSCName) {
       setError('No StorageClass selected');
       setProgress(false);
     } else {
-      k8sPatch(OCSServiceModel, ocsConfig, patch)
+      k8sPatch(OCSServiceModel, ocsConfig, [patch])
         .then(() => {
           setProgress(false);
           close();
@@ -97,21 +114,20 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
         <div className="ceph-add-capacity__modal">
           <div
             className={classNames('ceph-add-capacity__sc-dropdown', {
-              'ceph-add-capacity__sc-dropdown--margin':
-                storageClass?.provisioner !== NO_PROVISIONER,
+              'ceph-add-capacity__sc-dropdown--margin': !isNoProvionerSC,
             })}
           >
             <label className="control-label" htmlFor="storageClass">
               Storage Class
               <FieldLevelHelp>{storageClassTooltip}</FieldLevelHelp>
             </label>
-            <OCSStorageClassDropdown onChange={setStorageClass} />
+            <OCSStorageClassDropdown onChange={onChange} />
           </div>
-          {storageClass?.provisioner === NO_PROVISIONER ? (
+          {isNoProvionerSC ? (
             <PVsAvailableCapacity
-              replica={ocsConfig.spec.storageDeviceSets[0].replica}
+              replica={replica}
               data-test-id="ceph-add-capacity-pvs-available-capacity"
-              sc={storageClass}
+              storageClass={storageClass}
             />
           ) : (
             <div>
@@ -129,9 +145,11 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
                   disabled
                   data-test-id="requestSize"
                 />
-                <div className="ceph-add-capacity__input--info-text">
-                  x 3 replicas = <strong>{provisionedCapacity} TiB</strong>
-                </div>
+                {provisionedCapacity && (
+                  <div className="ceph-add-capacity__input--info-text">
+                    x {replica} replicas = <strong>{provisionedCapacity} TiB</strong>
+                  </div>
+                )}
               </div>
               <div className="ceph-add-capacity__current-capacity">
                 <div className="text-secondary ceph-add-capacity__current-capacity--text">
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
@@ -20,7 +20,13 @@ import {
 import { setFlag } from '@console/internal/actions/features';
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
 import { getName } from '@console/shared';
-import { minSelectedNode, defaultRequestSize, NO_PROVISIONER } from '../../../constants';
+import {
+  minSelectedNode,
+  defaultRequestSize,
+  NO_PROVISIONER,
+  OCS_INTERNAL_CR_NAME,
+  OCS_DEVICE_SET_REPLICA,
+} from '../../../constants';
 import { OCSServiceModel } from '../../../models';
 import AttachedDevicesNodeTable from './sc-node-list';
 import { PVsAvailableCapacity } from '../pvs-available-capacity';
@@ -95,7 +101,7 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
       history.push(
         `/k8s/ns/${ns}/clusterserviceversions/${appName}/${referenceForModel(
           OCSServiceModel,
-        )}/${getName(getOCSRequestData())}`,
+        )}/${OCS_INTERNAL_CR_NAME}`,
       );
     });
   };
@@ -108,9 +114,9 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
       <Form className="co-m-pane__body-group">
         <StorageClassSection handleStorageClass={handleStorageClass} filterSC={onlyNoProvSC}>
           <PVsAvailableCapacity
-            replica={getOCSRequestData().spec.storageDeviceSets[0].replica}
+            replica={OCS_DEVICE_SET_REPLICA}
             data-test-id="ceph-ocs-install-pvs-available-capacity"
-            sc={storageClass}
+            storageClass={storageClass}
           />
         </StorageClassSection>
         <h3 className="co-m-pane__heading co-m-pane__heading--baseline ceph-ocs-install__pane--margin">
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
@@ -1,10 +1,36 @@
 import { K8sResourceKind } from '@console/internal/module/k8s';
-import { NO_PROVISIONER, OCS_INTERNAL_CR_NAME, CEPH_STORAGE_NAMESPACE } from '../../constants';
+import {
+  NO_PROVISIONER,
+  OCS_INTERNAL_CR_NAME,
+  CEPH_STORAGE_NAMESPACE,
+  OCS_DEVICE_SET_REPLICA,
+} from '../../constants';
+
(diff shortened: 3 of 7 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -15,35 +43,34 @@ export const getOCSRequestData = (
...
-            },
-          },
-        },
-      ],
+      storageDeviceSets: [createDeviceSet(scName, storage, 1, true)],
     },
   } as K8sResourceKind;
 
-  if (infra === NO_PROVISIONER) {
+  if (provisioner === NO_PROVISIONER) {
```
afreen23, line 50: need to do a corresponding change in the caller function.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,95 @@ import {
...
-  const [response, loadError, loading] = usePrometheusPoll({
-    endpoint: PrometheusEndpoint.QUERY,
-    namespace: CEPH_STORAGE_NAMESPACE,
-    query: CAPACITY_USAGE_QUERIES[StorageDashboardQuery.CEPH_CAPACITY_USED],
-  });
+  const [response, loadError, loading] = usePrometheusPoll(cephCapacityResource);
   const [storageClass, setStorageClass] = React.useState<StorageClassResourceKind>(null);
+  /* TBD(Afreen): Show installation storage class as preselected 
+                  Change state metadata
+  */
```
afreen23, line 41: As part of enhancements and refactors in a followup PR

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,94 @@ import {
...
 export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const { ocsConfig, close, cancel } = props;
+  const deviceSets: DeviceSet[] = ocsConfig?.spec?.storageDeviceSets || [];
 
-  const [response, loadError, loading] = usePrometheusPoll({
-    endpoint: PrometheusEndpoint.QUERY,
-    namespace: CEPH_STORAGE_NAMESPACE,
-    query: CAPACITY_USAGE_QUERIES[StorageDashboardQuery.CEPH_CAPACITY_USED],
-  });
+  const [response, loadError, loading] = usePrometheusPoll(cephCapacityResource);
```
gnehapk, line 36: response -> capacity/cephCapacityData

gnehapk:
Can you also attach screenshots here in this PR?

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -97,21 +116,20 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
-                storageClass?.provisioner !== NO_PROVISIONER,
+              'ceph-add-capacity__sc-dropdown--margin': !isNoProvionerSC,
             })}
           >
             <label className="control-label" htmlFor="storageClass">
               Storage Class
               <FieldLevelHelp>{storageClassTooltip}</FieldLevelHelp>
             </label>
-            <OCSStorageClassDropdown onChange={setStorageClass} />
+            <OCSStorageClassDropdown onChange={onChange} />
```
gnehapk, line 126: `setStorageClass` can be directly called here. No need to create a separate function.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -15,35 +38,34 @@ export const getOCSRequestData = (
...
-              resources: {
-                requests: {
-                  storage,
-                },
-              },
-            },
-          },
-        },
-      ],
+      storageDeviceSets: [createDeviceSet(scName, storage, true)],
```
gnehapk, line 41: Its better  to pass the `provisioner` too as parameter in `createDeviceSet` for deciding the `portable` value instead of passing just `true`. This will help in case of multiple storage device sets creation.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
```diff
@@ -108,9 +108,9 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
       <Form className="co-m-pane__body-group">
         <StorageClassSection handleStorageClass={handleStorageClass} filterSC={onlyNoProvSC}>
           <PVsAvailableCapacity
-            replica={getOCSRequestData().spec.storageDeviceSets[0].replica}
+            replica={getOCSRequestData(getName(storageClass)).spec.storageDeviceSets[0].replica}
```
gnehapk, line 111: Why do you need to pass the storage class name here?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -1,10 +1,33 @@
 import { K8sResourceKind } from '@console/internal/module/k8s';
 import { NO_PROVISIONER, OCS_INTERNAL_CR_NAME, CEPH_STORAGE_NAMESPACE } from '../../constants';
 
+export const createDeviceSet = (scName: string, osdSize: string, portable: boolean): DeviceSet => {
```
gnehapk, line 4: `scName` and `storage` are optional param in `getOCSRequestData`, here also it should be optional too. 
`getOCSRequestData` without these parameters are used at some places for just getting few details.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
gnehapk: This will be solved once `storageClass` param is marked as optional one.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,94 @@ import {
...
-  );
-  const presentCount = _.get(ocsConfig, 'spec.storageDeviceSets[0].count');
-  const capacity = _.get(response, 'data.result[0].value[1]');
-
-  const osdSizeWithoutUnit = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
+  const cephCapacity: string = response?.data?.result?.[0]?.value[1];
+  const osdSizeWithUnit = getRequestedPVCSize(deviceSets[0].dataPVCTemplate);
+  const osdSizeWithoutUnit: number = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
   const provisionedCapacity = getProvisionedCapacity(osdSizeWithoutUnit);
+  const isNoProvionerSC: boolean = storageClass?.provisioner === NO_PROVISIONER;
```
bipuladh, line 48:
```suggestion
  const isNoProvionerSC = storageClass?.provisioner === NO_PROVISIONER;
```
Implicit from the boolean comparison

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,94 @@ import {
...
+  const selectedSCName: string = getName(storageClass);
```
bipuladh, line 49:
```suggestion
  const selectedSCName = getName(storageClass);
```
IIRC the function has the proper return type set to it.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,94 @@ import {
...
+        <strong>{`${humanizeBinaryBytes(+cephCapacity / replica).string} / ${deviceSets[
+          deviceSetIndex
+        ].count * osdSizeWithoutUnit} TiB`}</strong>
       </div>
     );
   }
 
+  const onChange = (sc: StorageClassResourceKind) => {
+    setStorageClass(sc);
+  };
```
bipuladh, line 73:
```suggestion
  const onChange = (sc: StorageClassResourceKind) => setStorageClass(sc);
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,94 @@ import {
...
     );
-  } else if (loadError || !capacity || !presentCount || !osdSizeWithoutUnit) {
+  } else if (loadError || !cephCapacity || !osdSizeWithoutUnit || deviceSetIndex === -1) {
     currentCapacity = <div className="text-muted">Not available</div>;
   } else {
     currentCapacity = (
       <div className="text-muted">
-        <strong>{`${humanizeBinaryBytes(capacity / 3).string} / ${presentCount *
-          osdSizeWithoutUnit} TiB`}</strong>
+        <strong>{`${humanizeBinaryBytes(+cephCapacity / replica).string} / ${deviceSets[
```
bipuladh, line 64:
```suggestion
        <strong>{`${humanizeBinaryBytes(Number(cephCapacity) / replica).string} / ${deviceSets[
```
Or you can let it be as it is `cephCapacity / replica`. It converts automatically.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: Its defined below as `cephCapacity`.

> Can you also attach screenshots here in this PR?

If you read the epic and description there are no visual changes

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: The underlying function definition accepts a function like `onChange`

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
afreen23: I have changed the definition, the device set needs the storage class name to be created.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
afreen23: > Its better to pass the `provisioner` too as parameter in `createDeviceSet` for deciding the `portable` value instead of passing just `true`. 


Check again `createDeviceSet` is taking `portable` value after checking for provisioner.
The function takes in the value that re relevant for device set creation. 

> This will help in case of multiple storage device sets creation.

This is working

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
afreen23: You want me to create device set without storage class ? That is not possible.
I have fixed the places where it was getting used.
See here https://github.com/openshift/console/pull/6028#discussion_r457850445

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: I understand, but I prefer to read that as well :)

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: I understand, but I prefer to read that as well :)

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: ack

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: Yeah it was doing automatically back then, but its not right to do that.
Operations or usage might change later, so adding type and converting seems fine.

afreen23:
> Can you also attach screenshots here in this PR?

If you see the description and epic, then this epic is not making any visual changes

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
bipuladh: If you want type coercion than let it handle by `/` else if you want to make it explicit use `Number` but using `+` is exactly the same as letting `/` handle it.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -1,10 +1,33 @@
...
+  return {
```
bipuladh, line 5:
```suggestion
export const createDeviceSet = (scName: string, osdSize: string, portable: boolean): DeviceSet => ({
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,92 @@ import {
...
+    const osdSize = isNoProvionerSC ? defaultRequestSize.BAREMETAL : osdSizeWithUnit;
+    if (deviceSetIndex === -1) {
+      patch.op = 'add';
+      patch.path = `/spec/storageDeviceSets/-`;
+      patch.value = createDeviceSet(selectedSCName, osdSize, portable);
+    } else {
+      patch.op = 'replace';
+      patch.path = `/spec/storageDeviceSets/${deviceSetIndex}/count`;
+      patch.value = deviceSets[deviceSetIndex].count + 1;
+    }
```
bipuladh, line 91:
```suggestion
    const portable = !isNoProvionerSC;
    const osdSize = isNoProvionerSC ? defaultRequestSize.BAREMETAL : osdSizeWithUnit;
   const patch = deviceSetIndex === -1 ?  {
      op : 'add';
      path : `/spec/storageDeviceSets/-`;
      value : createDeviceSet(selectedSCName, osdSize, portable);
    : {
      op : 'replace';
      path : `/spec/storageDeviceSets/${deviceSetIndex}/count`;
      value : deviceSets[deviceSetIndex].count + 1;
    }
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: You are right. Makes sense

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
gnehapk: Are you saying that this `onChange` is used at other place too as I don't see any?
I was suggesting to leave this like this(the way it was earlier) - 
`<OCSStorageClassDropdown onChange={setStorageClass} />`

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
gnehapk: No, I am not saying that. I was referring to the way `getOCSRequestData` method has been used at some places(in order to use the same code), without passing the storage class, just to fetch few values.
It has been used at other places too, where passing storage class name is not required -
1. https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx#L102
2. https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx#L98

For `getOCSRequestData` -> `scName` and `storage` are optional params, and `createDeviceSet` is expecting these values. So, `createDeviceSet` should be called only when these params are present. Can you add the check atleast?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -15,35 +36,34 @@ export const getOCSRequestData = (
...
-              resources: {
-                requests: {
-                  storage,
-                },
-              },
-            },
-          },
-        },
-      ],
+      storageDeviceSets: [createDeviceSet(scName, storage, true)],
```
gnehapk, line 39: refer this comment - https://github.com/openshift/console/pull/6028#discussion_r457867571

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: design preferences :)

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,90 @@ import {
...
+import { cephCapacityResource } from '../../../constants/resources';
 
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
 
+const getCurrentDeviceSet = (deviceSets: DeviceSet[], selectedSCName: string): number =>
+  deviceSets.findIndex((ds) => ds.dataPVCTemplate.spec.storageClassName === selectedSCName);
+
 export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const { ocsConfig, close, cancel } = props;
+  const deviceSets: DeviceSet[] = ocsConfig?.spec?.storageDeviceSets || [];
```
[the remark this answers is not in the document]
afreen23: yeah, my miss!

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: It will throw an error , we are handling that case.
This TODO is to show storage class used  @ installation as the preselected one.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: I need to change the typings here and look for changes at other places.
I dont want to do refactors in this PR.
https://github.com/openshift/console/blob/2d0cfb2b0b27b65126cf79f4c03aab028223e5b5/frontend/packages/ceph-storage-plugin/src/components/modals/storage-class-dropdown.tsx#L52

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
afreen23: I am aware and I must say that its not nice to have request data to be used for fetching name when we have a constant.
I planned to do in followup refactors but I did the change now.
The change is not to check for sc but only use ot for creating storage cluster.
Also `getOCSRequestData` -> `createInternalAndAttachedStorageCluster`
`createInternalAndAttachedStorageCluster` is using internally the cluster name as `OCS_INTERNAL_CR_NAME`, so using one constant at all requirements of storage cluster name is right.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -10,74 +9,95 @@ import {
...
-
-  const osdSizeWithoutUnit = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
+  const cephCapacity: string = response?.data?.result?.[0]?.value[1];
+  const osdSizeWithUnit = getStorageSize(deviceSets[0].dataPVCTemplate.spec.resources);
+  const osdSizeWithoutUnit: number = OSD_CAPACITY_SIZES[osdSizeWithUnit]?.size;
   const provisionedCapacity = getProvisionedCapacity(osdSizeWithoutUnit);
+  const isNoProvionerSC: boolean = storageClass?.provisioner === NO_PROVISIONER;
+  const selectedSCName: string = getName(storageClass);
+  const deviceSetIndex: number = getCurrentDeviceSet(deviceSets, selectedSCName);
+  const { replica } = deviceSets[deviceSetIndex];
```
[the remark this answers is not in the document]
afreen23: ` deviceSets[-1] ` will return `undefined` I assume. 
The PR will not break because if a storage class is not selected, an error will be thrown that  a storage class is not selected.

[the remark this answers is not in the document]
afreen23: I checked with older version of PR as well, works well.
Thanks!

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
bipuladh: I prefer we not mutating objects when we have already defined them as `const`.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
gnehapk: `getOCSRequestData` was used just not for the name, it was used for getting other values too like replica etc. Hence, this approach was used. 
It should be named as `getOCSRequestData` because here we are fetching the request data and not creating a storage cluster. You can replace `OCS` word, if you want but naming this like `createInternalAndAttachedStorageCluster` doesn't seem right.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -1,10 +1,36 @@
...
+      resources: {
+        requests: {
+          storage: osdSize,
+        },
+      },
+    },
+  },
+});
+
+export const createInternalAndAttachedStorageCluster = (
```
gnehapk, line 30: This function is basically returning the request data after considering all the conditions and not creating a storage cluster. Hence it was named as `getOCSRequestData`. You can replace OCS word if you want to be very specific which is not necessary IMO. But naming this like createInternalAndAttachedStorageCluster doesn't fit here.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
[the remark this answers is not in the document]
afreen23: Thanks

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
afreen23: @gnehapk I request you to review my PR, I have added replica constant.
this function will be under makeOCSRequest.

https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx#L60-L67

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
afreen23: Its not a generic function We would make it generic for all internal, external and attached devices.
It is used only for attached devices and internal
But something not needed now for this PR @gnehapk

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: Change will happen once in each if -else. I would prefer to keep it.
Thanks for suggesting

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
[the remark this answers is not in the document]
cloudbehl: correct it will return undefine but the destruction of the object will throw an error because undefined will not have replica property.

PR #6029, opened 2020-07-20
Bug 1856511: Correct alignment of help icons

pcbailey:
This PR corrects help icon styles to comply with PF4 styles and accessibility guidelines.

Before:
Selection_676

After:
Selection_675

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
-import { FormGroup } from '@patternfly/react-core';
+import { FormGroup, Popover, PopoverPosition } from '@patternfly/react-core';
 import { HelpIcon } from '@patternfly/react-icons';
-import * as classnames from 'classnames';
 import { LoadingInline } from '@console/internal/components/utils';
-import { PopoverStatus, ValidationErrorType, StatusIconAndText } from '@console/shared';
+import { ValidationErrorType } from '@console/shared';
+
 import './form-row.scss';
 
 export const FormRow: React.FC<FormRowProps> = ({
@@ -19,9 +19,6 @@ export const FormRow: React.FC<FormRowProps> = ({
   children,
   className,
 }) => {
-  const [isActive, setActive] = React.useState(false);
-  const onHide = React.useCallback(() => setActive(false), [setActive]);
-  const onShow = React.useCallback(() => setActive(true), [setActive]);
   if (isHidden) {
     return null;
   }
@@ -39,30 +36,25 @@ export const FormRow: React.FC<FormRowProps> = ({
         type === ValidationErrorType.Info || type === ValidationErrorType.Warn ? message : undefined
       }
       className={className}
-    >
-      {help && (
-        <span className="kubevirt-form-row__icon-status-container">
-          <PopoverStatus
-            title={`${fieldId} help`}
-            hideHeader
-            onHide={onHide}
-            onShow={onShow}
-            statusBody={
-              <StatusIconAndText
-                title={`${fieldId} help`}
-                iconOnly
-                icon={
-                  <HelpIcon
-                    className={classnames({ 'kubevirt-form-row__help-icon--hidden': !isActive })}
-                  />
-                }
-              />
-            }
+      labelIcon={
+        help && (
+          <Popover
+            position={PopoverPosition.right}
+            aria-label={`${fieldId} help`}
+            bodyContent={help}
           >
-            {help}
-          </PopoverStatus>
-        </span>
-      )}
+            <button
+              type="button"
+              onClick={(e) => e.preventDefault()}
+              className="pf-c-form__group-label-help"
+              aria-label={`${fieldId} help`}
+            >
+              <HelpIcon noVerticalAlign />
+            </button>
+          </Popover>
+        )
+      }
+    >
       {isLoading && (
         <span className="kubevirt-form-row__loading-container">
           <LoadingInline />


frontend/packages/console-shared/src/components/status/PopoverStatus.tsx
```diff
@@ -22,9 +22,7 @@ const PopoverStatus: React.FC<PopoverStatusProps> = ({
       isVisible={isVisible}
       shouldClose={shouldClose}
     >
-      <Button variant="link" isInline>
-        {statusBody}
-      </Button>
+      <Button variant="plain">{statusBody}</Button>
```
spadgett, line 25: This component is used in about 5 places. Do we know if this could break the layout in any of those places?

pcbailey: I didn't see any issues during my testing, which included a lot of the places that `GeneralStatus` is used. I decided to just fix it in the kubevirt plugin anyway, since the problem is only taking place there.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
```diff
@@ -39,30 +40,34 @@ export const FormRow: React.FC<FormRowProps> = ({
         type === ValidationErrorType.Info || type === ValidationErrorType.Warn ? message : undefined
       }
       className={className}
+      labelIcon={
+        help && (
+          <span>
```
matthewcarleton, line 45: do we need this span?

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
```diff
@@ -39,30 +40,34 @@ export const FormRow: React.FC<FormRowProps> = ({
...
+                  icon={
+                    <HelpIcon
+                      className={classnames({
+                        'kubevirt-form-row__help-icon--hidden': !isActive,
+                      })}
+                    />
+                  }
+                />
+              </Button>
+            </Popover>
```
matthewcarleton, line 66: Maybe I'm missing something but shouldn't this code for the popover be closer to what PF supplies? It seems very different.
```<Popover
              headerContent={
                <div>
                  The <a href="https://schema.org/name">name</a> of a <a href="https://schema.org/Person">Person</a>
                </div>
              }
              bodyContent={
                <div>
                  Often composed of{' '}
                  <a href="https://schema.org/givenName" target="_blank">
                    givenName
                  </a>{' '}
                  and{' '}
                  <a href="https://schema.org/familyName" target="_blank">
                    familyName
                  </a>
                  .
                </div>
              }
            >
              <button
                aria-label="More info for name field"
                onClick={e => e.preventDefault()}
                aria-describedby="simple-form-name"
                className="pf-c-form__group-label-help"
              >
                <HelpIcon noVerticalAlign />
              </button>
            </Popover>
```

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
pcbailey: No, we don't. I removed it.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
pcbailey: I have updated it to align closer to what PF has. I'm not sure why it was originally written to use the `StatusIconAndText` component since it's not providing a status. I also don't know why it was using a `Button` component rather than a button element since it's not being used as a standard button.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
```diff
@@ -39,30 +40,31 @@ export const FormRow: React.FC<FormRowProps> = ({
...
-      )}
+            <button
+              type="button"
+              onClick={(e) => e.preventDefault()}
+              className="pf-c-form__group-label-help"
+            >
+              <HelpIcon
+                noVerticalAlign
+                className={classnames({
+                  'kubevirt-form-row__help-icon--hidden': !isActive,
```
matthewcarleton, line 60: can we remove this class? I'm not sure what it's purpose is but it's adding a color override.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
```diff
@@ -39,30 +40,31 @@ export const FormRow: React.FC<FormRowProps> = ({
...
-              />
-            }
+            bodyContent={help}
           >
-            {help}
-          </PopoverStatus>
-        </span>
-      )}
+            <button
+              type="button"
```
matthewcarleton, line 53: do you need type=button?

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
```diff
@@ -39,30 +40,31 @@ export const FormRow: React.FC<FormRowProps> = ({
...
-                }
-              />
-            }
+            bodyContent={help}
           >
-            {help}
-          </PopoverStatus>
-        </span>
-      )}
+            <button
```
matthewcarleton, line 52: can we add the aria tags to the button?

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
pcbailey: I removed the class and the logic that was used to apply it. I'm not sure why it was added originally, but I didn't see it being used anywhere else, so it doesn't appear to be any type of convention used for consistency.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
pcbailey: I didn't add it at first, but there's an eslint check that requires buttons to have a type. In my opinion, we shouldn't override those checks without a good reason.

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/components/form/form-row.tsx
matthewcarleton: ok - I was comparing to what PF has for the implementation and didn't see it. Including shouldn't be harmful in anyway though so it's probably fine.

PR #6030, opened 2020-07-20
Correct the vertical alignment of operatorhub icons

sg00dwin:
Screen Shot 2020-07-15 at 3 24 47 PM
Screen Shot 2020-07-20 at 9 40 05 AM

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx
@@ -383,7 +383,14 @@ export const OperatorHubTileView: React.FC<OperatorHubTileViewProps> = (props) =
     const badges = [COMMUNITY_PROVIDER_TYPE, MARKETPLACE_PROVIDER_TYPE].includes(item.providerType)
       ? [badge(item.providerType)]
       : [];
-    const icon = <img className="catalog-tile-pf-icon" loading="lazy" src={imgUrl} alt="" />;
+    const icon = (
+      <img
+        className="catalog-tile-pf-icon catalog-tile-pf-icon--align-top"
+        loading="lazy"
+        src={imgUrl}
+        alt=""
+      />
+    );
     return (
       <CatalogTile
         className="co-catalog-tile"
--- a/frontend/public/components/catalog/_catalog.scss
+++ b/frontend/public/components/catalog/_catalog.scss
@@ -6,10 +6,6 @@ $catalog-item-icon-size-sm: 24px;
 $co-modal-ignore-warning-icon-width: 30px;
 $catalog-tile-width: $co-m-catalog-tile-width;
 
-// Until Patternfly-React-Extensions is updated: https://github.com/patternfly/patternfly-react/issues/1146
-.catalog-tile-pf-title {
-  @include co-break-word;
-}
 
 // reset font size back to 13px since console h5 font size is 14px
 .catalog-item-header-pf-subtitle {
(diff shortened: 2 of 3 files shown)


frontend/public/components/catalog/_catalog.scss
```diff
@@ -20,6 +16,15 @@ $catalog-tile-width: $co-m-catalog-tile-width;
   font-size: ($font-size-base - 1);
 }
 
+.catalog-tile-pf-icon {
```
rhamilto, line 19: Is this also an upstream bug?  If so, need a link to an upstream issue.

sg00dwin: I don't believe it's an upstream bug. There is a difference of markup structure between the OperatorHub and Dev Catalog tiles ...  OperatorHub has `<div class="pf-c-card__header"><span>` while Dev Catalog has `<div class="pf-c-card__header">`  not sure why... 

- https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx#L386-L407 
- https://github.com/openshift/console/blob/master/frontend/public/components/catalog/catalog-tile.tsx

rhamilto: We [pass our own `` to `icon` in order to lazy load](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx#L386), resulting in the [`span.catalog-tile-pf-icon`](https://github.com/patternfly/patternfly-react/blob/master/packages/react-catalog-view-extension/src/components/CatalogTile/CatalogTile.tsx#L113).

rhamilto: These changes (including the one you reordered) should live in `_overrides.scss` since they're overrides/additions to PatternFly rules.

Please add a comment to this rule explaining why it is necessary.

rhamilto: On second thought, instead of adding to an existing PatternFly class, you should probably create a new class with a modifier in the classname (e.g., `catalog-tile-pf-icon--aligin-top`) and add the new className in the JSX so the fix is scoped to only where it is needed.

sg00dwin: Added modifier class.

frontend/public/components/catalog/_catalog.scss
```diff
@@ -20,6 +16,11 @@ $catalog-tile-width: $co-m-catalog-tile-width;
   font-size: ($font-size-base - 1);
 }
 
+// Until Patternfly-React-Extensions is updated: https://github.com/patternfly/patternfly-react/issues/1146
```
rhamilto, line 19: Move this to `_overrides.scss` rather than reordering it?

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-items.tsx
```diff
@@ -383,7 +383,14 @@ export const OperatorHubTileView: React.FC<OperatorHubTileViewProps> = (props) =
     const badges = [COMMUNITY_PROVIDER_TYPE, MARKETPLACE_PROVIDER_TYPE].includes(item.providerType)
       ? [badge(item.providerType)]
       : [];
-    const icon = <img className="catalog-tile-pf-icon" loading="lazy" src={imgUrl} alt="" />;
+    const icon = (
+      <img
+        className="catalog-tile-pf-icon catalog-tile-pf-icon--align-top"
```
rhamilto, line 388: 👍

PR #6032, opened 2020-07-20
Add kind and apiVersion to resources returned from k8sList

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4351

**Description**
This PR adds `kind` and `apiVersion` fields to the items returned from `k8sList`. Without those fields, the overview page needed to add them to the resources causing comparisons with newly retrieved items to fail though no data had changed.
It makes sense that these items contain these fields and upon retrieval seems to be the best single place to set them.

/kind bug

### DIFF
--- a/frontend/packages/console-shared/src/utils/resource-utils.ts
+++ b/frontend/packages/console-shared/src/utils/resource-utils.ts
@@ -810,12 +810,6 @@ export const createDeploymentConfigItem = (
   resources: any,
   utils?: OverviewResourceUtil[],
 ): OverviewItem => {
-  if (!deploymentConfig.apiVersion) {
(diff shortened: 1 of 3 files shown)


andrewballantyne:
@spadgett Would you know why it doesn't return the `apiVersion` and `kind` for each of the items? I assume this is for saving the bytes of data transfer, but it's near insignificant...

spadgett:
> @spadgett Would you know why it doesn't return the `apiVersion` and `kind` for each of the items? I assume this is for saving the bytes of data transfer, but it's near insignificant...

I don't know the history except that it's been this way since k8s 1.0. The 3.x console did what @jeff-phillips-18 is doing here and injected the api version and kind in our data service.

PR #6033, opened 2020-07-20
Fix topology view updates on search and item select

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4346
https://issues.redhat.com/browse/ODC-4355

**Description**
Topology graph view needs to update node renderings when the selections change and when the filter value changes. Add a watcher on the match parameter to re-render when it changes.

/kind bug


christianvogt:
@jeff-phillips-18 can we use react-router `useParams` instead of `getQueryArgument` ? This way the hook is reactive to changes in the URL. Looks like this is available in react-router v5.

frontend/packages/dev-console/src/components/topology/Topology.tsx
```diff
@@ -280,8 +278,8 @@ const Topology: React.FC<ComponentProps> = ({
       if (selectedIds.length && !visualization.getElementById(selectedIds[0])) {
         setSelectedIds([]);
       } else {
-        const selectId = getQueryArgument('selectId');
-        const selectTab = getQueryArgument('selectTab');
+        const selectId = queryParams.get('selectId');
+        const selectTab = queryParams.get('selectTab');
```
christianvogt, line 282: `queryParams` should be a dependency.
We should remove the comment `// eslint-disable-next-line react-hooks/exhaustive-deps` because that is an old comment and the dependencies have changed since.

frontend/packages/console-shared/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,5 @@
+import { useLocation } from 'react-router-dom';
+
+export const useQueryParams = () => {
+  return new URLSearchParams(useLocation().search);
+};
```
christianvogt, line 5: We want the result to be stable and only change when search has changed.
```suggestion
export const useQueryParams = () => {
  const search = useLocation().search;
  const queryParams = React.useMemo(() => new URLSearchParams(search), [search]);
  return queryParams;
};
```

frontend/packages/dev-console/src/components/topology/TopologyDataRenderer.tsx
```diff
@@ -1,37 +1,38 @@
...
   render,
   resources,
-  namespace,
   kindsInFlight,
   trafficData,
 }) => {
   const dataModelContext = React.useContext<ExtensibleModel>(ModelContext);
   const [model, setModel] = React.useState<Model>(null);
   const [loadError, setLoadError] = React.useState<string>(null);
+  const params = useParams();
```
christianvogt, line 31: I think we are better off to pass in props than couple the component to URL params. Especially when the component is utility in nature.

frontend/packages/dev-console/src/components/topology/TopologyDataController.tsx
```diff
@@ -13,30 +13,22 @@ interface StateProps {
 }
 
 export interface TopologyDataControllerProps extends StateProps {
-  match: RMatch<{
-    name?: string;
-  }>;
```
christianvogt, line 18: Can we just accept `namespace` here instead as to not couple this component to router or url.

frontend/packages/dev-console/src/components/topology/Topology.tsx
jeff-phillips-18: ✔️

PR #6034, opened 2020-07-20
hides operatorBacked group if deployments in it are backing eventsources in topology view

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4245

Issue : https://issues.redhat.com/browse/ODC-4357 

**Analysis / Root cause**: 
Deployment owned by camelConnectors are shown in topology view as operator backed service and should be shown only in eventSources sidebar

**Solution Description**: 
Don't show Deployment owned by camelConnectors  in topology view as operator backed service and show only in eventSources sidebar

image

image

**Test setup:**
- Install Serverless Operator, create `knativeServing/knativeEventing` CR

**To test one camel Source**
-  Install Knative camel Operator
-  go to DevConsole -> Add -> EventSources
- Create Camel telegram source

**To test various camel Connectors**
- oc apply -f {file: https://gist.github.com/nicolaferraro/68afc4e07cfa7ef1a10f926967b76063 } on 4.4 Cluster (dependency is on 4.4 cluster for testing now as knative-camel operator in operator Hub doesn't have alm-examples for camel connectors and to test here have installed via operator source which works till 4.4)
- Install Serverless/Knative camel (custom) will show up in operator hub post applying above
- go to DevConsole -> Add -> EventSources

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/operators/operators-data-transformer.ts
+++ b/frontend/packages/dev-console/src/components/topology/operators/operators-data-transformer.ts
@@ -31,6 +31,7 @@ import {
 } from '../data-transforms/transform-utils';
 import { WORKLOAD_TYPES } from '../topology-utils';
 import { TYPE_SERVICE_BINDING } from '../components';
+import { isOperatorBackedKnResource } from '@console/knative-plugin/src/topology/knative-topology-utils';
 
 export const edgesFromServiceBinding = (
   source: K8sResourceKind,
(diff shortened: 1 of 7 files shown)


frontend/packages/dev-console/src/components/topology/operators/operators-data-transformer.ts
```diff
@@ -112,9 +113,14 @@ const OBSModelProps = {
...
   obj: K8sResourceKind,
   installedOperators: ClusterServiceVersionKind[],
+  resources?: TopologyDataResources,
 ): boolean => {
   const kind = _.get(obj, 'metadata.ownerReferences[0].kind', null);
   const ownerUid = _.get(obj, 'metadata.ownerReferences[0].uid');
+  // added this as needs to hide oprator backed if belong to source
+  if (isOperatorBackedKnResource(obj, resources)) {
+    return false;
+  }
```
jeff-phillips-18, line 123: Did you verify that this is necessary? I thought the depicters logic would prevent these items from being added to the model since knative is claiming them as depicted.

invincibleJai: yes @jeff-phillips-18 for other sources i.e `ping/sinkBinding/container etc` deployments (backing) created are owned by respective eventSources i.e in `ownerReference` of those can see eventSources and these deployments doesn't qualify for Operator backed service as their owner is nor CSV nor they are owned by CRDs owned by CSV. So as you said depicters logic would prevent these items from being added to the model since knative is claiming them as depicted.

but for camel sources(Connectors) deployments (backing) created are not owned by `camelSource` i.e in `ownerReference` of those can see `Integrations` and these deployments qualify for **Operator backed service** as they are owned by CRDs (Integrations) owned by CSV. So depicters logic wouldn't prevent these items from being added to the model as these deployments will be added as children to OperatorBackedService (group).

Let me know if there is a better way to handle it, will look into that. 

cc @christianvogt

PR #6035, opened 2020-07-20
Bug 1871996: Made create role bindings links consistent

rebeccaalpert:
The "Create Binding" button under the Role Bindings tab on the Role details page linked to a namespaced or cluster bindings URL. Meanwhile, other pages only use the cluster URL. I made the URLs consistent.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1871996.

### DIFF
--- a/frontend/public/components/RBAC/role.jsx
+++ b/frontend/public/components/RBAC/role.jsx
@@ -32,9 +32,10 @@ const menuActions = [
   // }),
   (kind, role) => ({
     label: 'Add Role Binding',
-    href: `/k8s/cluster/rolebindings/~new?rolekind=${roleKind(role)}&rolename=${
-      role.metadata.name
-    }`,
+    href: `/k8s/${
+      role.metadata.namespace ? `ns/${role.metadata.namespace}` : 'cluster'
+    }/rolebindings/~new?rolekind=${roleKind(role)}&rolename=${role.metadata.name}${role.metadata
+      .namespace && `&namespace=${role.metadata.namespace}`}`,
   }),
   Kebab.factory.Edit,
   Kebab.factory.Delete,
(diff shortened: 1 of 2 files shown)


frontend/public/components/RBAC/role.jsx
```diff
@@ -228,9 +228,7 @@ export const BindingsForRolePage = (props) => {
       canCreate={true}
       createButtonText="Create Binding"
       createProps={{
-        to: `/k8s/${
-          ns ? `ns/${ns}` : 'cluster'
-        }/rolebindings/~new?rolekind=${kind}&rolename=${name}`,
+        to: `/k8s/cluster/rolebindings/~new?rolekind=${kind}&rolename=${name}`,
```
spadgett, line 231: Won't this simply break the form for a namespaced Role? I would only expect this URL to work for a ClusterRole.

rebeccaalpert: No, it seems ok. The other buttons all use this link already.

spadgett: How does the form know which namespace to get the Role from when we're removing it from the URL?

spadgett: Can you point me to the other buttons that use this link for context?

rebeccaalpert: Sure, they're here:

Bindings Create button: https://github.com/openshift/console/blob/master/frontend/public/components/RBAC/bindings.jsx#L291
Role kebab: https://github.com/openshift/console/blob/master/frontend/public/components/RBAC/role.jsx#L35

spadgett: @rebeccaalpert You've found another bug. Those links are incorrect as well. I tested to confirm.

You *must* specify a namespace in addition to the role name if it's a `Role` (as opposed to a `ClusterRole`). The way the page should work is that the namespace is preselected and can't change. Right now you can select any namespace on the create role page, which doesn't make sense. (Selecting a namespace is fine for a `ClusterRole`.)

You could fix this by adding a `namespace` query parameter to the URL for `Role` and lock the choice to that namespace.

rebeccaalpert: I updated all three links to use this format (as well as a namespace parameter if available). This will show the resource link dropdown when a namespace is defined and none if it's a ClusterRole or if there's a namespace selected on the Role Bindings page. Let me know if you have other thoughts.

spadgett: @rebeccaalpert Sorry for not getting back sooner, I missed this update.

This change is fine, except it doesn't fix the original bug :/ I think we should make the original bug no plan to fix if we want to do this, then open a second bug for the other pages using the wrong URL for namespaced roles.

rebeccaalpert: I closed the original and attached the new bug to this PR.

PR #6037, opened 2020-07-20
Delete HPA In-Context

andrewballantyne:
## Built on #6022 - See the 2nd Commit

**Fixes**: 
https://issues.redhat.com/browse/ODC-4270

**Analysis / Root cause**: 
As a user, I want a way to delete an in-context HPA.

**Solution Description**: 
Add a Kebab action and modal to support deleting the associated HPA on a workload.

Screen Shot 2020-07-22 at 7 32 11 AM
Screen Shot 2020-07-22 at 7 32 19 AM

```
  doesHpaMatch checks if it aligns to a workload
    ✓ expect not to match when hpa does not target workload
    ✓ expect to match when hpa does target workload

------------------------------------------------------------|----------|----------|----------|----------|-------------------|
File                                                        |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------------------------------------------|----------|----------|----------|----------|-------------------|
  src/components/hpa/hpa-utils.ts                           |    97.26 |    90.63 |      100 |    96.83 |            92,100 |
------------------------------------------------------------|----------|----------|----------|----------|-------------------|
```

**Test setup:**

- Add an HPA to a D/DC Workload that is in Topology
- Use the Kebab action to delete

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/actions/modify-hpa.ts
+++ b/frontend/packages/console-app/src/actions/modify-hpa.ts
@@ -1,0 +1,16 @@
+import { history, KebabAction } from '@console/internal/components/utils';
+import { K8sKind, K8sResourceCommon } from '@console/internal/module/k8s';
+import { HorizontalPodAutoscalerModel } from '@console/internal/models';
+
+export const AddHorizontalPodAutoScaler: KebabAction = (kind: K8sKind, obj: K8sResourceCommon) => ({
+  label: `Add ${HorizontalPodAutoscalerModel.label}`,
+  callback: () => {
+    history.push(`/workload-hpa/${obj.metadata.namespace}/${kind.kind}/${obj.metadata.name}`);
+  },
+  accessReview: {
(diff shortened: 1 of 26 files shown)


andrewballantyne:
/hold

Waiting on dependant PR #6022.

christianvogt:
@andrewballantyne the remove action seems to only be available in topology. But not from the resource list or details pages.

andrewballantyne:
> @andrewballantyne the remove action seems to only be available in topology. But not from the resource list or details pages.

🤔  I suppose it probably shows the Add HPA action because it's in with the generics. I'll have to look into that. It was in context, but I guess that means more than Topology.

andrewballantyne:
See https://github.com/openshift/console/pull/6150

PR #6038, opened 2020-07-20
Add a new flow for create sample application

rohitkrai03:
**Fixes**:  https://issues.redhat.com/browse/ODC-4091

**Analysis / Root cause**: 
There's no quick way for users to create applications out of samples provided by openshift right now.

**Solution Description**: 
Add a new import flow for creating samples quickly.

cc: @openshift/team-devconsole-ux @beaumorley @serenamarie125 
Sample Flow

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/utils/imagestream-utils.ts
```diff
@@ -105,6 +106,7 @@ export const normalizeBuilderImages = (
       'annotations',
       'openshift.io/display-name',
     ]);
+    const description = _.get(recentTag, 'annotations.description');
```
sahil143, line 109: use optional chaining

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -532,6 +534,32 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+    type: 'Page/Route',
+    properties: {
+      exact: true,
+      path: ['/import-sample/all-namespaces', '/import-sample/ns/:ns'],
+      loader: async () =>
+        (
+          await import(
+            './components/import/ImportSamplePage' /* webpackChunkName: "dev-console-import-sample" */
+          )
+        ).default,
```
sahil143, line 560: `'/import-sample/all-namespaces'` this route results in a white screen with error `Cannot read property 'tags' of undefined`
and so will this `'/import-sample/ns/:ns'`. I guess we should redirect the user to `'samples/ns/:ns'` if there is no query params present in the URL.

sahil143: Since `/import-sample` is highly dependent on the query params, can we remove it and use `/samples/ns/:ns` with query params for the `ImportSamplePage` component. Not sure about this. WDYT?

rohitkrai03: Fixed.

frontend/packages/dev-console/src/utils/imagestream-utils.ts
rohitkrai03: Done.

bkrikori:
This looks good to me!

debsmita1:
@rohitkrai03  Just a question, do you think we should have the namespace bar with applications dropdown in the Sample app form? I am asking this because suppose in the topology user is in the context of app `x` & then tries to create a sample app. The user will expect the workload to be grouped under app `x`. Have a look at the gif below
sample-app

rohitkrai03:
> @rohitkrai03 Just a question, do you think we should have the namespace bar with applications dropdown in the Sample app form? I am asking this because suppose in the topology user is in the context of app `x` & then tries to create a sample app. The user will expect the workload to be grouped under app `x`. Have a look at the gif below
> sample-app

Ohh, I already had a conversation around this with PM and UX. Just forgot to update the code. Removed the application selector from both the pages.

frontend/packages/dev-console/src/components/import/SamplesCatalog.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <NamespacedPage hideApplications>
+        {namespace ? (
+          <>
+            <div className="odc-empty-state__title">
+              <PageHeading title="Samples" />
+              <div className="co-catalog-page__description odc-empty-state__hint-block">
+                Get Started using applications by choosing a code sample.
+              </div>
+            </div>
+            <div className="odc-empty-state__content">
```
sahil143, line 75: How come we are using empty-state styles here?

rohitkrai03: Well, it's the same exact style that we needed for this page as well as it looks exactly same as `EmptyState`. So I just re-used it directly.

frontend/packages/dev-console/src/components/import/ImportSamplePage.tsx
```diff
@@ -0,0 +1,200 @@
...
+  getSampleRef,
+  getSampleContextDir,
+} from '../../utils/imagestream-utils';
+import { detectGitType, validationSchema } from './import-validation-utils';
+import { createOrUpdateResources } from './import-submit-utils';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { Formik } from 'formik';
+import ImportSampleForm from './ImportSampleForm';
+import NamespacedPage, { NamespacedPageVariants } from '../NamespacedPage';
+import Helmet from 'react-helmet';
```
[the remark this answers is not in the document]
rohitkrai03: Fixed.

frontend/packages/dev-console/src/components/import/SamplesCatalog.tsx
christianvogt: I'd prefer that these were made into a reusable component or into a shared style file.

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -532,6 +534,32 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+            './components/import/SamplesCatalog' /* webpackChunkName: "dev-console-samples-catalog" */
+          )
+        ).default,
+    },
+  },
+  {
+    type: 'Page/Route',
+    properties: {
+      exact: true,
+      path: ['/samples/ns/:ns/is/:is/isNs/:isNs'],
```
christianvogt, line 554: Why the intermediate segments `is` and `isNs`? Seems unnecessary as you're not going to clash with any other route.

rohitkrai03: Do you mean I should have the route like `/samples/:ns/:is/:isNs`?

rohitkrai03:
Updated screens - 

Screenshot from 2020-07-23 04-42-12
Screenshot from 2020-07-23 04-42-17

PR #6039, opened 2020-07-20
Edit HPA In-Context

andrewballantyne:
## Built on #6037 - See the 3rd Commit

**Fixes**: 
https://issues.redhat.com/browse/ODC-4271

**Analysis / Root cause**: 
As a user, I'd like to be able to in-context edit an associated HPA.

**Solution Description**: 
Using the same form from the Add HPA, allow editing.

When a custom metric is in play, we auto select the YAML - they can still use the Form when they so desire to get access to cpu / memory (if applicable), the hpa name and min/max replica counts.
Screen Shot 2020-07-22 at 7 50 51 AM

```
  hasCustomMetrics accurately determines if an hpa contains non-default metrics
    ✓ expect no metrics to mean no custom metrics
    ✓ expect cpu and memory metrics to not be custom
    ✓ expect any unknown metric to be custom
    ✓ expect an unknown metric to trump known metrics

------------------------------------------------------------|----------|----------|----------|----------|-------------------|
File                                                        |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
------------------------------------------------------------|----------|----------|----------|----------|-------------------|
  src/components/hpa/hpa-utils.ts                           |    97.47 |    90.35 |      100 |    97.06 |            92,100 |
------------------------------------------------------------|----------|----------|----------|----------|-------------------|
```

**Test setup:**

- Have an HPA on a D/DC Workload
- Using the Kebab option in the Topology sidebar, Edit the HPA

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/hpa/HPAPage.tsx
+++ b/frontend/packages/dev-console/src/components/hpa/HPAPage.tsx
@@ -1,0 +1,54 @@
+import * as React from 'react';
+import { Helmet } from 'react-helmet';
+import { PageBody } from '@console/shared';
+import { LoadingInline, PageComponentProps } from '@console/internal/components/utils';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { HorizontalPodAutoscalerModel } from '@console/internal/models';
+import NamespacedPage, { NamespacedPageVariants } from '../NamespacedPage';
+import HPAFormikForm from './HPAFormikForm';
+import HPAPageHeader from './HPAPageHeader';
+import { getLimitWarning, VALID_HPA_TARGET_KINDS } from './hpa-utils';
+
+const HPAPage: React.FC<PageComponentProps> = (props) => {
+  const {
+    match: {
+      params: { ns, kind, name },
+    },
+  } = props;
+
+  const resource = React.useMemo(
+    () => ({
+      kind,
+      namespace: ns,
+      name,
+    }),
+    [kind, ns, name],
+  );
+  const [data, loaded, loadError] = useK8sWatchResource<K8sResourceKind>(resource);
+
+  const validSupportedType = VALID_HPA_TARGET_KINDS.includes(kind);
+  const title = `Add ${HorizontalPodAutoscalerModel.label}`;
+  return (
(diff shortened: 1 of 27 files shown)


andrewballantyne:
/hold

Waiting for dependant PR - #6037

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -808,7 +808,7 @@ const plugin: Plugin<ConsumedExtensions> = [
     type: 'Page/Route',
     properties: {
       exact: true,
-      path: ['/workload-hpa/:ns/:kind/:name'],
+      path: ['/workload-hpa/:ns/:resourceRef/:name', '/workload-hpa/:ns/:resourceRef/:name/edit'],
```
christianvogt, line 811: hmmm... Should only have one route. If the HPA already exists, it's an Edit scenario since we have no intention to allow a user to add a second HPA ... ?

frontend/packages/dev-console/src/components/hpa/HPAPage.tsx
```diff
@@ -3,20 +3,24 @@ import { Helmet } from 'react-helmet';
...
 const HPAPage: React.FC<PageComponentProps> = (props) => {
   const {
     match: {
-      params: { ns, kind, name },
+      params: { ns, resourceRef, name },
+      path,
     },
   } = props;
-
+  const editHPA = path === '/workload-hpa/:ns/:resourceRef/:name/edit';
```
christianvogt, line 21: Derive edit based on presence of HPA?

frontend/packages/dev-console/src/plugin.tsx
andrewballantyne: Ah; didn't think of that. Might be a bit odd if the route is invoked from "Add HPA" and ends up on "Edit HPA" but we don't want to allow them to add two+. I'll code that up.

andrewballantyne:
@christianvogt Since it might be difficult to track changes I made here, I'll detail them out:

- I updated the two parts you suggested. Majority of changes are in `HPAPage.tsx` to handle as a single route and just toggle to Edit when an HPA exists
- I updated the Form to disable the name field when in edit, as it'll just prompt an error anyways.

andrewballantyne:
See https://github.com/openshift/console/pull/6150

PR #5933, opened 2020-07-09
Add support to configure multiple helm chart repo  using HelmChartRepository CRD

akashshinde:
This PR adds support to configure multiple helm chart repos in openshift. 

It uses `HelmChartRepository` CRD to add multiple repository support. ref PR: https://github.com/openshift/api/pull/598

### How to test

Register `HelmChartRepository` CRD 
```
oc apply -f https://raw.githubusercontent.com/openshift/api/master/helm/v1beta1/0000_10-helm-chart-repository.crd.yaml
```

### Add two helm chart repository
1. Add Redhat chart repo.
```
apiVersion: helm.openshift.io/v1beta1
kind: HelmChartRepository
metadata:
  name: redhat-chart-repo
spec:
  name: redhat-chart-repo
  connectionConfig:
    url: https://redhat-developer.github.io/redhat-helm-charts
```

2. Add Azure sample helm chart repo.
```
apiVersion: helm.openshift.io/v1beta1
kind: HelmChartRepository
metadata:
  name: azure-sample-repo
spec:
  name: azure-sample-repo
  connectionConfig:
    url: https://raw.githubusercontent.com/Azure-Samples/helm-charts/master/docs
```

### Run console in dev mode
Build 
```
./build.sh
```
Setup environment to use openshift cluster context
```
source ./contrib/oc-environment.sh
```
Start console backend
```
./bin/bridge
```

### Make rest call to helm chart endpoint
```
 curl localhost:9000/api/helm/charts/index.yaml
```

Demo Link: https://youtu.be/r68ZWHzZMVA

PR #6044, opened 2020-07-21
Use HeaderBlacklist to filter unwated headers in reverse proxy

akashshinde:
This PR provides a way to delete unwated headers from proxy response.

`Proxy.Config` already has a field(`HeaderBlacklist`) to blacklist/filter headers, but It is not currently used anywhere. ref: https://github.com/openshift/console/blob/ca7021b644f481f82c0c10aac1df7fa12d8117bd/pkg/proxy/proxy.go#L23
This PR deletes headers provided in the `HeaderBlacklist` 

- [x] Added Unit test

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -31,8 +31,9 @@ type Proxy struct {
 	config       *Config
 }
 
-func filterHeaders(r *http.Response) error {
+func filterHeaders(r *http.Response, headers []string) error {
 	badHeaders := []string{"Connection", "Keep-Alive", "Proxy-Connection", "Transfer-Encoding", "Upgrade"}
+	badHeaders = append(badHeaders, headers...)
 	for _, h := range badHeaders {
 		r.Header.Del(h)
 	}
@@ -54,7 +55,9 @@ func NewProxy(cfg *Config) *Proxy {
 	reverseProxy := httputil.NewSingleHostReverseProxy(cfg.Endpoint)
 	reverseProxy.FlushInterval = time.Millisecond * 100
 	reverseProxy.Transport = transport
-	reverseProxy.ModifyResponse = filterHeaders
+	reverseProxy.ModifyResponse = func(r *http.Response) error {
+		return filterHeaders(r, cfg.HeaderBlacklist)
+	}
 
 	proxy := &Proxy{
 		reverseProxy: reverseProxy,
--- a/pkg/proxy/proxy_test.go
+++ b/pkg/proxy/proxy_test.go
@@ -12,7 +12,8 @@ import (
 )
 
 func TestProxyWebsocket(t *testing.T) {
-	proxyURL, closer, err := startProxyServer(t)
+	config := &Config{}
+	proxyURL, closer, err := startProxyServer(t, config)
 	if err != nil {
 		t.Fatalf("problem setting up proxy server: %v", err)
 	}
@@ -43,7 +44,10 @@ func TestProxyWebsocket(t *testing.T) {
 }
 
 func TestProxyHTTP(t *testing.T) {
-	proxyURL, closer, err := startProxyServer(t)
+	config := &Config{
+		HeaderBlacklist: []string{"Unwanted-Header"},
+	}
+	proxyURL, closer, err := startProxyServer(t, config)
 	if err != nil {
 		t.Fatalf("problem setting up proxy server: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


pkg/proxy/proxy.go
```diff
@@ -54,7 +55,9 @@ func NewProxy(cfg *Config) *Proxy {
 	reverseProxy := httputil.NewSingleHostReverseProxy(cfg.Endpoint)
 	reverseProxy.FlushInterval = time.Millisecond * 100
 	reverseProxy.Transport = transport
-	reverseProxy.ModifyResponse = filterHeaders
+	reverseProxy.ModifyResponse = func(r *http.Response) error {
```
spadgett, line 58: This is intended to remove request headers, not response headers.

spadgett:
It looks like we have the request headers hard-coded here, so we're not sending anything we shouldn't.

https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/pkg/proxy/proxy.go#L88

Maybe we can remove the `HeaderBlacklist` in the proxy config?

spadgett:
@akashshinde Do you have a need to filter additional response headers or did you just notice it was unused?

akashshinde:
@spadgett 
We have a use case where proxied url copies all headers along with response which is correct but origin url/proxied url copies `Cache-Control` header which caches the response in the browser.
Therefore when UI makes second request to same proxy url, browser uses the cached response instead of the one sent by server.
So we want a way to filter out that header.
If we can't use `HeaderBlacklist` then we need to find alternative way to tackle this issue.
Let me know what do you suggest.

akashshinde:
> It looks like we have the request headers hard-coded here, so we're not sending anything we shouldn't.
> 
> https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/pkg/proxy/proxy.go#L88
> 
> Maybe we can remove the `HeaderBlacklist` in the proxy config?

Please check https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/cmd/bridge/main.go#L328
We pass `Cookie` and `X-CSRFToken` here in `HeaderBlacklist` (same headers we've hard coded in https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/pkg/proxy/proxy.go#L88)

So I'm guessing we should remove hard coded headers and keep the field `HeaderBlacklist`, This way we will have control over response headers.

spadgett:
> Please check https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/cmd/bridge/main.go#L328
> We pass `Cookie` and `X-CSRFToken` here in `HeaderBlacklist` (same headers we've hard coded in https://github.com/spadgett/console/blob/bc9a54150429b081278e2024fb6cb769f5ca3a86/pkg/proxy/proxy.go#L88)

Right these are definitely request headers, not response headers. `X-CSRFToken` comes from the console client, and `Cookie` is a request header.

> So I'm guessing we should remove hard coded headers and keep the field `HeaderBlacklist`, This way we will have control over response headers.

But this was intended to filter request headers. I would rather have the hard-coded list to be honest since we should *never* send these for security reasons. It would be too easy to forget to set that on one of our proxies.

Do you have a specific need to filter a different response header?

akashshinde:
> Do you have a specific need to filter a different response header?

this is the reason - https://github.com/openshift/console/pull/6044#issuecomment-661885925 
Let me know if I am not clear.

spadgett:
I'd make this `ReponseHeaderDenylist` and remove `HeaderBlacklist` from the config and the references in main.go.

spadgett:
The problem though is that removing `Cache-Control` is not enough. It leaves it up to the browser. You really want to replace it with another value.

spadgett:
Any chance of fixing the Cache-Control header in the upstream service if it shouldn't be cached?

akashshinde:
> Any chance of fixing the Cache-Control header in the upstream service if it shouldn't be cached?

https://redhat-developer.github.io/redhat-helm-charts/index.yaml this is the URL we are serving through the proxy and it's being served through GitHub pages. I doubt If we could remove/configure http headers in GitHub service @sbose78 wdyt?
Github page adds two headers which are causing this issue.
```
Cache-Control: 600
Etag: unique-id
```

> The problem though is that removing Cache-Control is not enough. It leaves it up to the browser. You really want to replace it with another value.

Well I was thinking of replacing `Cache-Control` default value to `no-cache, private`, this would avoid caching in browser, But to modify the response headers we would need to add `ModifyResponse func(r *http.Response)` instead of `HeaderBlacklist string[]` in `Config`

`Config` struct would look like 

```
type Config struct {
	ModifyHTTPProxyResponse func(r *http.Response)
	Endpoint        *url.URL
	TLSClientConfig *tls.Config
	Origin          string
}
```

spadgett:
@akashshinde If GitHub pages is setting a valid ETag on the response, why is the Cache-Control header a problem? You should get the cached version only when the ETag hasn't changed.

akashshinde:
@spadgett we are no longer using proxy in https://github.com/openshift/console/pull/5933, therefore we can close this PR unless you have a reason to keep it open

sbose78:
Yea, let's close this PR.

PR #6048, opened 2020-07-21
Add missing catches to handlePromise

glekner:
This will solve Uncaught errors from `handlePromise`

### DIFF
--- a/frontend/public/components/utils/promise-component.tsx
+++ b/frontend/public/components/utils/promise-component.tsx
@@ -4,19 +4,20 @@ export const withHandlePromise: WithHandlePromise = (Component) => (props) => {
   const [inProgress, setInProgress] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState('');
 
-  const handlePromise = (promise) => {
+  const handlePromise = (promise, onFulfill, onError) => {
     setInProgress(true);
-    return promise.then(
+    promise.then(
       (res) => {
         setInProgress(false);
         setErrorMessage('');
-        return res;
+        onFulfill && onFulfill(res);
       },
       (error) => {
         const errorMsg = error.message || 'An error occurred. Please try again.';
         setInProgress(false);
         setErrorMessage(errorMsg);
-        return Promise.reject(errorMsg);
+        // eslint-disable-next-line no-console
+        onError ? onError(errorMsg) : console.error(error);
       },
     );
   };
@@ -69,7 +70,11 @@ export class PromiseComponent<P, S extends PromiseComponentState> extends React.
 }
 
 export type HandlePromiseProps = {
-  handlePromise: <T>(promise: Promise<T>) => Promise<T>;
+  handlePromise: <T>(
+    promise: Promise<T>,
+    onFulfill?: (res) => void,
+    onError?: (errorMsg: string) => void,
+  ) => void;
   inProgress: boolean;
   errorMessage: string;
 };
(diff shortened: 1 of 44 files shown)


glekner:
@suomiy @yaacov @rawagner can you review?

yaacov:
lgtm :+1:

I think it's better then not catching the errors :cake:

rawagner:
we could also change the `handlePromise` implementation to something like
```javascript
const handlePromise = (promise, okCallback, errorCallback) => {
    setInProgress(true);
    promise.then(
      (res) => {
        setInProgress(false);
        setErrorMessage('');
        okCallback && okCallback(res);
      },
      (error) => {
        const errorMsg = error.message || 'An error occurred. Please try again.';
        setInProgress(false);
        setErrorMessage(errorMsg);
        errorCallback && errorCallback(errorMsg);
      },
    );
  };
```
which would not force us to `catch`
With the current implementation we do something like
```javascript
handlePromise(k8sCreate(PersistentVolumeClaimModel, pvcCloneObj)).then(close).catch(consoleError)
```
With proposed one
```javascript
handlePromise(k8sCreate(PersistentVolumeClaimModel, pvcCloneObj, close))
//or optional error callback
handlePromise(k8sCreate(PersistentVolumeClaimModel, pvcCloneObj, close, optionalErrorCallback))
```

glekner:
@rawagner 
I changed `handlePromise` to the following

`const handlePromise = (promise, onFulfill, onError)`

if no error handler is supplied, we put it in console
`(error) => onError ? onError(errorMsg) : consoleError(errorMsg);`

frontend/packages/console-shared/src/utils/utils.ts
```diff
@@ -43,3 +43,6 @@ export const isValidUrl = (url: string): boolean => {
     return false;
   }
 };
+
+// eslint-disable-next-line no-console
+export const consoleError = (err: any) => console.error(err);
```
suomiy, line 48: do we want to expose this function?

frontend/public/components/utils/promise-component.tsx
```diff
@@ -1,22 +1,23 @@
...
         setErrorMessage('');
-        return res;
+        onFulfill && onFulfill(res);
       },
       (error) => {
         const errorMsg = error.message || 'An error occurred. Please try again.';
         setInProgress(false);
         setErrorMessage(errorMsg);
-        return Promise.reject(errorMsg);
+        onError ? onError(errorMsg) : consoleError(errorMsg);
```
suomiy, line 20: maybe we could display the original error here?

glekner:
fixed @suomiy suggestions

frontend/public/components/utils/promise-component.tsx
```diff
@@ -4,19 +4,20 @@ export const withHandlePromise: WithHandlePromise = (Component) => (props) => {
...
-        return res;
+        onFulfill && onFulfill(res);
       },
       (error) => {
         const errorMsg = error.message || 'An error occurred. Please try again.';
         setInProgress(false);
         setErrorMessage(errorMsg);
-        return Promise.reject(errorMsg);
+        // eslint-disable-next-line no-console
+        onError ? onError(error) : console.error(error);
```
suomiy, line 20: `onError` callback should be backwards compatible. I just mean changing the console.error

glekner:
fixed tests @rawagner @spadgett

rawagner:
I like it, @spadgett thoughts ?

spadgett:
/hold

Sorry needs rebase again :(

glekner:
rebased @spadgett

spadgett:
/hold

Build error:

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
(112,91): Property 'then' does not exist on type 'void'.
```

glekner:
@spadgett fixed

PR #6055, opened 2020-07-21
Add new channel notification

rhamilto:
Resolves https://issues.redhat.com/browse/CONSOLE-2272

Q4WvyESvKa
Note the channel name cannot be bolded per the design as the description must be a string.
Screen Shot 2020-07-21 at 1 29 09 PM
Screen Shot 2020-07-21 at 1 49 58 PM

### DIFF
--- a/frontend/public/components/_about-modal.scss
+++ b/frontend/public/components/_about-modal.scss
@@ -16,10 +16,13 @@
 
 .co-about-modal__alert {
   margin-bottom: var(--pf-global--spacer--xl);
-}
-
-// Temp fix until https://github.com/patternfly/patternfly-next/issues/2447 is fixed upstream
-.pf-c-backdrop {
-  backdrop-filter: none !important;
-  -webkit-backdrop-filter: none !important; // so Safari is consistent with Chrome
+  // PatternFly does not have an `update` alert variant
+  // See https://github.com/patternfly/patternfly-react/issues/4594
+  .pf-c-alert__icon {
+    display: none;
+    &--alt {
+      display: inline-flex;
+      vertical-align: -0.25em !important;
+    }
+  }
 }
--- a/frontend/public/components/about-modal.tsx
+++ b/frontend/public/components/about-modal.tsx
@@ -9,7 +9,7 @@ import {
 } from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
 
-import { FLAGS } from '@console/shared';
+import { BlueArrowCircleUpIcon, FLAGS } from '@console/shared';
 import { connectToFlags, FlagsObject } from '../reducers/features';
 import { getBrandingDetails } from './masthead';
 import { Firehose, useAccessReview } from './utils';
@@ -49,12 +49,14 @@ const AboutModalItems: React.FC<AboutModalItemsProps> = ({ closeAboutModal, cv }
       {clusterVersion && hasAvailableUpdates(clusterVersion) && clusterVersionIsEditable && (
         <Alert
           className="co-alert co-about-modal__alert"
-          variant="info"
           title={
             <>
-              Update available.{' '}
+              {/* PatternFly does not have an `update` alert variant
+              See https://github.com/patternfly/patternfly-react/issues/4594 */}
+              <BlueArrowCircleUpIcon className="pf-c-alert__icon pf-c-alert__icon--alt" />
+              Cluster update available.{' '}
               <Link to="/settings/cluster" onClick={closeAboutModal}>
-                View cluster settings
+                Update cluster
               </Link>
             </>
           }
--- a/frontend/public/components/cluster-settings/cluster-settings.tsx
+++ b/frontend/public/components/cluster-settings/cluster-settings.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import * as _ from 'lodash-es';
 import * as classNames from 'classnames';
-import * as semver from 'semver';
 import { Helmet } from 'react-helmet';
 import {
   Button,
@@ -16,6 +15,7 @@ import { HashLink } from 'react-router-hash-link';
 
 import { AddCircleOIcon, SyncAltIcon, PencilAltIcon } from '@patternfly/react-icons';
 import { BlueInfoCircleIcon } from '@console/shared/src/components/status';
+import { removeQueryArgument } from '@console/internal/components/utils/router';
 
 import { ClusterOperatorPage } from './cluster-operator';
 import {
@@ -37,7 +37,6 @@ import {
   ClusterVersionConditionType,
   ClusterVersionKind,
   clusterVersionReference,
-  getAvailableClusterChannels,
   getAvailableClusterUpdates,
   getClusterID,
   getClusterOperatorVersion,
@@ -46,8 +45,10 @@ import {
   getCurrentVersion,
   getDesiredClusterVersion,
   getLastCompletedUpdate,
+  getNewerClusterVersionChannel,
   getOCMLink,
   getReleaseNotesLink,
+  getSimilarClusterVersionChannels,
   getSortedUpdates,
   k8sPatch,
   K8sResourceConditionStatus,
@@ -56,6 +57,7 @@ import {
   MachineConfigPoolKind,
   referenceForModel,
   showReleaseNotes,
+  splitClusterVersionChannel,
   UpdateHistory,
 } from '../../module/k8s';
 import {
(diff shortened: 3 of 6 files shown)


frontend/public/components/_about-modal.scss
```diff
@@ -16,10 +16,12 @@
 
 .co-about-modal__alert {
   margin-bottom: var(--pf-global--spacer--xl);
-}
-
-// Temp fix until https://github.com/patternfly/patternfly-next/issues/2447 is fixed upstream
-.pf-c-backdrop {
-  backdrop-filter: none !important;
-  -webkit-backdrop-filter: none !important; // so Safari is consistent with Chrome
+  // PatternFly does not have an `update` alert variant
```
rhamilto, line 19: Add link to upstream issue once it is created

frontend/public/components/about-modal.tsx
```diff
@@ -49,9 +49,10 @@ const AboutModalItems: React.FC<AboutModalItemsProps> = ({ closeAboutModal, cv }
       {clusterVersion && hasAvailableUpdates(clusterVersion) && clusterVersionIsEditable && (
         <Alert
           className="co-alert co-about-modal__alert"
-          variant="info"
           title={
             <>
+              {/* PatternFly does not have an `update` alert variant */}
```
rhamilto, line 54: Add link to upstream issue once it is created

frontend/public/components/notification-drawer.tsx
```diff
@@ -108,20 +111,37 @@ const getUpdateNotificationEntries = (
...
           title="Cluster update available"
           toggleNotificationDrawer={toggleNotificationDrawer}
           targetPath="/settings/cluster"
         />,
+        newerChannel && (
+          <NotificationEntry
+            actionPath="/settings/cluster?showChannels"
+            actionText="Edit channel"
+            key="channel-update"
+            description={`OpenShift ${newerChannelVersion} is available. If you are
```
spadgett, line 135: We probably want to tweak the text here. I believe there is a gap between when the channel is first available and the new version actually GAs, so this message could be misleading. cc @wking @megan-hall

frontend/public/components/notification-drawer.tsx
```diff
@@ -108,20 +111,37 @@ const getUpdateNotificationEntries = (
   isLoaded: boolean,
   updateData: ClusterUpdate[],
   isEditable: boolean,
+  newerChannel: string,
+  newerChannelVersion: string,
```
spadgett, line 115: Instead of passing `updateData`, `newerChannel`, and `newerChannelVersion` separately as parameters, can we just pass the `ClusterVersion` object? We should be able to calculate the above from the `ClusterVersion` (moving that logic into this helper).

frontend/public/components/notification-drawer.tsx
```diff
@@ -108,20 +111,37 @@ const getUpdateNotificationEntries = (
...
+            actionText="Edit channel"
+            key="channel-update"
+            description={`OpenShift ${newerChannelVersion} is available. If you are
+            interested in updating this cluster to ${newerChannelVersion} in the
+            future, change the update channel to ${newerChannel} to receive recommended
+            updates.`}
+            type={NotificationTypes.update}
+            title={`OpenShift ${newerChannelVersion} is available`}
+            toggleNotificationDrawer={toggleNotificationDrawer}
+            targetPath="/settings/cluster?showChannels"
```
spadgett, line 142: The other update notification takes you to the Cluster Settings page, but doesn't open the dialog. We probably want to be consistent. cc @megan-hall

megan-hall: I was thinking about this too. 

Do we want users to be able to view the details for the update before navigating them to the update modal? Or do you think it is ok to open the update dialog directly from the notification?

frontend/public/components/notification-drawer.tsx
wking: Makes sense to me.  Maybe just `The ${newerChannel} channel is available.  If you are...`?

frontend/public/components/notification-drawer.tsx
```diff
@@ -108,20 +111,37 @@ const getUpdateNotificationEntries = (
...
           description={updateData[0].version || 'Unknown'}
           type={NotificationTypes.update}
           title="Cluster update available"
           toggleNotificationDrawer={toggleNotificationDrawer}
           targetPath="/settings/cluster"
         />,
+        newerChannel && (
+          <NotificationEntry
+            actionPath="/settings/cluster?showChannels"
+            actionText="Edit channel"
```
megan-hall, line 133: Should we change this to 'Update channel' to match the title of the modal dialog that opens?

frontend/public/components/notification-drawer.tsx
rhamilto: Another consideration is the fact we link to `/settings/cluster` several other places.  Do we want consistency when linking?  

I agree with @spadgett that these two notifications should have a consistent behavior, however.

frontend/public/components/notification-drawer.tsx
rhamilto: I agree.

frontend/public/components/notification-drawer.tsx
rhamilto: Done.

frontend/public/components/notification-drawer.tsx
rhamilto: I've made both notifications open the appropriate modal, but am happy to change.

frontend/public/components/notification-drawer.tsx
rhamilto: Done.

frontend/public/components/notification-drawer.tsx
rhamilto: Should the `title` still be `OpenShift ${newerChannelVersion} is available`?

rhamilto: Probably this instead?  
Screen Shot 2020-07-21 at 1 27 27 PM

rhamilto: gonna drop `is` in the title so it is parallel to the first

rhamilto: Screen Shot 2020-07-21 at 1 29 09 PM

frontend/public/components/notification-drawer.tsx
megan-hall: Agree as well that the two notifications should be consistent. Looks like we link directly to the update modal on the cluster overview today. So I am good with opening dialogs directly for those two alerts. Makes sense.

And totally agree they should be consistent - should have articulated that better :)

rhamilto:
/hold cancel

I believe all comments are addressed.  All good @megan-hall?  If so, please review again @spadgett.

frontend/public/components/notification-drawer.tsx
megan-hall: I think in certain places its ok to link to `/settings/cluster `like on the cluster overview details card in the top right hand corner where it says View details. But the about modal alert should probably match the notification drawer alert and open the dialog directly. Is that what you were referring to @rhamilto?

rhamilto: Exactly, @megan-hall.  We're in agreement.

rhamilto:
> The only small thing I see is that the alert on the about modal should match the alert in the notification drawer to say:
> 
> Cluster update available. Update cluster

Updated.

frontend/public/components/notification-drawer.tsx
```diff
@@ -105,25 +108,47 @@ const getAlertNotificationEntries = (
...
-          key="cluster-udpate"
+          key="cluster-update"
           description={updateData[0].version || 'Unknown'}
           type={NotificationTypes.update}
           title="Cluster update available"
           toggleNotificationDrawer={toggleNotificationDrawer}
-          targetPath="/settings/cluster"
+          targetPath="/settings/cluster?showVersions"
         />,
+        newerChannel && (
```
spadgett, line 134: Won't this only show the new channel notification if there an also an update available? These checks should be independent.

frontend/public/components/notification-drawer.tsx
```diff
@@ -105,25 +108,47 @@ const getAlertNotificationEntries = (
...
-): React.ReactNode[] =>
-  isLoaded && !_.isEmpty(updateData) && isEditable
+): React.ReactNode[] => {
+  const updateData: ClusterUpdate[] = getSortedUpdates(cv);
+  const currentChannel = cv?.spec?.channel;
+  const currentPrefix = splitClusterVersionChannel(currentChannel)?.prefix;
+  const similarChannels = getSimilarClusterVersionChannels(currentPrefix);
+  const newerChannel = getNewerClusterVersionChannel(similarChannels, currentChannel);
+  const newerChannelVersion = splitClusterVersionChannel(newerChannel)?.version;
+  return isLoaded && !_.isEmpty(updateData) && isEditable
```
spadgett, line 122: I think `isLoaded` can be removed. The behavior should be the same with no other changes.

frontend/public/components/notification-drawer.tsx
rhamilto: Yeah, you're right.  Lemme fix.

rhamilto: Done.

frontend/public/components/notification-drawer.tsx
```diff
@@ -105,25 +108,48 @@ const getAlertNotificationEntries = (
...
-          actionPath="/settings/cluster"
-          actionText="Update cluster"
-          key="cluster-udpate"
-          description={updateData[0].version || 'Unknown'}
-          type={NotificationTypes.update}
-          title="Cluster update available"
-          toggleNotificationDrawer={toggleNotificationDrawer}
-          targetPath="/settings/cluster"
-        />,
+        !_.isEmpty(updateData) && (
```
spadgett, line 123: This puts `false` in the array when there are no updates, which I'd think would cause problems. Similar problem below when `newerChannel` isn't defined.

spadgett:
Last test failure was flake https://bugzilla.redhat.com/show_bug.cgi?id=1858832

/retest

rhamilto:
Had to rebase.  Please retag, @spadgett.

frontend/public/components/cluster-settings/cluster-settings.tsx
```diff
@@ -653,6 +645,16 @@ export const ClusterVersionDetailsTable: React.FC<ClusterVersionDetailsTableProp
     verb: 'patch',
     name: cv.metadata.name,
   });
+  if (new URLSearchParams(window.location.search).has('showVersions')) {
+    clusterUpdateModal({ cv })
+      .then(() => removeQueryArgument('showVersions'))
+      .catch(_.noop);
+  }
+  if (new URLSearchParams(window.location.search).has('showChannels')) {
```
spadgett, line 653: I missed this on previous review, but I'd suggest an `else if` here to prevent ever opening two dialogs (if the URL has both query parameters for whatever reason).

rhamilto: Updated.  One more tag, @spadgett.

PR #6057, opened 2020-07-21
Add an alert action extension

afreen23:
- Extension for adding an alert action to the alerts body on status card and notification drawer
- Also adds the consumer for extension in lso-plugin

https://issues.redhat.com/browse/RHSTOR-1204

Persistent Dashboard and Notification Drawer
Screenshot from 2020-07-21 22-19-43
Cluster dashboard 
Screenshot from 2020-07-21 22-15-16

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/notification-alerts.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/notification-alerts.ts
@@ -1,0 +1,21 @@
+import { Extension } from './base';
+import { Alert } from '@console/internal/components/monitoring/types';
+
+namespace ExtensionProperties {
+  export interface AlertAction {
+    /* Alert name as defined by `alert.rule.name` property */
+    alert: string;
+    /* Action text */
+    text: string;
+    /* Action href link */
+    path: (alert: Alert) => string;
+  }
+}
+
+export interface AlertAction extends Extension<ExtensionProperties.AlertAction> {
+  type: 'AlertAction';
+}
+
+export function isAlertAction(e: Extension): e is AlertAction {
+  return e.type === 'AlertAction';
+}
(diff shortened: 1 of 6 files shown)


frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
```diff
@@ -42,6 +44,10 @@ export const StatusItem: React.FC<StatusItemProps> = ({ Icon, timestamp, message
 };
 
 const AlertItem: React.FC<AlertItemProps> = ({ alert }) => {
+  const actionsExtensions = useExtensions(isAlertAction);
+  actionsExtensions.forEach(({ properties }) =>
+    alertActions.set(properties.alert, properties.value),
+  );
   const action = alertActions.get(alert.rule.name);
```
rawagner, line 51: would be better if we change `alertActions` to function. Something like
```javascript
const getAlertActions = (acionsExtensions: AlertAction[]) => ....
```

and use it everywhere like
```javascript
const actionsExtensions = useExtensions(isAlertAction);
const action = getAlertActions(actionsExtensions).get(alert.rule.name);
```

that way a component does not always have to do `forEach` on its own

frontend/packages/console-plugin-sdk/src/typings/notification-alerts.ts
```diff
@@ -0,0 +1,24 @@
+import { Extension } from './base';
+import { Alert } from '@console/internal/components/monitoring/types';
+
+namespace ExtensionProperties {
+  export interface AlertAction {
+    /* Alert name for the key in alertAction Map */
```
rawagner, line 6: I would add comment like this, which is tied to implementation. Better would be something like
```
/* Alert name as defined by `alert.rule.name` property */
```

frontend/packages/console-plugin-sdk/src/typings/notification-alerts.ts
```diff
@@ -0,0 +1,24 @@
...
+    alert: string;
+    /* Value for the alertAction Map */
+    value: {
```
rawagner, line 9: nicer interface would be without `value` (which is again implementation detail which dont need to be reflected into interface)
```typescript
export interface AlertAction {
  /* Alert name as defined by `alert.rule.name` property */
  alert: string;

  /* Action text */
  text: string;
  
  /* Action href link */
  path: (alert: Alert) => string;
};
```

PR #6060, opened 2020-07-21
Align `Update cluster` label and links with the one in notifications

rhamilto:
so that the label and action is consistent (goes to Cluster Settings and opens the Update Cluster modal)

Requires #6055

Screen Shot 2020-07-21 at 4 59 24 PM
Screen Shot 2020-07-21 at 5 06 21 PM


rhamilto:
Had to rebase.  Please retag, @spadgett.

PR #5977, opened 2020-07-14
Kubevirt golden image

yaacov:
Description:
Common templates will get a sane default PVC name for the boot disk.

  - [x] The UI should by default use this name fom the template and use it to by default "attach a cloned disk" to VMs created of a template.
  - [x] The user should still have the ability to choose a different PVC

Assumptions:
a - The common namespace for common base images will be:
`openshift-cnv-base-images`
b - common templates will have only one datavolume

JIRA: https://issues.redhat.com/browse/CNV-5097

Screesnshot:
Usage flow:
Peek 2020-07-21 13-47

OS Template with base image
screenshot-localhost_9000-2020 07 14-13_09_24

OS Template with base image, lilnk to storage tab is active
screenshot-localhost_9000-2020 07 14-13_13_53

OS Template conntaining image name, but PVC is missing
screenshot-localhost_9000-2020 07 14-13_09_49

OS Template without a base image
screenshot-localhost_9000-2020 07 14-13_09_24

Usage Clip:
Peek 2020-07-14 12-22

PR #6066, opened 2020-07-22
add selectors for datavolume annotations

yaacov:
Add selectors specific for datavolume annotations from common templates

Part of: #5977

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/data-volume-wrapper.ts
+++ b/frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/data-volume-wrapper.ts
@@ -106,6 +106,18 @@ export class DataVolumeWrapper extends K8sResourceObjectWithTypePropertyWrapper<
     return this;
   };
 
+  setPesistentVolumeClaimName = (name: string) => {
+    this.ensurePath('spec.source.pvc');
+    this.data.spec.source.pvc.name = name;
+    return this;
+  };
+
+  setPesistentVolumeClaimNamespace = (namespace: string) => {
+    this.ensurePath('spec.source.pvc');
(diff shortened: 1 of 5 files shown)


frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/data-volume-wrapper.ts
```diff
@@ -106,6 +106,18 @@ export class DataVolumeWrapper extends K8sResourceObjectWithTypePropertyWrapper<
     return this;
   };
 
+  setPesistentVolumeClaimName = (name: string) => {
```
irosenzw, line 109: What if `name` is null or undefined ?
Is there any need to validate the name (to check for non-allowed chars) ?

frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/data-volume-wrapper.ts
```diff
@@ -106,6 +106,18 @@ export class DataVolumeWrapper extends K8sResourceObjectWithTypePropertyWrapper<
...
+    this.ensurePath('spec.source.pvc');
+    this.data.spec.source.pvc.name = name;
+    return this;
+  };
+
+  setPesistentVolumeClaimNamespace = (namespace: string) => {
```
irosenzw, line 115: same here.

frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/data-volume-wrapper.ts
yaacov: we don't do this checks in other places ... IMHO it's ok to let the user of this method do the checks before calling this method

PR #6067, opened 2020-07-22
Added support for start VM after migration from RHV

ibragins:
Very little change, just added support to start VM after migration from RHV

### DIFF
--- a/frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.configs.ts
+++ b/frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.configs.ts
@@ -41,6 +41,7 @@ export const rhvVMConfig: VMImportConfig = {
   sourceVMName: vmName,
   provider: RHV,
   instanceConfig: newInstanceConfig,
+  startOnCreation: true,
 };
 
 export const rhvVMMultiNicConfig: VMImportConfig = {
--- a/frontend/packages/kubevirt-plugin/integration-tests/tests/models/rhvImportWizard.ts
+++ b/frontend/packages/kubevirt-plugin/integration-tests/tests/models/rhvImportWizard.ts
@@ -82,6 +82,7 @@ export class RhvImportWizard extends ImportWizard {
       sourceVMName,
       storageResources,
       networkResources,
+      startOnCreation,
     } = config;
     await this.openWizard(VirtualMachineModel);
 
@@ -133,6 +134,9 @@ export class RhvImportWizard extends ImportWizard {
 
     // Review
     await this.validateReviewTab(config);
+    if (startOnCreation) {
+      await this.startOnCreation();
+    }
 
     // Import
     await this.confirmAndCreate();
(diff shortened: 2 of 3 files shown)


frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.configs.ts
```diff
@@ -41,6 +41,7 @@ export const rhvVMConfig: VMImportConfig = {
   sourceVMName: vmName,
   provider: RHV,
   instanceConfig: newInstanceConfig,
+  startOnCreation: true,
```
suomiy, line 44: We should also wait for `Running` status in v2v.rhv.import.scenario.ts where we test this config.

Maybe it would be better to have two tests - one for `Off` and one for `Running`

ibragins: Do you mean running VM? I don't expect VM to run during migration in this test

ibragins: I mean on side of RHV

suomiy: no I mean on kubevirt side:

we could have two test cases 

`startOnCreation: true,`

and 

`startOnCreation: false,`

ibragins: This parameter is optional, each test where this parameter is not specified will work as your second line

ibragins: Added separate tests, like you requested

frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.import.scenario.no.start.vm.ts
```diff
@@ -1,5 +1,5 @@
 import { withResource } from '@console/shared/src/test-utils/utils';
```
suomiy, line 1: let's use the a convention of vm.wizard.scenario and have similar test cases in the same file

frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.configs.ts
suomiy: thanks!

frontend/packages/kubevirt-plugin/integration-tests/v2v/rhv/v2v.rhv.import.scenario.no.start.vm.ts
ibragins: Done, tests merged to the same file

suomiy:
thanks!
/lgtm

PR #6069, opened 2020-07-22
Add Attached devices flag to ceph plugin

afreen23:
- The flag name is OCS_ATTACHED_DEVICES
- The flag will allow to detect ocs installation on baremetal
- The flag will be consumed for disk replacement kebab action in 4.6
- Updated Readme
- https://issues.redhat.com/browse/RHSTOR-1196

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
@@ -4,6 +4,7 @@ import {
   OCS_INTERNAL_CR_NAME,
   CEPH_STORAGE_NAMESPACE,
   OCS_DEVICE_SET_REPLICA,
+  ATTACHED_DEVICES_ANNOTATION,
 } from '../../constants';
 
 export const createDeviceSet = (scName: string, osdSize: string, portable: boolean): DeviceSet => ({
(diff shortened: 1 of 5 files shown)


frontend/packages/ceph-storage-plugin/src/constants/index.ts
```diff
@@ -15,3 +15,4 @@ export const OCS_INTERNAL_CR_NAME = 'ocs-storagecluster';
 export const NO_PROVISIONER = 'kubernetes.io/no-provisioner';
 export const OCS_SUPPORT_ANNOTATION = 'features.ocs.openshift.io/enabled';
 export const OCS_DEVICE_SET_REPLICA = 3;
+export const PLATFORM_ANNOTATION = 'cluster.ocs.openshift.io/local-devices';
```
cloudbehl, line 18:
```suggestion
export const ATTACH_DEVICE_ANNOTATION = 'cluster.ocs.openshift.io/local-devices';
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -48,6 +49,12 @@ export const getOCSRequestData = (
   if (provisioner === NO_PROVISIONER) {
     requestData.spec.monDataDirHostPath = '/var/lib/rook';
     requestData.spec.storageDeviceSets[0].portable = false;
+    requestData.metadata = {
+      ...requestData.metadata,
+      annotations: {
+        [PLATFORM_ANNOTATION]: 'true',
+      },
+    };
```
bipuladh, line 57: nit: We should these as args of the request data generator function. No point in having a function if we have to modify again and again.

afreen23: I could not do refactors for this PR. Just using an existing functionality.

afreen23: Not relevant to this PR, but I think we should make the request function generic.
Post the install PR or in the install PR itself , it can be considered @gnehapk ^^

PR #6018, opened 2020-07-17
Added Auto discovery of volumes and create SC wizard

gnehapk:
Completed -
- Auto Detect Volume Flow for Local Storage Operator View
- Create Storage Cluster Wizard for ocs installation on attached devices
- Total Capacity available on the discovered nodes

Screenshot from 2020-07-20 13-04-44
Screenshot from 2020-07-20 13-05-10
Screenshot from 2020-07-20 13-07-06

Changes to be included in next PR -
- Donut chart using the total capacity calculated
- Update the selected capacity based on filter
- Do changes to incorporate CLI scenario in all UI views
- Hide non OCS specific options from LVS and ADV

PR #6072, opened 2020-07-22
Added Donut chart inside for ocs installation on attached devices

gnehapk:
Screenshot from 2020-07-24 12-17-13


gnehapk:
Depends on #6018

cloudbehl:
Selected capacity is confusion as a user I will understand that my cluster will be of the size that is selected but out of those disk on 3 disks will be selected & made available to OCS cluster use.

gnehapk:
v

PR #6076, opened 2020-07-22
Bug 1858832: Fix e2e flakes by switching from dev to admin perspective

rohitkrai03:
**Fixes**:   https://issues.redhat.com/browse/ODC-4352

**Analysis / Root cause**: The new perspective detection changed the default perspective for a basic user to dev perspective. Tis led to e2e test failures which depend of admin perspective.

**Solution Description**:  Add a test case to switch perspective from dev to admin.


frontend/integration-tests/tests/login.scenario.ts
```diff
@@ -1,4 +1,9 @@
 import { $, browser, ExpectedConditions as until } from 'protractor';
+import {
+  sideHeader,
+  switchPerspective,
+  Perspective,
+} from '@console/dev-console/integration-tests/views/dev-perspective.view';
```
spadgett, line 6: I'm debating if these should move since they're no longer specific to the dev console tests, maybe frontend/integration-tests/views/sidenav.view.ts

@christianvogt opinion?

frontend/packages/dev-console/integration-tests/views/dev-perspective.view.ts
```diff
@@ -18,7 +18,7 @@ export enum Perspective {
 }
 
 export const pageSidebar = $('#page-sidebar .pf-c-nav .pf-c-nav__list');
-export const sideHeader = $('#page-sidebar .oc-nav-header h1');
+export const sideHeader = $('#page-sidebar .oc-nav-header h2');
```
spadgett, line 21: Existing code, but it would be better to give these `data-test` attributes.

frontend/integration-tests/tests/login.scenario.ts
rohitkrai03: My first instinct was to move them as you suggested but then didn't do it. I also think it would be better to just move them here.

rohitkrai03: @spadgett I moved the necessary code in `sidenav.view.ts`.

rohitkrai03: We can remove it from `dev-console/integration-tests` later as none of the dev console based scenarios are being run along with other e2e tests right now. They are outdated and need to be updated based on my last conversation with our QE team.

frontend/packages/dev-console/integration-tests/views/dev-perspective.view.ts
rohitkrai03: Updated the code to use `data-test-id` now.

PR #6078, opened 2020-07-22
Bug 1855318: Add external link icon to 'Open Console' VM menu action

pcbailey:
Design: http://openshift.github.io/openshift-origin-design/designs/virtualization/4.6/vnc-in-new-window/vnc-in-new-window.html

Screenshots:
Selection_678

Selection_677

### DIFF
--- a/frontend/public/components/utils/kebab.tsx
+++ b/frontend/public/components/utils/kebab.tsx
@@ -4,7 +4,7 @@ import * as classNames from 'classnames';
 import * as FocusTrap from 'focus-trap-react';
 import { connect } from 'react-redux';
 import { KEY_CODES, Tooltip } from '@patternfly/react-core';
-import { EllipsisVIcon, AngleRightIcon } from '@patternfly/react-icons';
+import { AngleRightIcon, EllipsisVIcon } from '@patternfly/react-icons';
 import Popper from '@console/shared/src/components/popper/Popper';
 import {
   annotationsModal,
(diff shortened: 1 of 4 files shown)


pcbailey:
@spadgett Please review. I couldn't find a way to accomplish this without modifying the kebab code. I'm open to an alternative and/or more generic approach if you prefer.

frontend/public/components/utils/kebab.tsx
```diff
@@ -506,6 +511,7 @@ export type KebabOption = {
   // Eg. `Menu 1/Menu 2/Menu 3`
   path?: string;
   icon?: React.ReactNode;
+  isExternalLink?: boolean;
```
yaacov, line 514: @rawagner @pcbailey hi, 
does changing line [L499](https://github.com/openshift/console/pull/6078/files#diff-44eaa3d336fe8f3b11ac4f13e1ebf60dL499) to:
`label: string | React.ReactElement;`

and then using label to pass a component with the icon included makes sense to you ?

rawagner: `label: React.ReactNode` would be the best

pcbailey: I tried making this change, but hit an issue. The `label` field is used to create the `key` for the items in `KebabMenuItems`. I tried to cast the label to a string, but it caused a type conflict for the `key` in `frontend/packages/dev-console/src/components/topology/components/nodeContextMenu.tsx` on line 34. I could cast that one to a string as well, but that will cause the PR to spread across three different packages. 

@yaacov  @rawagner Thoughts?

rawagner: Nothing better than just using `index` comes to my mind

```
{_.map(options, (o, index) => (
      <li key={index}>
```

pcbailey: Gotcha. I've seen where it's considered a discouraged practice and wasn't sure if I should go that route or not. Thanks! =)

pcbailey: It's still going to require approval for three different packages, though.

PR #6079, opened 2020-07-22
Minor changes to metrics pagination and removal of orphaned rule

sg00dwin:
Screen Shot 2020-07-22 at 4 40 09 PM
Screen Shot 2020-07-22 at 4 07 16 PM

### DIFF
--- a/frontend/public/components/monitoring/_monitoring.scss
+++ b/frontend/public/components/monitoring/_monitoring.scss
@@ -324,13 +324,6 @@ $monitoring-line-height: 18px;
   }
 }
 
-.query-browser__pagination {
-  margin-top: 15px;
-  ul {
-    list-style: none;
-  }
-}
-
 .query-browser__query {
   line-height: 1;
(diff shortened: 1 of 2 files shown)


frontend/public/components/monitoring/_monitoring.scss
```diff
@@ -324,13 +324,6 @@ $monitoring-line-height: 18px;
   }
 }
 
-.query-browser__pagination {
```
spadgett, line 327: Looks like this is still used here: https://github.com/openshift/console/blob/master/frontend/public/components/monitoring/metrics.tsx#L598

sg00dwin: Removed since it looks to be no longer needed.

spadgett:
/cc @kyoto 

It would be good to get @kyoto to review since he would know why that class was added originally.

spadgett:
Although I see @rhamilto tagged this as I was writing that comment, and I trust his review :)

PR #6088, opened 2020-07-23
Group multiple KSVC revision URLs under the same Route

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3827

**Analysis / Root cause**: 
As an user i would like to view associated Routes for selected Revisions/Service along with respective locations.

**Solution Description**: 
Group multiple revision URLs under the same route.

- Sidebar for `KSVC`: 
p0

- Sidebar for `REV` when there is no traffic split:
p1

- Sidebar for `REV` when there is traffic split:
p2

cc: @openshift/team-devconsole-ux @invincibleJai @beaumorley 

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.scss
+++ b/frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.scss
@@ -1,0 +1,3 @@
+.odc-ksroute-split-list-item {
+  padding-left: var(--pf-global--spacer--lg);
+}
--- a/frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.tsx
@@ -1,0 +1,26 @@
+import * as React from 'react';
+import { ExternalLink } from '@console/internal/components/utils';
+import { RoutesOverviewListItem } from '../../types';
+import './KSRouteSplitListItem.scss';
+
+type KSRouteSplitListItemProps = {
+  route: RoutesOverviewListItem;
+};
+
+const KSRouteSplitListItem: React.FC<KSRouteSplitListItemProps> = ({ route: { percent, url } }) =>
+  url.length > 0 && percent.length > 0 ? (
+    <li className="list-group-item">
+      <div className="odc-ksroute-split-list-item">
+        <div className="row">
+          <div className="col-xs-10">
+            <span>
+              {`${percent} → `}
+              <ExternalLink href={url} additionalClassName="co-external-link--block" text={url} />
+            </span>
+          </div>
+        </div>
+      </div>
+    </li>
+  ) : null;
+
+export default KSRouteSplitListItem;
--- a/frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react';
+import { K8sResourceKind, referenceForModel } from '@console/internal/module/k8s';
+import { ResourceLink, ExternalLink } from '@console/internal/components/utils';
+import { RouteModel } from '../../models';
+
+type KSRoutesOverviewListItemProps = {
+  ksroute: K8sResourceKind;
+};
+
+const KSRoutesOverviewListItem: React.FC<KSRoutesOverviewListItemProps> = ({ ksroute }) => {
+  const {
+    metadata: { name, namespace },
+    status,
+  } = ksroute;
+  return (
+    status && (
+      <li className="list-group-item">
+        <div className="row">
+          <div className="col-xs-10">
+            <ResourceLink kind={referenceForModel(RouteModel)} name={name} namespace={namespace} />
+            {status.url?.length > 0 && (
+              <>
+                <span className="text-muted">Location: </span>
+                <ExternalLink
+                  href={status.url}
+                  additionalClassName="co-external-link--block"
+                  text={status.url}
+                />
+              </>
+            )}
+          </div>
+        </div>
+      </li>
+    )
+  );
+};
+
+export default KSRoutesOverviewListItem;
--- a/frontend/packages/knative-plugin/src/components/overview/RoutesOverviewList.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/RoutesOverviewList.tsx
@@ -4,6 +4,7 @@ import { K8sResourceKind } from '@console/internal/module/k8s';
 import { SidebarSectionHeading } from '@console/internal/components/utils';
 import { getKnativeRoutesLinks } from '../../utils/resource-overview-utils';
 import RoutesOverviewListItem from './RoutesOverviewListItem';
+import KSRoutes from './KSRoutes';
 
 export type RoutesOverviewListProps = {
   ksroutes: K8sResourceKind[];
@@ -19,9 +20,10 @@ const RoutesOverviewList: React.FC<RoutesOverviewListProps> = ({ ksroutes, resou
       <ul className="list-group">
         {_.map(ksroutes, (route) => {
           const routeLinks = getKnativeRoutesLinks(route, resource);
-          return routeLinks.map((routeLink) => (
-            <RoutesOverviewListItem key={routeLink.uid} routeLink={routeLink} />
-          ));
+          if (routeLinks.length === 1) {
+            return <RoutesOverviewListItem key={route.metadata.uid} routeLink={routeLinks[0]} />;
+          }
+          return <KSRoutes key={route.metadata.uid} route={route} routeLinks={routeLinks} />;
         })}
       </ul>
     )}
--- a/frontend/packages/knative-plugin/src/components/overview/KSRoutes.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/KSRoutes.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { RoutesOverviewListItem } from '../../types';
+import KSRoutesOverviewListItem from './KSRoutesOverviewListItem';
+import KSRouteSplitListItem from './KSRouteSplitListItem';
+
+type KSRoutesProps = {
+  route: K8sResourceKind;
+  routeLinks: RoutesOverviewListItem[];
+};
+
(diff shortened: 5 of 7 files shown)


rottencandy:
Updated sidebar for `Revision`:
p1
p0
cc: @openshift/team-devconsole-ux @invincibleJai

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewList.tsx
```diff
@@ -0,0 +1,34 @@
...
+    ) : (
+      <ul className="list-group">
+        {ksroutes.map((route) => (
+          <>
+            <KSRoutesOverviewListItem ksroute={route} />
+            {getKnativeRoutesLinks(route, resource).map((splitRoute) => (
+              <KSRouteSplitListItem route={splitRoute} />
+            ))}
+          </>
+        ))}
```
debsmita1, line 28: create a re-useable component named `KSRoutes`(may be) and
```suggestion
{ksroutes.map((route) => {
          const routeLinks = getKnativeRoutesLinks(route, resource);
          return <KSRoutes key={route.metadata.uid} route={route} routeLinks={routeLinks} />;
        })}
```

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ksroute: K8sResourceKind;
+};
+
+const KSRoutesOverviewListItem: React.FC<KSRoutesOverviewListItemProps> = ({ ksroute }) => {
+  const {
+    metadata: { name, namespace },
+    status: { url, uid },
+  } = ksroute;
+  return (
+    <li key={uid} className="list-group-item">
```
debsmita1, line 16: if you create the re-useable component mentioned above then you won't need to pass the key here

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewList.tsx
debsmita1: `export const KSRoutes: React.FC<KSRoutesProps> = ({ route, routeLinks }) => (
  <>
    <KSRoutesOverviewListItem ksroute={route} />
    {routeLinks.map((splitRoute) => (
      <KSRouteSplitListItem key={splitRoute.uid} route={splitRoute} />
    ))}
  </>
);`

frontend/packages/knative-plugin/src/components/overview/RoutesOverviewList.tsx
```diff
@@ -4,28 +4,44 @@ import { K8sResourceKind } from '@console/internal/module/k8s';
...
+          if (routeLinks.length === 1) {
+            return <RoutesOverviewListItem routeLink={routeLinks[0]} />;
+          }
+          return (
+            <>
+              <KSRoutesOverviewListItem ksroute={route} />
+              {routeLinks.map((routeLink) => (
+                <KSRouteSplitListItem route={routeLink} />
+              ))}
+            </>
```
debsmita1, line 39: you should be able to re-use the above mentioned component here

debsmita1:
This is how it looks when there are no revisions. Maybe we should show the split route when the percentage is available? cc @invincibleJai 

no-revision

serenamarie125:
> Maybe we should show the split route when the percentage is available? 
Nice catch @debsmita1! Agreed, this section shouldn't be shown until there is at least one revision in the active traffic block.

frontend/packages/knative-plugin/src/components/overview/RoutesOverviewList.tsx
```diff
@@ -4,28 +4,44 @@ import { K8sResourceKind } from '@console/internal/module/k8s';
...
-    ) : (
+const RoutesOverviewList: React.FC<RoutesOverviewListProps> = ({ ksroutes, resource }) => {
+  if (_.isEmpty(ksroutes)) {
+    return (
+      <>
+        <SidebarSectionHeading text="Routes" />
+        <span className="text-muted">No Routes found for this resource.</span>;
+      </>
+    );
+  }
```
invincibleJai, line 23: will it make sense to have it as ternary as `<SidebarSectionHeading text="Routes" />` will always be there

frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.scss
```diff
@@ -0,0 +1,3 @@
+.odc-ksvc-route-list {
```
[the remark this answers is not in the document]
rottencandy: Updated.

frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { ExternalLink } from '@console/internal/components/utils';
+import { RoutesOverviewListItem } from '../../types';
+import './KSRouteSplitListItem.scss';
+
+export type KSRouteSplitListItemProps = {
```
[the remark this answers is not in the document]
rottencandy: Removed export.

frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+const KSRouteSplitListItem: React.FC<KSRouteSplitListItemProps> = ({
+  route: { percent, url, uid },
+}) =>
+  url.length > 0 ? (
+    <li key={uid} className="list-group-item">
+      <div className="odc-ksvc-route-list">
+        <div className="row">
+          <div className="col-xs-10">
+            <span>
+              {`${percent} → `}
```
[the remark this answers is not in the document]
rottencandy: Added check for `percent`.

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewList.tsx
rottencandy: Thanks. Updated.

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { RouteModel } from '../../models';
+
+export type KSRoutesOverviewListItemProps = {
+  ksroute: K8sResourceKind;
+};
+
+const KSRoutesOverviewListItem: React.FC<KSRoutesOverviewListItemProps> = ({ ksroute }) => {
+  const {
+    metadata: { name, namespace },
+    status: { url, uid },
```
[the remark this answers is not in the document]
rottencandy: Added check for `status`.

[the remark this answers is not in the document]
rottencandy: Was using it for `key`, but removed since I have set it in the parent component.

frontend/packages/knative-plugin/src/components/overview/RoutesOverviewList.tsx
rottencandy: Updated.

frontend/packages/knative-plugin/src/components/overview/KSRouteSplitListItem.scss
```diff
@@ -0,0 +1,3 @@
+.odc-lsroute-split-list-item {
```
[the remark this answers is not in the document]
rottencandy: Fixed.

invincibleJai:
@rottencandy thanks looks good , overall. One comment to check from UXD

incase of just base route and it receiving 100% traffic by default `ksvc` sidebar Route looks like this 

image

But Revision sidebar Route looks like this

image

_Note : in above images it's same and just one route url(just base route and it receiving 100% traffic)_

Should be consistent IMO for both revision and ksvc sidebar,
1.  Either show base route always in location and then traffic percent(split) with route
2.  Or in case of just base route and it receiving 100% traffic , show Percentage as well in route block itself as shown for revision
 
@beaumorley @serenamarie125 any thoughts/comments?

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+    metadata: { name, namespace },
+    status,
+  } = ksroute;
+  return (
+    status && (
+      <li className="list-group-item">
+        <div className="row">
+          <div className="col-xs-10">
+            <ResourceLink kind={referenceForModel(RouteModel)} name={name} namespace={namespace} />
+            {status?.url?.length > 0 && (
```
invincibleJai, line 21: nit: on line 16 you checked for status so may not need here 
```suggestion
            {status.url?.length > 0 && (
```

rottencandy: > Should be consistent IMO for both revision and ksvc sidebar,
> 
>     1. Either show base route always in location and then traffic percent(split) with route
> 
>     2. Or in case of just base route and it receiving 100% traffic , show Percentage as well in route block itself as shown for revision
> 
> 
> @beaumorley @serenamarie125 any thoughts/comments?

I think the second method of displaying the route once might be better, since it wouldn't be useful to show the same URL twice.

I've updated the `KSVC` sidebar to match the route section for `Revision`:
img

rottencandy:
> Should be consistent IMO for both revision and ksvc sidebar,
> 
>     1. Either show base route always in location and then traffic percent(split) with route
> 
>     2. Or in case of just base route and it receiving 100% traffic , show Percentage as well in route block itself as shown for revision
> 
> 
> @beaumorley @serenamarie125 any thoughts/comments?

I think the second method of displaying the route once might be better, since it wouldn't be useful to show the same URL twice.

I've updated the `KSVC` sidebar to match the route section for `Revision`:
img

frontend/packages/knative-plugin/src/components/overview/KSRoutesOverviewListItem.tsx
rottencandy: Removed.

beaumorley:
@debsmita1 @rottencandy @invincibleJai LGTM. Last few tweaks are great. Thank you all.

invincibleJai:
Works as expected, verified the changes

PR #6095, opened 2020-07-24
fixes issue with ksvc decorator not showing base route

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4253

**Analysis / Root cause**: 
On OpenShift 4.5 deploying a Knative Service with traffic distribution will show an invalid URL(route of revision) instead of base route for KSVC. The URL should be the same as advised by Knative Service, but now it's the same as one of the revisions of traffic distribution.

**Solution Description**: 
Fixed the route decorator on topology for KSVC to redirect to base URL 

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified locally, works great
/lgtm

PR #6096, opened 2020-07-24
Implement inactivity timeout for logging user out

jhadvig:
Console side of the https://issues.redhat.com/browse/CONSOLE-740 story.

/assign @spadgett

### DIFF
--- a/cmd/bridge/main.go
+++ b/cmd/bridge/main.go
@@ -77,6 +77,8 @@ func main() {
 	fUserAuthOIDCClientSecretFile := fs.String("user-auth-oidc-client-secret-file", "", "File containing the OIDC OAuth2 Client Secret.")
 	fUserAuthLogoutRedirect := fs.String("user-auth-logout-redirect", "", "Optional redirect URL on logout needed for some single sign-on identity providers.")
 
+	fInactivityTimeout := fs.Int("inactivity-timeout", 0, "Number of seconds, after which user will be logged out if inactive. Ignored if less than 300 seconds (5 minutes).")
+
 	fK8sMode := fs.String("k8s-mode", "in-cluster", "in-cluster | off-cluster")
 	fK8sModeOffClusterEndpoint := fs.String("k8s-mode-off-cluster-endpoint", "", "URL of the Kubernetes API server.")
 	fK8sModeOffClusterSkipVerifyTLS := fs.Bool("k8s-mode-off-cluster-skip-verify-tls", false, "DEV ONLY. When true, skip verification of certs presented by k8s API server.")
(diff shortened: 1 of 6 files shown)


frontend/public/components/app.jsx
```diff
@@ -167,6 +199,12 @@ class App extends React.PureComponent {
   }
 }
 
+const appStateToProps = (state) => ({
+  user: state.UI.get('user'),
+});
+
+const App = connect(appStateToProps, {})(connectToFlags(FLAGS.OPENSHIFT)(App_));
```
spadgett, line 206: I wonder if we shouldn't add this logic to the masthead instead. It already has the user and the OpenShift flag, and it already handles the logic for logout. Since we're adding the listeners to `window` it doesn't strictly need to be in the `App` component.

alecmerdler: Interestingly, I have a TamperMonkey script which deletes the masthead (because it is huge and unnecessary for my daily usage of the console). Obviously this is not supported, but would this mean I would never get logged out?

cmd/bridge/main.go
```diff
@@ -77,6 +77,8 @@ func main() {
 	fUserAuthOIDCClientSecretFile := fs.String("user-auth-oidc-client-secret-file", "", "File containing the OIDC OAuth2 Client Secret.")
 	fUserAuthLogoutRedirect := fs.String("user-auth-logout-redirect", "", "Optional redirect URL on logout needed for some single sign-on identity providers.")
 
+	fInactivityTimeout := fs.Int("inactivity-timeout", 0, "Number of seconds, after which user will be logged out if inactive")
```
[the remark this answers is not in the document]
jhadvig: > We should add some validation here. This can only be set with certain auth modes for instance, and we should reject negative values.

By validation you mean the more than 300 seconds ?

jhadvig:
@spadgett comments addressed. Makes sense to put into masthead 👍

frontend/public/components/app.jsx
spadgett: Heh, maybe? :) I guess it depends on whether it's hiding it with CSS or blocking the code from running.

If someone were determined to get around inactivity timeout, they could. For instance, by using an extension that occasionally generates mouse clicks. There's nothing we can do to block that. Even with the CLI inactivity timeout, which is all server-side, you could have a cronjob that occasionally hit the API to keep the token fresh.

The tokens do have a lifetime (24 hours by default) that you can't get around. So it will still expire.

cmd/bridge/main.go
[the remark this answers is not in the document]
spadgett: > By validation you mean the more than 300 seconds ?

No, I mean it should only be valid when you have user authentication enabled. And we probably should reject obviously incorrect values like `-23`.

cmd/bridge/main.go
```diff
@@ -204,6 +206,10 @@ func main() {
 		}
 	}
 
+	if *fInactivityTimeout < 300 {
+		log.Warning("Flag inactivity-timeout is set to less then 300 seconds and will be ignored!")
```
spadgett, line 210: We should fail if `fInactivityTimeout` is set and `fK8sAuth` is *not* `oidc` or `openshift`.

jhadvig: @spadgett comment addressed. PTAL

PR #4831, opened 2020-03-26
Cypress integration testing

dtaylor113:
Todo:
- [ ] Balaji Makram @makambalaji on the devconsole ui team is working on a [cypress cucumber](https://cucumber.io/tools/cucumber-open/) integration.
- [ ] ~~we do need to think about how static plugins can contribute tests. #2051 is what we did for protractor.~~ Notes from @spadgett "this might be less critical since we are looking at pulling plugins out of the console into their own repos, so they would have separate ci"
- [ ] ~~add ability to run suites of cypress tests ('release', 'e2e', etc). before we can move these tests out of Protractor. Investigate [Cypress Groups](https://docs.cypress.io/guides/guides/command-line.html#cypress-run-group-lt-name-gt)~~ Tracked via [JIRA](https://issues.redhat.com/browse/CONSOLE-2327)
- [ ] ~~investigate integrating [patternfly-a11y / axe](https://github.com/patternfly/patternfly-a11y) with [cypress axe plugin](https://github.com/avanslaars/cypress-axe).~~ Tracked via [JIRA](https://issues.redhat.com/browse/CONSOLE-2295) 
- [x] ~~turn on video file support?~~
- [x] ~~add `getError` and `getLogs` checks like [this](https://github.com/openshift/console/blob/master/frontend/integration-tests/protractor.conf.ts#L263-L271) in our Cypress tests to catch runtime errors in the browser.~~
- [x] ~~limit use of class selectors in favor of `data-test`~~
- [x] ~~remove specific timeouts~~
- [x] ~~change to `byTestID()`, `byTestActionID()`, & `byLegacyTestID`~~
- [x] ~~refactor: remove `utils`, rename `cy-command-additions` to `support`~~ 
- [x] ~~read BRIDGE_KUBEADMIN_IDP from an env var set in the test prow script~~
- [x] ~~implement multiple cypress reporters~~
- [x] ~~integrate with using [OpenShift with Authentication](https://github.com/openshift/console#openshift-with-authentication)~~
- [x] ~~update README with instructions for running cypress.~~
- [x] ~~move cypress to monorepo/packages directory structure~~
- [x] ~~do simple CRUD test~~
- [x] ~~split `commands.js` file into several context helper files~~
- [x] ~~enable eslint for cypress tests~~
- [x] ~~convert all cypress config files to typescript (ie. 'commands.js')~~

Todo (CI):
[...]

PR #6100, opened 2020-07-24
Comment out failing monitoring protractor tests, to be replaced by Cypress monitoring tests

dtaylor113:
Will be overwritten by https://github.com/openshift/console/pull/4831

### DIFF
--- a/frontend/integration-tests/tests/monitoring.scenario.ts
+++ b/frontend/integration-tests/tests/monitoring.scenario.ts
@@ -131,7 +131,8 @@ describe('Monitoring: Alerts', () => {
     testDetailsPage('Alert Details', testAlertName);
   });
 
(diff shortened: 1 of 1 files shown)


frontend/integration-tests/tests/monitoring.scenario.ts
```diff
@@ -131,6 +131,7 @@ describe('Monitoring: Alerts', () => {
     testDetailsPage('Alert Details', testAlertName);
   });
 
+  /* comment out failing tests, to be replaced by cypress test/pr
```
spadgett, line 134: Use `xit` instead

dtaylor113: ok, updated to `xit`

PR #6101, opened 2020-07-24
CONSOLE-2378: Add infrastructure for Console dynamic plugins

vojtechszocs:
See [CONSOLE-2378](https://issues.redhat.com/browse/CONSOLE-2378)

## :memo: Summary

This PR sets the foundation for supporting dynamic plugins in Console UI.

Existing [static plugins](https://github.com/openshift/console/blob/master/frontend/packages/console-plugin-sdk/README.md) are not affected and continue to work as expected.

For now, the only way to load and enable (activate) dynamic plugins is via browser JS console:

```js
// This only works in *non-production* builds
window.loadPluginFromURL('http://127.0.0.1:9001')
window.pluginStore.setDynamicPluginEnabled('@openshift/console-example-plugin@0.0.1', true)
```

## :gear: Details

Using the concept of webpack [module federation](https://webpack.js.org/concepts/module-federation/), it's now possible to load and interpret Console plugins from remote sources at runtime. Dynamic plugin metadata and extensions are managed by `PluginStore`.

Example dynamic plugin structure:
```
example-plugin
├── src
│   ├── components
│   │   └── Foo.tsx
│   └── utils
│       └── bar.ts
├── console-extensions.json
├── package.json
├── tsconfig.json
└── webpack.config.ts
```

### `package.json`

`name` and `version` together identify the plugin instance at runtime (`pluginID`).

Additional metadata is declared via the `consolePlugin` object.

```jsonc
{
  "name": "@openshift/console-example-plugin",
  "version": "0.0.1",
  "dependencies": {
    "@openshift/console-dynamic-plugin-sdk": "0.0.1"
  },
  // devDependencies, scripts ...
  "consolePlugin": {
    "displayName": "Example Plugin",
    "description": "Plasma reactors online. Initiating hyper drive.",
    "exposedModules": {
      "barUtils": "./utils/bar"
    },
    "dependencies": {
      "@console/pluginAPI": "~0.0.1",
      "@openshift/foo": "^1.2.3",
      "@openshift/bar": "=4.5.6"
    }
  }
}
```

### `console-extensions.json`

Declares all extensions contributed by the plugin. The `$schema` property is optional but recommended.

An extension's `properties` may contain code references, encoded as object literal `{ $codeRef: string }`.

The value of `$codeRef` is `moduleName.exportName` or `moduleName` (equivalent to `moduleName.default`).

```jsonc
{
  "$schema": "/path/to/schema/console-extensions.json",
  "data": [
    {
      "type": "console.flag",
      "properties": {
        "handler": {
          "$codeRef": "barUtils.testHandler"
        }
      }
    },
    {
      "type": "console.flag/model",
      "properties": {
        "flag": "EXAMPLE",
        "model": {
          "group": "kubevirt.io",
          "version": "v1alpha3",
          "kind": "ExampleModel"
        }
      }
    }
  ]
}
```

### `webpack.config.ts`

Use `ConsoleRemotePlugin` to generate and validate all plugin-related assets.

`ConsoleRemotePlugin` auto-detects your plugin's metadata and extension declarations.

```ts
const config: webpack.Configuration = {
  // entry is optional, but unrelated to Console plugin assets
  // context, output, resolve, module ...
  plugins: [new ConsoleRemotePlugin()],
  // mode, devtool, optimization ...
};
```

### Generated assets

`plugin-manifest.json` contains both metadata and extension declarations.

`plugin-entry.js` is the webpack [container entry](https://webpack.js.org/concepts/module-federation/#low-level-concepts) module. It provides asynchronous access to specific modules exposed by the plugin. A separate chunk is generated for each exposed module.

`utils_bar_ts-chunk.js` corresponds to the exposed `barUtils` module.

```
example-plugin/dist
├── plugin-entry.js
├── plugin-manifest.json
└── utils_bar_ts-chunk.js
```

## :package: Loading

Loading a dynamic plugin involves the following steps:

1. **Process Manifest**
  a. Fetch and validate `plugin-manifest.json`.

2. **Process Entry**
  a) Ensure that a plugin with the same `name` is not loaded yet.
  b) Evaluate `plugin-entry.js`, expecting callback to `window.loadPluginEntry`.

3. **Entry Callback**
  a) Ensure that the plugin is one we expect and that the callback was not fired yet for this plugin.
  b) [Override](https://webpack.js.org/concepts/module-federation/#overriding) shared modules. This way, a single version of React etc. is used by Console UI.
  c) Resolve encoded code references in extensions, replacing properties of type `{ $codeRef: string }` with ones of type `() => Promise<T>`.
  d) Add plugin to `PluginStore` (manifest and resolved extensions).

## :checkered_flag: Activation

Once added to `PluginStore`, the plugin is disabled by default.

Enabling the plugin via `PluginStore.setDynamicPluginEnabled` makes all of its extensions available to consuming code via `PluginStore.getAllExtensions`.

## :robot: Consuming APIs

`subscribeToExtensions(listener, ...typeGuards)`
- Low-level subscription based mechanism for consuming extensions.
- Handles subscriptions to `PluginStore` and Redux `Store` and invokes provided listeners whenever the list of computed extensions (narrowed by the given type guards and gated by feature flags) changes.
- _**This API is not meant for direct usage in a React app.**_

`useExtensions(...typeGuards)`
- [React hook](https://reactjs.org/docs/hooks-intro.html) for use with functional components.
- _**This is the preferred way of consuming Console extensions.**_

`withExtensions(typeGuardMapper)(WrappedComponent)`
- [React higher-order component](https://reactjs.org/docs/higher-order-components.html) for use with class components.
- _**This is the alternative (but less preferable) way of consuming Console extensions.**_


vojtechszocs:
/hold

Checklist still has some TODO items.

frontend/packages/console-plugin-sdk/src/withExtensions.tsx
```diff
@@ -45,50 +41,57 @@ import { Extension, ExtensionTypeGuard } from './typings';
...
+        Object.keys(typeGuardMapper).reduce((acc, propName) => {
+          acc[propName] = extensionsInUse.filter(typeGuardMapper[propName]);
+          return acc;
+        }, {} as ExtensionProps<E>),
+      [extensionsInUse],
+    );
 
-  // 2.a) Compute flags relevant for gating matched extensions
-  const gatingFlagNames = getGatingFlagNames(matchedExtensions);
+    const combinedProps = React.useMemo(
```
rawagner, line 74: is there actually any benefit using `useMemo` ? Why not just `<WrappedComponent {...props} {...extensionProps} />`

frontend/packages/console-plugin-sdk/src/useExtensions.ts
```diff
@@ -40,47 +34,46 @@ import { Extension, ExtensionTypeGuard } from './typings';
...
+  if (typeGuards.length === 0) {
+    throw new Error('You must pass at least one type guard to useExtensions');
+  }
 
-  // 1) Narrow extensions according to type guard
-  const matchedExtensions = React.useMemo(() => allExtensions.filter(typeGuard), [
-    allExtensions,
-    typeGuard,
-  ]);
+  const unsubscribeRef = React.useRef<() => void>(null);
```
rawagner, line 47: you could also use `VoidFunction`
```suggestion
  const unsubscribeRef = React.useRef<VoidFunction>(null);
```

but whatever you prefer, I dont mind

frontend/packages/console-plugin-sdk/src/subscribeToExtensions.ts
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  let lastExtensions: Extension[] = [];
+  let lastFlags: FeatureFlags = {};
+
+  const invokeAllListeners = () => {
+    const nextExtensions = getAllExtensions();
+    const nextFlags = getAllFlags();
+
+    if (_.isEqual(nextExtensions, lastExtensions) && _.isEqual(nextFlags, lastFlags)) {
```
rawagner, line 54: if you do not covert the Flags ImmutableMap to Object you wont have to use isEqual, just `nextFlags === lastFlags` would be sufficient.
I think its worth doing since `invokeAllListeners` is invoked on any redux state change - instead of calling `toObject` and `isEqual` you would need to do just `===` . And once you get inside the `if` block you can convert flags to object

frontend/dynamic-plugin-prototype/dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * _Does not execute `CodeRef` functions to load the referenced objects._
+ */
+export const resolveEncodedCodeRefs = (
+  extensions: SupportedExtension[],
+  entryModule: RemoteEntryModule,
+  pluginID: string,
+  errorCallback: () => void,
+): Extension[] => {
+  return extensions.map(_.cloneDeep).map((e) => {
```
rawagner, line 73:
```suggestion
  return _.cloneDeep(extensions).map((e) => {
```

frontend/packages/console-plugin-sdk/src/withExtensions.tsx
vojtechszocs: > Why not just `<WrappedComponent {...props} {...extensionProps} />`

`WrappedComponent` is `React.ComponentType<TCombinedProps>` so the props object needs to be of type `TCombinedProps`, otherwise TS compiler will complain.

@rawagner Agreed that there's no real benefit to `React.useMemo` here, but we still need to retain the `as TCombinedProps` cast and therefore create a separate (combined) props object:

```ts
const combinedProps = {
  ...props,
  ...extensionProps,
} as TCombinedProps;
```

frontend/packages/console-plugin-sdk/src/useExtensions.ts
vojtechszocs: Nice, I'll replace `() => void` with `VoidFunction` throughout the plugin SDK packages.

frontend/packages/console-plugin-sdk/src/subscribeToExtensions.ts
vojtechszocs: Agreed and done :smiley:

frontend/dynamic-plugin-prototype/dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
vojtechszocs: Agreed, this seems more correct since the resulting `Extension[]` array is handled separately from the original `pluginData.manifest.extensions` array.

vojtechszocs:
/hold cancel

Fixed all TODO items & addressed @rawagner's review comments.

Will rebase and re-test.

vojtechszocs:
Added README file for dynamic plugin SDK.

Excluded `dynamic-demo-plugin` directory from Console frontend root ESLint & TypeScript configs.

vojtechszocs:
Looks like `ci/prow/analyze` is failing due to

```
Couldn't read webpack bundle stats from "/home/vszocs/go/src/github.com/openshift/console/frontend/public/dist/stats.json":
SyntaxError: Unexpected token U in JSON at position 0
```

since the above JSON file contains

```
Using ^[[1m1 worker^[[22m with ^[[1m2048MB^[[22m memory limit
```

vojtechszocs:
`fork-ts-checker-webpack-plugin` bumped from `0.4.1` to `4.1.6`.

This should fix the polluted JSON output issue coming from `yarn analyze --no-open`.

spadgett:
Adding @ahardin-rh for docs approval. @ahardin-rh Initially, this is only an internal API. We expect no docs impact.

rawagner:
/lgtm

but it needs rebase again
/hold

vojtechszocs:
Update: in `public/components/app.jsx` moved the plugin init statement

```ts
initConsolePlugins(store);
```

outside the `graphQLReady.onReady` callback to ensure `onSubscriptionAdded` in `packages/console-plugin-sdk/src/subscribeToExtensions.ts` module is initialized as part of immediate application startup flow.

frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
```diff
@@ -0,0 +1,86 @@
...
+import { SupportedExtension } from '../schema/console-extensions';
+import { RemoteEntryModule, EncodedCodeRef, CodeRef } from '../types';
+
+const isEncodedCodeRef = (obj: object): obj is EncodedCodeRef =>
+  _.isPlainObject(obj) &&
+  _.isEqual(Object.getOwnPropertyNames(obj), ['$codeRef']) &&
+  typeof (obj as EncodedCodeRef).$codeRef === 'string';
+
+export const filterCodeRefProperties = (properties: object) =>
+  _.pickBy(properties, isEncodedCodeRef) as { [propName: string]: EncodedCodeRef };
```
christianvogt, line 14: Not sure it's fair to assume that all code refs appear as root properties.

frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleAssetPlugin.ts
```diff
@@ -0,0 +1,67 @@
...
+const validateExtensionsFile = (ext: ConsoleExtensionsJSON) => {
+  return new SchemaValidator(extensionsFile).validate(consoleExtensionsSchema, ext);
+};
+
+const emitJSON = (compilation: webpack.Compilation, filename: string, data: any) => {
+  const source = JSON.stringify(data, null, 2);
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  compilation.emitAsset(filename, {
```
christianvogt, line 20: You should make it a habit wherever you disable rules to add a comment as to why.

frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleRemotePlugin.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as _ from 'lodash';
+import { ConsoleAssetPlugin } from './ConsoleAssetPlugin';
+import { ConsolePackageJSON } from '../schema/plugin-package';
+import { sharedVendorModules } from '../shared-modules';
+import { SchemaValidator } from '../validation/SchemaValidator';
+import { remoteEntryFile } from '../constants';
+import consolePkgMetadataSchema from '../../dist/schema/plugin-package';
+
+// eslint-disable-next-line
+const { ReplaceSource } = require('webpack-sources');
```
christianvogt, line 13: Why a require statement and not an import?

frontend/packages/console-plugin-sdk/OWNERS
```diff
@@ -7,4 +7,4 @@ approvers:
   - spadgett
   - vojtechszocs
 labels:
-  - component/sdk
+  - component/sdk-static
```
christianvogt, line 10: I say keep it as `component/sdk` and reuse the same label for other packages that are sdk related.

frontend/packages/console-plugin-sdk/src/useExtensions.ts
```diff
@@ -40,47 +34,48 @@ import { Extension, ExtensionTypeGuard, LoadedExtension } from './typings';
...
 
-  // 1) Narrow extensions according to type guard
-  const matchedExtensions = React.useMemo(() => allExtensions.filter(typeGuard), [
-    allExtensions,
-    typeGuard,
-  ]);
+  if (!_.isEqual(latestTypeGuardsRef.current, typeGuards)) {
+    latestTypeGuardsRef.current = typeGuards;
+    tryUnsubscribe();
+    trySubscribe();
```
christianvogt, line 73: Unnecessary because this function is called immediate after the `if` block
```suggestion
```

frontend/packages/console-dynamic-plugin-sdk/OWNERS
```diff
@@ -0,0 +1,10 @@
+reviewers:
+  - christianvogt
+  - spadgett
+  - vojtechszocs
+approvers:
+  - christianvogt
+  - spadgett
+  - vojtechszocs
+labels:
+  - component/sdk-dynamic
```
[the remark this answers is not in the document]
vojtechszocs: Right, I'll post a PR similar to https://github.com/openshift/release/pull/4526.

frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -0,0 +1,137 @@
...
+        }
+      }
+    }
+  ]
+}
+```
+
+Depending on extension `type`, the `properties` object may contain code references, encoded as object
+literals `{ $codeRef: string }`. The `$codeRef` value should be formatted as `moduleName.exportName`
+(referring to a named export) or `moduleName` (referring to the `default` export).
```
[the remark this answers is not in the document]
vojtechszocs: `exposedModuleName` sounds correct so we can go with that.

frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## webpack config
```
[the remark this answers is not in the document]
vojtechszocs: In most docs, webpack is spelled all-lowercase :stuck_out_tongue: anyway I'll do the change.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
```diff
@@ -0,0 +1,83 @@
...
+  entryCallbackFired: boolean;
+};
+
+let pluginEntryCallbackRegistered = false;
+
+const pluginMap = new Map<string, ConsolePluginData>();
+
+const getPluginID = (m: ConsolePluginManifestJSON) => `${m.name}@${m.version}`;
+
+export const loadDynamicPlugin = (baseURL: string, manifest: ConsolePluginManifestJSON) => {
```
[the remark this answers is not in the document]
vojtechszocs: Good point, will do.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (existingPluginData) {
+    console.error(`Attempt to reload plugin ${getPluginID(existingPluginData.manifest)}`);
+    return;
+  }
+
+  pluginMap.set(pluginID, { manifest, entryCallbackFired: false });
+
+  const script = document.createElement('script');
+  script.src = new URL(remoteEntryFile, baseURL).toString();
+  script.async = true;
```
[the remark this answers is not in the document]
vojtechszocs: Good point, I think it will be useful to track down dynamic plugin entry script errors.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-manifest.ts
```diff
@@ -0,0 +1,18 @@
+import { coFetch } from '@console/internal/co-fetch';
+import { ConsolePluginManifestJSON } from '@console/dynamic-plugin-sdk/src/schema/plugin-manifest';
+import { pluginManifestFile } from '@console/dynamic-plugin-sdk/src/constants';
+import pluginManifestSchema from '@console/dynamic-plugin-sdk/dist/schema/plugin-manifest';
```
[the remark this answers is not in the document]
vojtechszocs: Right, will fix.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable no-console */
+
+import { PluginStore } from '@console/plugin-sdk/src/store';
+import { overrideSharedModules } from '@console/dynamic-plugin-sdk/src/shared-modules';
+import { ConsolePluginManifestJSON } from '@console/dynamic-plugin-sdk/src/schema/plugin-manifest';
+import { resolveEncodedCodeRefs } from '@console/dynamic-plugin-sdk/src/coderefs/coderef-resolver';
+import { remoteEntryFile } from '@console/dynamic-plugin-sdk/src/constants';
+import { RemoteEntryModule } from '@console/dynamic-plugin-sdk/src/types';
```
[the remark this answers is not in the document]
vojtechszocs: Will fix.

frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
vojtechszocs: Maybe let's start with supporting code refs as root properties only and add nested code ref support in a follow-up?

frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleAssetPlugin.ts
vojtechszocs: OK, will do. In this particular case, the `emitAsset` webpack API requires `Source` argument with lots of mandatory methods, but most aren't really required or necessary at runtime, so it's essentially a webpack typing issue.

frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleRemotePlugin.ts
vojtechszocs: Because static import failed for some reason, if I recall correctly. Will check again.

frontend/packages/console-plugin-sdk/OWNERS
vojtechszocs: OK, sounds good :+1:

frontend/packages/console-plugin-sdk/src/useExtensions.ts
vojtechszocs: Right, I fixed this in follow-up #6655 but will do the change here as suggested.

frontend/packages/console-dynamic-plugin-sdk/OWNERS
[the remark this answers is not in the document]
christianvogt: or we just use `component/sdk`

frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
vojtechszocs: Will add a comment to implement support for code refs appearing on any level within the extension's properties object.

yapei:
Launched a new 4.7 cluster and follow the steps in the test case, everything goes good. Approving

spadgett:
/assign @ahardin-rh @sferich888 
for docs / CEE review

sferich888:
/label px-approved
All in all this is a enablement work (prep work for the the customer facing side of the feature), to be implemented at a later time.

ahardin-rh:
:+1: Ack from the docs side

spadgett (approved):
/lgtm

PR #6102, opened 2020-07-24
Support Schema in the YAMLEditorField

andrewballantyne:
**Analysis / Root cause**: 
Our Formik YAMLEditorField lacked the ability to show schema. So when used in the new Formik variation of the yaml/form switcher, it wouldn't show the schema for the editing.

**Solution Description**: 
I wrote a separate component that deals with adding a wrapper around it. It's another component to call. There is no usage implementation part of this PR.

/cc @rohitkrai03

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/YAMLEditorField.scss
+++ b/frontend/packages/console-shared/src/components/formik-fields/YAMLEditorField.scss
@@ -1,0 +1,16 @@
+.osc-yaml-editor {
+  display: flex;
+  flex: 1 0 auto;
+  flex-direction: row;
+
+  &__editor {
+    display: flex;
+    flex: 1 0 auto;
+    flex-direction: column;
+  }
+  &__sidebar {
+    display: flex;
+    flex: 1 0 0;
+    margin-left: 20px;
+  }
+}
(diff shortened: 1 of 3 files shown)


rohitkrai03:
/approve cancel

Approved it by mistake Instead of adding a comment.

rohitkrai03:
The unit tests are failing due to some moncao-editor error. Might be related to this change.

andrewballantyne:
> @andrewballantyne If this is the only change needed to support sidebar in the editor, I wonder if we should add the support directly to `YAMLEditorField` instead of creating another component that wraps it. Also, how would one add another tab in the sidebar if needed for something similar to how we have the snippets tab?

If we wanted the samples or snippets... I think this is a problem. We already have an implementation of the yaml editor that we are duplicating... this needs to either stop and we consolidate or we have a different implementation for the YAMLEditorField for a different purpose. I think this is a small change to want to just add the schema (if possible) to the YAMLEditorField.

As for why not in the same file? 🤔  No real reason, files and components are cheap. But I guess thinking back on it now, it sorta doesn't have a lot of value as another file on the account of it being optional based on a schema anyways. I'll merge them together and make the prop optional.

andrewballantyne:
Screen Shot 2020-07-29 at 1 20 50 PM
Screen Shot 2020-07-29 at 1 21 21 PM
@rohitkrai03 Tested it does not break the YAML pages on other pages :)

PR #6103, opened 2020-07-24
Support schema-grouped descriptors on operand details page

TheRealJon:
Implement descriptor groups based on shared root path segment.

screenshot-localhost_9000-2020 07 24-17_23_25

Story: https://issues.redhat.com/browse/CONSOLE-2283
Sub-task: https://issues.redhat.com/browse/CONSOLE-2308

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/descriptors/index.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/descriptors/index.tsx
@@ -1,0 +1,159 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import { DetailsItem } from '@console/internal/components/utils';
+import { getSchemaAtPath } from '@console/shared';
+import { Descriptor, DescriptorType } from './types';
+import { K8sKind, K8sResourceKind } from '@console/internal/module/k8s';
+import { JSONSchema6 } from 'json-schema';
+import { SpecDescriptorDetailsItem } from './spec';
+import { StatusDescriptorDetailsItem } from './status';
+import { withFallback } from '@console/shared/src/components/error/error-boundary';
+
+export const DescriptorDetailsItem = withFallback<DescriptorDetailsItemProps>(
+  ({ descriptor, model, obj, onError, schema, type }) => {
+    const propertySchema = getSchemaAtPath(schema, descriptor.path);
+    const description = descriptor?.description || propertySchema?.description;
+    const fullPath = [type, ..._.toPath(descriptor.path)];
+    const label = descriptor.displayName || propertySchema?.title || _.startCase(_.last(fullPath));
+    const value = _.get(obj, fullPath, descriptor.value);
+    switch (type) {
+      case DescriptorType.spec:
+        return (
+          <SpecDescriptorDetailsItem
+            description={description}
+            descriptor={descriptor}
+            label={label}
+            model={model}
+            obj={obj}
+            onError={onError}
+            fullPath={fullPath}
+            value={value}
+          />
+        );
+      case DescriptorType.status:
+        return (
+          <StatusDescriptorDetailsItem
+            description={description}
+            descriptor={descriptor}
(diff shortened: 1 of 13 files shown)


frontend/packages/operator-lifecycle-manager/src/components/descriptors/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+            description={description}
+            descriptor={descriptor}
+            label={label}
+            model={model}
+            obj={obj}
+            onError={onError}
+            fullPath={fullPath}
+            value={value}
+          />
+        );
```
spadgett, line 45: Assuming the props are the same you could avoid some duplication with the following. TypeScript should still check the props.

```suggestion
    const descriptorProps = { description, descriptor, label, model, obj, onError, fullPath, value };
    switch (type) {
      case DescriptorType.spec:
        return <SpecDescriptorDetailsItem {...descriptorProps} />;
      case DescriptorType.status:
        return <StatusDescriptorDetailsItem {...descriptorProps} />;
```

frontend/packages/operator-lifecycle-manager/src/components/descriptors/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+  schema,
+  type,
+}) => {
+  const propertySchema = getSchemaAtPath(schema, groupName) ?? {};
+  const { root, descendants } = _.groupBy(descriptors, (descriptor) =>
+    descriptor.path === groupName ? 'root' : 'descendants',
+  );
+  const description = root?.[0]?.description || propertySchema?.description;
+  const label = root?.[0]?.displayName || propertySchema?.title || _.startCase(groupName);
+  return descendants?.length > 0 ? (
```
spadgett, line 67: I would do this instead so I don't have to think about what `undefined > 0` evaluates to.

```suggestion
  return descendants?.length ? (
```

frontend/packages/operator-lifecycle-manager/src/components/descriptors/status/index.tsx
```diff
@@ -178,69 +114,46 @@ const MainStatus: React.SFC<StatusCapabilityProps> = ({
...
->()
-  .set(StatusCapability.podStatuses, PodStatuses)
-  .set(StatusCapability.conditions, StatusConditions)
-  .set(StatusCapability.w3Link, Link)
-  .set(StatusCapability.k8sPhase, K8sPhase)
-  .set(StatusCapability.k8sPhaseReason, K8sPhaseReason)
-  .set(StatusCapability.k8sResourcePrefix, K8sResourceLink)
-  .set(StatusCapability.hidden, null);
+export const StatusDescriptorDetailsItem: React.FC<StatusCapabilityProps> = (props) => {
+  const capability = _.get(props.descriptor, ['x-descriptors', 0], null) as StatusCapability;
```
spadgett, line 118: nit: prefer optional chaining

```suggestion
  const capability = props.descriptor?.['x-descriptors']?.[0] as StatusCapability;
```

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
```diff
@@ -482,35 +480,36 @@ export const OperandDetails = connectToModel(({ crd, csv, kindObj, obj }: Operan
...
-    ].find((def) => def.name.split('.')[0] === kindObj.plural) ?? {};
-
-  const { podStatusDescriptors, otherStatusDescriptors } = _.reduce<any, any>(
-    statusDescriptors,
-    (descriptorAccumulator, descriptor) => {
-      // exclude Conditions since they are included in their own section
-      if (descriptor.path === 'conditions') {
-        return descriptorAccumulator;
-      }
+    ].find((def) => def.name === crd?.metadata?.name) ?? {};
```
spadgett, line 483: 👍

PR #6112, opened 2020-07-27
Fix Size dropdown help text margin in Pvc forms

glekner:
Margin is too big
Screen Shot 2020-07-27 at 12 59 58

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
@@ -304,11 +304,9 @@ export const UploadPVCForm: React.FC<UploadPVCFormProps> = (props) => {
               dropdownUnits={dropdownUnits}
               describedBy="request-size-help"
               inputID="request-size-input"
+              helpText="Ensure your PVC size covers the requirements of the uncompressed image and any other
+              space requirements"
             />
-            <p className="help-block" id="request-size-help">
-              Ensure your PVC size covers the requirements of the uncompressed image and any other
-              space requirements
-            </p>
           </SplitItem>
         </Split>
       </div>
--- a/frontend/public/components/utils/request-size-input.tsx
+++ b/frontend/public/components/utils/request-size-input.tsx
@@ -19,7 +19,7 @@ export class RequestSizeInput extends React.Component<RequestSizeInputProps> {
   };
 
   render() {
-    const { describedBy, name, inputID } = this.props;
+    const { describedBy, name, inputID, helpText } = this.props;
     const inputName = `${name}Value`;
     const dropdownName = `${name}Unit`;
     return (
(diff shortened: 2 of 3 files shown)


glekner:
@rawagner @yaacov can you review

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
```diff
@@ -304,11 +304,9 @@ export const UploadPVCForm: React.FC<UploadPVCFormProps> = (props) => {
               dropdownUnits={dropdownUnits}
               describedBy="request-size-help"
```
spadgett, line 305: We should remove the `describedBy` prop if it's not needed and pick a value inside the component.

frontend/public/components/utils/request-size-input.tsx
```diff
@@ -49,6 +49,11 @@ export class RequestSizeInput extends React.Component<RequestSizeInputProps> {
             ariaLabel={`Number of ${this.props.dropdownUnits[this.props.defaultRequestSizeUnit]}`}
           />
         </div>
+        {helpText && (
+          <p className="help-block" id="request-size-help">
```
spadgett, line 53: This won't be unique if the component appears on the page more than once.

spadgett: The `aria-describedby` attribute also won't match if a different value was passed into the component.

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
glekner: why is it not needed?

frontend/public/components/utils/request-size-input.tsx
glekner: I removed the `id`, 

regarding `ResourceLimitField`, its under a `FormGroup` that already takes care of the `helpText`.
the prop here is used for implementations not using a parent `FormGroup`

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
spadgett: > why is it not needed?

Because we're managing the help text inside the component now. `aria-describedby` should always reference the `id` of the help text we're adding.

frontend/public/components/utils/request-size-input.tsx
spadgett: > I removed the `id`,

You can't remove `id`. It's needed for `aria-describedby` to associate the help text with the input for screen readers.

glekner: in this case why not have the id of the help text as `id={describedBy}`?

spadgett: > in this case why not have the id of the help text as `id={describedBy}`?

1. Because it's an optional param that might not be set.
2. There's no reason the outer component should need to set that since the help text is managed internally by the dropdown.

PR #6113, opened 2020-07-27
LTR alignment for sources, PubSub, KSVC

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4237

**Analysis / Root cause**: 
Currently eventSources/channels/brokers/ksvc doesn't follow any order as a developer i would like to have left to right layout such that event sources always layout initially to the left of the channel/broker/ksvc likewise for ksvc and sink URI node

**Solution Description**: 
made sources follow the alignments / order wrt channel/Broker/KSVC and sinkURI node
- EventSources to be on left for connection to KSVC, SinkUri, Channel or Broker
- Subscriptions and Triggers should layout to their targets going out to the right.  

image

@openshift/team-devconsole-ux

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/topology/layouts/layoutConstraints.ts
+++ b/frontend/packages/knative-plugin/src/topology/layouts/layoutConstraints.ts
@@ -6,7 +6,13 @@ import {
   getGroupPadding,
   LayoutOptions,
 } from '@patternfly/react-topology';
-import { TYPE_EVENT_SOURCE_LINK, TYPE_KNATIVE_SERVICE } from '../const';
+import {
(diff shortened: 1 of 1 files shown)


invincibleJai:
> ~@invincibleJai left to right alignment of event source and services works fine but they are not aligned on y-axis at all. I would expect these to be on y-axis as well when there is enough space available but layout is always on x-axis even when there is enough space.~

@sahil143 it's all aligned on x-axis  wrt to node it's connected to , maybe we can revisit this later and see based on input from Dev's/users if there is need.

PR #6114, opened 2020-07-27
Bug 1860858: About modal for regular user shouldn't error out

jhadvig:
/assign @rhamilto

### DIFF
--- a/frontend/public/components/about-modal.tsx
+++ b/frontend/public/components/about-modal.tsx
@@ -41,7 +41,7 @@ const AboutModalItems: React.FC<AboutModalItemsProps> = ({ closeAboutModal, cv }
     group: ClusterVersionModel.apiGroup,
     resource: ClusterVersionModel.plural,
     verb: 'patch',
-    name: clusterVersion.metadata.name,
+    name: 'version',
   });
 
   return (


frontend/public/components/about-modal.tsx
```diff
@@ -41,7 +41,7 @@ const AboutModalItems: React.FC<AboutModalItemsProps> = ({ closeAboutModal, cv }
     group: ClusterVersionModel.apiGroup,
     resource: ClusterVersionModel.plural,
     verb: 'patch',
-    name: clusterVersion.metadata.name,
+    name: clusterVersion?.metadata?.name,
```
spadgett, line 44: Let's hard-code `'version`'. That way the access review is correct even if `clusterVersion` resource isn't defined. The resource name is an API, so it won't change.

```suggestion
    name: 'version',
```

spadgett:
/hold
since I had a small comment

frontend/public/components/about-modal.tsx
jhadvig: @spadgett maybe ?
```suggestion
    name: clusterVersion?.metadata?.name || 'version',
```

PR #6115, opened 2020-07-27
Bug 1841517: removing URL CD-Rom doesnt free PVC

glekner:


glekner:
@suomiy can you review?
Asking if to keep the DV/PVC is tricky because the Modal can add/delete multiple cds at once. we could add an extra modal when deleting a specific CD but it would get too bloated IMO, so the DV/PVCs are always getting deleted.

glekner:
closing in favour of #6132

PR #6117, opened 2020-07-27
Add support to configure the runtime icon for container images

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3847

**Analysis / Root cause**: 
Users want to add a runtime icon for container images to the topology view, similar to source to image builder images.

**Solution Description**:
When adding a new image Deployment from a container image the user can now also select an icon. The default icon is the OpenShift circle. It's also possible to change the icon on the edit application page.

This PR adds a new IconSection which shows a new field "Runtime Icon" which overrides the deployment label `app.openshift.io/runtime`. It uses the PatternFly Select box to show an icon select box. (See movies below.)

@openshift/team-devconsole-ux

Looks same for Deployment, DeploymentConfig and Serverless config. And its only available for Container Images:

Deployments:
odc-3847-deployment-2 webm

DeploymentConfig:
odc-3847-deploymentconfig-2 webm

Knative Service:
odc-3847-knative-service-2 webm

Not affected

**Test setup:**
* Add > Container Image
* Setup a container and select an icon

**Known issue**
Currently the interface crashs when the user inputs an icon which does not exist. There is a [PatternFly PR available](https://github.com/patternfly/patternfly-react/pull/4616) to fix this. After merging the PF PR we need to update PF in the console. In best case with this PR.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/import/icon/IconSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }, []);
+
+  const onToggle = () => setIsOpen(!isOpen);
+  const onSelect = (_, selection: string) => {
+    onValueChanged(selection);
+    setIsOpen(false);
+  };
+
+  return (
+    <Select
```
andrewballantyne, line 47: I think we want to use `Dropdown` from `frontend/public/components/utils/dropdown.jsx`.

This is the same component used for the Project header and seems to support what we are trying to do more cleanly without having to do the useEffect hack.

```
    <Dropdown
      id="icon-select"
      items={selectOptions}
      onChange={onSelect}
      title={selectOptions[value] || selectOptions['openshift']}
      autocompleteFilter={(filterString: string, node: React.ReactNode, itemName: string) => fuzzy(filterString, itemName)}
      autocompletePlaceholder="Select an icon"
    />
```
This is very rough, but I wanted to just showcase a potential solution. I reworked your `selectOptions` into a map:
```
  const selectOptions = React.useMemo<{ [key: string]: React.ReactElement }>(() => {
    const options: { [key: string]: React.ReactElement } = {};

    getIcons().forEach(({ label, url }) => {
      options[label] = (
        <div key={label}>
          <img src={url} width="24" height="24" alt={label} className="icon" />
          {label}
        </div>
      );
    });

    return options;
  }, []);
```
Again, very rough - but gives you an idea.

frontend/packages/dev-console/src/components/import/icon/IconSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+      ref={selectRef}
+      variant={isOpen ? SelectVariant.typeahead : SelectVariant.single}
+      placeholderText="Select Icon or enter an URL"
+      typeAheadAriaLabel="Select Icon or enter an URL"
```
andrewballantyne, line 51: Please make sure any text is matching what UX wants... this is one of those annoying little things that can slip through.

jerolimov: Done

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -275,6 +276,14 @@ export const getCommonInitialValues = (
   return commonInitialValues;
 };
 
+export const getIconInitialValues = (editAppResource: K8sResourceKind) => {
+  const runtimeLabel = _.get(editAppResource, 'metadata.labels["app.openshift.io/runtime"]', '');
```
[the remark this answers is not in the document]
jerolimov: Done

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -407,39 +416,37 @@ export const getInitialValues = (
...
-    appName,
-    namespace,
-  );
-  const gitDockerValues = getGitAndDockerfileInitialValues(
-    _.get(appResources, 'buildConfig.data'),
-    _.get(appResources, 'route.data'),
-  );
+  const editAppResourceData = _.get(appResources, 'editAppResource.data');
+  const routeData = _.get(appResources, 'route.data');
+  const buildConfigData = _.get(appResources, 'buildConfig.data');
```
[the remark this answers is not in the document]
jerolimov: Done

frontend/packages/dev-console/src/components/import/import-types.ts
```diff
@@ -96,6 +97,7 @@ export interface GitImportFormData {
   serverless?: ServerlessData;
   pipeline?: PipelineData;
   image: ImageData;
+  runtimeIcon?: string;
```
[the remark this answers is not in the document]
jerolimov: Yeah I do not like this as well! The problem is that in `create-knative-utils.ts` the types GitImportFormData and DeployImageFormData was used together and I want to extract runtimeIcon from them. And that works only if the property is part of both types.

https://github.com/openshift/console/blob/c116d89a8aa71c87f6aee80b5c66ca15d5f392c8/frontend/packages/knative-plugin/src/utils/create-knative-utils.ts#L11-L35

I could also change `create-knative-utils.ts` to to something like this ... but this looks much more confusing. What do you think? :thinking: 

```diff
 export const getKnativeServiceDepResource = (
-  formData: GitImportFormData | DeployImageFormData,
+  formData: (GitImportFormData & { runtimeIcon: never }) | DeployImageFormData,
   imageStreamUrl: string,
   imageStreamName?: string,
```

Combing both with an & instead of an | does not work because many callers apply only one of this types.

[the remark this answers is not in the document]
andrewballantyne: Perhaps we just leave it for now. There are type-guards we could probably use here, but I am unsure it's overly needed. Splitting the type values (ie where you use `: never`) is not going to be a fun maintenance issue. I don't think it overly matters if we have an optional property not used.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
```diff
@@ -0,0 +1,20 @@
+.icon-dropdown {
+  img {
+    margin-right: var(--pf-global--spacer--sm);
+  }
+
+  .pf-c-dropdown__toggle-text {
+    display: flex;
+    align-items: center;
+  }
```
andrewballantyne, line 9:
```suggestion
```

frontend/packages/dev-console/src/components/import/DeployImage.tsx
```diff
@@ -69,6 +69,7 @@ const DeployImage: React.FC<Props> = ({
       status: { metadata: {}, status: '' },
       ports: [],
     },
+    runtimeIcon: null,
```
andrewballantyne, line 72:
```suggestion
    runtimeIcon: 'openshift',
```
Design says the default is openshift.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
```diff
@@ -0,0 +1,20 @@
...
+
+  &__menu {
+    overflow: overlay;
```
andrewballantyne, line 12:
```suggestion
```
Not officially supported.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
```diff
@@ -0,0 +1,20 @@
...
+    max-height: 400px;
+
+    li > a {
+      display: flex;
+      align-items: center;
+    }
```
andrewballantyne, line 18: Don't think you need this.
```suggestion
```

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -275,6 +276,14 @@ export const getCommonInitialValues = (
   return commonInitialValues;
 };
 
+export const getIconInitialValues = (editAppResource: K8sResourceKind) => {
+  const runtimeLabel = editAppResource?.metadata?.labels?.['app.openshift.io/runtime'];
+  const runtimeIcon = runtimeLabel && hasIcon(runtimeLabel) ? runtimeLabel : null;
```
andrewballantyne, line 281: Now that there is a default (see my other comment)... we probably should have it here as well. Maybe as a shared constant in like the `imagestream-utils` file or something. 

My only concern is I don't want to default to the `openshift` icon if it has an icon and for some reason we don't support it. Maybe the logic is:

- if `runtimeLabel` is not found => `openshift` icon
- if `runtimeLabel` is found BUT `hasIcon` is false => `null`

frontend/packages/dev-console/src/components/import/icon/IconDropdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const Icon: React.FC<{ url: string; label: string }> = ({ url, label }) => (
+  <>
+    <img src={url} width="24" height="24" alt={label} className="icon" />
+    {label}
+  </>
+);
+
+const iconLabelAutocompleteFilter = (text: string, item: any) => fuzzy(text, item.props.label);
```
andrewballantyne, line 21:
```suggestion
const iconLabelAutocompleteFilter = (text: string, item: React.ReactElement) => fuzzy(text, item.props.label);
```

frontend/packages/dev-console/src/components/import/section/IconSection.scss
```diff
@@ -0,0 +1,27 @@
+.icon-section {
+  .custom-icon {
+    overflow: hidden;
+    margin-right: 8px;
```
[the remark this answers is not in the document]
jerolimov: Done

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
jerolimov: Without this lines it looks like this:

image

And with this vertical alignment it looks like this:

image

So I think we need then another vertical alignment (maybe a PatternFly css class I do not know yet) or keep this styles?

andrewballantyne: Ah yeah... my bad. I didn't see it when I was scanning. That does look better. Let's keep it.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
jerolimov: I just removed it now, the max-height is enough here to ensure that the scrollable is not going outside the screen.

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
jerolimov: I changed the implementation to this and keep the default on null. So if the user do not select another icon we do not need to save the label. If the user selects one we save the additional label. So we can change the default behavior at any time. OK?

```jsx
    const icon = getIcon(value || 'openshift');
    return icon ? (
      <Icon label={icon.label} url={icon.url} />
    ) : (
      <span className="btn-dropdown__item--placeholder">{placeholder}</span>
    );
```

jerolimov:
@parvathyvr @andrewballantyne I updated the videos above. After Andrews request I changed the design from just a Select field to the similar design you used in the UX documentation and we use two fields below for the Application name picker.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
andrewballantyne: Confused by your comment. 

I was just pointing out [MDN says it's not officially supported](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#Values):
> overlay 
> Behaves the same as auto, but with the scrollbars drawn on top of content instead of taking up space. Only supported in WebKit-based (e.g., Safari) and Blink-based (e.g., Chrome or Opera) browsers.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
andrewballantyne: Wait, I thought this was on the other one... this is not a patternfly component right? This shouldn't impact the styles...

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
jerolimov: I removed the overflow: overlay because it is not required.

I kept the max-height.

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
andrewballantyne: We could fake the default... I think it's helpful for the user to see the icon. If it's null we could set the view to render the OpenShift icon instead of the placeholder. But I think we lose the fidelity of informing when there just isn't an icon... maybe check with UX. Maybe this is a moot point?

frontend/packages/dev-console/src/components/import/icon/IconDropdown.scss
andrewballantyne: Yeah, this should be deleted, it's unused styles.

jerolimov: Both styles are used. One for the list placeholder and one for the list items. But I will create one comma separated entry for both!

frontend/packages/dev-console/src/components/import/DeployImage.tsx
jerolimov: The code defines now openshift as default in the IconDropdown, so I would prefer to just keep the default there and start with null, same as editing a Deployment which does not have a runtime label yet.

frontend/packages/dev-console/src/components/import/icon/IconDropdown.tsx
jerolimov: I go one step further :smile: 

```jsx
const iconLabelAutocompleteFilter = (text: string, item: React.ReactElement<IconProps>) =>
  fuzzy(text, item.props.label);
```

frontend/packages/dev-console/src/components/import/section/IconSection.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const [field] = useField<string>('runtimeIcon');
+  const formik = useFormikContext<{ runtimeIcon: string }>();
+
+  const onChanged = (value: string) => {
+    formik.setFieldValue('runtimeIcon', value);
+    formik.setFieldTouched('runtimeIcon');
+  };
+
+  return (
+    <div className="icon-section">
```
christianvogt, line 18: Do you need this div? The class name isn't defined and has no prefix.

jerolimov: Good catch. This is left over from moving styles from IconSection to the IconDropdown. Removed it now.

PR #6118, opened 2020-07-27
Add create CR button on CSV details page

zherman0:
This adds a create CR button to the Operator details page for operators that have required CRs. 
It also adds the `Required` badge to the subscribe page so users know it is needed.

Design doc: http://openshift.github.io/openshift-origin-design/designs/administrator/olm/operand-creation-at-operator-install/

Screenshot_2020-07-27 Operator Installation · OKD
Screenshot_2020-07-28 portworx-operator v1 3 0 · Details · OKD

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
@@ -15,7 +15,14 @@ import {
   Popover,
   CardTitle,
 } from '@patternfly/react-core';
-import { ALL_NAMESPACES_KEY, Status, WarningStatus, getNamespace, getUID } from '@console/shared';
+import {
+  ALL_NAMESPACES_KEY,
+  Status,
+  WarningStatus,
+  getNamespace,
+  getUID,
+  StatusIconAndText,
+} from '@console/shared';
 import {
   DetailsPage,
   Table,
@@ -29,11 +36,13 @@ import {
   modelFor,
   referenceForModel,
   referenceFor,
+  groupVersionFor,
   GroupVersionKind,
   K8sKind,
   k8sKill,
   k8sPatch,
   k8sGet,
+  K8sResourceCommon,
 } from '@console/internal/module/k8s';
 import { ResourceEventStream } from '@console/internal/components/events';
 import { Conditions } from '@console/internal/components/conditions';
@@ -46,6 +55,7 @@ import {
   Timestamp,
   SectionHeading,
   ResourceSummary,
+  ResourceStatus,
   ScrollToTopOnMount,
   AsyncComponent,
   ExternalLink,
@@ -58,6 +68,7 @@ import {
   resourceObjPath,
   KebabAction,
 } from '@console/internal/components/utils';
+import { useK8sGet } from '@console/internal/components/utils/k8s-get-hook';
 import { useAccessReview } from '@console/internal/components/utils/rbac';
 import { RootState } from '@console/internal/redux';
 import {
@@ -90,8 +101,10 @@ import {
   upgradeRequiresApproval,
   UpgradeApprovalLink,
 } from './subscription';
+import { RedExclamationCircleIcon } from '@console/shared/src/components/status/icons';
 import { ClusterServiceVersionLogo, referenceForProvidedAPI, providedAPIsFor } from './index';
 import { getBreadcrumbPath } from '@console/internal/components/utils/breadcrumbs';
+import { CreateInitializationResourceButton } from './operator-install-page';
 
 const clusterServiceVersionStateToProps = (state: RootState): ClusterServiceVersionStateProps => {
   return {
@@ -694,20 +707,28 @@ export const MarkdownView = (props: {
 };
 
 export const CRDCard: React.SFC<CRDCardProps> = (props) => {
-  const { csv, crd, canCreate } = props;
+  const { csv, crd, canCreate, required = false } = props;
   const reference = referenceForProvidedAPI(crd);
   const model = modelFor(reference);
   const createRoute = () =>
     `/k8s/ns/${csv.metadata.namespace}/${ClusterServiceVersionModel.plural}/${csv.metadata.name}/${reference}/~new`;
+
   return (
     <Card>
       <CardTitle>
-        <ResourceLink
-          kind={referenceForProvidedAPI(crd)}
-          title={crd.name}
-          linkTo={false}
-          displayName={crd.displayName || crd.kind}
-        />
+        <span className="co-resource-item">
+          <ResourceLink
+            kind={referenceForProvidedAPI(crd)}
+            title={crd.name}
+            linkTo={false}
+            displayName={crd.displayName || crd.kind}
+          />
+          {required && (
+            <ResourceStatus badgeAlt>
+              <StatusIconAndText icon={<RedExclamationCircleIcon />} title="Required" />
+            </ResourceStatus>
+          )}
+        </span>
       </CardTitle>
       <CardBody>
         <MarkdownView content={crd.description} truncateContent />
@@ -761,6 +782,53 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
   );
 });
 
+const InitializationResourceAlert: React.SFC<InitializationResourceAlertProps> = (props) => {
+  const { initializationResource, csv } = props;
+
+  const initializationResourceKind = initializationResource?.kind;
+  const { group: initializationResourceGroup } = groupVersionFor(
+    initializationResource?.apiVersion,
+  );
+  const model = modelFor(referenceFor(initializationResource));
+
+  // Check if the CR is already present - only checks for the specific name/ns in the initialization-resource
+  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceCommon>(
+    model,
+    initializationResource?.metadata.name,
+    model?.namespaced ? initializationResource?.metadata.namespace || csv.metadata.namespace : null,
+  );
+  const canCreateCustomResource = useAccessReview({
+    group: initializationResourceGroup,
+    resource: model?.plural,
+    namespace: model?.namespaced
+      ? initializationResource?.metadata.namespace || csv.metadata.namespace
+      : null,
+    verb: 'create',
+  });
+
+  if (!customResource && customResourceLoaded && canCreateCustomResource) {
+    return (
+      <Alert
+        isInline
+        className="co-alert"
+        variant="warning"
+        title={`${initializationResourceKind} Required`}
+      >
+        <p>Create a {initializationResourceKind} instance to use this operator.</p>
+        <CreateInitializationResourceButton
+          obj={props.csv}
+          targetNamespace={
+            model?.namespaced
+              ? initializationResource?.metadata.namespace || csv.metadata?.namespace
+              : null
+          }
+        />
+      </Alert>
+    );
+  }
+  return null;
+};
+
 export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetailsProps> = (
   props,
 ) => {
@@ -768,7 +836,18 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
   const {
     'marketplace.openshift.io/support-workflow': marketplaceSupportWorkflow,
     'olm.targetNamespaces': olmTargetNamespaces = '',
+    'operatorframework.io/initialization-resource': initializationResourceJSON,
   } = metadata.annotations || {};
+
+  let initializationResource = null;
+  if (initializationResourceJSON) {
+    try {
+      initializationResource = JSON.parse(initializationResourceJSON);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error.message);
+    }
+  }
 
   return (
     <>
@@ -784,6 +863,12 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
                   className="co-alert"
                   variant="danger"
                   title={`${status.phase}: ${status.message}`}
+                />
+              )}
+              {initializationResource && (
+                <InitializationResourceAlert
+                  initializationResource={initializationResource}
+                  csv={props.obj}
                 />
               )}
               <SectionHeading text="Provided APIs" />
@@ -1085,6 +1170,7 @@ export type CRDCardProps = {
   crd: CRDDescription | APIServiceDefinition;
   csv: ClusterServiceVersionKind;
   canCreate: boolean;
+  required?: boolean;
 };
 
 export type CRDCardRowProps = {
@@ -1147,6 +1233,11 @@ export type CSVSubscriptionProps = {
 
 type ClusterServiceVersionStateProps = {
   activeNamespace?: string;
+};
+
+type InitializationResourceAlertProps = {
+  csv: ClusterServiceVersionKind;
+  initializationResource: K8sResourceCommon;
 };
 
 type Header = {
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
@@ -24,6 +24,7 @@ import {
   k8sGet,
   k8sListPartialMetadata,
   kindForReference,
+  referenceFor,
   referenceForModel,
 } from '@console/internal/module/k8s';
 import { RadioGroup, RadioInput } from '@console/internal/components/radio';
@@ -97,6 +98,26 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
     currentCSVDesc.annotations?.['operatorframework.io/suggested-namespace'];
   const operatorRequestsMonitoring =
     currentCSVDesc.annotations?.['operatorframework.io/cluster-monitoring'] === 'true';
+  const initializationResourceJSON =
+    currentCSVDesc.annotations?.['operatorframework.io/initialization-resource'];
+
+  let initializationResourceReference = null;
+  if (initializationResourceJSON) {
+    let initializationResource = null;
+    try {
+      initializationResource = JSON.parse(initializationResourceJSON);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(
+        err.message || 'Operator Hub Subscribe: Could not get initialization resource.',
+      );
+    }
+
+    initializationResourceReference = initializationResource
+      ? referenceFor(initializationResource)
+      : null;
+  }
+
   const internalObjects = getInternalObjects(currentCSVDesc, 'annotations');
 
   const globalNS =
@@ -665,6 +686,7 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
                     canCreate={false}
                     crd={api}
                     csv={null}
+                    required={referenceForProvidedAPI(api) === initializationResourceReference}
                   />
                 ))
               )}
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
@@ -18,12 +18,7 @@ import {
   ResourceStatus,
   resourcePathFromModel,
 } from '@console/internal/components/utils';
-import {
-  groupVersionFor,
-  k8sPatch,
-  referenceForModel,
-  referenceForGroupVersionKind,
-} from '@console/internal/module/k8s';
+import { k8sPatch, modelFor, referenceForModel, referenceFor } from '@console/internal/module/k8s';
 import { errorModal } from '@console/internal/components/modals';
 import {
   ClusterServiceVersionModel,
@@ -45,7 +40,9 @@ import {
 } from '@console/shared/src/components/status/icons';
 import { InstallPlanPreview } from './install-plan';
 
-const getCRDFromObject = (obj: ClusterServiceVersionKind | InstallPlanKind | SubscriptionKind) => {
+const getInitializationResourceJSON = (
+  obj: ClusterServiceVersionKind | InstallPlanKind | SubscriptionKind,
+) => {
   return obj?.metadata?.annotations?.['operatorframework.io/initialization-resource'];
 };
 
@@ -68,14 +65,14 @@ const ViewInstalledOperatorsButton: React.FC<ViewOperatorButtonProps> = ({ names
 );
 
 const InstallFailedMessage: React.FC<InstallFailedMessageProps> = ({ namespace, csvName, obj }) => {
-  const crdMessage = getCRDFromObject(obj)
+  const initializationResourceMessage = getInitializationResourceJSON(obj)
     ? "The required custom resource can be created in the Operator's details view."
     : '';
 
   return (
     <>
       <h2 className="co-clusterserviceversion-install__heading">Operator installation failed</h2>
-      <p>The operator did not install successfully. {crdMessage}</p>
+      <p>The operator did not install successfully. {initializationResourceMessage}</p>
       <ActionGroup className="pf-c-form pf-c-form__group--no-top-margin">
         <Link to={resourcePathFromModel(ClusterServiceVersionModel, csvName, namespace)}>
           <Button variant="primary">View Error</Button>
@@ -118,28 +115,33 @@ const InstallNeedsApprovalMessage: React.FC<InstallNeedsApprovalMessageProps> = 
   </>
 );
 
-const CreateCRDButton: React.FC<CRDButtonProps> = ({ obj, disabled }) => {
-  if (!getCRDFromObject(obj)) {
+export const CreateInitializationResourceButton: React.FC<InitializationResourceButtonProps> = ({
+  obj,
+  disabled,
+  targetNamespace,
+}) => {
+  if (!getInitializationResourceJSON(obj)) {
     return null;
   }
-  let cr = null;
+  let initializationResource = null;
   try {
-    cr = JSON.parse(getCRDFromObject(obj));
+    initializationResource = JSON.parse(getInitializationResourceJSON(obj));
   } catch (error) {
     const errorMsg = error.message;
     errorModal({ errorMsg });
   }
-  const crNamespace = cr?.[0]?.metadata.namespace;
-  const apiVersion = cr?.[0]?.apiVersion;
-  const kind = cr?.[0]?.kind;
-  const { group, version } = groupVersionFor(apiVersion);
-  const reference = referenceForGroupVersionKind(group)(version)(kind);
+  const reference = referenceFor(initializationResource);
+  const model = modelFor(reference);
+  const initializationResourceNamespace = model?.namespaced
+    ? initializationResource?.metadata?.namespace || targetNamespace
+    : null;
+  const kind = initializationResource?.kind;
   return (
     <Link
       to={`${resourcePathFromModel(
         ClusterServiceVersionModel,
         obj.metadata.name,
-        crNamespace,
+        initializationResourceNamespace,
       )}/${reference}/~new`}
     >
       <Button isDisabled={disabled} variant="primary">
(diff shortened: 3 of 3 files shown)


frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -768,6 +769,7 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
   const {
     'marketplace.openshift.io/support-workflow': marketplaceSupportWorkflow,
     'olm.targetNamespaces': olmTargetNamespaces = '',
+    'operatorframework.io/initialization-resource': crInitializationResource,
```
spadgett, line 772: nit: I'd just call this `initializationResource`. It matches the annotation and whether it's a CR is unimportant.

```suggestion
    'operatorframework.io/initialization-resource': initializationResource,
```

zherman0:
Adding WIP until I add the logic to see if CR is already created.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -786,6 +806,17 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
                 />
               )}
+              {initializationResource && (
+                <Alert
+                  isInline
+                  className="co-alert"
+                  variant="warning"
+                  title="Custom resource required"
+                >
+                  <p>Create the required custom resource to be able to use this operator.</p>
```
TheRealJon, line 816: This is more of a design/content issue, but the wording of the title and content sound a bit weird to me. The phrase "to be able to" should just be "to", and IMO we should use the cr kind in the title and content. For example (made up a variable for the interpolated kind):
```suggestion
                  title={`${requiredCR.kind} Required`}
                >
                  <p>Create a ${requiredCR.kind} instance to use this operator.</p>
```

TheRealJon:
/lgtm cancel

Removing lgtm since this is now WIP. Since you are still addressing other issues, maybe my comment about the alert wording could be discussed/addressed.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
TheRealJon: @tlwu2013 @itsptk WDYT?

itsptk: Agreed. If we have access to the resource name we may as well use it. Thanks!

zherman0:
Logic added to check if a CR already exists instead of showing create button. Removing WIP.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
   );
 });
 
+const AlertToCreateCustomResource: React.SFC<AlertToCreateCustomResourceProps> = (props) => {
```
zherman0, line 787: Hooks cannot be in conditionals and I need to only call the hooks when we actually have a CustomResource in the operator, so I created a component here that only gets called when we have a CR.

zherman0:
@spadgett @TheRealJon - Added logic to only show create button if the CR is not already present. Ready for final review.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
spadgett: Let's call this `InitializationResourceAlert`, which matches the annotation and is more clear.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+};
+
 export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetailsProps> = (
   props,
 ) => {
   const { spec, metadata, status } = props.obj;
   const {
     'marketplace.openshift.io/support-workflow': marketplaceSupportWorkflow,
     'olm.targetNamespaces': olmTargetNamespaces = '',
+    'operatorframework.io/initialization-resource': initializationResource,
```
spadgett, line 824: Let's call this `initializationResourceJSON` and the parsed value `initializationResource` below.

```suggestion
    'operatorframework.io/initialization-resource': initializationResourceJSON,
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
   } = metadata.annotations || {};
 
+  let customResource = null;
```
spadgett, line 827:
```suggestion
  let initializationResource = null;
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,23 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
     currentCSVDesc.annotations?.['operatorframework.io/suggested-namespace'];
   const operatorRequestsMonitoring =
     currentCSVDesc.annotations?.['operatorframework.io/cluster-monitoring'] === 'true';
+  const operatorInitializationResouce =
```
spadgett, line 102:
```suggestion
  const initializationResourceJSON =
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,23 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+    currentCSVDesc.annotations?.['operatorframework.io/initialization-resource'];
+
+  let crReference = null;
```
spadgett, line 105:
```suggestion
  let initializationResourceReference = null;
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,23 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+  if (operatorInitializationResouce) {
+    let cr = null;
```
spadgett, line 107:
```suggestion
    let initializationResource = null;
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,23 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+    try {
+      cr = JSON.parse(operatorInitializationResouce);
+    } catch (err) {
+      setError(err.message || 'Could not get custom resource.');
+    }
+    if (cr) {
+      const { group: crGroup, version: crVersion } = groupVersionFor(cr?.[0]?.apiVersion);
```
spadgett, line 114:
```suggestion
      const { group: initializationResourceGroup, version: initializationResourceVersion } = groupVersionFor(initializationResource?.[0]?.apiVersion);
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -786,6 +850,9 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
                   title={`${status.phase}: ${status.message}`}
                 />
               )}
+              {customResource && (
+                <AlertToCreateCustomResource cr={customResource} obj={props.obj} />
```
spadgett, line 854:
```suggestion
                <InitializationResourceAlert initializationResource={initializationResource} obj={props.obj} />
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
spadgett: We should change `cr` to `initializationResource` throughout.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
spadgett: If `obj` is always a CSV, it would be better to call the prop `csv`. Otherwise it's confusing because there are two resources (the CSV and the initialization resource). `obj` is too vague.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1149,6 +1217,11 @@ type ClusterServiceVersionStateProps = {
   activeNamespace?: string;
 };
 
+type AlertToCreateCustomResourceProps = {
+  obj: ClusterServiceVersionKind;
+  cr: any;
```
spadgett, line 1222: It would be good to have a type definition for what we expect the annotation to be.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  const crDefinition = props.obj.spec?.customresourcedefinitions?.owned?.find(
+    (api) => api.kind === crName,
+  );
+  const crModelName = crDefinition?.name;
+  const { group: crGroup } = groupVersionFor(cr?.[0]?.apiVersion);
+  const { findModel } = useModelFinder();
+  const model = findModel(crGroup, crModelName.split('.').shift());
+
+  // Check if the CR is already present - only checks for the specific name/ns in the initialization-resource
+  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceKind>(
```
spadgett, line 800: Prefer `K8sResourceCommon` for unknown types. (It guards against trying to access `spec` and `status` values that might not be there.)

```suggestion
  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceCommon>(
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+    model,
+    cr?.[0]?.metadata.name,
+    cr?.[0]?.metadata.namespace,
+  );
+
+  if (!customResource && customResourceLoaded) {
+    return (
+      <Alert isInline className="co-alert" variant="warning" title={`${crName} Required`}>
+        <p>Create a {crName} instance to use this operator.</p>
+        <CreateCRDButton obj={props.obj} />
```
spadgett, line 810: Can we pass the parsed JSON and the model through here instead of parsing the annotation again inside of `CreateCRDButton`? There's a lot of code duplication around handling this annotation.

I'm not sure how `CreateCRDButton` works since it won't have the model.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,56 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  const { cr } = props;
+
+  const crName = cr?.[0]?.kind;
+  const crDefinition = props.obj.spec?.customresourcedefinitions?.owned?.find(
+    (api) => api.kind === crName,
+  );
+  const crModelName = crDefinition?.name;
+  const { group: crGroup } = groupVersionFor(cr?.[0]?.apiVersion);
+  const { findModel } = useModelFinder();
+  const model = findModel(crGroup, crModelName.split('.').shift());
```
spadgett, line 797: We need to handle cases where the model is not found.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
zherman0: Agreed.  The current `CreateCRDButton` component does need the obj for the name so I cannot remove obj completely.  I would like to refactor this function but I would need to refactor the install page component as well.  I will open a new Jira ticket to do those changes after this in order to prevent scope creep.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
zherman0: Is there a generic kind I can use?

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
zherman0: `useK8sGet` will survive a null model value.  In that case we just continue to show the `Create` button.  What would you prefer?

spadgett: I'd expect the link to be broken when the resource doesn't exist yet on the cluster. I suspect we have this issue on the install operator status as well.

We should also use `useAccessReview` here to only show the alert to users who have authority to create the resource.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
spadgett: > Is there a generic kind I can use?

Is there a design somewhere that describes the annotation?

It looks like we're treating it as an array in the code, but the issue says, "The "CR_TEMPLATE" will be a complete YAML template for the CR." It looks like the implementation doesn't match the epic description.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
zherman0: For all the links, we use `referenceForGroupVersionKind` to build the URL, so if the reference does not exist, then we get 404.  We should probably discuss if there is anything we can do in those situations further, but preferably out of the scope of this PR.

zherman0: `useAccessReview` has been added.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,73 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  const { findModel } = useModelFinder();
+  const model = findModel(
+    initializationResourceGroup,
+    initializationResourceModelName.split('.').shift(),
+  );
+  // Check if the CR is already present - only checks for the specific name/ns in the initialization-resource
+  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceCommon>(
+    model,
+    initializationResource?.metadata.name,
+    initializationResource?.metadata.namespace || csv.metadata?.namespace,
```
spadgett, line 807: You should check if it's a namespaced resource in `model`. You only need a namespace if it's a namespaced resource.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,73 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  );
+  const initializationResourceModelName = initializationResourceDefinition?.name;
+  const { group: initializationResourceGroup } = groupVersionFor(
+    initializationResource?.apiVersion,
+  );
+  const { findModel } = useModelFinder();
+  const model = findModel(
+    initializationResourceGroup,
+    initializationResourceModelName.split('.').shift(),
+  );
```
spadgett, line 802: You have the sample resource, right? Can you just call `referenceFor(initializationResource)` followed by `modelFor`? Then you don't need `useModelFinder`.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,73 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  // Check if the CR is already present - only checks for the specific name/ns in the initialization-resource
+  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceCommon>(
+    model,
+    initializationResource?.metadata.name,
+    initializationResource?.metadata.namespace || csv.metadata?.namespace,
+  );
+  const canCreateCustomResource = useAccessReview({
+    group: initializationResourceGroup,
+    resource: initializationResourceModelName.split('.').shift(),
```
spadgett, line 811:
```suggestion
    resource: model.plural,
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,73 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+    verb: 'create',
```
spadgett, line 812: You need to add a namespace as well (unless it's a cluster-scoped resource).

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +784,73 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+    'operatorframework.io/initialization-resource': initializationResourceJSON,
   } = metadata.annotations || {};
 
+  let initializationResource = null;
+  if (initializationResourceJSON) {
+    try {
+      initializationResource = JSON.parse(initializationResourceJSON);
+    } catch (error) {
+      const errorMsg = error.message;
+      errorModal({ errorMsg });
```
spadgett, line 850: I wouldn't show a modal because the user can't do anything. Just use `console.error`

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1149,6 +1237,11 @@ type ClusterServiceVersionStateProps = {
   activeNamespace?: string;
 };
 
+type InitializationResourceAlertProps = {
+  csv: ClusterServiceVersionKind;
+  initializationResource: any;
```
spadgett, line 1242:
```suggestion
  initializationResource: K8sResourceCommon;
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,29 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+  const initializationResourceJSON =
+    currentCSVDesc.annotations?.['operatorframework.io/initialization-resource'];
+
+  let initializationResourceReference = null;
+  if (initializationResourceJSON) {
+    let initializationResource = null;
+    try {
+      initializationResource = JSON.parse(initializationResourceJSON);
+    } catch (err) {
+      setError(err.message || 'Could not get custom resource.');
```
spadgett, line 111: I'd use `console.log` here as well.

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +99,29 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+    }
+
+    if (initializationResource) {
+      const {
+        group: initializationResourceGroup,
+        version: initializationResourceVersion,
+      } = groupVersionFor(initializationResource?.apiVersion);
+      initializationResourceReference = referenceForGroupVersionKind(initializationResourceGroup)(
```
spadgett, line 119: You should be able to use `referenceFor(initializationResource)`

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -45,7 +45,9 @@ import {
 } from '@console/shared/src/components/status/icons';
 import { InstallPlanPreview } from './install-plan';
 
-const getCRDFromObject = (obj: ClusterServiceVersionKind | InstallPlanKind | SubscriptionKind) => {
+const getInitializationResourceJSONFromCSV = (
```
spadgett, line 48: This says from `CSV` but it takes other types. Should this only take `ClusterServiceVersionKind`?

I would just call this `getInitializationResourceJSON`

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -118,28 +120,33 @@ const InstallNeedsApprovalMessage: React.FC<InstallNeedsApprovalMessageProps> =
...
   }
-  const crNamespace = cr?.[0]?.metadata.namespace;
-  const apiVersion = cr?.[0]?.apiVersion;
-  const kind = cr?.[0]?.kind;
+  const initializationResourceNamespace =
+    initializationResource?.metadata?.namespace || targetNamespace;
+  const apiVersion = initializationResource?.apiVersion;
+  const kind = initializationResource?.kind;
   const { group, version } = groupVersionFor(apiVersion);
   const reference = referenceForGroupVersionKind(group)(version)(kind);
```
spadgett, line 143: `referenceFor(initializationResource)`

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -149,24 +156,30 @@ const CreateCRDButton: React.FC<CRDButtonProps> = ({ obj, disabled }) => {
...
+  const initializationResourceKind = initializationResource?.kind;
+  const initializationResourceNamespace = initializationResource?.metadata?.namespace;
   const description = obj?.metadata?.annotations?.description;
   return (
     <div className="co-clusterserviceversion__box">
       <span className="co-resource-item">
-        <ResourceLink kind={crdKind} name={crdKind} namepace={crNamespace} />
+        <ResourceLink
+          kind={initializationResourceKind}
+          name={initializationResourceKind}
```
spadgett, line 180: Not new, but this seems wrong. Why are we passing kind as the `name` prop?

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -149,24 +156,30 @@ const CreateCRDButton: React.FC<CRDButtonProps> = ({ obj, disabled }) => {
   );
 };
 
-const InstallPageCRDRequiredMessage: React.FC<InstallCRDMessageProps> = ({ obj }) => {
-  if (!getCRDFromObject(obj)) {
+const InstallPageInitializationResourceRequiredMessage: React.FC<InstallInitializationResourceMessageProps> = ({
```
spadgett, line 159: Maybe just `InitializationResourceRequiredMessage`?

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
zherman0: For the examples I have, the kind is `StorageCluster` while the name is `portworx` or `example-storagecluster`.  I want the Link to show the kind name.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +782,67 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  const { group: initializationResourceGroup } = groupVersionFor(
+    initializationResource?.apiVersion,
+  );
+  const model = modelFor(referenceFor(initializationResource));
+
+  // Check if the CR is already present - only checks for the specific name/ns in the initialization-resource
+  const [customResource, customResourceLoaded] = useK8sGet<K8sResourceCommon>(
+    model,
+    initializationResource?.metadata.name,
+    model.namespaced ? initializationResource?.metadata.namespace || csv.metadata?.namespace : '',
```
spadgett, line 798: The `?` should either go directly after `csv` or should be removed.

```suggestion
    model.namespaced ? initializationResource?.metadata.namespace || csv.metadata.namespace : '',
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +782,67 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+  );
+  const canCreateCustomResource = useAccessReview({
+    group: initializationResourceGroup,
+    resource: model.plural,
+    namespace: initializationResource?.metadata.namespace || csv.metadata?.namespace,
```
spadgett, line 803: Need to check `model.namespaced` here.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
spadgett: You need to handle the case where `model` is null. This is a runtime error otherwise.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -761,15 +782,67 @@ export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProp
...
+      <Alert
+        isInline
+        className="co-alert"
+        variant="warning"
+        title={`${initializationResourceKind} Required`}
+      >
+        <p>Create a {initializationResourceKind} instance to use this operator.</p>
+        <CreateInitializationResourceButton
+          obj={props.csv}
+          targetNamespace={initializationResource?.metadata.namespace || csv.metadata?.namespace}
```
spadgett, line 818: Need to check `model.namespaced`

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -97,6 +98,26 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(
+        err.message || 'Operator Hub Subscribe: Could not get initialization resource.',
+      );
+    }
+
+    if (initializationResource) {
+      initializationResourceReference = referenceFor(initializationResource);
+    }
```
spadgett, line 118: We can make this a `const`

```ts
const initializationResourceReference = initializationResource ? referenceFor(initializationResource) : null;
```

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -118,28 +115,31 @@ const InstallNeedsApprovalMessage: React.FC<InstallNeedsApprovalMessageProps> =
...
     const errorMsg = error.message;
     errorModal({ errorMsg });
   }
-  const crNamespace = cr?.[0]?.metadata.namespace;
-  const apiVersion = cr?.[0]?.apiVersion;
-  const kind = cr?.[0]?.kind;
-  const { group, version } = groupVersionFor(apiVersion);
-  const reference = referenceForGroupVersionKind(group)(version)(kind);
+  const initializationResourceNamespace =
+    initializationResource?.metadata?.namespace || targetNamespace;
```
spadgett, line 134: We should be checking `model.namespaced`. (There's some code duplication here, which makes things harder.)

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
zherman0: This definition is inside the conditional check `if (initializationResourceJSON)` so I still need to define the value outside of that block for the code later on.  I went ahead and made this ternary though.

zherman0:
@spadgett - Ready for another review, thanks.

frontend/packages/operator-lifecycle-manager/src/components/operator-install-page.tsx
```diff
@@ -118,28 +115,33 @@ const InstallNeedsApprovalMessage: React.FC<InstallNeedsApprovalMessageProps> =
...
+    ? initializationResource?.metadata?.namespace || targetNamespace
+    : null;
+  const kind = initializationResource?.kind;
   return (
     <Link
       to={`${resourcePathFromModel(
         ClusterServiceVersionModel,
         obj.metadata.name,
-        crNamespace,
+        initializationResourceNamespace,
```
spadgett, line 144: This isn't quite right if the initialization resource is from a different namespace than the CSV. The operand details form won't work in that case. We probably need to link to a YAML editor.

@tlwu2013 FYI

spadgett: It sounds like I'm mistaken and that this works.

PR #6119, opened 2020-07-27
Bug 1859256: Fix blank page/error on ImageManifestVulnPage

rebeccaalpert:
There was no check for an undefined props.match in ImageManifestVulnPage, leading to a blank page/error message. Adding a check seems to fix the issue.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1859256.

### DIFF
--- a/frontend/packages/container-security/src/components/image-manifest-vuln.tsx
+++ b/frontend/packages/container-security/src/components/image-manifest-vuln.tsx
@@ -302,16 +302,14 @@ export const ImageManifestVulnList: React.FC<ImageManifestVulnListProps> = (prop
 };
 
 export const ImageManifestVulnPage: React.FC<ImageManifestVulnPageProps> = (props) => {
-  const namespace = props.match.params?.ns;
-
   return (
     <MultiListPage
       {...props}
-      namespace={namespace}
+      namespace={props.namespace}
       resources={[
         {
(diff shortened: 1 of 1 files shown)


frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -302,7 +302,7 @@ export const ImageManifestVulnList: React.FC<ImageManifestVulnListProps> = (prop
 };
 
 export const ImageManifestVulnPage: React.FC<ImageManifestVulnPageProps> = (props) => {
-  const namespace = props.match.params?.ns;
+  const namespace = props.match?.params?.ns;
```
spadgett, line 305: This fixes the runtime error, but the list will presumably show vulnerabilities for *all projects*, not just the current project. We should compare this to other list views to see what's different. I believe the namespace is passed through in a different way for other pages.

rebeccaalpert: I looked through the other components using MultiListPage and they're either getting the namespace passed in as a prop directly or using props.match.params.ns.

rebeccaalpert:
I updated the namespace parameter to use props.namespace. It seems to fix the issue and allow for the correct behavior.

PR #6120, opened 2020-07-27
Bug 1849555: Clarify VM import messages

pcbailey:
This PR clarifies the type of VM import that is in progress in user-facing messages.


pcbailey:
@suomiy Please review.

PR #6121, opened 2020-07-27
Bug 1850513: Disable stop action in VM menu for in process CDI imports

pcbailey:
Backend does not support stopping VMs that are still importing. This PR disables the stop action during the import process.

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
@@ -102,9 +102,14 @@ export const menuActionStart = (
   };
 };
 
-const menuActionStop = (kindObj: K8sKind, vm: VMKind, { vmi }: ActionArgs): KebabOption => {
+const menuActionStop = (
+  kindObj: K8sKind,
+  vm: VMKind,
+  { vmi, vmStatusBundle }: ActionArgs,
+): KebabOption => {
   const title = 'Stop Virtual Machine';
   return {
+    isDisabled: vmStatusBundle?.status?.isPending(),
     hidden: !isVMExpectedRunning(vm),
     label: title,
     callback: () =>


frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
```diff
@@ -102,9 +102,14 @@ export const menuActionStart = (
...
 
-const menuActionStop = (kindObj: K8sKind, vm: VMKind, { vmi }: ActionArgs): KebabOption => {
+const menuActionStop = (
+  kindObj: K8sKind,
+  vm: VMKind,
+  { vmi, vmStatusBundle }: ActionArgs,
+): KebabOption => {
   const title = 'Stop Virtual Machine';
   return {
+    isDisabled: vmStatusBundle?.status?.isPending(),
```
suomiy, line 112: why not use `isImporting` instead? `isPending` has overlap with `VMI_WAITING` which does not have anything to do with importing flow.
Plus this does not include for example `CDI_IMPORTING`.

also just to be sure (not really against it) are we really introducing isDisabled semantic into our actions? user's are used to having only actions which they can execute.

pcbailey: I discussed the issue with @matthewcarleton and he would like to move toward disabling actions that are unavailable. I understand both perspectives, but think this is an issue worth discussing further either in a meeting or via some other venue than these comments to make sure it gets the discussion it deserves. Thoughts?

Regarding which status(es) I'm disabling here, I see your point. I'm currently working on https://bugzilla.redhat.com/show_bug.cgi?id=1843532, which is somewhat related. I will include any updates in that PR and let you know when it's posted so we can discuss further, if needed.

suomiy: > I discussed the issue with @matthewcarleton and he would like to move toward disabling actions that are unavailable. I understand both perspectives, but think this is an issue worth discussing further either in a meeting or via some other venue than these comments to make sure it gets the discussion it deserves. Thoughts?
> 

sure, we can discuss it in a meeting

> Regarding which status(es) I'm disabling here, I see your point. I'm currently working on https://bugzilla.redhat.com/show_bug.cgi?id=1843532, which is somewhat related. I will include any updates in that PR and let you know when it's posted so we can discuss further, if needed.

+1

PR #6125, opened 2020-07-28
Bug 1844944: No default storage class in add disk 

yaacov:
Fix: No default storage class in add disk modal

Screenshots:
Add a new disk, has the default storageClass pre selected:
screenshot-localhost_9000-2020 08 05-15_10_13

Edit a disk with no storageClass, the missing storageClass is not fixed:
screenshot-localhost_9000-2020 08 05-15_10_38

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
@@ -333,6 +333,8 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
     }
   };
 
+  const isStorageClassDataLoading = !isLoaded(storageClasses) || !isLoaded(_storageClassConfigMap);
+
   return (
     <div className="modal-content">
       <ModalTitle>
@@ -531,7 +533,7 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
             <K8sResourceSelectRow
               key="storage-class"
               id={asId('storage-class')}
-              isDisabled={isDisabled('storageClass') || !isLoaded(storageClasses)}
+              isDisabled={isDisabled('storageClass') || isStorageClassDataLoading}
               name={storageClassName}
               data={storageClasses}
               model={StorageClassModel}
@@ -559,7 +561,7 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
                     onChange={(vMode) => setVolumeMode(VolumeMode.fromString(vMode))}
                     value={asFormSelectValue(volumeMode?.getValue())}
                     id={asId('volume-mode')}
-                    isDisabled={isDisabled('volumeMode')}
+                    isDisabled={isDisabled('volumeMode') || isStorageClassDataLoading}
                   >
                     <FormSelectPlaceholderOption
                       isDisabled={inProgress}
@@ -585,7 +587,7 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
                     onChange={(aMode) => setAccessMode(AccessMode.fromString(aMode))}
                     value={asFormSelectValue(accessMode?.getValue())}
                     id={asId('access-mode')}
-                    isDisabled={isDisabled('accessMode')}
+                    isDisabled={isDisabled('accessMode') || isStorageClassDataLoading}
                   >
                     <FormSelectPlaceholderOption
                       isDisabled={inProgress}
--- a/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
@@ -19,7 +19,7 @@ import {
   PersistentVolumeClaimModel,
   StorageClassModel,
 } from '@console/internal/models';
-import { getName, getAnnotations } from '@console/shared/src';
+import { getName } from '@console/shared/src';
 import { getLoadedData, isLoaded, prefixedID, resolveDataVolumeName } from '../../../utils';
 import { validateDisk } from '../../../utils/validations/vm';
 import { isValidationError } from '../../../utils/validations/common';
@@ -31,6 +31,7 @@ import {
 import {
   getDefaultSCAccessModes,
   getDefaultSCVolumeMode,
+  getGefaultStorageClass,
 } from '../../../selectors/config-map/sc-defaults';
 import {
   ADD,
@@ -46,7 +47,6 @@ import { DiskWrapper } from '../../../k8s/wrapper/vm/disk-wrapper';
 import { DataVolumeWrapper } from '../../../k8s/wrapper/vm/data-volume-wrapper';
 import { VolumeWrapper } from '../../../k8s/wrapper/vm/volume-wrapper';
 import { AccessMode, DiskBus, DiskType, VolumeMode } from '../../../constants/vm/storage';
-import { DEFAULT_SC_ANNOTATION } from '../../../constants/sc';
 import { getPvcStorageSize } from '../../../selectors/pvc/selectors';
 import { K8sResourceSelectRow } from '../../form/k8s-resource-select-row';
 import { SizeUnitFormRow } from '../../form/size-unit-form-row';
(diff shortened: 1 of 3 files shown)


yaacov:
@matthewcarleton @suomiy @pcbailey @irosenzw @glekner please review

suomiy:
this is a nice improvement, but it causes incosistencies

eg.
1. create a template with no sc/default sc datavolume
2. create a vm with this template
3. you will see in review that no sc is selected
4. but the disk modal behaves like the default sc is the real target sc, thus giving incorrect information

- also cancel/save with no change have a different meaning - modal should be idempotent in this case
- it should be possible to select no sc

yaacov:
> this is a nice improvement, but it causes incosistencies
> 
> eg.
> 
>     1. create a template with no sc/default sc datavolume

The console does not allow creating PVCs without SC from the UI, we should not let users create a VMIwith PVCs that are lacking SC.
 
>     4. but the disk modal behaves like the default sc is the real target sc, thus giving incorrect information

The SC used is the default SC so this is the correct information
 
>     * also cancel/save with no change have a different meaning - modal should be idempotent in this case

Pressing the save button will restore the PVC to a valid state with the default SC in the yaml,
while pressing cancel will leave the SC empty in the yaml ( uing the default SC )

>     * it should be possible to select no sc

No, if the console does not allow it, we should no allow it too. another reason may be to not let
users create unusable PVC with none valid SC

yaacov:
@matthewcarleton @jelkosz we need you help here ^^ can you comment ?

suomiy:
for completeness

> > this is a nice improvement, but it causes incosistencies
> > eg.
> > ```
> > 1. create a template with no sc/default sc datavolume
> > ```
> 
> The console does not allow creating PVCs without SC from the UI, we should not let users create a VMIwith PVCs that are lacking SC.

kubevirt allows it and some flows depend on it - templates/v2v
for example you might not want to distribute a template with a specific class. Instead you might want an empty SC and just have VMs pick up the default one from the environment they are in

> 
> > ```
> > 4. but the disk modal behaves like the default sc is the real target sc, thus giving incorrect information
> > ```
> 
> The SC used is the default SC so this is the correct information
>
> Pressing the save button will restore the PVC to a valid state with the default SC in the yaml,
> while pressing cancel will leave the SC empty in the yaml ( uing the default SC )

the final YAMLs will be different and can cause unwanted behaviours
> 
> > ```
> > * it should be possible to select no sc
> > ```
> 
> No, if the console does not allow it, we should no allow it too. another reason may be to not let
> users create unusable PVC with none valid SC

this depends on the use case

matthewcarleton:
are we still pursuing this @yaacov?  It sounded like we were dropping it after the call this morning.

yaacov:
> are we still pursuing this @yaacov? It sounded like we were dropping it after the call this morning.

@matthewcarleton yes, it's a BZ so we need to do something ... I'll thing of something that will make everyone happy :-)

yaacov:
@suomiy please re-review

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -141,6 +142,17 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
     combinedDisk.getStorageClassName() || '',
   );
 
+  React.useEffect(() => {
+    if (!isEditing) {
```
suomiy, line 146: ok seems reasonable, but let's check if isLoaded here

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -141,6 +142,17 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
...
+      const defaultStorageClass = getLoadedData(storageClasses, []).find(
+        (sc) => getAnnotations(sc, {})[DEFAULT_SC_ANNOTATION] === 'true',
```
suomiy, line 148: this is already used once in the codebase. Having a `isDefaultSC(storageClass)` function would be nice

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
suomiy: this component supports passing half initialized values when creating so lets's also check if storageClassName is not empty as well

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -141,6 +142,17 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
...
+      );
+
+      setStorageClassName(getName(defaultStorageClass) || '');
```
suomiy, line 151: let's check if not empty before calling it

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -141,6 +142,17 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isLoaded(storageClasses)]);
+
   const [size, setSize] = React.useState<string>(
     combinedDiskSize ? `${combinedDiskSize.value}` : '',
   );
```
suomiy, line 158: we should disable the storageClass field until storageClasses are loaded

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
suomiy: hm, this should be combined with storageClassConfigMap useEffect since they depend on each other and should set values accordingly

yaacov: added the checks for isLoaded and storageClassName +1
moved inside storageClassConfigMap useEffect +2 :-)

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
yaacov: moved to a common class

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
yaacov: checking :t-rex:

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
yaacov: storageClass dropdown is now disabled until all sc are loaded

yaacov:
@suomiy please re-review

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -167,8 +168,17 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
         setAccessMode(getDefaultSCAccessModes(storageClassConfigMap)[0]);
```
suomiy, line 168: we should pass SC to get the correct access modes and volume modes

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -521,7 +531,7 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
             <K8sResourceSelectRow
               key="storage-class"
               id={asId('storage-class')}
-              isDisabled={isDisabled('storageClass')}
+              isDisabled={isDisabled('storageClass') || !isLoaded(storageClasses)}
```
suomiy, line 534: since this can affect volume/access modes it should wait also for storageClassConfigMap

we should also do the same thing for access/volume modes fields. 

It would be nice to have a common variable for that, or we could add if/switch statement to the isDisabled function and do the same thing by fieldName.

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
suomiy: hm, now I see we have a very similar logic in onStorageClassNameChanged, but sligtly different. Maybe we could extract the logic on lines 287-294 into a function?

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
yaacov: done

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
yaacov: a - using the sc now to pick the access and volume
b - the logic is far enough to make merging the onChange and onInit code look forced

suomiy: ok

PR #6126, opened 2020-07-28
Add validation for IP Family match

bipuladh:
Screenshot from 2020-07-28 13-48-51

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/independent-mode/install.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/independent-mode/install.tsx
@@ -14,6 +14,7 @@ import {
   k8sCreate,
   referenceForModel,
   apiVersionForModel,
+  k8sList,
 } from '@console/internal/module/k8s';
 import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager';
 import {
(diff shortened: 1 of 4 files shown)


frontend/packages/ceph-storage-plugin/src/components/independent-mode/install.tsx
```diff
@@ -52,10 +53,19 @@ const CreateExternalCluster = withHandlePromise((props: CreateExternalClusterPro
...
   const dispatch = useDispatch();
 
   const plainKeys = _.concat(configMaps, storageClasses);
 
+  React.useEffect(() => {
+    // eslint-disable-next-line promise/catch-or-return
+    k8sList(PodModel).then((pods) => {
+      const address = pods[0].status.podIP;
+      setIPFamily(getIPFamily(address));
+    });
```
gnehapk, line 66: instead of disabling the eslint, you can do `setIPFamily(IP_FAMILY.IPV4` in catch block. We should disable eslint only if required.

bipuladh: but it's already IPv4 by default. I don't have good enough reason to add a catch block.

bipuladh:
/hold
frontend tests are failing

PR #6128, opened 2020-07-28
Start moving PromQL queries to a central location

kyoto:
The intention is to begin moving all of Console's PromQL queries (including those under `frontend/packages/`) to a central location over time. This will allow the Monitoring team to help with reviewing queries as they are added and modified.


frontend/packages/console-shared/src/promql/OWNERS
```diff
@@ -0,0 +1,16 @@
...
+  - spadgett
+approvers:
+  - bparees
+  - christianvogt
+  - jhadvig
+  - kyoto
+  - rawagner
+  - rhamilto
+  - spadgett
+  - vojtechszocs
```
spadgett, line 16: Most of the approvers are listed in `frontend/OWNERS`, so they don't all need to be listed explicitly again here. I'm happy adding some folks from the monitoring team to approvers for this folder if you like.

kyoto: 👍 Updated

PR #6130, opened 2020-07-28
Add ability to hide all groups in topology

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4317

**Description**
As a developer, I want to view my “Grouped Resources”  as flat workloads, hiding the grouping, in the Topology View.

In this mode:
- No groups are shown
- Children of groups are shown as individual nodes
- the corresponding workload node for grouped nodes are shown (ie the Revisions for a knative service are shown as well as the deployment that backs it).
- Edges are not shown
- Edges are not allowed to be created

**Screenshots**

FlatGroups

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

/kind feature


jeff-phillips-18:
Updated to disable the expand options and added unit tests for them.

andrewballantyne:
HideGroupsByDesignHideNodes

I feel this is a little odd... but I can also bend my brain to understand it. @bgliwa01 By design?

bgliwa01:
> HideGroupsByDesignHideNodes
> 
> I feel this is a little odd... but I can also bend my brain to understand it. @bgliwa01 By design?

@andrewballantyne yes this is correct implementation of the design

jeff-phillips-18:
@andrewballantyne @bgliwa01 Actually it is not the correct implementation. The children should be shown. It should not matter the order in which things were selected. However, there is an issue with PF topology that prevents nodes whose parent no longer exists from being re-rendered to the graph. Have a PR up to fix it: https://github.com/patternfly/patternfly-react/pull/4629

bgliwa01:
@andrewballantyne @jeff-phillips-18 I see what you are saying now, I thought it was around how expand becomes disabled. +1 to everything jeff said

christianvogt:
Can someone explain to me why edges are hidden and creation is disabled when the user chooses to hide groupings? Why are groups and edges even related?

invincibleJai:
Verified locally, seems to work as expected

PR #6139, opened 2020-07-28
Continue to show Workers progress bar if not 100% when CV is updated

rhamilto:
88720100-51b99000-d0f2-11ea-92ea-b96cbad83f1f copy

Video of switch:  https://youtu.be/KglxwTcOeYI

Bonus fix: update Worker Nodes help icon and popover

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = {
     browser: true,
     es6: true,
     jasmine: true,
+    jest: true,
   },
   extends: [
     'eslint:recommended',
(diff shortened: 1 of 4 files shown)


megan-hall:
lgtm!

megan-hall:
One small question - should we just link the word Nodes and not Master Nodes and Worker Nodes - just so it is clear that its the complete Nodes list and not filtered? 

Other than that lgtm!!

rhamilto:
> One small question - should we just link the word Nodes and not Master Nodes and Worker Nodes - just so it is clear that its the complete Nodes list and not filtered?
> 
> Other than that lgtm!!

The links are to the filtered Nodes list.

PR #6141, opened 2020-07-29
Added quick start controller and complete flow

rohitkrai03:
**Fixes**:  https://issues.redhat.com/browse/ODC-4133

Also adds changes from https://issues.redhat.com/browse/ODC-4201.

**Analysis / Root cause**: 
- All the previous stories that were part of quick starts epic were only pure components that rendered stuff based on props. No real state was being used and there was no controller to change any state.

**Solution Description**: 
This PR - 
- Brings together all the previous work for the quick start epic.
- Adds actions and reducers for quick start state along with ability to persist state in localstorage.
- Adds a controller that controls the whole state of quick start tours.
- Refactors and cleans up a ton of previous code to make sure it all works correctly with the actual state.
- Completes the whole quick start flow.

@openshift/team-devconsole-ux @beaumorley @serenamarie125 

Quick Starts

Screenshot from 2020-07-29 09-37-15

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


beaumorley:
@rohitkrai03 

A few things. It looks like the copy is wrong on the Quick Starts card on the Add page. The first link should be Create Serverless App and should link to the Serverless Applications quick start.  The Explore Serverless quick start should not show up on the Dev console side at all since it about adding an operator.

So the card will look like this:
----------------------
**Quick Starts**

Create a Serverless App  [goes to Serverless Applications]
Build a Pipeline  [goes to Build Pipelines]
Create a Sample App [goes to a sample Serena is working on we need to verify this one with Serena]

See all quick starts -->  [can we move the arrow to the end]
------------------------

The interaction of the side panel and the responsiveness look great.

If we have copy edits, will those go in as bugs?

Thank you!

@bkrikori can you add the card to this PR so they can see the final design?

cc @serenamarie125

bkrikori:
+Add page@2x
Dev Add Page (kebab)@2x
Just updated the Quick Starts card - "See all quick starts" now comes before the arrow on the bottom of the card.

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -22,26 +31,51 @@ type DispatchProps = {
...
-  // const tourCompleteAlert = (
-  //   <Alert variant="success" isInline title="This tour has already been completed" />
-  // );
+
+  const panelContent = quickStart ? (
+    <DrawerPanelContent>
+      <DrawerHead>
+        <div className="co-quick-start-drawer__title">
+          <Title headingLevel="h1" size="xl">
+            {quickStart?.name}
```
invincibleJai, line 45: nit: as in line 40 already checked for `quickStart` so can avoid here
```suggestion
            {quickStart.name}
```

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartCatalog.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type QuickStartCatalogProps = OwnProps & DispatchProps & StateProps;
+
+const QuickStartCatalog: React.FC<QuickStartCatalogProps> = ({
+  quickStarts,
+  activeQuickStartID,
+  allQuickStartStates,
+  setActiveQuickStart,
+}) => {
+  return (
```
invincibleJai, line 39: nit: can do implicit return here

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTileHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+          variant="outline"
+          color={statusColorMap[status]}
+          icon={<StatusIcon status={status} />}
+        >
+          {status}
+        </Label>
+      )}
+      <Label variant="outline" icon={<OutlinedClockIcon />}>
+        {`${duration} minutes`}
+      </Label>
```
invincibleJai, line 36: will duration be shown even for `QuickStartStatus.NOT_STARTED`?

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTasks.tsx
```diff
@@ -0,0 +1,65 @@
...
+}) => (
+  <>
+    {tasks
+      .filter((_, index) => index <= taskNumber)
+      .map((task, index) => {
+        const { title, description, review, recapitulation, taskHelp } = task;
+        const isActiveTask = index === taskNumber;
+        const taskStatus = allTaskStatuses[index];
+        return (
+          <React.Fragment key={task.title}>
```
invincibleJai, line 32: nit
```suggestion
          <React.Fragment key={title}>
```

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-utils.ts
```diff
@@ -1,21 +1,10 @@
...
-  return items.map((item) => {
-    return {
-      ...item,
-      ...mockStatus[item.name],
-      unmetPrerequisite: mockPrerequisiteStatus[item.name],
-    };
-  });
-};
+export const getQuickStart = (quickStartID: string): QuickStart =>
+  mockQuickStarts.find((qs) => qs.id === quickStartID);
```
invincibleJai, line 7: just curious , will it be mock only for 4.6?

frontend/packages/console-app/src/redux/reducers/quick-start-reducer.ts
```diff
@@ -1,31 +1,143 @@
...
 
 export const quickStartReducerName = 'quickStart';
 
-type State = {
-  isExpanded: boolean;
-  activeQuickStartID: string;
-};
+export const QUICK_START_REDUX_STATE_LOCAL_STORAGE_KEY = 'bridge/quick-start-redux-state';
+
+type State = Map<string, any>;
```
andrewballantyne, line 15:
```suggestion
type State = Map<string, { activeQuickStartId: string }>;
```
I think we know more than `any`. It's a constructed state that we are saving... can we elaborate the type? I added the only one I saw in this file.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
```diff
@@ -88,7 +88,7 @@ const QuickStartTaskHeader: React.FC<QuickStartTaskHeaderProps> = ({
         </>
       }
       step={taskIndex}
-      onNavItemClick={onTaskSelect}
+      onNavItemClick={() => onTaskSelect(taskIndex - 1)}
       navItemComponent="Button"
```
andrewballantyne, line 92: This does not seem supported. I get a React error:

image

andrewballantyne:
Don't think this is a question anyone has asked yet... but I think we need to look at improving the content so it's better numbered:
image

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
rohitkrai03: That's a patternfly component warning. Not related to code that i have.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const resetQuickStart = () => setQuickStartStatus(id, QuickStartStatus.NOT_STARTED);
+
+  const handleQuickStartChange = (quickStartId: string) => setActiveQuickStart(quickStartId);
+
+  const handleTaskStatusChange = (newTaskStatus: QuickStartTaskStatus) =>
+    setQuickStartTaskStatus(newTaskStatus);
+
+  const handleNext = () => {
+    const totalTasks = tasks?.length;
```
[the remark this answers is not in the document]
rohitkrai03: Number of tasks is not going to change during the render cycle. So don't think there will be any issues with stale closures.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -0,0 +1,169 @@
...
+      return handleTaskStatusChange(QuickStartTaskStatus.REVIEW);
+
+    if (taskNumber < totalTasks) return setQuickStartTaskNumber(taskNumber + 1);
+
+    return null;
+  };
+
+  const handleBack = () => {
+    if (status === QuickStartStatus.COMPLETE && taskNumber === tasks.length - 1)
+      return startQuickStart();
```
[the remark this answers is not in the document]
rohitkrai03: Yup. Back should work in exact same sequence as the next works. So, going back from complete will make it in progress.

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTileHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+type QuickStartTileHeaderProps = {
+  status: string;
+  duration: number;
+  name: string;
+};
+
+const statusColorMap = {
+  [QuickStartStatus.COMPLETE]: 'green',
+  [QuickStartStatus.IN_PROGRESS]: 'purple',
+  [QuickStartStatus.NOT_STARTED]: 'grey',
```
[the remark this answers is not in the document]
rohitkrai03: It's actually not CSS colors. It's for the `color` prop of Patternfly `<Label />` component.

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTileHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+        </Label>
+      )}
+      <Label variant="outline" icon={<OutlinedClockIcon />}>
+        {`${duration} minutes`}
+      </Label>
+    </div>
+  </div>
+);
+
+// there could be various statuses in future
```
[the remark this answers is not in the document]
rohitkrai03: Will remove. Old comment went unnoticed.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
```diff
@@ -0,0 +1,65 @@
...
+      return PrimaryButtonText.CLOSE;
+    case QuickStartStatus.IN_PROGRESS:
+      return PrimaryButtonText.NEXT;
+    default:
+      return PrimaryButtonText.START;
+  }
+};
+
+const QuickStartFooter: React.FC<QuickStartFooterProps> = ({ status, onNext, onBack }) => (
+  <div className="co-quick-start-footer">
```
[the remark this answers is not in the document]
rohitkrai03: The styles are coming from older implementation. I can convert it to flex. Just didn't wanted to spend much time on it since it worked fine.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-types.ts
```diff
@@ -0,0 +1,45 @@
...
+  IN_PROGRESS = 'In Progress',
+  NOT_STARTED = 'Not started',
+}
+
+export enum QuickStartTaskStatus {
+  INIT = 'Initial',
+  REVIEW = 'Review',
+  SUCCESS = 'Success',
+  FAILED = 'Failed',
+}
```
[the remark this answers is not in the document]
rohitkrai03: Just have these two. So thought it's okay to have then in types since we do actually use them as types as well.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskReview.tsx
```diff
@@ -0,0 +1,63 @@
...
+        isChecked={taskStatus === QuickStartTaskStatus.SUCCESS}
+        className="co-quick-start-task__radio-button-field"
+      />
+      <Radio
+        name="review-negative"
+        onChange={() => onTaskReview(QuickStartTaskStatus.FAILED)}
+        label="No"
+        id="review-negative"
+        isChecked={taskStatus === QuickStartTaskStatus.FAILED}
+        className="co-quick-start-task__radio-button-field"
```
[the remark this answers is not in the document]
rohitkrai03: Just some leftover from the refactoring. Forgot to split the styles.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
rohitkrai03: Oh I got it now. I think it's the prop that is being passed on. It should be `button` instead of `Button`. Again leftover from older implementation.

frontend/packages/console-app/src/redux/reducers/quick-start-reducer.ts
rohitkrai03: Actually I cannot define it's type like `Map<string, { activeQuickStartId: string }>`. `activeQuickStartId` is the key to the `Map` itself and then it's value is of type `string` as well. But the second key there would be `allQuickStartStatuses` and it's value would be again of type `Map<string, any>`. Even the main `UI` redux reducer state follows the same type. i couldn't come up with a smart way to define all the type safe params here.

rohitkrai03: Open to suggestion though.

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTileHeader.tsx
rohitkrai03: Yes, the duration needs to be shown always.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-utils.ts
rohitkrai03: Yes, it's almost a mock only but with a name that does not say mock. LOL. We have a separate PR to add final quick start tour data.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
andrewballantyne: Yup, understood. We can leave it or fix it, doesn't make a big difference.

frontend/packages/console-app/src/redux/reducers/quick-start-reducer.ts
andrewballantyne: Fair enough.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-types.ts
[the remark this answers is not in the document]
andrewballantyne: Types are not JS, enums are just objects in JS. So not _quite_ the same thing but I guess it doesn't _really_ matter.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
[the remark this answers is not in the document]
andrewballantyne: I would think it's worth a few moments... but if you're not super comfortable with flex we can ignore it. It's not going to hurt the maintenance cycle all that much.

frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTileHeader.tsx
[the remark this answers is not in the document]
andrewballantyne: Interesting.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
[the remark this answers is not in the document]
andrewballantyne: This is also not the only variable. I think it's bad form not to recompute the callback when the internal state changes from props provided to the component.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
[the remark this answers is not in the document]
andrewballantyne: Hmm - interesting. Fair enough.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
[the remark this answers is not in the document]
rohitkrai03: I didn't find any other variable in the scope of a callback. But anyways I updated the code to use `React.useCallback` now.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
```diff
@@ -0,0 +1,66 @@
...
+        }}
+        type="submit"
+        variant="secondary"
+        onClick={onBack}
+        isInline
+      >
+        Back
+      </Button>
+    )}
+    {status === QuickStartStatus.COMPLETE && (
```
andrewballantyne, line 58: I think you might want to check to see if we are on this route... otherwise you end up with a link that does nothing:

Screen Shot 2020-07-29 at 10 06 57 PM

andrewballantyne:
The background does not scale with content - some css fixes to be done on the quick start tile page (likely unrelated to this PR - I could log a bug if you prefer).

Screen Shot 2020-07-30 at 10 04 13 AM

andrewballantyne:
This seems odd... I skipped over a validation step and my `tour` is completed but not visually completed?

image

jerolimov:
When the quick start is opened and I press the button "Restart the tour" the tile onClick automatically closes the sidebar. I would prefer if the button only resets the tour and keeps it open. (The tour was reset successfully, but I need to click the tile or reset button again.)

pr-6141 webm

frontend/packages/console-app/src/redux/reducers/quick-start-reducer.ts
```diff
@@ -1,31 +1,142 @@
...
+          })
+          .updateIn(['allQuickStartStates', quickStartId], (quickStartState) => {
+            if (!quickStartId || quickStartState) {
+              return quickStartState;
+            }
+
+            return getDefaultQuickStartState(totalTasks);
+          }),
+      );
+      persistState(newState);
```
jerolimov, line 66: Reducers should be free of side effects, incl. storing data. A good option is the library [redux-persist](https://github.com/rt2zz/redux-persist). It's also possible to persist just single reducers, it is not required to persist everything.

I expect this takes some time, so I'm fine with this at the moment. WDYT? Should we create a ticket to change this later?

andrewballantyne: Yeah, this is not the first reducer that we have written that had side-effects. I believe the Topology filters work like this too. I'll find the ticket I logged for that, perhaps we just expand to cover this scenario too.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
rohitkrai03: Done.

rohitkrai03:
> The background does not scale with content - some css fixes to be done on the quick start tile page (likely unrelated to this PR - I could log a bug if you prefer).
> 
> Screen Shot 2020-07-30 at 10 04 13 AM

Fixed.

rohitkrai03:
> This seems odd... I skipped over a validation step and my `tour` is completed but not visually completed?
> 
> image

Thats how UX has it. User should be able to complete the tour even if he skips the validations.

frontend/packages/console-app/src/redux/reducers/quick-start-reducer.ts
rohitkrai03: I actually looked at `redux-persist` for this but didn't use it considering it would e a lot of overhead to firrst setup everything to work with it. As it is the PR has become huge. We can look at introducing `redux-persist` later on.

rohitkrai03:
> When the quick start is opened and I press the button "Restart the tour" the tile onClick automatically closes the sidebar. I would prefer if the button only resets the tour and keeps it open. (The tour was reset successfully, but I need to click the tile or reset button again.)
> 
> pr-6141 webm

Fixed.

rohitkrai03:
> There are still some wonky text due to the indentation... Recommend you do the following steps:
> 
> 1. Open up the content file(s) in your IDE -- I am looking at `quick-start-mocks` specifically
> 2. Select all content and remove indentation (Shift+Tab for me)
> 3. Run prettier on the file
> 
> This will remove all unintended indentation as part of template literals but return the content back to a JS object structure that Prettier likes. It seems to fix the content for me when I do it.

Fixed all the markdown issues. Updated the gif in description with new data.

andrewballantyne:
Some inconsistencies. If you cannot fix them early tomorrow, we should probably just convert them to bugs and address them after Feature Freeze @rohitkrai03.

CSS Issue that we can fix to make the click smoother. If the selection highlight doesn't take up space we can avoid the UI bump. This may be a PatternFly issue, not sure where the code lies.
QuickStart-CatalogShifting

Odd and inconsistent start tour functionality.
QuickStart-ClickBeforeStart
QuickStart-StartBeforeClick

andrewballantyne:
We really shouldn't have a scrollbar issue. We should be using flex or some other bounding box model not to have this kind of (I suspect) "100% height" issue. Again, this can be a bug.

Screen Shot 2020-07-30 at 9 43 52 PM
Screen Shot 2020-07-30 at 9 44 01 PM

rohitkrai03:
> Some inconsistencies. If you cannot fix them early tomorrow, we should probably just convert them to bugs and address them after Feature Freeze @rohitkrai03.
> 
> CSS Issue that we can fix to make the click smoother. If the selection highlight doesn't take up space we can avoid the UI bump. This may be a PatternFly issue, not sure where the code lies.
> QuickStart-CatalogShifting
> 
> Odd and inconsistent start tour functionality.
> QuickStart-ClickBeforeStart
> QuickStart-StartBeforeClick

Fixed the start quick start issue. The catalog tile active CSS is coming directly from PF catalog component so not sure anything can be done right now to make it smoother. Adding a border changes the size of the container which is why it seems like it moves from the position a bit.

rohitkrai03:
> We really shouldn't have a scrollbar issue. We should be using flex or some other bounding box model not to have this kind of (I suspect) "100% height" issue. Again, this can be a bug.
> 
> Screen Shot 2020-07-30 at 9 43 52 PM
> Screen Shot 2020-07-30 at 9 44 01 PM

Another style that is coming directly from PF Drawer component. It was either having a bit of scroll with a sticky footer or having a full scroll without sticky footer. I thought the experience without sticky footer was even worse.

andrewballantyne:
> Fixed the start quick start issue. The catalog tile active CSS is coming directly from PF catalog component so not sure anything can be done right now to make it smoother. Adding a border changes the size of the container which is why it seems like it moves from the position a bit.

We should log a Patternfly Bug for this. This is just a mishap in the css of the selection and it _should_ be fixed at the source.

andrewballantyne:
> Another style that is coming directly from PF Drawer component. It was either having a bit of scroll with a sticky footer or having a full scroll without sticky footer. I thought the experience without sticky footer was even worse.

This is poor quality and we should strive to fix it... if we don't own the container around the sidebar, then it's another PF issue. The sidebar loses it's left border when it scrolls and looks kinda wonky imo.

PR #6143, opened 2020-07-29
Changed local-storage namespace to openshift-local-storage

gnehapk:
- Changed local-storage namespace to openshift-local-storage
- Few minor fixes

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
@@ -18,7 +18,7 @@ const getRows: GetRows = ({ componentProps, customData }) => {
   const { data } = componentProps;
   const { nodes, setNodes, filteredNodes } = customData;
 
-  const nodeList = filteredNodes?.length ? filteredNodes : data.map(getName);
+  const nodeList = filteredNodes ?? data.map(getName);
   const filteredData = data.filter((node) => nodeList.includes(getName(node)));
 
   const rows = filteredData.map((node) => {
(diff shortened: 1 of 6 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
```diff
@@ -18,7 +18,7 @@ const getRows: GetRows = ({ componentProps, customData }) => {
   const { data } = componentProps;
   const { nodes, setNodes, filteredNodes } = customData;
 
-  const nodeList = filteredNodes?.length ? filteredNodes : data.map(getName);
+  const nodeList = filteredNodes ?? data.map(getName);
```
cloudbehl, line 21: FilteredNodes is a list? what are you trying to check?

cloudbehl: This check will always return true IMO.

gnehapk: Yes, its a list only and its required when sc does not have any nodes associated with it. In that case `filteredNodes.length`  will be 0 and all nodes will be displayed, which is not correct.

cloudbehl: I'm trying to say you will never hit `data.map(getName)` condition. Because if `filteredNodes` is empty or having value only filternodes will be passed so why do this?

gnehapk: I have hit this condition, hence modified it. So are you saying that, if `filteredNodes = [ ](empty)` then still `filteredNodes` will be picked? Well `filteredNodes.length` will be a false value and hence `data.map(getName)` will be picked.

gnehapk: ` data.map(getName)` will be picked when `filteredNodes` is undefined as its optional para.

PR #6146, opened 2020-07-29
Fix Context Providers extension not rendering multiple providers

glekner:
Reducing the Context Providers extensions array inside the JSX seems to mess up the order of the hooks which React expects. Creating one merged global provider seems to fix the issue.

can you take a look @rawagner @christianvogt @spadgett @rohitkrai03 @sahil143

### DIFF
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -174,14 +174,18 @@ class App_ extends React.PureComponent {
       </>
     );
 
+    const GlobalProvider = contextProviderExtensions.reduce(
+      (AccProvider, curProvider) => ({ children }) => (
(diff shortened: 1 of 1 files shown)


vojtechszocs:
LGTM but IMO the correct solution is to ensure each component rendered in a cycle has a stable `key` across multiple render passes.

spadgett:
/lgtm cancel

This is consistently failing on the same test, "displays created RoleBinding in list view"

spadgett:
/hold

We need to understand why the tests were failing instead of just rebasing. The Cypress tests that merged recently might not catch the problem, and this looks like it was a legitimate bug.

christianvogt:
The correct fix is to use a key. Do not recreate a react component inline on each render.

PR #6149, opened 2020-07-29
Add uid property to all extension instances

vojtechszocs:
This PR adds a `uid` property to all Console extension instances at runtime.

The `uid` property is computed when processing Console plugins via `PluginStore`.

Extension consumer APIs `useExtensions` & `withExtensions` now work with `LoadedExtension` type:

```ts
/**
 * Runtime extension interface, exposing additional metadata.
 */
export type LoadedExtension<E extends Extension> = E & {
  pluginName: string;
  uid: string;
};
```

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/store.ts
+++ b/frontend/packages/console-plugin-sdk/src/store.ts
@@ -1,5 +1,5 @@
 import * as _ from 'lodash';
-import { Extension, ExtensionWithMetadata, ActivePlugin } from './typings';
+import { Extension, LoadedExtension, ActivePlugin } from './typings';
 import { ExtensionRegistry } from './registry';
 
 export const sanitizeExtension = (e: Extension): Extension => {
@@ -9,9 +9,15 @@ export const sanitizeExtension = (e: Extension): Extension => {
   return e;
 };
 
-export const augmentExtension = (e: Extension, p: ActivePlugin): ExtensionWithMetadata => {
-  return Object.assign(e, { plugin: p.name });
-};
+export const augmentExtension = (
+  e: Extension,
+  pluginName: string,
+  index: number,
+): LoadedExtension<typeof e> =>
+  Object.assign(e, {
+    pluginName,
+    uid: `${pluginName}[${index}]`,
+  });
 
 export const isExtensionInUse = (e: Extension, flags: FlagsObject): boolean =>
   e.flags.required.every((f) => flags[f] === true) &&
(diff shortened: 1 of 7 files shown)


vojtechszocs:
cc @spadgett @christianvogt @glekner

frontend/packages/console-plugin-sdk/src/withExtensions.tsx
```diff
@@ -92,7 +92,7 @@ export const withExtensions = <
 };
 
 type ExtensionProps<E extends Extension> = {
-  [propName: string]: E[];
+  [propName: string]: E[] | LoadedExtension<E>[];
```
vojtechszocs, line 95: FYI - this type union is to allow `withExtension` users to keep injecting extension types instead of `LoadedExtension` types.

frontend/packages/console-plugin-sdk/src/store.ts
```diff
@@ -31,14 +35,14 @@ export const getGatingFlagNames = (extensions: Extension[]): string[] =>
...
-  private readonly extensions: ExtensionWithMetadata[];
+  private readonly extensions: Extension[];
 
   public readonly registry: ExtensionRegistry; // TODO(vojtech): legacy, remove
 
   public constructor(plugins: ActivePlugin[]) {
     this.extensions = _.flatMap(
       plugins.map((p) =>
-        p.extensions.map((e) => Object.freeze(augmentExtension(sanitizeExtension({ ...e }), p))),
+        p.extensions.map((e) => Object.freeze(augmentExtension(sanitizeExtension(e), p))),
```
andrewballantyne, line 45: This change now makes the extension mutated by these actions. Is there a benefit to mutating the extension?

vojtechszocs: `PluginStore` takes ownership of all extensions passed to it (right now via `ActivePlugin` interface).

There isn't an inherent benefit to mutating passed extension instances, this was done because `augmentExtension` computes index as `p.extensions.indexOf(e)` which doesn't work for new objects.

Anyway I think you're right so I'll update the code to keep creating new extension objects as before.

andrewballantyne: `.map` has the index available if you prefer to provide a 3rd argument to `augmentExtensions`?

vojtechszocs: Oh right, forgot about that. I've used `p.extensions.indexOf(e)` via closure but I can update it to use the `index` argument of `map` function.

PR #6132, opened 2020-07-28
Bug 1856507: Refine CDROM UI Flow

glekner:
JIRA  - https://issues.redhat.com/browse/CNV-5160

* Removed CDROM Section VM/VMT Details + Modal
* Removed Virtual Hardware Tab from VM Wizard - Cloud-init is now a single option in the Advanced step.
* Removed Tests
* Added Type column to Disks Table
* Added Type Select to Disk Modal + CD Validation

Screen Shot 2020-07-28 at 14 56 18

PR #6152, opened 2020-07-30
Bug 1861957: Overhaul kubevirt plugin tests

gouyang:
Overhaul failure tests, verified here: https://cnv-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/test-kubevirt-console-4.5-cnv-2.4-prod-gouyang/5/testReport/.

Note:
The verify job is running against http://0.0.0.0:9000/ which has login disabled, so non-admin test is failed.
If run non-admin test against https://console-openshift-console.apps.gouyang0727.cnv-qe.rhcloud.com/, it's PASS.


gouyang:
will rebase on #6132.

frontend/packages/console-shared/src/test-utils/utils.ts
```diff
@@ -46,6 +48,8 @@ export function removeLeakableResource(leakedResources: Set<string>, resource) {
 }
 
 export function createResource(resource) {
+  // ensure not return error even the source is existed by deleting it before adding it
```
suomiy, line 51: where is this necessary?

IMO we should have a specific method for this like `recreate` so the use case would be apparent by looking at the name.

We should not use this in places where it isn't necessary, so it would not hide bugs in the future

frontend/packages/console-shared/src/test-utils/utils.ts
```diff
@@ -56,7 +60,7 @@ export function createResources(resources) {
 export function deleteResource(resource) {
   const kind = resource.kind === 'NetworkAttachmentDefinition' ? 'net-attach-def' : resource.kind;
   execSync(
-    `kubectl delete -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
+    `kubectl delete --ignore-not-found=true -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
```
suomiy, line 63: the same. Where is this used?

we could also pass an additional argument for this if needed

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/vmBuilderPresets.ts
```diff
@@ -87,18 +87,18 @@ export const computedVMBuilders = deepFreeze(
   ),
 );
 
-export const getBasicVMBuilder = () =>
+export const getBasicVMBuilder = (namespace?: string) =>
   new VMBuilder()
-    .setNamespace(testName)
+    .setNamespace(namespace || testName)
```
suomiy, line 92: could we call setNamespace 2nd time on the returned builder instead?

It could get quite confusing if start introducting additional params here

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/vmBuilderPresets.ts
```diff
@@ -87,18 +87,18 @@ export const computedVMBuilders = deepFreeze(
...
     .setDescription('Default vm description')
     .setFlavor(flavorConfigs.Tiny)
     .setOS(OperatingSystem.RHEL7)
     .setWorkload(Workload.DESKTOP)
     .setStartOnCreation(false);
 
-export const getBasicVMTBuilder = () =>
+export const getBasicVMTBuilder = (namespace?: string) =>
```
suomiy, line 99: plus here

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.action.base.scenario.ts
```diff
@@ -118,7 +119,9 @@ describe('Test VM actions', () => {
       'ID(CNV-1794) Unpauses VM',
       async () => {
         pauseVM(vmName, testName);
+        await vm.waitForStatus(VM_STATUS.Paused);
```
suomiy, line 122: nice

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
```diff
@@ -42,7 +42,7 @@ describe('KubeVirt VM console - RDP', () => {
     const vmName = 'windows-rdp';
 
     // for cmd-line scripts only
-    execSync(`kubectl config set-context --current --namespace=${testName}`, { stdio: 'inherit' });
```
suomiy, line 45: why was this here?

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.dashboard.scenario.ts
```diff
@@ -12,37 +12,20 @@ import { VirtualMachine } from './models/virtualMachine';
...
-    try {
-      await browser.wait(
-        until.not(until.textToBePresentInElement(dashboardView.vmStatus, VM_STATUS.Off)),
-        PAGE_LOAD_TIMEOUT_SECS,
-      );
-    } catch (ex) {
-      // continue, this is optional condition
-      // we want to wait for import to start but in some cases it may have already completed
-    }
-    await browser.wait(
```
suomiy, line 41: is this not needed anymore?

gouyang: I don't see `importing` status on the latest environment, it's `pending` -> `starting`. Seems there is no `importing` now.

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
gouyang: I don't know why it output to stdout, just want to keep the tests output clean if it's not necessay?

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/vmBuilderPresets.ts
gouyang: yes, it's better to call `setNamespace` instead of changing here.

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: I've seen some tests failed because of existed resource when I run tests locally, I was think it would be safe to make sure they're not returning error as they're used in setup/teardown.

I don't want to introduce `recreate` as they're not on the same purpose. I can revert this.

suomiy: they are not, but we could have both createResource and reCreateResource

ok, let's revert if you are not seeing this anymore. If you think it is necessary in some case, I would vote for having already mentioned functions and using it just in place where it is needed.

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
```diff
@@ -138,17 +137,13 @@ describe('KubeVirt VM console - RDP', () => {
        */
 
       await vm.navigateToDetail();
-      // eslint-disable-next-line no-console
-      console.log(
-        'Waiting for static IP to be reported by the guest-agent (can take up to several minutes ...)',
-      );
       // Waiting for instllation & start of the guest-agent and reporting the static IP back
```
suomiy, line 140: can we add this statement `(can take up to several minutes ...)` as a comment here?

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.dashboard.scenario.ts
suomiy: +1 after inspecting ` getVMManifest('Container',` the VM should not contain any datavolume to wait for

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
suomiy: maybe it was to make clear that we are changing namespace to the active user?

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
gouyang: NP

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: What about add a check via `kubectl get`, if the resource is existing, do nothing. Then it don't throw error.

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: If the resource is not created successfully, the test is failed earlier and will have an error in tear down step in `deleteResource`. This is trying to make it not throwing error, as it seems always been used in tear down step.

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
gouyang: I would like to make it quite because a single line output does not help.

frontend/packages/console-shared/src/test-utils/utils.ts
suomiy: ok, can we at least add the additional optional argument ignoreNotFound ?

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
suomiy: it could help somebody, but ok, I will leave it to your judgment

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: Added additional optional argument `ignoreNotFound` with default value `true` because it's used in tear down, in case it's called somewhere else, it can specify it to `false`.

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: Are you meaning to use `kubectl create` for createResource and use `kubectl apply` for applyResource?
I changed it to `kubectl apply` in createResource.

frontend/packages/kubevirt-plugin/integration-tests/tests/vm.tab.console.rdp.scenario.ts
gouyang: thanks.

frontend/packages/console-shared/src/test-utils/utils.ts
suomiy: yes, IMO it would be nicer to have both createResource and applyResource and use them accordingly

frontend/packages/console-shared/src/test-utils/utils.ts
suomiy: +1

frontend/packages/console-shared/src/test-utils/utils.ts
```diff
@@ -46,18 +48,24 @@ export function removeLeakableResource(leakedResources: Set<string>, resource) {
...
 
-export function deleteResource(resource) {
+export function deleteResource(resource, ignoreNotFound = true) {
   const kind = resource.kind === 'NetworkAttachmentDefinition' ? 'net-attach-def' : resource.kind;
-  execSync(
-    `kubectl delete -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
-  );
+  if (ignoreNotFound) {
+    execSync(
+      `kubectl delete --ignore-not-found=true -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
```
suomiy, line 62:
```suggestion
      `kubectl delete --ignore-not-found=${ignoreNotFound} -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
```

gouyang: It seems better to don't add `if esle` in this way, changed to use `--ignore-not-found=${ignoreNotFound}` and not use `if else`.

frontend/packages/console-shared/src/test-utils/utils.ts
gouyang: I added applyResource, I will use it accordingly in future based on test experience.

suomiy:
@rawagner could you take a look?

PR #6155, opened 2020-07-30
Show alert message and hide pods when ImageStream is not ready yet

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3875

**Analysis / Root cause**: 
When creating a Deployment from Git or Dockerfile (s2i) the web console creates a Deployment, ImageStream and BuildConfig. The Deployment creates Pods while the first build is running. This problem is similar to https://issues.redhat.com/browse/ODC-3175 / https://github.com/openshift/console/pull/6138 where we wait until the first ImageStream informations are fetched from the remote image registry. But this is not possible for builds which takes much longer.

**Solution Description**: 
Adding an alert message, close to the pod list in the topology sidebar for Deployments. It detects if the Deployment was created with the Web Console and the first generation of pods are "Pending" because their have no imageID.

It was also possible to check if the Deployment has the `generated` version === 1. But when checking this the alert message already disappears as soon as the build finishes. In this case the failing Pod was shown for a second.

@openshift/team-devconsole-ux @beaumorley  I added the option to show / hide the failing pods. We could easily remove this option again. The dialog now automatically disappear when the build finish. What do you think?

The message shows now:
> **Waiting for the build**
>
> Waiting for the first build to run successfully. You may temporarily see "ImagePullBackOff" and "ErrImagePull" errors while waiting.
>
> [Hide/Show waiting pods with errors](#)

Screenshots when the pods are hidden:
3875-closed-alert

Screenshots when the pods are shown:
3875-open-alert

Complete process as gif:
odc-3875 webm

Unchanged

**Test setup:**
* Create a Deployment from a Git Repository (for example `https://github.com/jerolimov/docker` with context dir `/nodeinfo`)

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/overview/pods-overview.tsx
+++ b/frontend/public/components/overview/pods-overview.tsx
@@ -2,6 +2,7 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 
+import { Alert, AlertActionLink } from '@patternfly/react-core';
 import { Status } from '@console/shared';
 import { ResourceLink, resourcePath, SidebarSectionHeading } from '../utils';
 import { podPhase, PodKind, K8sResourceKind, referenceFor } from '../../module/k8s';
@@ -52,6 +53,14 @@ const errorPhases = [
 ];
 
 const isPodError = (pod: PodKind) => _.includes(errorPhases, podPhase(pod));
+
+const isDeploymentGeneratedByWebConsole = (obj: K8sResourceKind) =>
+  obj.kind === 'Deployment' &&
+  obj.metadata?.annotations?.['openshift.io/generated-by'] === 'OpenShiftWebConsole';
+
+const isPodWithoutImageId = (pod: PodKind) =>
+  pod.status?.phase === 'Pending' &&
+  pod.status?.containerStatuses?.some((containerStatus) => !containerStatus.imageID);
 
 const podCompare = (pod1: PodKind, pod2: PodKind): number => {
   const error1 = isPodError(pod1);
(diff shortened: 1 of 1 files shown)


beaumorley:
This is great. Can you make a slight nit text edits?

1. Switch the word see and temporarily 
2. Make the word error plural by adding an "s" 

so like this:

"... You may temporarily see "imagePullBackOff" and "ErrImagePull" errors ..."

jerolimov:
> Can you make 2 slight nit text edits?

Yep, thanks & done :)

frontend/public/components/overview/pods-overview.tsx
```diff
@@ -123,25 +132,49 @@ export const PodsOverview: React.SFC<PodsOverviewProps> = ({
...
-  pods.sort(podCompare);
 
   return (
     <>
       <SidebarSectionHeading text="Pods">
-        {_.size(pods) > podsShown && (
+        {_.size(filteredPods) > podsShown && (
           <Link className="sidebar__section-view-all" to={linkTo}>
-            {`View all (${_.size(pods)})`}
+            {`View all (${_.size(filteredPods)})`}
```
rottencandy, line 155: Small nit: I think this should still show the size for `pods`. Because when you go to the details page, you would see the list of all pods, including the ones that aren't filtered. Not entirely sure though, WDYT?

jerolimov: I think you're right here, will change it back. Thanks Saud :+1:

jerolimov: Done

PR #6157, opened 2020-07-30
Bump monaco-editor packages, fix critical dependency error

glekner:
Before
image

After
image

Updating `"react-monaco-editor"` to above `0.37.x` causes typescript errors. the latest is `0.39.0`


jeff-phillips-18:
This fixes a build error I am seeing after upgrading PF react versions in https://github.com/openshift/console/pull/6130

spadgett:
@glekner I think this is causing the text to be preselected when you open the YAML editor.

Screenshot_2020-07-31 OKD

PR #6163, opened 2020-07-30
Bug 1860535: show project description

### LINKED ISSUE #802: Project display name and description unused
Unique names for projects are used in all places and display names only appear in the details, which seems to be the opposite of their intended purpose. Additionally, project descriptions don't appear anywhere that I've been able to find. I wonder if we should emphasize these fields more, or remove them if they are not used.

debsmita1:
**fixes:**
https://issues.redhat.com/browse/ODC-1854
Fixes #802

**Root Analysis:**
The project description that is being asked for during project creation is not being displayed anywhere in the console

**Solution description:**
- displaying the project description in the details card in the Project Overview tab
- displaying the project description in the Project Details tab
- added Popover if the description exceeds 10 lines
- showing `No description` in the Details tab when there isn't one

**Screenshots & GIF:**
details
pd
project-des
detailscard
nodd
detailscard

### DIFF
--- a/frontend/public/components/_resource.scss
+++ b/frontend/public/components/_resource.scss
@@ -158,3 +158,9 @@
     }
   }
 }
+
+.co-namespace-summary__description {
+  max-height: 15rem;
+  overflow-y: auto;
+  padding-right: var(--pf-global--spacer--sm);
+}
--- a/frontend/public/components/dashboard/project-dashboard/details-card.tsx
+++ b/frontend/public/components/dashboard/project-dashboard/details-card.tsx
@@ -1,5 +1,6 @@
 import * as _ from 'lodash';
 import * as React from 'react';
+import cx from 'classnames';
 import DashboardCard from '@console/shared/src/components/dashboard/dashboard-card/DashboardCard';
 import DashboardCardBody from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardBody';
 import DashboardCardHeader from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardHeader';
@@ -17,6 +18,7 @@ export const DetailsCard: React.FC = () => {
   const keys = _.keys(obj.metadata.labels).sort();
   const labelsSubset = _.take(keys, 3);
   const firstThreelabels = _.pick(obj.metadata.labels, labelsSubset);
+  const description = obj.metadata.annotations?.['openshift.io/description'];
   const detailsLink = `${resourcePathFromModel(ProjectModel, obj.metadata.name)}/details`;
   const serviceMeshEnabled = obj.metadata?.labels?.['maistra.io/member-of'];
   return (
@@ -38,6 +40,16 @@ export const DetailsCard: React.FC = () => {
               <LabelList kind={ProjectModel.kind} labels={firstThreelabels} />
               {keys.length > 3 && <DashboardCardLink to={detailsLink}>View all</DashboardCardLink>}
             </div>
+          </DetailItem>
+          <DetailItem isLoading={!obj} title="Description">
+            <span
+              className={cx({
+                'text-muted': !description,
+                'co-project-dashboard-details-card__description': description,
+              })}
+            >
+              {description || 'No description'}
+            </span>
           </DetailItem>
           {serviceMeshEnabled && (
             <DetailItem isLoading={!obj} title="Service Mesh">
--- a/frontend/public/components/namespace.jsx
+++ b/frontend/public/components/namespace.jsx
@@ -822,6 +822,7 @@ const ResourceUsage = requirePrometheus(({ ns }) => (
 
 export const NamespaceSummary = ({ ns }) => {
   const displayName = getDisplayName(ns);
+  const description = getDescription(ns);
   const requester = getRequester(ns);
   const serviceMeshEnabled = ns.metadata?.labels?.['maistra.io/member-of'];
   const canListSecrets = useAccessReview({
@@ -830,13 +831,32 @@ export const NamespaceSummary = ({ ns }) => {
     verb: 'patch',
     namespace: ns.metadata.name,
   });
+
   return (
     <div className="row">
       <div className="col-sm-6 col-xs-12">
         {/* Labels aren't editable on kind Project, only Namespace. */}
         <ResourceSummary resource={ns} showLabelEditor={ns.kind === 'Namespace'}>
-          {displayName && <dt>Display Name</dt>}
-          {displayName && <dd>{displayName}</dd>}
+          <dt>Display Name</dt>
+          <dd
+            className={classNames({
+              'text-muted': !displayName,
+            })}
+          >
+            {displayName || 'No display name'}
+          </dd>
+          <dt>Description</dt>
+          <dd>
+            <p
+              className={classNames({
+                'text-muted': !description,
+                'co-pre-wrap': description,
+                'co-namespace-summary__description': description,
+              })}
+            >
+              {description || 'No description'}
+            </p>
+          </dd>
           {requester && <dt>Requester</dt>}
           {requester && <dd>{requester}</dd>}
         </ResourceSummary>
--- a/frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
+++ b/frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
@@ -89,14 +89,16 @@ describe('Project Dashboard', () => {
       const items = projectDashboardView.detailsCardList.$$('dt');
       const values = projectDashboardView.detailsCardList.$$('dd');
 
-      expect(items.count()).toBe(3);
-      expect(values.count()).toBe(3);
+      expect(items.count()).toBe(4);
+      expect(values.count()).toBe(4);
       expect(items.get(0).getText()).toEqual('Name');
       expect(values.get(0).getText()).toEqual(testName);
       expect(items.get(1).getText()).toEqual('Requester');
       expect(values.get(1).getText()).toEqual('kube:admin');
       expect(items.get(2).getText()).toEqual('Labels');
       expect(values.get(2).getText()).toEqual('No labels');
+      expect(items.get(3).getText()).toEqual('Description');
+      expect(values.get(3).getText()).toEqual('No description');
     });
     it('has View all link', async () => {
       const link = projectDashboardView.detailsCard.$(
--- a/frontend/public/components/dashboard/project-dashboard/_details-card.scss
+++ b/frontend/public/components/dashboard/project-dashboard/_details-card.scss
@@ -1,0 +1,3 @@
+.co-project-dashboard-details-card__description {
+  @include co-line-clamp(1);
+}
--- a/frontend/public/style.scss
+++ b/frontend/public/style.scss
@@ -107,5 +107,6 @@
 @import 'components/dashboard/dashboards-page/cluster-dashboard/status-card';
 
 @import 'components/dashboard/project-dashboard/activity-card';
+@import 'components/dashboard/project-dashboard/details-card';
 
 @import 'components/nav/nav-header';


lwrigh:
This is looking good to me! Thank you for starting the PR @debsmita1!

frontend/public/components/dashboard/project-dashboard/details-card.scss
```diff
@@ -0,0 +1,3 @@
+.details-card__description {
+  -webkit-line-clamp: 1 !important;
```
spadgett, line 2:
```suggestion
    @include co-line-clamp(1);
```

frontend/public/components/dashboard/project-dashboard/details-card.tsx
```diff
@@ -11,12 +11,14 @@ import { getName, getRequester, GreenCheckCircleIcon } from '@console/shared';
...
 import { ProjectModel } from '../../../models';
 import { ProjectDashboardContext } from './project-dashboard-context';
+import './details-card.scss';
 
 export const DetailsCard: React.FC = () => {
   const { obj } = React.useContext(ProjectDashboardContext);
   const keys = _.keys(obj.metadata.labels).sort();
   const labelsSubset = _.take(keys, 3);
   const firstThreelabels = _.pick(obj.metadata.labels, labelsSubset);
+  const description = obj?.metadata?.annotations?.['openshift.io/description'];
```
spadgett, line 21: This would blow up above if `obj` is null.

```suggestion
  const description = obj.metadata.annotations?.['openshift.io/description'];
```

frontend/public/components/dashboard/project-dashboard/details-card.tsx
```diff
@@ -44,6 +46,11 @@ export const DetailsCard: React.FC = () => {
               <GreenCheckCircleIcon /> Service Mesh Enabled
             </DetailItem>
           )}
+          {description && (
+            <DetailItem isLoading={!obj} title="Description">
+              <span className="co-line-clamp details-card__description">{description}</span>
```
spadgett, line 51: With the CSS change above this can be

```suggestion
              <span className="details-card__description">{description}</span>
```

frontend/public/components/dashboard/project-dashboard/details-card.scss
```diff
@@ -0,0 +1,3 @@
+.details-card__description {
```
spadgett, line 1: If this is specific to projects, that should go in the class name.

spadgett:
Description column is being added to the table in #6169

frontend/public/components/namespace.jsx
```diff
@@ -531,8 +533,22 @@ export const NamespaceSummary = ({ ns }) => {
...
+          {<dt>Display Name</dt>}
+          <dd
+            className={classNames({
+              'text-muted': !displayName,
+            })}
+          >
+            {displayName || 'No display name'}
+          </dd>
+          {<dt>Description</dt>}
+          <dd
```
spadgett, line 545: We probably want line clamp here as well, even if it's set to a larger value like 10 lines. There's nothing in the API preventing a user from adding up to a 1MB description, which would blow up the layout.

spadgett: I'd add `co-pre-line` as well to preserve line breaks.

debsmita1: > We probably want line clamp here as well, even if it's set to a larger value like 10 lines. There's nothing in the API preventing a user from adding up to a 1MB description, which would blow up the layout.

should I set the max lines to 10? cc @lwrigh

debsmita1:
> Description column is being added to the table in #6169

Should I add `description name` next to the `display name`?

spadgett:
> > Description column is being added to the table in #6169
> 
> Should I add `description` next to the `display name`?

No, we're adding it to the table as an extra customizable column.

spadgett:
/lgtm cancel

This is consistently failing on the "has all fields populated" test.

andrewballantyne:
/hold

andrewballantyne:
/hold cancel
/retest

🤞

frontend/public/components/namespace.jsx
```diff
@@ -834,8 +836,26 @@ export const NamespaceSummary = ({ ns }) => {
       <div className="col-sm-6 col-xs-12">
         {/* Labels aren't editable on kind Project, only Namespace. */}
         <ResourceSummary resource={ns} showLabelEditor={ns.kind === 'Namespace'}>
-          {displayName && <dt>Display Name</dt>}
-          {displayName && <dd>{displayName}</dd>}
+          {<dt>Display Name</dt>}
```
andrewballantyne, line 839:
```suggestion
          <dt>Display Name</dt>
```

frontend/public/components/namespace.jsx
lwrigh: I would suggest truncating at the 10 line limit like you're suggesting for the description section in the UI. Would it be possible to set a character limit for both the UI and the YAML? That would help to constraint the length of the description text in both views.  

If it isn't an option, we could go the route of truncating the description text and then having it become clickable when the user hovers over it and they can expand it. We could use either a popover or add an expanding/collapsing caret to the description section.

andrewballantyne: @lwrigh There is a collapsible feature in the margin for the YAML editor. I don't think any change is needed there.

Not sure any character limit can be imposed here, but it is an openshift custom resource -- it is possible to think on that but not for 4.6 (as this would be a feature). 

Would you be able to provide a quick mock what you're thinking of for the expand feature? I don't want us going crazy to support this -- again it's post feature freeze so it's limited what we can do here.

lwrigh: Expandable description 1
Expandable description 2

I was thinking of something like this if we went the caret route.

debsmita1: @spadgett ^

christianvogt: @lwrigh for the serverless reason text in the table, the suggestion is currently a popover on hover IIRC? Wouldn't we want to have a similar experience in both cases.
Also the description shouldn't be hidden if it's of reasonable size.

lwrigh: Yes that's the current suggestion for the reason, We can go the popover route here too. The description here could truncate after 10 lines or so since there's more space around it.

lwrigh:
@debsmita1 Can you post the most recent screenshots of the description box if anything else has changed from the previous ones?

debsmita1:
> @debsmita1 Can you post the most recent screenshots of the description box if anything else has changed from the previous ones?

@lwrigh It's added in the PR description. Below is the screenshot of the most recent change
desc

lwrigh:
Great thank you that looks good to me!

frontend/public/components/namespace.jsx
```diff
@@ -822,21 +822,83 @@ const ResourceUsage = requirePrometheus(({ ns }) => (
...
+                <Popover aria-label="Description" bodyContent={description} maxWidth="25rem">
+                  <Button variant="link" isInline>
+                    {projectDescription}
+                  </Button>
+                </Popover>
+              ) : (
+                projectDescription
+              )
+            ) : (
+              <p className="text-muted">{'No description'}</p>
```
jerolimov, line 899:
```suggestion
              <p className="text-muted">No description</p>
```

jerolimov:
From coding perspective this lgtm. But I have some UX issues with the Popover solution:

1. When adding a multi line description, line breaks was shown in the details page, but not in the Popover:



2. With a really long description the Popover extends the viewport size which shows then a second scrollbar (to scroll the window). This enables me to read all informations but looks buggy:





@debsmita1 @lwrigh @andrewballantyne @openshift/team-devconsole-ux What do you think about an alternative idea:

Adding a small pencil icon similar and add an edit modal with an textarea. This allow the users to read the complete description and makes it also easy to change this. But maybe this would keep this back for 4.7? WDYT?

andrewballantyne:
@jerolimov All good points... Maybe we should fix this one issue as it's changing the format:

> 1. When adding a multi line description, line breaks was shown in the details page, but not in the Popover:
> 
> 

But I don't think we need to do anything else for 4.6 on this. If the popup doesn't give a beautiful experience for extremely long text, they still have the YAML they can go to. Might be worth looking at a UX improvement here if @lwrigh thinks there is a desire here.

lwrigh:
I'd still push for showing like 10 lines of the description first and then truncate the lines that goes past that limit and have the user expand a 'View more' link or the option where an expand caret says 'View full description' for a better experience. The super long popover is a lot. I think this is a good example of a component that we should take a broader look at to see if there are any other cases in the console that feature a similar description experience and design an experience that's consistent across the console. For 4.6 I think this is fine for the time being.

andrewballantyne:
> I'd still push for showing like 10 lines of the description first and then truncate the lines that goes past that limit and have the user expand a 'View more' link or the option where an expand caret says 'View full description' for a better experience. The super long popover is a lot. I think this is a good example of a component that we should take a broader look at to see if there are any other cases in the console that feature a similar description experience and design an experience that's consistent across the console. For 4.6 I think this is fine for the time being.

@lwrigh That's fair... we are basing this off of another usage of "expanded text" for tables... so definitely encourage UX to created a DTUX and chase down a more preferred solution for this kinda problem. I'm sure we'll have more of them. In an app with user-driven content you sorta have this as a common problem.

lwrigh:
Most defs, yeah I would love to come up with a more standardized approach for it. I think this is a good example of a component that we can do an audit of and provide a more consistent design solution for. I can create a DTUX story for taking a closer and broader look at it.

spadgett:
Have considered just adding a max-height and scrollbar for very long descriptions?

debsmita1:
> Have we considered just adding a max-height and scrollbar for very long descriptions?
@andrewballantyne @jerolimov @lwrigh ^

jerolimov:
> Have we considered just adding a max-height and scrollbar for very long descriptions?

For me, removing the Popover and adding a scrollable, max-height inline area sounds good :+1: 

We have a similar approach also in the OperatorHub for the long READMEs.

lwrigh:
That's totally cool with me since it's a simpler solution. That would be easier to implement into other similar areas of the UI as well.

debsmita1:
@lwrigh @jerolimov @andrewballantyne I have updated the screens in the PR description. PTAL

lwrigh (approved):
The scrollbar update looks good to me!

andrewballantyne:
> That's totally cool with me since it's a simpler solution. That would be easier to implement into other similar areas of the UI as well.

Definitely a more elegant and to the point fix... @lwrigh this should be documented by UX so we draw a consistency around this.

frontend/public/components/_resource.scss
```diff
@@ -158,3 +158,9 @@
     }
   }
 }
+
+.co-namespace-summary__description {
+  max-height: 15rem;
+  overflow-y: auto;
+  padding-right: var(--pf-global--spacer--sm);
```
jerolimov, line 165: @debsmita1 Sorry I have still the opinion that we should keep line spaces here. So we could you a markdown component or add just an additional css rule here:

```css
  white-space: pre-wrap;
```

Then also an description like this works:

```yaml
  annotations:
    openshift.io/description: |
      1 first line
      2 second line
            3 line with intent

      4 line break above
      5fasddfasdfafasd
      65faasdfasfdsafasdf
      .....
```

jerolimov (changes requested):

frontend/public/components/_resource.scss
jerolimov: I searched for `white-space: pre` in console, there are also a class for that (`co-pre-wrap`). Its up to you, I see other descriptions which uses this also.

debsmita1: used the class `co-pre-wrap`

jerolimov (approved):
:+1: 
/lgtm

rohitkrai03 (approved):
/approve

PR #6165, opened 2020-07-30
Add ability to filter topology by resource types

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4319

**Description**
As a user, I want to show/hide specific resource types in the topology view

When a resource is selected, that resource and its children are shown.

**Screenshots**

image

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

/kind feature

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/TopologyView.tsx
+++ b/frontend/packages/dev-console/src/components/topology/TopologyView.tsx
@@ -1,31 +1,32 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import {
-  Model,
-  Visualization,
-  SELECTION_EVENT,
-  SelectionEventListener,
-  GROUPS_LAYER,
-  TOP_LAYER,
+  BaseEdge,
   BOTTOM_LAYER,
   DEFAULT_LAYER,
-  isNode,
+  GROUPS_LAYER,
   isEdge,
-  BaseEdge,
+  isNode,
+  Model,
+  SELECTION_EVENT,
+  SelectionEventListener,
+  TOP_LAYER,
+  Visualization,
 } from '@patternfly/react-topology';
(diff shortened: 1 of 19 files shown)


andrewballantyne:
Odd... Knative selected, but I don't see Knative - I see the Revision. This would be confusing, no?

Screen Shot 2020-07-30 at 4 08 32 PM
Screen Shot 2020-07-30 at 4 08 37 PM

andrewballantyne:
There appears to be a rendering issue with groups too... It seems I can toggle off my group of Deployments, then re-enable them... and the grouping doesn't come back immediately. I wanna say it comes back after "some time" (feels like a webhook kick) or some sorta mouse-over triggering it. I don't think it's a "svg rendering issue" because the alignment is off:

Screen Shot 2020-07-30 at 4 26 28 PM
Screen Shot 2020-07-30 at 4 26 37 PM

jeff-phillips-18:
> Odd... Knative selected, but I don't see Knative - I see the Revision. This would be confusing, no?

Fixed.

invincibleJai:
Thanks @jeff-phillips-18 , works as expected and i could do the filter as mentioned (some cases i see slight delay in filter but they depends on nodes we have). Another observation is In design there is equal gap between `Display Options`, `Filter by Resource`

image

jeff-phillips-18:
> Another observation is In design there is equal gap between `Display Options`, `Filter by Resource`

Fixed.

andrewballantyne:
/retitle Add ability to filter topology by resource types
/hold

Putting on hold so it doesn't look like this is incomplete, just waiting on dependant PR to merge.

andrewballantyne:
Only pending issue I can see is this one: https://issues.redhat.com/browse/ODC-4417

PR #6170, opened 2020-07-31
Monitoring: Use <dl> for popover help text

kyoto:
These help texts are description lists, so `<dl>` is better for accessibility.

Do you prefer this approach @spadgett?

Before | After
-|-
before | after

### DIFF
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -280,57 +280,66 @@ const PopoverField: React.FC<{ body: React.ReactNode; label: string }> = ({ body
 );
 
 const alertStateHelp = (
-  <>
-    <p>
-      <AlertStateIcon state={AlertStates.Firing} /> <strong>Firing:</strong> The alert condition is
-      true for the duration of the timeout.
-    </p>
-    <p>
-      <AlertStateIcon state={AlertStates.Pending} /> <strong>Pending:</strong> The alert condition
-      is true, but the timeout has not been reached.
-    </p>
-    <p>
-      <AlertStateIcon state={AlertStates.Silenced} /> <strong>Silenced:</strong> The alert is now
-      silenced.
-    </p>
-  </>
+  <dl className="co-inline">
+    <dt>
+      <AlertStateIcon state={AlertStates.Firing} /> <strong>Firing: </strong>
+    </dt>
+    <dd>The alert condition is true for the duration of the timeout.</dd>
+    <dt>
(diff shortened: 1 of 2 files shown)


frontend/public/components/monitoring/_monitoring.scss
```diff
@@ -1,5 +1,15 @@
 $tooltip-background-color: #151515;
 
+dl.inline {
```
spadgett, line 3: I'd suggest a `co-` prefix so we never conflict with pf or another dependency.

We should either move this to a file like `_common.scss` or give it a BEM name for the monitoring view.

kyoto: Thanks @spadgett. Now renamed to `co-inline` and moved to `_common.scss`.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -280,57 +280,66 @@ const PopoverField: React.FC<{ body: React.ReactNode; label: string }> = ({ body
...
-    <p>
-      <SeverityIcon severity={AlertSeverity.Info} /> <strong>Info:</strong> The alert is provided
-      for informational purposes only.
-    </p>
-    <p>
-      <SeverityIcon severity={AlertSeverity.None} /> <strong>None:</strong> The alert has no defined
-      severity.
-    </p>
-  </>
+  <dl className="inline">
```
spadgett, line 300: `co-inline`?

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -280,57 +280,66 @@ const PopoverField: React.FC<{ body: React.ReactNode; label: string }> = ({ body
...
-      <strong>Platform:</strong> Platform-level alerts relate only to OpenShift namespaces.
-      OpenShift namespaces provide core OpenShift functionality.
-    </p>
-    <p>
-      <strong>User:</strong> User workload alerts relate to user-defined namespaces and are
-      customizable. User workload monitoring can be enabled post-installation to provide
-      observability into your own services.
-    </p>
-  </>
+  <dl className="inline">
```
spadgett, line 327: `co-inline`?

frontend/public/components/monitoring/alerting.tsx
kyoto: Oops 🤦

frontend/public/components/monitoring/alerting.tsx
kyoto: 🤦

PR #6176, opened 2020-07-31
Bug 1871092: Catch unhandeled promise when metrics are not available (shows an error overlay in the topology in dev mode)

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4416

**Analysis / Root cause**: 
There is a full screen error message each time I open the Topology on a local CRC (in dev mode) because there is no metrics available on a CRC.

**Solution Description**: 
Handle the promise also in the error case. => Do nothing here.

Before:
unhandled-promise

After:
No error message was shoiwn.

Not covered

**Test setup:**
* Requires a CRC setup
* Start local developer console and open the topology

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/overview/index.tsx
+++ b/frontend/public/components/overview/index.tsx
@@ -305,9 +305,10 @@ class OverviewMainContent_ extends React.Component<
           throw new Error(`Could not fetch metrics, status: ${status}`);
         }
       })
-      .then(() => {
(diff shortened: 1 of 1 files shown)


rohitkrai03 (approved):
/approve

PR #6180, opened 2020-07-31
[WIP]: Watch ConsoleLinks so they update in the UI without refreshing

rhamilto:


rhamilto:
Need to revist next week when I'm not working from a moving truck.  :)

frontend/packages/dev-console/src/components/guided-tour/GuidedTourText.tsx
```diff
@@ -53,11 +51,12 @@ export const searchTourText = (
...
-  const consoleLinks = useSelector<K8sResourceCommon[]>((state: RootState) =>
-    state.UI.get('consoleLinks'),
-  );
+  const [consoleLinks] = useK8sWatchResource<K8sResourceKind[]>({
+    isList: true,
+    kind: referenceForModel(ConsoleLinkModel),
+  });
   const openshiftBlogLink = consoleLinks.filter(
-    (link: K8sResourceCommon) => link.metadata.name === 'openshift-blog',
+    (link: K8sResourceKind) => link.metadata.name === 'openshift-blog',
```
spadgett, line 59: I think we probably want to just hard-code links like the OpenShift blog in console instead of using a `ConsoleLink` manifest in the operator repo. It's unnecessarily complicating the frontend logic, and we won't be able to translate the label.

cc @jhadvig

frontend/packages/dev-console/src/components/topology/TopologyEdgePanel.tsx
```diff
@@ -36,7 +32,11 @@ const connectorTypeToTitle = (type: string): string => {
   }
 };
 
-const TopologyEdgePanel: React.FC<TopologyEdgePanelProps> = ({ edge, consoleLinks }) => {
+const TopologyEdgePanel: React.FC<TopologyEdgePanelProps> = ({ edge }) => {
+  const [consoleLinks] = useK8sWatchResource<K8sResourceKind[]>({
```
spadgett, line 36: I'm a little hesitant to open a watch per topology node, even though there should only be one watch. It would be better if we can watch this once in the topology view.

This seems to be used for the Kiali link, but the logic looks incorrect here:

https://github.com/spadgett/console/blob/e47665c50e6f5ea842a2de42e88e60a3c2090452/frontend/packages/dev-console/src/components/topology/topology-utils.ts#L59-L61

It's assuming any namespace link is Kiali, but that's not necessarily the case. Maybe we can add a label to the Kiali `ConsoleLink` so we know.

cc @christianvogt

frontend/packages/dev-console/src/components/topology/components/nodes/WorkloadNode.tsx
```diff
@@ -40,18 +37,21 @@ export type WorkloadNodeProps = {
...
   dropTarget,
   dropTooltip,
-  cheURL,
   ...rest
 }) => {
+  const [consoleLinks] = useK8sWatchResource<K8sResourceKind[]>({
+    isList: true,
+    kind: referenceForModel(ConsoleLinkModel),
+  });
+  const cheURL = getCheURL(consoleLinks);
```
spadgett, line 54: You should be able to watch the one link with name `che`. But again I think we should avoid a separate watch for each node.

https://github.com/spadgett/console/blob/e47665c50e6f5ea842a2de42e88e60a3c2090452/frontend/packages/dev-console/src/components/topology/topology-utils.ts#L36-L37

frontend/public/components/dashboard/project-dashboard/project-dashboard.tsx
```diff
@@ -49,10 +45,11 @@ export const getNamespaceDashboardConsoleLinks = (
...
 };
 
-const ProjectDashboard_: React.FC<ProjectDashboardReduxProps & ProjectDashboardProps> = ({
-  obj,
-  consoleLinks,
-}) => {
+export const ProjectDashboard: React.FC<ProjectDashboardProps> = ({ obj }) => {
+  const [consoleLinks] = useK8sWatchResource<K8sResourceKind[]>({
+    isList: true,
+    kind: referenceForModel(ConsoleLinkModel),
```
spadgett, line 51: I think we probably want `optional: true` each place we watch these links.

rhamilto: D'oh.  Thanks.  I got distracted trying to figure out how to avoid the copy/pasting of the resource and completely forgot to add it.  Fixed.

frontend/packages/dev-console/src/components/guided-tour/GuidedTourText.tsx
rhamilto: I assume this is because there already is a `ConsoleLink` for the blog that's being used in the help menu?

spadgett: > I assume this is because there already is a `ConsoleLink` for the blog that's being used in the help menu?

That's right

frontend/packages/dev-console/src/components/topology/TopologyEdgePanel.tsx
rhamilto: Waiting on input from @christianvogt (who I believe is out on PTO).

christianvogt: @spadgett why can't we use `useK8sWatchResource` multiple times on the page with the same resource?  i thought we can trust the framework to manage only a single watch when using the hook multiple times on the same page. 

I ask because I want to move in this direction because it makes it easier for each node to manage the data it needs instead of always needing to surface this information to the topology root.

rhamilto:
/hold

for broken merge queue

frontend/packages/dev-console/src/components/topology/TopologyEdgePanel.tsx
spadgett: @christianvogt I was worried that we might have a race if two components start watching at the same time, but I think we're probably OK. The code to handle it is here:

https://github.com/spadgett/console/blob/b6ce43d2a4247538016b88e565288770feebe016/frontend/public/actions/k8s.ts#L146-L150

We should test to make sure only one list request and WebSocket gets opened.

rhamilto: > We should test to make sure only one list request and WebSocket gets opened.

@christianvogt, can you either do said testing or help me understand what I need to do to view `<TopologyEdgePanel>`?

christianvogt: `TopologyEdgePanel` is only rendered when selecting a link. Therefore there will only be a single watch here.

As for watching per workload node, I saw no extra network calls for the console links since a socket was already opened. Switching from and to topology multiple times with several workloads presents shows no additional console link watches.

rhamilto: @spadgett, are you therefore good with this PR as is?

rhamilto:
Rebase required.  Please retag, @spadgett.

spadgett:
/hold

Needs rebase again :(

rhamilto:
/hold cancel

Rebased.  @spadgett, PTAL.

PR #6187, opened 2020-08-03
Bug 1856675: Fix filter toolbar initialization from URL parameters

kyoto:
Fix a bug on the Monitoring pages where filters where preserved when
switching tabs (list was filtered, but the filter toolbar did not show
the applied filters).

With this change, the filters revert to default when clicking to a
different tab. When URL parameters exist (e.g. when using the browser
back button), those filters are applied to the list and shown in the
filter bar.

With this change, any `name` filter in the URL now causes the text input
to be populated with that value. This change applies to all pages that
use the filter toolbar. This seems like the correct behaviour since we
already set the text input value when the user switches from label
filter to name filter.

### DIFF
--- a/frontend/public/components/factory/table-filters.ts
+++ b/frontend/public/components/factory/table-filters.ts
@@ -48,10 +48,10 @@ export const tableFilters: TableFilterMap = {
   },
 
   alerts: (values, alert: Alert) => {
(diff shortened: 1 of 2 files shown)


kyoto:
\cc @bipuladh

PR #6188, opened 2020-08-03
Bug 1870622: Update topology list view to be more consistent with workloads page

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4418

**Description**
Update the topology list view to contain separate cells for memory and cpu metrics. Add polling for metric and alert updates. 

**Screenshots**

Before:
image

After:
image

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/list-view/TopologyListView.tsx
+++ b/frontend/packages/dev-console/src/components/topology/list-view/TopologyListView.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { DataList } from '@patternfly/react-core';
+import * as _ from 'lodash';
+import { connect } from 'react-redux';
 import {
   observer,
   isGraph,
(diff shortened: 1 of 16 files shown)


debsmita1:
@jeff-phillips-18 when the sidebar is open, the cells for memory & cpu usage don't show up. have a look at the below gif
listview

jeff-phillips-18:
@debsmita1 Correct. That is by design as in the project workloads tab in the admin perspective.

debsmita1:
verified this locally
/lgtm

frontend/packages/dev-console/src/components/topology/list-view/TopologyListView.tsx
```diff
@@ -139,6 +172,91 @@ const TopologyListView: React.FC<TopologyListViewProps> = ({
...
+          if (_.includes([401, 403, 502, 503], status) && _.isEmpty(metrics)) {
+            throw new Error(`Could not fetch metrics, status: ${status}`);
+          }
+        })
+        .then(() => {
+          metricsInterval = setTimeout(fetchMetrics, METRICS_POLL_INTERVAL);
+        })
+        .catch((e) => {
+          console.error(e); // eslint-disable-line no-console
+        });
```
invincibleJai, line 224: this logic is same as used in Admin side but that's a class based and this is functional connected component https://github.com/openshift/console/blob/master/frontend/public/components/overview/index.tsx#L267 , can we separate it and use across or it'll be beyond scope of this issue WDYT?

jeff-phillips-18: I can look into doing this. Eventually we will be replacing the workloads tab contents with the topology list view and the duplication will be removed (see https://github.com/openshift/console/pull/6224)

invincibleJai:
/lgtm

verified the changes, works as expected

jeff-phillips-18:
@invincibleJai squashed and rebased

christianvogt:
Unit test failure related to `observer`

PR #6189, opened 2020-08-03
Bug 1862934: Trigger the inactivity logout method when Masthead component is mounted

jhadvig:
We need to trigger the `_inactivityTimeout()` when `Masthead` component is mounted to start the countdown to cover cases when user will just login and wont do any activity.

/assign @spadgett

### DIFF
--- a/frontend/public/components/masthead-toolbar.jsx
+++ b/frontend/public/components/masthead-toolbar.jsx
@@ -81,8 +81,8 @@ class MastheadToolbarContents_ extends React.Component {
     this._onHelpDropdownToggle = this._onHelpDropdownToggle.bind(this);
     this._onAboutModal = this._onAboutModal.bind(this);
     this._closeAboutModal = this._closeAboutModal.bind(this);
-    this._onInactivityTimeout = this._onInactivityTimeout.bind(this);
-    this.userInctivityTimeout = null;
+    this._resetInactivityTimeout = this._resetInactivityTimeout.bind(this);
+    this.userInactivityTimeout = null;
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


frontend/public/components/masthead-toolbar.jsx
```diff
@@ -93,6 +93,7 @@ class MastheadToolbarContents_ extends React.Component {
     if (window.SERVER_FLAGS.inactivityTimeout >= 300) {
       window.addEventListener('click', _.throttle(this._onInactivityTimeout, 500));
       window.addEventListener('keydown', _.throttle(this._onInactivityTimeout, 500));
+      this._onInactivityTimeout();
```
spadgett, line 96: Do you know if the `OPENSHIFT` flag and `user` prop is resolved as this point? We could be racing.

Not new, but this method would be better called `_resetInactivityTimeout`. `_onInactivityTimeout` sounds like you were inactive for too long.

jhadvig:
> Do you know if the OPENSHIFT flag and user prop is resolved as this point? We could be racing.

Dont think thats the issue here. It's more likely due to a fact that we are event needs to be triggered. ATM we are only adding a listener.

spadgett:
> > Do you know if the OPENSHIFT flag and user prop is resolved as this point? We could be racing.
> 
> @spadgett dont think thats the issue here. It's more likely due to a fact that we are event needs to be triggered. ATM we are only adding a listener. If the user just logs in and wont do any interaction (wont trigger the event) the countdown wont start.

Right, I understand. But I think we're still racing since `_onInactivityTimeout` won't do anything if the user or the flag haven't resovled yet.

spadgett:
In other words, I'm not sure the fix will always work.

jhadvig:
@spadgett yes, the race was also part of the issue. Updated the PR so only if all conditions are met the timeout is being set.
PTAL

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -91,36 +91,41 @@ class MastheadToolbarContents_ extends React.Component {
...
+      !_.isEqual(user, prevProps.user)
     ) {
       this._updateUser();
     }
+    // Start the inactivity countdown when the flag and user are loaded and the timeout is not set
+    if (
+      window.SERVER_FLAGS.inactivityTimeout >= 300 &&
+      !this.userInactivityTimeout &&
+      flags[FLAGS.OPENSHIFT] &&
+      !_.isEmpty(user)
```
spadgett, line 113: I'm not sure this check is still necessary since we check auth mode on startup.

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -91,36 +91,41 @@ class MastheadToolbarContents_ extends React.Component {
...
+      flags[FLAGS.OPENSHIFT] !== prevProps.flags[FLAGS.OPENSHIFT] ||
+      !_.isEqual(user, prevProps.user)
     ) {
       this._updateUser();
     }
+    // Start the inactivity countdown when the flag and user are loaded and the timeout is not set
+    if (
+      window.SERVER_FLAGS.inactivityTimeout >= 300 &&
+      !this.userInactivityTimeout &&
+      flags[FLAGS.OPENSHIFT] &&
```
spadgett, line 112: Are we meaning to disable this when not OpenShift or are we simply checking if the flag is set? If the latter, you should use `flagPending`. Otherwise, we should move the OpenShift-specific logic below for logout. I don't think we need to restrict this to OpenShift, though.

spadgett: I'm not sure we would need to block on the flag either. Presumably the flag will resolve well before the inactivity timeout kicks in.

jhadvig:
@spadgett I've put the initial `this._resetInactivityTimeout()` call back into `componentDidMount` since we need to call it only once. Also I've realised that we dont need to pro-actively check for the flag of user.
PTAL

PR #6190, opened 2020-08-03
Bug 1862508: Create PVC dropdown text update

glekner:


glekner:
@matthewcarleton @yfrimanm wdyt?

frontend/packages/kubevirt-plugin/src/plugin.tsx
```diff
@@ -494,7 +494,7 @@ const plugin: Plugin<ConsumedExtensions> = [
   {
     type: 'PVCCreateProp',
     properties: {
-      label: 'New with Data Upload',
+      label: 'With Data upload form',
```
yaacov, line 497: the BZ say "With form and Data upload" ?

glekner: yes and I'm asking from Matt to rethink it :)

PR #6192, opened 2020-08-03
Bug 1854196: Use consistent title case for operand form labels

TheRealJon:
Apply lodash startCase to operand form field labels that are not explicitely set by descriptor (`displayName` property) or schema (`title` property).


spadgett:
/hold

It looks like this breaks CI.

TheRealJon:
> It looks like this breaks CI.

@spadgett Yeah, there was an integration test that needed to be updated.

PR #6193, opened 2020-08-03
Bug 1863402: Developer console tables show extra columns due to projects table columns change

DaoDaoNoCode:
Add `isColumnManagementEnabled` props to `ProjectTableRow` so it will not render the two additional columns even if they are chosen in the page under admin view.

### DIFF
--- a/frontend/public/components/namespace.jsx
+++ b/frontend/public/components/namespace.jsx
@@ -537,7 +537,13 @@ const ProjectTableRow = connect(projectRowStateToProps)(
   ({ obj: project, index, rowKey, style, customData = {}, metrics, selectedColumns }) => {
     const name = getName(project);
     const requester = getRequester(project);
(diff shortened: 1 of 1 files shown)


spadgett:
@DaoDaoNoCode Is it possible to enable custom columns for the Developer perspective as well? As a user, I'd want my columns to be used in both perspectives.

I'm not familiar with how the Developer perspective is changing the Project list. cc @christianvogt

DaoDaoNoCode:
@spadgett Yes I think it's possible to enable custom columns for the Developer perspective. However, this is likely adding new features that is not included in version 4.6. What's more, if it is added, all the pages with a table under Developer perspective can use column management modal when "all projects" is selected. Do we still need to implement it now?

spadgett:
Closing in favor of #6601

/close

PR #6194, opened 2020-08-03
i18n support

rebeccaalpert:
Opening [Joachim's i18n localization research PR](https://github.com/jschuler/console/pull/1) against master for review.

See original PR for implementation details and suggestions. I updated the code to support locales folders in packages. Each package can have its own locales folder and its own namespaces. (Note that the public namespace is merged with the one in public/locales/${lng}/public.json.) I did three components in console-shared as an example of how to use i18n in packages.

I also added a pseudolocalization plugin and five utilities to make it easier to handle files and file conversions:
* `build-i18n.sh` (runs the parser for public and every package)
* `consolidate-public-folders.js` (runs after the parser and merges any namespace files in packages that overlap with namespace files in public if there are no key conflicts)
* `set-english-defaults.js` (runs after the parser and pre-fills English i18n files with the text set in the keys -- otherwise they would be blank by default)
* `yarn i18n-to-pot` (creates a .pot file in frontend/public/locales/pots or frontend/packages/${package}/locales/pot based on the parameters you pass in)
* `yarn po-to-i18n` (looks for files in frontend/public/locales/pos or frontend/packages/${package}/locales/pos and creates i18n files in the correct language folders)

`build-i18n.sh`, `consolidate-public-folders.js`, and `set-english-defaults.js` all run when you call `yarn i18n`.

Use the query parameter `?pseudolocalization=true&lng=en` to see pseudolocalization on strings you've marked.

This PR also includes a custom Jest resolver by Christian Vogt so i18n will play nice with our tests and a cool Cypress child command from Dave Taylor.


rebeccaalpert:
Working on resolving conflicts.

frontend/i18next-parser.config.js
```diff
@@ -0,0 +1,86 @@
...
+  sort: false,
+  // Whether or not to sort the catalog
+
+  useKeysAsDefaultValue: true,
+  // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
+  // The option `defaultValue` will not work if this is set to true
+
+  verbose: false
+  // Display info about the parsing including some stats
+}
```
spadgett, line 86: Missing final newline

frontend/i18n.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }, () => { 
+    moment.locale(i18n.language === 'zh' ? 'zh-cn' : i18n.language); 
+  });
+
+i18n.on('languageChanged', function(lng) {
+  moment.locale(lng === 'zh' ? 'zh-cn' : lng);
+});
+
+export default i18n;
```
spadgett, line 63: Missing final newline

frontend/locales/ar/nav.json
```diff
@@ -0,0 +1,64 @@
...
+  "Logging": "عربي Logging",
+  "Home": "عربي Home",
+  "Overview": "عربي Overview",
+  "Projects": "عربي Projects",
+  "Search": "عربي Search",
+  "Explore": "عربي Explore",
+  "Events": "عربي Events",
+  "Operators": "عربي Operators",
+  "Workloads": "عربي Workloads",
+  "Pods": "عربي Pods",
```
spadgett, line 15: Many of these items we probably don't want to translate as they're proper names. I think anything that is a k8s resource name probably should not be translated.

frontend/public/components/app.jsx
```diff
@@ -136,11 +139,13 @@ class App extends React.PureComponent {
   render() {
     const { isNavOpen, isDrawerInline } = this.state;
     const { productName } = getBrandingDetails();
+    const { t, i18n } = this.props;
 
     return (
       <>
         <Helmet titleTemplate={`%s · ${productName}`} defaultTitle={productName} />
         <ConsoleNotifier location="BannerTop" />
+    <div>Current locale: {i18n.language} | moment locale: {moment.locale()} | {t('Date: It is now {{date, MM/DD/YYYY}}', { date: new Date() })} | {t('Number: ${{value, number}}', { value: 1550.95 })}</div>
```
spadgett, line 148: We'll want to remove this when we're ready to merge.

frontend/public/components/lang-selector.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
```
spadgett, line 1: We should remove the lang selector from the masthead before we merge and use the browser language settings insetad. At some point we can add our own selector, but that can be a follow on that needs design input.

frontend/__tests__/components/utils/datetime.spec.ts
```diff
@@ -33,36 +33,36 @@ describe('fromNow', () => {
       fromNow(new Date('Jan 01 1970 00:00:00'), new Date('Jan 01 1970 00:00:02'), {
         omitSuffix: true,
       }),
-    ).toEqual('few seconds');
+    ).toEqual('a few seconds');
```
spadgett, line 36: I don't think we need these tests any longer if we're just using the moment.js implementation. We're basically testing the moment.js code :)

rebeccaalpert:
Rebased; working through PR feedback now.

frontend/locales/ar/nav.json
rebeccaalpert: The Kubernetes documentation ([Pods example](https://kubernetes.io/docs/concepts/workloads/pods/)) is actually localized, and I noticed things like pluralization differences in Chinese/Japanese compared to English/Spanish even though they're using English for the term. I'm wondering if it's still something that needs to be localized even if it's not translated? I don't know any Chinese and my Japanese is pretty bad. Do you think it would be useful to get translation to weigh in?

rebeccaalpert:
Not ready for review; pushing updates that'll make it easier to do a pre-test with the localization QE team.

rebeccaalpert:
I removed the language select menu and language display bar at the top. Language now depends on the browser; to test this, just change the language on the browser, clear localStorage, and refresh.

rebeccaalpert:
Added dummy translations from the globalization QE team so they can take a look.

rebeccaalpert:
I removed language caching so we're consistent with other products that the g11n QE team has worked on. You just need to refresh after changing the language setting/locale.

rebeccaalpert:
I added three command-line utilities to make it easier to handle files and file conversions:

- `yarn i18n-to-pot` (asks you what file you want to export and creates a .pot file in frontend/locales/pots)
- `yarn po-to-i18n` (looks for files in frontend/locales/pos and creates i18n files in the correct language folders)
- `set-english-defaults.js` (runs after `yarn i18n` and pre-fills English i18n files with the text set in the keys)

`yarn i18n` was added by Joachim and is used to automatically generate the i18n files based on the strings marked in the application.

I'm really rusty on this kind of stuff, so feedback would be great.

rebeccaalpert:
Added plugin for psuedo-language. This is applied to English by default. We have to disable this and enable it only when we're working on i18n stuff.

Screen Shot 2020-09-10 at 12 44 54 PM
Screen Shot 2020-09-10 at 12 49 44 PM

rebeccaalpert:
Use the "?pseudolocalization" query parameter to turn on pseudolocalization for English-language marked strings.

frontend/i18n.js
```diff
@@ -0,0 +1,71 @@
...
+import Pseudo from 'i18next-pseudo';
+
+// Load moment.js locales (en is default and always loaded)
+import 'moment/locale/zh-cn';
+import 'moment/locale/ja';
+import moment from 'moment';
+
+const { FALLBACK_LOCALE } = require('./i18next-parser.config');
+const params = new URLSearchParams(window.location.search);
+const isPseudoLocalizationEnabled = params.get('pseudolocalization');
```
spadgett, line 14:
```suggestion
const pseudolocalizationEnabled = params.get('pseudolocalization') === 'true';
```

frontend/i18n.js
```diff
@@ -0,0 +1,71 @@
+import i18n from 'i18next';
```
spadgett, line 1: This should probably go under `frontend/public` if it's part of the console app.

frontend/locales/ja/nav.json
```diff
@@ -0,0 +1,69 @@
+{
```
spadgett, line 1: I'd suggest we remove these if they aren't real translations. We can use the pseudolocalizations to test and do translation verification testing when we get the real translations back.

frontend/public/components/app.jsx
```diff
@@ -249,17 +254,21 @@ if ('serviceWorker' in navigator) {
...
-        <Route
-          path="/k8s/ns/:ns/virtualmachineinstances/:name/standaloneconsole"
-          render={(componentProps) => <AsyncComponent loader={consoleLoader} {...componentProps} />}
-        />
-        <Route path="/terminal" component={CloudShellTab} />
-        <Route path="/" component={App} />
-      </Switch>
-    </Router>
-  </Provider>,
+  <React.Suspense fallback={<Spinner />}>
```
spadgett, line 257: Is this needed? If it is, we should use the common loading component we use throughout console.

frontend/locales/ja/nav.json
rebeccaalpert: Let me just check with the globalization team. They provided these as part of their workflow. I'll see if they're flexible.

frontend/i18n.js
rebeccaalpert: Should other config files and the conversion js go there too?

spadgett: > Should other config files and the conversion js go there too?

No, not if they're not part of the console app itself.

frontend/locales/ja/nav.json
spadgett: If they're real translations, we can keep them.

spadgett:
@rebeccaalpert We should add some a few Cypress e2e tests that make sure the pseudolocalizations work. We shouldn't need to check every string, just a few from different areas of the UI to make sure it's working.

rebeccaalpert:
Working on cypress test; debugging a weird permissions issue it has when I try to run it right now.

frontend/packages/integration-tests-cypress/views/navigation.ts
```diff
@@ -0,0 +1,9 @@
+export const navigation = {
```
dtaylor113, line 1: Thanks for doing this helper, unfortunately a similar helper was just [merged](https://github.com/openshift/console/pull/6521/files#diff-bfd2ae71744a1dd4227111900750f99fR14). Pls uses `nav` helper -thanks

frontend/packages/integration-tests-cypress/tests/pseudotranslation.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  after(() => {
+    cy.deleteProject(testName);
+    cy.logout();
+  });
+
+  it('pseudolocalizes navigation', () => {
+    cy.log('test navigation');
+    cy.visit('/dashboards?pseudolocalization=true');
+    nav.sidenav.switcher.shouldHaveText('[Ḥṓṓṃḛḛ]');
```
spadgett, line 22: I'd suggest we add a helper that checks if the string starts with `[`, ends with `]`, and doesn't have `a-zA-Z`. Then we can have a list of URLs / CSS selectors to check and add tests really easily without needing to hard-code every pseudo-localization.

dtaylor113: Hi, it's not `nav.sidenav.switcher`, rather `nav.sidenav.shouldHaveNavSection`.  The `switcher` is 'Administration' .vs 'Developer' views.

TheRealJon:
There may be some limitations or restrictions of the i18next package, but I'm wondering if it would make sense for us to have a `frontend/i18n` folder where most of these new files live? Also, I think we need to allow static packages to provide their own localization files.

rebeccaalpert:
Added Cypress tests for nav and timestamps. @dtaylor113 - would you mind taking a look sometime this week? I'd appreciate feedback.

frontend/packages/integration-tests-cypress/tests/i18n/pseudotranslation.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  after(() => {
+    cy.logout();
+  });
+
+  it('pseudolocalizes navigation', () => {
+    cy.log('test navigation');
+    cy.visit('/dashboards?pseudolocalization=true&lng=en');
+    nav.sidenav.shouldMatchRegex(/\[[αḅͼḍḛϝḡḥḭĵḳḽṃṇṓṗʠṛṡṭṵṽẁẋẏẓḀḂḈḌḚḞḠḤḬĴḰḺṀṄṎṔǪṚṢṪṲṾŴẊŶŻ]+\]/);
```
spadgett, line 21:
```suggestion
    nav.sidenav.shouldMatchRegex(/\[[^a-zA-Z]+\]/);
```

We should make this a reusable helper.

rebeccaalpert: I tried the [^a-zA-Z] approach yesterday in one of those regex testing websites and it wasn't passing.

rebeccaalpert:
Updated tests.

rebeccaalpert:
Seemingly fixed typescript issue with test. Used suggestions from here: https://stackoverflow.com/questions/40406728/typescript-throw-error-property-to-does-not-exist-on-type-matchers-when-comp

rebeccaalpert:
Added big angry error for when there are missing keys. It should break CI tests based on what I've seen locally.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { checkErrors } from '../../support';
+import { timestamp } from '../../views/timestamp';
+import { shouldMatchRegex } from '../../views/common';
+
+describe('Localization', () => {
+  before(() => {
+    cy.login();
```
dtaylor113, line 7: I guess you don't have to create a test project?

rebeccaalpert: Do I need to? If I need to, I'll add it back in.

frontend/public/components/nav/items.tsx
```diff
@@ -29,7 +29,13 @@ export const stripNS = (href) => {
 
 export const ExternalLink = ({ href, name }) => (
   <NavItem isActive={false}>
-    <a className="pf-c-nav__link" href={href} target="_blank" rel="noopener noreferrer">
+    <a
+      className="pf-c-nav__link"
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
+      data-cy="nav"
```
dtaylor113, line 37: This should be `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress)

frontend/public/components/nav/items.tsx
```diff
@@ -83,6 +89,7 @@ class NavLink<P extends NavLinkProps> extends React.PureComponent<P> {
           onClick={onClick}
           title={tipText}
           data-tour-id={dataTourId}
+          data-cy="nav"
```
dtaylor113, line 92: This should be `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress)

frontend/public/components/nav/section.tsx
```diff
@@ -209,6 +209,7 @@ export const NavSection = connect(navSectionStateToProps)(
             isActive={isActive}
             isExpanded={isOpen}
             onExpand={this.toggle}
+            data-cy="nav"
```
dtaylor113, line 212: This should be `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress)

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,11 @@
+import { expect } from 'chai';
+
 export const resourceStatusShouldContain = (desiredStatus: string, options?: any) =>
   cy.contains('[data-test="status-text"]', desiredStatus, options);
+
+export const shouldMatchRegex = (selector: string, regex: any) => {
```
dtaylor113, line 6: Maybe should add `regex` to  `cy.byTestID(id: string, regex:any)` and put the regex handling in `cy.byTestID` if regex param detected.

frontend/packages/integration-tests-cypress/views/timestamp.ts
```diff
@@ -0,0 +1,3 @@
+export const timestamp = {
+  isLoaded: () => cy.get('[data-cy=timestamp]').should('exist'),
```
dtaylor113, line 2: This should be `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress) and `cy.byTestID('timestamp').should('exist')`

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  after(() => {
+    cy.logout();
+  });
+
+  it('pseudolocalizes navigation', () => {
+    cy.log('test navigation');
+    cy.visit('/dashboards?pseudolocalization=true&lng=en');
+    shouldMatchRegex('[data-cy=nav]', /\[[^a-zA-Z]+\]/);
```
dtaylor113, line 21: Maybe should add `regex` to  `cy.byTestID(id: string, regex:any)` and put the regex handling in `cy.byTestID` if regex param detected.  We should be using `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress) which lends itself to `cy.byTestID`.

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,11 @@
+import { expect } from 'chai';
```
dtaylor113, line 1: Do we need this import?  I've been using `expect` without declaring an import.

rebeccaalpert: I was getting this weird Typescript error without it: https://stackoverflow.com/questions/40406728/typescript-throw-error-property-to-does-not-exist-on-type-matchers-when-comp

If you have suggestions, I'm here for them!

dtaylor113:
Hi @rebeccaalpert, reviewed the Cypress tests, in general we should be using the attribute `data-test` per [our best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#migrating-protractor-tests-to-cypress).  -thanks

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: That what I was questioning :-)  I guess not if all the tests i18n tests work.  I believe we only create projects if we CRUD resources, so when we delete the test project all artifacts are also cleanned up.  So, if i18n doesn't CRUD anything, I think we're good

rebeccaalpert:
Updated tests per @dtaylor113 's feedback.

frontend/packages/integration-tests-cypress/views/timestamp.ts
```diff
@@ -0,0 +1,3 @@
+export const timestamp = {
```
dtaylor113, line 1: Hi, do you feel there will be many other helper methods for timestamp?  Not sure it needs to be in it's own file, perhaps move to `views/common.ts`

dtaylor113: Hi @rebeccaalpert, thanks for the cypress updates, LGTM!  Last little nit. ^^

rebeccaalpert: All set @dtaylor113! Thanks.

rebeccaalpert:
I'm getting a test failure on CI for timestamp test - I'll investigate more when I'm back from PTO. It runs ok locally but probably needs fine-tuning.

spadgett:
Adding @ahardin-rh for docs approval. @ahardin-rh We're starting work on localizing the console. We plan to add a language switcher in a future PR. Right now it will pick the language based on the browser settings.

frontend/locales/en/nav.json
```diff
@@ -0,0 +1,68 @@
+{
```
spadgett, line 1: @rebeccaalpert We'll need to make sure the checked in i18n files are current in CI so we don't forget to run `yarn i18n`. I'd probably just update test-frontend.sh to run `yarn i18n` and check if there is a git diff. Something like this (untested)

```bash
GIT_STATUS="$( git status --short --untracked-files -- frontend/locales )"
if [[ -n "$GIT_STATUS" ]]; then
  echo "i18n files are not up to date. Run `yarn i18n` to re-generate."
  git diff
  exit 1
fi
```

rebeccaalpert:
Updated ./test-frontend.sh. I tested it without the linter since that never completes on my machine.

test-frontend.sh
```diff
@@ -12,10 +12,18 @@ if [ "$OPENSHIFT_CI" = true ]; then
     JEST_SUITE_NAME="OpenShift Console Unit Tests" JEST_JUNIT_OUTPUT_DIR="$ARTIFACT_DIR" yarn run test --ci --reporters=default --reporters=jest-junit
 else
     yarn run test
+yarn i18n
```
spadgett, line 15: Move this outside of the if/else. It needs to happen in CI, too. I've move it just before the `GIT_STATUS` check below.

frontend/i18next-parser.config.js
```diff
@@ -0,0 +1,88 @@
...
+
+  indentation: 2,
+  // Indentation of the catalog files
+
+  keepRemoved: false,
+  // Keep keys from the catalog that are no longer in code
+
+  keySeparator: false,
+  // Key separator used in your translation keys
+  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.
```
ahardin-rh, line 37: s/plan english/Plain English

    `namespaceSeparator: false` so that, for example, `t('Status: Loading...')` will not think that there is a namespace and three separator dots.

frontend/i18next-parser.config.js
```diff
@@ -0,0 +1,88 @@
...
+
+  lineEnding: 'auto',
+  // Control the line ending. See options at https://github.com/ryanve/eol
+
+  locales: Object.keys(SUPPORTED_LOCALES),
+  // An array of the locales in your applications
+
+  namespaceSeparator: '~',
+  // Namespace separator used in your translation keys
+  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.
```
ahardin-rh, line 64: same suggestion as above

frontend/i18next-parser.config.js
```diff
@@ -0,0 +1,88 @@
...
+  // see below for more details
+  lexers: {
+    hbs: ['HandlebarsLexer'],
+    handlebars: ['HandlebarsLexer'],
+
+    htm: ['HTMLLexer'],
+    html: ['HTMLLexer'],
+
+    mjs: ['JavascriptLexer'],
+    js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
```
ahardin-rh, line 48: s/you're/you are

ahardin-rh:
:+1: from the docs side. I realize that some of my comments are on commented out text, so feel free to disregard if they are superfluous :smile:

rebeccaalpert:
Updated script, updated comments, and I think I fixed the failing CI test.

rebeccaalpert:
I updated the parser to handle multiple locale folders in packages, as well as our locale webpack handling and scripts. Webpack seems to be working ok; it seems like `yarn run dev` just needs to be run again after `yarn i18n` is run. I'll work with Joachim to see if we can get this to be smoother.

I added an `en-gb` locale so we could adjust the date format, but it looks like we will have to maintain "translations" for text if we provide date support.

Updated script details:
- `yarn i18n-to-pot` now parses all packages in addition to public
  - Options:
    - -h, -help
    - -l, -language (default is Japanese - “ja”)
    - -p, -package (defaults to public folder if none is specified)
    - -f, -file (any number of files)
  - Examples
    - Create files from package: `yarn i18n-to-pot -f='modal' -p='console-shared'`
    - Create files from public: `yarn i18n-to-pot -f='nav' -f='modal'`
  - POT files are placed in the `locales/pot` folder for public or the package (if a package is provided)
- `yarn po-to-i18n` now parses all packages in addition to public
    - PO files must be placed in `frontend/public/locales/pos` or `frontend/packages/${package}/locales/pos` in order to be processed
- English defaulter now parses all packages in addition to public. Note that this doesn't include English locale variants at the moment, such as English (GB).

rebeccaalpert:
Removed `en-gb` from the parser config so we don't need to maintain non-date/number language files manually. This means we'll need to manually enter dates/times for English locales we want to support though.

frontend/public/i18n.js
```diff
@@ -0,0 +1,76 @@
...
+import httpBackend from 'i18next-http-backend';
+import Pseudo from 'i18next-pseudo';
+
+// Load moment.js locales (en is default and always loaded)
+import 'moment/locale/zh-cn';
+import 'moment/locale/ja';
+import 'moment/locale/en-gb';
+import moment from 'moment';
+
+const { FALLBACK_LOCALE } = require('../../i18next-parser.config');
```
rhamilto, line 13:
```suggestion
const { FALLBACK_LOCALE } = require('../i18next-parser.config');
```

rebeccaalpert:
This is currently broken. Don't rebase!

rebeccaalpert:
Updated strings in masthead, activity card, and utilization card and added tests. Still not safe to rebase as jest tests are still broken.

rebeccaalpert:
Removed blank language file generation.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    cy.log('test navigation');
+    cy.visit('/dashboards?pseudolocalization=true&lng=en');
+    cy.byTestID('nav', /\[[^a-zA-Z]+\]/);
+  });
+
+  it('pseudolocalizes activity card', () => {
+    cy.log('test activity card components');
+    cy.visit('/dashboards?pseudolocalization=true&lng=en');
+    cy.byTestID('activity', /\[[^a-zA-Z]+\]/);
+    cy.byTestID('activity-recent-title', /\[[^a-zA-Z]+\]/);
```
dtaylor113, line 35: Can we create a const for this in `frontend/packages/integration-tests-cypress/views/common.ts`?  Something like `export const pseudoi18nRegex = /\[[^a-zA-Z]+\]/`  -thanks

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    cy.log('test utilization card components');
+    cy.visit('/dashboards?pseudolocalization=true&lng=en');
+    cy.byTestID('utilization-card-item-text', /\[[^a-zA-Z]+\]/);
+  });
+
+  it('pseudolocalizes timestamps', () => {
+    cy.log('test timestamps');
+    cy.visit('/k8s/all-namespaces/events?pseudolocalization=true&lng=en');
+    isLoaded('timestamp');
+    cy.byTestID('timestamp', /\[[^a-zA-Z]+\]/);
```
dtaylor113, line 51: Not sure you need the `isLoaded`.   `cy.byTestID('timestamp', /\[[^a-zA-Z]+\]/)` will fail if the elem doesn't exist.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: Found a better way to do this, I think, pls see: https://github.com/rebeccaalpert/console/pull/11

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
rebeccaalpert: Seems happy enough without it!

frontend/packages/console-shared/locales/en/activity-card.json
```diff
@@ -0,0 +1,8 @@
+{
```
spadgett, line 1: This feels like it might be a little too granular. Should there be a cluster-overview or dashboard namespace for all of these strings?

The browser will have to make a separate XHR for all of these files, which has a performance impact.

frontend/packages/console-shared/src/components/dashboard/utilization-card/UtilizationItem.tsx
```diff
@@ -212,12 +215,28 @@ export const UtilizationItem: React.FC<UtilizationItemProps> = React.memo(
...
+              <span
+                className="co-utilization-card__item-text"
+                data-test="utilization-card-item-text"
+              >
+                {humanLimit && (
+                  <span>{t('utilization-card~{{humanLimit}} total limit', { humanLimit })}</span>
+                )}
+                {!humanLimit && humanMax && (
+                  <span>{t('utilization-card~of {{humanMax}}', { humanMax })}</span>
+                )}
```
spadgett, line 239: I'm not sure we can break this phrase up like this. You'll want separate properties for the case where there is a just a limit and another property with the full phrase for limit of max.

frontend/public/i18n.js
```diff
@@ -0,0 +1,77 @@
...
+        },
+        escapeValue: false, // not needed for react as it escapes by default
+      },
+      react: {
+        useSuspense: true,
+        wait: true,
+      },
+      saveMissing: true,
+      missingKeyHandler: function(lng, ns, key) {
+        throw new Error(`Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`);
```
spadgett, line 65: Will this cause a whitespace or other user-visible error in console when it happens? If so, we should change how this works. We could just to set `window.windowError` here and make sure it's logged to the console.

rebeccaalpert: @spadgett - How do I look for a whitespace error? i18n is already configured to log the missing key to console.

spadgett: Sorry I meant whitescreen error. If I have a missing key, does it break console entirely if we throw an error here?

rebeccaalpert: It does. I can remove it.

rebeccaalpert:
@spadgett - Addressed all PR feedback; let me know if you have other feedback. 

Missing keys now only generate an "info" message like the following: `i18next::translator: missingKey en-GB dashboard Usage Usage.`

frontend/i18n-scripts/build-i18n.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -exuo pipefail
+
+i18next "public/!(dist)**/**/*.{js,jsx,ts,tsx}" [-oc] -c "./public/i18next-parser.config.js" -o "public/locales/\$LOCALE/\$NAMESPACE.json"
```
spadgett, line 5: Where do `$LOCALE` and `$NAMESPACE` get defined? I'd expect this script to fail when you run it since you have `set -u`.

spadgett:
> @spadgett - Addressed all PR feedback; let me know if you have other feedback.
> 
> Missing keys now only generate an "info" message like the following: `i18next::translator: missingKey en-GB dashboard Usage Usage.`

Will these fail the e2e tests? We want to fail if there is a key missing.

spadgett:
@yapei This is ready for QE review.

@rebeccaalpert Can you let @yapei know what strings to are expected to be translated with this change how to test pseudolocalization? Thanks!

frontend/i18n-scripts/build-i18n.sh
rebeccaalpert: They don't get defined in the file. I was following the parser docs here: https://github.com/i18next/i18next-parser. The locale and namespace allegedly get injected.

rebeccaalpert:
@spadgett - Let me double check.

@yapei - The only strings that are actually translated are timestamps. You can see these by appending these query parameters to the URL:

- `?lng=ja` for Japanese
- `?lng=zh` for Chinese
- `?lng=en-gb` for UK 24-hour time

You can test pseudolocalization by using  `?pseudolocalization=true&lng=en`. This will put some bracketed text in strings marked for translation. You should see these in the nav, parts of the masthead (some of the strings in the action menu aren't accessible to us), and two of the boxes in the Overview dashboard.

frontend/i18n-scripts/consolidate-public-folders.js
```diff
@@ -2,67 +2,70 @@ const fs = require('fs');
 const path = require('path');
 const common = require('./common.js');
 
+// eslint-disable-next-line no-undef
```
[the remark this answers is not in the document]
rebeccaalpert: I can look into it as part of the common components PR if that's ok with you.

rhamilto:
When I run `yarn i18n` against my branch based on this branch, I get an unexpected `public_old.json` file.  I assume that's because of an issue in this branch as I am not adding any keys to the `public` namespace in my branch.

rebeccaalpert:
Hey @rhamilto - That's expected behavior. There are a couple of keys in the public namespace in the package which are being consolidated with the public namespace after the parser is run in order to avoid namespace conflicts. The parser creates the _old files, and the scripts are configured to ignore them. If @spadgett is interested in scrapping the _old files, it's a  small change we can address in a follow-on.

rhamilto:
I suspect what will happen is `public_old.json` will be inadvertently committed to the repo at some point since any run of `yarn i18n` is going to result in the file on disk.

spadgett:
We'll want to add it to `.gitignore`.

@rebeccaalpert Do you mind rebasing so QE will be able to test?

spadgett:
```
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-shared/src/components/dashboard/activity-card/EventItem.tsx
(46,19): Type '{ alt: string; title: string; className: string; }' is not assignable to type 'IntrinsicAttributes & ColoredIconProps & { children?: ReactNode; }'.
  Property 'alt' does not exist on type 'IntrinsicAttributes & ColoredIconProps & { children?: ReactNode; }'.
```

yanpzhan:
Spined up a cluster with cluster-bot against the pr, the ja/zh/en-gb language works for timestamp, and use 
"?pseudolocalization=true&lng=en", could see bracketed text in strings marked for translation on nav menus, overview page and masthead dropdown items.

yanpzhan:
/lgtm
Does QE need use /lgtm label? I'm not sure if /qe-approved takes effect, since I didn't see any change after add the label.

spadgett:
Added qe-approved and docs-approved based on comments from @yanpzhan and @ahardin-rh (thanks!).

reestr:
lgtm

spadgett (approved):
Thanks, @rebeccaalpert! 👍 

/lgtm

pweil-:
> Does QE need use /lgtm label? I'm not sure if /qe-approved takes effect, since I didn't see any change after add the label.

@yanpzhan (cc @ahardin-rh @reestr) Currently you'd need to use the label command.  IIRC there are no restrictions on who can add them.  So give this a shot on the next one:

/label qe-approved

PR #6195, opened 2020-08-04
Bug 1856709: Remove single instance of whitespace from user input

bipuladh:
Screenshot from 2020-08-04 10-56-02


bipuladh:
/hold
A better approach would be to do group capture

frontend/public/components/autocomplete.tsx
```diff
@@ -57,8 +57,10 @@ const AutocompleteInput: React.FC<AutocompleteInputProps> = (props) => {
   React.useEffect(() => {
     if (textValue && visible && showSuggestions) {
       const processed = labelParser(data, labelPath);
+      // User input without whitespace
+      const processedText = textValue.trim().replace(/([^\s]*)\s*(=|^)\s*(.*)/, '$1$2$3');
```
kyoto, line 61: Couldn't we just do `replace(/\s*=\s*/, '=')`?

PR #6198, opened 2020-08-04
Bug 1858229: Make Mode Detection independent of Storage Cluster CR name

bipuladh:
Detection of storage cluster mode should be based on `spec.externalStorage` and not on the CR name. 
/assign @cloudbehl @afreen23 
cc @umangachapagain

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/features.ts
+++ b/frontend/packages/ceph-storage-plugin/src/features.ts
@@ -9,14 +9,13 @@ import { fetchK8s } from '@console/internal/graphql/client';
 import { StorageClassModel } from '@console/internal/models';
 import { OCSServiceModel } from './models';
 import {
-  OCS_EXTERNAL_CR_NAME,
   CEPH_STORAGE_NAMESPACE,
   OCS_SUPPORT_ANNOTATION,
   ATTACHED_DEVICES_ANNOTATION,
-  OCS_INTERNAL_CR_NAME,
   RGW_PROVISIONER,
   SECOND,
 } from './constants';
+import { OCSStorageClusterKind } from './types';
 
 export const OCS_INDEPENDENT_FLAG = 'OCS_INDEPENDENT';
 export const OCS_CONVERGED_FLAG = 'OCS_CONVERGED';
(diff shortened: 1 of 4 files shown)


frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -70,33 +68,32 @@ export const detectRGW: FeatureDetector = async (dispatch) => {
 
 export const detectOCS: FeatureDetector = async (dispatch) => {
   try {
-    const storageCluster = await fetchK8s(
-      OCSServiceModel,
-      OCS_INTERNAL_CR_NAME,
-      CEPH_STORAGE_NAMESPACE,
-    );
+    const storageClusters = await k8sList(OCSServiceModel, { ns: CEPH_STORAGE_NAMESPACE });
+    const storageCluster = storageClusters?.[0];
```
cloudbehl, line 72: The one with the not Ignored needs to be checked. Because there will be only one in that case.

I have seen other dashboards getting enabled when 0th index is picked.

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -70,33 +68,32 @@ export const detectRGW: FeatureDetector = async (dispatch) => {
...
+    const isInternal = _.isEmpty(storageCluster.spec.externalStorage);
```
afreen23, line 73: Is the `spec.externalStorage` always present ?
If not, then `_,isEmpty` will return `true`  when `spec.externalStorage === undefined`.
If, yes, then I guess it will be always `{}` for attached devices and internal mode ?

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -70,33 +68,32 @@ export const detectRGW: FeatureDetector = async (dispatch) => {
...
-    dispatch(setFlag(OCS_CONVERGED_FLAG, true));
-    dispatch(setFlag(OCS_INDEPENDENT_FLAG, false));
-  } catch (e) {
-    if (e?.response?.status !== 404) handleError(e, [OCS_CONVERGED_FLAG], dispatch, detectOCS);
-    else {
+    if (isInternal) {
+      dispatch(setFlag(OCS_CONVERGED_FLAG, true));
+      dispatch(setFlag(OCS_INDEPENDENT_FLAG, false));
+    } else {
+      dispatch(setFlag(OCS_INDEPENDENT_FLAG, true));
```
afreen23, line 82:
```suggestion
      dispatch(setFlag(OCS_CONVERGED_FLAG, isInternal));
      dispatch(setFlag(OCS_INDEPENDENT_FLAG, !isInternal));
```

frontend/packages/ceph-storage-plugin/src/features.ts
bipuladh: yes.

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -70,34 +69,30 @@ export const detectRGW: FeatureDetector = async (dispatch) => {
...
   try {
-    const storageCluster = await fetchK8s(
-      OCSServiceModel,
-      OCS_INTERNAL_CR_NAME,
-      CEPH_STORAGE_NAMESPACE,
+    const storageClusters = await k8sList(OCSServiceModel, { ns: CEPH_STORAGE_NAMESPACE });
+    const storageCluster = storageClusters.find(
+      (sc: OCSStorageClusterKind) => sc.status.phase !== 'Ignored',
     );
+    const isInternal = _.isEmpty(storageCluster.spec.externalStorage);
```
afreen23, line 76: What if I make it `false` ? 
Should not this check be based on the value for `spec.externalStorage.enable`  ?
```
        externalStorage: {
          enable: false,
        },
```

bipuladh: In external mode, we set it to false. For every other mode, it's an empty object. Never is there a situation where `externalStorage.enable` is `false`. You can't flip a storage cluster to external/internal by just changing that field.

afreen23: Checking over spec.externalStorage.enable would be still a better option IMO.
It will scale well wrt any changes in future.
Right now UI is assumed to be not sending this field for internal and attached devices, operator by default makes it {} .
If any change happen we need to track to make the fix here as well.

spadgett:
/hold

PR #6199, opened 2020-08-04
Bug 1866214: Delete only KSVC, not Route associated with KSVC

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4259

**Analysis / Root cause**: 
If user choose to remove that application with knative service, it succeeds but an error is thrown intermittently as even KSVC tried to delete all associated resources i.e CFG, REV, Route etc 

**Solution Description**: 
Should Delete only KSVC

**Unit test**
image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
On deleting KSVC, it automatically removes its related resources like routes, revisions etc, so it makes sense why we shouldn't be deleting the route explicitly.
Verfied locally, works as expected
/lgtm

PR #6203, opened 2020-08-04
Add namespace choice to obc creation function

ebondare:


afreen23:
@ebondare can you specify in commit the way you made it stable.
It will help track. We can have many patches go by `Make obc tests more stable` , which will make us lose track of things.

Thanks for PR :)

afreen23:
/retitle Add namespace choice to obc creation function
/lgtm

spadgett:
/hold

PR #6204, opened 2020-08-04
Bug 1865828: Components with withHandlePromise wrapper should be called accordingly

jhadvig:
https://github.com/openshift/console/pull/6048 followUp 

/assign @spadgett 
@glekner fyi


frontend/public/components/modals/configure-count-modal.tsx
```diff
@@ -3,29 +3,29 @@ import * as React from 'react';
...
     const invalidateState = props.invalidateState || _.noop;
 
     invalidateState(true, _.toInteger(value));
-    props
-      .handlePromise(k8sPatch(props.resourceKind, props.resource, patch))
-      .then(props.close)
-      .catch((error) => {
+    handlePromise(
+      k8sPatch(resourceKind, resource, patch),
+      () => close(),
```
spadgett, line 23:
```suggestion
      close,
```

spadgett:
/hold

PR #6208, opened 2020-08-04
Bug 1870469: Check for podName before displaying Task logs

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4273

**Analysis / Root cause**: 
When checking logs for a failed PipelineRun that has not run any pods, a random pod log will be shown from any running pod.

**Solution Description**: 
Ensure that `podName` is not empty in the pipelinerun spec before trying to display logs.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelineruns/detail-page-tabs/PipelineRunLogs.tsx
+++ b/frontend/packages/dev-console/src/components/pipelineruns/detail-page-tabs/PipelineRunLogs.tsx
@@ -90,15 +90,17 @@ class PipelineRunLogs extends React.Component<PipelineRunLogsProps, PipelineRunL
             obj.metadata?.name,
           )
         : undefined;
-    const resources = taskCount > 0 && [
-      {
-        name: _.get(taskRunFromYaml[activeItem], ['status', 'podName'], ''),
-        kind: 'Pod',
-        namespace: obj.metadata.namespace,
-        prop: `obj`,
-        isList: false,
-      },
-    ];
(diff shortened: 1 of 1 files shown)


frontend/packages/dev-console/src/components/pipelineruns/detail-page-tabs/PipelineRunLogs.tsx
```diff
@@ -138,7 +138,7 @@ class PipelineRunLogs extends React.Component<PipelineRunLogsProps, PipelineRunL
           )}
         </div>
         <div className="odc-pipeline-run-logs__container">
-          {activeItem ? (
+          {activeItem && resources[0].name ? (
```
divyanshiGupta, line 141: Checking for the name here seems a bit odd, a better approach would be to create the resources only when a pod exists.

```
const podName = taskRunFromYaml[activeItem]?.status?.podName;
const resources = taskCount > 0 && podName && [
      {
        name: podName,
        kind: 'Pod',
        namespace: obj.metadata.namespace,
        prop: `obj`,
        isList: false,
      },
    ];
```
```suggestion
          {activeItem && resources ? (
```

rottencandy: Thanks! Checking during resource creation is definitely better. I've updated with your suggestion.

rohitkrai03 (approved):
/approve

PR #6210, opened 2020-08-04
Bug 1858317: Change type of VLAN attribute for CNV bridge in NADs

pcbailey:
Currently, the type of the VLAN attribute for network attachment definitions with a network type of CNV bridge is string, but it should be number. This PR makes that change.


spadgett:
/hold

PR #6212, opened 2020-08-04
Bug 1873143: fix pipeline details page breadcrumbs in admin perspective

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4220

**Analysis / Root cause**: 
 Pipeline details pages on Admin to have breadcrumbs that take the user to default list page of Pipelines instead of to the tabbed list page from where the user came from

**Solution Description**: 
Pipeline details pages on Admin to have breadcrumbs that take the user back to the tabbed list page from where the user came from.

Kapture 2020-08-05 at 3 26 28

### DIFF
--- a/frontend/packages/dev-console/src/components/cluster-tasks/ClusterTaskDetailsPage.tsx
+++ b/frontend/packages/dev-console/src/components/cluster-tasks/ClusterTaskDetailsPage.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import { DetailsPageProps, DetailsPage } from '@console/internal/components/factory';
+import { navFactory, Kebab } from '@console/internal/components/utils';
+import { DetailsForKind } from '@console/internal/components/default-resource';
+import { useTasksBreadcrumbsFor } from '../pipelines/hooks';
+
+const ClusterTaskDetailsPage: React.FC<DetailsPageProps> = (props) => {
+  const { kindObj, match, kind } = props;
+  const breadcrumbsFor = useTasksBreadcrumbsFor(kindObj, match);
+
+  return (
+    <DetailsPage
+      {...props}
+      menuActions={Kebab.factory.common}
+      breadcrumbsFor={() => breadcrumbsFor}
+      pages={[navFactory.details(DetailsForKind(kind)), navFactory.editYaml()]}
+    />
+  );
+};
+
+export default ClusterTaskDetailsPage;
--- a/frontend/packages/dev-console/src/components/pipelines/hooks.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/hooks.ts
@@ -1,0 +1,43 @@
+import { match as RMatch } from 'react-router-dom';
+// FIXME upgrading redux types is causing many errors at this time
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useSelector } from 'react-redux';
+import { getBreadcrumbPath } from '@console/internal/components/utils/breadcrumbs';
+import { K8sKind } from '@console/internal/module/k8s';
+import { RootState } from '@console/internal/redux';
+import { getActivePerspective, getActiveNamespace } from '@console/internal/reducers/ui';
+import { ALL_NAMESPACES_KEY } from '@console/shared';
+import { pipelinesTab } from '../../utils/pipeline-utils';
+
+type Match = RMatch<{ url: string }>;
+
+const useTabbedTableBreadcrumbsFor = (kindObj: K8sKind, match: Match, navOption: string) => {
+  const currentNamespace = useSelector((state: RootState) => getActiveNamespace(state));
+  const isAdminPerspective =
+    useSelector((state: RootState) => getActivePerspective(state)) === 'admin';
+  const nsURL =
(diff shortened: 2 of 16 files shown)


debsmita1:
you would need to fix the same for the Task details page as well

vikram-raj:
> you would need to fix the same for the Task details page as well

Thanks @debsmita1, I fix the same for tasks and triggers pages as well.

vikram-raj:
@andrewballantyne I added a hook to avoid code repeat. PTAL again.

vikram-raj:
/retitle Bug 1873143: fix pipeline details page breadcrumbs in admin perspective

frontend/packages/dev-console/src/components/cluster-tasks/ClusterTaskDetailsPage.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { DetailsPageProps, DetailsPage } from '@console/internal/components/factory';
+import { getActiveNamespace } from '@console/internal/reducers/ui';
+import { RootState } from '@console/internal/redux';
+import { navFactory, Kebab } from '@console/internal/components/utils';
+import { DetailsForKind } from '@console/internal/components/default-resource';
+import { usePipelinesBreadcrumbsFor } from '../pipelines/hooks';
+
+const ClusterTaskDetailsPage: React.FC<DetailsPageProps> = (props) => {
+  const { kindObj, match, kind } = props;
+  const activeNamespace = useSelector((state: RootState) => getActiveNamespace(state));
```
andrewballantyne, line 15: I think you can avoid doing this out here if you just bake this functionality into the hook.

This is a bit of annoying usage due to the 3 imports + ts-ignore override.

frontend/packages/dev-console/src/components/pipelines/hooks.ts
```diff
@@ -0,0 +1,31 @@
...
+  namespace: string,
+  navOption: string,
+) => {
+  const activePerspective = useSelector((state: RootState) => getActivePerspective(state));
+  return [
+    {
+      name: kindObj.labelPlural,
+      path:
+        activePerspective === 'admin'
+          ? `/${navOption}/ns/${namespace}${
```
andrewballantyne, line 24: You have a bug here if you're in "all-namespaces" when you go into the details page - leaving via the breadcrumb causes a not found issue with `#ALL_NS#`.

frontend/packages/dev-console/src/models/pipelines.ts
```diff
@@ -105,7 +105,7 @@ export const ConditionModel: K8sKind = {
   id: 'condition',
   labelPlural: 'Conditions',
   crd: true,
-  badge: BadgeType.DEV,
+  badge: BadgeType.TECH,
```
andrewballantyne, line 108: Thanks for aligning this.

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -403,3 +414,26 @@ export const getSecretAnnotations = (annotation: KeyValuePair) => {
   }
   return annotations;
 };
+
+export const pipelinesTab = (kindObj: K8sKind) => {
+  switch (kindObj.kind) {
```
andrewballantyne, line 419: I think for this to be an accurate usage, we need to cover the situation where they pass valid models that don't have unique values (the "root" pages).

```
    case PipelineModel.kind:
    case TaskModel.kind:
    case EventListenerModel.kind:
      return '';
```

That way there is a distinction between `null` (not a pipeline tab) and `''` a "root" pipeline tab. This would be my expectation when using a function called `pipelinesTab` to get a tab extension for the pipelines pages.

vikram-raj:
@andrewballantyne Nothing wrong with your suggestion. I was just trying to do it using one hook. Thanks.

andrewballantyne:
> @andrewballantyne Nothing wrong with your suggestion. I was just trying to do it using one hook. Thanks.

It's important to remember in the functional world, functions are not bad. If you can get a readable and easily understandable set of functions, which avoid magic exposed strings, you have less errors in usage and more understandable use-cases. 

Wrapping functions should not be frowned upon in the functional world. This is one of the main wins over the OO world.

andrewballantyne:
> i.e, something like this:

Oops, I created a bug with my implementation. Was testing around and I noticed `all-namespaces` is not `ns/all-namespaces`. Please correct that @vikram-raj 

There is another bug here but I think it's unrelated, and that's viewing this tab view in the Dev perspective, it breaks the "single project" methodology... as it has no "dev perspective" blocks to show projects. Not sure what the best solution is for that right now. I will raise it in the tmp-odc-pipelines channel.

vikram-raj:
> Oops, I created a bug with my implementation. Was testing around and I noticed all-namespaces is not ns/all-namespaces. Please correct that @vikram-raj

Fixed this issue.

andrewballantyne:
cc @divyanshiGupta FYI on this solution as we had discussed this several weeks back.

Vikram needs another reviewer, if @debsmita1 wants to finish her review or if you want to do it @divyanshiGupta

divyanshiGupta (approved):
Code looks good to me, also verified it locally as well and it works fine for me.
/lgtm

PR #6213, opened 2020-08-05
Adding lso plugin as a dependency for ceph-storage-plugin

gnehapk:


spadgett:
/hold

PR #6215, opened 2020-08-05
Unit test for IPV6 Validation

bipuladh:
Validates that the function used can correctly determine IP Family mismatch.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/independent-mode/utils.spec.ts
+++ b/frontend/packages/ceph-storage-plugin/src/components/independent-mode/utils.spec.ts
@@ -32,6 +32,18 @@ const JSON_CORRECT_MULTIPLE_KEYNAME_VARIATION = [
     name: 'rook-csi-cephfs-provisioner',
   },
 ];
+const JSON_IPV6_VARIATION = [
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/components/independent-mode/utils.spec.ts
```diff
@@ -32,6 +32,18 @@ const JSON_CORRECT_MULTIPLE_KEYNAME_VARIATION = [
...
+    kind: 'ConfigMap',
+    data: { maxMonId: '0', data: 'a=10:106:31:93:6789', mapping: {} },
+    name: 'rook-ceph-mon-endpoints',
+  },
+  {
+    kind: 'Secret',
+    data: { userKey: 'AQBV66pedfkUIhAAn/tnB0cvIih5n9DwwxOwBg==', userID: 'csi-cephfs-node' },
+    name: 'rook-csi-cephfs-node',
+  },
+];
```
afreen23, line 46: I am finding it same as `JSON_CORRECT` . Am I missing something ?

spadgett:
/hold

PR #6216, opened 2020-08-05
Bug 1866084: AWS EBS parameters for EBS CSI storageclass

cloudbehl:
Made parameters & options available looking at https://github.com/kubernetes-sigs/aws-ebs-csi-driver#createvolume-parameters doc.

588DD614-C5A8-4ED7-9F66-81DC2427793D

### DIFF
--- a/frontend/public/components/storage-class-form.tsx
+++ b/frontend/public/components/storage-class-form.tsx
@@ -103,6 +103,44 @@ export class StorageClassForm_ extends React.Component<
   // For 'csi' storage type
   CSIStorageTypes = Object.freeze({
     ...this.getExtensionsStorageClassProvisioners(Provisioner.CSI),
+    'ebs.csi.aws.com': {
+      title: 'AWS CSI',
+      provisioner: 'ebs.csi.aws.com',
+      allowVolumeExpansion: true,
+      parameters: {
+        type: {
+          name: 'Type',
+          values: { gp2: 'gp2', io1: 'io1', sc1: 'sc1', st1: 'st1', standard: 'standard' },
+          hintText: 'Select AWS Type. Default is gp2',
+        },
+        iopsPerGB: {
+          name: 'IOPS Per GiB',
+          hintText: 'I/O operations per second per GiB',
+          validation: (params) => {
+            if (params.iopsPerGB.value && !params.iopsPerGB.value.match(/^\d+$/)) {
+              return 'IOPS per GiB must be a number';
+            }
+            return null;
+          },
+          visible: (params) => _.get(params, 'type.value') === 'io1',
+        },
+        fsType: {
+          name: 'Filesystem Type',
+          hintText: 'Filesystem type to use during volume creation. Default is ext4.',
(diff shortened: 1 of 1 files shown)


frontend/public/components/storage-class-form.tsx
```diff
@@ -103,6 +103,44 @@ export class StorageClassForm_ extends React.Component<
...
+        type: {
+          name: 'Type',
+          values: { io1: 'io1', gp2: 'gp2', sc1: 'sc1', st1: 'st1', standard: 'standard' },
+          hintText: 'Select AWS Type. Default is gp2',
+        },
+        iopsPerGB: {
+          name: 'IOPS Per GiB',
+          hintText: 'I/O operations per second per GiB',
+          validation: (params) => {
+            if (params.iopsPerGB.value !== '' && !params.iopsPerGB.value.match(/^[1-9]\d*$/)) {
```
spadgett, line 120: Isn't this the same as `^\d+$`? Are fractional values allowed? (If so, the regex is probably wrong.)

```suggestion
            if (params.iopsPerGB.value && !params.iopsPerGB.value.match(/^\d+$/)) {
```

spadgett: `Number.isFinite(Number(params.iopsPerGB.value))` might be better. Although it looks like you've copied what's below.

frontend/public/components/storage-class-form.tsx
```diff
@@ -103,6 +103,44 @@ export class StorageClassForm_ extends React.Component<
...
+              return 'IOPS per GiB must be a number';
+            }
+            return null;
+          },
+          visible: (params) => _.get(params, 'type.value') === 'io1',
+        },
+        fsType: {
+          name: 'Filesystem Type',
+          hintText: 'Select Filesystem to Be Laid Out. Default is ext4',
```
spadgett, line 129: This is odd case, but I see it's the same as below. Should it be "to be laid out"?

cloudbehl: How about going with this?
`Selected File system type that will be formatted during volume creation. Default Type is ext4`

frontend/public/components/storage-class-form.tsx
cloudbehl: `(/^\d+$/)` && `(/^\d+$/)` is same and fractional values are not allowed looking at code[1].

[1] . https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/44ec7de2c953b52c9fd39e85ca2183aab6465d32/pkg/driver/controller.go#L128

`Number.isFinite(Number(params.iopsPerGB.value))` should I change it at both places?

spadgett: `/^\d+$/` is good if we don't allow fractions.

frontend/public/components/storage-class-form.tsx
spadgett: Maybe "Filesystem type to use during volume creation. Default is ext4." Punctuation on both sentences. Filesystem is one word.

PR #6217, opened 2020-08-05
Bug 1856645: Query Browser: Fix poller time range

kyoto:
`now` should have been set immediately before each poller tick. This bug
caused the query range start and end times to be off by an amount
approximately equal to current refresh interval.


spadgett:
/hold

PR #6218, opened 2020-08-05
Change image stream dropdown labels

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-2425

**Analysis / Root cause**: 
There are a couple things that should be improved since this is a static field and not something dynamically created.

* The title `ImageStreams` should have a space
* The disabled dropdown shouldn't have an ellipsis to a predefined label ("Select Image Strea...")

**Solution Description**: 
Changed the text of "Projects" to "Project" and ImageStreams to "Image Stream". I used singular because the user can select only one Project here. This matches also some other dropdown field labels, like "Application Name".
* The text "Select Image Stream" was ellipsis on my Chrome and Firefox (both Fedora 32) only when the windows is smaller than 1024 pixels. I think this is "acceptable", or? @andrewballantyne 

@openshift/team-devconsole-ux 

Before:
before

After:
after

No tests changed

**Test setup:**
* Create a Deployment from a container image with an external image (for example jerolimov/nodeinfo)
* Open the Import from container image page again and select "Image stream tag from internal registry"

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/import/image-search/ImageStreamDropdown.tsx
+++ b/frontend/packages/dev-console/src/components/import/image-search/ImageStreamDropdown.tsx
@@ -80,7 +80,7 @@ const ImageStreamDropdown: React.FC = () => {
   return (
     <ResourceDropdownField
       name="imageStream.image"
-      label="ImageStreams"
+      label="Image Stream"
       resources={getImageStreamResource(imageStream.namespace)}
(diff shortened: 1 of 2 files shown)


rottencandy:
/lgtm
Tested locally, looks good

PR #6219, opened 2020-08-05
Bug 1870670: Improve spacing for RadioGroupFields and the image search

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-2411

**Analysis / Root cause**: 
These radio buttons on the Import from container image page are not spaced appropriately.

**Solution Description**: 
Improve the general RadioGroup spacing. See screenshots below.

@openshift/team-devconsole-ux There is also an video available which shows to difference:
https://issues.redhat.com/secure/attachment/12482887/Screencast.mp4

Import from container image, image area...
Before:
external-image-empty-before

After:
external-image-empty-after

When selecting "internal image registry"
Before:
internal-registry-no-error-before

After:
internal-registry-no-error-after

:warning: This changes also the Resource section!
Before:
resource-select-before

After:
resource-select-after

The change in the Resource section matches now all other areas. Checkout the movie I uploaded to the issue!
https://issues.redhat.com/secure/attachment/12482887/Screencast.mp4

Not touched

**Test setup:**
* Open the Import from container image screen and select the both options "from external image registry" and "from internal image registry". Test this also when there is no images available in the internal image registry which shows an additional alert then.
* Verify / Compare also the "Resource" section on the Import from container image page.
* The RadioGroupField was also used with inline prop in the Helm chart. To check that this behavior is not changed, go to the developer perspective, select "Add" > "Helm Chart" > Select for example "Nodejs Ex K v0.2.0" > "Install Helm Chart". The configuration switch for "Form View" and "YAML View" should looks like before.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/RadioGroupField.scss
+++ b/frontend/packages/console-shared/src/components/formik-fields/RadioGroupField.scss
@@ -5,8 +5,16 @@
     margin-top: 0;
   }
 
-  &__children {
-    padding: var(--pf-global--spacer--md) 0;
+  &:not(.ocs-radio-group-field--inline) {
+    .pf-c-radio {
+      padding-bottom: var(--pf-global--spacer--md);
+    }
+    .pf-c-radio:last-of-type {
+      padding-bottom: 0;
+    }
+    .ocs-radio-group-field__children {
+      padding-top: var(--pf-global--spacer--sm);
+    }
   }
 
   &--inline {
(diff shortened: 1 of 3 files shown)


andrewballantyne:
Need bugzilla now.

/retitle Bug 1870670: Improve spacing for RadioGroupFields and the image search

PR #6220, opened 2020-08-05
Bug 1846894: remove namespace if resource is not namespaced and namespace is provided

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1846894

### DIFF
--- a/frontend/public/components/edit-yaml.jsx
+++ b/frontend/public/components/edit-yaml.jsx
@@ -324,6 +324,10 @@ export const EditYAML_ = connect(stateToProps)(
           return;
         }
 
+        if (obj.metadata.namespace && !model.namespaced) {
+          delete obj.metadata.namespace;
+        }
+
         // If this is a namespaced resource, default to the active namespace when none is specified in the YAML.
         if (!obj.metadata.namespace && model.namespaced) {
           if (this.props.activeNamespace === ALL_NAMESPACES_KEY) {


frontend/public/components/edit-yaml.jsx
```diff
@@ -324,6 +324,12 @@ export const EditYAML_ = connect(stateToProps)(
           return;
         }
 
+        // If this is NOT a namespaced resource and namespace is provided, display an error.
+        if (obj.metadata.namespace && !model.namespaced) {
+          this.handleError('"metadata.namespace" field found in YAML is not necessary.');
```
rhamilto, line 329: Wondering if we remove the `namespace` key instead?  The CLI appears to do that.

spadgett, line 329: I might make the message a little stronger since it's an error, something like

```
${obj.kind} is a cluster-scoped resource and cannot have a "metadata.namespace" field.

spadgett: > Wondering if we remove the `namespace` key instead? The CLI appears to do that.

Yeah, I had the same thought. I think that's a reasonable fix since it's consistent with the CLI. The API server also ignores unknown fields.

rhamilto:
Updated to remove `namespace` if not a namespaced resource.  Thanks, @spadgett.

spadgett:
/hold
for merge queue fix #6225

PR #6222, opened 2020-08-05
Bug 1862537: Change pending changes alert when making a change

irosenzw:
While the VM is running and the user make a change
in the Flavor or the boot order modals, the alert text is changing in the modal
as well as the alert variant.


spadgett:
/hold

PR #6226, opened 2020-08-05
Bug 1865930: Fix runtime error on project details page

TheRealJon:
In area chart component, filter empty sets from the `data` prop before using it to calculate legend data or render a chart.

### DIFF
--- a/frontend/public/components/graphs/area.tsx
+++ b/frontend/public/components/graphs/area.tsx
@@ -62,13 +62,12 @@ export const AreaChart: React.FC<AreaChartProps> = ({
   // @ts-ignore
   const theme = getCustomTheme(ChartThemeColor.blue, ChartThemeVariant.light, areaTheme);
   const [containerRef, width] = useRefWidth();
-
-  const [processedData, unit] = React.useMemo(() => {
+  const { processedData, unit } = React.useMemo(() => {
+    const nonEmptyDataSets = data.filter((dataSet) => dataSet?.length);
     if (byteDataType) {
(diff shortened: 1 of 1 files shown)


frontend/public/components/graphs/area.tsx
```diff
@@ -91,9 +91,9 @@ export const AreaChart: React.FC<AreaChartProps> = ({
   const container = React.useMemo(() => {
     if (multiLine) {
       const legendData = processedData.map((d) => ({
-        childName: d[0].description,
-        name: d[0].description,
-        symbol: d[0].symbol,
+        childName: d[0]?.description,
+        name: d[0]?.description,
+        symbol: d[0]?.symbol,
```
spadgett, line 96: Do we know if it's safe to pass `undefined` for these values? How does it impact the chart when this happens?

TheRealJon: I followed this down through the ChartLegendTooltip component and it looks like it should be okay if these are undefined. I'm not 100% sure what effect this would have on the chart since in this case `d[0] === undefined` means there is an empty data set. Let me dig a little bit more.

TheRealJon:
/hold

Testing implications of this

TheRealJon:
/hold cancel

@spadgett Looked a little more deeply at this and I think solving the 'empty data' set issue before even getting to the legend logic makes more sense. I added logic to filter the `data` prop down to only data sets that have data points and removed the optional chaining from my previous solution.

spadgett:
/hold
for queue fix #6225

PR #6227, opened 2020-08-05
Bug 1866114: Don't store OpenAPI definitions in localStorage

spadgett:
The document is too large, so we hit browser limits for localStorage.
This can break Safari. Since the API server supports ETags for the
OpenAPI document, rely on browser caching instead.

/assign @jhadvig @andrewballantyne

### DIFF
--- a/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
+++ b/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
@@ -5,7 +5,7 @@ import {
 } from 'monaco-languageclient/lib/monaco-converter';
 import { getLanguageService, TextDocument } from 'yaml-language-server';
 import { openAPItoJSONSchema } from '@console/internal/module/k8s/openapi-to-json-schema';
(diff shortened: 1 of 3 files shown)


andrewballantyne:
@spadgett I'm still not getting a successful render in Safari, I recall there being another issue here that caused rendering but I am unable to find the ticket (hopefully I'm not dreaming of tickets 😛 ).

I get some React warnings and websocket errors (closed before it established) but I don't think they would prevent rendering.

spadgett:
> @spadgett I'm still not getting a successful render in Safari, I recall there being another issue here that caused rendering issues but I am unable to find the ticket (hopefully I'm not dreaming of tickets 😛 ).

Yeah, it's a separate known issue related to the notification drawer. I think it's the same underlying problem as the drawer not opening, which has a bug. We should confirm that the PF update fixes it, though.

cc @rhamilto @jcaianirh @dtaylor113

TheRealJon:
Also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1856354

PR #6229, opened 2020-08-05
Hide topology display options that are not effective for the current view

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4451

**Description**
Removes the options for show labels, show pod count from the display options when in the topology list view. Also remove the show <object type> options from the display options as these should now be filtered via the 'Filter by Resource' options.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


invincibleJai:
Verified locally works as described in issue

/lgtm

PR #6230, opened 2020-08-05
Add fix to prevent application reset

divyanshiGupta:
Jira issue - https://issues.redhat.com/browse/ODC-3992

**Problem:**
In deploy-image form in case of external-registry the application is getting reset even when a component is being created in context of an existing application using application context menu.

**Solution:**
Check if the component being created is in context of some application, if yes do not reset the application.

**Gif:**
Peek 2020-08-05 23-00

/kind bug


karthikjeeyar:
Verified locally, it creates the application in  context.
/lgtm

PR #6232, opened 2020-08-05
Fix crash when knative deployment is selected in topology

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4453

**Description**
Fixes an issue where the topology view crashes upon selection of a knative deployment. This occurs when a knative revision's deployment resource is selected while in 'flat' mode.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


invincibleJai:
/lgtm

verified locally seems to work as expected

PR #6238, opened 2020-08-05
Bug 1866090: Fix selector spec descriptor link behavior

TheRealJon:
Correctly format `group~version~kind` string for spec descriptor selector links.


spadgett:
/lgtm cancel

Test failure:

```
FAIL packages/operator-lifecycle-manager/src/components/descriptors/spec/index.spec.tsx
  ● Spec descriptors › renders a basic selector

    expect(received).toEqual(expected)
    
    Expected value to equal:
      "core:v1:Service"
    Received:
      "core~v1~Service"
      
      at Object.it (packages/operator-lifecycle-manager/src/components/descriptors/spec/index.spec.tsx:168:49)
          at new Promise (<anonymous>)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #6239, opened 2020-08-05
Bug 1866033: Use Monaco APIs to set YAML editor value

spadgett:
Previously we were tracking the value in state, but this isn't
necessary. We read the current value from the Monaco edit API already,
and it caused a side effect of selecting the entire document on changes.

/assign @rebeccaalpert @rhamilto


rhamilto:
lgtm, but will wait for @rebeccaalpert to give final approval.

spadgett:
I missed a reference to `this.state.yaml`, updated.

PR #6248, opened 2020-08-06
Bug 1872336: Use metadata.generatorName when create a pipeline trigger

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4236

**Analysis / Root cause**: 
When adding a new Trigger for a Pipeline the web console creates an `EventListener` and a `TriggerTemplate`. The trigger templates contains a spec of a PipelineRun configuration (incl. parameters and resources) which will be created when an event is received. The current PipelineRun spec defines a fix name which could be created only once. 

**Solution Description**: 
To create multiple PipelineRun objects it is required to the a metaname `generateName` instead of `name` in the TriggerTemplate spec.

Some background infos I received from Andrew:
* The TriggerTemplate is created in [packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts#L14-L29)
* [TriggerTemplate documentation](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md) contains infos about the `generateName` attribute

YAML diff:

```diff
apiVersion: triggers.tekton.dev/v1alpha1
kind: TriggerTemplate
metadata:
  name: trigger-template-docker-kvqrbt
spec:
  resourcetemplates:
    - apiVersion: tekton.dev/v1beta1
      kind: PipelineRun
      metadata:
-       name: docker-3p01ab
+       generateName: docker-
...
```

Nothing changed in the UI

But here you see the relevant part of the test setup below:
odc-4236

Added some unit tests, no big change here.

Before:
```
=============================== Coverage summary ===============================
Statements   : 27.99% ( 20197/72165 )
Branches     : 17.94% ( 7515/41880 )
Functions    : 16.2% ( 2628/16222 )
Lines        : 29.82% ( 19255/64575 )
================================================================================
```

After:
```
=============================== Coverage summary ===============================
Statements   : 28% ( 20203/72165 )
Branches     : 17.96% ( 7523/41886 )
Functions    : 16.22% ( 2631/16222 )
Lines        : 29.83% ( 19261/64575 )
================================================================================
```

**Test setup:**
* Install the OpenShift Pipeline operator
* Create a Pipeline, for example Import from Git and select "Add Pipeline"
* Open the pipeline and press the menu button in the top right corner and select "Add Trigger"
* Fill the form and validate the generator trigger template
* Please test also Start Pipeline, Re-run pipeline
* And finally you can test the event listener with postman or curl. You find a GitHub POST JSON in the ticket. Ensure that the copied JSON does not contain JSON errors! You can just remove the lines with JSON errors!

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts
@@ -2,7 +2,7 @@ import { getRandomChars } from '@console/shared';
 import { apiVersionForModel, RouteKind } from '@console/internal/module/k8s';
 import { RouteModel } from '@console/internal/models';
 import { EventListenerModel, TriggerTemplateModel } from '../../../../models';
-import { PipelineRun } from '../../../../utils/pipeline-augment';
+import { Pipeline, PipelineRun } from '../../../../utils/pipeline-augment';
 import { PIPELINE_SERVICE_ACCOUNT } from '../../const';
 import {
   TriggerBindingKind,
@@ -12,6 +12,7 @@ import {
 } from '../../resource-types';
 
 export const createTriggerTemplate = (
+  pipeline: Pipeline,
   pipelineRun: PipelineRun,
   params: TriggerTemplateKindParam[],
 ): TriggerTemplateKind => {
@@ -19,7 +20,7 @@ export const createTriggerTemplate = (
     apiVersion: apiVersionForModel(TriggerTemplateModel),
     kind: TriggerTemplateModel.kind,
     metadata: {
-      name: `trigger-template-${pipelineRun.metadata.name}`,
+      name: `trigger-template-${pipeline.metadata.name}-${getRandomChars()}`,
     },
     spec: {
       params,
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/__tests__/utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/__tests__/utils.spec.ts
@@ -1,6 +1,6 @@
-import * as _ from 'lodash';
 import { PipelineRun, Pipeline } from '../../../../../utils/pipeline-augment';
-import { getPipelineRunData, migratePipelineRun } from '../utils';
+import { getPipelineRunData, getPipelineRunFromForm, migratePipelineRun } from '../utils';
+import { CommonPipelineModalFormikValues } from '../types';
 
 export const actionPipelines: Pipeline[] = [
   {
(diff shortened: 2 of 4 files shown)


frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
```diff
@@ -12,7 +12,6 @@ export type PipelineModalFormResource = {
 };
 
 export type CommonPipelineModalFormikValues = FormikValues & {
-  namespace: string;
```
jerolimov, line 15: @andrewballantyne I did not find a usage or formik field for this field, so I removed it. Missed I something here?!

jerolimov:
I do not find a "really short way" to add this case also to to this method which was used before to create PipelineRuns based on Pipeline OR PipelineRuns.
https://github.com/openshift/console/blob/2aee537dd54cb18e1108996a991b1b0f684e0bf4/frontend/packages/dev-console/src/components/pipelines/modals/common/utils.ts#L42-L83
Plus
https://github.com/openshift/console/blob/2aee537dd54cb18e1108996a991b1b0f684e0bf4/frontend/packages/dev-console/src/components/pipelines/modals/common/utils.ts#L136-L157

So I splitted this into this seperated methods and extracted some helper:
https://github.com/openshift/console/blob/66d6922bcbeaa2998089f6f9e3edb5079652076a/frontend/packages/dev-console/src/components/pipelines/modals/common/utils.ts#L117-L173

Its a little bit opinionated that this is easier and type-stricter. I hope it finds some support. :smirk:

frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
andrewballantyne: Yup still [being used](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/pipelines/modals/common/PipelineResourceDropdownField.tsx#L18). Please undo.

jerolimov:
> Looks like you did a bit of a refactor to solve a long standing issue ([ODC-2184](https://issues.redhat.com/browse/ODC-2184))... unfortunately this is not going to be acceptable during the stability sprints. We should look to trim this down to just managing a use of `generateName` for the trigger modal.
> 
> If you want, you can copy this code out and open up another PR we can get into master when it switches to 4.6 - but that's likely not until Oct 2nd.

@andrewballantyne I removed the refactoring and monitor ODC-2184 so that we can maybe apply these changes later. The original changes are available here: https://github.com/openshift/console/compare/master...jerolimov:odc-4236-with-refactoring

I keep the extended unit tests. I'm open for any feedback. :smile:

frontend/packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts
```diff
@@ -12,14 +12,15 @@ import {
 } from '../../resource-types';
 
 export const createTriggerTemplate = (
+  pipeline: Pipeline,
   pipelineRun: PipelineRun,
   params: TriggerTemplateKindParam[],
 ): TriggerTemplateKind => {
   return {
```
andrewballantyne, line 19: I wonder if we could just override it for this one use-case rather than baking it into the (more or less) broken flow that exists today getting a Pipeline Run.

I think the refactor portion of your PR hit the nose of the problem, and that definitely is the better choice once we can refactor the code. Just not sure we need to pollute "crafting a pipeline run" if we are going to rewrite it in a few anyways.

I just suggest the modification here: 
```
  const resource = omit(pipelineRun, 'metadata.name');
  resource.metadata.generateName = `${pipeline.metadata.name}-`;
```

jerolimov: @andrewballantyne I though multiple times (and longer than I should) about this.. I would be happy if we stay with this (reduced) change. Keep the additional options parameter allow us to extend the unit tests for this case. Keep the additional tests instead of just "override" the generated json here is worth enough to touch 3 files.

I'm on your side that its still not perfect, but for me it looks like the right direction. :thinking: 

I'm also happy if we want split this method into 3 when starting with 4.7.

andrewballantyne (approved):
Going with Christoph's suggestion to keep this change.

Imo, a full refactor in 4.7 is much better than a partial change here, but I left it up to him to make the call. Let's hope this fragile piece of the code doesn't come back to haunt us.

rottencandy (approved):
Tested locally
/lgtm

PR #6252, opened 2020-08-06
Bug 1859258: Pf package update with fix for nested drawers

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1859258

Ran a build and test for this for nested drawers successfully. 

Also updated notification badge as follows:

- new design with no grey square border

- force isread to always equal true

- provide alert count to badge to show total alerts next to badge

Screen Shot 2020-08-10 at 9 38 16 PM
Screen Shot 2020-08-10 at 9 39 15 PM


jcaianirh:
/hold
there is one bug the pf team is working which will bump the pf core repo

dtaylor113:
Hi @jcaianirh & @jschuler, is the Notification icon supposed to have a blue background?
image

dtaylor113:
Verified Notification Drawer does open on small/medium screens.

jcaianirh:
> Hi @jcaianirh & @jschuler, is the Notification icon supposed to have a blue background?
> 
> image
> 
> 

Yes @dtaylor11 this is the new unread design for the notification badge.

spadgett:
Hey @jcaianirh. I have concerns about that. In OpenShift, you're pretty much always going to have a notification on most clusters. We blue background seems distracting and way too strong. I strongly prefer the subtle dot indicator. What is the problem the blue background is intending to fix?

cc @beanh66

jschuler:
Hi, is the logic surrounding the alert notifications based on number of alerts? e.g. show a visual indicator if there are 1 or more alerts? Is there any way to acknowledge, mark as read, or silence these alerts?
I believe the original intent behind the notification badge is to be able to show whether alerts have been read or unread, and if there are any unread alerts it is assumed that you can change the state.

dtaylor113:
> Hi, is the logic surrounding the alert notifications based on number of alerts? e.g. show a visual indicator if there are 1 or more alerts? Is there any way to acknowledge, mark as read, or silence these alerts?

Hi, I see `NotificationBadge` being set [here](https://github.com/openshift/console/blob/5acc7626602021093282d77cc559d7fef425d565/frontend/public/components/masthead-toolbar.jsx#L565) which gets its `isRead/notificationsRead` value from Redux [here](https://github.com/openshift/console/blob/5acc7626602021093282d77cc559d7fef425d565/frontend/public/components/masthead-toolbar.jsx#L639). 
Still investigating how `isRead` is set.

jschuler:
We are still discussing the ramifications of the changes within the team.
Possibly as an alternative or workaround: Are you able to get a count? If so, you could also choose to display the 'read' variant with a count. Some examples here: https://codesandbox.io/s/pedantic-sky-w51mq?file=/index.js

spadgett:
Hey @jschuler, those options all seem pretty strong to me.

One challenge we have with the count is handling different alert levels. If we have a red background with 30, does that mean we have 30 critical alerts or could it be 1 critical alert and 29 normal alerts?

We don't currently have a way to mark the alerts as read. Part of the problem with marking read is how to handle an alert that is resolved then comes back later. We want it to surface again.

Maybe my opinion is being skewed by always having the `AlertmanagerReceiversNotConfigured` alert firing on every development cluster. There are definitely some alerts you shouldn't ignore.

Adding @cshinn for feedback.

dtaylor113:
> We don't currently have a way to mark the alerts as read. Part of the problem with marking read is how to handle an alert that is resolved then comes back later. We want it to surface again.

@spadgett  When one [clicks](https://github.com/openshift/console/blob/5acc7626602021093282d77cc559d7fef425d565/frontend/public/components/masthead-toolbar.jsx#L564) on the notification badge it toggles the [expanded/collapse](https://github.com/openshift/console/blob/016065ff398d8e2f3892c341a4d9efbe2b4b7fc7/frontend/public/reducers/ui.ts#L257) of the drawer.  @spadgett & @jcaianirh we could also have clicking on the notification badge toggle the [read/unread](https://github.com/openshift/console/blob/016065ff398d8e2f3892c341a4d9efbe2b4b7fc7/frontend/public/reducers/ui.ts#L263) of the notifications to acknowledge that the user opened the drawer and 'read' the notifications.

spadgett:
@dtaylor113 Do we have enough data in the alert to make sure we're not permanently muting an alert that can come back days or months later? I'm also concerned with accumulation of cruft over time if we use `localStorage` for tracking read alerts. If we track in-memory instead, it will be annoying to open a new tab or come back tomorrow and have the same alert be marked unread. We might have to give this some careful thought. I'm a little uncomfortable feeling like we need to add a feature after feature freeze to fix a basic problem with the drawer.

dtaylor113:
> @dtaylor113 Do we have enough data in the alert to make sure we're not permanently muting an alert that can come back days or months later?

Hi @spadgett, sorry for the churn, taking time to chase down the redux actions & reducers.   I believe [this logic](https://github.com/openshift/console/blob/736fceaa41de71ce32b1d85e6302422b1177bc30/frontend/public/components/notification-drawer.tsx#L351) is supposed to toggle the read/unread of notifications.  Debugging through it, it doesn't toggle though.

spadgett:
So we're tracking in-memory in Redux? Will it be annoying that it won't remember across browser tabs or sessions?

mceledonia:
@spadgett Just jumping in here - 

The attention/critical state is to signify that one or more critical alerts have been fired, so you could indeed have 1 critical and 29 noncritical. This was intentional, as it is intended to exist as a sign to say "you really should check the notification drawer right now" regardless of if there are one or multiple critical alerts. Whereas the unread state (all non-critical) is more of "there are new alerts firing, but nothing is on fire" indication.

If you're always going to have alerts firing, perhaps it makes sense to just drop the blue/non-critical unread state and leave it with default + critical. It might make sense to then bring in a count so the user is aware of how many alerts are firing at a given time to give some indication of change over time, unless there really is no need for awareness until a critical alert fires. 

The attention state is also built in such a way that should allow admins to specify exactly which alerts trigger the state, rather than always assuming all critical alerts are worthy of the state. Though, that is how it works out of the box as we figured that was a reasonable default. 

As for read/unread, I think of it more as opening the drawer resets the state until a new critical alert fires. This is making an assumption that the user now has awareness of the firing alert, but without capabilities to detect actual reading or some sort of confirmation/dismissal of the firing alert, we thought it was a good next option. I think this combined with the ability to silence alerts in the future should also help provide more flexibility to the user in terms of how they handle their experience.

CC @cshinn

dtaylor113:
> So we're tracking in-memory in Redux? Will it be annoying that it won't remember across browser tabs or sessions?

"annoying", possibly, I think the bigger issue is that we really don't have a read/acknowledge alert/notification strategy.  At the moment it seems to be "if you have alerts, they are unread, unless you silence all of them".

dtaylor113:
Hi, just verified if I Silence all alerts, they notification badge does become 'read':
image

This means that if a new alert comes in, notification badge will again become 'unread'.

spadgett:
That looks a lot like a focus border around the bell icon.

dtaylor113:
> That looks a lot like a focus border around the bell icon.

Yes, that is persistent, not going away if I reload page or click elsewhere on the page.

mceledonia:
> That looks a lot like a focus border around the bell icon.

> ...At the moment it seems to be "if you have alerts, they are unread, unless you silence all of them".

Given this feedback I would suggest a couple design changes. We can help out with those if needed, but I think these should solve those issues:
1. Eliminate "unread/noncritical" blue state, so only state change is from default to critical
1a. Potentially consider adding "firing alert count" number to give some awareness of firing alerts over time if desired. 
2. Eliminate border around the icon in default state. Given that this is being read as a focus state we may remove this from the component itself as well, it also does not fit in with the styling of the rest of the icons. I think the hover/focus states are fine to leave as-is, just removing the border to match the other icons should be good.

dtaylor113:
> As for read/unread, I think of it more as opening the drawer resets the state until a new critical alert fires. This is making an assumption that the user now has awareness of the firing alert, but without capabilities to detect actual reading or some sort of confirmation/dismissal of the firing alert, we thought it was a good next option.

@mceledonia,  in OpenShift we a way to tag/compare 'these are the last read [critical?] alerts when you last opened the drawer' .vs 'there are now new [critical?] alerts that weren't there the last time you opened the drawer'  -if that is the direction we are going in.

mceledonia:
@dtaylor113 I see, in terms of how it worked beforehand was it the same but [all] alerts rather than filtered for [critical]?

dtaylor113:
Our current implementation doesn't seem correct:
image

Seems like we should have the red background for critical alerts, ie:  
image

Either way, I think we need to figure out a short term .vs longer term solution.

dtaylor113:
> @dtaylor113 I see, in terms of how it worked beforehand was it the same but [all] alerts rather than filtered for [critical]?

@mceledonia, currently it is 'all' alerts for the notification badge, we don't differentiate between critical and non-critical, but looking at the code we do have the information so it might be a straight forward change, but maybe too much to do after our feature freeze date.

mceledonia:
Gotcha, thanks for showing that.

I think for short term... Given that there's not an existing capability to filter between critical and non-critical, in addition to the fact that clusters likely always have alerts firing, I would question what the real value of a "firing alerts" state is. Maybe just switching to the default-only state (removing all unread state stylings) along with a count is a better short term solution? The count would at least provide some insight into how many alerts are currently firing rather than just being a static button with no information.

For long term, assuming we could add in the capability to distinguish between critical/non-critical, I would stick with the suggestions I made above. 

Thoughts? @dtaylor113 @spadgett @cshinn

dtaylor113:
> Maybe just switching to the default-only state (removing all unread state stylings) along with a count is a better short term solution?

Hi @mceledonia, I agree.   Passing in a `isRead={true}` for the `<NotificationBadge/>` is easy, only remaining  issue is the focus border discussed above.   Unfortunately, we don't have the count at the mastheadtoolbar level, but we are storing them in Redux, so we could pull them.  Looking into this now....

mceledonia:
> ... only remaining issue is the focus border discussed above.

Sounds good, I checked with Michael Coker on that and he offered to help with the CSS there if needed!

dtaylor113:
Hi @spadgett, @mceledonia,  @jcaianirh, 
Not too difficult to count all the `notificationAlerts` from redux, here is what it looks like:
image
I think it's more difficult to 'count by severity` as that is done via notification/alerting code below the masthead code.  -thoughts?

cshinn:
Part of the issue that this was intended to solve is that I've gotten repeated reports that the current icon isn't salient enough when there are things present in the drawer. I think that including the count is certainly helpful if it's easy enough and we have to space to do so. It does seem to me that we're doing something wrong if we _always_ have some alerts firing. What's the point of them then?

spadgett:
> It does seem to me that we're doing something wrong if we _always_ have some alerts firing. What's the point of them then?

That's fair, and my experience might be biased by always connecting to clusters used for development, which usually seem to have something going on.

Even if there are no firing alerts, I suspect the update / channel available notification will often be present. Not everyone wants to update immediately.

cshinn:
To be honest, I wasn't expecting this icon change to come as early as 4.6. I have several designs for filters that would let people choose what they want to see in the drawer. With that kind of functionality, I think a more salient icon probably makes more sense. In the mean time however, perhaps it's better to go with something less in-your-face.

spadgett:
> To be honest, I wasn't expecting this icon change to come as early as 4.6.

Yeah, we're more or less required to take the indicator change to pick up an unrelated PF bug fix.

dtaylor113:
Hi @jcaianirh, created a PR for this PR to add the count to notifications badge: https://github.com/jcaianirh/console/pull/6  
PTAL, thanks :-)

spadgett:
@jcaianirh looks like this is causing unit test failures :/

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -600,12 +601,13 @@ class MastheadToolbarContents_ extends React.Component {
           </PageHeaderToolsGroup>
           <PageHeaderToolsGroup>
             {/* mobile -- (notification drawer button) */
-            alertAccess && !notificationsRead && (
+            alertAccess && notificationAlerts?.data?.length > 0 && (
               <PageHeaderToolsItem className="visible-xs-block">
```
dtaylor113, line 605: Hi @jcaianirh, does this mean if there are no notifications the badge won't show up in the masthead?  Is that what we want? -thanks

dtaylor113: nvm, I see this is for mobile.  :+1:

christianvogt:
@jcaianirh can this be addressed or is it (or something like it) still required?
https://github.com/openshift/console/blob/b115ab3f91decd113431da43aa8491ea92b99fd4/frontend/public/style/_overrides.scss#L355-L356

jcaianirh:
> https://github.com/patternfly/patternfly/pull/3268

I believe this can be addressed. removing this and testing.

spadgett:
Hey @jcaianirh, can you verify this fixes the problem in Safari as well where the main content area is white?

Do you have a link to the PF bug fix we're waiting on?

jcaianirh:
@spadgett openshift does appear in safari with this update. see image below

Screen Shot 2020-08-17 at 1 21 08 PM

PR #6255, opened 2020-08-06
Bug 1868644: Add resource badges in the add trigger form dropdown

karthikjeeyar:
**Fixes**: https://issues.redhat.com/browse/ODC-4284

**Problem**:
In Add trigger form, Git provider type dropdown lists TriggerBindings and ClusterTriggerBindings without resource badges which makes it difficult for the user to distinguish

**Solution**:
Added the resource badges in the dropdown and autocomplete filter to be consistent with other resource dropdown in the Devconsole

**Screenshots**:
TriggerBinding and ClusterTriggerBinding with same name
image

trigger-dropdown
Test coverage updated:

image

cc: @andrewballantyne @siamaksade @openshift/team-devconsole-ux


frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -17,6 +17,7 @@ import {
 
 type DropdownItemProps = {
   model: K8sKind;
+  resource: K8sResourceKind;
```
abhinandan13jan, line 20: should this be optional?

karthikjeeyar: resource is new field which is always passed from this component https://github.com/openshift/console/blob/977aca12752a297cac1124d0d106313ff569ab4b/frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx#L174.

andrewballantyne, line 20: I'm confused -- what am I missing here. This isn't used by the component, why did you add it to the props?

christianvogt: @karthikjeeyar  ^^

karthikjeeyar: @andrewballantyne  When dropdownItem is selected, we are passing the selected resource (item from firehose) to the onChange handler, which takes K8sResourceKind as input. https://github.com/openshift/console/blob/977aca12752a297cac1124d0d106313ff569ab4b/frontend/packages/dev-console/src/components/pipelines/modals/triggers/TriggerBindingSelector.tsx#L22

andrewballantyne: But the component that uses these props:

```
const DropdownItem: React.FC<DropdownItemProps> = ({ model, name }) => (
    <span>
    </span>
    <span className="co-truncate show co-nowrap small">{name}</span>
  </span>
);
```
Doesn't use it... like at all. Why add it to the props? Firehose pushes a lot of props around into components that we don't use... is this somehow a TS error that `DropdownItem` doesn't define a prop?

karthikjeeyar: OnChange emitter from ResourceDropdown.tsx sends `key` and `value` (default value is  **string**) as arguments, In case of dropdownItem with badge the `value` argument is being set as a **React.Element**, so the newly introduced `resource` can be accessible as `ele?.props.resource` in the parent component. But I agree that the actual component does not require the `resource` at all . However I can add the key attribute
```
<span className="co-resource-item" key={resource.metadata.id} > 
....
</span>
``` 
let me know your thoughts on this

andrewballantyne: Keys really only have 1 main purpose (and a few less common ones)... and that's for Array rendering of a list of elements. Using a key on a non list item has really only one valid purpose I can think of and that's to force a re-mount of a component. I do not believe I can see any benefit from having a key on a `span` that is not rendered as part of a `.map` (or equiv).

I do not believe there is any value in saying this component gets a prop it does not use. It's not really future proofing the code against accidental changes... if anything it's doing the opposite and coupling this component to a prop it does not use, reducing its reusability.

karthikjeeyar: Currently ResourceDropdown does not have a way to emit the selected resource to the parent component. I will add the logic to emit the selected resource on `onChange` event instead of having this resource prop tied up to the dropdownItem component.

andrewballantyne: OH! NOW I understand what you're doing. Yeah... that's a very bad practice that you shouldn't do. Do not stuff variables into props of a component to extract them later from the internals of React. Components are not unbound objects you can just mutate to store data. 

Let me see what I can think of here.

andrewballantyne: So yesterday night I tried to investigate sending the data in the onChange handler, but it's not available and a rework of the Dropdown would be needed.

I then back back to the drawing board and maybe tried to manually add the badge to the old implementation - but it seems the selection fails to go through -- I called it for a day there before figuring out why. I'll try to find some time to continue my investigation today but I have a very full plate, so no promises.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -215,26 +218,28 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
...
+    const resource = resourceList[key];
+    this.setState({ resources: resourceList });
     const { actionItems, onChange, selectedKey } = this.props;
     const selectedActionItem = actionItems && actionItems.find((ai) => key === ai.actionKey);
     const title = selectedActionItem ? selectedActionItem.actionTitle : name;
     if (title !== this.state.title) {
       this.setState({ title });
     }
     if (key !== selectedKey) {
-      onChange && onChange(key, name, _.isEmpty(items));
```
karthikjeeyar, line 232: @andrewballantyne Modified the onChange handler in ResourceDropdown.tsx to pass the selected resource.
`.isEmpty(items)`  was introduced in early developement phase for PipelineResourceDropdownField.tsx in https://github.com/karthikjeeyar/console/commit/0f8674a0855692f0af066559502f9ee0affbde79 and is no longer needed as the PipelineResourceDropdownField's implementation got changed.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -215,26 +218,28 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
...
+      selectedItem && this.handleChange(selectedItem, sortedList, resourceList);
     }
     return sortedList;
   };
 
-  private handleChange = (key, items) => {
+  private handleChange = (key, items, resourceList) => {
     const name = items[key];
+    const resource = resourceList[key];
+    this.setState({ resources: resourceList });
```
andrewballantyne, line 229: This is very odd... you're setting state for something you should have in state unmutated?

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
andrewballantyne: Ah gotcha. Wasn't sure we wanted to modify this component. It's a Class component which made me think we didn't create it and thus shouldn't be cavalier modifying it... but sounds like we are good here.

frontend/packages/dev-console/src/components/pipelines/modals/triggers/TriggerBindingSelector.tsx
```diff
@@ -16,27 +16,40 @@ type TriggerBindingSelectorProps = {
 const TriggerBindingSelector: React.FC<TriggerBindingSelectorProps> = (props) => {
   const { description, label = TriggerBindingModel.label, onChange } = props;
   const { values } = useFormikContext<AddTriggerFormValues>();
+  const autoCompleteFilter = (strText: string, item: React.ReactElement): boolean =>
+    fuzzy(strText, item?.props?.name);
+  const onTriggerChange = (key: string, value: string, selectedResource) => {
```
andrewballantyne, line 21: Type please

frontend/packages/console-shared/src/components/formik-fields/ResourceDropdownField.tsx
```diff
@@ -62,8 +62,8 @@ const ResourceDropdownField: React.FC<ResourceDropdownFieldProps> = ({
           dropDownClassName={cx({ 'dropdown--full-width': fullWidth })}
           onLoad={onLoad}
           resourceFilter={resourceFilter}
-          onChange={(value: string, name) => {
-            props.onChange && props.onChange(value, name);
+          onChange={(value: string, name, resource) => {
```
andrewballantyne, line 65: These params really should have types.

andrewballantyne (changes requested):

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
karthikjeeyar: I have moved this setState out of this function, and made sure the setState is called only when firehose results are loaded or if it has new data.

frontend/packages/console-shared/src/components/formik-fields/ResourceDropdownField.tsx
```diff
@@ -62,8 +62,8 @@ const ResourceDropdownField: React.FC<ResourceDropdownFieldProps> = ({
           dropDownClassName={cx({ 'dropdown--full-width': fullWidth })}
           onLoad={onLoad}
           resourceFilter={resourceFilter}
-          onChange={(value: string, name) => {
-            props.onChange && props.onChange(value, name);
+          onChange={(value: string, name: string, resource: K8sResourceKind) => {
```
andrewballantyne, line 65:
```suggestion
          onChange={(value: string, name: string | object, resource: K8sResourceKind) => {
```
Technically, in our case, this is a React Element... but that's also just an object. We don't want to say `string` and get something that is not a `string` that defeats the purpose of TypeScript.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -81,6 +82,7 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
   constructor(props) {
     super(props);
     this.state = {
+      resources: {},
```
andrewballantyne, line 85: If the item is already loaded coming into the component, you have a chance of not setting it (ever or at least not immediately). I think you need to bring your logic from `UNSAFE_componentWillReceiveProps` here too (see the `items` line below this one).

andrewballantyne (changes requested):

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -158,6 +180,7 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
     updateSelection: boolean,
   ) => {
     const unsortedList = { ...appendItems };
+    const resourceList = {};
```
andrewballantyne, line 183: Hmm... is this needed anymore, you're setting it up front as props come in... or am I not following things?

karthikjeeyar: You are absolutely right, this isn't needed anymore after the last suggested change. My mistake, i have removed it

frontend/packages/console-shared/src/components/dropdown/__tests__/ResourceDropdown.spec.tsx
```diff
@@ -32,7 +32,7 @@ describe('ResourceDropdown test suite', () => {
     const spy = jest.fn();
     const component = shallow(componentFactory({ onChange: spy }));
     component.setProps({ resources: [] });
-    expect(spy).toHaveBeenCalledWith('#CREATE_APPLICATION_KEY#', undefined, true);
+    expect(spy).toHaveBeenCalledWith('#CREATE_APPLICATION_KEY#', undefined, undefined);
```
andrewballantyne, line 35: Interesting... pretty sure you don't need to provide `undefined` as the last argument to a function, testing or otherwise... `undefined` is implied by the nature of not providing it in JS.

karthikjeeyar: `toHaveBeenCalledWith` function expects the exact number of arguments to be passed.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -140,6 +142,25 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
...
+    _.each(resources, ({ data }) => {
+      _.each(data, (resource) => {
+        let key;
+        if (resourceFilter && resourceFilter(resource)) {
+          key = _.get(resource, dataSelector);
+        } else if (!resourceFilter) {
+          key = _.get(resource, dataSelector);
+        }
+        if (key) {
+          resourceList[key] = resource;
```
andrewballantyne, line 157: You're squashing duplicate entries of the same name -- which was the explicit issue mentioned in the ticket. Which confuses me even more because you noted that as a comment 🤔  This is an issue we need to fix in addition to adding the badge.

Screen Shot 2020-09-09 at 12 46 26 PM
Screen Shot 2020-09-09 at 12 46 49 PM

andrewballantyne (changes requested):

frontend/packages/console-shared/src/components/dropdown/__tests__/ResourceDropdown.spec.tsx
andrewballantyne: Interesting, fair enough... seemed odd at a glance.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
karthikjeeyar: In order to avoid the squashing of the entries with same name but different resource, we must need to change the logic of data accumulation in  `getDropdownList` function in ResourceDropdown.  This might require us to change the onChange handlers in some of the ResourceDropdown usages.
Can we have this additional issue addressed in a seperate PR & Ticket ? 
cc: @christianvogt

andrewballantyne: Ah right, the caller defines the firehose and does not actually have the data 🤔

andrewballantyne: So I just added a resource key generator and got it to work:

image

Diff is here if you're curious: https://gist.github.com/andrewballantyne/f0d82506da9aca7ccd688edd0911ac28

karthikjeeyar: Now, the generated resource Key will be passed via onChange handler, so the caller has to extract the resourceName from the custom-generated-key or use the new third argument (resource), right?

karthikjeeyar: My thought process was ResourceDropdown to implicitly handle multiple fiehose resources , thats why i was talking about a seperate PR, but based on your previous comment, i will update the component with `customResourceKey`  prop.Thanks @andrewballantyne

andrewballantyne: > Now, the generated resource Key will be passed via onChange handler, so the caller has to extract the resourceName from the custom-generated-key or use the new third argument (resource), right?

They crafted it, so yes they will have to extract it... the owner understands the what they did, they can reverse it.

frontend/packages/dev-console/src/components/pipelines/modals/triggers/TriggerBindingSelector.tsx
```diff
@@ -16,27 +16,42 @@ type TriggerBindingSelectorProps = {
 const TriggerBindingSelector: React.FC<TriggerBindingSelectorProps> = (props) => {
   const { description, label = TriggerBindingModel.label, onChange } = props;
   const { values } = useFormikContext<AddTriggerFormValues>();
+  const autoCompleteFilter = (strText: string, item: React.ReactElement): boolean =>
+    fuzzy(strText, item?.props?.name);
```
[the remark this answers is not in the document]
karthikjeeyar: @andrewballantyne Today, Almost all the dropdowns with multiple resources has this autoCompleteFilter feature by Design. In some places, we are customizing the dropdown item label, so this is the reason  i could think of (but its just me) for forcing this search implementation in caller functions. This autocompleteFilter function is defined this way throughout the console, do we cleanup in all the instances ? Let me know your thoughts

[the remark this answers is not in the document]
karthikjeeyar: AutocompleteFilter receives the third argument (key: string) as well, we can use that instead of reading it from the React.Element.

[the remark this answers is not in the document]
andrewballantyne: Let us leave it, log a bug and we can address this in 4.7 better.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -140,52 +143,76 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
     return true;
   }
 
-  private getDropdownList = (
-    {
+  private craftResourceKey = (
+    resource: K8sResourceKind,
+    props: ResourceDropdownProps,
+  ): { customKey: any; key: any } => {
```
divyanshiGupta, line 149: nit: it will be good if you can add the types here

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -140,52 +143,76 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
...
+    const { customResourceKey, resourceFilter, dataSelector } = props;
+    let key;
+    if (resourceFilter && resourceFilter(resource)) {
+      key = _.get(resource, dataSelector);
+    } else if (!resourceFilter) {
+      key = _.get(resource, dataSelector);
```
divyanshiGupta, line 155: use optional chaining here

divyanshiGupta (changes requested):

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
karthikjeeyar: dataSelector here is a prop which is been passed to this Resource dropdown component. The values of the dataSelector will be an array. 
eg: ```dataSelector={['metadata', 'labels', 'app.kubernetes.io/part-of']}```.  Hence the lodash _.get.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
karthikjeeyar: Added proper types here.

andrewballantyne (approved):

PR #6258, opened 2020-08-06
Add create project link in all the dev console pages

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-4448

**Analysis / Root cause**: Create project link was missing from project list page on Topology/Monitoring/Builds/Pipelines/Helm pages.

**Solution Description**: Use `CreateProjectListPage` instead of `ProjectListPage` in all pages.

@openshift/team-devconsole-ux @parvathyvr 

Peek 2020-08-07 01-37

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/AddPage.tsx
+++ b/frontend/packages/dev-console/src/components/AddPage.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import * as _ from 'lodash';
 import { Helmet } from 'react-helmet';
 import { match as RMatch } from 'react-router';
-import { history, Firehose, FirehoseResource, HintBlock } from '@console/internal/components/utils';
+import { Firehose, FirehoseResource, HintBlock } from '@console/internal/components/utils';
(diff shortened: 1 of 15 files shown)


serenamarie125:
> Hi @rohitkrai03 ! I see that we have the same text as +Add page in all the other views.What we wanted to do was just append the " or create a project" to the existing text on the other views. Eg: For Topology "Select a project to view the topology or create a project"

Agreed - original text should remain, and just append "or _create a project_"

rohitkrai03:
@parvathyvr @serenamarie125 I actually didn't change text for any of the pages. If you look at the diff in code you'll see that I am just using a different component which adds that functionality. I just didn't realize that the component which was earlier written for +Add page was hard coded with its text. I've updated the PR, take a look at the updated gif below - 

Peek 2020-08-07 20-33

PR #6259, opened 2020-08-06
Bug 1857399: Fix  chip to show All when no other selected

zherman0:
Bug was leaving empty `Resource` badge when last selected item removed.  Events should default to "All" when no specific resource is chosen, which also prevents a blank chip. UPDATE: If user clears initial `All` selection, we will show no resources selected which means no filter and show all events.
event-bug-3

### DIFF
--- a/frontend/public/components/events.jsx
+++ b/frontend/public/components/events.jsx
@@ -180,8 +180,14 @@ export class EventsList extends React.Component {
     }
   };
 
+  removeResource = (selection) => {
+    const updateItems = new Set(this.state.selected);
+    updateItems.delete(selection);
+    this.setState({ selected: updateItems });
+  };
+
   clearSelection = () => {
-    this.setState({ selected: new Set(['All']) });
+    this.setState({ selected: new Set() });
   };
 
   render() {
@@ -195,7 +201,6 @@ export class EventsList extends React.Component {
             <ResourceListDropdown
               onChange={this.toggleSelected}
               selected={Array.from(selected)}
-              showAll
               clearSelection={this.clearSelection}
               className="co-search-group__resource"
             />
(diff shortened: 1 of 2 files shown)


spadgett:
Hey @zherman0, I think the right fix is to make this filter work like all the other filters in console. We can remove the "All" item and show everything when no filters are selected. Even with your fix, there are some quirky behaviors. For instance, clicking the `x` when "All" is selected does nothing, and we're not consistent with other pages.

cc @beanh66

zherman0:
@spadgett @beanh66  `EventStream` defaults to all if no values are passed to it meaning the `ResourceDropdown` shows `0` selected but we still see all the events which does not make sense.  
Screenshot_2020-08-10 Events · OKD

zherman0:
@spadgett @beanh66 - Updated the description section above with new UX gif.  Check it out.  Sam, this is ready for another code review.

spadgett:
@zherman0 I think we should remove "All Resources" entirely from the dropdown. If no resource is selected, no filter is enabled, so we show all events.

zherman0:
@spadgett - Made this so no selection means show All events.  Ready for final review.

frontend/public/components/events.jsx
```diff
@@ -195,7 +201,6 @@ export class EventsList extends React.Component {
             <ResourceListDropdown
               onChange={this.toggleSelected}
               selected={Array.from(selected)}
-              showAll
```
spadgett, line 198: This is the only place `showAll` is used. We should go ahead and remove it from `ResourceListDropdown`

zherman0: I have no problem doing that, but the showAll flag has been around since the creation of the file.  It allows for the option to have select All Resources.  If I remove the flag then I will have to remove the All Resources logic as well.

zherman0: I guess my concern here is that this PR is a bug fix for the new toolbar chips showing when everything is cleared.  Now we have changed the UX to no longer show "All" selected and now we are discussing removing unused functionality in a common component.  I worry about the expanding scope of this PR.

zherman0:
/approved
@spadgett - ready for review

frontend/public/components/resource-dropdown.tsx
```diff
@@ -89,31 +89,16 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+  const items = resources
+    .map((model) => (
+      <DropdownItem
+        key={referenceForModel(model)}
+        model={model}
+        showGroup={isDup(model.kind)}
+        checked={isKindSelected(referenceForModel(model))}
+      />
+    ))
+    .toJS() as OrderedMap<string, JSX.Element>;
```
spadgett, line 101: This looks wrong... It's no longer an ordered map after calling `toJS()` right?

spadgett: I think you can just remove the `as`

```suggestion
    .toJS();
```

zherman0:
@spadgett - Ready for final review

spadgett (approved):
/lgtm

👍

PR #6261, opened 2020-08-07
Bug 1866997: Warn user that when HyperConverged and SR-IOVNetwork CRDs are missing

jhadvig:
If **HyperConverged** and **SR-IOVNetwork** CRDs are missing the Dropdown for the `NetworkType` is disabled, without any warning. We should be warning user that he needs to install the CRDs.

Screen:
Screenshot 2020-08-07 at 13 31 40

/assign @spadgett


jhadvig:
@pcbailey comment addressed. PTAL

PR #6262, opened 2020-08-07
Bug 1865817:  Values for cpu/memeory are empty in Metrics table on HPA detail page

cyril-ui-developer:
/assign @TheRealJon
/cc @spadgett

Screen Shot 2020-08-28 at 10 59 16 AM

@spadgett I wanted to test for currentMetrics values, but the yaml update returned `null` value for currentMetrics on save. I am thinking that could be a real time values generated from server api. Is that right?

### DIFF
--- a/frontend/public/components/hpa.tsx
+++ b/frontend/public/components/hpa.tsx
@@ -2,7 +2,11 @@ import * as React from 'react';
 import * as _ from 'lodash-es';
 import * as classNames from 'classnames';
 import { sortable } from '@patternfly/react-table';
-import { K8sResourceKind, K8sResourceKindReference } from '../module/k8s';
+import {
+  K8sResourceKind,
+  K8sResourceKindReference,
+  HorizontalPodAutoscalerKind,
+} from '../module/k8s';
 import { HorizontalPodAutoscalerModel } from '../models';
 import { Conditions } from './conditions';
 import { DetailsPage, ListPage, Table, TableRow, TableData, RowFunction } from './factory';
@@ -34,55 +38,51 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
 
 const externalRow = (metric, current, key) => {
   const { external } = metric;
-  const type = external.metricName;
-  // TODO: show metric selector for external metrics?
-  const currentValue = external.targetAverageValue
-    ? _.get(current, 'object.currentAverageValue')
-    : _.get(current, 'object.currentValue');
-  const targetValue = external.targetAverageValue
-    ? external.targetAverageValue
-    : external.targetValue;
+  const type = external.metric.name;
+  const currentValue =
+    current?.external?.current?.averageValue || current?.external?.current?.value;
+  const targetValue = external.target.averageValue || external.target.value;
 
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
 
-const objectRow = (metric, current, ns, key) => {
+const objectRow = (metric, current, ns, key, scaleTarget) => {
   const { object } = metric;
   const type = (
     <>
-      {object.metricName} on
+      {object.metric.name} on
       <ResourceLink
-        kind={object.target.kind}
-        name={object.target.name}
+        kind={scaleTarget.kind}
+        name={scaleTarget.name}
         namespace={ns}
-        title={object.target.name}
+        title={object.metric.name}
       />
     </>
   );
-  const currentValue = _.get(current, 'object.currentValue');
-  const targetValue = object.targetValue;
+  const currentValue = current?.object?.current.value;
+  const targetValue = object.target.value;
 
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
 
 const podRow = (metric, current, key) => {
   const { pods } = metric;
-  const type = `${pods.metricName} on pods`;
-  const currentValue = _.get(current, 'pods.currentAverageValue');
-  const targetValue = pods.targetAverageValue;
+  const type = `${pods.metric.name} on pods`;
+  const currentValue = current?.pods?.current.averageValue;
+  const targetValue = pods.target.averageValue;
 
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
 
 const getResourceUtilization = (currentMetric) => {
-  const currentUtilization = _.get(currentMetric, 'resource.currentAverageUtilization');
+  const currentUtilization = currentMetric?.resource?.current?.averageUtilization;
 
   // Use _.isFinite so that 0 evaluates to true, but null / undefined / NaN don't
   if (!_.isFinite(currentUtilization)) {
     return null;
   }
 
-  const currentAverageValue = _.get(currentMetric, 'resource.currentAverageValue');
+  const currentAverageValue = currentMetric?.resource?.current?.averageValue;
   // Only show currentAverageValue in parens if set and non-zero to avoid things like "0% (0)"
   return currentAverageValue && currentAverageValue !== '0'
     ? `${currentUtilization}% (${currentAverageValue})`
@@ -90,8 +90,9 @@ const getResourceUtilization = (currentMetric) => {
 };
 
 const resourceRow = (metric, current, key) => {
-  const targetUtilization = metric.resource.targetAverageUtilization;
-  const resourceLabel = `resource ${metric.resource.name}`;
+  const { resource } = metric;
+  const targetUtilization = resource.target.averageUtilization;
+  const resourceLabel = `resource ${resource.name}`;
   const type = targetUtilization ? (
     <>
       {resourceLabel}&nbsp;<span className="small text-muted">(as a percentage of request)</span>
@@ -101,10 +102,8 @@ const resourceRow = (metric, current, key) => {
   );
   const currentValue = targetUtilization
     ? getResourceUtilization(current)
-    : _.get(current, 'resource.currentAverageValue');
-  const targetValue = targetUtilization
-    ? `${targetUtilization}%`
-    : metric.resource.targetAverageValue;
+    : resource?.current?.averageValue;
+  const targetValue = targetUtilization ? `${targetUtilization}%` : resource.target.averageValue;
 
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
@@ -127,7 +126,13 @@ const MetricsTable: React.FC<MetricsTableProps> = ({ obj: hpa }) => {
               case 'External':
                 return externalRow(metric, current, i);
               case 'Object':
-                return objectRow(metric, current, hpa.metadata.namespace, i);
+                return objectRow(
+                  metric,
+                  current,
+                  hpa.metadata.namespace,
+                  i,
+                  hpa.spec.scaleTargetRef,
+                );
               case 'Pods':
                 return podRow(metric, current, i);
               case 'Resource':
@@ -339,7 +344,7 @@ export const HorizontalPodAutoscalersPage: React.FC<HorizontalPodAutoscalersPage
 HorizontalPodAutoscalersPage.displayName = 'HorizontalPodAutoscalersListPage';
 
 export type HorizontalPodAutoscalersDetailsProps = {
-  obj: any;
+  obj: HorizontalPodAutoscalerKind;
 };
 
 export type HorizontalPodAutoscalersPageProps = {
@@ -353,7 +358,7 @@ export type HorizontalPodAutoscalersDetailsPageProps = {
 };
 
 type MetricsTableProps = {
-  obj: any;
+  obj: HorizontalPodAutoscalerKind;
 };
 
 type MetricsRowProps = {
--- a/frontend/public/module/k8s/types.ts
+++ b/frontend/public/module/k8s/types.ts
@@ -364,16 +364,71 @@ export type DeploymentKind = {
   };
 } & K8sResourceCommon;
 
+type CurrentObject = {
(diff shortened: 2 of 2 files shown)


andrewballantyne:
Ah crud, sorry about that! I should have investigated the details page more when I upgraded the version to beta2. Thanks for the PR!

spadgett:
Looks like this broke in #6022 when we changed the API version. Let me look at the API between v2beta1 and v2beta2 to see if anything else changed that might impact us.

spadgett:
I think there are other changes needed as well for `externalRow`, `objectRow`, and `podRow` like `currentAverageValue` -> `current.averageValue`, etc.

spadgett:
You can try some of the examples here to get additional data:

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics

andrewballantyne:
Been thinking about how I could have avoided making this mistake... and just being more attentive to what already existed if an obvious one. 

However I remembered I had to create the [HorizontalPodAutoscalerKind](https://github.com/openshift/console/blob/master/frontend/public/module/k8s/types.ts#L377) type during my work. I think, hopefully part of this PR, we can start making use of the type too. That way TypeScript can help ensure we align when changes in spec happen.

spadgett:
Yeah, I think the best way is to make sure we have TypeScript definitions for everything and update them when we bump an API version. You can see the schema using `oc explain` or Home -> Explore.

@cyril-ui-developer Do you mind updating the `obj: any` to `obj: HorizontalPodAutoscalerKind` in this PR? I think we'll want to fill out the `currentMetrics` value from that type so it isn't `any`.

cyril-ui-developer:
> Yeah, I think the best way is to make sure we have TypeScript definitions for everything and update them when we bump an API version. You can see the schema using `oc explain` or Home -> Explore.
> 
> @cyril-ui-developer Do you mind updating the `obj: any` to `obj: HorizontalPodAutoscalerKind` in this PR? I think we'll want to fill out the `currentMetrics` value from that type so it isn't `any`.

@spadgett Sure, will do!

cyril-ui-developer:
> I think there are other changes needed as well for `externalRow`, `objectRow`, and `podRow` like `currentAverageValue` -> `current.averageValue`, etc.

@spadgett I am not too sure if the API changes impacted these `object` properties based on K8 HPA schema shared 
[https://github.com/openshift/console/blob/master/frontend/public/components/hpa.tsx#L55](url)

spadgett:
@cyril-ui-developer It should. You can verify with `oc explain` (or Home -> Explore in console).

```
❯ oc explain hpa.status.currentMetrics.object --api-version=autoscaling/v2beta2
KIND:     HorizontalPodAutoscaler
VERSION:  autoscaling/v2beta2

RESOURCE: object <Object>

DESCRIPTION:
     object refers to a metric describing a single kubernetes object (for
     example, hits-per-second on an Ingress object).

     ObjectMetricStatus indicates the current value of a metric describing a
     kubernetes object (for example, hits-per-second on an Ingress object).

FIELDS:
   current      <Object> -required-
     current contains the current value for the given metric

   describedObject      <Object> -required-

   metric       <Object> -required-
     metric identifies the target metric by name and selector
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,66 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
-  const type = external.metricName;
+  const type = external.metric.name;
   // TODO: show metric selector for external metrics?
-  const currentValue = external.targetAverageValue
-    ? _.get(current, 'object.currentAverageValue')
-    : _.get(current, 'object.currentValue');
-  const targetValue = external.targetAverageValue
-    ? external.targetAverageValue
-    : external.targetValue;
+  const currentValue = current?.external?.current?.averageValue
```
cyril-ui-developer, line 43: @spadgett I think line 43 should be `const currentValue = current?.external?.current?.averageValue` instead of `const currentValue = external.target.averageValue` since the conditional check is for `current?.external?.current?.averageValue`.

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,66 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
+    ? current?.external?.current?.averageValue
+    : current?.external?.currentValue;
+  const targetValue = external?.target?.averageValue
+    ? external.target.averageValue
+    : external.target.value;
```
spadgett, line 48:
```suggestion
  const targetValue = external?.target?.averageValue || external.target.value;
```

frontend/public/module/k8s/types.ts
```diff
@@ -390,6 +390,7 @@ export type HorizontalPodAutoscalerKind = K8sResourceCommon & {
     desiredReplicas: number;
     currentMetrics: any;
```
spadgett, line 391: Can you define this instead of using any?

frontend/public/module/k8s/types.ts
```diff
@@ -390,6 +390,7 @@ export type HorizontalPodAutoscalerKind = K8sResourceCommon & {
...
     conditions: NodeCondition[];
+    lastScaleTime: string;
```
spadgett, line 393:
```suggestion
    lastScaleTime?: string;
```

frontend/public/components/hpa.tsx
```diff
@@ -127,7 +131,7 @@ const MetricsTable: React.FC<MetricsTableProps> = ({ obj: hpa }) => {
               case 'External':
                 return externalRow(metric, current, i);
               case 'Object':
-                return objectRow(metric, current, hpa.metadata.namespace, i);
+                return objectRow(metric, current, hpa?.metadata?.namespace, i, hpa?.kind);
```
spadgett, line 134: This looks wrong. It shouldn't be HPA kind. It should be scale target kind.

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,66 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
+const objectRow = (metric, current, ns, key, kind) => {
   const { object } = metric;
   const type = (
     <>
-      {object.metricName} on
+      {object?.metric?.name} on
       <ResourceLink
-        kind={object.target.kind}
-        name={object.target.name}
+        kind={kind}
```
spadgett, line 59: This should come from `spec.scaleTargetRef.kind` in the HPA

frontend/public/module/k8s/types.ts
```diff
@@ -388,8 +425,9 @@ export type HorizontalPodAutoscalerKind = K8sResourceCommon & {
   status?: {
     currentReplicas: number;
     desiredReplicas: number;
-    currentMetrics: any;
+    currentMetrics: CurrentMetrics[];
```
spadgett, line 428:
```suggestion
    currentMetrics?: CurrentMetrics[];
```

frontend/public/module/k8s/types.ts
```diff
@@ -374,6 +374,43 @@ export type HPAMetric = {
...
+    metric: MetricObject;
+  };
+  pods: {
+    current: CurrentObject;
+    metric: MetricObject;
+  };
+  resource: {
+    name: string;
+    current: CurrentObject;
+  };
```
spadgett, line 410: Many of these are optional props.

```suggestion
  external?: {
    current: CurrentObject;
    metric: MetricObject;
  };
  object?: {
    current: CurrentObject;
    describedObject: {
      apiVersion?: string;
      kind: string;
      name: string;
    };
    metric: MetricObject;
  };
  pods?: {
    current: CurrentObject;
    metric: MetricObject;
  };
  resource?: {
    name: string;
    current: CurrentObject;
  };
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
+        name={object?.metric?.name}
```
spadgett, line 56: This doesn't look right. I think this should be the scale target, not the name of the metric.

You should set up the HPA to target a real resource and make sure the link works.

cyril-ui-developer: @spadgett Addressed, thanks

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
 
-const objectRow = (metric, current, ns, key) => {
+const objectRow = (metric, current, ns, key, scaleTarget) => {
   const { object } = metric;
   const type = (
     <>
-      {object.metricName} on
+      {object?.metric?.name} on
```
spadgett, line 53: `metric.name` is required, and we don't use optional chaining in `externalRow`. We should either always use it or never use it. It shouldn't be needed, though.

```suggestion
      {object.metric.name} on
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
       <ResourceLink
-        kind={object.target.kind}
-        name={object.target.name}
+        kind={scaleTarget.kind}
+        name={scaleTarget?.name}
```
spadgett, line 56: We should either use optional chaining on both of these lines or neither. `scaleTargetRef` is required, so optional chaining shouldn't be needed.

```suggestion
        kind={scaleTarget.kind}
        name={scaleTarget.name}
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
         namespace={ns}
-        title={object.target.name}
+        title={object?.metric?.name}
```
spadgett, line 58:
```suggestion
        title={object.metric.name}
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
       />
     </>
   );
-  const currentValue = _.get(current, 'object.currentValue');
-  const targetValue = object.targetValue;
+  const currentValue = current?.object?.current.value;
+  const targetValue = object?.target?.value;
```
spadgett, line 63:
```suggestion
  const targetValue = object.target.value;
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
 
   return <MetricsRow key={key} type={type} current={currentValue} target={targetValue} />;
 };
 
 const podRow = (metric, current, key) => {
   const { pods } = metric;
-  const type = `${pods.metricName} on pods`;
-  const currentValue = _.get(current, 'pods.currentAverageValue');
-  const targetValue = pods.targetAverageValue;
+  const type = `${pods?.metric?.name} on pods`;
```
spadgett, line 70:
```suggestion
  const type = `${pods.metric.name} on pods`;
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
+  const currentValue = current?.pods?.current?.averageValue;
+  const targetValue = pods?.target?.averageValue;
```
spadgett, line 72:
```suggestion
  const targetValue = pods.target.averageValue;
```

frontend/public/components/hpa.tsx
```diff
@@ -34,64 +38,61 @@ const MetricsRow: React.FC<MetricsRowProps> = ({ type, current, target }) => (
...
   return currentAverageValue && currentAverageValue !== '0'
     ? `${currentUtilization}% (${currentAverageValue})`
     : `${currentUtilization}%`;
 };
 
 const resourceRow = (metric, current, key) => {
-  const targetUtilization = metric.resource.targetAverageUtilization;
-  const resourceLabel = `resource ${metric.resource.name}`;
+  const { resource } = metric;
+  const targetUtilization = resource?.target?.averageUtilization;
```
spadgett, line 94:
```suggestion
  const targetUtilization = resource.target.averageUtilization;
```

frontend/public/components/hpa.tsx
```diff
@@ -127,7 +126,13 @@ const MetricsTable: React.FC<MetricsTableProps> = ({ obj: hpa }) => {
               case 'External':
                 return externalRow(metric, current, i);
               case 'Object':
-                return objectRow(metric, current, hpa.metadata.namespace, i);
+                return objectRow(
+                  metric,
+                  current,
+                  hpa?.metadata?.namespace,
```
spadgett, line 132:
```suggestion
                  hpa.metadata.namespace,
```

frontend/public/components/hpa.tsx
```diff
@@ -127,7 +126,13 @@ const MetricsTable: React.FC<MetricsTableProps> = ({ obj: hpa }) => {
...
+                  i,
+                  hpa?.spec?.scaleTargetRef,
```
spadgett, line 134:
```suggestion
                  hpa.spec.scaleTargetRef,
```

cyril-ui-developer:
@spadgett Addressed, thanks!

PR #6263, opened 2020-08-07
WIP Monitoring links

abhinandan13jan:
https://issues.redhat.com/browse/ODC-4397
Need to fix tests

Screencast from 08-07-2020 06_33_02 PM (1)
Screencast from 08-07-2020 06_32_21 PM


debsmita1:
Thank you for working on this. This works fine, I'm just unsure about the part where you are redirecting the user to `/monitoring/dashboards` when clicked on `view monitoring dashboard` in admin perspective. In the dev side, we do show workload specific dashboard so that part is fine. May be @cshinn  can help here

abhinandan13jan:
/retitle Bug 1868013: Point admin monitoring links to admin

frontend/packages/dev-console/src/components/monitoring/overview/MonitoringOverview.tsx
```diff
@@ -132,7 +133,11 @@ const MonitoringOverview: React.FC<MonitoringOverviewProps> = (props) => {
               <>
                 <div className="odc-monitoring-overview__view-monitoring-dashboard">
                   <Link
-                    to={`/dev-monitoring/ns/${resource?.metadata?.namespace}/?workloadName=${resource?.metadata?.name}&workloadType=${resource?.kind}`}
+                    to={
+                      currentPath.startsWith('/k8s')
```
sahil143, line 137: I think it would be better if we can do it based on the active perspective. WDYT?

cshinn:
@abhinandan13jan @debsmita1 It seems like the best thing to link to in the admin console would be the `grafana-dashboard-k8s-resources-workloads-namespace` board for the relevant namespace/workload. It has the same context as the dev version would and shows the same data as the sidebar
Screen Shot 2020-08-12 at 9 50 29 AM

frontend/packages/dev-console/src/components/monitoring/overview/MonitoringOverview.tsx
abhinandan13jan: @sahil143 The issue with checking last active perspective is that it returns 'admin' on `incognito` while You are on dev

abhinandan13jan:
> @abhinandan13jan @debsmita1 It seems like the best thing to link to in the admin console would be the `grafana-dashboard-k8s-resources-workload` board for the relevant namespace/workload. It has the same context as the dev version would and shows the same data as the sidebar
> Screen Shot 2020-08-12 at 9 53 27 AM

@cshinn  We can redirect to `monitoring/dashboards/grafana-dashboard-k8s-resources-workload` but the namespace or workload name cannot be sent over the URL. So, it will not have the context

debsmita1:
cc @invincibleJai @kyoto ^

frontend/packages/dev-console/src/components/monitoring/overview/MonitoringOverview.tsx
rohitkrai03: @abhinandan13jan Why would last active perspective return `admin` on `incognito` while you're on `dev`? Both localStorage and redux store gets initialized as expected on `incognito` as well otherwise the application wouldn't work.

rohitkrai03: I don't think it's a good idea to check for a specific URL pattern. What if this component is added to some other page later which doesn't conform to the above URL pattern? The whole logic would break then. I think it's better to use `activePerspective` from redux here..

sahil143: I agree with @rohitkrai03. We shouldn't be checking based on the URL pattern. If it is returning `admin` on incognito while in `dev` then it's a bug and we should look to fix it.

debsmita1: @sahil143 @rohitkrai03 The issue is, if you copy a dev-perspective URL and open it in incognito, the activePerspective in the redux is set as `admin` . Have a look at the below screenshot
Screenshot from 2020-08-18 22-49-32

sahil143: @debsmita1 That happens for any devconsole url in incognito.

abhinandan13jan:
@rohitkrai03 @sahil143 @debsmita1 I have updated to use getActivePerspective and it seems to work fine

frontend/packages/dev-console/src/components/monitoring/overview/MonitoringOverview.tsx
abhinandan13jan: I have updated to use getActivePerspective for convention as well and it seems to work fine

rohitkrai03 (approved):
/lgtm

invincibleJai:
in Admin perspective if i go from sidebar to dashboard , it selects in correct namespace

image

rohitkrai03:
> in Admin perspective if i go from sidebar to dashboard , it doesn't selects current namespace
> 
> > @cshinn We can redirect to monitoring/dashboards/grafana-dashboard-k8s-resources-workload but the namespace or workload name cannot be sent over the URL. So, it will not have the context cc: @debsmita1
> 
> @kyoto can add here , if there is a way to update proper ns/workloads in this scenario or redirecting to `grafana-dashboard-k8s-resources-workload` is the way now cc @abhinandan13jan
> 
> image

AFAIK, there's no support for passing namespace or workload context in the grafana dashboard route currently. Even the namespace dropdown selects first item in the dropdown instead of `activeNamespace` from redux store. This results in user being redirected to wrong namespace context. Not sure what we can do for the scope of this PR.

kyoto:
In the admin perspective Dashboards page, the "Namespace" dropdown represents a variable in the dashboard definition (same as "Workload" and "Type"). The page therefore doesn't have any knowledge of what the variable "Namespace" means, so it isn't connected to the active namespace.

Unfortunately, there's currently no way to automatically select a value from the "Namespace" dropdown (or any other variable dropdown).

PR #6264, opened 2020-08-07
Bug 1866035: fix util to get Kiali link from consoleLinks

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4458

**Analysis / Root cause**: 
Incorrectly assumes any "namespace dashboard" link is a Kiali link

**Solution Description**: 
Get the namespace Dashboard Kiali link as per `metadat.name` attributes of the consoleLink.


vikram-raj:
/retitle Bug 1866035: fix util to get Kiali link from consoleLinks

PR #6269, opened 2020-08-09
Bug 1844944: Add default labels to default sc

yaacov:
Add a "(default)" string to the default storage class option.

Screesnshots:
Before:
Screenshot from 2020-08-09 10-56-06

After:
Screenshot from 2020-08-09 11-09-19

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
@@ -19,7 +19,7 @@ import {
   PersistentVolumeClaimModel,
   StorageClassModel,
 } from '@console/internal/models';
-import { getName } from '@console/shared/src';
+import { getName, getAnnotations } from '@console/shared/src';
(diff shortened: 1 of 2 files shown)


yaacov:
@matthewcarleton please review

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -539,16 +541,30 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
             </FormSelect>
           </FormRow>
           {source.requiresStorageClass() && (
-            <K8sResourceSelectRow
+            <FormSelect
```
glekner, line 544: This change works but you lose the value that comes with `K8sResourceSelectRow` which is the loading state.
I suggest adding a generic `defaultAnnotation` prop to `K8sResourceSelectRow` and update the component to look for it in the annotations and if its there add `(default)` to the name

yaacov: :+1: done

glekner:
nice!
/lgtm

PR #6280, opened 2020-08-10
Bug 1866514: If CSV doesn't have status field, check it's labels for 'olm.copiedFrom' label

jhadvig:
When global installing/uninstalling operators a flickering is cause due to the fact that there is an extra CSV of the installed/uninstalled operator, that's without a `status` field and so was passing though our `isCopiedCSV` filter. If the CSV doesn't have the `status` field, check it's labels for the `olm.copiedFrom` label. 

/assign @spadgett


jhadvig:
/test e2e-gcp-console
--

PR #6281, opened 2020-08-10
Bug 1870632: Add context menu and details panel for SBRs in topology

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4461

**Description**
Adds missing context menu for Service Binding Requests in topology as well as the missing details tab and link to the SBR resource page.

**Screen Shots**
image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


frontend/packages/dev-console/src/components/topology/operators/actions/removeServiceBinding.tsx
```diff
@@ -1,35 +0,0 @@
...
-import { Edge } from '@patternfly/react-topology';
-import { k8sKill } from '@console/internal/module/k8s';
-import { ServiceBindingRequestModel } from '../../../../models';
-
-export const removeServiceBinding = (edge: Edge): Promise<any> => {
-  const message = (
-    <p>
-      Deleting the binding connector deletes the config details of the source and removes the
-      binding resources. Are you sure you want to delete the binding connector?
-    </p>
```
invincibleJai, line 13: Don't we need this custom message now for SBR while deletion?

invincibleJai:
@jeff-phillips-18  do we need to show delete icon as well on hover along with context menu?

image

frontend/packages/dev-console/src/components/topology/operators/actions/removeServiceBinding.tsx
jeff-phillips-18: We use the standard delete now so, no this message is not needed.

jeff-phillips-18:
@invincibleJai  There was no conclusion to a discussion raised about this so I have not proceeded with removing it. If we want to remove it we would need context menus on visual connectors as well so that we could remove it from all edges.

/cc @bgliwa01 @beaumorley

invincibleJai:
> @invincibleJai There was no conclusion to a discussion raised about this so I have not proceeded with removing it. If we want to remove it we would need context menus on visual connectors as well so that we could remove it from all edges.
> 
> /cc @bgliwa01 @beaumorley

Yeah just wanted confirmation on it as then i would need to log an issue to delete icon for subscription/trigger as well

bgliwa01:
@invincibleJai @jeff-phillips-18 if we can have the trash icon disappear and have in context menus for all of the connectors, not just ones that are resources, that would be the ideal scenario from UX perspective

invincibleJai:
> @invincibleJai @jeff-phillips-18 if we can have the trash icon disappear and have in context menus for all of the connectors, not just ones that are resources, that would be the ideal scenario from UX perspective

Thanks @bgliwa01 for regular connector which just has one option delete , trash makes sense on hover then context menu or can have both across. I raised this just to be consistent

@jeff-phillips-18 @christianvogt let us know your thoughts and for this PR POV if we are fine with both options if connector are resources and has more options then i'll file issue for subscription/trigger and handle in that

bgliwa01:
@invincibleJai Can we get rid of the trash can all together?

invincibleJai:
> @invincibleJai Can we get rid of the trash can all together?

yes we can @jeff-phillips-18 let me know if i shall create an issue to handle it separately?

jeff-phillips-18:
I have remove the hover over trash can to delete edges and added context menus for them:

image

rottencandy (approved):
/lgtm

rohitkrai03 (approved):
/approve

invincibleJai:
Looks good , thanks @jeff-phillips-18 !!

PR #6228, opened 2020-08-05
Bug 1866845: support pipeline task resources to be an optional field based on spec

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-4281

Problem: 
Pipelines resources are all set to mandatory while using the pipeline builder

Solution:
Pipeline resources are set as required based on the `optional` attribute from the pipeline task spec.

Screenshots:
Before:
Screenshot 2020-07-07 at 16 44 15
After: 
image

Gif:
optinal-pipeline

cc: @andrewballantyne @siamaksade

PR #6282, opened 2020-08-10
Bug 1867526: support pipeline task resources to be an optional field based on spec

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-4281

Problem:
Pipelines resources are all set to mandatory while using the pipeline builder

Solution:
Pipeline resources are set as required based on the optional attribute from the pipeline task spec.

This is a manual cherry pick of https://github.com/openshift/console/pull/6228


nemesis09:
verified locally.
/lgtm

PR #6283, opened 2020-08-10
Bug 1867238:  utilize channels from ClusterVersion when available

rhamilto:
Due to the recent addition of `cv.status.desired.channels`, we cannot make a hard switch over to make use of `cv.status.desired.channels` as many builds do not yet contain those changes, and removing the hard-coded values outright will break the console for those builds.

https://issues.redhat.com/browse/PD-579 documents some follow-on design items that will need to be addressed in the future.

Screen Shot 2020-08-10 at 2 16 34 PM

resulting from

```
apiVersion: config.openshift.io/v1
kind: ClusterVersion
metadata:
  creationTimestamp: '2020-08-10T17:44:11Z'
  generation: 6
  name: version
  resourceVersion: '72671'
  selfLink: /apis/config.openshift.io/v1/clusterversions/version
  uid: 00a05f63-ab93-42a7-82e4-8b58a2504203
spec:
  channel: candidate-4.6
  clusterID: caa924e0-11f7-4e8b-a293-0fd0f0a39600
  upstream: 'https://api.openshift.com/api/upgrades_info/v1/graph'
status:
  availableUpdates: null
  conditions:
    - lastTransitionTime: '2020-08-10T18:11:33Z'
      message: Done applying 4.6.0-fc.0
      status: 'True'
      type: Available
    - lastTransitionTime: '2020-08-10T17:56:19Z'
      status: 'False'
      type: Failing
    - lastTransitionTime: '2020-08-10T18:11:33Z'
      message: Cluster version is 4.6.0-fc.0
      status: 'False'
      type: Progressing
    - lastTransitionTime: '2020-08-10T18:59:34Z'
      status: 'True'
      type: RetrievedUpdates
  desired:
    channels:
      - candidate-4.6
    image: >-
      quay.io/openshift-release-dev/ocp-release@sha256:45e6bc583040384efb4033b22c58f054b12ac32c7874554885d74a0faf6fef79
    version: 4.6.0-fc.0
  history:
    - completionTime: '2020-08-10T18:11:33Z'
      image: >-
        quay.io/openshift-release-dev/ocp-release@sha256:45e6bc583040384efb4033b22c58f054b12ac32c7874554885d74a0faf6fef79
      startedTime: '2020-08-10T17:44:18Z'
      state: Completed
      verified: false
      version: 4.6.0-fc.0
  observedGeneration: 6
  versionHash: kVdi1UOYMBM=
```

cc:  @megan-hall


frontend/public/module/k8s/cluster-settings.ts
```diff
@@ -39,11 +39,21 @@ export const getSortedUpdates = (cv: ClusterVersionKind): ClusterUpdate[] => {
...
 };
 
-export const getAvailableClusterChannels = () => ({
-  'stable-4.6': 'stable-4.6',
-  'fast-4.6': 'fast-4.6',
-  'candidate-4.6': 'candidate-4.6',
-});
+export const getAvailableClusterChannels = (cv) => {
+  if (cv?.status?.desired?.channels) {
+    return cv?.status?.desired?.channels.reduce((o, val) => {
```
spadgett, line 44: No need for optional chaining here since you check above in the `if` condition

```suggestion
    return cv.status.desired.channels.reduce((o, val) => {
```

rhamilto: Fixed.  Thanks, @spadgett.

rhamilto:
/hold

It occurs to me it's a bit silly to transform an array to an object and then back to an array.

rhamilto:
/hold cancel

@spadgett, please take another look.

PR #6292, opened 2020-08-11
Bug 1777129: Tighten cipher usage to library-go

jhadvig:
Use the cipher suites that are provided by the OpenShift library-go as the reasonable defaults, based on https://github.com/openshift/oauth-proxy/pull/130

/assign @spadgett 
cc'ing @stlaz


cmd/bridge/main.go
```diff
@@ -19,12 +19,12 @@ import (
 
 	"github.com/openshift/console/pkg/auth"
 	"github.com/openshift/console/pkg/bridge"
-	"github.com/openshift/console/pkg/crypto"
```
spadgett, line 22: Can you remove `pkg/crypto` from console if it's no longer needed?

jhadvig: yeah, we probably should!

jhadvig: The line you you are mentioning was changed but the second commit:
Screenshot 2020-08-11 at 15 44 31

jhadvig:
@spadgett I've removed the console's `crypto` pkg.
PTAL :)

PR #6296, opened 2020-08-11
Bug 1868353: update v2v temporary label

suomiy:


suomiy:
@yaacov references correct bug, please take a look

PR #6299, opened 2020-08-11
Bug 1866034: Fix to reduce pod donut text size when text is long

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4046

**Description**
Fixes text size in pod ring donut when showing long text such that the text does not overflow.

**Screen Shots**
image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/console-shared/src/utils/pod-ring-text.scss
+++ b/frontend/packages/console-shared/src/utils/pod-ring-text.scss
@@ -5,9 +5,11 @@
     font-size: 14px !important;
     fill: var(--pf-global--Color--200) !important;
   }
-  tspan:first-of-type {
-    font-size: 24px !important;
-    fill: var(--pf-global--Color--100) !important;
+  &:not(.pod-ring__long-text) {
+    tspan:first-of-type {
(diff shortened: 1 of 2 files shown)


frontend/packages/console-shared/src/utils/pod-ring-text.scss
```diff
@@ -5,9 +5,11 @@
     font-size: 14px !important;
     fill: var(--pf-global--Color--200) !important;
   }
-  tspan:first-of-type {
-    font-size: 24px !important;
-    fill: var(--pf-global--Color--100) !important;
+  &:not(.pod-ring__long-text) {
+    tspan:first-of-type {
+      font-size: 24px !important;
```
karthikjeeyar, line 10:
```suggestion
      font-size: var(--pf-chart-global--FontSize--2xl) !important;
```

karthikjeeyar:
Verified locally, works fine.
/lgtm

jeff-phillips-18:
Thanks @andrewballantyne, fixed.

PR #6302, opened 2020-08-11
Bug 1858422: update CRD apiVersion to v1

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1858422

With an assist from @TheRealJon 

Doc links for reference:
* https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
* https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinition-v1-apiextensions-k8s-io

### DIFF
--- a/frontend/__mocks__/crds.ts
+++ b/frontend/__mocks__/crds.ts
@@ -1,7 +1,7 @@
 import { CRDVersion, CustomResourceDefinitionKind } from '../public/module/k8s';
 
 export const testCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
+  apiVersion: 'apiextensions.k8s.io/v1',
   kind: 'CustomResourceDefinition',
   metadata: {
     name: 'testresources.testapp.coreos.com',
@@ -15,16 +15,23 @@ export const testCRD: CustomResourceDefinitionKind = {
   },
   spec: {
     group: 'testapp.coreos.com',
-    version: 'v1alpha1',
-    validation: {
-      openAPIV3Schema: {},
-    },
+    versions: [
+      {
+        name: 'v1alpha1',
+        storage: true,
+        served: true,
+        schema: {
+          openAPIV3Schema: {},
+        },
+      },
+    ],
     names: {
       kind: 'TestResource',
       plural: 'testresources',
       singular: 'testresource',
       listKind: 'TestResourceList',
     },
+    scope: 'Namespaced',
   },
 };
 
@@ -32,65 +39,91 @@ export const testCRDVersionV1Alpha1: CRDVersion = {
   name: 'v1alpha1',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV1Alpha2: CRDVersion = {
   name: 'v1alpha2',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV8Unserved: CRDVersion = {
   name: 'v8',
   storage: true,
   served: false,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV2Beta1: CRDVersion = {
   name: 'v2beta1',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV2Alpha1: CRDVersion = {
   name: 'v2alpha1',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV3Beta1: CRDVersion = {
   name: 'v3beta1',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDVersionV1: CRDVersion = {
   name: 'v1',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDInvalidVersion: CRDVersion = {
   name: 'versionthreedotone',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testCRDInvalidVersionTwo: CRDVersion = {
   name: '7.45',
   storage: true,
   served: true,
+  schema: {
+    openAPIV3Schema: {},
+  },
 };
 
 export const testForValidVersionsCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
+  apiVersion: 'apiextensions.k8s.io/v1',
   kind: 'CustomResourceDefinition',
   metadata: {
     name: 'testresources.testapp.coreos.com',
   },
   spec: {
     group: 'testapp.coreos.com',
-    version: 'wrong-version',
     versions: [
       testCRDVersionV1Alpha1,
       testCRDVersionV1Alpha2,
@@ -106,18 +139,18 @@ export const testForValidVersionsCRD: CustomResourceDefinitionKind = {
       singular: 'testresource',
       listKind: 'TestResourceList',
     },
+    scope: 'Namespaced',
   },
 };
 
 export const testForUnservedVersionsCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
+  apiVersion: 'apiextensions.k8s.io/v1',
   kind: 'CustomResourceDefinition',
   metadata: {
     name: 'testresources.testapp.coreos.com',
   },
   spec: {
     group: 'testapp.coreos.com',
-    version: 'correct-version',
     versions: [testCRDVersionV8Unserved],
     names: {
       kind: 'TestResource',
@@ -125,42 +158,6 @@ export const testForUnservedVersionsCRD: CustomResourceDefinitionKind = {
       singular: 'testresource',
       listKind: 'TestResourceList',
     },
-  },
-};
-
-export const testForInvalidVersionsCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
-  kind: 'CustomResourceDefinition',
-  metadata: {
-    name: 'testresources.testapp.coreos.com',
-  },
-  spec: {
-    group: 'testapp.coreos.com',
-    version: 'correct-version',
-    versions: [testCRDInvalidVersion, testCRDInvalidVersionTwo],
-    names: {
-      kind: 'TestResource',
-      plural: 'testresources',
-      singular: 'testresource',
-      listKind: 'TestResourceList',
-    },
-  },
-};
-
-export const testForNoVersionsCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
-  kind: 'CustomResourceDefinition',
-  metadata: {
-    name: 'testresources.testapp.coreos.com',
-  },
-  spec: {
-    group: 'testapp.coreos.com',
-    version: 'correct-version',
-    names: {
-      kind: 'TestResource',
-      plural: 'testresources',
-      singular: 'testresource',
-      listKind: 'TestResourceList',
-    },
+    scope: 'Namespaced',
   },
 };
--- a/frontend/__tests__/module/k8s/k8s-models.spec.ts
+++ b/frontend/__tests__/module/k8s/k8s-models.spec.ts
@@ -20,10 +20,8 @@ import {
   testCRDVersionV2Alpha1,
   testCRDVersionV3Beta1,
   testCRDVersionV1,
-  testForNoVersionsCRD,
   testForValidVersionsCRD,
   testForUnservedVersionsCRD,
-  testForInvalidVersionsCRD,
 } from '../../../__mocks__/crds';
 import { testNamespace, testOwnedResourceInstance } from '../../../__mocks__/k8sResourcesMocks';
 import {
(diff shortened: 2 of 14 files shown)


rhamilto:
/hold

for broken merge queue

rhamilto:
/hold

OLM descriptors.scenario.ts is broken, and I'm struggling with how to fix it.

TheRealJon:
@rhamilto Taking a look

frontend/__mocks__/crds.ts
```diff
@@ -15,16 +15,23 @@ export const testCRD: CustomResourceDefinitionKind = {
   },
   spec: {
     group: 'testapp.coreos.com',
-    version: 'v1alpha1',
-    validation: {
-      openAPIV3Schema: {},
-    },
+    versions: [
+      {
+        name: 'v1',
```
TheRealJon, line 20: I believe this can remain 'v1alpha1' since it's actually the version of the custom resource, not the CRD api version.

frontend/packages/operator-lifecycle-manager/mocks.ts
```diff
@@ -367,7 +367,16 @@ export const testCRD: CustomResourceDefinitionKind = {
   },
   spec: {
     group: 'testapp.coreos.com',
-    version: 'v1alpha1',
+    versions: [
+      {
+        name: 'v1',
```
TheRealJon, line 372:
```suggestion
        name: 'v1alpha1',
```

frontend/packages/operator-lifecycle-manager/mocks.ts
```diff
@@ -367,7 +367,16 @@ export const testCRD: CustomResourceDefinitionKind = {
...
+        served: true,
+        storage: true,
+        schema: {
+          openAPIV3Schema: {},
+        },
+      },
+    ],
     validation: {
       openAPIV3Schema: {},
     },
```
TheRealJon, line 382:
```suggestion
```

frontend/__tests__/module/k8s/k8s-models.spec.ts
```diff
@@ -47,7 +47,7 @@ describe('referenceFor', () => {
 
 describe('referenceForCRD', () => {
   it('returns a reference for custom resource definitions', () => {
-    expect(referenceForCRD(testCRD)).toEqual('testapp.coreos.com~v1alpha1~TestResource');
+    expect(referenceForCRD(testCRD)).toEqual('testapp.coreos.com~v1~TestResource');
```
TheRealJon, line 50: If you revert the change to `testCRD.spec.versions[0].name`, you can also revert this change:
```suggestion
    expect(referenceForCRD(testCRD)).toEqual('testapp.coreos.com~v1alpha1~TestResource');
```

frontend/public/components/custom-resource-definition.tsx
```diff
@@ -231,7 +231,7 @@ const Details: React.FC<{ obj: CustomResourceDefinitionKind }> = ({ obj: crd })
                   <Established crd={crd} />
                 </dd>
                 <DetailsItem label="Group" obj={crd} path="spec.group" />
-                <DetailsItem label="Version" obj={crd} path="spec.version" />
+                <DetailsItem label="Version" obj={crd} path="spec.versions[0].name" />
```
TheRealJon, line 234: Should this be updated to handle multiple versions better?

TheRealJon: Or maybe the column heading should be "Latest Version" if we just show the first item in the versions array.

frontend/__mocks__/crds.ts
```diff
@@ -16,15 +16,13 @@ export const testCRD: CustomResourceDefinitionKind = {
   spec: {
     group: 'testapp.coreos.com',
     version: 'v1alpha1',
```
TheRealJon, line 18: I think if we update the CRD API version to v1, we have to update this to use `versions` instead of `version`. The version property is removed in v1.
```suggestion
    versions: [
      {
        name: 'v1alpha1',
        storage: true,
        served: true,
        schema: {
          openAPIV3Schema: {}
        }
```

frontend/__mocks__/crds.ts
```diff
@@ -144,23 +147,26 @@ export const testForInvalidVersionsCRD: CustomResourceDefinitionKind = {
...
 export const testForNoVersionsCRD: CustomResourceDefinitionKind = {
-  apiVersion: 'apiextensions.k8s.io/v1beta1',
+  apiVersion: 'apiextensions.k8s.io/v1',
   kind: 'CustomResourceDefinition',
   metadata: {
     name: 'testresources.testapp.coreos.com',
   },
   spec: {
     group: 'testapp.coreos.com',
     version: 'correct-version',
```
TheRealJon, line 162: `version` is not a valid field for API version `apiextensions.k8s.io/v1`. We will also likely need to update the test or tests that use this object.
```suggestion
```

frontend/public/module/k8s/types.ts
```diff
@@ -483,25 +483,29 @@ export type CRDVersion = {
...
       shortNames?: string[];
     };
-    scope?: 'Namespaced';
+    scope: 'Cluster' | 'Namespaced';
     validation?: {
       // NOTE: Actually a subset of JSONSchema, but using this type for convenience
       openAPIV3Schema: JSONSchema6;
     };
-    versions?: CRDVersion[];
   };
```
TheRealJon, line 509: Might need some input from @spadgett on this. This type is structured to support both v1 and v1beta1 because they are both supported in Kubernetes v1.18, but I'm not sure if that's necessary or not.

frontend/public/components/custom-resource-definition.tsx
rhamilto: Updated col head to `Latest Version`

frontend/public/models/yaml-templates.ts
```diff
@@ -289,28 +289,46 @@ spec:
...
-    singular: crontab
-    # kind is normally the CamelCased singular type. Your resource manifests use this.
-    kind: CronTab
-    # shortNames allow shorter string to match your resource on the CLI
-    shortNames:
-    - ct
+    apiVersion: apiextensions.k8s.io/v1
+    kind: CustomResourceDefinition
+    metadata:
+      # name must match the spec fields below, and be in the form: <plural>.<group>
```
spadgett, line 295: The comments are nice, but I suspect they get stripped out when we insert the namespace into the YAML. OK to leave in case we fix that bug.

frontend/public/components/custom-resource-definition.tsx
spadgett, line 234: I don't think we can assume the list is already sorted. We probably have to call `getLatestVersionForCRD`. We probably want `useMemo` to avoid resorting each array on every render.

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
```diff
@@ -472,7 +472,8 @@ export const OperandDetails = connectToModel(({ crd, csv, kindObj, obj }: Operan
   const { kind, status } = obj;
   const [errorMessage, setErrorMessage] = React.useState(null);
   const handleError = (err: Error) => setErrorMessage(err.message);
-  const schema = crd?.spec?.validation?.openAPIV3Schema ?? (definitionFor(kindObj) as JSONSchema6);
+  const schema =
+    crd?.spec?.versions[0]?.schema?.openAPIV3Schema ?? (definitionFor(kindObj) as JSONSchema6);
```
spadgett, line 476: I'd think we need to search for the specific version we need here, @TheRealJon ?

frontend/__mocks__/crds.ts
```diff
@@ -68,6 +75,9 @@ export const testCRDVersionV1: CRDVersion = {
   name: 'v1',
   storage: true,
   served: true,
+  schema: {
```
rhamilto, line 78: Not sure if `schema` is optional or not (I specified it is it [types.ts](https://github.com/openshift/console/pull/6302/files#diff-26cde32482ccee5968a1f4d2d69ffe03R486)).  @TheRealJon or @spadgett, do you know?

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.spec.tsx
```diff
@@ -140,7 +140,7 @@ xdescribe('[https://issues.redhat.com/browse/CONSOLE-2136] CreateOperandForm', (
         model={testModel}
         providedAPI={testClusterServiceVersion.spec.customresourcedefinitions.owned[0]}
         csv={testClusterServiceVersion}
-        schema={testCRD.spec.validation.openAPIV3Schema}
+        schema={testCRD.spec.versions[0].schema.openAPIV3Schema}
```
spadgett, line 143: Do we need to search for a specific version?

frontend/__mocks__/crds.ts
rhamilto: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinitionversion-v1-apiextensions-k8s-io sure would be helpful if it specified.

rhamilto: I believe `schema` is required, so I've changed it.

frontend/public/models/yaml-templates.ts
rhamilto: They do get stripped, but as with the other examples, this is a copy/paste from the Kube docs.

frontend/public/components/custom-resource-definition.tsx
rhamilto: I think this can be `<DetailsItem label="Version" obj={crd} path={getLatestVersionForCRD(crd)} />` since it's the details page and not the list?  We may want `useMemo` in `<TableData className={tableColumnClasses[2]}>{getLatestVersionForCRD(crd)}</TableData>` on line 205?

frontend/public/components/custom-resource-definition.tsx
```diff
@@ -82,8 +82,8 @@ const CRDTableHeader = () => {
       props: { className: tableColumnClasses[1] },
     },
     {
-      title: 'Version',
-      sortField: 'spec.version',
+      title: 'Latest Version',
+      sortField: 'spec.versions[0]',
```
rhamilto, line 86: This is probably incorrect as it assumes the list is already sorted.

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.spec.tsx
TheRealJon: I think in this case, the testCRD is hard coded with only one version

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
TheRealJon: Yeah, @rhamilto We already have logic that finds the correct owned/required CRD in the csv, so we just need to destructure the `version` field on L479 as well. Then you can move this declaration below that one, and use `version` to find the appropriate schema in `crd.spec.versions`.

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
```diff
@@ -472,7 +472,8 @@ export const OperandDetails = connectToModel(({ crd, csv, kindObj, obj }: Operan
...
 
   // Find the matching CRD spec for the kind of this resource in the CSV.
   const { displayName, specDescriptors, statusDescriptors } =
```
TheRealJon, line 479:
```suggestion
  const { displayName, specDescriptors, statusDescriptors, version } =
```

frontend/public/components/custom-resource-definition.tsx
TheRealJon: Yeah, the API says this field is _**used**_ to sort versions, but not that it _**is**_ sorted

frontend/public/module/k8s/types.ts
```diff
@@ -483,25 +483,24 @@ export type CRDVersion = {
...
   served: boolean;
   storage: boolean;
+  schema: {
+    // NOTE: Actually a subset of JSONSchema, but using this type for convenience
+    openAPIV3Schema: JSONSchema6;
+  };
 };
 
 export type CustomResourceDefinitionKind = {
   spec: {
```
TheRealJon, line 493: nit: alpha

frontend/public/components/custom-resource-definition.tsx
TheRealJon: > I think this can be `<DetailsItem label="Version" obj={crd} path={getLatestVersionForCRD(crd)} />` since it's the details page and not the list? We may want `useMemo` in `<TableData className={tableColumnClasses[2]}>{getLatestVersionForCRD(crd)}</TableData>` on line 205?

`<DetailsItem label="Version" obj={crd} path={getLatestVersionForCRD(crd)} />` won't work as expected. The `path` is a path string which points to the object property we want to display. We can omit path and just do `<DetailsItem label="Latest Version" obj={crd}>{getLatestVersionForCRD(crd)}</DetailsItem>`

rhamilto:
@TheRealJon, @spadgett I've taken a swing at addressing feedback.

Notable fixes:
* [get correct operand schema](https://github.com/openshift/console/pull/6302/commits/53be07cb41415f5198c9bd713299dabf4feb0a9b#diff-6c9977d579b6b822a24efbc27cfeb084R483-R486)
* [add sortFunc to CRD Latest Version col](https://github.com/openshift/console/pull/6302/commits/53be07cb41415f5198c9bd713299dabf4feb0a9b#diff-7a6938f70a1ec48994c8cfb5433438d2R86)
* [change CRD details `Version` to `Latest Version`](https://github.com/openshift/console/pull/6302/commits/53be07cb41415f5198c9bd713299dabf4feb0a9b#diff-7a6938f70a1ec48994c8cfb5433438d2R234-R235)

PR #5481, opened 2020-05-18
Bug 1835348: Replace hideNameFilter with hideToolbar

bipuladh:
- `hideNameFilter` and `hideLabelFilter` should stop label and name chips from being rendered. 
Before:
Screenshot from 2020-05-18 21-14-49
After:
Screenshot from 2020-05-18 21-12-05
cc @zherman0 
/assign @spadgett

PR #5081, opened 2020-04-16
Update PatternFly to major release v4.0

dtaylor113:
### Overview
This WIP PR contains OpenShift console code changes required to update to the PatternFly v4 RC2!  Thanks to @redallen, @jschuler, and everyone on the PatternFly team for making the transition as easy as possible.

Please note that this PR touches many, many files across plug-ins and dev console.

**More work is needed to address the visual changes introduced**

image

### Additional PF v4 Upgrade Resources
- [PF-Core (HTMl, css) Upgrade Guide](https://patternfly-pr-3139.surge.sh/documentation/core/overview/upgrade-guide)
- [PF-Core: breaking-change-notes](https://github.com/redallen/patternfly-scripts/blob/master/github/patternfly/patternfly/pull_requests/breaking-change-notes.md)
- [PF-React Upgrade Guide](https://patternfly-react-pr-4298.surge.sh/documentation/react/overview/upgrade-guide)
- [PF-React: breaking-change-notes](https://github.com/redallen/patternfly-scripts/blob/master/github/patternfly/patternfly-react/pull_requests/breaking-change-notes.md)
- [Victory Chart fixes for RC1](https://github.com/patternfly/patternfly-react/pull/4330)

PR #6304, opened 2020-08-11
Bug 1856593: Change from hideNameFilter flag to hideToolbar

zherman0:
The `hideNameFilter` was removed and replaced with `hideToolbar`.  The search page and the FilterToolbar component were not updated to handle the new flag name so the Name filter was showing on the search page when it did not use to. 
EDIT: @spadgett @bipuladh  I am confused what happened here.  I reviewed Bipul's [changes](https://github.com/openshift/console/pull/5481) when he switched this to the `hideToolbar` flag but the FilterToolbar code was reverted at some point I assume because of needing just the `hideLabelFilter`. Anyone have more detail and suggestions on what we should name the flag?

Screenshot_2020-08-11 Search · OKD

### DIFF
--- a/frontend/public/components/filter-toolbar.tsx
+++ b/frontend/public/components/filter-toolbar.tsx
@@ -84,7 +84,7 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
   const {
     rowFilters = [],
     data,
-    hideNameFilter,
+    hideToolbar,
     columnLayout,
     hideLabelFilter,
     location,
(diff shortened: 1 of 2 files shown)


zherman0:
I am not sure why the name was changed, but the change does not make sense inside the FilterToolbar component since it only affects the Name filed and not the entire toolbar.  However, I did not want possibly overwrite anything that is being reworked - so I simply updated the flag.

zherman0:
EDIT: @spadgett @bipuladh  I am confused what happened here.  I reviewed Bipul's [changes](https://github.com/openshift/console/pull/5481) when he switched this to the `hideToolbar` flag but the FilterToolbar code was reverted at some point I assume because of needing just the `hideLabelFilter`. Anyone have more detail and suggestions on what we should name the flag?

bipuladh:
https://github.com/openshift/console/pull/5081/files#diff-250131ae4d03b64ce9bae143cec0b46aR83 made the change to `hideNameFilter`. @dtaylor113 might be able to help us. :)

spadgett:
> https://github.com/openshift/console/pull/5081/files#diff-250131ae4d03b64ce9bae143cec0b46aR83 made the change to `hideNameFilter`. @dtaylor113 might be able to help us. :)

Bad rebase maybe?

PR #6306, opened 2020-08-11
Bug 1868150: Add click-able links to notifications

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1868150

Use linkify to allow clickable links in the body of a notification description in the notification drawer:

notification-link

### DIFF
--- a/frontend/public/components/notification-drawer.tsx
+++ b/frontend/public/components/notification-drawer.tsx
@@ -46,6 +46,7 @@ import {
 import { ClusterVersionModel } from '../models';
 import { useK8sWatchResource, WatchK8sResource } from './utils/k8s-watch-hook';
 import { useAccessReview } from './utils/rbac';
+import { LinkifyExternal } from './utils';
 
 const criticalCompare = (a: Alert): boolean => getAlertSeverity(a) === 'critical';
 const otherAlertCompare = (a: Alert): boolean => getAlertSeverity(a) !== 'critical';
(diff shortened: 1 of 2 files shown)


frontend/public/components/notification-drawer.tsx
```diff
@@ -106,7 +107,9 @@ const getAlertNotificationEntries = (
           return (
             <NotificationEntry
               key={`${i}_${alert.activeAt}`}
-              description={getAlertDescription(alert) || getAlertMessage(alert)}
+              description={
+                <Linkify>{getAlertDescription(alert) || getAlertMessage(alert)}</Linkify>
```
spadgett, line 111: Use `LinkifyExternal` instead.

jcaianirh: @spadgett , so i played around with this, and i'm not sure its the behavior that we want. Maybe i'm wrong, but the link in the slack convo that created this pr is an internal link (image below). Are we expecting internal links here or external. While i agree if these links point us outside of the app we should use external, but otherwise, we have to re authenticate and open a new window. It seems more seamless not to use the external option. Thoughts on this?

image

spadgett: @jcaianirh I'd expect these to almost always be external links. This one is the exception imo. We could have special handling if the link starts with the console, but that gets a bit involved. I'm not sure it's worth it.

spadgett:
@beanh66 Do you think we should keep our existing notification for cluster upgrades? We have an alert for this now, so it's showing up twice.

frontend/public/components/notification-drawer.tsx
spadgett: Maybe we can do something special for `UpdateAvailable` alerts.

jcaianirh: @spadgett ok, sounds good. I'll try that. Thanks for the feedback!

frontend/public/components/notification-drawer.tsx
```diff
@@ -90,6 +92,16 @@ export const getAlertActions = (actionsExtensions: AlertAction[]) => {
   return alertActions;
 };
 
+const AlertDescription: React.FC<AlertDescriptionProps> = ({ alert }) => {
+  if (
+    getAlertName(alert) === 'UpdateAvailable' ||
+    getAlertName(alert) === 'CannotRetrieveUpdates'
+  ) {
+    return <Linkify>{getAlertDescription(alert) || getAlertMessage(alert)}</Linkify>;
```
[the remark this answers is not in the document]
jcaianirh: removed special case for internal links

PR #6311, opened 2020-08-12
Removes `ceph` from the error message for pool creation

a2batic:


frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/storage-efficiency-card/storage-efficiency-card.tsx
```diff
@@ -59,18 +59,18 @@ const StorageEfficiencyCard: React.FC<DashboardItemProps> = ({
...
     const savingsValue = Number(saved);
     const savedBytes = humanizeBinaryBytes(savingsValue).string;
-    return savingsValue === 0 ? 'No Savings' : savedBytes;
+    return savedBytes;
   };
 
   const compressionRatioProps = {
     isLoading: !poolCapacityRatioResult,
-    error: !!poolCapacityRatioResultError || !ratio,
+    error: !!poolCapacityRatioResultError || !ratio || Number(ratio) === 0,
```
afreen23, line 73: Instead you might want to add this as a condition in `ItemsBody` alongwith `error` .
Since 0 stats is not technically an error.

a2batic: Every Item can have its own reason to have `error = true`. It's true here for both the items, but we plan to have more items in the card in future releases, so IMO we shouldn't generalise it. wdyt ?

afreen23: The same piece of check is required in noobaa reduction card.
It will work for all four use cases currently.

afreen23:
/approve 
https://github.com/openshift/console/pull/6311#discussion_r471255682 can be resolved in later PR as the fix required is more.

PR #6312, opened 2020-08-12
remove the explicit chain method in channel selection

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-4494

Problem: Explicit lodash chain method breaks the UI with the error

Solution: Removed the explict chain method and changed the logic to filter the available channels
image

/assign @invincibleJai 
/kind bug


invincibleJai:
/lgtm
/approve

verified on cluster bot https://console-openshift-console.apps.ci-ln-v58jrdb-f76d1.origin-ci-int-gce.dev.openshift.com/channel/ns/karthik

PR #6313, opened 2020-08-12
Bug 1873140: fix Alert State filter for dev perspective Alerts

vikram-raj:
Fix - https://issues.redhat.com/browse/ODC-4424

**Screenshots**
Kapture 2020-09-03 at 20 17 13

**Test coverage**
Screenshot 2020-08-27 at 6 45 29 PM

### DIFF
--- a/frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
@@ -1,4 +1,4 @@
-import { RuleStates } from '@console/internal/components/monitoring/types';
+import { AlertStates } from '@console/internal/components/monitoring/types';
 import { getAlertsAndRules } from '@console/internal/components/monitoring/utils';
 import { rules } from '@console/shared/src/utils/__mocks__/alerts-and-rules-data';
 import { monitoringAlertRows } from '../monitoring-alerts-utils';
@@ -16,8 +16,14 @@ describe('monitoring-alerts-utils', () => {
   });
   it('row should be collapse if rule state is not FIRING', () => {
     const alertRules = getAlertsAndRules(rules?.data).rules;
-    alertRules[0].state = RuleStates.Inactive;
+    alertRules[1].alerts[0].state = AlertStates.NotFiring;
     const rows = monitoringAlertRows(alertRules, [''], 'ns');
-    expect(rows[0].isOpen).toBe(undefined);
+    expect(rows[1].isOpen).toBe(undefined);
+    alertRules[0].alerts[0].state = AlertStates.Pending;
+    const rows1 = monitoringAlertRows(alertRules, [''], 'ns');
+    expect(rows1[0].isOpen).toBe(false);
+    alertRules[0].alerts[0].state = AlertStates.Firing;
+    const rows2 = monitoringAlertRows(alertRules, [''], 'ns');
+    expect(rows2[0].isOpen).toBe(true);
   });
 });
--- a/frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
@@ -13,9 +13,10 @@ import {
   StateCounts,
   Severity,
   AlertState,
-  alertsRowFilters,
   severityRowFilter,
+  alertStateFilter,
 } from '@console/internal/components/monitoring/alerting';
+import { RowFilter } from '@console/internal/components/filter-toolbar';
 import { Kebab } from '@console/internal/components/utils';
 import {
   alertDescription,
@@ -23,7 +24,12 @@ import {
   alertSeverityOrder,
   alertingRuleStateOrder,
 } from '@console/internal/reducers/monitoring';
-import { Alert, Rule, RuleStates } from '@console/internal/components/monitoring/types';
+import {
+  Alert,
+  Rule,
+  RuleStates,
+  AlertStates,
+} from '@console/internal/components/monitoring/types';
 import { YellowExclamationTriangleIcon } from '@console/shared';
 import { labelsToParams } from '@console/internal/components/monitoring/utils';
 import SilenceAlert from './SilenceAlert';
@@ -77,22 +83,22 @@ export const monitoringAlertRows = (
 ) => {
   const rows = [];
   _.forEach(alertrules, (rls) => {
+    const states = _.map(rls.alerts, (a) => a.state);
     rows.push({
-      ...(rls.state !== RuleStates.Inactive && {
+      ...(!_.isEmpty(states) && {
         isOpen:
-          (rls.state === RuleStates.Firing && !_.includes(collapsedRowsIds, rls.id)) ||
-          (rls.state !== RuleStates.Firing && _.includes(collapsedRowsIds, rls.id)),
+          (_.includes(states, AlertStates.Firing) && !_.includes(collapsedRowsIds, rls.id)) ||
+          (!_.includes(states, AlertStates.Firing) && _.includes(collapsedRowsIds, rls.id)),
       }),
       cells: [
         {
-          title:
-            rls.state !== 'firing' ? (
-              rls.name
-            ) : (
-              <>
-                <YellowExclamationTriangleIcon /> {rls.name}
-              </>
-            ),
+          title: !_.includes(states, AlertStates.Firing) ? (
+            rls.name
+          ) : (
+            <>
+              <YellowExclamationTriangleIcon /> {rls.name}
+            </>
+          ),
           id: rls.id,
         },
         {
@@ -145,15 +151,7 @@ export const monitoringAlertRows = (
   return rows;
 };
 
-export const alertFilters = [
-  {
-    filterGroupName: 'Alert State',
-    type: 'alert-state',
-    reducer: alertState,
-    items: [...alertsRowFilters[0].items, ...[{ id: 'inactive', title: 'Inactive' }]],
-  },
-  severityRowFilter,
-];
+export const alertFilters: RowFilter[] = [alertStateFilter, severityRowFilter];
 
 const setOrderBy = (orderBy: SortByDirection, data: Rule[]): Rule[] => {
   return orderBy === SortByDirection.asc ? data : data.reverse();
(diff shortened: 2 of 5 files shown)


kyoto:
I realized that we should be making a similar change to the admin perspective too. I will take a look into that and consider if the two perspectives can share code for this fix.

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -145,11 +151,41 @@ export const monitoringAlertRows = (
...
+        filteredObjects = _.filter(filteredObjects, (o) => filter(value, o));
+      }
+    }
+  });
+
+  return filteredObjects;
+};
+
+const alertStateReducer = (rule: Rule) => {
+  return _.map(rule.alerts, (a) => a.state)[0] || 'inactive';
```
kyoto, line 181: Why only take the state of the first alert? Don't we need to iterate through all alerts looking for any alert that has a matching state?

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -145,11 +151,41 @@ export const monitoringAlertRows = (
...
+  _.each(_filters, (value, name) => {
+    const filter = allTableFilters[name];
+    if (_.isFunction(filter)) {
+      if (name === 'alert-state') {
+        filteredObjects = _.filter(filteredObjects, (o) =>
+          !_.isEmpty(value.selected)
+            ? _.filter(o.alerts, (a) => filter(value, a)).length > 0 ||
+              (value.selected.has('inactive') && _.isEmpty(o.alerts))
+            : o,
+        );
```
jerolimov, line 170: Maybe its just for myself to understand this.. If none filter is selected, you use the o (entry in filteredObjects) as filter value. Isn't it easier and cleaner to wrap this in an if, in the case we want keep 'all' filteredObjects? WDYT? :thinking: 

```jsx
    if (_.isFunction(filter)) {
      if (name === 'alert-state') {
        if (_.isEmpty(value.selected)) {
          filteredObjects = _.filter(filteredObjects, (o) =>
            _.filter(o.alerts, (a) => filter(value, a)).length > 0 ||
              (value.selected.has('inactive') && _.isEmpty(o.alerts)),
          );
        }
      } else {
        filteredObjects = _.filter(filteredObjects, (o) => filter(value, o));
      }
    }
```

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
jerolimov: Or if not, you could just write rule.alerts[0]?.state, but I think @kyoto question here is valid

jerolimov:
Could verify the fix on a cluster setup. Thanks @vikram-raj for providing some help and yaml files here. Will get a lgtm when you response to the open questions :smirk:

kyoto:
Opened a similar change for the admin perspective: https://github.com/openshift/console/pull/6460. The solutions will need to be different because the dev perspective combines rule rows and alert rows in the same table, but the admin perspective does not.

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
vikram-raj: I am mapping through all the alerts of the rule and it returns an array of a single element that is why I am taking the first element.

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
vikram-raj: Agree, Thanks for the suggestion.

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
jerolimov: But why mapping first over all elements and converting alerts to states, and then picking the first state. You could also pick the first alert state directly, or?

```js
  return rule.alerts[0].state || 'inactive';
```

I did not understand if we could reduce the alerts in any way instead of picking just the first one. /cc @kyoto

vikram-raj:
/retitle Bug 1873140: fix Alert State filter for dev perspective Alerts

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
kyoto: It looks to me like this won't work when a rule has multiple active alerts with different states. For example, `_.map(rule.alerts, (a) => a.state)` could return `['firing', 'pending']`, and then this would ignore the `pending` alert.

frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
```diff
@@ -16,8 +16,8 @@ describe('monitoring-alerts-utils', () => {
   });
   it('row should be collapse if rule state is not FIRING', () => {
     const alertRules = getAlertsAndRules(rules?.data).rules;
-    alertRules[0].state = RuleStates.Inactive;
+    alertRules[1].alerts[0].state = AlertStates.Pending;
```
jerolimov, line 19: Can you keep the tests for both rules here?

```jsx
    alertRules[0].alerts[0].state = AlertStates.Inactive; // or what ever
    alertRules[1].alerts[0].state = AlertStates.Pending;
```

frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
```diff
@@ -16,8 +16,8 @@ describe('monitoring-alerts-utils', () => {
...
     const rows = monitoringAlertRows(alertRules, [''], 'ns');
-    expect(rows[0].isOpen).toBe(undefined);
+    expect(rows[1].isOpen).toBe(undefined);
```
jerolimov, line 21: Same here? Why this changed? :thinking:

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1307,19 +1307,21 @@ const ruleHasAlertState = (rule: Rule, state: AlertStates): boolean =>
 const ruleAlertStateFilter = (filter, rule: Rule) =>
   _.some(rule.alerts, (a) => filter.selected.has(a.state)) || _.isEmpty(filter.selected);
 
+export const AlertStateFilter: RowFilter = {
```
jerolimov, line 1310: Please use lowercase a for variables (which are not a react component)

```suggestion
export const alertStateFilter: RowFilter = {
```

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -145,15 +152,35 @@ export const monitoringAlertRows = (
...
-export const alertFilters = [
-  {
-    filterGroupName: 'Alert State',
-    type: 'alert-state',
-    reducer: alertState,
-    items: [...alertsRowFilters[0].items, ...[{ id: 'inactive', title: 'Inactive' }]],
-  },
-  severityRowFilter,
-];
+export const getFilteredRows = (_filters, rowFilters, objects) => {
```
jerolimov, line 155:
```suggestion
export const getFilteredRows = (_filters, rowFilters, rules: Rule[]) => {
```

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -145,15 +152,35 @@ export const monitoringAlertRows = (
...
+  if (_.isEmpty(_filters)) {
+    return objects;
+  }
+
+  const allTableFilters = getAllTableFilters(rowFilters);
+  let filteredObjects = objects;
```
jerolimov, line 161:
```suggestion
  let filteredRules = rules;
```

and so on..

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -145,15 +152,35 @@ export const monitoringAlertRows = (
...
+  _.each(_filters, (value, name) => {
+    const filter = allTableFilters[name];
+    if (_.isFunction(filter)) {
+      if (name === 'alert-state') {
+        if (!_.isEmpty(value.selected)) {
+          filteredObjects = _.filter(
+            filteredObjects,
+            (o) =>
+              _.filter(o.alerts, (a) => filter(value, a)).length > 0 ||
```
jerolimov, line 170:
```suggestion
              _.some(o.alerts, (a) => filter(value, a)) ||
```

wdyt?

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
jerolimov: Great final solution with `inMatch`. Nice work @vikram-raj and @kyoto

frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
vikram-raj: Added the tests for other states as well.

frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/monitoring-alerts-utils.spec.tsx
vikram-raj: handle the test for rows 0 as well.

jerolimov (approved):
Tested it locally and on an cluster (4.6.0-0.ci.test-2020-09-07-093240-ci-ln-fl8bwh2), works as expected. :+1: 

/lgtm

frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
vikram-raj: This function is not required. So, removed it.

jerolimov (approved):
Tested it again on 4.6.0-0.ci.test-2020-09-07-115219-ci-ln-qgpzqit

Thanks @vikram-raj for the support here :)

/lgtm

PR #6315, opened 2020-08-12
Bug 1868394: Fix boot-order change alert when VM is running

irosenzw:
Show the alert when the disk order is changing in the yaml
while the VM is running.

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/selectors/vm-like/next-run-changes.ts
+++ b/frontend/packages/kubevirt-plugin/src/selectors/vm-like/next-run-changes.ts
@@ -84,7 +84,7 @@ export const isBootOrderChanged = (vm: VMWrapper, vmi: VMIWrapper): boolean => {
     const vmDevices = getDevices(vm.asResource());
     const vmiDevices = getVMIDevices(vmi.asResource());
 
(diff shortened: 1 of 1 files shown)


frontend/packages/kubevirt-plugin/src/selectors/vm-like/next-run-changes.ts
```diff
@@ -84,7 +84,9 @@ export const isBootOrderChanged = (vm: VMWrapper, vmi: VMIWrapper): boolean => {
     const vmDevices = getDevices(vm.asResource());
     const vmiDevices = getVMIDevices(vmi.asResource());
 
-    return vmDevices.every((bootableDevice, index) => _.isEqual(bootableDevice, vmiDevices[index]));
+    return !vmDevices.every((bootableDevice, index) =>
```
glekner, line 87: why not
 `vmDevices.some((bootableDevice, index) => _.isEqual(bootableDevice, vmiDevices[index])`

irosenzw: you mean:
`vmDevices.some((bootableDevice, index) => !_.isEqual(bootableDevice, vmiDevices[index])` ?

PR #6316, opened 2020-08-12
Bug 1868426: Unfilter OS dropdown in VM wizard when choosing flavor or workload

irosenzw:
When switching OS, the flavor and workload will be set to NULL
 thus the user would have to pick suitable flavor & workload
 for the new OS.
 
### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
@@ -90,59 +90,6 @@ const selectedUserTemplateUpdater = (options: UpdateOptions) => {
   prefillVmTemplateUpdater(options);
 };
 
-const provisioningSourceUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
-  const state = getState();
-  if (
-    !hasVmSettingsChanged(
-      prevState,
-      state,
-      id,
-      VMSettingsField.PROVISION_SOURCE_TYPE,
-      VMSettingsField.USER_TEMPLATE,
-    )
-  ) {
-    return;
-  }
-  const source = iGetProvisionSource(state, id);
-  const isContainer = source === ProvisionSource.CONTAINER;
-  const isUrl = source === ProvisionSource.URL;
-
-  dispatch(
-    vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
-      [VMSettingsField.CONTAINER_IMAGE]: {
-        isHidden: asHidden(!isContainer, VMSettingsField.PROVISION_SOURCE_TYPE),
-      },
-      [VMSettingsField.IMAGE_URL]: {
-        isHidden: asHidden(!isUrl, VMSettingsField.PROVISION_SOURCE_TYPE),
-      },
-    }),
-  );
-};
-
-const flavorUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
-  const state = getState();
-  if (!hasVmSettingsChanged(prevState, state, id, VMSettingsField.FLAVOR)) {
-    return;
-  }
-  const flavor = iGetVmSettingValue(state, id, VMSettingsField.FLAVOR);
-
-  const isHidden = asHidden(flavor !== CUSTOM_FLAVOR, VMSettingsField.FLAVOR);
-  const isRequired = asRequired(flavor === CUSTOM_FLAVOR, VMSettingsField.FLAVOR);
-
-  dispatch(
-    vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
-      [VMSettingsField.MEMORY]: {
-        isHidden,
-        isRequired,
-      },
-      [VMSettingsField.CPU]: {
-        isHidden,
-        isRequired,
-      },
-    }),
-  );
-};
-
 const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
   const state = getState();
   if (iGetCommonData(state, id, VMWizardProps.isProviderImport)) {
@@ -252,9 +199,17 @@ const cloneCommonBaseDiskImageUpdater = ({ id, prevState, dispatch, getState }: 
   );
 };
 
-const workloadConsistencyUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
+const templateConsistencyUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
   const state = getState();
-  if (!hasVMSettingsValueChanged(prevState, state, id, VMSettingsField.WORKLOAD_PROFILE)) {
+  if (
+    !hasVMSettingsValueChanged(
+      prevState,
+      state,
+      id,
+      VMSettingsField.WORKLOAD_PROFILE,
+      VMSettingsField.OPERATING_SYSTEM,
+    )
+  ) {
     return;
   }
   const selectors = iGetRelevantTemplateSelectors(state, id);
@@ -262,15 +217,44 @@ const workloadConsistencyUpdater = ({ id, prevState, dispatch, getState }: Updat
   const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
 
   if (!iGetRelevantTemplate(iUserTemplates, iCommonTemplates, selectors)) {
-    // reset workload profile if no relevant template found - could be triggered by provider prefil
+    // Reset workload and flavor profile if no relevant template found
+    // Could be triggered by provider prefil or os selection
     dispatch(
-      vmWizardInternalActions[InternalActionType.UpdateVmSettingsField](
-        id,
-        VMSettingsField.WORKLOAD_PROFILE,
-        { value: null },
-      ),
+      vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
+        [VMSettingsField.WORKLOAD_PROFILE]: { value: null },
+        [VMSettingsField.FLAVOR]: { value: null },
+      }),
     );
   }
+};
+
+const provisioningSourceUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
+  const state = getState();
+  if (
+    !hasVmSettingsChanged(
+      prevState,
+      state,
+      id,
+      VMSettingsField.PROVISION_SOURCE_TYPE,
+      VMSettingsField.USER_TEMPLATE,
+    )
+  ) {
+    return;
+  }
+  const source = iGetProvisionSource(state, id);
+  const isContainer = source === ProvisionSource.CONTAINER;
+  const isUrl = source === ProvisionSource.URL;
+
+  dispatch(
+    vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
+      [VMSettingsField.CONTAINER_IMAGE]: {
+        isHidden: asHidden(!isContainer, VMSettingsField.PROVISION_SOURCE_TYPE),
+      },
+      [VMSettingsField.IMAGE_URL]: {
+        isHidden: asHidden(!isUrl, VMSettingsField.PROVISION_SOURCE_TYPE),
+      },
+    }),
+  );
 };
 
 const nativeK8sUpdater = ({ id, dispatch, getState, changedCommonData }: UpdateOptions) => {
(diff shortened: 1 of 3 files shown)


frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -34,22 +34,13 @@ export const getFlavorLabel = (flavor: string) => {
...
-    templatesWithLabels = [
+    return getTemplateOperatingSystems([
       getTemplatesOfLabelType(templates, TEMPLATE_TYPE_VM).find((t) => getName(t) === userTemplate),
-    ];
-  } else {
-    templatesWithLabels = getTemplatesWithLabels(
-      getTemplatesOfLabelType(templates, TEMPLATE_TYPE_BASE),
-      [getWorkloadLabel(workload), getFlavorLabel(flavor)],
-    );
+    ]);
```
suomiy, line 41: why? 
this will cause to return the wrong OSs not reflecting the workload and flavor selection.

irosenzw: https://bugzilla.redhat.com/show_bug.cgi?id=1868426

yaacov: @suomiy hi, 
from UX side this is wrong to change the selection options of a field by using another field below it.

from a user perspective a common problem this can fix is:
a - a user selects an OS then selects a Flavor or Workload
b - a user want to fix their selection of OS, but the correct OS is missing from the list.

the fix:
a - a user select and OS
b - the user select some flavor
c - all OS options are presented and they can currect the OS selection.

suomiy: this is not about UX, but about correct template selection. The user should be aware of comon templates and create his own if such template is missing (based on os/flavor etc.)

disscussing more in BZ

yaacov: Note:
in the BZ discussion we decided that a viable fix for the problem wil be to:

> 1: override the user selection of flavor/workload upon OS selection

Ido, can we override the values to "null" (or what ever the default value is) ?
this will make sure users know that the values where changed ?

irosenzw: yeah good idea

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -156,11 +156,22 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
-      id,
-      VMSettingsField.MOUNT_WINDOWS_GUEST_TOOLS,
-      { isHidden: asHidden(!isWindows, VMSettingsField.OPERATING_SYSTEM), value: isWindows },
-    ),
+    vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
+      [VMSettingsField.MOUNT_WINDOWS_GUEST_TOOLS]: {
+        isHidden: asHidden(!isWindows, VMSettingsField.OPERATING_SYSTEM),
+        value: isWindows,
+      },
+      [VMSettingsField.WORKLOAD_PROFILE]: { value: null },
```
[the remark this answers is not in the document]
suomiy: additional idea: we could also show a small warning to the user that a combination is invalid and reset the warning on any new value selection

thoughts?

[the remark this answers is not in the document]
irosenzw: I agree with the first comment.
for the warning I think it's redundant. when following the first comment behavior, it is clear that if the workload/flavor would set to null it is because the combination is invalid, thus no need for any warning message.

what do you think @matthewcarleton  ?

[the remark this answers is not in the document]
yaacov: Note:
two os may have the same flavor label, but it may mean different things, for example:
flavor: tiny os:ubuntu => mem: 2Gi
flavor: tiny os:fedora => mem: 1Gi

IMHO, not re-seting the flavor each time we change an os is not nice to users 

https://github.com/kubevirt/common-templates/blob/e127881c7fb7f77ba84f0571478d4507e3e6b8e7/generate-templates.yaml#L204
https://github.com/kubevirt/common-templates/blob/e127881c7fb7f77ba84f0571478d4507e3e6b8e7/generate-templates.yaml#L162

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -159,9 +159,67 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
+      );
+    } else {
+      const keepWorkload = Object.assign({}, relevantOptions);
+      Object.assign(keepWorkload, { flavor: null });
+      // Check for a valid common-template that contain the chosen Workload
+      if (iGetRelevantTemplate(null, iCommonTemplates, keepWorkload)) {
+        dispatch(
+          vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
+            [VMSettingsField.FLAVOR]: { value: null },
+            [VMSettingsField.MEMORY]: {
```
suomiy, line 194: this is done by flavorUpdater. Let's move the flavorUpdater down if we need its functionality later

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -159,9 +159,67 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
+
+  const relevantOptions = iGetRelevantTemplateSelectors(state, id);
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iTemplate =
+    iCommonTemplates && iGetRelevantTemplate(null, iCommonTemplates, relevantOptions);
+
+  if (os && !iTemplate) {
+    const keepFlavor = Object.assign({}, relevantOptions);
+    Object.assign(keepFlavor, { workload: null });
+    // Check for a valid common-template that contain the chosen Flavor
```
suomiy, line 177: let's not try to find combinations which could work. We don't know if the user prefers to keep his flavor or workload so let's reset both if we have to reset one of them.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -159,9 +159,67 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
+            },
+            [VMSettingsField.CPU]: {
+              isHidden: asHidden(true, VMSettingsField.FLAVOR),
+              isRequired: asRequired(false, VMSettingsField.FLAVOR),
+            },
+          }),
+        );
+      } else {
+        // Reset both Workload and Flavor as no valid common-template has been found
+        dispatch(
```
suomiy, line 206: btw workloadConsistencyUpdater does a similar thing. It might be easier for understanding to have the same logic in one place for both

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
suomiy: why is it not nice? the user just sees the label tiny and probably does not know which template will be used, ie does not have an association with how much memory. And if he does know, it shouldn't matter either. The user is selecting a tiny flavor for a reason.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -294,13 +312,13 @@ export const updateVmSettingsState = (options: UpdateOptions) =>
...
     selectUserTemplateOnLoadedUpdater,
     selectedUserTemplateUpdater,
-    flavorUpdater,
     osUpdater,
     baseImageUpdater,
     cloneCommonBaseDiskImageUpdater,
     workloadConsistencyUpdater,
     provisioningSourceUpdater,
     nativeK8sUpdater,
+    flavorUpdater,
```
suomiy, line 321: please move the code as well. We want to ensure that the updaters can be read from top to bottom in the same order they are executed in

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
yaacov: > why is it not nice? 

All the times I ordered a VM, I used numbers (or looked up doc to see the numbers) to specify the mem/cpu/disk I needed , I think that a user like me, will be confused if they choose 2Gb (tiny on ubuntu) and got 1Gb (tiny on fedora) while they should have chosen small because it's the 2Gb for fedora.

But I don't feel strongly about this, so just ignore if you feel differently, the bug is about having the ability to choose any OS, so any fix that will show all the OS options for any given flavour/workload is good for me :-)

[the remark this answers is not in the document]
suomiy: I know it is a bit confusing, I just meant it in the context of this PR. Sure if we plan to redesign this feature, we should tackle this problem in a separate effort. As this concerns the basic flow as well, and not just this fix.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -252,27 +199,58 @@ const cloneCommonBaseDiskImageUpdater = ({ id, prevState, dispatch, getState }:
...
-      ),
+      vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
+        [VMSettingsField.WORKLOAD_PROFILE]: { value: null },
+        [VMSettingsField.FLAVOR]: { value: null },
+      }),
     );
   }
 };
 
+const provisioningSourceUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
```
suomiy, line 225: +1 for moving this as well

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -252,27 +199,58 @@ const cloneCommonBaseDiskImageUpdater = ({ id, prevState, dispatch, getState }:
   );
 };
 
-const workloadConsistencyUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
+const templateConsistencyUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
   const state = getState();
-  if (!hasVMSettingsValueChanged(prevState, state, id, VMSettingsField.WORKLOAD_PROFILE)) {
+  if (
+    !hasVMSettingsValueChanged(prevState, state, id, VMSettingsField.WORKLOAD_PROFILE) &&
```
suomiy, line 205: this can be done in one call, look at the args of this func

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -252,27 +199,58 @@ const cloneCommonBaseDiskImageUpdater = ({ id, prevState, dispatch, getState }:
...
+  ) {
     return;
   }
   const selectors = iGetRelevantTemplateSelectors(state, id);
   const iUserTemplates = iGetLoadedCommonData(state, id, VMWizardProps.userTemplates);
   const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
 
   if (!iGetRelevantTemplate(iUserTemplates, iCommonTemplates, selectors)) {
-    // reset workload profile if no relevant template found - could be triggered by provider prefil
+    // reset workload and flavor profile if no relevant template found - could be triggered by provider prefil
```
suomiy, line 215:
```suggestion
    // reset workload and flavor profile if no relevant template found - could be triggered by provider prefil or os selection
```

PR #6317, opened 2020-08-12
Bug 1868353: v2v: do not let backend delete saved secrets

suomiy:
https://github.com/openshift/console/pull/6296 will cause backend to react to our secrets and mark them for deletion.

it should be ensured that secrets which were requested to be saved are preserved


suomiy:
@yaacov please take a look

PR #6318, opened 2020-08-12
Fetch HelmChartRepository instances on behalf of console service account      

pedjak:
With this change, non-admin users can browse through UI Helm charts available from the configured Helm repositories within the cluster      

Additional fixes/improvements:      
* Deserialze TLSConfig for correct secret 
* Refactored Helm chart repo model for better testability 
* More tests added


sbose78:
> For other users, we could document how to add requested roles. 

* Reading `HelmChartRepository` CRs should be trivial by providing `view` role on `HelmChartRepository` to `'system:authenticated'`. Console service account doesn't need any changes for that. Console backend can use user's token to read CRs.

* Issue is with reading the admin-configured credentials for private repos.

pedjak:
giving up on reading helm chart repositories on behalf of service account.

PR #6320, opened 2020-08-12
Bug 1866085: Ensure we not submit an invalid selector on label search

zherman0:
We have a bug where we do not verify that the selector string for the label search ends in alphanumeric (0-9A-z). This is updated for `key`,`!key`, `key==val`, and `key!=val`.

### DIFF
--- a/frontend/public/components/search-filter-dropdown.tsx
+++ b/frontend/public/components/search-filter-dropdown.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
+import { Dropdown, DropdownToggle, DropdownItem, Tooltip } from '@patternfly/react-core';
 import { CaretDownIcon, FilterIcon } from '@patternfly/react-icons';
 import { TextFilter } from './factory';
 
@@ -12,7 +12,7 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
   const [isOpen, setOpen] = React.useState(false);
   const [selected, setSelected] = React.useState(searchFilterValues.Label);
 
-  const { onChange, nameFilterInput, labelFilterInput } = props;
+  const { onChange, nameFilterInput, labelFilterInput, validated, validationMsg } = props;
 
   const onToggle = (open: boolean) => setOpen(open);
   const onSelect = (event: React.SyntheticEvent) => {
@@ -52,16 +52,23 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
         isOpen={isOpen}
         dropdownItems={dropdownItems}
       />
-      <TextFilter
-        parentClassName="co-search__filter-input"
-        onChange={handleInputValue}
-        placeholder={selected === searchFilterValues.Label ? 'app=frontend' : 'my-resource'}
-        name="search-filter-input"
-        id="search-filter-input"
-        value={selected === searchFilterValues.Label ? labelFilterInput : nameFilterInput}
-        onKeyDown={handleKeyDown}
-        aria-labelledby="toggle-id"
-      />
+      <Tooltip
+        content={validationMsg}
+        isVisible={!!validationMsg}
+        trigger={validationMsg ? 'mouseenter focus' : 'manual'}
+      >
+        <TextFilter
+          parentClassName="co-search__filter-input"
+          onChange={handleInputValue}
+          placeholder={selected === searchFilterValues.Label ? 'app=frontend' : 'my-resource'}
+          name="search-filter-input"
+          id="search-filter-input"
+          value={selected === searchFilterValues.Label ? labelFilterInput : nameFilterInput}
+          onKeyDown={handleKeyDown}
+          aria-labelledby="toggle-id"
+          validated={validated}
+        />
+      </Tooltip>
     </div>
   );
 };
@@ -70,4 +77,6 @@ export type SearchFilterDropdownProps = {
   onChange: (type: string, value: string, endOfString: boolean) => void;
   nameFilterInput: string;
   labelFilterInput: string;
+  validated?: string;
+  validationMsg?: string;
 };
--- a/frontend/public/module/k8s/selector-requirement.js
+++ b/frontend/public/module/k8s/selector-requirement.js
@@ -8,42 +8,40 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key) => /^(([A-Za-z0-9][0-9A-Za-z/\-_.]*)?[A-Za-z0-9])$/.test(key);
+
 export const requirementFromString = (string) => {
   string = string.trim();
 
   // "key"
   if (/^[0-9A-Za-z/\-_.]+$/.test(string)) {
-    return {
-      key: string,
-      operator: 'Exists',
-      values: [],
-    };
+    if (validateKey(string)) {
+      return { key: string, operator: 'Exists', values: [] };
+    }
   }
 
   // "!key"
   if (/^!\s*[0-9A-Za-z/\-_.]+$/.test(string)) {
-    return {
-      key: string.split(/!\s*/)[1],
-      operator: 'DoesNotExist',
-      values: [],
-    };
+    const key = string.split(/!\s*/)[1];
+    if (validateKey(key)) {
+      return { key, operator: 'DoesNotExist', values: [] };
+    }
   }
 
   // "key=value" OR "key==value"
   if (/^[0-9A-Za-z/\-_.]+\s*==?\s*[0-9A-Za-z/\-_.]+$/.test(string)) {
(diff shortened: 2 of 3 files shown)


frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,40 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key, operator, values) => {
+  // Need to validate key after parts split or else we don't take correct branch
+  if (/^(.*[0-9A-Za-z])$/.test(key)) {
+    return { key, operator, values };
+  }
+};
```
spadgett, line 16: This method is a bit overloaded. It should only validate the key, and the callers should be responsible for returning the object.

```suggestion
const validateKey = (key) => /^(.*[0-9A-Za-z])$/.test(key);
```

frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,40 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key, operator, values) => {
+  // Need to validate key after parts split or else we don't take correct branch
+  if (/^(.*[0-9A-Za-z])$/.test(key)) {
```
spadgett, line 13: I don't think this regex is correct. `.*` matches any character, but not every character is allowed in a label key.

frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,40 @@ export const createEquals = (key, value) => ({
...
   // "key!=value"
   if (/^[0-9A-Za-z/\-_.]+\s*!=\s*[0-9A-Za-z/\-_.]+$/.test(string)) {
-    return {
-      key: string.split(/\s*!=\s*/)[0],
-      operator: 'NotEquals',
-      values: [string.split(/\s*!=\s*/)[1]],
-    };
+    const parts = string.split(/\s*!=?\s*/);
+    const key = parts[0];
+    const value = parts[1];
```
spadgett, line 43:
```suggestion
    const [key, value] = string.split(/\s*!=?\s*/);
```

frontend/public/module/k8s/selector-requirement.js
spadgett: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

spadgett: Here is the error you get from the API server on an invalid label with the validation regex

> name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')" for field "metadata.labels".

spadgett:
@zherman0 Don't we want to show an error to the user to let them know the selector is not valid?

zherman0:
@spadgett @beanh66 @megan-hall  - Ok, I have 2 suggestions for displaying that a label input is invalid, the first is very robust but very obtrusive:
Screenshot_2020-09-03 Search · OKD(1)

Or something more subtle but not exactly telling for the user:
Screenshot_2020-09-03 Search · OKD(2)

frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,36 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key, operator, values) => {
+  // Need to validate key after parts split or else we don't take correct branch
```
zherman0, line 12: Remove this comment

megan-hall:
@zherman0 @spadgett @beanh66

In talking with @gdoyle1, we were wondering if it would make sense to use the same recommended behavior from Patternfly for the Search page as we do on pages like Pods. We use a type ahead and provide the labels in a dropdown below. This pattern doesn't allow users to submit an invalid label. 

Screen Shot 2020-09-03 at 4 14 21 PM

If we can't or don't want to repeat that behavior for some reason, I think it would be ok to add the error to the text input field. From a usability perspective, if would be even better (and probably expected) if we could surface how a user can resolve the error - possibly in a tooltip (as the form pattern for an invalid field where the message is below in red really does not suite the toolbar well). A few down sides to the tool tip would be: I am not sure how hard that would be able to do and I know that is something we often try to avoid for mobile reasons and screen readers as well. It would also be custom and outside of what we do in the console today. 

Screen Shot 2020-09-03 at 4 29 11 PM

Otherwise I wonder if we will have to go with the more in your face approach that explicitly tells users what they need to do and how to resolve the error for now.

Screen Shot 2020-09-03 at 4 38 47 PM

Any thoughts on the above? Hopefully I articulated it so that it makes some sense ...

megan-hall:
@zherman0 

Chatted with @beanh66 this morning. Could we go with the validation in the text field approach? If we can add a tooltip, could we change the messaging to something easier to consume like: "Label must start and end with alphanumeric characters."

Screen Shot 2020-09-03 at 4 29 11 PM

zherman0:
@beanh66 @megan-hall @gdoyle1 @spadgett - Here is my next attempt given limitations on PF:

search-tooltip-error

frontend/public/module/k8s/selector-requirement.js
zherman0: it is overloaded, but it seems much cleaner to do it this way then to either have:
```
if (validateKey(string)) { 
  return {key, 'Exists', []}
}
```
or 
```
validateKey(string) ? return {string, 'Exists' [] } : return;
```
That adds an extra conditional for 8 different branches.

megan-hall:
@zherman0 LGTM! Thank you for taking the time to get a tooltip working for it!

frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,35 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key, operator, values) => {
+  if (/^(([A-Za-z0-9][-A-Za-z0-9.]*)?[A-Za-z0-9])$/.test(key)) {
+    return { key, operator, values };
```
spadgett, line 13: I still believe this method should return a boolean, not an object.

zherman0: No problem, I will make the changes.

frontend/public/components/search-filter-dropdown.tsx
```diff
@@ -52,16 +52,23 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
...
-        name="search-filter-input"
-        id="search-filter-input"
-        value={selected === searchFilterValues.Label ? labelFilterInput : nameFilterInput}
-        onKeyDown={handleKeyDown}
-        aria-labelledby="toggle-id"
-      />
+      <Tooltip
+        content={validationMsg}
+        isVisible={validated !== ''}
+        trigger={validated === '' ? 'manual' : 'mouseenter focus'}
```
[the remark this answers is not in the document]
zherman0: I like it.

frontend/public/module/k8s/selector-requirement.js
```diff
@@ -8,42 +8,35 @@ export const createEquals = (key, value) => ({
   values: [value],
 });
 
+const validateKey = (key, operator, values) => {
+  if (/^(([A-Za-z0-9][-A-Za-z0-9.]*)?[A-Za-z0-9])$/.test(key)) {
```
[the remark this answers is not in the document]
zherman0: Yeah, that was not right. I switched it to:
```
const validateKey = (key) => /^(([A-Za-z0-9][0-9A-Za-z/\-_.]*)?[A-Za-z0-9])$/.test(key);
```
Testing looks correct.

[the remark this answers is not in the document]
spadgett: That's still not accurate, though. There are two parts: an optional prefix and a name. They have different rules.

Should we consider just fixing the runtime error when there are no results and showing the error message from the server when the selector is bad? I feel like we're duplicating too much server logic here to be honest.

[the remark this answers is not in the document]
zherman0: The runtime error appears in the virtualized vendor code.  Can you direct me on what I should look at? 

I am okay not checking this on the console side only because we are replicating the work.  That is, only if we can figure out how to prevent the empty table from exploding.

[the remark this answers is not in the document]
spadgett: I guess I'm unclear why the virtualization code is running at all. If there's an error, we shouldn't have a table to show. Is the table still being rendered?

zherman0:
PR6320 was created to catch the regex error message before it hit the backend but we decided to let the error message from the backend remain instead of trying to duplicate code that shows the same message from the front end.  As for the runtime error, it has been resolved and is no longer an issue in 4.6.  For this reasons, I am going to close this PR.

PR #6322, opened 2020-08-12
Bug 1868471: Add paused state to MachineConfigPool list view

jcaianirh:
Fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1868471

Screen Shot 2020-08-12 at 5 42 09 PM

### DIFF
--- a/frontend/public/components/machine-config-pool.tsx
+++ b/frontend/public/components/machine-config-pool.tsx
@@ -214,11 +214,12 @@ export const MachineConfigPoolDetailsPage: React.SFC<any> = (props) => (
 );
 
 const tableColumnClasses = [
-  classNames('col-lg-4', 'col-xs-6'),
+  classNames('col-lg-3', 'col-xs-6', 'col-sm-4'),
   classNames('col-lg-5', 'hidden-md', 'hidden-sm', 'hidden-xs'),
-  classNames('col-lg-1', 'col-sm-2', 'col-xs-3'),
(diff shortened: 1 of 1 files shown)


frontend/public/components/machine-config-pool.tsx
```diff
@@ -245,17 +246,28 @@ const MachineConfigPoolTableHeader = () => {
...
+    },
   ];
 };
 MachineConfigPoolTableHeader.displayName = 'MachineConfigPoolTableHeader';
 
+const getMCPPausedState = (obj: MachineConfigPoolKind) => {
+  if (_.isUndefined(obj.spec?.paused)) {
+    return 'Unknown';
+  }
+  return obj.spec?.paused ? 'True' : 'False';
```
spadgett, line 268: If it's undefined, it's not paused. So this can be just
```suggestion
  return obj.spec?.paused ? 'True' : 'False';
```

I'd inline it below.

jcaianirh: @spadgett I started with that but noticed that the other status differentiates between unknown and false. When you create a new mcp you can see them all show the same result.

spadgett: @jcaianirh Conditions have different semantics. We want false here for paused because `spec.paused` is always treated as false if missing.

jcaianirh: @spadgett thanks for clearing that up!

jcaianirh:
@spadgett added to details page, and removed unknown state.

PR #6323, opened 2020-08-13
Bug 1843532: Fix VM import status reporting for CDI imports

pcbailey:
Status reporting was broken for VMs that are pending for in progress CDI imports due to the removal of the `cdi.kubevirt.io/storage.import.importPvcName` label on data volumes. This PR fixes the reporting status.

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/dashboards-page/overview-dashboard/inventory.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/dashboards-page/overview-dashboard/inventory.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { OffIcon } from '@patternfly/react-icons';
 import { getNamespace } from '@console/shared/src/selectors/common';
 import { createBasicLookup } from '@console/shared/src/utils/utils';
-import { K8sResourceKind, PodKind } from '@console/internal/module/k8s';
+import { K8sResourceKind, PersistentVolumeClaimKind, PodKind } from '@console/internal/module/k8s';
 import { StatusGroupMapper } from '@console/shared/src/components/dashboard/inventory-card/InventoryItem';
 import { InventoryStatusGroup } from '@console/shared/src/components/dashboard/inventory-card/status-group';
 import { getVMStatus } from '../../../statuses/vm/vm-status';
@@ -25,11 +25,13 @@ export const getVMStatusGroups: StatusGroupMapper = (
     vmis,
     pods,
     migrations,
+    pvcs,
     dataVolumes,
     vmImports,
   }: {
     vmis?: VMIKind[];
     pods?: PodKind[];
+    pvcs?: PersistentVolumeClaimKind[];
     dataVolumes?: V1alpha1DataVolume[];
     migrations?: K8sResourceKind[];
     vmImports?: VMImportKind[];
@@ -113,6 +115,7 @@ export const getVMStatusGroups: StatusGroupMapper = (
         vmi,
         pods,
         migrations,
+        pvcs,
         dataVolumes,
         vmImports,
       }).status.getSimpleLabel();
--- a/frontend/packages/kubevirt-plugin/src/plugin.tsx
+++ b/frontend/packages/kubevirt-plugin/src/plugin.tsx
@@ -23,7 +23,7 @@ import {
   PVCStatus,
 } from '@console/plugin-sdk';
 import { DashboardsStorageCapacityDropdownItem } from '@console/ceph-storage-plugin';
-import { TemplateModel, PodModel } from '@console/internal/models';
+import { TemplateModel, PodModel, PersistentVolumeClaimModel } from '@console/internal/models';
 import { getName } from '@console/shared/src/selectors/common';
 import { AddAction } from '@console/dev-console/src/extensions/add-actions';
 import { FirehoseResource } from '@console/internal/components/utils';
@@ -101,6 +101,12 @@ const virtualMachineConfigurations = (namespace: string): FirehoseResource[] => 
       namespace,
       prop: 'migrations',
       optional: true,
+    },
+    {
+      isList: true,
+      optional: true,
+      kind: PersistentVolumeClaimModel.kind,
+      prop: 'pvcs',
     },
     {
       isList: true,
@@ -333,6 +339,11 @@ const plugin: Plugin<ConsumedExtensions> = [
           isList: true,
           kind: models.VirtualMachineInstanceMigrationModel.kind,
         },
+        pvcs: {
+          isList: true,
+          kind: PersistentVolumeClaimModel.kind,
+          optional: true,
+        },
         dataVolumes: {
           kind: models.DataVolumeModel.kind,
           isList: true,
@@ -404,6 +415,12 @@ const plugin: Plugin<ConsumedExtensions> = [
           isList: true,
           kind: models.VirtualMachineInstanceMigrationModel.kind,
           prop: 'migrations',
+        },
+        {
+          isList: true,
+          optional: true,
+          kind: PersistentVolumeClaimModel.kind,
+          prop: 'pvcs',
         },
         {
           isList: true,
--- a/frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
+++ b/frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
@@ -1,17 +1,14 @@
 import { get } from 'lodash';
 import { getName, getNamespace, getOwnerReferences, getUID } from '@console/shared/src/selectors';
 import { compareOwnerReference } from '@console/shared/src/utils/owner-references';
-import { PodKind } from '@console/internal/module/k8s';
-import { getLabelValue } from '../selectors';
+import { createBasicLookup } from '@console/shared/src/utils/utils';
+import { PersistentVolumeClaimKind, PodKind } from '@console/internal/module/k8s';
 import { VMKind, VMIKind } from '../../types';
-import { getDataVolumeTemplates } from '../vm';
-import {
-  CDI_KUBEVIRT_IO,
-  STORAGE_IMPORT_PVC_NAME,
-  VIRT_LAUNCHER_POD_PREFIX,
-} from '../../constants';
+import { VIRT_LAUNCHER_POD_PREFIX } from '../../constants';
 import { buildOwnerReferenceForModel } from '../../utils';
 import { VirtualMachineInstanceModel } from '../../models';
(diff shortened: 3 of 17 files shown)


frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
```diff
@@ -88,20 +84,17 @@ export const findVMIPod = (
...
 ) => {
   if (!pods) {
     return null;
   }
 
-  const datavolumeNames = getDataVolumeTemplates(vm)
-    .map((dataVolumeTemplate) => getName(dataVolumeTemplate))
-    .filter((dataVolumeTemplate) => dataVolumeTemplate);
-
+  const importerPodNames = pvcs.map((pvc) => getPvcImportPodName(pvc));
```
yaacov, line 93:
```suggestion
  const importerPodNames = (pvcs || []).map((pvc) => getPvcImportPodName(pvc));
```

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
```diff
@@ -88,20 +84,17 @@ export const findVMIPod = (
 export const getVMImporterPods = (
```
suomiy, line 84: this function should return only pods that belong to or the vm is dependent on.

So we should return only pods which import VM's DV/PVC volumes

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
suomiy: +1

also better to create a lookup here

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
```diff
@@ -88,20 +84,17 @@ export const findVMIPod = (
...
 
-  const datavolumeNames = getDataVolumeTemplates(vm)
-    .map((dataVolumeTemplate) => getName(dataVolumeTemplate))
-    .filter((dataVolumeTemplate) => dataVolumeTemplate);
-
+  const importerPodNames = pvcs.map((pvc) => getPvcImportPodName(pvc));
   return pods.filter(
     (p) =>
       getNamespace(p) === getNamespace(vm) &&
       getLabelValue(p, CDI_KUBEVIRT_IO) === 'importer' &&
```
suomiy, line 97: we don't need to check the label anymore since we have the full name

frontend/packages/kubevirt-plugin/src/statuses/vm/vm-status.ts
```diff
@@ -143,15 +142,16 @@ const isBeingMigrated = (
...
 ): VMStatusBundle => {
-  const importerPods = getVMImporterPods(vm, pods);
+  const importerPods = getVMImporterPods(vm, pods, pvcs);
   if (importerPods && importerPods.length > 0 && !isVMCreated(vm)) {
     const dvLookup = createBasicLookup(dataVolumes, getName);
     const importerPodsStatuses = importerPods.map((pod) => {
       const podStatus = getPodStatus(pod);
-      const dvName = getLabelValue(pod, `${CDI_KUBEVIRT_IO}/${STORAGE_IMPORT_PVC_NAME}`);
-      const dataVolume = dvLookup[dvName];
+      const pvcName = getPvcByImporterPodName(pvcs, getName(pod));
```
suomiy, line 153: this logic here is getting too tangled. Lets return a map from getVMImporterPods `{ [pvcName]: pod }`

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
```diff
@@ -7,29 +7,33 @@ export const getKind = (value) => _.get(value, 'kind') as K8sResourceKind['kind'
...
-  _.get(entity, ['metadata', 'labels', label]);
-
 export const getStorageSize = (value): string => _.get(value, 'requests.storage');
 
 export const getValueByPrefix = (obj = {}, keyPrefix: string): string => {
   const objectKey = Object.keys(obj).find((key) => key.startsWith(keyPrefix));
   return objectKey ? obj[objectKey] : null;
 };
 
+// Labels
```
suomiy, line 20: this should probably go to console/shared, but I am okay with doing this later.

suomiy: Nevertheless it should have better typing

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/components/dashboards-page/overview-dashboard/inventory.tsx
```diff
@@ -25,11 +25,13 @@ export const getVMStatusGroups: StatusGroupMapper = (
     vmis,
     pods,
     migrations,
+    pvcs,
     dataVolumes,
     vmImports,
   }: {
     vmis?: VMIKind[];
     pods?: PodKind[];
+    pvcs?: PersistentVolumeClaimKind[];
```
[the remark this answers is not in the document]
pcbailey: Thanks for that. I had no idea. Fixed!

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
pcbailey: That makes a lot of sense. =) Fixed!

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/statuses/vm/vm-status.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
pcbailey: What in particular are you concerned about in terms of typing?

frontend/packages/kubevirt-plugin/src/plugin.tsx
```diff
@@ -333,6 +333,11 @@ const plugin: Plugin<ConsumedExtensions> = [
           isList: true,
           kind: models.VirtualMachineInstanceMigrationModel.kind,
         },
+        pvcs: {
```
suomiy, line 336: also `virtualMachineConfigurations`. It is a mess: we should fix all of this soon :)

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
```diff
@@ -77,31 +74,33 @@ export const findVMIPod = (
...
   pods?: PodKind[],
-  pvcNameLabel = `${CDI_KUBEVIRT_IO}/${STORAGE_IMPORT_PVC_NAME}`,
+  pvcs?: PersistentVolumeClaimKind[],
 ) => {
   if (!pods) {
     return null;
   }
 
-  const datavolumeNames = getDataVolumeTemplates(vm)
+  const dataVolumeNames = getDataVolumeTemplates(vm)
```
suomiy, line 94: we could create a lookup here as well

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
suomiy: `defaultValue?: string` and better return types

frontend/packages/kubevirt-plugin/src/plugin.tsx
pcbailey: Done! And I agree! =)

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
pcbailey: I only need the names. Are you suggesting I make a lookup and then take the keys from it?

suomiy: ah you are right, maybe creating a `Set` would make more sense then.

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/selectors/pod/selectors.ts
pcbailey: Good idea! Done!

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
```diff
@@ -7,29 +7,38 @@ export const getKind = (value) => _.get(value, 'kind') as K8sResourceKind['kind'
...
+export const getLabels = (
+  entity: K8sResourceKind,
+  defaultValue?: { [key: string]: string },
+): { [key: string]: string } =>
+  _.get(entity, 'metadata.labels', defaultValue) as K8sResourceKind['metadata']['labels'];
```
suomiy, line 25: type is declared twice here

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
```diff
@@ -7,29 +7,38 @@ export const getKind = (value) => _.get(value, 'kind') as K8sResourceKind['kind'
...
+// Annotations
+export const getAnnotations = (
+  vm: VMGenericLikeEntityKind,
+  defaultValue?: { [key: string]: string },
+): { [key: string]: string } => _.get(vm, 'metadata.annotations', defaultValue);
+
+export const getAnnotation = (
+  entity: K8sResourceKind,
+  annotationName: string,
+  defaultValue?: any,
```
suomiy, line 39:
```suggestion
  defaultValue?: string,
```

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
pcbailey: Fixed!

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
pcbailey: Of course I missed one! =)

PR #6325, opened 2020-08-13
Bug 1867900: Fixed  incorrect node name in the nodeSelector while creating Auto Discovery CR

gnehapk:
- Fixed incorrect node name getting passed in the nodeSelector while creating Auto Discovery CR for AWS infra
- Fixed the post creation URL for Auto Detect Volume and Create Local volume Set
- Passing the units too along with `minSize` and `maxSize` on LVS which were not getting passed before.

### DIFF
--- a/frontend/packages/local-storage-operator-plugin/src/constants/index.ts
+++ b/frontend/packages/local-storage-operator-plugin/src/constants/index.ts
@@ -13,8 +13,8 @@ export const allNodesSelectorTxt =
 export const AUTO_DISCOVER_ERR_MSG = 'Failed to update the Auto Detect Volume!';
 
 export const diskSizeUnitOptions = {
-  TiB: 'TiB',
-  GiB: 'GiB',
+  Ti: 'TiB',
+  Gi: 'GiB',
 };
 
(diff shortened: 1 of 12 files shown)


frontend/packages/local-storage-operator-plugin/src/constants/index.ts
```diff
@@ -21,3 +21,8 @@ export const DISCOVERY_CR_NAME = 'auto-discover-devices';
 export const LOCAL_STORAGE_NAMESPACE = 'openshift-local-storage';
 export const HOSTNAME_LABEL_KEY = 'kubernetes.io/hostname';
 export const LABEL_OPERATOR = 'In';
+
+export const dropdownUnits = {
```
a2batic, line 25: We already have `dropdownUnits` here https://github.com/openshift/console/blob/master/frontend/public/components/storage/shared.ts#L35. We can make use of them directly.

gnehapk: @a2batic If you look closely, its the other way, i.e.  `Mi` is mapped to `MiB` not `MiB` to `Mi`

PR #6326, opened 2020-08-13
Bug 1870473: Make quick start headers sticky

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4437

**Analysis / Root cause**: 
Quick starts should have sticky headers in the side panel when content overflows.

**Solution Description**: 
Make the header sticky and add a drop shadow when scrolled.

sticky
@openshift/team-devconsole-ux 

Unchanged.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.scss
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.scss
@@ -1,3 +1,10 @@
+.co-quick-start-drawer-head {
+  position: sticky;
+  top: 0px;
+  background: inherit;
+  z-index: var(--pf-global--ZIndex--xs);
+}
+
 .co-quick-start-drawer {
   &__title {
     display: flex;
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 import { Dispatch } from 'redux';
 import { connect } from 'react-redux';
 import {
@@ -15,6 +16,7 @@ import {
 import { RootState } from '@console/internal/redux';
 import { AsyncComponent } from '@console/internal/components/utils';
 import { confirmModal } from '@console/internal/components/modals';
+import { useScrollDirection, ScrollDirection } from '@console/shared/';
 import {
   getActiveQuickStartID,
   getActiveQuickStartStatus,
@@ -43,6 +45,7 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
   onClose,
 }) => {
   const quickStart = getQuickStartByName(activeQuickStartID);
+  const [scrollDirection, handleScrollCallback] = useScrollDirection();
 
   const handleClose = () => {
     if (activeQuickStartStatus === QuickStartStatus.IN_PROGRESS) {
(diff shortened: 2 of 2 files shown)


dtaylor113:
Cypress CRUD (flake?) for Jobs and Routes waiting for cypress `yamlEditorReady` var to be set. 
/retest

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -60,25 +62,35 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
     return onClose();
   };
 
+  const handleScroll = (e: React.UIEvent<HTMLDivElement>) =>
+    setHeaderStyles(
+      classNames(
+        'co-quick-start-drawer-head',
+        e.currentTarget.scrollTop && 'pf-u-box-shadow-md-bottom',
```
sahil143, line 69:
```suggestion
       { 'pf-u-box-shadow-md-bottom':  !!e.currentTarget.scrollTop },
```

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -60,25 +62,35 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
...
-            {`${quickStart?.spec.duration} minutes`}
-          </Title>
-        </div>
-        <DrawerActions>
-          <DrawerCloseButton onClick={handleClose} />
-        </DrawerActions>
-      </DrawerHead>
+    <DrawerPanelContent onScroll={handleScroll}>
+      <div className={headerStyles}>
+        <DrawerHead>
```
sahil143, line 76: can we add the header styles to DrawerHead directly instead of wrapping them into a new div.

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -60,25 +62,35 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
...
-          </Title>
-          <Title headingLevel="h6" size="md" className="text-secondary">
-            {`${quickStart?.spec.duration} minutes`}
-          </Title>
-        </div>
-        <DrawerActions>
-          <DrawerCloseButton onClick={handleClose} />
-        </DrawerActions>
-      </DrawerHead>
+    <DrawerPanelContent onScroll={handleScroll}>
```
sahil143, line 74: If we are going with this approach. I would suggest using the hook `useScrollDirection` from [console-shared](https://github.com/openshift/console/blob/master/frontend/packages/console-shared/src/hooks/scroll.ts). Take a look at it

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
rottencandy: The `DrawerHead` component passes `className` to it's child, so only the title text gets the shadow applied.

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
rottencandy: Didn't know about `useScrollDirection`. Updated.

sahil143 (commented):
@rottencandy buttons are coming over the header
Screenshot 2020-08-19 at 9 19 14 AM

rottencandy:
> @rottencandy buttons are coming over the header
> Screenshot 2020-08-19 at 9 19 14 AM

Fixed.
scro

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
rohitkrai03: TBH, I don't really like this approach of using scroll position to add some styles but unfortunately there's no clean CSS solution for this as well. I guess it's okay to do it like this but I wish there was some CSS pseudo selector for sticky elements. 

FYI there is a CSS solution to this. You can add a shadow to the sticky element and add a separate cover element that covers this shadow in normal position and shows the shadow when scrolled up.

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.scss
```diff
@@ -1,3 +1,11 @@
+.co-quick-start-drawer-head {
+  box-sizing: border-box;
```
rohitkrai03, line 2: Don't really need to add this explicitly.

```suggestion
```

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -60,25 +63,32 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
     return onClose();
   };
 
+  const headerStyles = classNames('co-quick-start-drawer-head', {
```
rohitkrai03, line 66:
```suggestion
  const headerClasses = classNames('co-quick-start-drawer-head', {
```

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -60,25 +63,32 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
...
+    'pf-u-box-shadow-md-bottom':
```
rohitkrai03, line 67: Probably best to used `sm` here. Better aligned with UX mocks.
```suggestion
    'pf-u-box-shadow-sm-bottom':
```

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
sahil143: > TBH, I don't really like this approach of using scroll position to add some styles but unfortunately there's no clean CSS solution for this as well. I guess it's okay to do it like this but I wish there was some CSS pseudo selector for sticky elements.

I agree here that a CSS pseudo selector solution would be good. I looked around on web and found the same solution that you mentioned with css. But that's not a clean solution and I think current solution is okay. So, I guess we can keep the current solution.

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
rottencandy: Updated.

rohitkrai03 (approved):
/lgtm

PR #6330, opened 2020-08-13
Bug 1866296: wait for Jaeger tile before trying to click

spadgett:
/assign @dtaylor113

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts
@@ -98,6 +98,7 @@ describe('Interacting with an `AllNamespaces` install mode Operator (Jaeger)', (
     await catalogView.categoryTabsPresent();
     await catalogView.categoryTabs.get(0).click();
     await catalogPageView.clickFilterCheckbox(customProviderUID);
+    await browser.wait(until.visibilityOf(catalogPageView.catalogTileByID(jaegerTileID)));
     await catalogPageView.catalogTileByID(jaegerTileID).click();
     await browser.wait(until.visibilityOf(operatorHubView.operatorModalInstallBtn));
     await operatorHubView.operatorModalInstallBtn.click();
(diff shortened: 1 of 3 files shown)


frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts
```diff
@@ -98,6 +98,7 @@ describe('Interacting with an `AllNamespaces` install mode Operator (Jaeger)', (
     await catalogView.categoryTabsPresent();
     await catalogView.categoryTabs.get(0).click();
     await catalogPageView.clickFilterCheckbox(customProviderUID);
+    await browser.wait(until.visibilityOf(catalogPageView.catalogTileByID(jaegerTileID)));
```
dtaylor113, line 101: I recommend putting the `browser.wait' into `catalogPageView.catalogTileByID` as there are other tests which use the `catalogPageView.clickFilterCheckbox(customProviderUID)` then `catalogPageView.catalogTileByID(jaegerTileID)`.
Or.  add the 'browser.wait' to the following tests as well:
- displays subscription creation form for selected Operator
- displays Operator as subscribed in OperatorHub
- displays subscription creation form for selected Operator

spadgett: We don't want it in `catalogPageView.catalogTileByID`. This logic is specific to a particular test and does not belong in a view file. It doesn't fit the scenario/view separation of concerns.

There is one other place I could add a wait. I'll push an update.

spadgett: Putting it in the view file would also prevent us from writing tests to check that a tile is not displayed (e.g. when a certain filter is selected).

dtaylor113: Ok, sounds good.  I'd add it to the other tests which need it.

spadgett:
Last failure was an unrelated OLM test error.

/retest

dtaylor113:
@spadgett Looks like "displays subscription creation form for selected Operator" is still failing

spadgett:
@dtaylor113 It looks like every CI failure has been for a different reason. I'll dig into it. Adding a hold for now.

PR #6331, opened 2020-08-13
tests: use consistent test names in base scenario

spadgett:
Use a consistent name for the test that creates a namespace. Otherwise it looks like different tests to tools like testgrid, which creates a separate row for every run.

https://testgrid.k8s.io/redhat-openshift-ocp-release-4.6-informing#release-openshift-ocp-installer-console-aws-4.6&show-stale-tests=

Screen Shot 2020-08-13 at 4 54 20 PM

/assign @dtaylor113


spadgett:
Thanks @dtaylor, can you use the bot command so the lgtm label is applied to the PR?

PR #6332, opened 2020-08-14
Improved Cypress e2e reporting

dtaylor113:
This PR adds `test-gui-cypress-report.html` to the `gui_test_screenshots` directory.

### Before
```
.../gui_test_screenshots
  ├── cypress
          ├── screenshots
          └── videos
  ├── junit_cypress-nnnnnnnn.xml                         (one per spec)
  ├── ...
  ├── 83716afa13d47633853b248e51f0df05.png               (png for failed protractor test)
  ├── browser.log
  ├── junit_protractor.xml                               (xml for failed protractor test)
  └── test-gui-report.html                               (html report for protractor)
```

### After

- renamed `test-gui-report.html` to `protractor-report.html`
- created `cypress-report.html`

```
.../gui_test_screenshots
  ├── cypress
          ├── assets              (assets for test-gui-cypress-report.html)
          ├── screenshots
          └── videos
  ├── ...
  ├── ...
  ├── protractor-report.html
  └── cypress-report.html
```

### `cypress-report.html` initially shows failed tests:

image

**One can click on the passing tests badge in masthead to show passing tests:**

image

**Hamburger menu shows more filter options:**

image


frontend/integration-tests/protractor.conf.ts
```diff
@@ -33,7 +33,7 @@ const htmlReporter = new HtmlScreenshotReporter({
   dest: `./${screenshotsDir}`,
   inlineImages: true,
   captureOnlyFailedSpecs: true,
-  filename: 'test-gui-report.html',
+  filename: 'test-gui-protractor-report.html',
```
spadgett, line 36: Maybe just `protractor-report.html`

frontend/packages/integration-tests-cypress/reporter-config.json
```diff
@@ -1,7 +1,14 @@
...
-  "reporterEnabled": "spec, mocha-junit-reporter",
+  "reporterEnabled": "mocha-junit-reporter, mochawesome",
   "mochaJunitReporterReporterOptions": {
     "mochaFile": "../../gui_test_screenshots/junit_cypress-[hash].xml",
     "toConsole": false
+  },
+  "mochawesomeReporterOptions": {
+    "reportDir": "../../gui_test_screenshots/",
+    "reportFilename": "junit_cypress",
+    "overwrite": false,
```
spadgett, line 10: Does this mean the report will fail if it's already there? I don't see this in the options, but maybe it's not documented.

https://github.com/adamgruber/mochawesome#available-options

frontend/packages/integration-tests-cypress/reporter-config.json
```diff
@@ -1,7 +1,14 @@
 {
-  "reporterEnabled": "spec, mocha-junit-reporter",
+  "reporterEnabled": "mocha-junit-reporter, mochawesome",
   "mochaJunitReporterReporterOptions": {
     "mochaFile": "../../gui_test_screenshots/junit_cypress-[hash].xml",
     "toConsole": false
+  },
+  "mochawesomeReporterOptions": {
+    "reportDir": "../../gui_test_screenshots/",
+    "reportFilename": "junit_cypress",
```
spadgett, line 9: If this is for the JSON/HTML files, we don't need this prefix. `junit_` here is misleading since it's not a junit file. Maybe

```suggestion
    "reportFilename": "cypress-report",
```

frontend/packages/integration-tests-cypress/reporter-config.json
dtaylor113: hmm...I copied this from https://docs.cypress.io/guides/tooling/reporters.html#Config-file-4, but yeah, they don't match up?
Oh, I see, line under 'Avail Options' reads `The options below are specific to the reporter. For a list of all available options see mochawesome-report-generator options.`
We do need overwrite:
https://github.com/adamgruber/mochawesome-report-generator#overwrite

frontend/packages/integration-tests-cypress/reporter-config.json
dtaylor113: Since we specify `"html": false`, this is just for the JSON files.  I can change the name.

frontend/integration-tests/protractor.conf.ts
dtaylor113: Hi, had to trap for exit and then run the report generator, should be all set now.
Renamed to `protractor-report.html`.

dtaylor113:
Hi @spadgett, updated PR's description to show current set of files.

frontend/packages/integration-tests-cypress/reporter-config.json
```diff
@@ -1,7 +1,14 @@
 {
-  "reporterEnabled": "spec, mocha-junit-reporter",
+  "reporterEnabled": "mocha-junit-reporter, mochawesome",
   "mochaJunitReporterReporterOptions": {
     "mochaFile": "../../gui_test_screenshots/junit_cypress-[hash].xml",
     "toConsole": false
+  },
+  "mochawesomeReporterOptions": {
+    "reportDir": "../../gui_test_screenshots/",
+    "reportFilename": "cypress_report",
```
spadgett, line 9: nit: `cypress-report` so it follows the same format as the protractor report name

frontend/packages/integration-tests-cypress/reporter-config.json
spadgett: We don't need the JSON files for testgrid, only the JUnit XML report. Deleting is fine.

frontend/packages/integration-tests-cypress/reporter-config.json
dtaylor113: note: these are for the temp .json files which are deleted at the end.  The 'cypress-report' is passed in cmd line from package.json :-)

PR #6334, opened 2020-08-17
Cleanup and refactors for data resiliency and buckets card

afreen23:
- using gauge value function for resiliency
- minor refactor for resiliency activity
- cleaned up unused components for buckets-card


afreen23:
Cleanup for breakdown card is required as well.

afreen23:
Minimal deployment and encryption flags cleanup needs to be done as well.

afreen23:
/close
Forgot to close this , but this is already done.

PR #6337, opened 2020-08-17
Bug 1869127: Don't lowercase names in `k8sCreate`

spadgett:
Some resources like `Group` allow uppercase names.

/assign @rhamilto


spadgett:
I pushed a test update.

PR #6340, opened 2020-08-17
Bug 1872340: Tests for Snapshot Creation, List and Deletion 

bipuladh:
/assign @spadgett @dtaylor113

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
@@ -227,9 +227,9 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
               {PersistentVolumeModel.label}
             </label>
             <PVCDropdown
+              dataTest="pvc-dropdown"
               namespace={namespace}
               onChange={handlePVCName}
-              id="claimName"
               selectedKey={pvcName}
               desc={`Persistent Volume Claim in ${namespace} namespace`}
             />
@@ -255,8 +255,8 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
               Volume Snapshot Class
             </label>
             <SnapshotClassDropdown
+              dataTest="snapshot-dropdown"
               onChange={setSnapshotClassName}
-              id="claimName"
               selectedKey={snapshotClassName}
               pvcSC={pvcObj?.spec?.storageClassName}
             />
@@ -310,8 +310,9 @@ export const VolumeSnapshot = connectToPlural(VolumeSnapshotComponent);
 type SnapshotClassDropdownProps = {
   selectedKey: string;
   onChange: (string) => void;
-  id: string;
+  id?: string;
   pvcSC: string;
+  dataTest?: string;
 };
 
 type SnapshotResourceProps = HandlePromiseProps & {
--- a/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
@@ -51,7 +51,7 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
               </>
             )}
             <dt>Persistent Volume Claim</dt>
-            <dd>
+            <dd data-test="details-item-value__PVC">
               <ResourceLink
                 kind={PersistentVolumeClaimModel.kind}
                 name={pvcName}
@@ -61,7 +61,7 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
             {snapshotContent && (
               <>
                 <dt>Volume Snapshot Content</dt>
-                <dd>
+                <dd data-test="details-item-value__VSC">
                   <ResourceLink
                     kind={referenceForModel(VolumeSnapshotContentModel)}
                     name={snapshotContent}
@@ -70,7 +70,7 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
               </>
             )}
             <dt>Volume Snapshot Class</dt>
-            <dd>
+            <dd data-test="details-item-value__SC">
               <ResourceLink
                 kind={referenceForModel(VolumeSnapshotClassModel)}
                 name={obj?.spec?.volumeSnapshotClassName}
--- a/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
+++ b/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
@@ -1,0 +1,82 @@
+export const testerDeployment = {
+  apiVersion: 'apps/v1',
+  kind: 'Deployment',
+  metadata: {
+    name: 'busybox-deployment',
+    labels: {
+      app: 'busybox',
+    },
+  },
+  spec: {
+    replicas: 1,
+    strategy: {
+      type: 'RollingUpdate',
+    },
+    selector: {
+      matchLabels: {
+        app: 'busybox',
+      },
+    },
+    template: {
+      metadata: {
+        labels: {
+          app: 'busybox',
+        },
+      },
+      spec: {
+        volumes: [
+          {
+            name: 'testpvc',
+            persistentVolumeClaim: {
+              claimName: 'testpvc',
+            },
+          },
+        ],
+        containers: [
+          {
+            name: 'busybox',
+            image: 'busybox',
+            imagePullPolicy: 'IfNotPresent',
+            volumeDevices: [
+              {
+                name: 'testpvc',
+                devicePath: '/data',
+              },
+            ],
+            command: ['sh', '-c', 'echo Container 1 is Running ; sleep 3600'],
+          },
+        ],
+      },
+      nodeSelector: {
+        overload: 'true',
+      },
+    },
+  },
+};
+
+export const PVC = {
+  apiVersion: 'v1',
+  kind: 'PersistentVolumeClaim',
+  metadata: {
+    name: 'testpvc',
+  },
+  spec: {
+    storageClassName: 'gp2-csi',
+    accessModes: ['ReadWriteOnce'],
+    resources: {
+      requests: {
+        storage: '1Gi',
+      },
+    },
+  },
+};
+
+export const SnapshotClass = {
+  apiVersion: 'snapshot.storage.k8s.io/v1beta1',
+  kind: 'VolumeSnapshotClass',
+  metadata: {
+    name: 'csi-hostpath-snapclass',
+  },
+  driver: 'ebs.csi.aws.com',
+  deletionPolicy: 'Delete',
+};
--- a/frontend/packages/integration-tests-cypress/support/nav.ts
+++ b/frontend/packages/integration-tests-cypress/support/nav.ts
@@ -2,14 +2,14 @@ export {}; // needed in files which don't have an import to trigger ES6 module u
 declare global {
   namespace Cypress {
     interface Chainable<Subject> {
-      clickNavLink(path: [string, string]): Chainable<Element>;
+      clickNavLink(path: string[]): Chainable<Element>;
     }
   }
 }
 
 // any command added below, must be added to global Cypress interface above
 
-Cypress.Commands.add('clickNavLink', (path: [string, string]) => {
+Cypress.Commands.add('clickNavLink', (path: string[]) => {
   cy.get('#page-sidebar')
     .contains(path[0])
     .click();
--- a/frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
@@ -1,0 +1,91 @@
+import { testName, checkErrors } from '../../support';
+import { SnapshotDetails, dropdownFirstItem } from '../../views/storage/snapshot';
+import { listPage } from '../../views/list-page';
+import { PVC, testerDeployment, SnapshotClass } from '../../mocks/snapshot';
+import { detailsPage } from '../../views/details-page';
+import { modal } from '../../views/modal';
+import { resourceStatusShouldContain } from '../../views/common';
+
+const snapshotName = `${PVC.metadata.name}-snapshot`;
+
+// These tests are meant to be run on AWS as only AWS supports CSI storage classes(gp2-csi)
+if (Cypress.env('BRIDGE_AWS')) {
+  describe('Snapshot Tests', () => {
+    before(() => {
+      cy.login();
+      cy.createProject(testName);
+      cy.exec(`echo '${JSON.stringify(PVC)}' | oc apply -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(testerDeployment)}' | oc apply -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc apply -f -`);
+      cy.clickNavLink(['Storage', 'Persistent Volume Claims']);
+      listPage.filter.byName(PVC.metadata.name);
+      resourceStatusShouldContain('Bound');
+    });
+
+    afterEach(() => {
+      checkErrors();
+    });
+
+    after(() => {
+      cy.exec(`echo '${JSON.stringify(testerDeployment)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(PVC)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc delete -f -`);
+      cy.deleteProject(testName);
+      cy.logout();
+    });
+
+    it('Creates Snapshot', () => {
+      cy.clickNavLink(['Volume Snapshots']);
+      listPage.clickCreateYAMLbutton();
+      cy.byTestID('pvc-dropdown').click();
+      cy.get(dropdownFirstItem)
+        .first()
+        .click();
+      cy.byTestID('snapshot-dropdown').click();
+      cy.get(dropdownFirstItem)
+        .first()
+        .click();
+      modal.submit();
+      cy.location('pathname').should(
+        'include',
+        `snapshot.storage.k8s.io~v1beta1~VolumeSnapshot/${PVC.metadata.name}-snapshot`,
+      );
+      detailsPage.titleShouldContain(PVC.metadata.name);
+      resourceStatusShouldContain('Ready');
+      cy.exec(`oc get VolumeSnapshot ${PVC.metadata.name}-snapshot -n ${testName} -o json`)
+        .its('stdout')
+        .then((res) => {
+          const volumeSnapshot = JSON.parse(res);
+          cy.get(SnapshotDetails.name).contains(volumeSnapshot.metadata.name);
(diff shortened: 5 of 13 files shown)


frontend/packages/integration-tests-cypress/mocks/snapshot.ts
```diff
@@ -0,0 +1,82 @@
+export const testerDeployment = {
+  apiVersion: 'apps/v1',
```
dtaylor113, line 2: Hi, Can this be handled by the standard [Cypress Fixtures](https://docs.cypress.io/api/commands/fixture.html) api?

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+describe('Snapshot Tests', () => {
+  before(() => {
+    cy.login();
+    cy.createProject(testName);
+    cy.exec(`echo '${JSON.stringify(PVC)}' | oc apply -n ${testName} -f -`);
+    cy.exec(`echo '${JSON.stringify(testerDeployment)}' | oc apply -n ${testName} -f -`);
+    cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc apply -f -`);
+    // Let the PVC get bound
+    cy.wait(10 * 1000);
```
dtaylor113, line 27: `waits` are frowned upon in general, is there a GUI page or component we can test for which will be in the DOM after PVC 'gets bound'?

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  after(() => {
+    cy.deleteProject(testName);
+    cy.logout();
+  });
+
+  it('Creates Snapshot', () => {
+    cy.clickNavLink(['Storage', 'Volume Snapshots']);
+    cy.location('pathname', { timeout: 60000 }).should(
```
dtaylor113, line 41: I see you added `timeout` to your `cy.location` calls, if this is going to be a common issue, please look at overwriting `cy.location` and always including the `timeout`.  Similar to: https://github.com/openshift/console/pull/6372/commits/5bf3d4c90df42c15895a4158ad717f5912dbb6af

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      .first()
+      .click();
+    cy.get(submitButton).click();
+    cy.location('pathname', { timeout: 60000 }).should(
+      'include',
+      `snapshot.storage.k8s.io~v1beta1~VolumeSnapshot/${PVC.metadata.name}-snapshot`,
+    );
+    detailsPage.titleShouldContain(PVC.metadata.name);
+    // eslint-disable-next-line promise/catch-or-return
+    cy.exec(`oc get VolumeSnapshot ${PVC.metadata.name}-snapshot -n ${testName} -o json`)
```
dtaylor113, line 61: Instead of adding `// eslint-disable-next-line promise/catch-or-return`, you can just `return cy.exec(...`

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        cy.get(SnapshotDetails.name).contains(volumeSnapshot.metadata.name);
+        cy.get(SnapshotDetails.namespace).contains(volumeSnapshot.metadata.namespace);
+        cy.get(SnapshotDetails.vsc).contains(volumeSnapshot.status.boundVolumeSnapshotContentName);
+        cy.get(SnapshotDetails.sc).contains(volumeSnapshot.spec.volumeSnapshotClassName);
+        cy.get(SnapshotDetails.pvc).contains(volumeSnapshot.spec.source.persistentVolumeClaimName);
+      });
+  });
+
+  it('Lists Snapshot', () => {
+    cy.clickNavLink((['Volume Snapshots'] as unknown) as [string, string]);
```
dtaylor113, line 74: Rather than `...as unknown) as [string, string]`, could we somehow make second argument optional [here](https://github.com/openshift/console/blob/5150fde959976b79608956031a555c0bbef01598/frontend/packages/integration-tests-cypress/support/nav.ts#L6)?

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -46,12 +46,12 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
...
               <>
-                <dt>Size</dt>
-                <dd>{sizeMetrics}</dd>
+                <dt data-test-selector="details-item-label__Size">Size</dt>
+                <dd data-test-selector="details-item-value__Size">{sizeMetrics}</dd>
               </>
             )}
-            <dt>Persistent Volume Claim</dt>
-            <dd>
+            <dt data-test-selector="details-item-label__PVC">Persistent Volume Claim</dt>
```
dtaylor113, line 53: I do not see where the following are used?  If not used, please remove and add back when needed -thanks
```
data-test-selector="details-item-label__Size"
data-test-selector="details-item-value__Size"
data-test-selector="details-item-label__PVC"
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -46,12 +46,12 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
...
+            <dd data-test-selector="details-item-value__PVC">
```
dtaylor113, line 54: Can this be changed to `data-test="details-item-value__PVC"` per [best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#best-practices)?
Then in `frontend/packages/integration-tests-cypress/views/snapshot.ts` it could be:
```
export namespace SnapshotDetails {
  ...
  export const pvc = cy.byTestID('details-item-value__PVC');
  ...
```
[cy.byTestID](https://github.com/openshift/console/blob/9d9cb1f9a5279cf04c4a32f060716edda9951363/frontend/packages/integration-tests-cypress/support/selectors.ts#L15)
If this ID is unique on the page, then should be ok.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { testName, checkErrors } from '../../support';
+import {
+  SnapshotDetails,
+  snapshotClassDropdown,
+  dropdownFirstItem,
+  pvcDropdown,
+} from '../../views/snapshot';
+import { listPage } from '../../views/list-page';
+import { PVC, testerDeployment, SnapshotClass } from '../../mocks/snapshot';
```
dtaylor113, line 9: Hi, Can this be handled by the standard [Cypress Fixtures](https://docs.cypress.io/api/commands/fixture.html) api?

frontend/packages/integration-tests-cypress/views/snapshot.ts
```diff
@@ -0,0 +1,11 @@
+export namespace SnapshotDetails {
+  export const name = 'dd[data-test-selector="details-item-value__Name"]';
+  export const namespace = 'dd[data-test-selector="details-item-value__Namespace"] a';
+  export const pvc = 'dd[data-test-selector="details-item-value__PVC"] a';
+  export const vsc = 'dd[data-test-selector="details-item-value__VSC"] a';
+  export const sc = 'dd[data-test-selector="details-item-value__SC"] a';
+}
```
dtaylor113, line 7: Please see previous comment about changing `data-test-selector=` to `data-test=`.

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -229,7 +229,7 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
             <PVCDropdown
               namespace={namespace}
               onChange={handlePVCName}
-              id="claimName"
+              id="pvcDropdown"
```
dtaylor113, line 232: This should be changed to `data-test=...` per our [best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#best-practices).  Also, pls read Cypress's [Selecting Elements: Best Practices](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) -thanks

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -256,7 +256,7 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
             </label>
             <SnapshotClassDropdown
               onChange={setSnapshotClassName}
-              id="claimName"
+              id="snapshotDropdown"
               selectedKey={snapshotClassName}
```
dtaylor113, line 260: This should be changed to `data-test=...` per our [best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#best-practices).

frontend/packages/integration-tests-cypress/views/snapshot.ts
```diff
@@ -0,0 +1,11 @@
+export namespace SnapshotDetails {
+  export const name = 'dd[data-test-selector="details-item-value__Name"]';
+  export const namespace = 'dd[data-test-selector="details-item-value__Namespace"] a';
+  export const pvc = 'dd[data-test-selector="details-item-value__PVC"] a';
```
dtaylor113, line 4: If changed to `data-test=...`, then this could be:
`export const pvc = cy.byTestID('details-item-value__PVC');`

frontend/packages/integration-tests-cypress/views/snapshot.ts
```diff
@@ -0,0 +1,11 @@
...
+  export const vsc = 'dd[data-test-selector="details-item-value__VSC"] a';
+  export const sc = 'dd[data-test-selector="details-item-value__SC"] a';
+}
+
+export const pvcDropdown = '#pvcDropdown';
+export const snapshotClassDropdown = '#snapshotDropdown';
+export const dropdownFirstItem = '.dropdown-menu__autocomplete-filter li a';
```
dtaylor113, line 11: Should not be using # ids or class selectors (unless a third party lib where we can't add our own `data-test=...`.
Pls see: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

frontend/packages/integration-tests-cypress/mocks/snapshot.ts
bipuladh: Cypress fixtures are loaded asynchronously. Since these are used in multiple places(across various code blocks). It makes the design a bit complicated if we load it async.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
bipuladh: So to track the bound/failed status we would either have to poll the command line or go to the PVC page and wait on this resource.

frontend/packages/integration-tests-cypress/views/snapshot.ts
bipuladh: We can use `cy.byTestID` but for `PVC` we need to select the anchor tag as it contains the text we are interested in.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
dtaylor113: "go to the PVC page and wait on this resource" sounds more Cypress-y :-)

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
dtaylor113: You should no longer need these eslint-disable rules, recently [udpated](https://github.com/openshift/console/pull/6398/files#diff-8949092bfde58c674fd96f762a30519b)

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    it('Lists Snapshot', () => {
+      cy.clickNavLink(['Volume Snapshots']);
+      listPage.rows.shouldBeLoaded();
+      listPage.rows.shouldExist(snapshotName);
+      listPage.rows.shouldNotExist(`${snapshotName}dup`);
+    });
+
+    it('Deletes Snapshot', () => {
+      listPage.rows.clickKebabAction(snapshotName, 'Delete Volume Snapshot');
+      cy.get('#confirm-action').click();
```
dtaylor113, line 85: Can we add a `data-test` prop to this? Also, would `modal.submit();` work in this case?  Might want to also see `modal.shouldBeOpened() && modal.shouldBeClosed()` helper methods.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        .first()
+        .click();
+      cy.get(submitButton).click();
+      cy.location('pathname').should(
+        'include',
+        `snapshot.storage.k8s.io~v1beta1~VolumeSnapshot/${PVC.metadata.name}-snapshot`,
+      );
+      detailsPage.titleShouldContain(PVC.metadata.name);
+      cy.contains(status, 'Ready');
+      return cy
```
dtaylor113, line 59: Shouldn't need this `return` anymore, recently disabled this eslint check.

frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -56,3 +56,5 @@ export const listPage = {
       cy.get(`[data-test-id="${resourceName}"]`, { timeout: 90000 }).should('not.be.visible'),
   },
 };
+
+export const status = '[data-test="camel-case-wrapped-text"]';
```
dtaylor113, line 60: This isn't very specific to `status`.  There is a potential for a list page to have multiple tables, sections, statuses.  Can we use something more specific, ie `data-test='pvc-status'`.
Also, not sure `status` is common enough on list pages to warrant putting it in `list-page.ts`.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
bipuladh: This rule related to "promise/catch-or-return" is not disabled. I will disable it and remove the return.

frontend/packages/integration-tests-cypress/support/nav.ts
```diff
@@ -3,14 +3,14 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace,no-redeclare
```
dtaylor113, line 3: Should be able to remove this :-)

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      cy.exec(`echo '${JSON.stringify(PVC)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc delete -f -`);
+      cy.deleteProject(testName);
+      cy.logout();
+    });
+
+    it('Creates Snapshot', () => {
+      cy.clickNavLink(['Volume Snapshots']);
+      listPage.clickCreateYAMLbutton();
+      cy.byTestID(pvcDropdown).click();
```
dtaylor113, line 44: Not sure why this needs to be in a variable, just as easy to write `pvc-dropdown`.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      cy.get(dropdownFirstItem)
+        .first()
+        .click();
+      cy.byTestID(snapshotClassDropdown).click();
```
dtaylor113, line 48: Not sure why this needs to be in a variable, just as easy to write `snapshot-dropdown`, and don't see that this var is reused anywhere.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
dtaylor113: Hmm...thought I did disable that one, but I see it isn't :-)  Thanks, please disable, and perhaps search for and fix any `return cy...`

frontend/packages/integration-tests-cypress/mocks/snapshot.ts
```diff
@@ -0,0 +1,82 @@
+export const testerDeployment = {
```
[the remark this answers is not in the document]
bipuladh: This deployment can be used for PVC testing too. But we don't have any plans of moving PVC tests to it right now. We can easily move it when we have the requirement to do so. For now, I will keep it as it is.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
bipuladh: modal.submit() works!

frontend/packages/integration-tests-cypress/mocks/snapshot.ts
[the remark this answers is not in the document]
cloudbehl: Correct even if no plans to have PVC creation tests we still need this for snapshot creation, clone & also for restore. Clone is not related to Snapshot so it'll not make sense to use it from this file. 

Let's have storage-common.ts & then respective files. WDYT?

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -0,0 +1,2 @@
+export const waitForResourceStateToBe = (desiredState: string) =>
```
dtaylor113, line 1: Suggest `waitForResourceStatusToBe`

frontend/packages/integration-tests-cypress/mocks/snapshot.ts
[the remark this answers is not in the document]
bipuladh: Let's do that when we add clone and restore tests.

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -0,0 +1,2 @@
+export const resourceStatusShouldBe = (desiredState: string) =>
```
dtaylor113, line 1: Hi, really should be `resourceStatusShouldContain` as that is the funciton.  `...ShouldBe` implies 'equals'.

spadgett:
/hold

This has failed roughly 30 consecutive runs. We need to investigate the failures.

dtaylor113:
Hi, you'll probably need to revert changes to `yaml-editor.ts`, they were causing failures in my PR. 
You can revert the `disabled eslint promise/catch-or-return` changes as I implemented them in the PR below while addressing other cypress issues:
https://github.com/openshift/console/pull/6475/commits/02f1cf6ea2bc816c94b32c21a08a9d28e6455497
- Thanks

PR #6343, opened 2020-08-17
Bug 1870636: Add ability to select unassigned as application filter in topology

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4447

**Description**
Add the `Unassigned` group to the list of filters for application groupings. 

**Screenshots**
image

image

image

image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

/cc @bgliwa01


christianvogt:
@jeff-phillips-18 import from GIT workflow still creates resources in an application instead of being unassigned.

jeff-phillips-18:
@christianvogt Yeah. This is to add the ability to filter by `Unassigned` in the application drop down in the topology view.

bgliwa01:
Please change 'Unassigned' to be 'unassigned' in the dropdown, as well as the other locations we discussed.

jeff-phillips-18:
@christianvogt Resources are now created as unassigned if the unassigned filter is set.

@bgliwa01 Changed `Unassigned` to `unassigned` in the dropdown as well as in the application selector (and help text) in the create flows.

jeff-phillips-18:
@parvathyvr I have made those changes, updated the screenshots above

rohitkrai03 (changes requested):
@jeff-phillips-18 Even if i go to git import form with application selected as `unassigned`, the app gets created in application name context. After digging into the code it seems this is happening due to `GitSection` using duplicate/wrong constant `UNASSIGNED_KEY` instead of `UNASSIGNED_APPLICATIONS_KEY`.

Peek 2020-08-19 19-32

jeff-phillips-18:
@rohitkrai03 Fixed, thanks.

christianvogt:
When i go through the import flows with “unassigned” application selected, the application section is showing.

But otherwise this works.

jerolimov:
@jeff-phillips-18 When selecting Topology from the sidebar the application dropdown shows the right entries:

image

But when I'm already on the Topology graph and press F5 to refresh the menu keeps empty also when the graph shows some applications:

image

When selecting another navigation entry and Topology again it works fine again.

jeff-phillips-18:
I don't believe we intended the application filter to persist thru sessions as it is not part of the URL. That would be a request for change.

jerolimov:
> I don't believe we intended the application filter to persist thru sessions as it is not part of the URL. That would be a request for change.

Oh, I do not expected that the selection was saved. But sometimes the dropdown was not filled and I could not can select the app or your new "unassigned". On the new cluster it works now. I expect that there is an race condition that the application dropdown select was not updated always. As written in my "edit" note above this happen also on master so this does not block this issue here. I will keep an eye on this.

jeff-phillips-18:
> Btw: Is it possible to show entry only if there is any node without assigned application?

The only way to do this would be to load up the resources in the same manner as the topology view. Currently we retrieve all resources that __could__ have the `part-of` label but not all of those resources are shown in the view or are considered to be belonging to an application or not. For instance pods and secrets.

jerolimov (approved):
/lgtm
/assign @rohitkrai03

rohitkrai03:
@jeff-phillips-18 Got merge conflicts on this one.

jeff-phillips-18:
@jerolimov Rebased, no other changes.

jerolimov:
@jeff-phillips-18 Unfortunately I found another small issue.

When I right click on a Deployment with an app name and select Edit Application Grouping and select "unassigned" in the modal (which was Unassigned before, but I think thats ok), the app name was changed to this name instead of 'none' app name.

jeff-phillips-18:
> which was Unassigned before, but I think thats ok

UX asked for it to be lowercased consistently during discussions about https://github.com/openshift/console/pull/6343#issuecomment-675113106

jerolimov:
Could verify "Edit Application Grouping" fix

/lgtm

rohitkrai03 (approved):
/approve

PR #6345, opened 2020-08-17
Bug 1869600: add resource icon to the items in the `select task` dropdown in the pipeline builder page

debsmita1:
**Fixes:**
- https://issues.redhat.com/browse/ODC-4280
- https://issues.redhat.com/browse/ODC-3182

**Root Analysis:**
Tasks and ClusterTasks are not visually distinguishable in the `Select Task` dropdown list in the pipeline builder page.

**Solution Description:**
- added resource badge to the items in the `Select Task` dropdown
- fixed the issue with Task & ClusterTask having the same name
- fixed the issue with Tasks not showing up again in the dropdown if it is already selected

**GIF:**
taskdropdown

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/hooks.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/hooks.ts
@@ -133,17 +133,14 @@ export const useNodes = (
     onUpdateTasks(taskGroupRef.current, { type: UpdateOperationType.CONVERT_LIST_TO_TASK, data });
   };
 
-  // TODO: Fix id collisions then remove this utility; we shouldn't need to trim the tasks
-  const noDuplicates = (resource: PipelineResourceTask) =>
-    !taskGroupRef.current.tasks.find((pt) => pt.name === resource.metadata.name);
   const newListNode = (
     name: string,
     runAfter?: string[],
     firstTask?: boolean,
   ): PipelineTaskListNodeModel =>
     createTaskListNode(name, {
-      namespaceTaskList: namespacedTasks?.filter(noDuplicates),
-      clusterTaskList: clusterTasks?.filter(noDuplicates),
+      namespaceTaskList: namespacedTasks,
+      clusterTaskList: clusterTasks,
       onNewTask: (resource: PipelineResourceTask) => {
         onNewTask(resource, name, runAfter);
       },
@@ -163,8 +160,8 @@ export const useNodes = (
   const soloTask = (name = 'initial-node') => newListNode(name, undefined, true);
   const newInvalidListNode = (name: string, runAfter?: string[]): PipelineTaskListNodeModel =>
     createInvalidTaskListNode(name, {
-      namespaceTaskList: namespacedTasks?.filter(noDuplicates),
-      clusterTaskList: clusterTasks?.filter(noDuplicates),
+      namespaceTaskList: namespacedTasks,
+      clusterTaskList: clusterTasks,
       onNewTask: (resource: PipelineResourceTask) => {
         const data: UpdateOperationFixInvalidTaskListData = {
           existingName: name,
--- a/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
@@ -1,6 +1,7 @@
 import { history, resourcePathFromModel } from '@console/internal/components/utils';
 import { apiVersionForModel, referenceForModel } from '@console/internal/module/k8s';
-import { ClusterTaskModel, PipelineModel } from '../../../models';
+import { getRandomChars } from '@console/shared';
+import { PipelineModel, TaskModel } from '../../../models';
 import {
   Pipeline,
   PipelineResourceTask,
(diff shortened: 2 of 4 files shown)


divyanshiGupta:
@debsmita1 I think we should also add the resource badge to the builder nodes as without the badge the user wont be able to distinguish between the resources with same name just by looking at the pipeline. For example in the attached image one resource is a `Task` and the other is `ClusterTask`
image

cc: @beaumorley @bgliwa01

debsmita1:
> @debsmita1 I think we should also add the resource badge to the builder nodes as without the badge the user wont be able to distinguish between the resources with same name just by looking at the pipeline. For example in the attached image one resource is a `Task` and the other is `ClusterTask`
> image
> 
> cc: @beaumorley @bgliwa01

@andrewballantyne ^

andrewballantyne:
> > @debsmita1 I think we should also add the resource badge to the builder nodes as without the badge the user wont be able to distinguish between the resources with same name just by looking at the pipeline. For example in the attached image one resource is a `Task` and the other is `ClusterTask`
> > image
> > cc: @beaumorley @bgliwa01
> 
> @andrewballantyne ^

No, I think at this point it does not matter. Selecting in a mixed list is a problem because you don't know which is which... but the name in the bubble is a custom name (it just pre-selects as the name of task you took). If you click on the bubble the badge is at the top of the sidebar and the "name" field is a custom name you can use for this task irrespective of what it was originally named.

So I think we are good with just the badge in the dropdown list. @bgliwa01 Please let me know your thoughts.

bgliwa01:
Agreed @andrewballantyne

frontend/packages/dev-console/src/components/pipelines/pipeline-topology/TaskListNode.tsx
```diff
@@ -82,7 +84,7 @@ const TaskListNode: React.FC<TaskListNodeProps> = ({ element, unselectedText })
           <div className="pf-c-dropdown pf-m-expanded">
             <ul className="pf-c-dropdown__menu pf-m-align-right oc-kebab__popper-items odc-task-list-node__list-items">
               {options.map((option) => (
-                <li key={option.label}>
+                <li key={`${option.label}-${option.icon?.['props']?.kind}`}>
```
andrewballantyne, line 87: Not a fan of digging into props unless you absolutely have to... I think since we have a little sandbox of code here, we could change things up a bit with the help of TypeScript.

```
type KeyedKebabOption = KebabOption & { key: string };

const taskToOption = (task: PipelineResourceTask, callback: NewTaskNodeCallback): KeyedKebabOption => {
```

That way we return a `option.key` value that we can use. You can set `key` simply to `${kind}-${name}` with the data.

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -33,11 +33,12 @@ export const convertResourceToTask = (
   resource: PipelineResourceTask,
   runAfter?: string[],
 ): PipelineTask => {
+  const kind = resource.kind ?? TaskModel.kind;
   return {
-    name: resource.metadata.name,
+    name: `${resource.metadata.name}-${kind.toLowerCase()}`,
```
andrewballantyne, line 38: If we do this, we are not really avoiding the problem entirely. We are likely to be okay, but let's avoid anything that will lead to potentially another bug (or if QE is on game, a failure of the QE Review).

Recommend a generate name function (maybe we can put this in shared, or here -- wherever for now):

```
export const safeName = (reservedNames: string[], desiredName: string): string => {
  if (reservedNames.includes(desiredName)) {
    const newName = `${desiredName}-${getRandomChars()}`; // maybe `getRandomChars(3)`?
    if (reservedNames.includes(newName)) {
      return safeName(reservedNames, desiredName);
    }
    return newName;
  }
  return desiredName;
};

// use like:
name: safeName(reservedNames, resource.metadata.name),
```

Then you update the two uses of `convertResourceToTask` to include a 3rd param (I'd make it the first param personally) of a `string[]`. 

Sample usage:
```
  const usedNames = tasks.map((t) => t.name);
  const newPipelineTask: PipelineTask = convertResourceToTask(usedNames, resource, runAfter);
```

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/hooks.ts
```diff
@@ -135,7 +135,10 @@ export const useNodes = (
 
   // TODO: Fix id collisions then remove this utility; we shouldn't need to trim the tasks
   const noDuplicates = (resource: PipelineResourceTask) =>
-    !taskGroupRef.current.tasks.find((pt) => pt.name === resource.metadata.name);
+    !taskGroupRef.current.tasks.find(
+      (pt) => pt.taskRef.name === resource.metadata.name && pt.taskRef.kind === resource.kind,
```
andrewballantyne, line 139: We can't do `resource.metadata.name` against `pt.taskRef.name` because this will prevent any double uses of a given task. This isn't the end of the world, but likely will be a problem for some use-cases.

However! With my recent changes suggested, I think we can actually get rid of this function since we'll always will generate a new name.

invincibleJai (approved):
Verified the changes looks good !!

PR #6348, opened 2020-08-18
Bug 1869196: Disable card model if using SRIOV interface

yaacov:
Currently, it is possible to select the model if sriov is selected. But this selection has no meaning in this situation and should be disabled

Screenshot:
After:
screenshot-localhost_9000-2020 08 18-11_39_39

Before:
screenshot-localhost_9000-2020 08 18-11_40_51


yaacov:
@suomiy @pcbailey @irosenzw @glekner please review

PR #6349, opened 2020-08-18
Add color to vm imports

yaacov:
Currently VM (VirtualMachine) and VIM (VIrtualMachineImports) share the same label colour,
We show VM, VMI and VIM objects on the same list in the Virtualization table.
It will be nice to have better visual hint to distinguish between VM and VIM.

This PR adds more colour to the VIM object.

After:
screenshot-localhost_9000-2020 08 18-13_22_11

Before:
screenshot-localhost_9000-2020 08 18-13_21_11


yaacov:
@matthewcarleton @Ranelim @yfrimanm PTAL

yaacov:
Closing, we will have a new import flow that may change the way we use the import objects, we will discuss how to expose the import objects once the new design formalize.

PR #6350, opened 2020-08-18
Bug 1870478: Avoid adding a newline between pipelinerun log fetches

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4361

**Analysis / Root cause**: 
There is a new line added after each api call, which shows incorrectly if the logs are streamed with one or more lines broken down into multiple parts between calls.

**Solution Description**: 
Check if the previously streamed chunk ends in a newline before displaying the next chunk, avoid adding a newline if it doesn't.

Before:
scr0
After:
scr1

Unchanged.

**Test setup:**
Prevalent in a PipelineRun with lots of logs.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelineruns/logs/Logs.tsx
+++ b/frontend/packages/dev-console/src/components/pipelineruns/logs/Logs.tsx
@@ -33,9 +33,14 @@ const Logs: React.FC<LogsProps> = ({
   const onCompleteRef = React.useRef<(name) => void>();
   onCompleteRef.current = onComplete;
   const appendMessage = React.useRef<(blockContent) => void>();
+  const prevFetchNewline = React.useRef(true);
   appendMessage.current = React.useCallback(
-    (blockContent) => {
+    (blockContent: string) => {
       const contentLines = blockContent.split('\n').filter((line) => !!line);
+      if (!prevFetchNewline.current && contentRef.current.lastChild) {
+        contentRef.current.lastChild.textContent += contentLines.shift();
+      }
+      prevFetchNewline.current = blockContent.endsWith('\n');
       if (contentRef.current && contentLines.length >= 0) {
         const elements = contentLines.map((content) => {
           const customElement = document.createElement('div');


andrewballantyne:
@rottencandy Please provide image of logs.

/auto-cc
/assign

nemesis09:
verified locally
/lgtm

jerolimov (approved):
Could verify this also locally (crc with provided pipeline in the linked Jira issue).

I was confused by the solution first. IMHO this code could be simpler if we do not skip empty lines in the log and I think this is the root of this issue before. But I expect that we do not want to change this behavior. In this case this also lgtm :+1: 

/cc @andrewballantyne

andrewballantyne:
Hmm... I am having issues reproducing the issue... did you do anything special to setup the Pipeline/PipelineRun @rottencandy ?

rottencandy:
> Hmm... I am having issues reproducing the issue... did you do anything special to setup the Pipeline/PipelineRun @rottencandy ?

@andrewballantyne I used the petclinic yaml attached in the jira issue. Does plenty of logging.

andrewballantyne:
> @andrewballantyne I used the petclinic yaml attached in the jira issue. Does plenty of logging.

If I remove your change, I do not see the original issue... which is what I find confusing.

rottencandy:
> If I remove your change, I do not see the original issue... which is what I find confusing.

Hmm, I don't think I used anything other than that yaml, it was enough for me to identify the issue. I'll try again.

rottencandy:
@andrewballantyne I can see the line breaks after removing the changes:
demo

andrewballantyne:
I am very confused lol... what version of the Operator are you using? @rottencandy

andrewballantyne (approved):
Not the most ideal implementation, but I can see what you're trying to do here. My bad for not getting to this earlier... I think this is a safe implementation that is not likely to spawn other bugs.

PR #6353, opened 2020-08-18
Bug 1869631: Mode Detection independent of Storage Cluster CR name

bipuladh:
Detection of storage cluster mode should be based on spec.externalStorage and not on the CR name.


gnehapk:
Backport PR for https://github.com/openshift/console/pull/6198

PR #6355, opened 2020-08-18
Bug 1862903: "Add disk" button exist in the wizard but not active

glekner:


glekner:
@yaacov @suomiy please review

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/networking-tab/networking-tab.tsx
```diff
@@ -127,9 +127,11 @@ const NetworkingTabComponent: React.FC<NetworkingTabComponentProps> = ({
             <Title headingLevel="h5" size="lg">
               No network interface added
             </Title>
-            <Button {...addButtonProps} icon={<PlusCircleIcon />} variant={ButtonVariant.link}>
-              {ADD_NETWORK_INTERFACE}
-            </Button>
+            {!isCreateDisabled && (
```
suomiy, line 130: we should include the buttons itself too (not just the empty flow)

glekner: added

PR #6360, opened 2020-08-18
Bug 1866296: Disable e2e OLM test flakes

dtaylor113:
This PR disables known e2e OLM flakes.

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts
@@ -18,7 +18,7 @@ import * as operatorHubView from '../views/operator-hub.view';
 import { click } from '@console/shared/src/test-utils/utils';
 import { NAME_FIELD_ID, formFieldIsPresent } from '../views/descriptors.view';
 
-describe('Interacting with an `AllNamespaces` install mode Operator (Jaeger)', () => {
+xdescribe('Interacting with an `AllNamespaces` install mode Operator (Jaeger)', () => {
   const jaegerResources = new Set([
     'Deployment',
     'Service',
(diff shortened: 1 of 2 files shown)


frontend/integration-tests/protractor.conf.ts
```diff
@@ -69,8 +69,8 @@ const testSuites = {
   olmFull: suite([
     '../packages/operator-lifecycle-manager/integration-tests/scenarios/descriptors.scenario.ts',
     '../packages/operator-lifecycle-manager/integration-tests/scenarios/operator-hub.scenario.ts',
-    '../packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts',
-    '../packages/operator-lifecycle-manager/integration-tests/scenarios/single-installmode.scenario.ts',
+    // '../packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts',
+    // '../packages/operator-lifecycle-manager/integration-tests/scenarios/single-installmode.scenario.ts',
```
spadgett, line 73: Not this way, use `xdescribe` in the scenario itself. Then it gets gets status disabled in the report so we know a test is disabled.

frontend/integration-tests/tests/alertmanager.scenario.ts
```diff
@@ -24,7 +24,7 @@ const getGlobalsAndReceiverConfig = (configName: string, yamlStr: string) => {
   };
 };
 
-describe('Alertmanager: PagerDuty Receiver Form', () => {
+xdescribe('Alertmanager: PagerDuty Receiver Form', () => {
```
spadgett, line 27: Do we know if every one of these is flaking, or is it only certain tests?

frontend/integration-tests/protractor.conf.ts
dtaylor113: ok, fixed

frontend/integration-tests/tests/alertmanager.scenario.ts
dtaylor113: Hi @spadgett,  Not exactly sure, I've seen a few flakes in a few of these tests.  Should I just disable OLM and then we can try to fix anything else that fails?

dtaylor113: Hi @spadgett, removed the monitoring/alerting so this PR just deals with OLM.   I think other PRs will have to be made to address Cypress CRUD and Monitoring/Alertmanagement test flakes.

PR #6362, opened 2020-08-18
Bug 1870627: Fix to maintain URL params on topology view switch

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4508
https://issues.redhat.com/browse/ODC-4509

**Description**
Fix to topology view to keep URL params on switching views. When no selection, remove the select parameter (rather then setting to undefined).

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


rohitkrai03 (approved):
/lgtm
/approve

PR #6364, opened 2020-08-18
[WIP] Implemented cypress-plugin-retries

dtaylor113:
https://github.com/Bkucera/cypress-plugin-retries

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -204,6 +204,7 @@
     "cypress": "^4.3.0",
     "cypress-jest-adapter": "^0.1.1",
     "cypress-multi-reporters": "^1.4.0",
+    "cypress-plugin-retries": "^1.5.2",
     "enzyme": "3.10.x",
     "enzyme-adapter-react-16": "1.15.2",
     "eslint-plugin-cypress": "^2.10.3",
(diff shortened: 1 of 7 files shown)


spadgett:
Lots of build errors with this change like

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/ceph-storage-plugin/integration-tests/tests/1-install/installFlow.scenario.ts
(52,5): No overload matches this call.
  Overload 1 of 6, '(title: string, config: TestConfigOverrides, fn?: Func): Test', gave the following error.
    Type '() => Promise<void>' has no properties in common with type 'TestConfigOverrides'.
  Overload 2 of 6, '(title: string, config: TestConfigOverrides, fn?: AsyncFunc): Test', gave the following error.
    Type '() => Promise<void>' has no properties in common with type 'TestConfigOverrides'.
```

dtaylor113:
> Lots of build errors with this change like
> 
> ```
> ERROR in /go/src/github.com/openshift/console/frontend/packages/ceph-storage-plugin/integration-tests/tests/1-install/installFlow.scenario.ts
> (52,5): No overload matches this call.
>   Overload 1 of 6, '(title: string, config: TestConfigOverrides, fn?: Func): Test', gave the following error.
>     Type '() => Promise<void>' has no properties in common with type 'TestConfigOverrides'.
> ```

Hi @spadgett, where do you see these errors?

PR #6366, opened 2020-08-19
Bug 1869942: Monitoring: Clarify popover help text

kyoto:
Update text to be more helpful and to align with the 4.6 docs.

Before | After
-|-
before-severity | after-severity
before-source | after-source
before-state | after-state

### DIFF
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -282,17 +282,28 @@ const PopoverField: React.FC<{ body: React.ReactNode; label: string }> = ({ body
 const alertStateHelp = (
   <dl className="co-inline">
     <dt>
-      <AlertStateIcon state={AlertStates.Firing} /> <strong>Firing: </strong>
-    </dt>
-    <dd>The alert condition is true for the duration of the timeout.</dd>
-    <dt>
       <AlertStateIcon state={AlertStates.Pending} /> <strong>Pending: </strong>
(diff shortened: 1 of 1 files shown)


frontend/public/components/monitoring/alerting.tsx
```diff
@@ -282,17 +282,28 @@ const PopoverField: React.FC<{ body: React.ReactNode; label: string }> = ({ body
...
-    <dd>The alert condition is true, but the timeout has not been reached.</dd>
+    <dd>
+      The alert is firing because the alert condition is true and the optional `for` duration has
+      passed. The alert will continue to fire as long as the condition remains true.
+    </dd>
     <dt>
       <AlertStateIcon state={AlertStates.Silenced} /> <strong>Silenced: </strong>
     </dt>
-    <dd>The alert is now silenced.</dd>
+    <dt></dt>
```
spadgett, line 301: Did you mean to add an empty `dt`?

kyoto: Yeah, I checked and technically, HTML5 doesn't allow `dd` without a `dt`.

https://html.spec.whatwg.org/#the-dd-element says `dd` can be used:
> After `dt` or `dd` elements inside `dl` elements.

Also, https://html.spec.whatwg.org/#the-dl-element says that `dl` contains:
> Zero or more groups each consisting of one or more `dt` elements followed by one or more `dd` elements...

PR #6372, opened 2020-08-19
Bug 1866296: Cypress CRUD tests flake fixes

dtaylor113:
Ran Cypress locally many times, was able to see and fix some flakes:

1. `yamlEditorReady` Cypress env var seemed to have a race condition; removed it.  Yaml Editor is completely loaded when it is color coded, added test for that.  Had to test by `class` as we can't add an ID due to it being a 3rd party component.
2. When using url with query params for Search page, noticed page would load initially with just name filter/chip, and show the resource so test would continue, then page would reload and display with all three query params/filter chip groups (resource, label, name).  This would cause flakes.  Fixed by looking for all three chips.  Had to search by `class` as it's a PF component.

### DIFF
--- a/frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
@@ -82,6 +82,8 @@ describe('Kubernetes resource CRUD operations', () => {
         if (resourcesWithCreationForm.has(kind)) {
           cy.byTestID('yaml-link').click();
         }
+        // sidebar needs to be fully loaded, else it sometimes overlays the Create button
+        cy.byTestID('resource-sidebar').should('exist');
         yamlEditor.isLoaded();
         let newContent;
         // get, update, and set yaml editor content.
@@ -143,6 +145,9 @@ describe('Kubernetes resource CRUD operations', () => {
             namespaced ? `ns/${testName}` : 'all-namespaces'
           }?kind=${kind}&q=${testLabel}%3d${testName}&name=${name}`,
         );
+
+        // filter should have 3 chip groups: resource, label, and name
+        listPage.filter.numberOfActiveFiltersShouldBe(3);
         listPage.rows.shouldExist(name);
 
         cy.log('link to to details page');
--- a/frontend/packages/integration-tests-cypress/views/list-page.ts
+++ b/frontend/packages/integration-tests-cypress/views/list-page.ts
@@ -24,6 +24,9 @@ export const listPage = {
     byName: (name: string) => {
       cy.byLegacyTestID('item-filter').type(name);
     },
+    numberOfActiveFiltersShouldBe: (numFilters: number) => {
+      cy.get("[class='pf-c-toolbar__item pf-m-chip-group']").should('have.length', numFilters);
+    },
   },
   rows: {
     shouldBeLoaded: () => {
--- a/frontend/packages/integration-tests-cypress/views/yaml-editor.ts
+++ b/frontend/packages/integration-tests-cypress/views/yaml-editor.ts
@@ -10,6 +10,8 @@ export const setEditorContent = (text: string) => {
   });
 };
 
-export const isLoaded = () => cy.window().should('have.property', 'yamlEditorReady', true);
+// initially yamlEditor loads with all grey text, finished loading when editor is color coded
+// class='mtk26' is the light blue color of property such as 'apiVersion'
+export const isLoaded = () => cy.get("[class='mtk26']").should('exist');
 export const clickSaveCreateButton = () => cy.byTestID('save-changes').click();
 export const clickReloadButton = () => cy.byTestID('reload-object').click();
--- a/frontend/public/components/edit-yaml.jsx
+++ b/frontend/public/components/edit-yaml.jsx
@@ -196,9 +196,6 @@ export const EditYAML_ = connect(stateToProps)(
 
       loadYaml(reload = false, obj = this.props.obj) {
         if (this.state.initialized && !reload) {
-          if (window.Cypress) {
-            window.yamlEditorReady = true;
-          }
           return;
         }
 
--- a/frontend/public/declarations.d.ts
+++ b/frontend/public/declarations.d.ts
@@ -46,7 +46,6 @@ declare interface Window {
   store?: {}; // Redux store
   pluginStore?: {}; // Console plugin store
   Cypress?: {};
-  yamlEditorReady?: boolean;
 }
 
 // From https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html


spadgett:
> When using url with query params for Search page, noticed page would load initially with just name filter/chip, and show the resource so test would continue, then page would reload and display with all three query params/filter chip groups (resource, label, name). This would cause flakes. Fixed by looking for all three chips. Had to search by class as it's a PF component.

@dtaylor113 When you say reload, do you mean the entire page reloaded in the browser? Or just that the page updated? We might want to dig into this and see if there is a bug here. We do have a bug tracking problems with the name filter.

frontend/packages/integration-tests-cypress/support/selectors.ts
```diff
@@ -14,7 +14,7 @@ declare global {
 
 Cypress.Commands.add('byTestID', (selector: string) => cy.get(`[data-test="${selector}"]`));
 Cypress.Commands.add('byTestActionID', (selector: string) =>
-  cy.get(`[data-test-action="${selector}"]:not(.pf-m-disabled)`),
+  cy.get(`[data-test-action="${selector}"]:not(.pf-m-disabled)`, { timeout: 90000 }),
```
spadgett, line 17: Do we know if increasing the timeout actually prevents the flake? Or is there some error that prevents the menu items from enabling? I don't think it should take nearly this long, so I'm concerned about increasing the timeout.

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
```diff
@@ -143,6 +145,9 @@ describe('Kubernetes resource CRUD operations', () => {
             namespaced ? `ns/${testName}` : 'all-namespaces'
           }?kind=${kind}&q=${testLabel}%3d${testName}&name=${name}`,
         );
+
+        // filter should have 3 chip groups: resource, label, and name
+        cy.get("[class='pf-c-toolbar__item pf-m-chip-group']").should('have.length', 3);
```
spadgett, line 150: Ideally this selector would move to a view file.

frontend/packages/integration-tests-cypress/support/selectors.ts
dtaylor113: I notice when I run Cypress locally there is a noticeable pause when the kabab button is clicked to when the menu items render so increasing the timeout seemed like a good possible solution.

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
dtaylor113: > @dtaylor113 When you say reload, do you mean the entire page reloaded in the browser? Or just that the page updated? We might want to dig into this and see if there is a bug here. We do have a bug tracking problems with the name filter.

I 'think' it was the page updating.  It initially loaded with name filter in place, then I saw the three dot (table?) loading indicators, then the table with the single resource reappeared.

dtaylor113:
> @dtaylor113 When you say reload, do you mean the entire page reloaded in the browser? Or just that the page updated? We might want to dig into this and see if there is a bug here. We do have a bug tracking problems with the name filter.

I 'think' it was the page updating.  It initially loaded with name filter in place, then I saw the three dot (table?) loading indicator, then the table with the single resource reappeared.

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
dtaylor113: ok, fixed

dtaylor113:
@spadgett, added: [set all cy.visit(...)'s to use retryOnStatusCodeFailure: true](https://github.com/openshift/console/pull/6372/commits/5bf3d4c90df42c15895a4158ad717f5912dbb6af)

frontend/packages/integration-tests-cypress/support/selectors.ts
spadgett: > I notice when I run Cypress locally there is a noticeable pause when the kabab button is clicked to when the menu items render so increasing the timeout seemed like a good possible solution.

But 90 seconds? That seems way longer than we need. If it takes that long, that's a kebab bug we need to fix imo.

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -9,6 +9,10 @@ Cypress.Cookies.defaults({
   whitelist: ['openshift-session-token', 'csrf-token'],
 });
 
+Cypress.Commands.overwrite('visit', (originalFn, url, options) => {
+  return originalFn(url, { retryOnStatusCodeFailure: true, ...options });
```
spadgett, line 13: I'm debating if we should do this or not. I wouldn't expect to have a status code failure, although I guess a problem with ingress could cause it. We might want to dig into the failures more closely before making this change.

dtaylor113: I'll look for network failures, but what is the harm in retrying on failure?

spadgett: > I'll look for network failures, but what is the harm in retrying on failure?

It masks real bugs. If we have a backend or operator PR that sometimes causes console to respond with an error, we won't catch it.

spadgett: I would leave this out for now until we understand the problem better.

dtaylor113: Ok, it is removed.

spadgett:
@dtaylor113 Can you remove the 90s timeout on the actions as well? That's far too long. We need the tests to fail if it takes that long.

dtaylor113:
> @dtaylor113 Can you remove the 90s timeout on the actions as well? That's far too long. We need the tests to fail if it takes that long.

@spadgett, default is 30, scale down to 60 or totally remove?

dtaylor113:
> @dtaylor113 Can you remove the 90s timeout on the actions as well? That's far too long. We need the tests to fail if it takes that long.

Hi @spadgett, changed the timeout to 60s for actions, that might give the test the extra time it needs.

frontend/packages/integration-tests-cypress/support/selectors.ts
```diff
@@ -14,7 +14,7 @@ declare global {
 
 Cypress.Commands.add('byTestID', (selector: string) => cy.get(`[data-test="${selector}"]`));
 Cypress.Commands.add('byTestActionID', (selector: string) =>
-  cy.get(`[data-test-action="${selector}"]:not(.pf-m-disabled)`),
+  cy.get(`[data-test-action="${selector}"]:not(.pf-m-disabled)`, { timeout: 60000 }),
```
spadgett, line 17: This is still way, way too long. If you were a user, you wouldn't wait a full minute for the action to enable :) If this is happening, we want the test to fail.

dtaylor113: ok, timeout has been removed.

spadgett:
> level=fatal msg="failed to initialize the cluster: Cluster operator authentication is reporting a failure: WellKnownReadyControllerDegraded: need at least 3 kube-apiservers, got 2"

/retest

PR #6374, opened 2020-08-19
Bug 1872340: Fix PVC E2E tests

bipuladh:
- Adjusted to changes of Size unit in PVC details Page.
 - Adjusted to changes in the PVC expansion default value and changes in the selector.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/integration-tests/tests/2-tests/pvc.scenario.ts
+++ b/frontend/packages/ceph-storage-plugin/integration-tests/tests/2-tests/pvc.scenario.ts
@@ -49,7 +49,7 @@ describe('Test PVC creation with options.', () => {
       accessMode: VOLUME_ACCESS_MODES.RWO,
     };
     await createNewPersistentVolumeClaim(testPvc, true);
-    expect(pvcSize.getText()).toEqual('2Gi');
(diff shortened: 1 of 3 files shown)


cloudbehl:
@bipuladh Add Bugzilla

frontend/packages/ceph-storage-plugin/integration-tests/tests/2-tests/pvc.scenario.ts
```diff
@@ -49,7 +49,7 @@ describe('Test PVC creation with options.', () => {
       accessMode: VOLUME_ACCESS_MODES.RWO,
     };
     await createNewPersistentVolumeClaim(testPvc, true);
-    expect(pvcSize.getText()).toEqual('2Gi');
+    expect(pvcSize.getText()).toEqual('2 GiB');
```
ebondare, line 52: Could we have different values for different OCP versions? It would be better not to break the tests for OCP 4.5, I think.

bipuladh: @ebondare https://github.com/openshift/console/blob/release-4.5/frontend/packages/ceph-storage-plugin/integration-tests/tests/2-tests/pvc.scenario.ts#L52
It is already present in 4.5 release branch.

PR #6375, opened 2020-08-19
Bug 1870174: Add start VM option during CDI import

pcbailey:
This PR adds the "Start Virtual Machine" option back to the actions menus for VMs that are in the CDI import process. It also adds a modal warning to notify the user that the VM won't start until after the import completes and that the choice cannot be changed.

Screenshot of modal:
Selection_680

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
@@ -90,12 +90,29 @@ export const menuActionStart = (
 ): KebabOption => {
   const title = 'Start Virtual Machine';
   return {
-    hidden:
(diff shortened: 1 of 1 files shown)


frontend/packages/kubevirt-plugin/src/components/vms/menu-actions.tsx
```diff
@@ -90,12 +90,29 @@ export const menuActionStart = (
...
+            </>
+          ),
+          btnText: _.capitalize(VMActionType.Start),
+          executeFn: () => startVM(vm),
+        });
+      }
+    },
     accessReview: asAccessReview(kindObj, vm, 'patch'),
   };
 };
```
suomiy, line 118: are we still okay with keeping the stop isDisabled in this release?

anyway, can you please at least use isImporting status instead of isPending there?

PR #6377, opened 2020-08-19
Bug 1866875:  YAML Editor: Background resource update clears error message

cyril-ui-developer:
/assign @TheRealJon
/cc @spadgett

### DIFF
--- a/frontend/public/components/edit-yaml.jsx
+++ b/frontend/public/components/edit-yaml.jsx
@@ -122,9 +122,6 @@ export const EditYAML_ = connect(stateToProps)(
         this.setState({ stale });
         if (nextProps.error) {
           this.handleError(nextProps.error);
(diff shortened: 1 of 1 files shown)


frontend/public/components/edit-yaml.jsx
```diff
@@ -122,9 +122,6 @@ export const EditYAML_ = connect(stateToProps)(
         this.setState({ stale });
         if (nextProps.error) {
           this.handleError(nextProps.error);
-        } else if (this.state.error) {
-          //clear stale error state
-          this.setState({ error: '' });
```
spadgett, line 127: Do errors still get cleared when you click Reload? We'd want to clear errors in that case.

cyril-ui-developer:
@spadgett Yes! The errors gets cleared on click of Reload. Thanks

PR #6381, opened 2020-08-19
Bug 1870360: Fix toolbar spacing issue on search page

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1870360

If both hide toolbar and hide label filters are enabled, then be sure to hide the entire div that houses them both.

Before:
Screen Shot 2020-08-19 at 4 53 49 PM

After:
Screen Shot 2020-08-19 at 4 50 53 PM

### DIFF
--- a/frontend/public/components/factory/list-page.jsx
+++ b/frontend/public/components/factory/list-page.jsx
@@ -59,7 +59,7 @@ export const TextFilter = (props) => {
 TextFilter.displayName = 'TextFilter';
 
 // TODO (jon) make this into "withListPageFilters" HOC
-/** @augments {React.PureComponent<{ListComponent: React.ComponentType<any>, kinds: string[], filters?:any, flatten?: function, data?: any[], rowFilters?: any[], hideToolbar?: boolean, hideLabelFilter?: boolean, columnLayout?: ColumnLayout }>} */
+/** @augments {React.PureComponent<{ListComponent: React.ComponentType<any>, kinds: string[], filters?:any, flatten?: function, data?: any[], rowFilters?: any[], hideNameLabelFilters?: boolean, hideLabelFilter?: boolean, columnLayout?: ColumnLayout }>} */
 export class ListPageWrapper_ extends React.PureComponent {
   render() {
     const {
@@ -68,7 +68,7 @@ export class ListPageWrapper_ extends React.PureComponent {
       reduxIDs,
       rowFilters,
       textFilter,
-      hideToolbar,
+      hideNameLabelFilters,
       hideLabelFilter,
       columnLayout,
     } = this.props;
@@ -79,7 +79,7 @@ export class ListPageWrapper_ extends React.PureComponent {
         data={data}
         reduxIDs={reduxIDs}
         textFilter={textFilter}
-        hideToolbar={hideToolbar}
+        hideNameLabelFilters={hideNameLabelFilters}
         hideLabelFilter={hideLabelFilter}
         columnLayout={columnLayout}
         {...this.props}
@@ -107,7 +107,7 @@ ListPageWrapper_.propTypes = {
   rowFilters: PropTypes.array,
   staticFilters: PropTypes.array,
   customData: PropTypes.any,
-  hideToolbar: PropTypes.bool,
+  hideNameLabelFilters: PropTypes.bool,
   hideLabelFilter: PropTypes.bool,
 };
 
@@ -306,7 +306,7 @@ FireMan_.propTypes = {
   title: PropTypes.string,
 };
 
-/** @type {React.SFC<{ListComponent: React.ComponentType<any>, kind: string, helpText?: any, namespace?: string, filterLabel?: string, textFilter?: string, title?: string, showTitle?: boolean, rowFilters?: any[], selector?: any, fieldSelector?: string, canCreate?: boolean, createButtonText?: string, createProps?: any, mock?: boolean, badge?: React.ReactNode, createHandler?: any, hideToolbar?: boolean, hideLabelFilter?: boolean, columnLayout?: ColumnLayout, customData?: any} >} */
+/** @type {React.SFC<{ListComponent: React.ComponentType<any>, kind: string, helpText?: any, namespace?: string, filterLabel?: string, textFilter?: string, title?: string, showTitle?: boolean, rowFilters?: any[], selector?: any, fieldSelector?: string, canCreate?: boolean, createButtonText?: string, createProps?: any, mock?: boolean, badge?: React.ReactNode, createHandler?: any, hideNameLabelFilters?: boolean, hideLabelFilter?: boolean, columnLayout?: ColumnLayout, customData?: any, hideColumnManagement?: boolean } >} */
 export const ListPage = withFallback((props) => {
   const {
     autoFocus,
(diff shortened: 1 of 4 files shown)


frontend/public/components/filter-toolbar.tsx
```diff
@@ -294,49 +294,51 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
-        <ToolbarItem className="co-filter-search--full-width">
-          <ToolbarFilter
-            deleteChipGroup={() => updateLabelFilter([])}
-            chips={[...labelFilters]}
-            deleteChip={(filter, chip: string) =>
-              updateLabelFilter(_.difference(labelFilters, [chip]))
-            }
-            categoryName="Label"
-          >
+        {!hideToolbar && !hideLabelFilter && (
```
spadgett, line 297: Won't this hide the name filter when only `hideLabelFilter` is specified?

jcaianirh: @spadgett updated logic here.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -294,49 +293,51 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
-        <ToolbarItem className="co-filter-search--full-width">
-          <ToolbarFilter
-            deleteChipGroup={() => updateLabelFilter([])}
-            chips={[...labelFilters]}
-            deleteChip={(filter, chip: string) =>
-              updateLabelFilter(_.difference(labelFilters, [chip]))
-            }
-            categoryName="Label"
-          >
+        {(!hideToolbar || !hideLabelFilter) && (
```
rhamilto, line 296: This should be
```suggestion
        {!hideToolbar && (
```
since you could still have a `Name` filter by itself, no?

frontend/public/components/filter-toolbar.tsx
```diff
@@ -294,49 +293,51 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
+                <div className="pf-c-input-group">
+                  {!hideLabelFilter && (
+                    <DropdownInternal
+                      items={FilterType}
+                      onChange={switchFilter}
+                      selectedKey={filterType}
+                      title={filterType}
+                    />
+                  )}
+                  {!hideToolbar && (
```
rhamilto, line 320: This check is no longer needed with the addition of the new one above.

frontend/public/components/filter-toolbar.tsx
rhamilto: Wondering if `hideToolbar` should return anything at all?  With these changes, we still get empty markup:

```
<div class="pf-c-toolbar" id="filter-toolbar"><div class="pf-c-toolbar__content"><div class="pf-c-toolbar__content-section"></div><div class="pf-c-toolbar__expandable-content" id="filter-toolbar-expandable-content-47"><div class="pf-c-toolbar__group"></div></div></div><div class="pf-c-toolbar__content pf-m-hidden" hidden=""><div class="pf-c-toolbar__group"></div></div></div>
```

rhamilto: Oh, I see that's there because of `columnLayout`.  Is there ever `columnLayout` without the filter(s)?

jcaianirh: @rhamilto it looks like both row filters and col management are both available, as this only seems to apply to the name/label filter as of now.

jcaianirh: @rhamilto so the markup you are showing is expected...whether or not thats desired is another thing.

rhamilto: `hideToolbar` suggests the toolbar will not be present, so outputting empty markup seems kind of silly and is generally something we try to avoid.  If it is deemed necessary, it seems like renaming the prop is in order (maybe `hideFilters`?).

jcaianirh: @rhamilto there may be empty markup if there are no row filters for the resource. So I think this is the intent as far as I can tell. We have a name/label filter at the top of the search page, right? That is available for all resources in the search page. Each resource, however, may have additional resource specific filters that are not available in the search toolbar, thus those filters are available. I can add code to remove column management...I don't think its important on this page. If you think the name of the variable should change, I can do that so that it reads hideNameLabelFilters since that is what it does.

rhamilto: I vote for a prop name change.  `hideToolbar` is misleading.

frontend/public/components/filter-toolbar.tsx
jcaianirh: done

frontend/public/components/filter-toolbar.tsx
jcaianirh: @rhamilto changed name of prop, added a new one for col management, and updated logic

frontend/public/components/search.tsx
```diff
@@ -63,7 +63,8 @@ const ResourceList = connectToModel(({ kindObj, mock, namespace, selector, nameF
       mock={mock}
       badge={getBadgeFromType(kindObj.badge)}
       hideLabelFilter
```
rhamilto, line 65: You shouldn't need both `hideLabelFilter` and `hideNameLabelFilters`

jcaianirh: @rhamilto you can have a name filter without the label filter, so there is a case for this.

jcaianirh: @rhamilto for this use case, though, you are right, both are not needed.

frontend/public/components/factory/list-page.jsx
```diff
@@ -331,7 +331,8 @@ export const ListPage = withFallback((props) => {
     textFilter,
     match,
     badge,
-    hideToolbar,
+    hideNameLabelFilters,
```
rhamilto, line 334: Nit:  put `hideNameLabelFilters` and `hideLabelFilter` on adjoining lines here and elsewhere so you can see they're related?

frontend/public/components/filter-toolbar.tsx
```diff
@@ -224,7 +225,7 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
 
   const clearAll = () => {
     updateRowFilterSelected(selectedRowFilters);
-    if (!hideToolbar) {
+    if (!hideNameLabelFilters || !hideLabelFilter) {
```
rhamilto, line 228:
```suggestion
    if (!hideNameLabelFilters) {
```
since this is only calling `updateNameFilter`, no?

frontend/public/components/filter-toolbar.tsx
```diff
@@ -224,7 +225,7 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
       updateNameFilter('');
     }
     if (!hideLabelFilter) {
```
rhamilto, line 231:
```suggestion
    if (!hideNameLabelFilters || !hideLabelFilter) {
```
since the label filter can be hidden by both, no?

PR #6386, opened 2020-08-20
 Bug 1872280: Rename GitOps in nav to Application Stages

divyanshiGupta:
**Jira:** https://issues.redhat.com/browse/ODC-4503

**Screenshot:**
image

### DIFF
--- a/frontend/packages/dev-console/src/plugin.tsx
+++ b/frontend/packages/dev-console/src/plugin.tsx
@@ -168,21 +168,6 @@ const plugin: Plugin<ConsumedExtensions> = [
       perspective: 'dev',
       group: 'top',
       componentProps: {
-        name: 'GitOps',
-        href: '/gitops',
-        testID: 'gitops-header',
-      },
-    },
-    flags: {
-      required: [FLAG_OPENSHIFT_GITOPS],
-    },
-  },
-  {
-    type: 'NavItem/Href',
-    properties: {
-      perspective: 'dev',
-      group: 'top',
-      componentProps: {
         name: 'Monitoring',
         href: '/dev-monitoring',
         testID: 'monitoring-header',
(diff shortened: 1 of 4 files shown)


divyanshiGupta:
cc: @serenamarie125 @bkrikori

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -236,6 +221,21 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+      perspective: 'dev',
+      group: 'resources',
+      componentProps: {
+        name: 'Application Stages',
+        href: '/applicationstages',
+        testID: 'application-stages-header',
+      },
+    },
+    flags: {
+      required: [FLAG_OPENSHIFT_GITOPS],
```
sahil143, line 236: Why not rename the flag as well 
```suggestion
      required: [FLAG_OPENSHIFT_APPLICATION_STAGES],
```

sahil143 (changes requested):

jerolimov:
In general it lgtm for the "smallest possible change"..

I'm unsure if we want rename also all the GitOps files? I expect that the answer is no, but I do not want make the decision here... @christianvogt? :thinking:

christianvogt:
I think we're fine keeping the code named around 'gitops' as it's much more identifiable than 'application stages'

jerolimov:
If you create and link a bugzilla ticket this will get a lgtm from me. :smile:

frontend/packages/dev-console/src/plugin.tsx
divyanshiGupta: The operator is still GitOps and the flag can also be used at multiple places and not just for `Application Stages` page.

divyanshiGupta:
cc: @jerolimov

jerolimov:
:+1:

/lgtm

PR #6388, opened 2020-08-20
Bug 1870630: Monitoring: Replace `Inactive` with `-` for alerting rules

kyoto:
`Inactive` was considered confusing because it looked like it
represented an additional alert state.


spadgett:
I'm unsure why the frontend test failed.

/retest

PR #6390, opened 2020-08-20
Bug 1870880: add Rohit as a console frontend approver

christianvogt:
Dev console team is growing and we're making more changes touching files outside of dev-console. Some changes are in`console-shared`. New console wide features are going into `console-app`. Building on existing features for multiple perspectives in `public`.

We need additional approvers so ensure our PR backlog can be attended to in a timely manner.

cc @spadgett @rhamilto

### DIFF
--- a/frontend/OWNERS
+++ b/frontend/OWNERS
@@ -15,5 +15,6 @@ approvers:
   - kyoto
   - rawagner
   - rhamilto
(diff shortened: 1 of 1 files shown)


spadgett:
@christianvogt we'll need a bugzilla bug to make the bot happy :/

PR #6391, opened 2020-08-20
fixes inconsistent pod donut status for knsvc

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4514

**Analysis / Root cause**: 
Pos status text mismatch for topology view and sidebar

**Solution Description**: 
Added right kind as passed for sidebar

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/topology/components/nodes/PodSet.tsx
```diff
@@ -77,7 +77,7 @@ const PodSet: React.FC<PodSetProps> = ({ size, data, x = 0, y = 0, showPodCount
   );
   const hpaControlledScaling = !!hpa;
 
-  const obj = get(data, ['current', 'obj'], null) || data.dc;
+  const obj = RevisionModel.kind === data.dc.kind ? data.dc : data.current.obj || data.dc;
```
jeff-phillips-18, line 80:
```suggestion
  const obj = RevisionModel.kind === data.dc.kind ? data.dc : data?.current?.obj || data.dc;
```

invincibleJai: Thanks Jeff, Data will always be there it's required prop or we see scenario where it can be undefined?

PR #6393, opened 2020-08-20
Bug 1862500: Use initialization-resource annotation for yaml

zherman0:
If the initialization-resource annotation value is set for a CR, use it on the yaml pages instead of the alm-example annotation.

Screenshot_2020-08-20 Create StorageCluster · OKD

Note to self: For testing, use TW's operators that have CR values set in the initialization-resource annotation and then do a patch to change the values from the alm-example: `oc patch csv portworx-operator.v1.3.1 --patch "$(cat ~/openshift/portworx-test.json)" --type=merge`
and portworx-test.json looks like:
```
{
  "metadata": {
    "annotations": {
      "operatorframework.io/initialization-resource": "{\n    \"apiVersion\": \"core.libopenstorage.org/v1alpha1\",\n    \"kind\": \"StorageCluster\",\n    \"metadata\": {\n      \"name\": \"portworx-initvalue\",\n      \"namespace\": \"test-operator\"\n    },\n    \"spec\": {\n      \"userInterface\": {\n        \"enabled\": true\n      },\n      \"autopilot\": {\n        \"enabled\": false\n      }\n    }\n  }\n"
    }
  }
}
```

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/index.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/index.tsx
@@ -112,8 +112,17 @@ export const providedAPIForModel = (csv: ClusterServiceVersionKind, model: K8sKi
     (crd) => referenceForProvidedAPI(crd) === referenceForModel(model),
   );
 
-export const parseALMExamples = (csv: ClusterServiceVersionKind) => {
+export const parseALMExamples = (
+  csv: ClusterServiceVersionKind,
+  useInitializationResource: boolean,
+) => {
   try {
+    if (useInitializationResource) {
+      const resource = JSON.parse(
+        csv?.metadata?.annotations?.['operatorframework.io/initialization-resource'],
+      );
+      return [resource];
+    }
     return JSON.parse(csv?.metadata?.annotations?.['alm-examples'] ?? '[]');
   } catch (e) {
     // eslint-disable-next-line no-console
@@ -130,7 +139,10 @@ export const exampleForModel = (csv: ClusterServiceVersionKind, model: K8sKind) 
       apiVersion: `${model.apiGroup}/${model.apiVersion}`,
     },
     _.find(
-      parseALMExamples(csv),
+      parseALMExamples(
+        csv,
+        new URLSearchParams(window.location.search).has('useInitializationResource'),
+      ),
       (s: K8sResourceKind) => referenceFor(s) === referenceForModel(model),
     ),
   );
--- a/frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
+++ b/frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
@@ -223,7 +223,7 @@ export const getEventSourceConnectorList = (
   _.reduce(
     csvData,
     (acm, cv) => {
-      const parseCSVData = parseALMExamples(cv);
+      const parseCSVData = parseALMExamples(cv, false);
       _.forEach(parseCSVData, (res) => {
         if (
           _.findIndex(acm, res) === -1 &&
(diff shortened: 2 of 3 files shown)


frontend/packages/operator-lifecycle-manager/src/components/index.tsx
```diff
@@ -114,7 +114,11 @@ export const providedAPIForModel = (csv: ClusterServiceVersionKind, model: K8sKi
 
 export const parseALMExamples = (csv: ClusterServiceVersionKind) => {
   try {
-    return JSON.parse(csv?.metadata?.annotations?.['alm-examples'] ?? '[]');
+    return JSON.parse(
+      csv?.metadata?.annotations?.['operatorframework.io/initialization-resource'] ??
```
spadgett, line 118: The initialization resource is just a single object, right, not an array? This would return a different type.

spadgett, line 118: I have a feeling this is going to break the examples for other resources provided by the operator. Won't this always return the initialization resource sample for any operand?

zherman0: Isn't that the goal here?  If there is an initialization resource sample, then we want to show that over the alm-example since that is what we use to know we need a CR?  I did testing with other non CR CSVs and things looked good, but I will go do some more.

zherman0: Correct, I used an array in the init resource during my test data which is incorrect.  I have fixed that and the code.

spadgett: @zherman0 No, what I'm saying is if there are multiple provided APIs, this will always use the YAML for the initialization resource for all types instead of the YAML for the correct type. Can you check?

zherman0:
@spadgett - Reworked my logic and now I handle the combination of almExamples and Initialization-resource examples.

zherman0:
@spadgett - Switched logic to use URL flag. Ready for review.

frontend/packages/operator-lifecycle-manager/src/components/index.tsx
```diff
@@ -114,6 +114,11 @@ export const providedAPIForModel = (csv: ClusterServiceVersionKind, model: K8sKi
 
 export const parseALMExamples = (csv: ClusterServiceVersionKind) => {
   try {
+    if (new URLSearchParams(window.location.search).has('useInitializationResource')) {
+      return JSON.parse(
+        `[${csv?.metadata?.annotations?.['operatorframework.io/initialization-resource']}]` ?? '[]',
+      );
```
spadgett, line 120: Parse it as-is and then return it as an array.

```suggestion
      const resource = JSON.parse(csv?.metadata?.annotations?.['operatorframework.io/initialization-resource']);
      return [resource];
```

frontend/packages/operator-lifecycle-manager/src/components/index.tsx
```diff
@@ -114,6 +114,11 @@ export const providedAPIForModel = (csv: ClusterServiceVersionKind, model: K8sKi
 
 export const parseALMExamples = (csv: ClusterServiceVersionKind) => {
   try {
+    if (new URLSearchParams(window.location.search).has('useInitializationResource')) {
```
spadgett, line 117: This should be passed as a parameter from caller instead of read directly from the URL here. (This helper shouldn't know about the query parameter, which is specific to a certain page.)

zherman0:
@spadgett - Moved the check for the URL flag out of the parse helper function.

spadgett (approved):
/lgtm

PR #6397, opened 2020-08-21
Bug 1871174: Fix CR badges and operator link

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4471
https://issues.redhat.com/browse/ODC-4640

**Analysis / Root cause**: 
The CSV badge is not of the right color.
The title of Operator Backed Service and it's resources do not link to the Operator page in the context of the CSV

**Solution Description**: 
- Fix badge color in CSV sidebar and topology node.
- Fix links in CSV sidebar title and it's resources sidebars.
- Fix `Managed by ...` link in details page of resources managed by operator service.

rec0

Unchanged.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/operators/TopologyOperatorBackedPanel.tsx
+++ b/frontend/packages/dev-console/src/components/topology/operators/TopologyOperatorBackedPanel.tsx
@@ -1,22 +1,17 @@
 import * as React from 'react';
-import {
-  ResourceIcon,
-  ResourceLink,
-  SimpleTabNav,
-  ResourceSummary,
-  SectionHeading,
-} from '@console/internal/components/utils';
-import { referenceFor } from '@console/internal/module/k8s';
+import { SimpleTabNav, ResourceSummary, SectionHeading } from '@console/internal/components/utils';
 import TopologyOperatorBackedResources from './TopologyOperatorBackedResources';
 import { TopologyDataObject } from '../topology-types';
 import { OperatorGroupData } from './operator-topology-types';
+import { ManagedByOperatorResourceLink } from '@console/internal/components/utils/managed-by';
 
 export type TopologyOperatorBackedPanelProps = {
   item: TopologyDataObject<OperatorGroupData>;
 };
 
 const TopologyOperatorBackedPanel: React.FC<TopologyOperatorBackedPanelProps> = ({ item }) => {
-  const { name, resource, data } = item;
+  const { name, resource } = item;
+  const csvName = resource.metadata.selfLink.split('/').pop();
   const ResourcesSection = () => <TopologyOperatorBackedResources item={item} />;
   const DetailsSection = () => (
     <div className="overview__sidebar-pane-body">
@@ -30,12 +25,15 @@ const TopologyOperatorBackedPanel: React.FC<TopologyOperatorBackedPanelProps> = 
       <div className="overview__sidebar-pane-head resource-overview__heading">
         <h1 className="co-m-pane__heading">
           <div className="co-m-pane__name co-resource-item">
-            <ResourceIcon kind={data.operatorKind || 'Operator'} />
-            <ResourceLink
-              kind={referenceFor(resource)}
-              name={name}
+            <ManagedByOperatorResourceLink
+              csvName={csvName}
               namespace={resource.metadata.namespace}
-              hideIcon
+              owner={{
+                name,
+                kind: resource.kind,
+                uid: resource.metadata.uid,
+                apiVersion: resource.apiVersion,
+              }}
             />
           </div>
         </h1>
--- a/frontend/public/components/utils/managed-by.tsx
+++ b/frontend/public/components/utils/managed-by.tsx
@@ -10,30 +10,39 @@ import {
   referenceForModel,
   referenceForOwnerRef,
   OwnerReference,
+  modelFor,
 } from '../../module/k8s';
 import { findOwner, matchOwnerAndCSV } from '../../module/k8s/managed-by';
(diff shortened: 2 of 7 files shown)


invincibleJai:
@rottencandy in case of `tekton` i see issue now with this change

Aug-24-2020 17-30-26

rottencandy:
> @rottencandy in case of `tekton` i see issue now with this change
> 
> Aug-24-2020 17-30-26

@invincibleJai This issue will be handled separately as part of https://issues.redhat.com/browse/ODC-4640 since the component used affects other parts of the console as well.

frontend/packages/dev-console/src/components/topology/operators/TopologyOperatorBackedPanel.tsx
```diff
@@ -30,13 +49,8 @@ const TopologyOperatorBackedPanel: React.FC<TopologyOperatorBackedPanelProps> =
...
           <div className="co-m-pane__name co-resource-item">
-            <ResourceIcon kind={data.operatorKind || 'Operator'} />
-            <ResourceLink
-              kind={referenceFor(resource)}
-              name={name}
-              namespace={resource.metadata.namespace}
-              hideIcon
-            />
+            <ResourceIcon kind={referenceFor(resource)} />
+            <Link to={getOperatorLink(resource)}>{name}</Link>
```
rohitkrai03, line 53: You can reuse `ManagedByOperatorResourceLink` here instead of custom implementation.

rohitkrai03 (changes requested):

rohitkrai03:
> > @rottencandy in case of `tekton` i see issue now with this change
> > Aug-24-2020 17-30-26
> 
> @invincibleJai This issue will be handled separately as part of https://issues.redhat.com/browse/ODC-4640 since the component used affects other parts of the console as well.

@rottencandy I think it would be better to include the fix for that as well in this PR. https://issues.redhat.com/browse/ODC-4640 issue is not there in master right now which essentially means this PR is introducing it in first place. Not sure I am confortable with getting a fix in which in turn breaks something else and then handle it separately.

frontend/packages/dev-console/src/components/topology/operators/TopologyOperatorBackedPanel.tsx
rottencandy: Updated.

rottencandy:
> @rottencandy I think it would be better to include the fix for that as well in this PR. https://issues.redhat.com/browse/ODC-4640 issue is not there in master right now which essentially means this PR is introducing it in first place. Not sure I am confortable with getting a fix in which in turn breaks something else and then handle it separately.

The issue does appear in master, on the details page where `ManagedByOperatorLink` is used. And this issue seems to be exclusively occurring in `cluster` config in the `openshift-pipelines` namespace.

frontend/public/components/utils/managed-by.tsx
```diff
@@ -59,12 +59,14 @@ export const ManagedByOperatorLink: React.SFC<ManagedByLinkProps> = (props) => {
...
-          owner={owner}
-        />
+        <span className="co-resource-item">
+          Managed by{' '}
+          <ManagedByOperatorResourceLink
+            namespace={namespace}
+            csvName={csv.metadata.name}
+            owner={owner}
+          />
+        </span>
```
rohitkrai03, line 69: Better to add an optional `className` prop to the component and send it from here.
```suggestion
        Managed by{' '}
          <ManagedByOperatorResourceLink
            className="co-resource-item"
            namespace={namespace}
            csvName={csv.metadata.name}
            owner={owner}
          />
```

rohitkrai03:
> > @rottencandy I think it would be better to include the fix for that as well in this PR. https://issues.redhat.com/browse/ODC-4640 issue is not there in master right now which essentially means this PR is introducing it in first place. Not sure I am confortable with getting a fix in which in turn breaks something else and then handle it separately.
> 
> The issue does appear in master, on the details page where `ManagedByOperatorLink` is used. And this issue seems to be exclusively occurring only in `cluster` config in the `openshift-pipelines` namespace.

Okay, so the issue is on details page but this current PR fixes issue in Topology page and it's sidebar while introducing the same issue here as well which was not there before this fix. I would still prefer all the topology page links work correctly. The details page link can be fixed later on.

rohitkrai03 (changes requested):

frontend/public/components/utils/managed-by.tsx
rottencandy: Updated.

rottencandy:
> Okay, so the issue is on details page but this current PR fixes issue in Topology page and it's sidebar while introducing the same issue here as well which was not there before this fix. I would still prefer all the topology page links work correctly. The details page link can be fixed later on.

@rohitkrai03 I've included the fix for https://issues.redhat.com/browse/ODC-4640 in this PR. PTAL.

rohitkrai03 (approved):
/approve

invincibleJai:
@rottencandy overall works great Saud, in one scenario i see issue , for postgres in a particular namespace i don't see managed by

image

rottencandy:
> @rottencandy overall works great Saud, in one scenario i see issue , for postgres in a particular namespace i don't see managed by
> 
> image

@invincibleJai this was seemingly happening because the `Deployment` was created directly by the `CSV` and not through a `CRD` created by the operator. I've updated the code to handle this scenario. PTAL.

invincibleJai:
/lgtm

Verified the changes looks good, thanks @rottencandy :)

PR #6401, opened 2020-08-21
Bug 1856353: Include installed version in operator hub item details

TheRealJon:
Update operator hub item details to include the installed version when it is not the latest version.
Update "Operator Version" detail to "Latest Version" to convey that it is not the actual installed
version.

image

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
@@ -8,6 +8,10 @@ import { ExternalLink, HintBlock, Timestamp } from '@console/internal/components
 import { RH_OPERATOR_SUPPORT_POLICY_LINK } from '@console/shared';
 import { MarkdownView } from '../clusterserviceversion';
 import { OperatorHubItem } from './index';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import { referenceForModel } from '@console/internal/module/k8s';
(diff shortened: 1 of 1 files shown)


frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx
```diff
@@ -225,12 +252,17 @@ export const OperatorHubPage = withFallback(
...
                   isList: true,
                   kind: referenceForModel(SubscriptionModel),
-                  prop: 'subscription',
+                  prop: 'subscriptions',
+                },
+                {
+                  isList: true,
+                  kind: referenceForModel(ClusterServiceVersionModel),
+                  prop: 'clusterServiceVersions',
                 },
```
spadgett, line 266: This is a really expensive watch. CSVs contain a lot of data, and then they're copied across all namespaces. Are we able to wait until the user clicks the tile and get the single CSV we need to show the version?

TheRealJon: Yeah, we could feasibly do that.

TheRealJon:
> This is a really expensive watch. CSVs contain a lot of data, and then they're copied across all namespaces. Are we able to wait until the user clicks the tile and get the single CSV we need to show the version?

@spadgett Looking into this, but I'm having issues with the details pane re-rendering very frequently. I'm trying to pinpoint the cause of that.

TheRealJon:
/hold

Depends on https://github.com/openshift/console/pull/6456

PR #6402, opened 2020-08-23
Bug 1871103: Remove 'Template' field from VM wizard

irosenzw:
When creating a VM from template the wizard title will be:
`Create Virtual Machine from <template name> template`


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/vm-settings-tab.tsx
```diff
@@ -73,14 +71,6 @@ export const VMSettingsTabComponent: React.FC<VMSettingsTabComponentProps> = ({
           />
         </FormField>
       </FormFieldMemoRow>
```
suomiy, line 73: how will the user know which template was selected in `virtualization/templates` ?

irosenzw: what do you propose ? keep the template's dropdown or add a label (or another component) to show the chosen template ?

suomiy: I guess having a label should be enough. Although there will probably be some logic , when the template doesn't exist, etc. So having a component might be beneficial.

irosenzw:
Screenshot from 2020-08-26 15-34-23

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/user-templates.tsx
```diff
@@ -1,108 +0,0 @@
-import * as React from 'react';
-import { FormSelect, FormSelectOption } from '@patternfly/react-core';
-import { iGetIsLoaded, iGetLoadedData } from '../../../../utils/immutable';
-import { FormFieldRow } from '../../form/form-field-row';
-import { FormField, FormFieldType } from '../../form/form-field';
-import { ignoreCaseSort } from '../../../../utils/sort';
-import { VMSettingsField } from '../../types';
-import {
```
suomiy, line 8: what about all these strings?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -290,6 +291,7 @@ class CreateVMWizardComponent extends React.Component<CreateVMWizardComponentPro
 const wizardStateToProps = (state, { reduxID }) => ({
   isLastTabErrorFatal: isLastStepErrorFatal(state, reduxID),
   tabsMetadata: getStepsMetadata(state, reduxID),
+  userTemplateName: iGetCommonData(state, reduxID, VMWizardProps.userTemplateName),
   // fetch data from store to detect and fire changes
   ...[...DetectCommonDataChanges]
     .filter((v) => v !== VMWizardProps.storageClassConfigMap) // passed directly
```
suomiy, line 297: we don't have to load all the userTemplates anymore. Just the one is enough.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/vm-settings-tab.tsx
```diff
@@ -73,14 +71,6 @@ export const VMSettingsTabComponent: React.FC<VMSettingsTabComponentProps> = ({
...
-      <UserTemplates
-        userTemplateField={getField(VMSettingsField.USER_TEMPLATE)}
```
suomiy, line 77: we can remove VMSettingsField.USER_TEMPLATE field as it is not used anymore

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
suomiy: plus we don't have to load it at all when we are not in the userTemplateMode

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/vm-settings-tab.tsx
suomiy: plus all the not needed logic

suomiy (changes requested):

suomiy:
@matthewcarleton what is your opinion on this design and what should we do when the template is missing? (somebody has wrong url)

matthewcarleton:
> @matthewcarleton what is your opinion on this design and what should we do when the template is missing? (somebody has wrong url)

Having the template name in the title of the wizard works. We could also have 
Template
template-name-here
as the first label - read only input
WDYT @irosenzw ?

For when the template is missing 
for the short term I think we'll have to have an error in the wizard that alerts them of the issue and how to fix it. More long term we are talking about Template status so that could help avoid selecting a template that has this issue.
This is if I follow you correctly. I wasn't aware of this type of issue.

irosenzw:
@matthewcarleton can you please attach a design proposal so we can all be on the same page of what you are suggesting ?

suomiy:
> For when the template is missing
> for the short term I think we'll have to have an error in the wizard that alerts them of the issue and how to fix it. More long term we are talking about Template status so that could help avoid selecting a template that has this issue.
> This is if I follow you correctly. I wasn't aware of this type of issue.

I don't expect anything to be wrong template. I think the only issue will be that the user saved a template creation url and is trying to use, while that template has been already deleted

matthewcarleton:
> > For when the template is missing
> > for the short term I think we'll have to have an error in the wizard that alerts them of the issue and how to fix it. More long term we are talking about Template status so that could help avoid selecting a template that has this issue.
> > This is if I follow you correctly. I wasn't aware of this type of issue.
> 
> I don't expect anything to be wrong template. I think the only issue will be that the user saved a template creation url and is trying to use, while that template has been already deleted
This would happen in the UI?

suomiy:
yes, but not that likely - you go to create from a template and somebody deletes it

nevertheless we should add a guard against it

matthewcarleton:
@irosenzw just want to make sure you saw my last comment. It posted strangely - but there is a response there to @suomiy latest comment.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -89,6 +94,8 @@ const CreateVMWizardFooterComponent: React.FC<CreateVMWizardFooterComponentProps
         const areMainTabsHidden = VM_WIZARD_DIFFICULT_TABS.some((tab) => steps[tab].isHidden);
         const isLastStepValid = steps[VMWizardTab.RESULT].isValid;
 
+        const isValidUserTemplate = !userTemplateName || (userTemplateName && isUserTemplateValid);
```
suomiy, line 97: it would make more sense to call this isInvalidUserTemplate IMO

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -307,7 +310,11 @@ const wizardDispatchToProps = (dispatch, props) => ({
         data: {
           isCreateTemplate: props.isCreateTemplate,
           isProviderImport: props.isProviderImport,
-          userTemplateName: props.userTemplateName,
+          userTemplateName: {
+            value: props.userTemplateName,
+            initialized: false,
+            isValid: true,
```
suomiy, line 316: we should not load all the userTemplate, but just the one we need.

In that case we don't need isValid value here, because the validity is already present in FirehoseResult object


It is a fluid value. It could easily happen that we open wizard and then somebody deletes the userTemplate - and in that case the wizard stops being valid

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/resource-load-errors.tsx
```diff
@@ -10,6 +10,32 @@ import { K8sKind } from '@console/internal/module/k8s';
 import { TemplateModel } from '@console/internal/models';
 import { DataVolumeModel, VirtualMachineModel } from '../../../models';
 import { getExtraWSQueries } from '../selectors/selectors';
+import { iGetUserTemplateName, iGetUserTemplateValidity } from '../selectors/immutable/template';
+
+const userTemplateError = (state: any, wizardReduxID: string): Error => {
```
suomiy, line 15: if we load one use template in that case we don't need a special handler like this one but a generic handler like VMWizardProps.userTemplates have right now

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/internal-actions.ts
```diff
@@ -37,6 +37,27 @@ export const vmWizardInternalActions: VMWizardInternalActions = {
     },
     type: InternalActionType.Update,
   }),
+  [InternalActionType.UpdateUserTemplateName]: (id, value) => ({
```
suomiy, line 40: that is too many actions and it will not scale. Please create just a generic action like `setCommonDataDataValue`

suomiy: and it seems you are dispatching it in just one place, so you don't need such granularity

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -50,36 +51,32 @@ const selectUserTemplateOnLoadedUpdater = (options: UpdateOptions) => {
...
+  dispatch(vmWizardInternalActions[InternalActionType.UpdateUserTemplateInit](id, true));
+  dispatch(vmWizardInternalActions[InternalActionType.UpdateUserTemplateValidity](id, true));
 };
 
 const selectedUserTemplateUpdater = (options: UpdateOptions) => {
   const { id, prevState, dispatch, getState } = options;
   const state = getState();
-  if (!hasVmSettingsChanged(prevState, state, id, VMSettingsField.USER_TEMPLATE)) {
+
+  if (!hasCommonDataChanged(prevState, state, id, VMWizardProps.userTemplateName)) {
```
suomiy, line 66: we don't expect the name to change so the logic has to be redone.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -183,13 +181,13 @@ const cloneCommonBaseDiskImageUpdater = ({ id, prevState, dispatch, getState }:
       state,
       id,
       VMSettingsField.CLONE_COMMON_BASE_DISK_IMAGE,
-      VMSettingsField.USER_TEMPLATE,
-    )
+    ) &&
+    !hasCommonDataChanged(prevState, state, id, VMWizardProps.userTemplateName)
```
suomiy, line 185: no need to have it here

suomiy (changes requested):

irosenzw:
Create a VM from an existing template:
valid-template


Create a VM from a non-existing template:
invalid-template

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -41,10 +37,9 @@ const validateVm: Validator = (field, options) => {
   const isCreateTemplate = iGetCommonData(state, id, VMWizardProps.isCreateTemplate);
 
   const entities = isCreateTemplate
-    ? concatImmutableLists(
-        iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates),
-        iGetLoadedCommonData(state, id, VMWizardProps.userTemplates),
-      )
+    ? concatImmutableLists(iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates), [
```
suomiy, line 40: this will not validate properly we need another prop (`userTemplates`) just for the templates flow - it should be needed just for this validation, if I am not mistaken.

also commonTemplates are not needed here

suomiy: Please do not revert the changes you have doen already for this.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,24 +71,30 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
       flavor,
       workload: workloadProfile,
       os,
     };
 
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const vanillaTemplates = isUserTemplateLoaded
+      ? immutableListToShallowJS(
```
suomiy, line 86: let's just use the the usertemplate here when selected and not mash them up with commontemplates

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -14,24 +15,32 @@ import { VMSettingsField } from '../../types';
...
   }) => {
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const isUserTemplateLoaded = iGetIsLoaded(userTemplate);
+    const userTemplateName: string = isUserTemplateLoaded
+      ? iGetName(iGetLoadedData(userTemplate))
+      : '';
+
+    const vanillaTemplates = isUserTemplateLoaded
```
suomiy, line 35: same here. Plus a bug with mixing immutable with vanillajs

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
```diff
@@ -261,7 +259,10 @@ export type CommonData = {
     isSimpleView?: boolean;
     isCreateTemplate?: boolean;
     isProviderImport?: boolean;
-    userTemplateName?: string;
+    userTemplateName?: {
```
suomiy, line 262: old version?

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
```diff
@@ -26,15 +24,12 @@ const flavorOrder = {
 };
 
 export const iGetRelevantTemplates = (
```
suomiy, line 26: as mentioned earlier, let's remove the iUserTemplate parameter

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -45,12 +45,14 @@ export const getOperatingSystems = (templates: TemplateKind[], userTemplate: str
 
 export const getWorkloadProfiles = (
   templates: TemplateKind[],
-  { flavor, os, userTemplate }: { flavor: string; os: string; userTemplate: string },
+  { flavor, os, userTemplateName }: { flavor: string; os: string; userTemplateName: string },
```
suomiy, line 48: we should get just the userTemplate here in templates and not have to look it up retrospectivelly through common templates + user template

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -63,12 +65,14 @@ export const getWorkloadProfiles = (
 
 export const getFlavors = (
```
suomiy, line 66: same here

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -27,59 +27,66 @@ import { ProvisionSource } from '../../../../constants/vm/provision-source';
...
+    );
+  }
 };
 
 const selectedUserTemplateUpdater = (options: UpdateOptions) => {
   const { id, prevState, dispatch, getState } = options;
   const state = getState();
-  if (!hasVmSettingsChanged(prevState, state, id, VMSettingsField.USER_TEMPLATE)) {
+
+  if (!hasCommonDataChanged(prevState, state, id, VMWizardProps.isUserTemplateInitialized)) {
```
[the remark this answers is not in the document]
irosenzw: move it or copy it ? if we move it what should be here instead ?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -27,59 +27,66 @@ import { ProvisionSource } from '../../../../constants/vm/provision-source';
...
 
 const selectUserTemplateOnLoadedUpdater = (options: UpdateOptions) => {
   const { id, dispatch, getState } = options;
   const state = getState();
 
   if (
-    iGetVmSettingAttribute(state, id, VMSettingsField.USER_TEMPLATE, 'initialized') ||
-    !options.changedCommonData.has(VMWizardProps.userTemplates)
+    iGetCommonData(state, id, VMWizardProps.isUserTemplateInitialized) ||
+    !iGetLoadedCommonData(state, id, VMWizardProps.userTemplate)
```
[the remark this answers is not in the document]
irosenzw: cause `!options.changedCommonData.has(VMWizardProps.userTemplate)` will always return `false`at the beginning thus the `isUserTemplateIniitialized` becomes `true` before we can pull the template and populate the fields in `selectedUserTemplateUpdater`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -131,14 +132,14 @@ class CreateVMWizardComponent extends React.Component<CreateVMWizardComponentPro
...
-    const { isCreateTemplate, isProviderImport } = this.props;
+    const { isCreateTemplate, isProviderImport, userTemplateName } = this.props;
     if (isCreateTemplate) {
       return CREATE_VM_TEMPLATE;
     }
     if (isProviderImport) {
       return IMPORT_VM;
     }
-    return CREATE_VM;
+    return userTemplateName ? `${CREATE_VM} from ${userTemplateName} template` : CREATE_VM;
```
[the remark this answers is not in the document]
matthewcarleton: yes we could include
template as the first field and a read only input with the name of the template

matthewcarleton:
> Create a VM from an existing template:
> valid-template
> 
> Create a VM from a non-existing template:
> invalid-template

What if we said

**Could not load requested template**
requested template vm-templates-example-5 can not be found. It may have been removed. 
Please contact your admin for more information or choose another template (this could link back to the template list)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/selectors/combined.ts
```diff
@@ -2,18 +2,18 @@ import { Map as ImmutableMap } from 'immutable';
...
-    ? getTemplateOperatingSystems(
-        immutableListToShallowJS<TemplateKind>(
-          concatImmutableLists(iUserTemplates, iCommonTemplates),
-        ),
-      )
+    ? getTemplateOperatingSystems(immutableListToShallowJS<TemplateKind>(iCommonTemplates))
     : operatingSystemsNative;
 
+  if (openshiftFlag) {
+    operatingSystems.push(iUserTemplate.toJS());
```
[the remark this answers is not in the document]
suomiy: can we refactor this function so that we gather info only from `iUserTemplate` when we have it. Otherwise from `iCommonTemplates`?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,34 +71,45 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
     const cloneBaseDiskImage = iGetFieldValue(cloneBaseDiskImageField);
     const mountWindowsGuestTools = iGetFieldValue(mountWindowsGuestToolsField);
+    const isUserTemplateValid = iGetIsLoaded(iUserTemplate) && !iGetLoadError(iUserTemplate);
+    const userTemplate = isUserTemplateValid ? iGetLoadedData(iUserTemplate).toJS() : null;
+    const userTemplateName: string = isUserTemplateValid
+      ? iGetName(iGetLoadedData(iUserTemplate))
+      : '';
     const params = {
-      userTemplate,
+      userTemplateName,
```
suomiy, line 80: can we remove it from params. It does not seem necessary to have it there.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,34 +71,45 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
       os,
     };
 
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const vanillaTemplates = immutableListToShallowJS(iGetLoadedData(commonTemplates));
+
+    if (isUserTemplateValid) {
+      vanillaTemplates.push(userTemplate);
```
suomiy, line 89: lets just use the one userTemplate when resolving os/flavor/workload

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -14,24 +15,32 @@ import { VMSettingsField } from '../../types';
...
     operatingSystem,
     flavor,
     cnvBaseImages,
     onChange,
   }) => {
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const isUserTemplateLoaded = iGetIsLoaded(iUserTemplate);
+    const userTemplateName: string = isUserTemplateLoaded
```
suomiy, line 31: same - no need to resolve the name here

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
```diff
@@ -230,16 +231,16 @@ export type CommonDataProp =
...
+  | VMWizardProps.isUserTemplateInitialized
   | ChangedCommonDataProp;
 
 export type ChangedCommonData = Set<ChangedCommonDataProp>;
 
 export const DetectCommonDataChanges = new Set<ChangedCommonDataProp>([
   VMWizardProps.activeNamespace,
   VMWizardProps.openshiftFlag,
+  VMWizardProps.userTemplate,
   VMWizardProps.virtualMachines,
```
suomiy, line 243: VMWizardProps.userTemplates are missing here

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
suomiy: repeating myself, but let's not pass the name but just the `[template]` for userTemplates

suomiy (changes requested):

suomiy:
> What if we said

> Could not load requested template
> requested template vm-templates-example-5 can not be found. It may have been removed.
> Please contact your admin for more information or choose another template (this could link back to the template list)

@matthewcarleton 

We could do this, but I am against it at this time. This component handles all types of errors (of all resources and casuses) and if we start adding customization like this one on top of each other iit will become a mess.

We have to come either with a better generic solution to these errors or add a good extension framework. 
If we try to do it here, this PR will get more out of scope.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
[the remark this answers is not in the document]
suomiy: IMO I quite like that it is part of the header since it is an identifying info of the wizard and you can see it from all the tabs. Can we try to play with the idea of a header a bit longer?

what about 
`Create VM from vm-template-example`

or 

`Create Virtual Machine from vm-template-example`

or some kind of highligting.

Nevertheless, we still could opt to not to do anything and keep the header as it is.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
suomiy: it worked before. You can set it up so it works. You just need to play with `changedCommonData.has` and `isLoaded`. Let me know if something still does not add up.

please also see https://github.com/openshift/console/pull/6402#discussion_r485127998

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -117,11 +120,12 @@ const CreateVMWizardFooterComponent: React.FC<CreateVMWizardFooterComponentProps
         const isFinishingStep = [VMWizardTab.REVIEW, VMWizardTab.RESULT].includes(activeStepID);
         const isLastStep = activeStepID === VMWizardTab.RESULT;
 
-        const isNextButtonDisabled = isAnyStepLocked;
-        const isReviewButtonDisabled = isAnyStepLocked;
+        const isNextButtonDisabled = isAnyStepLocked || isInvalidUserTemplate;
```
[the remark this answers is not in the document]
irosenzw: first, I don't understand why it won't disable the left nav when needed, can you provide an example.
second, can we maybe add the functionality as a followup ?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -27,59 +27,46 @@ import { ProvisionSource } from '../../../../constants/vm/provision-source';
...
-  ) {
-    return;
-  }
-
-  const userTemplateName = iGetCommonData(state, id, VMWizardProps.userTemplateName);
-  if (!userTemplateName) {
-    return;
+  const iUserTemplate = iGetLoadedCommonData(state, id, VMWizardProps.userTemplate);
+  if (iUserTemplate && iGetLoadError(iGetCommonData(state, id, VMWizardProps.userTemplate))) {
+    dispatch(
```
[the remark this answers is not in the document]
irosenzw: This is for the use case when a template is deleted while you try to create a VM from that template - then we reset the initialization variable.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
irosenzw: two reasons:
1. we don't want to disable the fields on each update if they already disabled.
2. we don't want to execute `prefillVmTemplateUpdater` on every update - the user would not be able to make changes at the flavor field for example

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
irosenzw: It is. we use it later on in this method and need it for `getFlavor` and `getWorkloadProfiles`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
irosenzw: why not? we use it on later on `getWorkloadProfiles`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/selectors/combined.ts
```diff
@@ -2,17 +2,19 @@ import { Map as ImmutableMap } from 'immutable';
...
-        immutableListToShallowJS<TemplateKind>(
-          concatImmutableLists(iUserTemplates, iCommonTemplates),
-        ),
-      )
-    : operatingSystemsNative;
+export const getOS = ({ osID, iUserTemplate, openshiftFlag, iCommonTemplates }: GetOSParams) => {
+  let operatingSystems = [];
+
+  if (openshiftFlag && iUserTemplate) {
+    operatingSystems.push(getTemplateOperatingSystems([iUserTemplate.toJS()]));
```
suomiy, line 12: please use toJSON here or toShallowJS

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -34,48 +34,46 @@ export const getFlavorLabel = (flavor: string) => {
...
+      ),
     ]);
   }
   return getTemplateOperatingSystems(getTemplatesOfLabelType(templates, TEMPLATE_TYPE_BASE));
 };
 
 export const getWorkloadProfiles = (
   templates: TemplateKind[],
-  { flavor, os, userTemplate }: { flavor: string; os: string; userTemplate: string },
+  { flavor, os, isUserTemplateValid }: { flavor: string; os: string; isUserTemplateValid: boolean },
```
suomiy, line 50: why do we need isUserTemplateValid variable? can we just pass templates and that is all?

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -34,48 +34,46 @@ export const getFlavorLabel = (flavor: string) => {
   return undefined;
 };
 
-export const getOperatingSystems = (templates: TemplateKind[], userTemplate: string) => {
-  if (userTemplate) {
+export const getOperatingSystems = (templates: TemplateKind[], userTemplateName: string) => {
```
suomiy, line 37: same, why do we need to pass userTemplateName?

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
```diff
@@ -34,48 +34,46 @@ export const getFlavorLabel = (flavor: string) => {
...
 };
 
 export const getFlavors = (
   templates: TemplateKind[],
-  { workload, os, userTemplate }: { workload: string; os: string; userTemplate: string },
+  {
+    workload,
+    os,
+    isUserTemplateValid,
+  }: { workload: string; os: string; isUserTemplateValid: boolean },
```
suomiy, line 68: same

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -17,21 +18,23 @@ import { iGetFieldValue } from '../../selectors/immutable/field';
...
     cnvBaseImages,
     onChange,
   }) => {
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const isUserTemplateValid = iGetIsLoaded(iUserTemplate) && !iGetLoadError(iUserTemplate);
+
+    const templates = isUserTemplateValid
+      ? [iGetLoadedData(iUserTemplate).toJS()]
```
suomiy, line 32: please prefer toShallowJS here

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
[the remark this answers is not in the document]
suomiy: delete the template why you are in a wizard. IMO it is better to do it here, as these inconsistencies add up and are hard to follow up.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
suomiy: I would not complicate it. Let's prefill it just once (that is also how vmimport works).

We are also not checking template updates at the moment.
Let's just assume the user can modify all the values once he clicks on the template.

The deletion of the template has to be notified though since we cannot continue with vm creation

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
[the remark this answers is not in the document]
suomiy: another valid option IMO is to disable just the `review and create` button and let the wizard be navigable with the error on the top. But the navigation would be at least consistent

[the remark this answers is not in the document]
irosenzw: when the template gets deleted all the nav buttons are disabled (`next`, `back`, `review and create`)
and when the template gets create while in the wizard they are enabled again. 
I don't see what is wrong with the logic here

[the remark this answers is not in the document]
suomiy: the nav is enabled and it is possible to jump to other tabs

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/selectors/combined.ts
```diff
@@ -2,17 +2,19 @@ import { Map as ImmutableMap } from 'immutable';
...
-        immutableListToShallowJS<TemplateKind>(
-          concatImmutableLists(iUserTemplates, iCommonTemplates),
-        ),
-      )
-    : operatingSystemsNative;
+export const getOS = ({ osID, iUserTemplate, openshiftFlag, iCommonTemplates }: GetOSParams) => {
+  let operatingSystems = [];
+
+  if (openshiftFlag && iUserTemplate) {
+    operatingSystems.push(getTemplateOperatingSystems([toShallowJS(iUserTemplate)]));
```
suomiy, line 12: array  in an array. better to have an assignment?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -43,7 +45,6 @@ export const WorkloadProfile: React.FC<WorkloadProps> = React.memo(
           field={workloadProfileField}
           fieldType={FormFieldType.SELECT}
           loadingResources={{
```
suomiy, line 47: it should be in loading resources if the wizard is in userTemplate mode

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,34 +71,36 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
     }
 
-    const flavors = flavorSort(getFlavors(vanillaTemplates, params));
+    const flavors = flavorSort(getFlavors(templates, params));
 
-    const workloadProfiles = getWorkloadProfiles(vanillaTemplates, params);
+    const workloadProfiles = getWorkloadProfiles(templates, params);
 
     const loadingResources = openshiftFlag
       ? {
```
suomiy, line 103: loadingResources: same here

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
irosenzw: I wasn't able to move the content of this function to `selectUserTemplateOnLoadedUpdater` the fields are not getting populated. if you have a suggestion please share it, otherwise let's do it as a followup. wdyt ?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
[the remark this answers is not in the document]
matthewcarleton: I'm ok with the second option

matthewcarleton:
> > What if we said
> 
> > Could not load requested template
> > requested template vm-templates-example-5 can not be found. It may have been removed.
> > Please contact your admin for more information or choose another template (this could link back to the template list)
> 
> @matthewcarleton
> 
> We could do this, but I am against it at this time. This component handles all types of errors (of all resources and casuses) and if we start adding customization like this one on top of each other iit will become a mess.
> 
> We have to come either with a better generic solution to these errors or add a good extension framework.
> If we try to do it here, this PR will get more out of scope.

+1 to improving this in general - I think errors are most helpful when they are specific. These generic errors don't help the user take action to fix the problem.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
suomiy: I don't see any issue with it. It seems pretty simple:

```ts
const selectUserTemplateOnLoadedUpdater = (options: UpdateOptions) => {
  const { id, dispatch, getState } = options;
  const state = getState();

  if (
    iGetCommonData(state, id, VMWizardProps.isUserTemplateInitialized) ||
    !iGetLoadedCommonData(state, id, VMWizardProps.userTemplate)
  ) {
    return;
  }

  dispatch(
    vmWizardInternalActions[InternalActionType.UpdateCommonDataValue](
      id,
      [VMWizardProps.isUserTemplateInitialized],
      true,
    ),
  );

  const iUserTemplate = iGetLoadedCommonData(state, id, VMWizardProps.userTemplate);
  const isDisabled = asDisabled(iUserTemplate != null, VMWizardProps.userTemplate);

  dispatch(
    vmWizardInternalActions[InternalActionType.UpdateVmSettings](id, {
      [VMSettingsField.PROVISION_SOURCE_TYPE]: { isDisabled },
      [VMSettingsField.CONTAINER_IMAGE]: { isDisabled },
      [VMSettingsField.IMAGE_URL]: { isDisabled },
      [VMSettingsField.OPERATING_SYSTEM]: { isDisabled },
      [VMSettingsField.CLONE_COMMON_BASE_DISK_IMAGE]: {
        isHidden: asHidden(iUserTemplate != null, VMWizardProps.userTemplate),
      },
    }),
  );

  prefillVmTemplateUpdater(options);
};
```

frontend/packages/kubevirt-plugin/src/selectors/vm-template/combined-dependent.ts
suomiy: @irosenzw please take a look. This should work the same way as getWorkloadProfiles.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
irosenzw: This is the first thing I tried. it doesn't work.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -17,36 +19,42 @@ import { iGetFieldValue } from '../../selectors/immutable/field';
...
     flavor,
     cnvBaseImages,
     onChange,
   }) => {
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const isUserTemplateValid = iGetIsLoaded(iUserTemplate) && !iGetLoadError(iUserTemplate);
+
+    const templates = isUserTemplateValid
```
suomiy, line 32: There are 3 options here 

```
if iUserTemplate {
  if isUserTemplateValid {
    use iUserTemplate
  } else {
     use []
  }
} else {
  use commonTemplates
}
```

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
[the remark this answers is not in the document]
irosenzw: OK. Second option it is: `Create Virtual Machine from vm-template-example`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,45 +72,48 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
-      userTemplate,
       flavor,
       workload: workloadProfile,
       os,
     };
 
-    const vanillaTemplates = immutableListToShallowJS(
-      concatImmutableLists(iGetLoadedData(commonTemplates), iGetLoadedData(userTemplates)),
-    );
+    const templates = isUserTemplateValid
```
suomiy, line 83: same here, there are 3 options

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,45 +72,48 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
     const loadingResources = openshiftFlag
       ? {
-          userTemplates,
           commonTemplates,
           cnvBaseImages,
         }
       : {};
 
+    if (openshiftFlag && iUserTemplate) {
+      Object.assign(loadingResources, { [iUserTemplate]: iUserTemplate });
```
suomiy, line 109:
```suggestion
      Object.assign(loadingResources, { iUserTemplate: iUserTemplate });
```
the key is just called iUserTemplate (this notation would resolve in a key called `[object Object]`)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -17,36 +19,42 @@ import { iGetFieldValue } from '../../selectors/immutable/field';
...
       }),
     );
 
+    const loadingResources = {
+      commonTemplates,
+      cnvBaseImages,
+    };
+
+    if (iUserTemplate) {
+      Object.assign(loadingResources, { [iUserTemplate]: iUserTemplate });
```
suomiy, line 49:
```suggestion
      Object.assign(loadingResources, { iUserTemplate: iUserTemplate });
```

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -144,7 +147,7 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
           checkboxDescription: '',
         };
 
-        if (!userTemplate) {
+        if (!isUserTemplateValid) {
```
suomiy, line 150:
```suggestion
        if (!iUserTemplate) {
```

like this?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os-flavor.tsx
```diff
@@ -72,45 +72,48 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
+      Object.assign(loadingResources, { [iUserTemplate]: iUserTemplate });
+    }
+
     let operatingSystemValidation;
     let flavorValidation;
 
     if (
-      iGetIsLoaded(commonTemplates) &&
-      iGetIsLoaded(userTemplates) &&
+      (iGetIsLoaded(commonTemplates) || isUserTemplateValid) &&
```
suomiy, line 116: just loaded not valid and `&&`

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -131,14 +133,14 @@ class CreateVMWizardComponent extends React.Component<CreateVMWizardComponentPro
...
-    const { isCreateTemplate, isProviderImport } = this.props;
+    const { isCreateTemplate, isProviderImport, iUserTemplate } = this.props;
     if (isCreateTemplate) {
       return CREATE_VM_TEMPLATE;
     }
     if (isProviderImport) {
       return IMPORT_VM;
     }
-    return CREATE_VM;
+    return iUserTemplate ? `${CREATE_VM} from ${iGetName(iUserTemplate)} template` : CREATE_VM;
```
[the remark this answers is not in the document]
irosenzw: already changed it

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -88,7 +56,7 @@ const selectedUserTemplateUpdater = (options: UpdateOptions) => {
       [VMSettingsField.IMAGE_URL]: { isDisabled },
       [VMSettingsField.OPERATING_SYSTEM]: { isDisabled },
```
suomiy, line 57: I suppose we should disable also workload profile since it cannot be changed for the user template

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -118,7 +121,7 @@ const CreateVMWizardFooterComponent: React.FC<CreateVMWizardFooterComponentProps
         const isLastStep = activeStepID === VMWizardTab.RESULT;
 
         const isNextButtonDisabled = isAnyStepLocked;
-        const isReviewButtonDisabled = isAnyStepLocked;
+        const isReviewButtonDisabled = isAnyStepLocked || isInvalidUserTemplate;
```
suomiy, line 124: this should also apply for the create button. Can you create isCreateButtonDisabed variable?

and apply the same logic like we do for the render?

quote:
```tsx
             {activeStepID === VMWizardTab.REVIEW
                  ? getCreateVMLikeEntityLabel(isCreateTemplate, isProviderImport)
                  : 'Next'}
```

suomiy (changes requested):

suomiy (changes requested):

PR #6404, opened 2020-08-23
Bug 1871646: Fix location of Minimal Deployment message and calculation for Internal Mode deployment

bipuladh:
- Fixes calculation for Internal mode where if the total CPU < 30 then only the minimal deployment will occur.
- Fixes calculation for an Attached mode where if the CPU per node < 10 then the minimal deployment will occur.
- Moves the message as an alert to the bottom. 
Screenshot from 2020-08-26 16-33-52

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
@@ -34,8 +34,9 @@ import {
   SelectNodesSection,
   StorageClassSection,
   EncryptSection,
+  MinimalDeploymentAlert,
 } from '../../utils/common-ocs-install-el';
-import { filterSCWithoutNoProv, shouldDeployMinimally } from '../../utils/install';
+import { filterSCWithoutNoProv, shouldDeployInternalAsMinimal } from '../../utils/install';
 import { OCS_INTERNAL_CR_NAME } from '../../constants';
 import './ocs-install.scss';
 
@@ -94,12 +95,28 @@ export const CreateInternalCluster = withHandlePromise<
   const dispatch = useDispatch();
   const [nodes, setNodes] = React.useState<NodeKind[]>([]);
 
-  const isMinimal = shouldDeployMinimally(nodes);
+  const isMinimal = shouldDeployInternalAsMinimal(nodes);
+  const [showMessage, setShowMessage] = React.useState(false);
+
+  const timeoutID = React.useRef(null);
+
+  React.useEffect(() => {
+    if (timeoutID.current !== null) {
+      clearTimeout(timeoutID.current);
+    }
+    if (nodes.length >= minSelectedNode) {
+      timeoutID.current = setTimeout(() => {
+        setShowMessage(true);
+      }, 1000);
+    } else {
+      setShowMessage(false);
+    }
+  }, [nodes, setShowMessage]);
 
   const submit = (event: React.MouseEvent<HTMLButtonElement>) => {
     event.preventDefault();
     // eslint-disable-next-line promise/catch-or-return
-    handlePromise(makeOCSRequest(nodes, storageClass, osdSize, isEncrypted), () => {
+    handlePromise(makeOCSRequest(nodes, storageClass, osdSize, isEncrypted, isMinimal), () => {
       dispatch(setFlag(OCS_CONVERGED_FLAG, true));
       dispatch(setFlag(OCS_INDEPENDENT_FLAG, false));
       dispatch(setFlag(OCS_FLAG, true));
@@ -154,14 +171,13 @@ export const CreateInternalCluster = withHandlePromise<
               Select at least 3 nodes in different zones you wish to use. The recommended
               requirements are 16 CPUs and 64 GiB of RAM per node.
             </div>
-            {isMinimal && (
-              <div className="ceph-ocs-install__minimal-msg">
-                Since the selected nodes do not satisfy the recommended requirements stated above, a
-                minimal cluster will be deployed, limited to a single storage device set.
-              </div>
-            )}
+            <div className="text-muted ceph-ocs-install__minimal-msg">
+              A minimal deployment is also available if needed with the requirements of 8 CPUs and
+              32 GiB RAM.
+            </div>
           </>
         </SelectNodesSection>
+        <>{isMinimal && showMessage && <MinimalDeploymentAlert isInternalMode />}</>
         <ButtonBar errorMessage={errorMessage} inProgress={inProgress}>
           <ActionGroup className="pf-c-form">
             <Button
--- a/frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
@@ -4,6 +4,7 @@ import { ListPage } from '@console/internal/components/factory';
 import { NodeModel } from '@console/internal/models';
 import { FieldLevelHelp } from '@console/internal/components/utils';
 import { StorageClassResourceKind } from '@console/internal/module/k8s';
+import { TechPreviewBadge } from '@console/shared';
 import { OCSStorageClassDropdown } from '../components/modals/storage-class-dropdown';
 import { storageClassTooltip } from '../constants/ocs-install';
 import '../components/ocs-install/ocs-install.scss';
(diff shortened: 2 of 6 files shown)


cloudbehl:
How does total make sense if the node itself doesn't satisfy the criteria? Let say you have 15 nodes with 2vcpu's?

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -58,12 +58,30 @@ export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
   return Array.from(nodes);
 };
 
+export const shouldDeployInternalAsMinimal = (nodes: NodeKind[]) => {
```
cloudbehl, line 61:
```suggestion
export const DeployCloudAsMinimal = (nodes: NodeKind[]) => {
```

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -58,12 +58,30 @@ export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
...
+    },
+    {
+      totalCPU: 0,
+      totalMemory: 0,
+    },
+  );
+  return totalCPU < 30 || totalMemory < 60;
+};
+
 export const shouldDeployMinimally = (nodes: NodeKind[]) =>
```
cloudbehl, line 79:
```suggestion
export const DeployBaremetalAsMinimal = (nodes: NodeKind[]) =>
```

frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
```diff
@@ -17,6 +18,30 @@ export const OCSAlert = () => (
...
+    title={
+      <div className="ceph-minimal-deployment-alert__header">
+        Minimal deployment node requirements
+        <TechPreviewBadge />
+      </div>
+    }
+    isInline
+  >
+    {isInternalMode
+      ? 'The selected nodes do not match the OCS storage cluster minimum recommended requirements of an aggregated 30 CPUs. If the selection won’t be modified, a minimal cluster will be deployed and may face some performance issues.'
```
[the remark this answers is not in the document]
cloudbehl: As we are checking the memory requirements in `shouldDeployInternalAsMinimal` user will not able to make out why error is coming if the memory requirements are failing.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: you can deploy internal mode in VMWare as well.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: You can deploy Attached devices in AWS as well.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: It's not a react component. (first character no need to be uppercase)

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: I will keep it as `shouldDeployAttachedAsMinimal` to keep it consistent with the other function

frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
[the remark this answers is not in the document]
bipuladh: @shirimordechay please have a look at this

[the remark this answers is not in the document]
shirimordechay: If there is a memory requirement we should mention it, I think the message should fit the user selection. After the "recommended requirement of..." we should add the requirement that doesn't fit. the rest of the warning text will stay the same

In this use case, the text will be: ‘The selected nodes do not match the OCS storage cluster minimum recommended requirements of a [total 60 GiB Memory]. If the selection won’t be modified, a minimal cluster will be deployed and may face some performance issues.’ 

For cloud/ LSO requirement text should be changed accordingly.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -58,12 +58,30 @@ export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
...
+      acc.totalCPU += cpus;
+      acc.totalMemory += memory;
+      return acc;
+    },
+    {
+      totalCPU: 0,
+      totalMemory: 0,
+    },
+  );
+  return totalCPU < 30 || totalMemory < 60;
```
cloudbehl, line 76: We have different values over here for total memory & the message is saying 72 GiB. We need to be consistent.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -58,12 +58,30 @@ export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
...
+};
+
+export const shouldDeployAttachedAsMinimal = (nodes: NodeKind[]) =>
   nodes.reduce((acc, curr) => {
     const cpus = humanizeCpuCores(getNodeCPUCapacity(curr)).value;
     const memoryRaw = getNodeAllocatableMemory(curr);
     const memory = humanizeBinaryBytes(convertToBaseValue(memoryRaw)).value;
-    if (memory < 60 || cpus < 16) {
+    if (memory < 60 || cpus < 10) {
```
cloudbehl, line 84: same as above

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: @cloudbehl We are putting as 24 Gigs per node. I am assuming the worst-case conversion will come to 20 so taking at 60. RAM is not an as hard constraint as CPUs.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
bipuladh: Same as above

frontend/packages/ceph-storage-plugin/src/utils/install.ts
cloudbehl: Understood but the issue is we can have either message say the same or the validation should match because when admin is going to match these numbers will not match.

The message will say 72 Gib required when total of nodes goes below 60. Which I think will be weird & QE will raise bugs.

bipuladh: If you have noticed then the memory limit is not exactly like we want it to be sometimes it goes down by 3 to 4 GBs. For ex: 64GiB usually shows up as 61 and as you mentioned in my other PR to even 60 GiB at times. Now I am doing this calculation based ob (24GiB - 4 Error limit) * 3 Nodes. Hope this clears things out for you.

cloudbehl: The thing that matters is allocatable memory, not the memory provisioned at the time cluster was created. 

Also, memory is something mentioned in the node list table already & admin is going to calculate based on those values.

 I'm just saying user & QE will be confused with these values & I don't see any harm in having the same values and we should do that.

gnehapk:
@bipuladh As discussed,  please change the `getOCSRequestData`function, as its replacing the `spec` property in case of minimal deployement and hence, breaking the ocs cluster creation on attached devices.

bipuladh:
> @bipuladh As discussed, please change the `getOCSRequestData`function, as its replacing the `spec` property in case of minimal deployement and hence, breaking the ocs cluster creation on attached devices.

Noted and added.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
cloudbehl: Skipping for now as this needs more discussion.

cloudbehl (approved):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
```diff
@@ -94,12 +95,28 @@ export const CreateInternalCluster = withHandlePromise<
...
+  const isMinimal = shouldDeployInternalAsMinimal(nodes);
+  const [showMessage, setShowMessage] = React.useState(false);
+
+  const timeoutID = React.useRef(null);
+
+  React.useEffect(() => {
+    if (timeoutID.current !== null) {
+      clearTimeout(timeoutID.current);
+    }
+    if (nodes.length >= 3) {
```
gnehapk, line 107:
```suggestion
    if (nodes.length >= minSelectedNode) {
```

PR #6406, opened 2020-08-24
Bug 1870515: PVC Network Upload error should be more detailed

glekner:
Errors coming from the CORS preflight are not catchable by the UI ([see thread](https://github.com/axios/axios/issues/838#issuecomment-294956982))

All other errors will be more detailed. For this specific case, I added the text below:
Screen Shot 2020-08-24 at 13 22 01

Please approve the message @jelkosz @matthewcarleton

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/cdi-upload-provider.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/cdi-upload-provider.tsx
@@ -89,6 +89,23 @@ export const useCDIUploadHook = (): CDIUploadContextProps => {
               ...newUpload,
               uploadStatus: UPLOAD_STATUS.CANCELED,
             });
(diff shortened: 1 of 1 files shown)


jelkosz:
@glekner I would like it to be a bit more actionable. e.g. containing something like: please follow this <link> and approve the certificate and try again.

glekner:
image

@jelkosz @matthewcarleton

matthewcarleton:
> image
> 
> @jelkosz @matthewcarleton

The link text could just be the "approve this certificate". 
I think this should open in a new tab, otherwise they lose their data from the form right? 
In the text you wrote a bad internet connection and now we are just saying it's the certificate. Do we know this for sure?

glekner:
@matthewcarleton I changed the text, and the link now opens a new tab.

When the error is undefined it means the error came from a request handled by the browser and not our code, so its 99% a certificate error.

PR #6407, opened 2020-08-24
Bug 1871742: can't delete a PVC created using a data upload

glekner:
For Normal PVCs the modal is the same:
Screen Shot 2020-08-24 at 16 36 25

For CDI Bound PVCs the modal is:
Screen Shot 2020-08-24 at 16 36 32

@matthewcarleton how do we want this to look?

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/pvc.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/pvc.ts
@@ -1,4 +1,5 @@
-import { K8sResourceCommon } from '@console/internal/module/k8s';
+import { PersistentVolumeClaimKind } from '@console/internal/module/k8s';
+import { AlertVariant } from '@patternfly/react-core';
 import { Extension, LazyLoader } from './base';
 
 namespace ExtensionProperties {
@@ -11,7 +12,7 @@ namespace ExtensionProperties {
     /** Loader for the corresponding Status component. */
     loader: LazyLoader<T>;
     /** predicate that tells if to render the status or not */
-    predicate: (pvc: K8sResourceCommon) => boolean;
+    predicate: (pvc: PersistentVolumeClaimKind) => boolean;
     /** priority for the corresponding Status component. bigger is prioritized */
     priority: number;
   }
@@ -21,6 +22,22 @@ namespace ExtensionProperties {
     label: string;
     /** path for the create prop */
     path: string;
+  }
+
+  export interface PVCDelete {
+    /** predicate that tells if to use the extension or not */
+    predicate: (pvc: PersistentVolumeClaimKind) => boolean;
+    /** method for the pvc delete operation */
+    onPVCKill: (pvc: PersistentVolumeClaimKind) => Promise<void>;
+    /** alert for additional info */
+    alert?: {
+      /** alert title */
+      title: string;
+      /** alert variant */
+      type: AlertVariant;
+      /** alert body */
+      body: LazyLoader<{ pvc: PersistentVolumeClaimKind }>;
+    };
   }
 }
 
@@ -36,6 +53,10 @@ export interface PVCCreateProp extends Extension<ExtensionProperties.PVCCreatePr
   type: 'PVCCreateProp';
 }
 
+export interface PVCDelete extends Extension<ExtensionProperties.PVCDelete> {
+  type: 'PVCDelete';
+}
+
 export const isPVCAlert = (e: Extension): e is PVCAlert => {
   return e.type === 'PVCAlert';
 };
@@ -46,4 +67,8 @@ export const isPVCStatus = (e: Extension): e is PVCStatus => {
 
 export const isPVCCreateProp = (e: Extension): e is PVCCreateProp => {
   return e.type === 'PVCCreateProp';
+};
+
+export const isPVCDelete = (e: Extension): e is PVCDelete => {
+  return e.type === 'PVCDelete';
 };
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/pvc-delete-extension.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/pvc-delete-extension.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { k8sKill, PersistentVolumeClaimKind } from '@console/internal/module/k8s';
+import { DataVolumeModel } from '../../models';
+import {
+  useK8sWatchResource,
+  WatchK8sResource,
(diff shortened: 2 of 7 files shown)


glekner:
@suomiy @rawagner

jelkosz:
I like it!

frontend/public/components/modals/delete-pvc-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  );
+
+  return (
+    <form onSubmit={submit} className="modal-content">
+      <ModalTitle>
+        <YellowExclamationTriangleIcon className="co-icon-space-r" /> Delete Persistent Volume Claim
+      </ModalTitle>
+      <ModalBody>
+        <Stack hasGutter>
+          {alertComponents.map((alert, i) => (
```
rawagner, line 46:
```suggestion
          {alertComponents}
```

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/pvc-delete-extension.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { Alert, AlertVariant } from '@patternfly/react-core';
+import { k8sKill, K8sResourceKind } from '@console/internal/module/k8s';
+import { DataVolumeModel } from '../../models';
+
+export const killCDIBoundPVC = (pvc: K8sResourceKind) => k8sKill(DataVolumeModel, pvc);
+
+export const PVCDeleteAlertExtension: React.FC<PVCDeleteAlertExtension> = ({ pvc }) => (
```
rawagner, line 8: you should also consider that the extension would define `alertTitle: string`, `alertBody: LazyLoader<{pvc: PersistentVolumeClaimKind}>`, `alertType: AlertVariant`. 

It makes the extension less error prone, you dont have to specify className, isInline etc..all would be handled by pvc delete modal

rawagner (changes requested):

matthewcarleton:
> For Normal PVCs the modal is the same:
> Screen Shot 2020-08-24 at 16 36 25
> 
> For CDI Bound PVCs the modal is:
> Screen Shot 2020-08-24 at 16 36 32
> 
> @matthewcarleton how do we want this to look?

We need to ensure that if the PVC is used to back any OS then we highlight that here. It's possible that a user could delete this without knowing that's being cloned as an OS source somewhere else right?

frontend/public/components/modals/delete-pvc-modal.tsx
glekner: I want to render a margin between every alert `<StackItem>`, so I need this map.

glekner:
fixed suggestions @rawagner 
@matthewcarleton well yes, deleting is a serious thing but I don't think we should check for every VM that uses the specific PVC. We could maybe check if that PVC is a golden one and let the user know its important?

matthewcarleton:
> fixed suggestions @rawagner
> @matthewcarleton well yes, deleting is a serious thing but I don't think we should check for every VM that uses the specific PVC. We could maybe check if that PVC is a golden one and let the user know its important?

Ya that's what I was thinking. If it's that important we shouldn't assume the user is thinking about it that way. It would be a nice addition to alert them that is being used to support and OS as a backing source.

glekner:
@matthewcarleton can you approve the text here?
Screen Shot 2020-08-30 at 12 27 13

jelkosz:
@glekner I dont think it will cause any malfunction. We clone the image. It will cause that the other VMs will not be able to use it but the ones which have already been deployed will be fine. Or am I missing something?

frontend/public/components/modals/delete-pvc-modal.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const extensionPromises = pvcDeleteExtensions.map(
+      ({ properties: { predicate, onPVCKill } }) => predicate(pvc) && onPVCKill(pvc),
+    );
+    return handlePromise(Promise.all([promise, ...extensionPromises]), close);
+  };
+
+  const alertComponents = pvcDeleteExtensions.map(
+    ({ properties: { predicate, alert }, uid }) =>
+      predicate(pvc) && (
+        <Alert
```
rawagner, line 34: If you wrap the alert in `StackItem` you wont need to `map` again later.

```suggestion
       <StackItem key={uid}>
          <Alert
```
you can also move key from `Alert` to `StackItem`. There's also no need to make `key` more complicated than just uid. Doing `pvc-alert-${uid}` doesnt bring any value

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/pvc-delete-extension.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const killCDIBoundPVC = (pvc: PersistentVolumeClaimKind) =>
+  k8sKill(DataVolumeModel, {
+    metadata: {
+      name: pvc?.metadata?.name,
+      namespace: pvc?.metadata?.namespace,
+    },
+  });
+
+export const PVCDeleteAlertExtension = ({ pvc }) => {
```
rawagner, line 19:
```suggestion
export const PVCDeleteAlertExtension: React.FC<{ pvc: PersistentVolumeClaimKind }> = ({ pvc }) => {
```

frontend/packages/console-plugin-sdk/src/typings/pvc.ts
```diff
@@ -22,6 +23,22 @@ namespace ExtensionProperties {
...
+
+  export interface PVCDelete {
+    /** predicate that tells if to use the extension or not */
+    predicate: (pvc: PersistentVolumeClaimKind) => boolean;
+    /** method for the pvc delete operation */
+    onPVCKill: (pvc: PersistentVolumeClaimKind) => Promise<void>;
+    /** alert for additional info */
+    alert?: {
+      /** alert title */
+      alertTitle: string;
```
rawagner, line 35: could be just `title`, 'type', 'Body'. Not a blocker though :)

frontend/public/components/modals/delete-pvc-modal.tsx
rawagner: See my comment https://github.com/openshift/console/pull/6407#discussion_r480074555

matthewcarleton:
> @matthewcarleton can you approve the text here?
> Screen Shot 2020-08-30 at 12 27 13

Much better!
The only thing I would ask is if we can provide which OS it's backing. It will help the user make the connection easier.

rawagner:
/hold

needs rebase

PR #6409, opened 2020-08-24
Bug 1871866: Monitoring: Fix list style for the table pagination dropdowns

kyoto:
Affects the Monitoring UI, "Metrics" and "Dashboards" pages. These are the only pages using the PatternFly `Pagination` component.

Before | After
-|-
before | after

### DIFF
--- a/frontend/public/style/_overrides.scss
+++ b/frontend/public/style/_overrides.scss
@@ -362,6 +362,11 @@ h6 {
   }
 }
 
+// Necessary due to setting $pf-global--enable-reset to false
+.pf-c-pagination .pf-c-options-menu__menu {
+  list-style: none;
+}
+
 // pf table overrides
 
 .pf-c-table tr > th {


frontend/public/style/_overrides.scss
```diff
@@ -362,6 +362,10 @@ h6 {
   }
 }
 
+.pf-c-pagination .pf-c-options-menu__menu {
```
kyoto, line 365: I guess this is necessary for the same reason as given in the comment above (`due to setting $pf-global--enable-reset to false`)

frontend/public/style/_overrides.scss
```diff
@@ -362,6 +362,10 @@ h6 {
...
+  list-style-type: none;
```
rhamilto, line 366: Nit
```suggestion
  list-style: none;
```

frontend/public/style/_overrides.scss
rhamilto: Yeah.  Mind duplicating the comment?

kyoto:
@rhamilto Addressed your feedback. Thanks!

PR #6414, opened 2020-08-24
Bug 1871927: update resource section text for knative

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-2438

**Analysis / Root cause**: misleading text for Knative in resource section

**Solution Description**: update the text in resource section

Screenshot 2020-08-24 at 9 18 57 PM

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


invincibleJai:
@sahil143  thanks for the fix , as per design D/DC description has `.` at end so we may need to add that for DC and @rachael-phillips shouldn't description for knative service as well end with `.`

image

rachael-phillips:
I know how we should solve your comment Jai - I agree it should be consistent, so the text could read "A Knative Service is a type of deployment that enables Serverless scaling to 0 when idle."
That way we can stay in a consistent format to the others.

invincibleJai:
/lgtm

Thanks @sahil143  for making the changes

PR #6416, opened 2020-08-24
Bug 1848512: Wrong interpretation of labels for resource quota diagram

dtaylor113:
Needed to provide `data.x` for chart labels/tooltip

### Before
image

### After
image

### DIFF
--- a/frontend/public/components/resource-quota.jsx
+++ b/frontend/public/components/resource-quota.jsx
@@ -191,16 +191,24 @@ const NoQuotaGuage = ({ title, className }) => (
 
 export const QuotaGaugeCharts = ({ quota, resourceTypes, chartClassName = null }) => {
   const resourceTypesSet = new Set(resourceTypes);
+  const cpuRequestUsagePercent = getResourceUsage(
+    quota,
+    resourceTypesSet.has('requests.cpu') ? 'requests.cpu' : 'cpu',
+  ).percent;
+  const cpuLimitUsagePercent = getResourceUsage(quota, 'limits.cpu').percent;
+  const memoryRequestUsagePercent = getResourceUsage(
+    quota,
+    resourceTypesSet.has('requests.memory') ? 'requests.memory' : 'memory',
+  ).percent;
+  const memoryLimitUsagePercent = getResourceUsage(quota, 'limits.memory').percent;
   return (
     <div className="co-resource-quota-chart-row">
       {resourceTypesSet.has('requests.cpu') || resourceTypesSet.has('cpu') ? (
         <div className="co-resource-quota-gauge-chart">
           <GaugeChart
             data={{
-              y: getResourceUsage(
-                quota,
-                resourceTypesSet.has('requests.cpu') ? 'requests.cpu' : 'cpu',
-              ).percent,
+              x: `${cpuRequestUsagePercent}%`,
+              y: cpuRequestUsagePercent,
             }}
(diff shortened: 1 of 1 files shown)


frontend/public/components/resource-quota.jsx
```diff
@@ -191,16 +191,32 @@ const NoQuotaGuage = ({ title, className }) => (
 
 export const QuotaGaugeCharts = ({ quota, resourceTypes, chartClassName = null }) => {
   const resourceTypesSet = new Set(resourceTypes);
+  const cpuRequestUsagePercent =
```
jcaianirh, line 194:
```suggestion
  const cpuRequestUsagePercent =getResourceUsage(
    quota,
    resourceTypesSet.has('requests.cpu') ? 'requests.cpu' : 'cpu',
  ).percent;
```

frontend/public/components/resource-quota.jsx
```diff
@@ -191,16 +191,32 @@ const NoQuotaGuage = ({ title, className }) => (
...
+    ? getResourceUsage(quota, 'limits.cpu').percent
+    : 0;
+  const memoryRequestUsagePercent =
+    resourceTypesSet.has('requests.memory') || resourceTypesSet.has('memory')
+      ? getResourceUsage(
+          quota,
+          resourceTypesSet.has('requests.memory') ? 'requests.memory' : 'memory',
+        ).percent
+      : 0;
+  const memoryLimitUsagePercent = resourceTypesSet.has('limits.memory')
```
jcaianirh, line 209:
```suggestion
  const memoryLimitUsagePercent = getResourceUsage(quota, 'limits.memory').percent;
```

frontend/public/components/resource-quota.jsx
dtaylor113: fixed

frontend/public/components/resource-quota.jsx
dtaylor113: fixed

dtaylor113:
Hi @jcaianirh, applied that patter to all 4 lines.   Had to revert the check between showing a `GaugeChart` chart or `NoQuotaGuage` chart, in order to actually display a `0%` value, which is the state when a resource quota is created but no pods applied yet.

dtaylor113:
Hi @spadgett, @kyoto, and @rhamilto,  looking for an approval in order to merge this PR.  -thanks :-)

frontend/public/components/resource-quota.jsx
```diff
@@ -191,16 +191,24 @@ const NoQuotaGuage = ({ title, className }) => (
...
       {resourceTypesSet.has('requests.cpu') || resourceTypesSet.has('cpu') ? (
         <div className="co-resource-quota-gauge-chart">
           <GaugeChart
             data={{
-              y: getResourceUsage(
-                quota,
-                resourceTypesSet.has('requests.cpu') ? 'requests.cpu' : 'cpu',
-              ).percent,
+              x: `${cpuRequestUsagePercent}%`,
+              y: cpuRequestUsagePercent,
```
spadgett, line 211: Is there a reason we append `%` for `x` and not for `y`?

dtaylor113: `y' is the actual numeric data, `x` is the category string/label, which is used in the tooltip.  
https://formidable.com/open-source/victory/docs/victory-pie/

PR #6417, opened 2020-08-24
Bug 1875600: fix empty default value for pipeline parameters

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-4493

**Root Analysis:**
Removing default value for parameter from devconsole UI replaces default value with <empty string>

**Solution Description:**
removing `default` property if null or empty string

**GIF:**
defaultval

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
@@ -1,0 +1,48 @@
+import * as _ from 'lodash';
+import { PipelineParam } from '@console/dev-console/src/utils/pipeline-augment';
+import { pipelineParameters, pipelineParametersWithoutDefaults } from './utils-data';
+import { removeEmptyDefaultFromPipelineParams } from '../utils';
+
+describe('removeEmptyDefaultFromPipelineParams omits empty default values', () => {
+  it('should return pipline parameters by only omitting empty default values', () => {
+    const result = removeEmptyDefaultFromPipelineParams(pipelineParameters);
+    const expectedPipelineParameters: PipelineParam[] = [
+      {
+        name: 'param1',
+        default: 'abc',
+        description: 'This is param 1',
+      },
+      {
+        name: 'param2',
+        description: 'This is param 2',
+      },
+      {
+        name: 'param3',
+        default: 'xyz',
+        description: 'This is param 3',
+      },
+    ];
+
+    expect(result).toEqual(expectedPipelineParameters);
+  });
+
+  it('should return empty array if pipline parameters is empty', () => {
+    let result = removeEmptyDefaultFromPipelineParams(null);
+    expect(result).toEqual([]);
+
+    result = removeEmptyDefaultFromPipelineParams([]);
+    expect(result).toEqual([]);
+  });
+
+  it('should return pipline parameters as is if default is non-empty', () => {
+    const pipelineParams = _.cloneDeep(pipelineParameters);
+    pipelineParams[1].default = 'mno';
+    const result = removeEmptyDefaultFromPipelineParams(pipelineParams);
+    expect(result).toEqual(pipelineParams);
+  });
+
+  it('should return pipline parameters as is if the default property is not present', () => {
+    const result = removeEmptyDefaultFromPipelineParams(pipelineParametersWithoutDefaults);
+    expect(result).toEqual(pipelineParametersWithoutDefaults);
+  });
+});
--- a/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
@@ -12,6 +12,7 @@ import {
 import { getTaskParameters } from '../resource-utils';
 import { TASK_ERROR_STRINGS, TaskErrorType } from './const';
 import { PipelineBuilderFormikValues, PipelineBuilderFormValues, TaskErrorMap } from './types';
+import { removeEmptyDefaultFromPipelineParams } from '../detail-page-tabs/utils';
 
 export const getErrorMessage = (errorTypes: TaskErrorType[], errorMap: TaskErrorMap) => (
   taskName: string,
@@ -113,7 +114,7 @@ export const convertBuilderFormToPipeline = (
     },
     spec: {
       ...existingPipeline?.spec,
-      params,
+      params: removeEmptyDefaultFromPipelineParams(params),
       resources,
       tasks: tasks.map((task) => removeEmptyDefaultParams(removeListRunAfters(task, listIds))),
     },
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineForm.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/PipelineForm.tsx
@@ -3,6 +3,7 @@ import * as _ from 'lodash';
 import { Formik } from 'formik';
 import { k8sUpdate, K8sResourceKind } from '@console/internal/module/k8s';
 import { PipelineModel } from '../../../models';
+import { removeEmptyDefaultFromPipelineParams } from './utils';
 
 export interface PipelineFormProps {
   PipelineFormComponent: React.ComponentType<any>;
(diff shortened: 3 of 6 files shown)


frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -98,7 +99,9 @@ export const convertBuilderFormToPipeline = (
     },
     spec: {
       ...existingPipeline?.spec,
-      params,
+      params: params.map((param) =>
+        _.isEmpty(param.default) ? { name: param.name, description: param.description } : param,
```
divyanshiGupta, line 103: @debsmita1 why do you have two different implementations for removing default when it is empty?

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -93,6 +95,13 @@ const removeEmptyDefaultParams = (task: PipelineTask): PipelineTask => {
   return task;
 };
 
+export const removeEmptyDefaultValueFromParameter = (parameters: PipelineParam[]) =>
```
divyanshiGupta, line 98:
```suggestion
export const removeDefaultFromParam = (parameters: PipelineParam[]) =>
```

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -93,6 +95,13 @@ const removeEmptyDefaultParams = (task: PipelineTask): PipelineTask => {
...
+  parameters.map((parameter) =>
+    _.isEmpty(parameter.default)
+      ? { name: parameter.name, description: parameter.description }
+      : parameter,
+  );
```
divyanshiGupta, line 103:
```suggestion
 _.map(parameters, (parameter) =>_.omit(parameter, _.isEmpty(parameter.default) ? ['default'] : []));
```
Just a suggestion, what you have done is also correct.

divyanshiGupta (changes requested):

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
debsmita1: we want to remove `default` only if it's empty hence the name `removeEmptyDefaultValueFromParameter`

divyanshiGupta: Yeah, but the name seems too big and can be shortened. If you can then it will be good otherwise we can let this be.

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
debsmita1: @divyanshiGupta  The issue with using `_.omit` is , when this util is used in `convertBuilderFormToPipeline` , `params` throws a type error saying  : Type 'Partial<PipelineParam>[]' is not assignable to type 'PipelineParam[]'.
  Type 'Partial<PipelineParam>' is not assignable to type 'PipelineParam'.
    Property 'name' is optional in type 'Partial<PipelineParam>' but required in type 'PipelineParam'. 
This is the reason why I had two different implementations earlier which I then updated to use the one without the `omit` in both places. Let me know your thoughts

divyanshiGupta: To avoid this you can simply do this:
```
_.map(parameters, (parameter) =>_.omit(parameter, _.isEmpty(parameter.default) ? ['default'] : []) as PipelineParam);
```

But @debsmita1 I just checked the utils.ts file and there is already a function `removeEmptyDefaultParams` (that removes empty default value from task params) above the function that you have created. You can make this function generic and use it wherever we need to remove empty default params.

divyanshiGupta (changes requested):

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
debsmita1: the util `removeEmptyDefaultParams` removes a name-value pair from task params if the value is empty
but the util that I created removes the `default` property form pipeline parameters if it is empty & keeps the name  & description

divyanshiGupta: Okay, then just change the name of your fxn to ` removeEmptyDefaultFromPipelineParams` to avoid any confusions.

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
debsmita1: changed it to `removeEmptyDefaultFromPipelineParams`

frontend/packages/dev-console/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -93,6 +95,13 @@ const removeEmptyDefaultParams = (task: PipelineTask): PipelineTask => {
   return task;
 };
 
+export const removeEmptyDefaultFromPipelineParams = (parameters: PipelineParam[]) =>
```
andrewballantyne, line 98: LGTM on implementation, but I think I am going to suggest 3 changes to help us out here.

1. Move this function `removeEmptyDefaultFromPipelineParams` out of the Pipeline Builder utils... maybe create a utility file in `frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/`
2. Explicitly define the return type of `PipelineParam[]` so the caller knows they are giving you params, and getting back params in the same structure - maybe modified
3. Please write tests - this is a very testable function

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { PipelineParam } from '@console/dev-console/src/utils/pipeline-augment';
+import { pipelineParameters } from './utils-data';
+import { removeEmptyDefaultFromPipelineParams } from '../utils';
+
+describe('removeEmptyDefaultFromPipelineParams omits empty default values', () => {
+  it('should return pipline parameters by omitting empty default values', () => {
```
andrewballantyne, line 6: When writing tests, you should always consider "what can the user of this function provide".

- `null` **value?** - sometimes we can consider this an invalid use of the function and not support the test, just depends
- **empty value** (in this case `[]`) - tests the function doesn't do anything unexpected
- **a value that does NOT need to be modified** - sometimes values are accidentally modified and thus this test helps know that things shouldn't always be touched
- **a value that DOES need to be modified** - this is what you have here, this is a good test to showcase it does do what it does

Essentially you should always look at writing tests that cement the current functionality. Sure, we know your function only modifies the field on the 2nd element, and sure we know that the first and third elements are unmodified thus showing it doesn't touch them... but the goal of tests is to guard against accidental "breaks" or bugs, which include paths that don't do anything.

Please add a few more tests, consider any other path you think you might want to test (those without `default`s?)... rarely is 1 test good enough to say a function is fully tested.

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
andrewballantyne: Every `removeEmptyDefaultFromPipelineParams` call deserves it's own `it` wrapper... Debugging tests can be quite annoying if they are not broken up.

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  it('should return empty array if pipline parameters is empty', () => {
+    let result = removeEmptyDefaultFromPipelineParams(null);
+    expect(result).toEqual([]);
+
+    result = removeEmptyDefaultFromPipelineParams([]);
+    expect(result).toEqual([]);
+  });
+
+  it('should return pipline parameters as is if default is non-empty', () => {
+    pipelineParameters[1].default = 'mno';
```
andrewballantyne, line 37: Probably want to avoid direct modifications. Object references come through imports and could impact other tests that want to use the same data - which usually rears its head at "random" depending on which order the tests are executed in. It's possible Jest protects us from this but I'd be surprised if they do.

Just use lodash to deep clone before modifying directly. Safety first 😄

andrewballantyne (approved):

rohitkrai03 (approved):
/lgtm

PR #6422, opened 2020-08-25
Support Service Binding breaking api changes in console

akashshinde:
This PR isn't meant to be merged, this is just to illustrate what changes UI needs to make in order to make use of updated Service Binding API. 

Ref: Backed PR for this https://github.com/redhat-developer/service-binding-operator/pull/608

**Note:**
This changes would cause problem when user wants to use latest Service Binding Operator with Openshift 4.5 or previous versions, because DevConsole would still create older version of Service Binding Request which won't be supported if latest version of SBO is installed. 
To fix this issues user can install SBO using different channel(This is being discussed how to use different channel to serve different version of SBO). This will be clearly documented in SBO repo. wdyt @sbose78 ? cc: @pedjak


sbose78:
Akash wouldn't we have a new channel for this?

pedjak:
> Akash wouldn't we have a new channel for this?

Yes, that is the idea so that we can avoid breaking existing 4.5 clusters.

pedjak:
solved in https://github.com/openshift/console/pull/6792

PR #6429, opened 2020-08-25
Bug 1872092:  Make 'Volume Snapshot *' labels consistent

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1872092

Note I did not add `Volume` to `Snapshot Class` and `Snapshot Content` on the Volume Content and Volume Snapshot Content table headers in order to keep them shorter.

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
@@ -92,7 +92,7 @@ const SnapshotClassDropdown: React.FC<SnapshotClassDropdownProps> = (props) => {
       dataFilter={filter}
       resources={resources}
       selectedKeyKind={kind}
-      placeholder="Select snapshot class"
+      placeholder="Select volume snapshot class"
       selectedKey={selectedKey}
     />
   );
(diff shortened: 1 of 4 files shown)


yapei:
@rhamilto  Hi, trying to verify the fix before PR is merged

yapei:
1. on Create VolumeSnapshot form page, 'Volume Snapshot Class' dropdown shows 'Select volume snapshot class'
2. on Volume Snapshots Details page, it shows  a field 'Volume Snapshot Class' and links with a 'vsc' icon
3. on Volume Snapshot Class Details page, the heading is shown as 'Volume Snapshot Class Details' , it shows a field 'Volume Snapshot Class' and links with a 'vsc' icon
4. on Volume Snapshot Content Details page, the heading is shown as 'Volume Snapshot Content Details' it shows a field 'Volume Snapshot Class' and links with a 'vsc' icon

But on `Volume Snapshots` lists page, the table heading is still showing `Snapshot Class` while it should be `Volume Snapshot Class`

yapei:
Screen Shot 2020-08-26 at 7 53 46 PM

rhamilto:
> But on `Volume Snapshots` lists page, the table heading is still showing `Snapshot Class` while it should be `Volume Snapshot Class`

@yapei, [as noted above](https://github.com/openshift/console/pull/6429#issue-473171341), I intentionally left the table header instances unchanged due to the length of the labels.  Adding `Volume` makes those column headers significantly longer, making the table columns narrower.  If you feel strongly, I'm happy to change it, but we may need to drop a right-most column or two at various viewport sizes.

yapei:
yeah you are right, I just forgot about that sentence.  Sorry

PR #6430, opened 2020-08-25
Bug 1872303: Fix delay in topology view on changing namespace

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4633

**Description**
Update code to reset the model on a namespace change so that the status box is shown during the loading process for the new namespace. Also noticed that the namespace bar jumps in size when switching due to the view switcher being hidden and re-displayed. The height of the switcher was larger than the rest of the bar causing the resize.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


invincibleJai:
/approve

Verified the changes , shows loading in case of ns change

PR #6433, opened 2020-08-25
Bug 1871108: usability bug: the ui selections are unclear

glekner:
As of now, I only updated the Provision Source dropdown so we can examine the differences.
Until now, we used a component which renders the browser's native select component

Example
Screen Shot 2020-08-25 at 17 18 10

For us to have a more robust UI with descriptions and other props, we need to convert to the new PF `Select` component, its a change that could break tests, and it could take time to change all the selects in the wizard.

1 Key difference between the components is the new Select rendering the dropdown to the bottom or top is not automatic like the native select. causing it to get hidden by the wizard footer: (changing z-index didn't fix the issue)
Screen Shot 2020-08-25 at 17 15 47

The solution is to change the direction manually for some of the selects.
Screen Shot 2020-08-25 at 17 12 29

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
@@ -1,0 +1,115 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+import { SelectOption } from '@patternfly/react-core';
+import { ValidationErrorType, asValidationObject } from '@console/shared/src/utils/validation';
+import {
+  iGetIsLoaded,
+  iGetLoadedData,
+  iGetLoadError,
+  immutableListToShallowJS,
+  toShallowJS,
+} from '../../../../utils/immutable';
+import { FormFieldRow } from '../../form/form-field-row';
+import { FormField, FormFieldType } from '../../form/form-field';
+import {
+  getFlavors,
+  getWorkloadProfiles,
+} from '../../../../selectors/vm-template/combined-dependent';
+import { flavorSort } from '../../../../utils/sort';
+import { VMSettingsField } from '../../types';
+import { iGetFieldValue } from '../../selectors/immutable/field';
+import { nullOnEmptyChange } from '../../utils/utils';
+import { FormPFSelect } from '../../../form/form-pf-select';
+
+export const Flavor: React.FC<FlavorProps> = React.memo(
+  ({
+    iUserTemplate,
+    commonTemplates,
+    os,
+    flavorField,
+    workloadProfile,
+    cnvBaseImages,
+    onChange,
+    openshiftFlag,
+  }) => {
+    const flavor = iGetFieldValue(flavorField);
+    const isUserTemplateValid = iGetIsLoaded(iUserTemplate) && !iGetLoadError(iUserTemplate);
+
+    const params = {
+      flavor,
+      workload: workloadProfile,
+      os,
+    };
+
+    const templates = iUserTemplate
+      ? isUserTemplateValid
+        ? [toShallowJS(iGetLoadedData(iUserTemplate))]
+        : []
+      : immutableListToShallowJS(iGetLoadedData(commonTemplates));
+
+    const flavors = flavorSort(getFlavors(templates, params));
+
+    const workloadProfiles = getWorkloadProfiles(templates, params);
+
+    const loadingResources = openshiftFlag
+      ? {
+          commonTemplates,
+          cnvBaseImages,
+        }
+      : {};
+
+    if (openshiftFlag && iUserTemplate) {
+      Object.assign(loadingResources, { iUserTemplate });
+    }
+
+    let flavorValidation;
+
+    if (
+      iGetIsLoaded(commonTemplates) &&
+      (!iUserTemplate || isUserTemplateValid) &&
+      (flavors.length === 0 || workloadProfiles.length === 0)
+    ) {
+      const validation = asValidationObject(
+        'There is no valid template for this combination. Please install required template or select different os/flavor/workload profile combination.',
+        ValidationErrorType.Info,
+      );
+      if (!flavorField.get('validation')) {
+        flavorValidation = validation;
+      }
+    }
+
+    return (
+      <FormFieldRow
+        field={flavorField}
+        fieldType={FormFieldType.PF_SELECT}
+        validation={flavorValidation}
+        loadingResources={loadingResources}
+      >
+        <FormField value={flavor} isDisabled={flavor && flavors.length === 1}>
+          <FormPFSelect
+            onSelect={(e, v) => nullOnEmptyChange(onChange, VMSettingsField.FLAVOR)(v.toString())}
+          >
+            {flavors.map((f) => {
+              return (
+                <SelectOption key={f} value={f}>
+                  {_.capitalize(f)}
+                </SelectOption>
+              );
+            })}
+          </FormPFSelect>
+        </FormField>
+      </FormFieldRow>
+    );
+  },
+);
+
+type FlavorProps = {
+  iUserTemplate: any;
+  commonTemplates: any;
+  flavorField: any;
+  os: string;
+  workloadProfile: string;
+  cnvBaseImages: any;
+  openshiftFlag: boolean;
+  onChange: (key: string, value: string | boolean) => void;
+};
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os.tsx
@@ -1,13 +1,5 @@
-import * as _ from 'lodash';
 import * as React from 'react';
-import {
-  FormSelect,
-  FormSelectOption,
-  Checkbox,
-  Text,
-  Button,
-  ButtonVariant,
-} from '@patternfly/react-core';
+import { Checkbox, Text, Button, ButtonVariant, SelectOption } from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
 import { ValidationErrorType, asValidationObject } from '@console/shared/src/utils/validation';
 import {
@@ -20,7 +12,6 @@ import {
 } from '../../../../utils/immutable';
 import { FormFieldRow } from '../../form/form-field-row';
 import { FormField, FormFieldType } from '../../form/form-field';
-import { FormSelectPlaceholderOption } from '../../../form/form-select-placeholder-option';
 import {
   getFlavors,
   getWorkloadProfiles,
@@ -29,7 +20,7 @@ import { flavorSort, ignoreCaseSort } from '../../../../utils/sort';
 import { pluralize } from '../../../../utils/strings';
 import { VMSettingsField } from '../../types';
 import { iGetFieldValue } from '../../selectors/immutable/field';
-import { getPlaceholder, getFieldId } from '../../utils/renderable-field-utils';
+import { getFieldId } from '../../utils/renderable-field-utils';
 import { nullOnEmptyChange } from '../../utils/utils';
 import { operatingSystemsNative } from '../../../../constants/vm-templates/os';
 import { OperatingSystemRecord } from '../../../../types';
@@ -51,22 +42,22 @@ import {
   CDI_UPLOAD_RUNNING,
 } from '../../../cdi-upload-provider/consts';
 import { getTemplateOperatingSystems } from '../../../../selectors/vm-template/advanced';
+import { FormPFSelect } from '../../../form/form-pf-select';
 
-export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
+export const OS: React.FC<OSProps> = React.memo(
   ({
     iUserTemplate,
     commonTemplates,
     operatinSystemField,
     cloneBaseDiskImageField,
     mountWindowsGuestToolsField,
-    flavorField,
+    flavor,
     workloadProfile,
     cnvBaseImages,
     onChange,
     openshiftFlag,
     goToStorageStep,
   }) => {
-    const flavor = iGetFieldValue(flavorField);
     const os = iGetFieldValue(operatinSystemField);
     const display = iGet(operatinSystemField, 'display');
     const displayOnly = !!display;
@@ -112,7 +103,6 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
     }
 
     let operatingSystemValidation;
-    let flavorValidation;
 
     if (
       iGetIsLoaded(commonTemplates) &&
@@ -125,8 +115,6 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
       );
       if (!operatinSystemField.get('validation')) {
         operatingSystemValidation = validation;
-      } else if (!flavorField.get('validation')) {
-        flavorValidation = validation;
       }
     }
 
@@ -199,22 +187,23 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
       <>
         <FormFieldRow
           field={operatinSystemField}
-          fieldType={FormFieldType.SELECT}
+          fieldType={FormFieldType.PF_SELECT}
           validation={operatingSystemValidation}
           loadingResources={loadingResources}
         >
-          <FormField value={displayOnly ? display : undefined}>
-            <FormSelect onChange={nullOnEmptyChange(onChange, VMSettingsField.OPERATING_SYSTEM)}>
-              {!displayOnly && (
-                <FormSelectPlaceholderOption
-                  placeholder={getPlaceholder(VMSettingsField.OPERATING_SYSTEM)}
-                  isDisabled={!!os}
-                />
-              )}
+          <FormField value={displayOnly ? display : os}>
+            <FormPFSelect
+              onSelect={(e, v) =>
+                nullOnEmptyChange(onChange, VMSettingsField.OPERATING_SYSTEM)(v.toString())
+              }
+            >
               {operatingSystemBaseImages.map(({ id, name, message }) => (
-                <FormSelectOption key={id} value={id} label={`${name || id} ${message}`} />
+                <SelectOption key={id} value={id}>
+                  {name || id}
+                  {message ? ` ${message}` : ''}
+                </SelectOption>
               ))}
-            </FormSelect>
+            </FormPFSelect>
           </FormField>
           {baseImage && baseImage?.longMessage && (
             <div className="pf-c-form__helper-text" aria-live="polite">
@@ -250,33 +239,15 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
           </FormField>
         </FormFieldRow>
         {mountedDisksHelpMsg}
-        <FormFieldRow
-          field={flavorField}
-          fieldType={FormFieldType.SELECT}
-          validation={flavorValidation}
-          loadingResources={loadingResources}
-        >
-          <FormField isDisabled={flavor && flavors.length === 1}>
-            <FormSelect onChange={nullOnEmptyChange(onChange, VMSettingsField.FLAVOR)}>
-              <FormSelectPlaceholderOption
-                placeholder={getPlaceholder(VMSettingsField.FLAVOR)}
-                isDisabled={!!flavor}
-              />
-              {flavors.map((f) => {
-                return <FormSelectOption key={f} value={f} label={_.capitalize(f)} />;
-              })}
-            </FormSelect>
-          </FormField>
-        </FormFieldRow>
       </>
     );
   },
 );
 
-type OSFlavorProps = {
+type OSProps = {
   iUserTemplate: any;
   commonTemplates: any;
-  flavorField: any;
+  flavor: string;
   operatinSystemField: any;
   cloneBaseDiskImageField: any;
   mountWindowsGuestToolsField: any;
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
@@ -1,13 +1,12 @@
 import * as React from 'react';
-import { FormSelect, FormSelectOption, Button, ButtonVariant } from '@patternfly/react-core';
-import { FormSelectPlaceholderOption } from '../../../form/form-select-placeholder-option';
+import { Button, ButtonVariant, SelectOption, Text, TextContent } from '@patternfly/react-core';
 import { ProvisionSource } from '../../../../constants/vm/provision-source';
 import { FormFieldRow } from '../../form/form-field-row';
 import { FormField, FormFieldType } from '../../form/form-field';
 import { iGetFieldValue } from '../../selectors/immutable/field';
 import { VMSettingsField } from '../../types';
-import { getPlaceholder } from '../../utils/renderable-field-utils';
 import { iGet } from '../../../../utils/immutable';
+import { FormPFSelect } from '../../../form/form-pf-select';
 
 const ProvisionSourceDiskHelpMsg: React.FC<ProvisionSourceDiskHelpMsgProps> = ({
   provisionSourceValue,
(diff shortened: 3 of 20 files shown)


glekner:
@suomiy @matthewcarleton thoughts?

matthewcarleton:
I guess an alternative to this would be to scroll the user down so the whole dropdown is visible.

glekner:
* OS and Flavor components are now separated.
* Source is now between them

Some of the dropdown's content is still hidden by the footer but at least the user will know to scroll.

@suomiy @yaacov @irosenzw can you review

matthewcarleton:
@glekner Is there a reason why the selection titles don't align with the design?
image

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/vm-settings-tab.tsx
```diff
@@ -128,6 +114,34 @@ export const VMSettingsTabComponent: React.FC<VMSettingsTabComponentProps> = ({
         onProvisionSourceStorageChange={updateStorage}
         provisionSourceStorage={provisionSourceStorage}
       />
+      <Flavor
```
yaacov, line 117: I like that we have flavor+mem+cpu together :+1:

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -0,0 +1,113 @@
...
+      </FormFieldRow>
+    );
+  },
+);
+
+type FlavorProps = {
+  userTemplates: any;
+  commonTemplates: any;
+  flavorField: any;
+  operatinSystemField: any;
```
suomiy, line 105: you can pass just the value, not the whole field

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os.tsx
```diff
@@ -42,7 +40,7 @@ import {
   NO_BASE_IMAGE_MESSAGE,
 } from '../../strings/strings';
 
-export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
+export const OS: React.FC<OSProps> = React.memo(
   ({
     userTemplates,
     commonTemplates,
```
suomiy, line 46: same here with the flavorField

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  cloneBaseDiskImageField: any;
```
suomiy, line 106: no need for these

suomiy (changes requested):

glekner:
* Source dropdown now overlaps the wizard footer
* removed unused props

can you approve @suomiy

jelkosz:
@glekner One small comment: please enrich the PXE part by a mention that you need a network attachment definition available for you in this namespace or in the default one. So the message might be like:
"Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. In order to PXE boot, you need to have a Network Attachment Definition available for you either in this namespace, or in the default one."

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -69,19 +75,33 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
...
+    };
+
     return (
       <FormFieldRow field={provisionSourceField} fieldType={FormFieldType.SELECT}>
-        <FormField>
-          <FormSelect onChange={(v) => onChange(VMSettingsField.PROVISION_SOURCE_TYPE, v)}>
-            <FormSelectPlaceholderOption
-              placeholder={getPlaceholder(VMSettingsField.PROVISION_SOURCE_TYPE)}
-              isDisabled={!!provisionSourceValue}
+        <Select
```
suomiy, line 87: can you please create a new type for FormField and use it here? This will allow to pass correct values and metadata from redux to this component.

suomiy (changes requested):

glekner:
@jelkosz updated PXE text
@suomiy added `PF_SELECT` field type. the only supported prop I can see is `isDisabled`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -69,18 +75,34 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
...
+      <FormFieldRow field={provisionSourceField} fieldType={FormFieldType.PF_SELECT}>
         <FormField>
-          <FormSelect onChange={(v) => onChange(VMSettingsField.PROVISION_SOURCE_TYPE, v)}>
-            <FormSelectPlaceholderOption
-              placeholder={getPlaceholder(VMSettingsField.PROVISION_SOURCE_TYPE)}
-              isDisabled={!!provisionSourceValue}
-            />
+          <Select
+            isOpen={isSourceOpen}
+            selections={provisionSourceValue}
```
suomiy, line 90: we can pass this one in the FormField as well

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -69,18 +75,34 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
...
+            onToggle={(isOpen) => setIsSourceOpen(isOpen)}
```
suomiy, line 91: maybe it would be beneficial to create a single select component that has all the defaults included and stores its own state.
This will come in handy once we include multiple of these in.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,16 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
 export const ADD_SNAPSHOT = 'Take Snapshot';
 export const RESTORE_SNAPSHOT = 'Restore Snapshot';
 
+// provision sources help text
+export const PROVISION_SOURCE_PXE_HELPTEXT =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. In order to PXE boot, you need to have a Network Attachment Definition available for you either in this namespace, or in the default one';
+export const PROVISION_SOURCE_CONTAINER_HELPTEXT =
+  'Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/cirrosregistry-disk-demo. This action will not create a persistent volume claim (PVC). This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_URL_HELPTEXT =
+  'Link to an operating system image via URL (HTTP or S3 endpoint). A new persistent volume claim (PVC) will be created to store this image. This will show up as a disk in the Storage step';
```
jelkosz, line 27: Maybe would be good to add an example as well, something like:

"For example, for CentOS you can visit http://isoredirect.centos.org/centos/8/isos/x86_64/ and pick a URL for an ISO which will look like: http://mirror.hosting90.cz/centos/8.2.2004/isos/x86_64/CentOS-8.2.2004-x86_64-boot.iso"

@matthewcarleton WDYT? Would something like this be helpful?

glekner:
@suomiy 
* added labels to `ProvisionSource`
* created `FormPFSelect` component

frontend/packages/kubevirt-plugin/src/utils/strings.ts
jelkosz: OK, that might be a bit too much and something which will be changing so we should not hardcode it here. But we could have something which will help. Something like:
"Link to an operating system image via URL (HTTP or S3 endpoint). An example might be a download link to an operating system iso you can copy from the downloads page of your favourite distribution. A new persistent volume claim (PVC) will be created to store this image. This will show up as a disk in the Storage step.".

@matthewcarleton WDYT? Would this be understandable?

matthewcarleton:
> @glekner One small comment: please enrich the PXE part by a mention that you need a network attachment definition available for you in this namespace or in the default one. So the message might be like:
> "Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. In order to PXE boot, you need to have a Network Attachment Definition available for you either in this namespace, or in the default one."

I put this together which could solve this
Screen Shot 2020-09-02 at 9 44 25 AM

glekner:
* Added `FormPFSelect` component
* Created `SelectDropdownObjectEnum` class for abstraction and reusability.
* Changed OS, Flavor, Workload, Disk Source, NIC Model, NIC Type to use the new component.
* Flavor, Workload don't have descriptions as of now.

Screen Shot 2020-09-02 at 18 36 10

Screen Shot 2020-09-02 at 18 35 46

Screen Shot 2020-09-02 at 18 38 21

Screen Shot 2020-09-02 at 18 38 29

frontend/packages/kubevirt-plugin/src/utils/strings.ts
matthewcarleton: Maybe:

Link to an operating system source via URL (HTTP or S3). This will show up as a disk in the storage step (backed by a persistent volume claim). **Example: A operating system ISO download link you can copy from your favorite distribution**.

For brevity.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -70,17 +70,32 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
...
+            selections={provisionSourceValue}
+            placeholderText={getPlaceholder(VMSettingsField.PROVISION_SOURCE_TYPE)}
+            onSelect={(e, v) => {
+              onChange(VMSettingsField.PROVISION_SOURCE_TYPE, v.toString());
+            }}
+          >
+            {(sources || [])
+              .sort(
+                (a, b) =>
+                  ProvisionSource.fromString(a).getOrder() -
```
suomiy, line 86: you can do the map first

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os.tsx
```diff
@@ -193,22 +178,25 @@ export const OSFlavor: React.FC<OSFlavorProps> = React.memo(
...
+              selections={os}
+              placeholderText={getPlaceholder(VMSettingsField.OPERATING_SYSTEM)}
+              onSelect={(e, v) =>
+                nullOnEmptyChange(onChange, VMSettingsField.OPERATING_SYSTEM)(v.toString())
+              }
+            >
               {operatingSystemBaseImages.map(({ id, name, message }) => (
-                <FormSelectOption key={id} value={id} label={`${name || id} ${message}`} />
+                <SelectOption key={id} value={id}>
+                  {`${name || id} ${message}`}
```
suomiy, line 196: can render an extra space

frontend/packages/kubevirt-plugin/src/constants/select-dropdown-object-enum.ts
```diff
@@ -0,0 +1,28 @@
...
+  private readonly order: number;
+
+  protected constructor(value: T, { description, label, order }: SelectDropdownData = {}) {
+    super(value);
+    this.label = label;
+    this.description = description;
+    this.order = order;
+  }
+
+  getLabel = () => this.label;
```
suomiy, line 23: can we use toString method here instead to be consistent. We use this approach in other places as well.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
suomiy: can we also specify the formats?

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
 export const ADD_SNAPSHOT = 'Take Snapshot';
 export const RESTORE_SNAPSHOT = 'Restore Snapshot';
 
+// provision sources descriptions
+export const PROVISION_SOURCE_PXE_DESC =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. Requires a PXE bootable network attachment definition';
+export const PROVISION_SOURCE_CONTAINER_DESC =
+  'Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/cirros-registry-disk-demo. This action will not create a persistent volume claim (PVC). This will show up as a disk in the Storage step';
```
suomiy, line 25: Do we have to specify this will not create a PVC? Kubernetes use should probably know what container is.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
+export const PROVISION_SOURCE_URL_DESC =
+  'Link to an operating system image via URL (HTTP or S3 endpoint). A new persistent volume claim (PVC) will be created to store this image. This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_DISK_DESC =
+  'Use a persistent volume claim (PVC) already available in the cluster. This will show up as a disk in the Storage step';
+
+// storage ui sources descriptions
+export const UI_SOURCE_BLANK_DESC = 'Create an empty disk (PVC)';
+export const UI_SOURCE_URL_DESC =
```
suomiy, line 33: +formats?

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
 
+// provision sources descriptions
+export const PROVISION_SOURCE_PXE_DESC =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. Requires a PXE bootable network attachment definition';
+export const PROVISION_SOURCE_CONTAINER_DESC =
+  'Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/cirros-registry-disk-demo. This action will not create a persistent volume claim (PVC). This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_URL_DESC =
+  'Link to an operating system image via URL (HTTP or S3 endpoint). A new persistent volume claim (PVC) will be created to store this image. This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_DISK_DESC =
+  'Use a persistent volume claim (PVC) already available in the cluster. This will show up as a disk in the Storage step';
```
suomiy, line 29: can we reuse the message already defined in UI_SOURCE_ATTACH_DISK_DESC and append the further description? Plust other types as well.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
+  'Upload content from a container located in a registry accessible from the cluster. Example: kubevirt/cirros-registry-disk-demo. A new persistent volume claim (PVC) will be created';
+export const UI_SOURCE_ATTACH_CLONED_DISK_DESC =
+  'Clone a persistent volume claim (PVC) already available in the cluster';
+export const UI_SOURCE_ATTACH_DISK_DESC =
+  'Use a persistent volume claim (PVC) already available on the cluster';
+export const UI_SOURCE_IMPORT_DISK_DESC = 'TBD';
+
+// nics descriptions
+export const NIC_MODEL_VIRTIO_DESC =
+  'Optimized for best performance. Windows VMs requires virtio drivers to function.';
```
suomiy, line 45:
```suggestion
  'Optimized for best performance. Windows VMs require virtio drivers to function.';
```

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
+export const NIC_TYPE_MASQUERADE_DESC =
+  'Put the VM behind a NAT Proxy for high compability with different network providers. The VMs IP will differ from the IP seen on the pod network';
+export const NIC_TYPE_BRIDGE_DESC =
+  'The VM will be bridged to the selected network, ideal for L2 devices';
+export const NIC_TYPE_SRIOV_DESC =
+  'Attach a virtual function network device to the VM for very high performance';
```
suomiy, line 51:
```suggestion
  'Attach a virtual function network device to the VM for high performance';
```
isn't this enough?

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/utils/strings.ts
jelkosz: I would explicitely add the download page there: Link to an operating system source via URL (HTTP or S3). This will show up as a disk in the storage step (backed by a persistent volume claim). Example: An operating system ISO download link you can copy from the download page of your favorite distribution.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
glekner: don't think so, the map returns a different array with `[<SelectOption>]`

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/os.tsx
suomiy: because the message can be empty

frontend/packages/kubevirt-plugin/src/constants/vm/network/network-interface-model.ts
```diff
@@ -1,15 +1,39 @@
...
+  static readonly E1000 = new NetworkInterfaceModel('e1000', {
+    label: 'E1000',
+    order: 2,
+  });
+  static readonly E1000E = new NetworkInterfaceModel('e1000e', {
+    label: 'E1000e',
+    order: 3,
+  });
+  static readonly NE2kPCI = new NetworkInterfaceModel('ne2kPCI', {
+    label: 'ne2kPCi',
```
suomiy, line 22: small i? Let's rather default it to the value here

frontend/packages/kubevirt-plugin/src/constants/vm/network/network-interface-model.ts
```diff
@@ -1,15 +1,39 @@
...
+    order: 4,
+  });
+  static readonly PCNET = new NetworkInterfaceModel('pcnet', {
+    label: 'Pcnet',
```
suomiy, line 26:
```suggestion
    label: 'PCnet',
```

looks better?

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
+
+// provision sources descriptions
+export const PROVISION_SOURCE_PXE_DESC =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. Requires a PXE bootable network attachment definition';
+export const PROVISION_SOURCE_CONTAINER_DESC =
+  'Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/cirros-registry-disk-demo. This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_URL_DESC =
+  'Link to an operating system image via URL (HTTP or S3 endpoint). A new persistent volume claim (PVC) will be created to store this image. This will show up as a disk in the Storage step';
+export const PROVISION_SOURCE_DISK_DESC = `
+  ${UI_SOURCE_ATTACH_DISK_DESC}. This will show up as a disk in the Storage step`;
```
gouyang, line 41: It's not showing up a disk in the Storage step.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,38 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
...
+  'Upload content from a container located in a registry accessible from the cluster. Example: kubevirt/cirros-registry-disk-demo. A new persistent volume claim (PVC) will be created';
+export const UI_SOURCE_ATTACH_CLONED_DISK_DESC =
+  'Clone a persistent volume claim (PVC) already available in the cluster';
+export const UI_SOURCE_ATTACH_DISK_DESC =
+  'Use a persistent volume claim (PVC) already available on the cluster';
+export const UI_SOURCE_IMPORT_DISK_DESC = 'TBD';
+
+// provision sources descriptions
+export const PROVISION_SOURCE_PXE_DESC =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. Requires a PXE bootable network attachment definition';
```
gouyang, line 35: The show up NIC is "Pod Networking" nic, it shows up for each provision source, maybe not worth mention it here.

frontend/packages/kubevirt-plugin/src/utils/strings.ts
```diff
@@ -18,6 +18,16 @@ export const ADD_NETWORK_INTERFACE = 'Add Network Interface';
 export const ADD_SNAPSHOT = 'Take Snapshot';
 export const RESTORE_SNAPSHOT = 'Restore Snapshot';
 
+// provision sources help text
+export const PROVISION_SOURCE_PXE_HELPTEXT =
+  'Boot an operating system from a server on a network. This will show up as a NIC on the Networking step. In order to PXE boot, you need to have a Network Attachment Definition available for you either in this namespace, or in the default one';
+export const PROVISION_SOURCE_CONTAINER_HELPTEXT =
+  'Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/cirrosregistry-disk-demo. This action will not create a persistent volume claim (PVC). This will show up as a disk in the Storage step';
```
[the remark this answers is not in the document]
jelkosz: So, we need to highlight the limitations of this and reference fedora instead of cirros:
Link to a bootable operating system container located in a registry accessible from the cluster. Example: kubevirt/fedora-cloud-container-disk-demo. This action will not create a persistent volume claim (PVC). This will show up as a disk in the Storage step. The container disk is meant to be used only for read-only filesystems such
as CD-ROMs or for small short-lived throw-away VMs.';

And then, once picked, under it it needs to show the example kubevirt/fedora-cloud-container-disk-demo so the user can actually copy paste it to the text box.

yaacov:
Gilad, hi, make sure that you didn't miss any `ProvisionSource[.].*[.]toString()` instance that should be `ProvisionSource[.].*[.]getValue()`.

It may be just some bad rebase, but while preparing the merged PR I found some.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -26,9 +27,31 @@ const ProvisionSourceDiskHelpMsg: React.FC<ProvisionSourceDiskHelpMsgProps> = ({
   const getStorageMsg = () => {
     switch (ProvisionSource.fromString(provisionSourceValue)) {
       case ProvisionSource.URL:
-        return <>Enter URL here or edit the mounted disk in the {storageBtn} step</>;
+        return (
+          <>
+            <Text> Enter URL here or edit the mounted disk in the {storageBtn} step</Text>
+            <Text>
+              For example, you can obtain the download link for {isUpstream ? 'Fedora' : 'RHEL'}{' '}
```
matthewcarleton, line 34: These examples should be moved below the relevant field. So in this case it would be under the URL input.

matthewcarleton:
Can we change the "Source" label to "boot source"?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
glekner: the URL field is right below it. changing the help providers for each field is not in the scope.
we should merge the changes and if we still find it wrong, we should open a new bug.

glekner:
- Rebased
- Changed title of provision source from 'Soruce' to 'Boot Source'

can we merge @yaacov @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/url-source.tsx
```diff
@@ -34,6 +36,17 @@ export const URLSource: React.FC<URLSourceProps> = React.memo(
             }
           />
         </FormField>
+        <div className="pf-c-form__helper-text" aria-live="polite">
+          For example, you can obtain the download link for {isUpstream ? 'Fedora' : 'RHEL'} cloud
```
matthewcarleton, line 40: can we change this to:
"Visit the Fedora cloud image[link] list and copy a url to paste into the url field here"

yaacov: @glekner ^^

PR #6434, opened 2020-08-25
Bug 1872421: Added a few points of clarification to README

reginapizza:
When both myself and another intern were attempting to follow the instructions for setting up the console for development for the first time, especially in the `Frontend Development` section, I noticed that we both got tripped up on the same things. Just wanted to make a few points of clarification for others so hopefully it's a little easier for folks who are trying to set it up for the first time.


jhadvig (approved):
/test e2e-gcp-console

PR #6437, opened 2020-08-25
bug 1871109: Add a user friendly error message in VM import wizard

irosenzw:

When non-admin user picks a provider to import a VM from and lacks the permissions to do so.
minimized_error

The error expended :
maximaized_error

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/utils.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/utils.ts
@@ -10,6 +10,10 @@ import { iGetIn, toJS } from '../../../utils/immutable';
 import { joinGrammaticallyListOfItems } from '@console/shared/src';
 import { getValidationNameByKey } from '../../../utils/validations/strings';
 import { iGetStorages } from '../selectors/immutable/storage';
+import {
+  INSUFFICIENT_PERMISSIONS_ERROR_TITLE,
+  INSUFFICIENT_PERMISSIONS_ERROR_DESC,
+} from '../../../constants/errors/common';
 
 export const computeVMSettingsErrors = (state, wizardReduxID) => {
   const vmSettingsFields = iGetVmSettings(state, wizardReduxID);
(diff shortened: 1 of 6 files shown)


matthewcarleton:
How about this for the content?
Screen Shot 2020-08-25 at 2 15 59 PM
I guess all our alerts surface at the bottom of the wizard right? 
I wonder if we should also show an error on the provider dropdown to?
"Insufficient permissions" WDYT?

irosenzw:
The error message suppose to be generic so it will suites other pages as well (like creating a VM), not just VM import.

matthewcarleton:
Does it have to be generic? If so, we could drop the part about importing but leave the rest.

irosenzw:
@matthewcarleton yeah, I think that can work - I do like the `insufficient permissions` title.
Also, can we drop "or view more details below" part at the end?
It seems redundant since the user can do it in addition to contacting the administrator and most of the time he will view the details even if we won't instruct him to do so.

irosenzw:
@matthewcarleton, updated error message:
Screenshot from 2020-08-26 11-36-51

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/utils.ts
```diff
@@ -163,3 +167,30 @@ export const computeStorageErrors = (state, wizardReduxID) => {
 
   return { hasAllRequiredFilled, isValid, errors };
 };
+
+const isPermissionError = (title: string, message: string): boolean =>
+  title.includes('roles.rbac.authorization.k8s.io') ||
```
suomiy, line 172: this is only a specific error for operating on roles. IMO we should check the http code and mention 403 all errors.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/utils.ts
```diff
@@ -163,3 +167,30 @@ export const computeStorageErrors = (state, wizardReduxID) => {
...
+
+  if (isPermissionError(errTitle, message)) {
+    return {
+      isLong: true,
+      errTitle: INSUFFICIENT_PERMISSIONS_ERROR_TITLE,
+      description: INSUFFICIENT_PERMISSIONS_ERROR_DESC,
+    };
+  }
+
+  const [desc] = message.split('\n');
```
suomiy, line 194: what is the use case? in what conditions does this occur?

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/error-components/utils.ts
irosenzw: agree.

matthewcarleton:
> @matthewcarleton yeah, I think that can work - I do like the `insufficient permissions` title.
> Also, can we drop "or view more details below" part at the end?
> It seems redundant since the user can do it in addition to contacting the administrator and most of the time he will view the details even if we won't instruct him to do so.

ya I think it's ok to drop that part of the text and just have the View details action.

suomiy:
this bug/PR is fixed and superseeded by #6480

PR #6439, opened 2020-08-25
Bug 1872480: Fixes node-list alignment on storageCluster creation

a2batic:
Before: 
before_install

After:
after_install

JIra: https://issues.redhat.com/browse/RHSTOR-1252

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/node-list.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/node-list.tsx
@@ -20,38 +20,35 @@ import { GetRows, NodeTableProps } from './types';
 import './ocs-install.scss';
 
 const tableColumnClasses = [
-  classNames('col-md-1', 'col-sm-1', 'col-xs-1'),
-  classNames('col-md-4', 'col-sm-8', 'col-xs-11'),
-  classNames('col-md-2', 'col-sm-3', 'hidden-xs'),
-  classNames('col-md-2', 'hidden-sm', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
-  classNames('col-md-2', 'hidden-sm', 'hidden-xs'),
+  classNames('pf-u-w-50-on-sm'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-sm', 'pf-u-w-10-on-sm'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-sm', 'pf-u-w-10-on-sm'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-sm', 'pf-u-w-10-on-sm'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-sm', 'pf-u-w-10-on-sm'),
 ];
 
-export const getColumns = () => {
-  return [
-    {
-      title: 'Name',
-      props: { className: tableColumnClasses[1] },
-    },
-    {
-      title: 'Role',
-      props: { className: tableColumnClasses[2] },
-    },
-    {
-      title: 'Zone',
-      props: { className: tableColumnClasses[3] },
-    },
-    {
-      title: 'CPU',
-      props: { className: tableColumnClasses[4] },
-    },
-    {
-      title: 'Memory',
-      props: { className: tableColumnClasses[5] },
-    },
-  ];
-};
+export const getColumns = () => [
+  {
+    title: 'Name',
+    props: { className: tableColumnClasses[0] },
+  },
+  {
+    title: 'Role',
+    props: { className: tableColumnClasses[1] },
+  },
+  {
+    title: 'Zone',
+    props: { className: tableColumnClasses[2] },
+  },
+  {
+    title: 'CPU',
+    props: { className: tableColumnClasses[3] },
+  },
+  {
+    title: 'Memory',
+    props: { className: tableColumnClasses[4] },
+  },
+];
 
 const getRows: GetRows = (
   { componentProps },
@@ -70,13 +67,13 @@ const getRows: GetRows = (
     const memSpec: string = getNodeAllocatableMemory(node);
     const cells: IRow['cells'] = [
       {
-        title: <ResourceLink kind="Node" name={getName(node)} title={node.metadata.uid} />,
+        title: <ResourceLink kind="Node" name={getName(node)} title={getName(node)} />,
       },
       {
-        title: roles.join(', ') || '-',
+        title: roles.join(', ') ?? '-',
       },
       {
-        title: node.metadata.labels?.['failure-domain.beta.kubernetes.io/zone'] || '-',
+        title: node.metadata.labels?.['failure-domain.beta.kubernetes.io/zone'] ?? '-',
       },
       {
         title: `${humanizeCpuCores(cpuSpec).string || '-'}`,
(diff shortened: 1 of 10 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-install.scss
```diff
@@ -43,3 +43,18 @@
   margin-top: var(--pf-global--spacer--sm);
   margin-bottom: var(--pf-global--spacer--sm);
 }
+
+// Restrict form widths
+.ceph-ocs-install__node-list {
```
gnehapk, line 48: @a2batic If I am not mistaken, these are the same styles that gets applied with `co-m-pane__form`

a2batic:
/hold
There are few more node lists in Attached devices installation that requires to be fixed.

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
```diff
@@ -42,3 +42,14 @@
...
+// Restrict node-list widths
+.lso-create-lvs__node-list {
+  @media (min-width: 769px) {
+    max-width: 100% !important;
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 70% !important;
+  }
+}
```
afreen23, line 55: missing new line

frontend/packages/local-storage-operator-plugin/src/components/auto-detect-volume/auto-detect-volume.scss
```diff
@@ -6,3 +6,13 @@
...
+
+.auto-detect-volume__node-list {
+  @media (min-width: 769px) {
+    max-width: 100% !important;
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 70% !important;
+  }
+}
```
afreen23, line 18: missing new line

frontend/packages/local-storage-operator-plugin/src/components/auto-detect-volume/auto-detect-volume.scss
```diff
@@ -6,3 +6,13 @@
...
   margin-top: 0;
 }
+
+.auto-detect-volume__node-list {
+  @media (min-width: 769px) {
+    max-width: 100% !important;
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 70% !important;
```
afreen23, line 16: can we avoid `!important` ?

afreen23 (changes requested):

frontend/packages/local-storage-operator-plugin/src/components/auto-detect-volume/auto-detect-volume.scss
a2batic: We need to override some media queries in `co-m-pane__form`, so we need to use `!important`

cloudbehl:
@a2batic should the text be expanded to 100% or should be of same width as node list page(70%)?

a2batic:
@cloudbehl as per design https://marvelapp.com/prototype/ai19e9g/screen/69554628, the text should be expanded 100%.

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
```diff
@@ -37,3 +37,14 @@
...
+
+// Restrict node-list widths
+.lso-create-lvs__node-list {
+  @media (min-width: 769px) {
+    max-width: 100% !important;
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 70% !important;
+  }
```
gnehapk, line 49: Instead of defining the same styles at two places, can we define at once and use at both?

a2batic:
@gnehapk have removed the style from the `auto-detect` component and used the same as defined in `local-volume-set`. Please review.

cloudbehl:
> @cloudbehl as per design https://marvelapp.com/prototype/ai19e9g/screen/69554628, the text should be expanded 100%.

Ok, then we need to increase the font to look similar & also gray out the text. Also, we have divided then into two different paragraphs but in UX it's only one.

gnehapk (approved):

a2batic:
@cloudbehl, ack, in this BZ we are targetting only node-list for the view, for other issues, we can address in a seperate BZ.

cloudbehl:
I think it's a smaller change we can accommodate in the same as alignment to 100% is not making it look good. & comment on Bugzilla about the change.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-install.scss
```diff
@@ -47,3 +47,14 @@
   display: flex;
   justify-content: space-between;
 }
+
+// Restrict node-list widths
+.ceph-ocs-install__node-list {
+  @media (min-width: 769px) {
+    max-width: 100% !important;
```
cloudbehl, line 54: Is it not working without `!important`? It's not a good practice to use it.

a2batic: We need to override some media queries in `co-m-pane__form`, so we need to use `!important`

cloudbehl: Ahh is it marked important? Then only we require important over here otherwise internal class always has priority.

cloudbehl (approved):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-install.scss
a2batic: yes, but since we are using both classes here as internal class https://github.com/openshift/console/pull/6439/files#diff-10211f95caf12c0c4e086214ec83e140R120, thats why used `!important`.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/node-list.tsx
```diff
@@ -120,7 +117,7 @@ const NodeTable: React.FC<NodeTableProps> = (props) => {
 
   return (
     <>
-      <div className="ceph-node-list__max-height">
+      <div className="ceph-node-list__max-height co-m-pane__form ceph-ocs-install__node-list">
```
afreen23, line 120:
```suggestion
      <div className="ceph-node-list__max-height  ceph-ocs-install__node-list">
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-install.scss
```diff
@@ -47,3 +47,14 @@
...
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 70% !important;
+  }
+}
```
afreen23, line 60:
```suggestion
.ceph-ocs-install__node-list {
  @media (min-width: 769px) {
    max-width: 100%;
  }

  @media (min-width: 1200px) {
    max-width: 70%;
  }
}
```

afreen23 (changes requested):

afreen23:
If you are setting your own custom widths then its redundant to override "console" based widths

afreen23 (approved):
/lgtm

PR #6448, opened 2020-08-26
Bug 1877860: Volume Snapshot basic CRUD tests

cloudbehl:

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
+import { Link } from 'react-router-dom';
 import { match } from 'react-router';
 
 import { Grid, GridItem, ActionGroup, Button, Alert } from '@patternfly/react-core';
@@ -142,7 +143,15 @@ const PVCSummary: React.FC<PVCSummaryProps> = ({ persistentVolumeClaim }) => {
 };
 
 const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
-  const { resourceName, namespace, kindObj, handlePromise, inProgress, errorMessage } = props;
+  const {
+    resourceName,
+    plural,
+    namespace,
+    kindObj,
+    handlePromise,
+    inProgress,
+    errorMessage,
+  } = props;
 
   const [pvcName, setPVCName] = React.useState(resourceName);
   const [pvcObj, setPVCObj] = React.useState<PersistentVolumeClaimKind>(null);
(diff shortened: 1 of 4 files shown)


dtaylor113:
Hi @cloudbehl are this tests in addition too, and will they impact/conflict with:  #6340 ?

bipuladh:
/hold
We need to separate the tests from the Bug fixes.
https://github.com/openshift/console/pull/6340 also handles CRUD tests of Volume Snapshot already.

cloudbehl:
@dtaylor113 Yes these are additional test which helps generic flows like YAML based creation, edit resource & deletion of the resource.

@bipuladh canceling hold as discussed. The bug is related to tests where YAML creation button was not present & the tests were not getting passed.

/hold cancel

dtaylor113:
Hi @cloudbehl, irt. the changes to Cypress integration test changes in `k8-openshift-cruds.spec.ts`, those LGTM.   I'm not familiar with the other Volume Snapshot changes made.

frontend/public/models/yaml-templates.ts
```diff
@@ -1210,6 +1210,48 @@ spec:
...
+deletionPolicy: Delete
+`,
+  )
+  .setIn(
+    [referenceForModel(k8sModels.VolumeSnapshotContentModel), 'default'],
+    `
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotContent
+metadata:
+  name: new-snapshot-content-test
```
spadgett, line 1245: Should we have "test" in the name if this is going to be the real volume snapshot template? (Here are elsewhere in the samples.)

We try to make the examples just work if you click save with no changes, but I imagine that's not possible here.

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -201,84 +210,97 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
...
-        </ButtonBar>
-      </form>
+          {pvcObj && (
+            <div className="form-group co-volume-snapshot__form">
+              <label className="control-label co-required" htmlFor="snapshot-class">
+                Snapshot Class
+              </label>
+              <SnapshotClassDropdown
+                onChange={setSnapshotClassName}
+                id="claimName"
```
spadgett, line 281: Duplicate ID.

frontend/public/models/yaml-templates.ts
cloudbehl: > Should we have "test" in the name if this is going to be the real volume snapshot template? (Here are elsewhere in the samples.)
Right, we can call it an example/test. I'll change it.
> We try to make the examples just work if you click save with no changes, but I imagine that's not possible here.
I agree, Right now volume snapshots are not adapted to all the storage providers as it's not GA'ed. When snapshot gets GA'ed we should get the default CSI hostpath driver out of the box. So for now just added an example of it as this is provided on k8s documentation as well.

cloudbehl:
@spadgett updated the PR please have a look.

spadgett (approved):
/lgtm

cloudbehl:
/hold cancel
Rebased the PR!

dtaylor113 (approved):
/lgtm -thanks

spadgett (approved):
/approve

PR #6449, opened 2020-08-26
Revert "Bug 1868724: Fix cypress test failures"

spadgett:
Reverts openshift/console#6319

Seeing if we can revert this now that #6431 has merged.
/cc @dtaylor113 @jhadvig


spadgett:
/bugzilla refresh

@dtaylor113 FYI, this passed CI. I think we're good to revert this change.

PR #6455, opened 2020-08-26
Bug 1873147: cypress crud tests - moved create & delete out of before/after[all] hooks and into specific tests

dtaylor113:
This will allow us to take advantage of Cypress's new native retry ability which does not work in before/after hooks as shown in this previous message: "Although you have test retries enabled, we do not retry tests when `before all` or `after all` hooks fail"

It also helps readability of logs and console output:
```
  Pod
      ✓ creates the resource instance (9304ms)
      ✓ displays detail view for newly created resource instance (806ms)
      ✓ displays a list view for the resource (3103ms)
      ✓ search view displays created resource instance (3589ms)
      ✓ edits the resource instance (8998ms)
      ✓ deletes the resource instance (13528ms)

```

### DIFF
--- a/frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
@@ -69,8 +69,7 @@ describe('Kubernetes resource CRUD operations', () => {
     describe(kind, () => {
       const name = `${testName}-${_.kebabCase(kind)}`;
 
-      const createResource = () => {
-        cy.log(`Create ${kind}: ${name}`);
+      it(`creates the resource instance`, () => {
         cy.visit(
           `${namespaced ? `/k8s/ns/${testName}` : '/k8s/cluster'}/${resource}?name=${testName}`,
         );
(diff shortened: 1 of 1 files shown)


frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
```diff
@@ -69,8 +69,7 @@ describe('Kubernetes resource CRUD operations', () => {
     describe(kind, () => {
       const name = `${testName}-${_.kebabCase(kind)}`;
 
-      const createResource = () => {
-        cy.log(`Create ${kind}: ${name}`);
+      it(`creates ${kind} named '${name}'`, () => {
```
spadgett, line 72: Don't add a generated name into the test. We want the title to be the same every run.

spadgett:
@bparees FYI, this should remove some of the before all / after all tests you were seeing.

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
dtaylor113: ok, fixed, pls see updated output in PR Description

spadgett:
Hey @dtaylor113, Bug 1868724 is specifically for CI failures due to the logout problem. I'd rather not overload these Bugzilla bugs or we'll get confused. Can you open a separate BZ?

dtaylor113:
@spadgett, created new BZ -thanks

spadgett:
Thanks @dtaylor113 👍

PR #6457, opened 2020-08-27
Bug 1872986: Replace Not Firing Alert State with - and Not Firing filter with Inactive

vikram-raj:
Fix - https://issues.redhat.com/browse/ODC-4659

**Screenshots**
Screenshot 2020-08-27 at 10 35 35 AM


vikram-raj:
/retitle Bug 1872986: Replace Not Firing Alert State with - and Not Firing filter with Inactive

PR #6460, opened 2020-08-27
Bug 1871676: Monitoring: Change rules list to filter by alert state, not rule state

kyoto:
Instead of filtering by rule state ("Active" or "Inactive") allow
filtering the rules by alert state ("Firing", "Pending", "Silenced" or
"Not Firing").

This now mirrors the states shown in the Alert State table column.
Having the filter states be different from the states in the column was
confusing.

screenshot

### DIFF
--- a/frontend/public/components/factory/table.tsx
+++ b/frontend/public/components/factory/table.tsx
@@ -428,6 +428,7 @@ export type TableProps = {
   loaded?: boolean;
   reduxID?: string;
   reduxIDs?: string[];
+  rowFilters?: any[];
   label?: string;
   columnManagementID?: string;
 };
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -36,7 +36,6 @@ import {
 import { K8sKind } from '../../module/k8s';
 import {
   alertDescription,
-  alertingRuleIsActive,
   alertingRuleSource,
   alertSource,
   alertState,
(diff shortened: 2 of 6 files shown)


frontend/public/components/factory/table.tsx
```diff
@@ -428,6 +428,7 @@ export type TableProps = {
   loaded?: boolean;
   reduxID?: string;
   reduxIDs?: string[];
+  rowFilters?: any[];
```
kyoto, line 431: Not sure why this wasn't here, but `Table` does accept the `rowFilters`, so I think it should be added.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1301,15 +1301,24 @@ const AlertsPage_: React.FC<Alerts> = ({ data, loaded, loadError }) => (
...
+
+const ruleAlertStateFilter = (filter, rule: Rule) =>
+  _.some(rule.alerts, (a) => filter.selected.has(a.state)) || _.isEmpty(filter.selected);
+
 const rulesRowFilters: RowFilter[] = [
   {
-    filterGroupName: 'Rule State',
+    filter: ruleAlertStateFilter,
+    filterGroupName: 'Alert State',
+    isMatch: ruleHasAlertState,
```
kyoto, line 1314: For this change, we can't use the `reducer` because we can't simply reduce each rule to a single value. Instead, we need to iterate through the rule's alerts looking for any one that matches the given alert state.

PR #6463, opened 2020-08-27
Bug 1873192: show empty state in monitoring when no datapoint available

nemesis09:
Fixes:
https://issues.redhat.com/browse/ODC-4469
https://issues.redhat.com/browse/ODC-3297

Analysis / Root cause:
the monitoring tab shows empty charts when there are no datapoints available
Solution Description:
show empty state when there are no datapoints

Screens:
**Before**
Screenshot from 2020-08-27 20-30-00
Screenshot from 2020-08-27 20-30-54
Screenshot from 2020-08-27 20-31-26

**After**
Screenshot from 2020-08-27 20-34-30
Screenshot from 2020-08-27 20-34-49
Screenshot from 2020-08-27 20-33-51

Test Coverage:
NA

Browser Conformance:
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/monitoring/query-browser.tsx
+++ b/frontend/public/components/monitoring/query-browser.tsx
@@ -680,10 +680,13 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
     setSpan(to - from);
   };
 
+  const isGraphDataEmpty = !graphData || graphData.every((d) => d.length === 0);
+
   return (
     <div
       className={classNames('query-browser__wrapper', {
-        'graph-empty-state': _.isEmpty(graphData),
+        'graph-empty-state': isGraphDataEmpty,
+        'graph-empty-state__loaded': isGraphDataEmpty && !updating,
       })}
     >
       {hideControls ? (
@@ -702,8 +705,8 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
         </div>
       )}
       {error && <Error error={error} />}
-      {_.isEmpty(graphData) && !updating && <GraphEmpty />}
-      {!_.isEmpty(graphData) && (
+      {isGraphDataEmpty && !updating && <GraphEmpty />}
+      {!isGraphDataEmpty && (
         <>
           {samples < maxSamplesForSpan && !updating && (
             <Alert
--- a/frontend/public/components/graphs/graph-empty.tsx
+++ b/frontend/public/components/graphs/graph-empty.tsx
@@ -9,6 +9,7 @@ export const GraphEmpty: React.FC<GraphEmptyProps> = ({ height = 180, loading = 
       justifyContent: 'center',
       padding: '5px',
       width: '100%',
+      flexGrow: 1,
     }}
   >
     {loading ? (
--- a/frontend/public/components/monitoring/_monitoring.scss
+++ b/frontend/public/components/monitoring/_monitoring.scss
@@ -19,6 +19,12 @@ $tooltip-background-color: #151515;
 
 .graph-empty-state {
   min-height: 310px;
+  &__loaded {
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+  }
 }
 
 .graph-wrapper__horizontal-bar {
@@ -194,7 +200,9 @@ $screen-phone-landscape-min-width: 567px;
 
 $monitoring-line-height: 18px;
 
-.monitoring-description, .monitoring-label-list, .monitoring-timestamp {
+.monitoring-description,
+.monitoring-label-list,
+.monitoring-timestamp {
   line-height: $monitoring-line-height;
   margin-top: 4px;
 }
@@ -239,7 +247,7 @@ $monitoring-line-height: 18px;
     align-items: baseline;
     display: flex;
     justify-content: space-between;
-    @media(min-width: $grid-float-breakpoint) {
+    @media (min-width: $grid-float-breakpoint) {
       padding-top: 26px;
     }
   }


nemesis09:
cc @beaumorley @parvathyvr

frontend/public/components/monitoring/query-browser.tsx
```diff
@@ -673,11 +676,12 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
     setXDomain([from, to]);
     setSpan(to - from);
   };
+  const isGraphDataEmpty = _.isEmpty(_.flattenDeep(graphData));
```
kyoto, line 679: I think we should avoid using `_.flattenDeep` because there can be a large amount of data in `graphData` and we already have performance concerns.

kyoto: Maybe
```
const isGraphDataEmpty = _.isEmpty(graphData) || _.every(graphData, (d) => d.length === 0);
```
or
```
const isGraphDataEmpty = !_.some(graphData, (d) => d.length > 0);
```

frontend/public/components/monitoring/query-browser.tsx
```diff
@@ -501,6 +501,9 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
   tickInterval,
   timespan,
 }) => {
+  // Min height for GraphEmpty(empty state)
+  const EMPTY_GRAPH_MIN_HEIGHT = 310;
```
kyoto, line 505: Do we need this? I think the height should still be reasonable without it. We already have a state where an error message is displayed without a graph and I think the height is OK in that case without forcing a `height`.

Is there a page that looks bad without `EMPTY_GRAPH_MIN_HEIGHT`?

Also, I notice that `310` is too much for some places, like the Admin perspective > Monitoring > Metrics page.

Ideally, we wouldn't use `EMPTY_GRAPH_MIN_HEIGHT` and just fallback to CSS to give a reasonable height.

kyoto (changes requested):

frontend/public/components/monitoring/query-browser.tsx
nemesis09: `QueryBrowser` already has a CSS class `graph-empty-state` which sets min height `310px` for empty state
but the `GraphEmpty` defaults to `180px` which made it seem floating towards the top.

**Topology Sidebar**
Screenshot from 2020-08-28 17-37-48

**Dev perspective -> Monitoring**
Screenshot from 2020-08-28 17-38-21

frontend/public/components/monitoring/query-browser.tsx
nemesis09: we could use `const isGraphDataEmpty = !graphData || graphData.every((d) => d.length === 0)`
since `every()` always returns true for any condition when array is empty, we could drop the check `_.isEmpty`
updated the checks. please take a look.

kyoto: 👍 Thanks

frontend/public/components/monitoring/query-browser.tsx
kyoto: One problem with this is that depending on the page, the `graph-empty-state` wrapper may contain other content in addition to the empty state message (like the graph controls on the Metrics page), so the `310px` actually makes the `graph-empty-state` taller than 310 pixels.

Not sure if there is an easy CSS fix here.

vikram-raj: If we don't want to set the height then we can do it by using `display: table` CSS property. WDYT @kyoto?

vikram-raj: ^ @christianvogt What will be the better way to fix this?

kyoto: The `display: table` table approach feels like a bit of a hack to me, but not sure what a cleaner solution would be? 🤔

christianvogt: Apply `display: flex` to `graph-empty-state` ?

vikram-raj: Yes, it works. This was the only thing I didn't try. Thanks @christianvogt

nemesis09: tried it out. The empty state message is aligned correctly. updated the code
Thanks @vikram-raj @christianvogt

kyoto: Unfortunately this introduces a bug where the graph controls are now vertically centered while the graph is loading.

screenshot

christianvogt: Looks like the styling needs to be applied more conditionally since it's only necessary when the empty state is shown

nemesis09: added conditional styling
Screenshot from 2020-09-15 02-17-44

kyoto: Unfortunately, I think this is introducing a different bug on the Metrics page. When the graph is loading, the graph wrapper loses its height, which I think is not what we want since it causes everything below the graph to jump up / down as the graph is loading.

screenshot

nemesis09: I've pushed a fix addressing the css issue. Please take a look

kyoto:
I tested and didn't see any layout issues with this now.

Thanks @nemesis09!

/lgtm

PR #6464, opened 2020-08-27
Add option for Deep Compare hook to use JSON.stringify.

rawagner:
Use the strinfigy version in k8s watch hooks.

Just an idea that we could reuse existing `useDeepCompareMemoize` hook, but instead of deep equal we could do JSON.strinfigy.  K8sWatch hooks would use JSON.stringify version of `useDeepCompareMemoize`. 

Results in cleaner code (IMO) and we also dont have to turn the eslint rule off, which is quite dangerous.

I've also improved Model error detection which was incorrect if caller changed watched model during rerenders (due to useEffect async nature)

@spadgett WDYT ?

### DIFF
--- a/frontend/packages/console-shared/src/hooks/deep-compare-memoize.ts
+++ b/frontend/packages/console-shared/src/hooks/deep-compare-memoize.ts
@@ -1,10 +1,14 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 
-export const useDeepCompareMemoize = <T = any>(value: T): T => {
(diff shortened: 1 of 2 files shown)


jeff-phillips-18:
This fixes an issue seen during resources load. Instance where resources were flagged as `loaded: true` with `loadError: new NoModelError()` while `modelsLoaded` is false. This results in a flicker of an error message during this cycle.

@spadgett Can you take a look? Otherwise I need to add the work around mentioned at https://github.com/openshift/console/pull/7034#pullrequestreview-522921832

spadgett (approved):
/lgtm

👍

PR #6468, opened 2020-08-27
Bug 1873303: fix integration test flake where launcher card isn't yet…

rhamilto:
… visible

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1873303

### DIFF
--- a/frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
+++ b/frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
@@ -35,7 +35,7 @@ const launcherLink = {
     namespaceDashboard: {
       namespaces: [testName],
     },
-    text: 'Help Menu Link',
+    text: 'Namespace Dashboard Link',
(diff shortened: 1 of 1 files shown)


frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
```diff
@@ -180,6 +180,7 @@ describe('Project Dashboard', () => {
       // refresh browser to reload launcher links
       await browser.refresh();
       await dashboardView.isLoaded();
+      await browser.wait(until.visibilityOf(projectDashboardView.launcherCard));
       expect(projectDashboardView.launcherCard.isDisplayed()).toBe(true);
```
spadgett, line 184: This line can be removed since the line above will fail if the card isn't displayed.

rhamilto: Removed.

PR #6469, opened 2020-08-28
Bug 1873420: SinUri shows when other resources are filtered

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4693

**Analysis / Root cause**: 
URI shows in topology when selecting any resource from filter by resource

**Solution Description**: 
Made changes to utils to show/hide only filtered resources

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


invincibleJai:
Closing this as would be fixed as part of https://issues.redhat.com/browse/ODC-4652 
cc @jeff-phillips-18

PR #6472, opened 2020-08-28
Bug 1873481: Show workspace volume-type item fields only when added

divyanshiGupta:
Jira - https://issues.redhat.com/browse/ODC-3571

This PR adds a fix to show workspace volume-type item fields only when added and also some UX enhancement changes to item fields.

Screenshots:
Screenshot from 2020-08-28 18-31-05
Screenshot from 2020-08-28 18-32-32

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleKeySelector.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleKeySelector.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
 import { FieldArray, useFormikContext, FormikValues } from 'formik';
 import * as _ from 'lodash';
-import cx from 'classnames';
 import { PlusCircleIcon, MinusCircleIcon } from '@patternfly/react-icons';
 import { DropdownField, InputField, getFieldId, useFormikValidationFix } from '@console/shared';
-import { TextInputTypes, Button, FormGroup, Tooltip } from '@patternfly/react-core';
+import { TextInputTypes, Button, FormGroup, Tooltip, Flex, FlexItem } from '@patternfly/react-core';
 import './MultipleKeySelector.scss';
 
 interface MultipleKeySelectorProps {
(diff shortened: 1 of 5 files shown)


divyanshiGupta:
cc: @bgliwa01

frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleKeySelector.tsx
```diff
@@ -39,23 +38,24 @@ const MultipleKeySelector: React.FC<MultipleKeySelectorProps> = ({
...
-                      type={TextInputTypes.text}
-                      placeholder="Enter a path"
-                      isDisabled={!item.key}
-                    />
-                    <div
-                      className={cx('odc-multiple-key-selector__deleteButton', {
-                        '--disabled': items.length <= 1,
-                      })}
-                    >
+                    <Grid hasGutter>
```
andrewballantyne, line 41: Maybe go with a Flex? [Responsive design](https://www.patternfly.org/v4/documentation/react/layouts/flex#responsive-layout-modifiers) for mobile would help a bit more.

image

frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleResourceKeySelector.tsx
```diff
@@ -93,7 +95,9 @@ const MultipleResourceKeySelector: React.FC<StateProps & MultipleResourceKeySele
         }}
         showBadge
       />
-      {field.value && <MultipleKeySelector name={resourceKeysField} keys={keys} />}
+      {field.value && (
```
andrewballantyne, line 98: Interestingly found a bug while testing this - mind fixing it? It's should be a tiny one.

On [line 64](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleResourceKeySelector.tsx#L64) `keyMap` should have a default of `{}` to prevent an NPE on the `Object.keys` call.

I think we also would need to avoid rendering this block:
```suggestion
      {field.value && Object.keys(keys).length > 0 && (
```

Maybe you can think of another solution, this was my quick hack. Feel free to implement something else if you see a better fix for the NPE.

frontend/packages/dev-console/src/components/pipelines/modals/common/MultipleKeySelector.tsx
divyanshiGupta: image
Done

rottencandy (approved):
Verified locally.
/lgtm

PR #6473, opened 2020-08-28
Bug 1873582: Add access review and feature flags to quick starts

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-4435

**Analysis / Root cause**: Quick starts had no way of checking users permissions or feature flags.

**Solution Description**: Add support to check for access review and feature flags for filtering quick starts.

@openshift/team-devconsole-ux @beaumorley @serenamarie125 @bmignano 

No Operators Installed - 
Peek 2020-08-28 21-13

Pipelines Operator installed - 
Peek 2020-08-28 22-01

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/console-app/src/components/quick-starts/data/quick-start-data.ts
```diff
@@ -2,10 +2,11 @@ import { exploreServerlessQuickStart } from './explore-serverless-quickstart';
...
 import { serverlessApplicationQuickStart } from './serverless-application-quickstart';
+import { QuickStart } from '../utils/quick-start-types';
 
 export const allQuickStarts = [
   exploreServerlessQuickStart,
   serverlessApplicationQuickStart,
   explorePipelinesQuickStart,
   installAssociatePipelineQuickStart,
-];
+] as QuickStart[];
```
jerolimov, line 12: Please add the type information to the variable instead of 'casting' your array. This will ensure that your types really matches. Try this by adding for a example a string to your array.

```diff
-export const allQuickStarts = [
+export const allQuickStarts: QuickStart[] = [
  exploreServerlessQuickStart,
  serverlessApplicationQuickStart,
  explorePipelinesQuickStart,
  installAssociatePipelineQuickStart,
-] as QuickStart[];
+];
```

This will require that you add `: Quickstart` also to the quick starts json definitions itself, but only then we use the types here to validate the data.

frontend/packages/console-app/src/components/quick-starts/data/serverless-application-quickstart.ts
```diff
@@ -1,3 +1,4 @@
+import { FLAG_KNATIVE_SERVING, FLAG_KNATIVE_EVENTING } from '@console/knative-plugin';
 import { serverlessIcon } from './tour-icons';
 
 export const serverlessApplicationQuickStart = {
```
jerolimov, line 4: Add `: QuickStart` here as to ensure that the format and type definition matches.

```suggestion
export const serverlessApplicationQuickStart: QuickStart = {
```

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStarts.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const filteredQuickStarts = quickStarts.filter((quickStart) => {
+    const {
+      spec: { accessReviewResources, flags },
+    } = quickStart;
+
+    const hasAccess =
+      !accessReviewResources ||
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      accessReviewResources.map((descriptor) => useAccessReview(descriptor)).every((x) => x);
```
jerolimov, line 17: `useAccessReview` uses `useAccessReview2` which use `useSafetyFirst` which use `useState` internally. Disabling this eslint rule here works only because the quickStarts are statically saved and the order does never change until we change this.

Can you please add an comment here why you disable the eslint rules in this case? WDYT? Then its fine for me.

jerolimov (commented):
Tested it locally and works fine. Requested two changes or feedback for my ideas. :)

jerolimov (changes requested):
Mark as "Request changes" :)

frontend/packages/console-app/src/components/quick-starts/data/quick-start-data.ts
rohitkrai03: Done

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStarts.ts
rohitkrai03: Done.

frontend/packages/console-app/src/components/quick-starts/data/serverless-application-quickstart.ts
rohitkrai03: Done

jerolimov (commented):
/lgtm

rohitkrai03:
/hold

Will rework the PR to remove using hooks in loop so that the transition to CRD based quick starts would go smoother.

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsLoader.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <QuickStartPermissionChecker
+            key={qs.metadata.name}
+            quickStart={qs}
+            onPermissionCheck={(hasPermission) => {
+              if (hasPermission && !checkedQuickStarts.includes(qs)) {
+                setCheckedQuickStarts([...checkedQuickStarts, qs]);
+              }
+              permissionChecks.current[qs.metadata.name] = hasPermission;
+            }}
+          />
```
christianvogt, line 33: Considering that `QuickStartPermissionChecker` has an effect whose dependency is the `onPermissionCheck` function. This should use `useCallback`.

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStartPermission.ts
```diff
@@ -0,0 +1,29 @@
...
+    !flags?.required ||
+    // flags for a specific quick start is going to be static so the order will not change
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    flags?.required?.map((flag) => useFlag(flag)).every((x) => x);
+
+  // flags for a specific quick start is going to be static so the order will not change
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const disallowedFlags = flags?.disallowed?.map((flag) => useFlag(flag)).some((x) => x);
+
+  return hasAccess && requiredFlags && !disallowedFlags;
```
christianvogt, line 26: Shouldn't you distinguish between access review loading and loaded?
Otherwise you may be replying that access is not yet available while the review is still pending.
`useAccessReview2` returns a tuple with a loading indicator.

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsLoader.tsx
rohitkrai03: Done

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStartPermission.ts
rohitkrai03: Done.

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsLoader.tsx
```diff
@@ -0,0 +1,50 @@
...
+import QuickStartPermissionChecker from './QuickStartPermissionChecker';
+
+type QuickStartsLoaderProps = {
+  children: (quickStarts: QuickStart[], loaded: boolean) => React.ReactNode;
+};
+
+const QuickStartsLoader: React.FC<QuickStartsLoaderProps> = ({ children }) => {
+  const [quickStarts, setQuickStarts] = React.useState<QuickStart[]>([]);
+  const [allowedQuickStarts, setAllowedQuickStarts] = React.useState<QuickStart[]>([]);
+  const [loaded, setLoaded] = React.useState<boolean>(false);
```
christianvogt, line 13: You don't need both of these.
If `allowedQuickStarts` is set, then show, otherwise consider it loading.

Or derive the allowed quick starts in render.

rohitkrai03: Done.

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsLoader.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { getQuickStarts } from '../utils/quick-start-utils';
+import QuickStartPermissionChecker from './QuickStartPermissionChecker';
+
+type QuickStartsLoaderProps = {
+  children: (quickStarts: QuickStart[], loaded: boolean) => React.ReactNode;
+};
+
+const QuickStartsLoader: React.FC<QuickStartsLoaderProps> = ({ children }) => {
+  const [quickStarts, setQuickStarts] = React.useState<QuickStart[]>([]);
+  const [allowedQuickStarts, setAllowedQuickStarts] = React.useState<QuickStart[]>([]);
```
christianvogt, line 12: When an array is set, this means that the quick start permissions have been checked. When unset, it means we're still loading. It is possible that in the future we don't have any quick starts if all CRs are removed.
```suggestion
  const [allowedQuickStarts, setAllowedQuickStarts] = React.useState<QuickStart[]>(quickStarts.length ? null : []);
```

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStartPermission.ts
```diff
@@ -0,0 +1,68 @@
...
+  const hasDisallowedFlags = disallowedFlags.some((x) => x);
+
+  const hasPermission = hasAccess && hasRequiredFlags && !hasDisallowedFlags;
+
+  const loaded =
+    accessReviewsLoading.every((loading) => !loading) &&
+    requiredFlagsPending.every((pending) => !pending) &&
+    disallowedFlagsPending.every((pending) => !pending);
+
+  return [hasPermission, loaded];
```
christianvogt, line 65: Perhaps be cautious and only set `hasPermission` if loaded.
```suggestion
  return [loaded && hasPermission, loaded];
```

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsLoader.tsx
rohitkrai03: This would still return loaded  as false all the time even if there are no CRs to begin with. I think my earlier implementation of having a separate state for loaded would be better to handle both use cases. Updated to have a separate loaded state.

frontend/packages/console-app/src/components/quick-starts/utils/useQuickStartPermission.ts
rohitkrai03: done

PR #6474, opened 2020-08-28
Bug 1866739: Filter incompatible x-descriptors before rendering operand form

TheRealJon:
When parsing descriptors, check to make sure that x-descriptors are compatible with
the schema data type they are being applied to. Incompatible x-descriptors are filtered out before
using them to build a ui schema. An x-descriptor is considered incompatible with a schema property if:

  - x-descriptor is deprecated
  - x-descriptor would render a primitive type input and the schema property is not a primitive type.
  - x-descriptor would render an object or array type input and the schema property is not matching data type.

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operand/utils.ts
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operand/utils.ts
@@ -2,6 +2,7 @@ import * as _ from 'lodash';
 import * as Immutable from 'immutable';
 import { JSONSchema6 } from 'json-schema';
 import { UiSchema } from 'react-jsonschema-form';
+import { getSchemaType } from 'react-jsonschema-form/lib/utils';
 import { modelFor } from '@console/internal/module/k8s';
 import { getJSONSchemaOrder } from '@console/shared/src/components/dynamic-form/utils';
 import { SpecCapability, Descriptor } from '../descriptors/types';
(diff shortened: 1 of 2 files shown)


frontend/packages/operator-lifecycle-manager/src/components/operand/utils.ts
```diff
@@ -103,13 +111,6 @@ export const capabilitiesToUISchema = (capabilities: SpecCapability[] = []) => {
     return k8sResourceCapabilityToUISchema(k8sResourceCapability);
   }
 
-  const fieldDependencyCapability = _.find(capabilities, (capability) =>
-    capability.startsWith(SpecCapability.fieldDependency),
-  );
-  if (fieldDependencyCapability) {
-    return fieldDependencyCapabilityToUISchema(fieldDependencyCapability);
-  }
-
```
TheRealJon, line 112: This was another potential bug. The field dependency should be applied independent of widget or field. The way this was written, a k8sResourcePrefix descriptor would take precedence over fieldDependency, and fieldDependency would take precedence over all other field and widget types. So you would never be able to have custom widget with a dependency.

frontend/packages/operator-lifecycle-manager/src/components/operand/utils.ts
```diff
@@ -139,27 +140,75 @@ export const capabilitiesToUISchema = (capabilities: SpecCapability[] = []) => {
   };
 };
 
+const getValidCapabilities = (descriptor: Descriptor<SpecCapability>, jsonSchema: JSONSchema6) => {
+  const schemaType = getSchemaType(jsonSchema ?? {}) as JSONSchemaType;
```
spadgett, line 144: Does this work? It's preferable to `as` when possible since it has stronger type checking. I realize we probably don't have typings for react-jsonschema-form.

```suggestion
  const schemaType: JSONSchemaType = getSchemaType(jsonSchema ?? {});
```

spadgett (approved):
/lgtm

PR #6475, opened 2020-08-28
Bug 1880118: Improve Cypress `checkErrors` output

dtaylor113:
This PR updates our usage of `window.windowError` in `app.js` and also improves Cypress `checkErrors()` output:
```
2) Monitoring: Alerts
       "after each" hook for "creates and expires a Silence":
     window error detected: {
 "message": "y is not defined",
 "stack": "ReferenceError: y is not defined
    at http://localhost:9000/static/main-e0abc76e4ffb6fa60c4c.js:95489:9
    at renderWithHooks (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:371358:18)
    at mountIndeterminateComponent (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:373592:13)
    at beginWork$1 (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:374736:16)
    at HTMLUnknownElement.callCallback (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:356597:14)
    at Object.invokeGuardedCallbackDev (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:356647:16)
    at invokeGuardedCallback (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:356704:31)
    at beginWork$$1 (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:379467:7)
    at performUnitOfWork (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:378458:12)
    at workLoopSync (http://localhost:9000/static/vendors~main-2e679479202da74be6e9.js:378435:22)"
```

### DIFF
--- a/frontend/packages/integration-tests-cypress/support/index.ts
+++ b/frontend/packages/integration-tests-cypress/support/index.ts
@@ -3,6 +3,7 @@ import './project';
 import './selectors';
 import './nav';
 import './resources';
+
 import { a11yTestResults } from './a11y';
 
 Cypress.Cookies.defaults({
@@ -38,11 +39,22 @@ after(() => {
   });
 });
 
+const formatWindowError = (windowError: Error | PromiseRejectionEvent) => {
+  if (!windowError) {
+    return 'no window error detected';
+  }
+  const { reason } = windowError as PromiseRejectionEvent;
+  if (reason) {
+    return reason;
+  }
+  const { message, stack } = windowError as Error;
+  const formattedStack = stack?.replace(/\\n/g, '\n');
+  return `window error detected: ${message} ${formattedStack}`;
+};
+
 export const checkErrors = () =>
   cy.window().then((win) => {
-    if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
-    }
+    assert.isTrue(!win.windowError, formatWindowError(win.windowError));
   });
 
 export const testName = `test-${Math.random()
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -214,11 +214,16 @@ setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000)
 fetchSwagger();
 
 // Used by GUI tests to check for unhandled exceptions
-window.windowError = false;
-window.onerror = window.onunhandledrejection = (e) => {
+window.windowError = null;
+window.onerror = (message, source, lineno, colno, error) => {
   // eslint-disable-next-line no-console
-  console.error('Uncaught error', e);
-  window.windowError = e || true;
+  console.error('Uncaught error', error);
+  window.windowError = error;
+};
+window.onunhandledrejection = (promiseRejectionEvent) => {
+  // eslint-disable-next-line no-console
+  console.error('Unhandled promise rejection', promiseRejectionEvent);
+  window.windowError = promiseRejectionEvent;
 };
 
 if ('serviceWorker' in navigator) {
--- a/frontend/public/declarations.d.ts
+++ b/frontend/public/declarations.d.ts
@@ -41,7 +41,7 @@ declare interface Window {
     GOOS: string;
     graphqlBaseURL: string;
   };
-  windowError?: boolean | string;
+  windowError?: Error | PromiseRejectionEvent;
   __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: Function;
   store?: {}; // Redux store
   pluginStore?: {}; // Console plugin store
--- a/frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
@@ -46,6 +46,6 @@ describe('Namespace', () => {
     cy.testA11y('Delete Namespace modal');
     modal.submit();
     modal.shouldBeClosed();
-    listPage.rows.shouldNotExist(newName);
+    cy.resourceShouldBeDeleted(testName, 'namespaces', newName);
   });
 });


frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -59,7 +59,9 @@ Cypress.Commands.add('testA11y', (target: string) => {
 export const checkErrors = () =>
   cy.window().then((win) => {
     if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
+      console.error(
+        `window/js runtime error detected: ${JSON.stringify(win.windowError, null, '')}`,
```
spadgett, line 63: `win.windowError.message` ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -59,7 +59,9 @@ Cypress.Commands.add('testA11y', (target: string) => {
 export const checkErrors = () =>
   cy.window().then((win) => {
     if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
```
spadgett, line 62: Will the test still fail if we remove the `throw`? Should it just be `throw win.windowError`?

dtaylor113: not sure, we can see what it looks like in CI

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -59,7 +59,7 @@ Cypress.Commands.add('testA11y', (target: string) => {
...
+      throw win.windowError;
```
spadgett, line 62: We should fix the types and only assign an `Error` object. Right now it's `boolean | string`, which is wrong. We should initialize to `null` instead of `false`.

dtaylor113: ok, fixed

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
```diff
@@ -87,16 +87,17 @@ describe('Kubernetes resource CRUD operations', () => {
...
           newContent = _.defaultsDeep(
             {},
             { metadata: { name, labels: { [testLabel]: testName } } },
             safeLoad(content),
           );
-          yamlEditor.setEditorContent(safeDump(newContent, { sortKeys: true }));
           cy.log('creates a new resource instance');
-          yamlEditor.clickSaveCreateButton();
-          cy.get(errorMessage).should('not.exist');
+          yamlEditor.setEditorContent(safeDump(newContent, { sortKeys: true })).then(() => {
```
spadgett, line 97: Is the `then` necessary here? Does Cypress not block automatically?

frontend/public/components/app.jsx
```diff
@@ -213,11 +213,11 @@ setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000)
 fetchSwagger();
 
 // Used by GUI tests to check for unhandled exceptions
-window.windowError = false;
+window.windowError = null;
 window.onerror = window.onunhandledrejection = (e) => {
   // eslint-disable-next-line no-console
   console.error('Uncaught error', e);
-  window.windowError = e || true;
+  window.windowError = e || null;
```
spadgett, line 220: No reason to use `|| null`.

```suggestion
  window.windowError = e;
```

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
dtaylor113: Cypress is asynchronous and they show examples of nested `.then()`'s

frontend/public/components/app.jsx
dtaylor113: ok, fixed

frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts
spadgett: I'm still not sure it's needed. Are you looking at https://docs.cypress.io/api/commands/window.html#Check-a-custom-property ?

That's reading a variable set inside the `then` later in the test, which doesn't apply here.

dtaylor113: We've seen flakes around saving from yamlEditor, this way ensures the new content is set...then, `yamlEditor.clickSaveCreateButton();`.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: Set `windowError` type to `PromiseRejectionEvent`.

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -59,7 +59,7 @@ Cypress.Commands.add('testA11y', (target: string) => {
 export const checkErrors = () =>
   cy.window().then((win) => {
     if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
+      throw new Error(`window error - unhandled promise rejection: ${win.windowError.reason}`);
```
spadgett, line 62: We lose the stack trace if you do this. It would be better. We need to print that out somehow or we won't be able to debug these errors.

frontend/public/declarations.d.ts
```diff
@@ -41,7 +41,7 @@ declare interface Window {
     GOOS: string;
     graphqlBaseURL: string;
   };
-  windowError?: boolean | string;
+  windowError?: PromiseRejectionEvent;
```
spadgett, line 44: This isn't always a promise rejection. It can be any error.

dtaylor113: Isn't [this](https://github.com/openshift/console/blob/b8325c2f9032b1e822e1f5f63fb5506394b4e515/frontend/public/components/app.jsx#L217) just trapping `onunhandledrejection`?

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: If `win.windowError` is always a `PromiseRejectionEvent` then it doesn't have a `e.stack` rather it has a `PromiseRejectionEvent.reason`.  If I throw a new Error, it's stack with be to the 'checkErrors' method, not the original error.

frontend/public/declarations.d.ts
spadgett: No, it's also `window.onerror`

frontend/packages/integration-tests-cypress/support/index.ts
spadgett: The code you pasted shows it's not always `PromiseRejectionEvent` :) It can be `window.onerror`, too.

frontend/public/declarations.d.ts
dtaylor113: Ok, so `window.onerror = (e Event | string) => ...`, either way we're trapping an Event | string and setting `windowError` to it.  It's not an `Error` with a stack trace.

spadgett: Ah, we should really be handling the `error`, which is passed, just not as the first arg.

https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

We probably want to switch to `window.addEventListener('error')`

dtaylor113: k, I was considering that as well, thanks

dtaylor113: It would have to be something like:
```
window.addEventListener('error', e:ErrorEvent => {
    // Get the error properties from the error event object
    const { message, source, lineno, colno, error } = e; 
   // construct new Error from ^^ and assign to win.windowError ?
   //  or set win.windowError to e (ErrorEvent)?
});
```

spadgett: How about?

```ts
window.addEventListener('error', ({error}) => window.windowError = error);
```

The key thing is that we retain the original stack trace and any error details for debugging this failure in CI.

dtaylor113: When I use `window.onerror = (e : Event | string) => {` or `window.addEventListener('error', (e: ErrorEvent) => {` in app.js it doesn't trap the `PromiseRejectionEvent` thrown -that is only caught if I use `window.onerror = window.onunhandledrejection = (e) => {`
If I use `window.onerror = window.onunhandledrejection = (e) => {` then the `PromiseRejectionEvent` is caught in app.js; but it is an Event not an Error.

I'm surprised the `window.onerror` or ` `window.addEventListener('error',` don't catch the Error thrown by co-fetch.js's `validateStatus` which processes the fetch.response and generates any appropriate [Error](https://github.com/openshift/console/blob/cf7744e55f10bd7f05eafd57d8485254862faba8/frontend/public/co-fetch.js#L73).  If I could catch the Error thrown, then I could use it's `error.response`.

dtaylor113: Seems like something is dispatching the `PromiseRejectionEvent` when url is not found (my way of triggering it).  I thought throwing an Error would be the same as dispatching an ErrorEvent, but it doesn't seem so.

spadgett: You still need a `window.onunhandledrejection` handler.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: resolved, pls see latest code

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: resolved, pls see latest code

frontend/public/declarations.d.ts
dtaylor113: resolved, pls see latest code

dtaylor113:
Hi @spadgett, PR paired down to just update `window.windowError`

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -59,7 +59,13 @@ Cypress.Commands.add('testA11y', (target: string) => {
 export const checkErrors = () =>
   cy.window().then((win) => {
     if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
+      throw new Error(
```
spadgett, line 62: Does `throw win.windowError` work instead of creating a new `Error`? Or adding an assertion that `windowError` is nullish?

frontend/public/components/app.jsx
```diff
@@ -214,11 +214,17 @@ setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000)
...
+window.windowError = null;
+window.onerror = (message, source, lineno, colno, error) => {
   // eslint-disable-next-line no-console
-  console.error('Uncaught error', e);
-  window.windowError = e || true;
+  console.error('Uncaught error', error);
+  window.windowError = error;
+};
+window.onunhandledrejection = (promiseRejectionEvent) => {
+  const error = new Error(`unhandled promise rejection: ${promiseRejectionEvent.reason}`);
```
spadgett, line 224: If this isn't already an error, maybe we just leave `window.windowError` as `Error | Event` and just check that it's not defined in Cypress rather than trying to rethrow it.

dtaylor113: I don't think we'd see any output for `promiseRejectionEvent` in Cypress if we did that.

dtaylor113: Note: previously the `window.windowError` was either an `ErrorEvent` or `promiseRejectionEvent`, it never was an `Error`.  I had to change the method signiture params in order to get the `Error`: `window.onerror = (message, source, lineno, colno, error)`.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: Hmm...I had tried it in the past but not since I truley changed `win.windowError` to be an `Error` not an `Event`.  I'll try to throw it again not that it's a true `Error`

frontend/public/components/app.jsx
dtaylor113: I do know that when `window.windowError` was a `promiseRejectionEvent`, throwing it didn't produce any output in headless.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: `throw win.windowError` results in the following; we lose stack-trace, but since it's minified, not sure how useful it is:
```
ReferenceError: i is not defined
```
As opposed to throwing new Error w/ JSON.stringify & `\n` processing:
```
     window error detected: {
 "message": "i is not defined",
 "stack": "ReferenceError: i is not defined
    at http://localhost:9000/static/main-35bf7a55e8c0461351e2.js:102081:5
    at renderWithHooks (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:371358:18)
    at mountIndeterminateComponent (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:373592:13)
    at beginWork$1 (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:374736:16)
    at HTMLUnknownElement.callCallback (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:356597:14)
    at Object.invokeGuardedCallbackDev (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:356647:16)
    at invokeGuardedCallback (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:356704:31)
    at beginWork$$1 (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:379467:7)
    at performUnitOfWork (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:378458:12)
    at workLoopSync (http://localhost:9000/static/vendors~main-47b21d89223a2a918b47.js:378435:22)"
}
```

frontend/public/components/app.jsx
spadgett: Right, you wouldn't throw the event. You would check that it's not null and log the content (if not already logged by Cypress).

dtaylor113: idk, I feel casting all into an `Error` and just handling/throwing the `Error` (directly or recreating w/ formatting), to me seems more straight forward.   Unless you are saying we should log the `promiseRejectionEvent`, not throw an Error, so that the cypress tests won't fail on a `promiseRejectionEvent`?

spadgett: I'm saying we have an assertion that causes the test to fail if `windowError` is never null. Something like

```ts
expect(win.windowError).to.not.exist;
```

dtaylor113: Hi, tried
```
assert.isNull(
      win.windowError,
      `${(win.windowError as PromiseRejectionEvent).reason || (win.windowError as Error).message} ${
        (win.windowError as Error).stack
      }`
```
Which, in addition to the message (2nd arg in assert(..)), also output the tested object.  Unfortunately, since `window` is part of the PromiseRejectionEvent, it gets spit out along with a lot of other attributes:
```
2) Monitoring: Alerts
       "after each" hook for "creates and expires a Silence":
     path /silence_WRONG_URL/63646b85-b4a6-4ead-8a35-609a77317edf was not found undefined: expected { isTrusted: [Getter],
  promise: {},
  reason: 'path /silence_WRONG_URL/63646b85-b4a6-4ead-8a35-609a77317edf was not found',
  NONE: 0,
  CAPTURING_PHASE: 1,
  AT_TARGET: 2,
  BUBBLING_PHASE: 3,
  type: 'unhandledrejection',
  target: 
   { parent: 
      { parent: [Circular],
        opener: null,
        top: [Circular],
        length: 2,
        frames: [Circular],
        closed: false,
        location: [Object],
        self: [Circular],
        window: [Circular],
        document: [Getter],
        name: [Getter/Setter],
        customElements: [Getter],
        history: [Getter],
        locationbar: [Getter/Setter],
        menubar: [Getter/Setter],
        personalbar: [Getter/Setter],
        scrollbars: [Getter/Setter],
        statusbar: [Getter/Setter],
        toolbar: [Getter/Setter],
        status: [Getter/Setter],
        frameElement: [Getter],
        navigator: [Getter],
        origin: [Getter/Setter],
        external: [Getter/Setter],
        screen: [Getter/Setter],
        innerWidth: [Getter/Setter],
        innerHeight: [Getter/Setter],
        scrollX: [Getter/Setter],
        pageXOffset: [Getter/Setter],
        scrollY: [Getter/Setter],
        pageYOffset: [Getter/Setter],
        visualViewport: [Getter/Setter],
        screenX: [Getter/Setter],
        screenY: [Getter/Setter],
        outerWidth: [Getter/Setter],
        outerHeight: [Getter/Setter],
        devicePixelRatio: [Getter/Setter],
        clientInformation: [Getter/Setter],
...
...
  -- additional two screens worth of output ---
```

dtaylor113: Latest code I think moves closer to what your after. Changed the assertion so it doesn't ouput entire contents of `win.windowError`.  It does output `...: expected false to be true` which doesn't add any value:

```
assert.isTrue(
      win.windowError === null,
      `window error detected: ${(win.windowError as PromiseRejectionEvent).reason ||
        (win.windowError as Error).message} ${
        (win.windowError as Error).stack
          ? (win.windowError as Error).stack.replace(/\\n/g, '\n')
          : ''
      }`,
    );
```
**Produces:**
```
...when windowError is a PromiseRejectionEvent:

 2) Monitoring: Alerts
       "after each" hook for "creates and expires a Silence":
     AssertionError: window error detected: path /silence_WRONG_URL/b0ef08c1-e09c-41ab-a046-761d47a71bee was not found : expected false to be true

...when windowError is an Error:

 2) Monitoring: Alerts
       "after each" hook for "creates and expires a Silence":
     window error detected: i is not defined ReferenceError: i is not defined
    at http://localhost:9000/static/main-66d801b01502734ad2c0.js:96412:5
    at renderWithHooks (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:377121:18)
    at mountIndeterminateComponent (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:379355:13)
    at beginWork$1 (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:380499:16)
    at HTMLUnknownElement.callCallback (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:362360:14)
    at Object.invokeGuardedCallbackDev (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:362410:16)
    at invokeGuardedCallback (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:362467:31)
    at beginWork$$1 (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:385230:7)
    at performUnitOfWork (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:384221:12)
    at workLoopSync (http://localhost:9000/static/vendors~main-9d6cea73c651a7e7c901.js:384198:22): expected false to be true
```

The only way to generate output in headless in `afterEach` is either throwing an Error or failing an assertion.   `cy.task('log',' ...)` which uses `console.log` doesn't output anything :-(

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -40,9 +41,17 @@ after(() => {
...
+      win.windowError !== null
+        ? `window error detected: ${(win.windowError as PromiseRejectionEvent).reason ||
+            (win.windowError as Error).message} ${
+            (win.windowError as Error).stack
+              ? (win.windowError as Error).stack.replace(/\\n/g, '\n')
+              : ''
+          }`
+        : 'no window error detected';
+
+    assert.isTrue(win.windowError === null, assertMsg);
```
spadgett, line 54: Don't check specifically for null. You want to handle undefined as well.

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -40,9 +41,17 @@ after(() => {
...
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
-    }
+    const assertMsg =
+      win.windowError !== null
+        ? `window error detected: ${(win.windowError as PromiseRejectionEvent).reason ||
+            (win.windowError as Error).message} ${
+            (win.windowError as Error).stack
+              ? (win.windowError as Error).stack.replace(/\\n/g, '\n')
+              : ''
+          }`
```
spadgett, line 51: This is too much on one line. I'd pull this out into a helper. Something like

```ts
const formatWindowError = (windowError: Error | PromiseRejectionEvent) => {
  const { reason } = windowError as PromiseRejectionEvent;
  if (reason) {
    return reason;
  }
  const { message, stack } = windowError as Error;
  const formattedStack = stack?.replace(/\\n/g, '\n');
  return `${message} ${formattedStack}`;
};
```

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -40,9 +41,17 @@ after(() => {
 
 export const checkErrors = () =>
   cy.window().then((win) => {
-    if (win.windowError) {
-      throw new Error(`window/js runtime error detected: ${win.windowError}`);
-    }
+    const assertMsg =
+      win.windowError !== null
```
spadgett, line 45: Handle `undefined` as well.

```suggestion
      win.windowError
```

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: thanks, fixed.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: fixed.

frontend/packages/integration-tests-cypress/support/index.ts
dtaylor113: fixed.

frontend/public/components/app.jsx
```diff
@@ -214,11 +214,16 @@ setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000)
...
+window.onerror = (message, source, lineno, colno, error) => {
   // eslint-disable-next-line no-console
-  console.error('Uncaught error', e);
-  window.windowError = e || true;
+  console.error('Uncaught error', error);
+  window.windowError = error;
+};
+window.onunhandledrejection = (promiseRejectionEvent) => {
+  // eslint-disable-next-line no-console
+  console.error(' Unhandled promise rejection', promiseRejectionEvent);
```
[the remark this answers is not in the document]
dtaylor113: fixed.

spadgett (approved):
/lgtm

PR #6478, opened 2020-08-31
Bug 1843387: update queryInput on change to pre-fill Query

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4096

**Description**:
Goto Dev Console, click "Monitoring" and select any project, then click the "Metrics" tab, click "Show PromQL" button, select any defined metrics from the dropdown list, example: CPU Usage
after the graph is shown, click "x" button in the metrics input text area to remove expression, select "CPU Usage" again, the metrics expression is not shown in the metrics input text area

**Analysis / Root cause**: 
Query Input text area not shows query in it's cleared by user and tried to select to same Query again

**Solution Description**: 
On Each change update the QueryInput

Aug-31-2020 12-26-26

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj (commented):
Verified, worked as expected. 
/lgtm

PR #6485, opened 2020-09-01
Bug 1872340: E2E test for restoring a claim from Snapshot

a2batic:
Test for successful restore of claim from a snapshot.

### DIFF
--- a/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
+++ b/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
@@ -139,6 +139,7 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
                 isRequired
                 type="text"
                 id="pvc-name"
+                data-test="pvc-name"
                 name="restore-pvc-modal__name"
                 value={restorePVCName}
                 onChange={setPVCName}
--- a/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
+++ b/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
@@ -1,3 +1,5 @@
+import { Patch } from '@console/internal/module/k8s';
+
 export const testerDeployment = {
   apiVersion: 'apps/v1',
   kind: 'Deployment',
@@ -79,4 +81,15 @@ export const SnapshotClass = {
   },
   driver: 'ebs.csi.aws.com',
   deletionPolicy: 'Delete',
+};
+
+export const patchForVolume: Patch = {
+  op: 'add',
+  path: '/spec/template/spec/volumes/-',
+  value: {
+    name: 'testpvc-snapshot-restore',
+    persistentVolumeClaim: {
+      claimName: 'testpvc-snapshot-restore',
+    },
+  },
 };
--- a/frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
@@ -1,7 +1,7 @@
 import { testName, checkErrors } from '../../support';
 import { SnapshotDetails, dropdownFirstItem } from '../../views/storage/snapshot';
 import { listPage } from '../../views/list-page';
-import { PVC, testerDeployment, SnapshotClass } from '../../mocks/snapshot';
+import { PVC, testerDeployment, SnapshotClass, patchForVolume } from '../../mocks/snapshot';
 import { detailsPage } from '../../views/details-page';
 import { modal } from '../../views/modal';
 import { resourceStatusShouldContain } from '../../views/common';
@@ -29,6 +29,7 @@ if (Cypress.env('BRIDGE_AWS')) {
     after(() => {
       cy.exec(`echo '${JSON.stringify(testerDeployment)}' | oc delete -n ${testName} -f -`);
       cy.exec(`echo '${JSON.stringify(PVC)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`oc delete pvc ${snapshotName}-restore -n ${testName}`);
       cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc delete -f -`);
       cy.deleteProject(testName);
       cy.logout();
@@ -51,7 +52,7 @@ if (Cypress.env('BRIDGE_AWS')) {
         `snapshot.storage.k8s.io~v1beta1~VolumeSnapshot/${PVC.metadata.name}-snapshot`,
       );
       detailsPage.titleShouldContain(PVC.metadata.name);
-      resourceStatusShouldContain('Ready');
+      resourceStatusShouldContain('Ready', { timeout: 40000 });
       cy.exec(`oc get VolumeSnapshot ${PVC.metadata.name}-snapshot -n ${testName} -o json`)
         .its('stdout')
         .then((res) => {
@@ -75,7 +76,28 @@ if (Cypress.env('BRIDGE_AWS')) {
       listPage.rows.shouldNotExist(`${snapshotName}dup`);
     });
 
+    it('Restore a Snapshot to create a new claim from it', () => {
+      cy.clickNavLink(['Volume Snapshots']);
+      listPage.rows.clickKebabAction(snapshotName, 'Restore as new PVC');
+      modal.shouldBeOpened();
+      cy.byTestID('pvc-name').should('have.value', `${snapshotName}-restore`);
+      cy.get(SnapshotDetails.scDropdown).click();
+      cy.get(dropdownFirstItem)
+        .eq(1)
+        .click();
+      modal.submit();
+      modal.shouldBeClosed();
+      cy.exec(
+        `oc patch Deployment ${
+          testerDeployment.metadata.name
+        } --type='json' -n ${testName} -p '[${JSON.stringify(patchForVolume)}]'`,
+      )
+        .its('stdout')
+        .then(() => resourceStatusShouldContain('Bound', { timeout: 40000 }));
+    });
+
     it('Deletes Snapshot', () => {
+      cy.clickNavLink(['Volume Snapshots']);
       listPage.rows.clickKebabAction(snapshotName, 'Delete Volume Snapshot');
       modal.shouldBeOpened();
       modal.submitShouldBeEnabled();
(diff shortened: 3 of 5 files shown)


frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -139,6 +139,7 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
                 isRequired
                 type="text"
                 id="pvc-name"
+                data-cy="pvc-name"
```
gnehapk, line 142: Why you need to add `data-cy` if `id` is already present? You can use `id` to select the pvc-name

a2batic: I tried to follow best practices mentioned here: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements.

a2batic:
@dtaylor113, please review.

frontend/packages/integration-tests-cypress/tests/storage/restore.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    afterEach(() => {
+      checkErrors();
+    });
+
+    after(() => {
+      cy.exec(`echo '${JSON.stringify(testerDeployment)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`echo '${JSON.stringify(PVC)}' | oc delete -n ${testName} -f -`);
+      cy.exec(`oc delete pvc ${Snapshot.metadata.name}-restore -n ${testName}`)
+        .its('stdout')
+        .then(() => {
```
bipuladh, line 40: If you're the using the `stdout` I don't see the reason to add this block.

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,5 @@
 export const resourceStatusShouldContain = (desiredStatus: string) =>
   cy.contains('[data-test="status-text"]', desiredStatus);
+
+export const waitForResourceStatusToContain = (desiredStatus: string) =>
+  cy.contains('[data-test="status-text"]', desiredStatus, { timeout: 30000 });
```
bipuladh, line 5: The default timeout is 30 seconds so I don't think this is required.

frontend/packages/integration-tests-cypress/tests/storage/restore.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+          cy.deleteProject(testName);
+          cy.exec(`echo '${JSON.stringify(SnapshotClass)}' | oc delete -f -`);
+          cy.logout();
+        });
+    });
+
+    it('Restore a Snapshot to create a new claim from it', () => {
+      listPage.rows.clickKebabAction(Snapshot.metadata.name, 'Restore as new PVC');
+      modal.shouldBeOpened();
+      cy.get(RestoreDetails.restoreName).should('have.value', `${Snapshot.metadata.name}-restore`);
```
bipuladh, line 51:
```suggestion
      cy.byTestId('pvc-name').should('have.value', `${Snapshot.metadata.name}-restore`);
```
No need to use a constant since it's only used in one place.

frontend/packages/integration-tests-cypress/views/storage/snapshot.ts
```diff
@@ -6,4 +6,9 @@ export namespace SnapshotDetails {
   export const sc = '[data-test="details-item-value__SC"] a';
 }
 
+export namespace RestoreDetails {
+  export const restoreName = '[data-cy=pvc-name]';
```
bipuladh, line 10:
```suggestion
```

bipuladh (changes requested):
I think we can move this test inside the snapshot file itself as most of it is setup code. And the restore test is only one `it` block.

frontend/packages/integration-tests-cypress/views/common.ts
a2batic: The default time for contain command (4 seconds - https://docs.cypress.io/api/commands/contains.html#Arguments), is not enough to get the desired status.

bipuladh: https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/cypress.json#L13
We have overridden that in the configuration

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -51,7 +52,7 @@ if (Cypress.env('BRIDGE_AWS')) {
         `snapshot.storage.k8s.io~v1beta1~VolumeSnapshot/${PVC.metadata.name}-snapshot`,
       );
       detailsPage.titleShouldContain(PVC.metadata.name);
-      resourceStatusShouldContain('Ready');
+      waitForResourceStatusToContain('Ready');
```
bipuladh, line 55: https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/cypress.json#L13
We already wait for 30 seconds which is enough.

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,5 @@
 export const resourceStatusShouldContain = (desiredStatus: string) =>
   cy.contains('[data-test="status-text"]', desiredStatus);
+
+export const waitForResourceStatusToContain = (desiredStatus: string) =>
+  cy.contains('[data-test="status-text"]', desiredStatus, { timeout: 40000 });
```
bipuladh, line 5:
```suggestion
```
resourceStatusShouldContain is enough.

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -75,7 +76,31 @@ if (Cypress.env('BRIDGE_AWS')) {
...
+      modal.shouldBeClosed();
+      cy.exec(
+        `oc patch Deployment ${
+          testerDeployment.metadata.name
+        } --type='json' -n ${testName} -p '[${JSON.stringify(PatchForVolume)}]'`,
+      )
+        .its('stdout')
+        .then((res) => {
+          expect(res).toBe(`deployment.apps/${testerDeployment.metadata.name} patched`);
+          waitForResourceStatusToContain('Bound');
```
bipuladh, line 98:
```suggestion
         resourceStatusShouldContain('Bound');
```

bipuladh (changes requested):

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,6 @@
-export const resourceStatusShouldContain = (desiredStatus: string) =>
-  cy.contains('[data-test="status-text"]', desiredStatus);
+/* If no timeout is given, the function will take the default Timeout of 30 seconds
+frontend/packages/integration-tests-cypress/cypress.json */
+export const resourceStatusShouldContain = (desiredStatus: string, timeout?: number) =>
+  timeout
+    ? cy.contains('[data-test="status-text"]', desiredStatus, { timeout })
+    : cy.contains('[data-test="status-text"]', desiredStatus);
```
bipuladh, line 6:
```suggestion
export const resourceStatusShouldContain = (desiredStatus: string) =>
    cy.contains('[data-test="status-text"]', desiredStatus, 60000)
```

bipuladh, line 6:
```suggestion
export const resourceStatusShouldContain = (desiredStatus: string, timeout?: number = 30000) =>
  cy.contains('[data-test="status-text"]', desiredStatus, { timeout });
```

bipuladh (commented):
/lgtm

bipuladh:
@dtaylor113 @spadgett, please review

bipuladh:
/retitle Bug 1872340: E2E test for restoring a claim from Snapshot

frontend/packages/integration-tests-cypress/tests/storage/snapshot.spec.ts
```diff
@@ -75,7 +76,31 @@ if (Cypress.env('BRIDGE_AWS')) {
...
+      modal.submit();
+      modal.shouldBeClosed();
+      cy.exec(
+        `oc patch Deployment ${
+          testerDeployment.metadata.name
+        } --type='json' -n ${testName} -p '[${JSON.stringify(PatchForVolume)}]'`,
+      )
+        .its('stdout')
+        .then((res) => {
+          expect(res).toBe(`deployment.apps/${testerDeployment.metadata.name} patched`);
```
spadgett, line 97: We shouldn't check this text. If the command is successful (return code 0), that is sufficient to know it worked.

frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -1,2 +1,2 @@
-export const resourceStatusShouldContain = (desiredStatus: string) =>
-  cy.contains('[data-test="status-text"]', desiredStatus);
+export const resourceStatusShouldContain = (desiredStatus: string, timeout = 30000) =>
+  cy.contains('[data-test="status-text"]', desiredStatus, { timeout });
```
spadgett, line 2: `timeout` should be undefined by default so it uses the global command timeout. It would probably be better to take an optional `options` parameter so you can set any of the options.

```suggestion
export const resourceStatusShouldContain = (desiredStatus: string, options) =>
  cy.contains('[data-test="status-text"]', desiredStatus, options);
```

frontend/packages/integration-tests-cypress/mocks/snapshot.ts
```diff
@@ -80,3 +80,14 @@ export const SnapshotClass = {
   driver: 'ebs.csi.aws.com',
   deletionPolicy: 'Delete',
 };
+
+export const PatchForVolume = {
```
spadgett, line 84: None of these variables should start with an uppercase letter. Use the `Patch` type.

```suggestion
export const patchForVolume: Patch = {
```

frontend/packages/integration-tests-cypress/views/common.ts
dtaylor113: Hi, also you removed the default `timeout = 30000` which I believe another snapshot test was relying on.

dtaylor113 (changes requested):
lgtm other than @spadgett comments, pls address -thanks

spadgett (approved):
/lgtm

dtaylor113 (approved):

PR #6486, opened 2020-09-01
Bug 1860985:  Update the minimum requirement for OCS

gnehapk:
- Updated the minimum requirement text
- Replaced the word `Location` with  `Zone` across all pages

Screenshot from 2020-09-03 11-25-58

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
@@ -168,8 +168,8 @@ export const CreateInternalCluster = withHandlePromise<
         >
           <>
             <div>
-              Select at least 3 nodes in different zones you wish to use. The recommended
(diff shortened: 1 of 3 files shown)


gnehapk:
/hold 
Message needs to be updated

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/types.ts
```diff
@@ -14,7 +14,7 @@ export enum DiskType {
   Partition = 'Partition',
 }
 
-export enum DiskMechanicalProperty {
+export enum DiskMechanicalProperties {
```
bipuladh, line 17:
```suggestion
```
Since it can only have one property of either SSD or HDD I find it perfect as it is (in singular)

bipuladh (changes requested):
Everything else looks good.

gnehapk:
/hold cancel
Message updated

bipuladh (approved):
/lgtm

PR #6487, opened 2020-09-01
Bug 1871768: Add supported types to file upload

glekner:
Now when browsing for a file, you can only pick one of the following files: `.iso,.img,.qcow2,.gz,.xz`
If a user drags an unsupported file to the file input, it will register, but when he submits the form he will see the following screenshot:

Screen Shot 2020-08-31 at 22 56 20


frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form-status.tsx
```diff
@@ -180,7 +180,7 @@ export type UploadPVCFormStatusProps = {
   upload: DataUpload;
   isSubmitting: boolean;
   isAllocating: boolean;
-  allocateError: string;
+  allocateError: string | React.ReactNode;
```
yaacov, line 183: ReactNode include strings, no ?

PR #6488, opened 2020-09-01
Bug 1873303: Protractor - added wait for Launcher Card to not initially exist, then is displayed when CR exists

dtaylor113:


frontend/integration-tests/tests/dashboards/project-dashboard.scenario.ts
```diff
@@ -174,6 +174,7 @@ describe('Project Dashboard', () => {
     });
 
     it('is displayed when CR exists', async () => {
+      await browser.wait(until.not(until.presenceOf(projectDashboardView.launcherCard)));
```
spadgett, line 177: Were you able to reproduce? I'm not sure this is the issue. I suspect the `isPresent` call below happens asynchronously, and `createResource` completes too fast in some cases.

dtaylor113: ok, fixed to use `await`

spadgett (approved):
/lgtm

PR #6490, opened 2020-09-01
Bug 1873612: Handle empty responses if Content-Length is not specified

spadgett:
`Content-Length` is not a required response header. Check if `response.text()` is empty instead of relying on the header.

/assign @dtaylor113


bipuladh:
/lgtm
My PR is somewhat dependent on it.
cc @dtaylor113

PR #6494, opened 2020-09-02
Bug 1874817: fixes issue with route creation in case of no ports exposed

invincibleJai:
**Fixes**: 
- https://issues.redhat.com/browse/ODC-4702
- https://issues.redhat.com/browse/ODC-4701

**Analysis / Root cause**: 
- For Go sample repo builder image is not recommended
- No Route created if user deploys a Go app via Git import flow/Dev catalog

**Solution Description**: 
- Update build-tools to support *.go
- If there are no Ports exposed and use user provided port as target port or default port

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


divyanshiGupta (approved):
/lgtm

divyanshiGupta:
@invincibleJai I just noticed that if we create a go app using the sample from the new sample catalog then the app is not created since now you have added a check for `Source` buildStrategy on creation of service and route and thus even when there are no ports and hence no `targetPort` it tries to create the route and fails.
image

This can be fixed easily by adding `defaultUnknownPort: 8080` to `ImportSamplePage` form initial-values. Since we are not giving an option for editing route properties here so in case of no ports `targetPort` is always going to be `8080`.

invincibleJai:
> @invincibleJai I just noticed that if we create a go app using the sample from the new samples catalog then the app is not created since now you have added a check for `Source` buildStrategy on creation of service and route and thus even when there are no ports and hence no `targetPort` it tries to create the route and fails.
> 
> This can be fixed easily by adding `defaultUnknownPort: 8080` to `ImportSamplePage` form initial-values. Since we are not giving an option for editing route properties here so in case of no ports `targetPort` is always going to be `8080`.

Thanks @divyanshiGupta , have updated as suggested

karthikjeeyar:
Route creation happens successfully now, but while editing the go application I am not able to edit the `TargetPort`. 
 While creation i have specified the port as 8081, and during editing i am trying to change it to 8080.

karthikjeeyar:
Verified locally, Editing targetPort values, route creation and builder image detection for on Golang applications works as expected.
/lgtm

PR #6495, opened 2020-09-02
Bug 1862918: add support for import SRIOV networks from oVirt

suomiy:


suomiy:
/hold

needs to be tested first

suomiy:
tested it and it should be working fine with this approach

please review
@glekner @irosenzw  @pcbailey @yaacov

PR #6496, opened 2020-09-02
Bug 1850583: add better connection error message for VM Import

suomiy:


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/import-providers-tab/providers/vm-import-provider-object-status.tsx
```diff
@@ -46,9 +46,9 @@ const ConnectionFailedInfra: React.FC<StatusProps> = ({ provider }) => (
   <Alert
     isInline
     variant={AlertVariant.warning}
-    title={`Can not verify ${getProviderEndpointName(provider)} credentials, connection to the ${
-      provider === VMImportProvider.OVIRT ? 'OvirtProvider' : 'V2V VMWare'
-    } failed.`}
+    title={`Provided connection info is not correct. Connection to the ${getProviderEndpointName(
```
pcbailey, line 49: I would spell out "information". "Info" sounds very informal.

I'm conflicted about the second sentence. "Connection to the RHV API failed." sounds correct, but "Connection to the vCenter failed." doesn't. Typically, you would say "Connection to vCenter failed." Thoughts?

suomiy: done. Ok I changed the `the` resolution. I had done it for all the other use cases. Can you please double check it is correct?

PR #6497, opened 2020-09-02
Bug 1873331: Adds alt-text to masthead icons

sg00dwin:
Added `alt-text` to `svg` icons.
Buttons have `aria-label`

Bug https://bugzilla.redhat.com/show_bug.cgi?id=1873331

### DIFF
--- a/frontend/public/components/masthead-toolbar.jsx
+++ b/frontend/public/components/masthead-toolbar.jsx
@@ -564,7 +564,7 @@ class MastheadToolbarContents_ extends React.Component {
                   isRead
                   count={notificationAlerts?.data?.length || 0}
                 >
-                  <BellIcon />
+                  <BellIcon alt="" />
                 </NotificationBadge>
               </PageHeaderToolsItem>
             )}
@@ -575,7 +575,7 @@ class MastheadToolbarContents_ extends React.Component {
                   className="pf-c-button pf-m-plain"
                   aria-label="Import YAML"
                 >
-                  <PlusCircleIcon className="co-masthead-icon" />
+                  <PlusCircleIcon className="co-masthead-icon" alt="" />
                 </Link>
               </Tooltip>
             </PageHeaderToolsItem>
@@ -597,7 +597,7 @@ class MastheadToolbarContents_ extends React.Component {
                   ),
                 )}
                 position="right"
-                toggleIcon={<QuestionCircleIcon />}
+                toggleIcon={<QuestionCircleIcon alt="" />}
                 isGrouped
               />
             </PageHeaderToolsItem>


frontend/public/components/masthead-toolbar.jsx
```diff
@@ -564,7 +564,7 @@ class MastheadToolbarContents_ extends React.Component {
                   isRead
                   count={notificationAlerts?.data?.length || 0}
                 >
-                  <BellIcon />
+                  <BellIcon alt-text="Notification Bell icon" />
```
spadgett, line 567:
```suggestion
                  <BellIcon alt-text="View Notifications" />
```

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -575,7 +575,7 @@ class MastheadToolbarContents_ extends React.Component {
                   className="pf-c-button pf-m-plain"
                   aria-label="Import YAML"
                 >
-                  <PlusCircleIcon className="co-masthead-icon" />
+                  <PlusCircleIcon className="co-masthead-icon" alt-text="Import YAML icon" />
```
spadgett, line 578:
```suggestion
                  <PlusCircleIcon className="co-masthead-icon" alt-text="Import YAML" />
```

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -597,7 +597,7 @@ class MastheadToolbarContents_ extends React.Component {
                   ),
                 )}
                 position="right"
-                toggleIcon={<QuestionCircleIcon />}
+                toggleIcon={<QuestionCircleIcon alt-text="Help Menu icon" />}
```
spadgett, line 600:
```suggestion
                toggleIcon={<QuestionCircleIcon alt-text="Help Menu" />}
```

jessiehuff (commented):
@spadgett is correct, typically words like "image", "icon", etc. are left off in labeling. This [WAI article](https://www.w3.org/WAI/tutorials/images/tips/) talks about recommendations for labeling, the section discussing this kind of wording is under "Superfluous information in the text alternative". In terms of the user experience for screen readers, the alt text here actually won't be read to users as there's currently an `aria-hidden` on the svg which tells screen readers to ignore it. It looks like when [creating our icons](https://github.com/patternfly/patternfly-react/blob/master/packages/react-icons/src/createIcon.tsx#L83), if you don't include a title, it adds the `aria-hidden` to the svg. There are some options here, but I'm thinking that this falls under the category of functional images and perhaps we should consider making the `alt=""` since we already have the button aria-label. Scott O'Hara is a thought leader in the a11y space, and he does an excellent job of laying out the options in this [article](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html#image-elements-within-buttons) covering our use case. In Option 2 the button already has an aria-label, they just add the empty alt text to avoid redundancy and repetition. We could add a title to the icon which would make the svg not hidden anymore, but then we run the risk of some screen readers reading "Help menu Help menu."

spadgett:
+1 if there is already a button with an aria-label, I agree we should use `alt-text=""`

spadgett:
@jessiehuff Any idea why this wasn't caught by Axe? Is there a rule we need to enable?

cc @dtaylor113

jessiehuff:
Technically speaking, when there is an `aria-hidden` on an image or icon, a label isn't required since it's not accessible to a screen reader. Axe describes the alt rule [here](https://dequeuniversity.com/rules/axe/3.5/input-image-alt?application=rocketvalidator) stating that an alt text shouldn't be empty unless hidden using CSS with `display: none` or `aria-hidden="true"`. Many people recommend it though since having the empty alt is more like a fallback.

frontend/public/components/masthead-toolbar.jsx
sg00dwin: included `alt-text=""` based on comments below

jessiehuff:
Hey @sg00dwin, I didn't catch it before and even wrote it incorrectly earlier myself, but I think it should be an alt attribute so `alt=""`

sg00dwin:
> Hey @sg00dwin, I didn't catch it before and even wrote it incorrectly earlier myself, but I think it should be an alt attribute so `alt=""`

Updated changes to use `alt=""`

spadgett (approved):
/lgtm

PR #6506, opened 2020-09-03
Bug 1875308: Add support for more params and create resources only if required

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4466
https://bugzilla.redhat.com/show_bug.cgi?id=1875308

**Analysis / Root cause**: 
With the new OpenShift Pipeline Operator 1.1 (currently available as preview) the Tasks changes and uses (new) parameters instead of resources to link the Git repository and Image registry.

**Solution Description**: 
* Check if the Pipeline (template) provides more known parameters (GIT_REPO, GIT_REVISION and IMAGE_NAME) and fill the default parameters of the new pipeline.
* Check if the Pipeline (template) defines resources for the Git repository and Image registry. Create new resources only if there are such parameters defined.  

**Unchanged behavior with the Pipeline Operator 1.0.x (with this PR):**

Pipeline Details:
odc-4466-details-v10

Pipeline Parameters:
odc-4466-parameters-v10

Pipeline Resources:
odc-4466-resources-v10

Start Pipeline Modal:
odc-4466-startpipeline-v10

**Changed behavior with the Pipeline Operator 1.1.x (with this PR):**

Pipeline Details:
odc-4466-details-v11

Pipeline Parameters:
odc-4466-parameters-v11

Pipeline Resources:
odc-4466-resources-v11

Start Pipeline Modal:
odc-4466-startpipeline-v11

Unchanged

**Test setup:**
* Install the Pipeline Operator 1.0.1 (4.5) or 1.1.1 (preview)
* Create a new Deployment with Pipeline with the Import from Git flow
* Check the generated Pipeline
* You should start it without adding additional parameters on the 1.1 version.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
+++ b/frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
@@ -1,8 +1,13 @@
 import * as _ from 'lodash';
 import { k8sCreate } from '@console/internal/module/k8s';
+import { Pipeline } from 'packages/dev-console/src/utils/pipeline-augment';
 import { PipelineModel } from '../../../models';
 import { GitImportFormData } from '../import-types';
 import { createPipelineResource } from '../../pipelines/pipeline-resource/pipelineResource-utils';
+
+const getImageUrl = (name: string, namespace: string) => {
+  return `image-registry.openshift-image-registry.svc:5000/${namespace}/${name}`;
+};
 
 export const createGitResource = (url: string, namespace: string, ref: string = 'master') => {
   const params = { url, revision: ref };
@@ -11,7 +16,7 @@ export const createGitResource = (url: string, namespace: string, ref: string = 
 
 export const createImageResource = (name: string, namespace: string) => {
   const params = {
-    url: `image-registry.openshift-image-registry.svc:5000/${namespace}/${name}`,
+    url: getImageUrl(name, namespace),
   };
 
   return createPipelineResource(params, 'image', namespace);
@@ -24,7 +29,7 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
     git,
     pipeline,
   } = formData;
-  const template = _.cloneDeep(pipeline.template);
+  const template = _.cloneDeep(pipeline.template) as Pipeline;
 
   template.metadata = {
     name: `${name}`,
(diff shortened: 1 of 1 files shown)


frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -24,28 +29,34 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
...
+    }
+  });
+
+  if (template.spec?.resources?.find((r) => r.type === 'git' && r.name === 'app-source')) {
     await createGitResource(git.url, namespace, git.ref);
+  }
+  if (template.spec?.resources?.find((r) => r.type === 'image' && r.name === 'app-image')) {
     await createImageResource(name, namespace);
-  } catch (err) {
-    throw err;
```
andrewballantyne, line 48: Hah, a fair catch. Not really doing anything if we are just catching and rethrowing. Good cleanup :)

frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -24,28 +29,34 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
...
+        return { ...param, default: getImageUrl(name, namespace) };
+      default:
+        return param;
+    }
+  });
+
+  if (template.spec?.resources?.find((r) => r.type === 'git' && r.name === 'app-source')) {
     await createGitResource(git.url, namespace, git.ref);
+  }
+  if (template.spec?.resources?.find((r) => r.type === 'image' && r.name === 'app-image')) {
```
andrewballantyne, line 58:
```suggestion
  if (template.spec.resources?.find((r) => r.type === 'git' && r.name === 'app-source')) {
    await createGitResource(git.url, namespace, git.ref);
  }
  if (template.spec.resources?.find((r) => r.type === 'image' && r.name === 'app-image')) {
```
Spec is required in the type.

andrewballantyne: I like this implementation that looks specifically for those names. It could blow back on us if they changed it upstream, but I think it's important that we try to aim for existing functionality. FWIW though, upstream recommends not to mix workspaces & pipeline resources... so this is likely overcautious code, but I see no downside really.

andrewballantyne (changes requested):
Minor change to keep the optional chaining honest.

frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
jerolimov: Removed the optional check of spec. (And rebased)

Yeah I do not know if we want check here for just the type or the name as well. I do not know if all pipeline templates use exactly this names. Should we create a git resource if there is a resource with type === 'git' and name === 'app' ?! :thinking: 

If we want the "highest compatibly" I could remove the name check here... wdyt?

andrewballantyne: Yeah, I was originally going to ask you to remove the name... but let us not. It's less intrusive if the user wants to use a git resource for something entirely different in their template. I think this implementation is fine 👍 .

andrewballantyne (approved):

rottencandy (approved):
/lgtm

PR #6509, opened 2020-09-03
Bug 1875345: Query Browser: Fix JS error when label key is `id`

kyoto:


kyoto:
Not handling `id` columns appears to be a PatternFly limitation: https://github.com/patternfly/patternfly-react/issues/4283

spadgett (approved):
/lgtm

PR #6511, opened 2020-09-03
Bug 1850574: add better cloud init message for ovirt import

suomiy:


matthewcarleton:
thanks @suomiy I think this looks good. I'm asking around for other outside opinions to make sure it makes sense too :)

matthewcarleton:
@suomiy I had a good chat with some other designers and they made the good point of surfacing this type of message in either the Connect to provider step where it would be more of a general list of things that might not show up. If we did that we could then consider hiding the advanced > cloud init step entirely. WDYT?

suomiy:
@matthewcarleton I to not think this is a very useful message to have in the connect to provider step. There is already too much happening there. Plus this is very specific

matthewcarleton:
@suomiy My line of thinking here is that for a user to know as soon as possible where they could see a gap in their expectations and what we have in the wizard. 
At the very least changing the title is probably a good idea ( I know I wrote this to begin with haha)

**Possible data missing from import**
This wizard may not show all data from the import. If data is not showing here view the details of the import after it has completed for a full data set.

suomiy:
updated the text 

aa

please review 
@irosenzw @yaacov @glekner @pcbailey

yaacov:
I assume @matthewcarleton 's :+1: means he approve :-)

/lgtm

suomiy:
updated the design 

bb

suomiy:
@yaacov can you relgtm?

matthewcarleton:
LGTM!

PR #6517, opened 2020-09-03
Bug 1875488: Fix sample application form create

divyanshiGupta:
Jira - https://issues.redhat.com/browse/ODC-4728

Keep the create button enabled even when the form is not dirty since it is a sample application form and user might not want to change anything at all and straightaway create the application.

**Screenshot:**
image

/kind bug


rohitkrai03 (approved):
/approve

sahil143 (approved):
/lgtm

verified changes locally

PR #6518, opened 2020-09-03
Bug 1875599: trigger groupings in Event Listener details page

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-3710

**Root Analysis:**
In the EventListener spec there is a property called "triggers" which can have 1 to n TriggerBindings and a single TriggerTemplate associated to it. Currently we do not have a good way to represent this. We flatten the content today, and it's a bit misleading.

**Solution Description:**
Nested the associated Trigger Bindings under Trigger Templates

**Screenshot:**
TB

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.scss
+++ b/frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.scss
@@ -1,3 +1,5 @@
 .odc-dynamic-resource-link-list {
-  margin-bottom: var(--pf-global--spacer--lg);
+  &--addSpaceBelow {
+    margin-bottom: var(--pf-global--spacer--lg);
+  }
 }
--- a/frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 import { ResourceLink } from '@console/internal/components/utils';
 import { K8sKind, referenceForModel } from '@console/internal/module/k8s';
 
@@ -13,21 +14,27 @@ export type ResourceModelLink = {
 type DynamicResourceLinkListProps = {
   links: ResourceModelLink[];
   namespace: string;
-  title: string;
+  title?: string;
+  removeSpaceBelow?: boolean;
 };
 
 const DynamicResourceLinkList: React.FC<DynamicResourceLinkListProps> = ({
   links = [],
   namespace,
   title,
+  removeSpaceBelow,
 }) => {
   if (links.length === 0) {
     return null;
   }
   return (
-    <div className="odc-dynamic-resource-link-list">
+    <div
+      className={classNames('odc-dynamic-resource-link-list', {
+        'odc-dynamic-resource-link-list--addSpaceBelow': !removeSpaceBelow,
+      })}
+    >
       <dl>
-        <dt>{title}</dt>
+        {title && <dt>{title}</dt>}
         <dd>
           {links.map(({ name, model, displayName = '' }) => {
             const kind = referenceForModel(model);
--- a/frontend/packages/dev-console/src/components/pipelines/utils/triggers.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/utils/triggers.ts
@@ -19,6 +19,7 @@ import {
   EventListenerKindTrigger,
   TriggerBindingKind,
   TriggerTemplateKind,
+  EventListenerKindBindingReference,
 } from '../resource-types';
 import { ResourceModelLink } from '../resource-overview/DynamicResourceLinkList';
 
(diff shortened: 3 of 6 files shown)


frontend/packages/dev-console/src/components/pipelines/utils/triggers.ts
```diff
@@ -162,28 +163,15 @@ export const useEventListenerURL = (
...
-  );
-};
+  bindings: EventListenerKindBindingReference[],
+): ResourceModelLink[] =>
+  bindings
+    ? bindings.reduce((acc, binding) => {
+        acc.push({ model: getResourceModelFromBindingKind(binding.kind), name: binding.name });
+        return acc;
+      }, [] as ResourceModelLink[])
+    : null;
```
andrewballantyne, line 174:
```suggestion
  bindings?.map((binding) => ({
    model: getResourceModelFromBindingKind(binding.kind),
    name: binding.name,
  }));
```
?

frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
```diff
@@ -25,9 +23,9 @@ const DynamicResourceLinkList: React.FC<DynamicResourceLinkListProps> = ({
     return null;
   }
   return (
-    <div className="odc-dynamic-resource-link-list">
```
andrewballantyne, line 28: You caused this issue on the Trigger Templates details page... careful removing styles without tracing all usages.

Screen Shot 2020-09-03 at 1 14 07 PM

Maybe make it a prop to disable spacing?

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/Triggers.tsx
```diff
@@ -0,0 +1,52 @@
...
+} from '../resource-overview/DynamicResourceLinkList';
+import { getEventListenerTriggerBindingNames } from '../utils/triggers';
+import './Triggers.scss';
+
+interface TriggersProps {
+  triggers: EventListenerKindTrigger[];
+  namespace: string;
+}
+
+const Triggers: React.FC<TriggersProps> = ({ namespace, triggers }) => {
```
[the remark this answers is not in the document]
debsmita1: Changed it to `EventListenerTriggers`

frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
debsmita1: fixed it ! using a prop to add spacing

frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
```diff
@@ -13,21 +14,27 @@ export type ResourceModelLink = {
...
 }) => {
   if (links.length === 0) {
     return null;
   }
   return (
-    <div className="odc-dynamic-resource-link-list">
+    <div
+      className={classNames({
+        'odc-dynamic-resource-link-list': addSpaceBelow,
+      })}
```
andrewballantyne, line 34: Technically [BEM](http://getbem.com/introduction/) wants us to use modifiers in this case. Suggest this:
```suggestion
      className={classNames('odc-dynamic-resource-link-list', {
        'odc-dynamic-resource-link-list--addSpaceBelow': addSpaceBelow,
      })}
```

frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.tsx
```diff
@@ -13,21 +14,27 @@ export type ResourceModelLink = {
...
-  title: string;
+  title?: string;
+  addSpaceBelow?: boolean;
 };
 
 const DynamicResourceLinkList: React.FC<DynamicResourceLinkListProps> = ({
   links = [],
   namespace,
   title,
+  addSpaceBelow = true,
```
andrewballantyne, line 25: A suggestion instead of having to place a default would be to make this a "removal" command. Easier to use as a react component prop & you don't need to place a default as it's a boolean. Could go with `removeSpacing`... this is all a suggestion, I am good with both ways.

andrewballantyne (changes requested):
One change and one suggestion left.

frontend/packages/dev-console/src/components/pipelines/resource-overview/DynamicResourceLinkList.scss
```diff
@@ -1,3 +1,3 @@
-.odc-dynamic-resource-link-list {
+.odc-dynamic-resource-link-list--addSpaceBelow {
```
andrewballantyne, line 1:
```suggestion
.odc-dynamic-resource-link-list {
  &--addSpaceBelow {
```

andrewballantyne (approved):

jerolimov:
I noticed that an trigger binding in an EventListener created by the console pipeline "Add Trigger" feature has no name. So it looks like this:

image

The name wasn't shown before, so I think this is okay in context of this UI cleanup.

jerolimov (approved):
/lgtm

andrewballantyne:
> I noticed that an trigger binding in an EventListener created by the console pipeline "Add Trigger" feature has no name. So it looks like this:
> 
> image
> 
> The name wasn't shown before, so I think this is okay in context of this UI cleanup.

That's not right 🤔  -- give me a moment to finish what I am doing and I'll see what is up... Is this related to 1.1.1? Or 1.0.1 of the Pipeline Operator @jerolimov

jerolimov:
> Is this related to 1.1.1?

Maybe?! At least I have seen this on 1.1.1. I will check this on 1.0 as well. But this does not affect this PR.

andrewballantyne:
> > Is this related to 1.1.1?
> 
> Maybe?! At least I have seen this on 1.1.1. I will check this on 1.0 as well. But this does not affect this PR.

I have no idea how you got this.

image

/lgtm cancel
/approve

andrewballantyne:
@jerolimov Please recheck, I think you might have a local issue? I don't see it on my cluster. If you can reproduce, maybe share your `EventListener.spec`? Because this seems very odd.

andrewballantyne:
Ah sorry, I was just impatient... this is the result of Pipeline Operator 1.1.1. Which [this bug](https://issues.redhat.com/browse/ODC-4496) is addressing. I think you had partial 1.1.1 data in play @jerolimov which is backwards incompatible with the current 1.0.1 logic.

/lgtm

PR #6523, opened 2020-09-04
Bug 1872919: Add routes for OCS OLM pages with full resource references

bipuladh:


frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -134,6 +134,19 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+  {
+    type: 'Page/Route',
+    properties: {
+      exact: true,
+      path: `/k8s/ns/:ns/${referenceForModel(
+        ClusterServiceVersionModel,
+      )}/:appName/${apiObjectRef}/~new`,
+      loader: () =>
+        import('./components/ocs-install/install-page' /* webpackChunkName: "install-page" */).then(
+          (m) => m.default,
```
gnehapk, line 146: Shouldn't the above `'Page/Route` should be updated with these details? Why another object is required?

bipuladh: We need to support both.

gnehapk (approved):

frontend/packages/noobaa-storage-plugin/src/plugin.ts
```diff
@@ -92,6 +123,25 @@ const plugin: Plugin<ConsumedExtensions> = [
       required: [NOOBAA_FLAG],
     },
   },
+  {
+    type: 'Page/Route',
+    properties: {
+      exact: true,
+      path: [
+        `/k8s/ns/:ns/${referenceForModel(ClusterServiceVersionModel)}/:appName/${referenceForModel(
```
afreen23, line 131: Would be better if you could extract `referenceForModel(ClusterServiceVersionModel)` once in  a variable and use everywhere.

afreen23 (changes requested):
We need to do this chnage for LSO as well, for `localvolumeset` and `autodetect`.

frontend/packages/noobaa-storage-plugin/src/plugin.ts
bipuladh: Keeping it this way IMO keeps it more readable. It conveys that we are using the GVK name of the CSVModel.

PR #6526, opened 2020-09-04
Bug 1875758: Remove extensions for Volume Snapshot from Ceph Plugin

bipuladh:


cloudbehl:
Can you remove from docs too on how to enable snapshot?

bipuladh:
> Can you remove from docs too on how to enable snapshot?

okay will do it.

afreen23 (approved):

PR #6529, opened 2020-09-04
Bug 1875794: Feature guards for Minimal Deployment and Cluster Encryption

bipuladh:


cloudbehl:
Is there parallel OCS-op PR for this? can you tag it

bipuladh:
> Is there parallel OCS-op PR for this? can you tag it

https://github.com/openshift/ocs-operator/pull/754

cloudbehl:
We need to handle `getOCSRequestData` as well as this flag is not there in 4.5
 https://github.com/openshift/console/blob/21472765fd769cec49194cee13f198611e9d5566/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx#L77

cloudbehl:
This needs to be conditional based on flag enabled.
https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts#L47..L49

PR #6531, opened 2020-09-04
Bug 1858578: Change max name length for VMs imported by RHV

pcbailey:
This PR limits the length of VM names to 63 characters for VMs that are imported from a RHV provider.


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/ovirt/ovirt-state-update.ts
```diff
@@ -61,6 +62,19 @@ const startControllerAndCleanup = (options: UpdateOptions) => {
   cleanupOvirtProvider(options); // call should be idempotent and called on every provider change
```
suomiy, line 63: this whole function `startControllerAndCleanup` will get called on every provider change

so we have to update the validations unconditionally every time like this:

```ts
dispatch(
  vmWizardInternalActions[InternalActionType.UpdateVmSettingsField](id, VMSettingsField.NAME, {
    validations:  isOvirtProvider(state, id) ? [
      {
        type: UIValidationType.LENGTH,
        settings: {
          min: 1,
          max: 63,
        },
      },
    ] : [],
  }),
);
```

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/vmware/vmware-state-update.ts
```diff
@@ -58,6 +58,11 @@ const startControllerAndCleanup = (options: UpdateOptions) => {
   cleanupVmWareProvider(options); // call should be idempotent and called on every provider change
 
   if (isVMWareProvider(state, id) && namespace) {
+    dispatch(
```
suomiy, line 61: should not be here

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -57,6 +58,7 @@ const validateVm: Validator = (field, options) => {
         : VIRTUAL_MACHINE_EXISTS,
       subject: getFieldTitle(iGetFieldKey(field)),
     },
+    immutableListToJS(iGet(field, 'validations')) as UIValidation[],
```
suomiy, line 61: probably better to enhance the object arg than positional arguments

frontend/packages/kubevirt-plugin/src/types/ui/ui.ts
```diff
@@ -0,0 +1,8 @@
+export enum UIValidationType {
+  LENGTH,
+}
+
+export type UIValidation = {
+  type: UIValidationType;
+  settings: { [k: string]: any };
```
suomiy, line 7: should be optional for a generic type

suomiy: I would prefer to put all the known types here compared to not knowing what is inside. min/max

frontend/packages/kubevirt-plugin/src/utils/validations/vm/vm.ts
```diff
@@ -16,14 +40,34 @@ export const validateVmLikeEntityName = (
...
-        subject,
-      });
+  const dnsValidationFailed = dnsValidation?.type === ValidationErrorType.Error;
+  const lengthValidation = validations?.[0]
+    ? validateUIValidationType(validations?.[0], value)
+    : null;
+  const lengthValidationFailed = lengthValidation?.type === ValidationErrorType.Error;
+
+  let validation = dnsValidation;
+  if (lengthValidationFailed && dnsValidationFailed) {
```
suomiy, line 53: I am sorry, but this all seems hacky. Can't we just add  optional `min` and `max` values to validateDNS1123SubdomainValue? This would make the code much simpler I think.

frontend/packages/kubevirt-plugin/src/utils/validations/vm/vm.ts
```diff
@@ -16,14 +40,34 @@ export const validateVmLikeEntityName = (
...
 ): ValidationObject => {
   const dnsValidation = validateDNS1123SubdomainValue(value, { subject });
-  return dnsValidation && dnsValidation.type === ValidationErrorType.Error
-    ? dnsValidation
-    : validateEntityAlreadyExists(value, namespace, vmLikeEntities, {
-        errorMessage: existsErrorMessage,
-        subject,
-      });
+  const dnsValidationFailed = dnsValidation?.type === ValidationErrorType.Error;
+  const lengthValidation = validations?.[0]
```
suomiy, line 47: please find the lengthValidation by type, not by index. Then we can just pass the values to validateDNS1123SubdomainValue as I mentioned. For now we don't need the generic validateUIValidationType function.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/utils/validations/vm/vm.ts
pcbailey: No need to apologize. I can definitely do that, but I'm not sure how it will allow me to clean this up much more. Since `validateDNS1123SubdomainValue` validates more than just length, we have to account for multiple scenarios. If it only validated length, the logic could be made much clearer. Perhaps I'm just not seeing a better way. Could you give me an idea of what you're thinking?

suomiy: It can only validate what it is validating already plus simple min/max params. No need for other scenarios. Let's not focus on future `UIValidation`s 
It will help us not having to dissect returned messages.

pcbailey: I was trying to avoid touching code outside of kubevirt, but that would make it much simpler.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/ovirt/ovirt-state-update.ts
pcbailey: I made this change, but it doesn't look like the `validations` property is changing back to `[]` when the provider changes away from ovirt.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/utils/validations/vm/vm.ts
pcbailey: Sorry for the confusion. Done!

frontend/packages/kubevirt-plugin/src/utils/validations/vm/vm.ts
pcbailey: Done!

frontend/packages/console-shared/src/utils/validation/validation.ts
```diff
@@ -54,7 +60,11 @@ export const validateDNS1123SubdomainValue = (
   const forbiddenCharacters = new Set<string>();
   const validationSentences = [];
 
-  if (value.length > DNS1123_MAX_LENGTH) {
+  if (min && value.length < min) {
+    validationSentences.push(getStringTooShortErrorMsg(min));
+  } else if (max && value.length > max) {
+    validationSentences.push(getStringTooLongErrorMsg(max));
+  } else if (value.length > DNS1123_MAX_LENGTH) {
```
suomiy, line 67: we can default max to DNS1123_MAX_LENGTH and then we can shorten these two cases under one. Plus we can delete DNS1123_TOO_LONG_ERROR constant

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/ovirt/ovirt-state-update.ts
suomiy: because you have it under `isOvirtProvider(state, id)` It should be called unconditionally (except the initil provider change condition)

frontend/packages/kubevirt-plugin/src/utils/validations/common.ts
```diff
@@ -95,3 +96,8 @@ export const validateBus = (value: DiskBus, allowedBuses: Set<DiskBus>): Validat
   }
   return null;
 };
+
+export const getValidationByType = (
+  validationList: UIValidation[],
+  type: UIValidationType,
+): UIValidation => validationList.find((val) => val.type === type);
```
suomiy, line 103:
```suggestion
): UIValidation => validationList?.find((val) => val.type === type);
```

suomiy (changes requested):

frontend/packages/console-shared/src/utils/validation/validation.ts
pcbailey: Very nice! I wish I had thought of that. =) Done!

frontend/packages/kubevirt-plugin/src/utils/validations/common.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/vmware/vmware-state-update.ts
pcbailey: Done!

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/ovirt/ovirt-state-update.ts
pcbailey: I took it outside of the conditional block, but it still isn't changing. I verified that this `dispatch` call has the right value, but haven't figure out why the reducer isn't doing its job, yet.

suomiy: because you are calling `UpdateVmSettingsField` action which has a special implementation for merging fields in `mergeDeepInSpecial` . So it wants to update the field and not set it.

Can you try to use `InternalActionType.SetInVmSettings` action for just setting the value?

note: also if we once add such validations to vmware provider, we should not reset the values here. Not ideal solution at the moment. Maybe we could add a comment describing this behaviour?

frontend/packages/kubevirt-plugin/src/types/ui/ui.ts
suomiy: ping

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/providers/ovirt/ovirt-state-update.ts
pcbailey: Fixed!

frontend/packages/kubevirt-plugin/src/types/ui/ui.ts
pcbailey: Done!

frontend/packages/console-shared/src/utils/validation/validation.ts
```diff
@@ -44,7 +45,11 @@ export const validateEmptyValue = (
 // DNS-1123 subdomain
 export const validateDNS1123SubdomainValue = (
   value: string,
-  { subject } = { subject: 'Name' },
+  { subject, min, max }: { subject: string; min?: number; max?: number } = {
```
suomiy, line 48: this will not default to DNS1123_MAX_LENGTH when empty object is passed. It should always default this value. For example this validation stoped working for basic wizard

suomiy (changes requested):

frontend/packages/console-shared/src/utils/validation/validation.ts
pcbailey: Fixed!

suomiy (approved):

suomiy:
@rawagner can you take a look?

PR #6532, opened 2020-09-04
Bug 1875797: remove vmware-to-kubevirt-os warning from vmWare import wizard

suomiy:


suomiy:
please review
@irosenzw @yaacov @glekner @pcbailey

yaacov:
Without this config map imports may fail because of bad os mapping, I'ts important to let users know this os mapping is missing, and that their imports may fail because of it.

We have a BZ on the backend to provide this config map on all deployments [1].

@jelkosz hi, what did we desiccated to do about this in the BZ that Ilanit opened ?

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1845408

suomiy:
@yaacov the import will not fail, they will just be required to pick the right os. The backend BZ is still valid and will just make life easiers for users.

yaacov:
> the import will not fail, they will just be required to pick the right os.

The next BZ after this one was IIRC "why do I need to pick an OS" :-) , the cause of this issue is missing config map, and the user need to know it's missing (for example to know why they need to pick an OS themselves).

jelkosz:
@yaacov I agree that it would be nice to have the backend bug fixed - on the other hand I dont agree we want to show this error to our users. There is nothing the user can do about it, it is just a scary error message. And the consequence is that he has to pick the OS. So it does not "feel" broken without the error. Just as a field I have to pick. With the message it feels like: "lets just quickly leave the dialog before it bites me"

yaacov:
ok :-)

/lgtm

PR #6535, opened 2020-09-04
Bug 1875878: Move pipelinerun log scroll handler to parent element

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4517

**Analysis / Root cause**: 
The scroll event listener was attached to a div that did not have the `overflow-y` property, hence no events were being fired.

**Solution Description**: 
Move the event listener to it's parent element, which had the `overflow-y` property set to `scroll`. It is likely that a CSS change introduced this bug.

rec
@openshift/team-devconsole-ux 

Unchanged.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (approved):
/lgtm

verified changes locally

jerolimov (approved):
Tested it locally works as expected.

/lgtm

rohitkrai03 (changes requested):
/approve

PR #6537, opened 2020-09-06
Bug 1875347: can't have a disk provision source and use cdrom type as a disk

glekner:


glekner:
@suomiy @yaacov can you review?

PR #6539, opened 2020-09-07
Bug 1876442: fixes custom query selection should not show any query

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4738

**Analysis / Root cause**: 
On selection of customQuery Query editor used to persist last entered query

**Solution Description**: 
On change on Query metric to Custom Query as well updated query editor with empty value

Sep-07-2020 14-16-09

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj (commented):
verified it works as expected
/lgtm

PR #6540, opened 2020-09-07
Bug 1866315: Inconsistent naming in storage cluster creation

cloudbehl:

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/models.ts
+++ b/frontend/packages/ceph-storage-plugin/src/models.ts
@@ -14,8 +14,8 @@ export const CephClusterModel: K8sKind = {
 };
 
 export const OCSServiceModel: K8sKind = {
-  label: 'OCS Cluster Service',
-  labelPlural: 'OCS Cluster Services',
+  label: 'Storage Cluster',
+  labelPlural: 'Storage Clusters',
   apiVersion: 'v1',
   apiGroup: 'ocs.openshift.io',
   plural: 'storageclusters',


frontend/packages/ceph-storage-plugin/src/models.ts
```diff
@@ -14,8 +14,8 @@ export const CephClusterModel: K8sKind = {
 };
 
 export const OCSServiceModel: K8sKind = {
-  label: 'OCS Cluster Service',
-  labelPlural: 'OCS Cluster Services',
+  label: 'Storage Cluster',
+  labelPlural: 'Storage Clusters',
```
afreen23, line 18: Why not change abbr -> SC ?

gnehapk: Even, id can be changed

afreen23:
/lgtm cancel
Lets resolve https://github.com/openshift/console/pull/6540#discussion_r484298590

frontend/packages/ceph-storage-plugin/src/models.ts
cloudbehl: SC is a storage class so make sense to have `OCS` as it differentiates. 

Id is internal and is not used to display

cloudbehl: @afreen23 @gnehapk ^^

gnehapk (approved):

frontend/packages/ceph-storage-plugin/src/models.ts
afreen23: Alright, I am assuming to have Storage Cluster model be aligned to its CRD spec.
Therefore changing abbr and id makes sense .
Also, we have Portworx operator which has Storage Cluster as API and it represents SC only in console.

So are we overriding intentionally the common pattern @cloudbehl ?

cloudbehl: I think Abbr and id are specific to UI only and not related to CRD spec. This is used to override the default functionality. 

I'm good making it default but I found OCS more relevant & unique as an abbr. So Let me know your thoughts & I'll change it accordingly.

afreen23 (approved):
/lgtm

frontend/packages/ceph-storage-plugin/src/models.ts
afreen23: Sure, for this bug fix lets have the current change in and we can discuss over other changes offline.
Thanks!

PR #6541, opened 2020-09-07
Bug 1876515: Fix inline taskSpec error in pipeline page

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-4724

Problem:
Issue with inline spec over taskref causes the pipeline details page to break

Solution:
Handle the inline spec issue and skip the visualization

Screenshots:
image

Test cases:
image
PipelineVisualization tests
image

pipeline-actions tests
image

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Alert } from '@patternfly/react-core';
 import { Pipeline, PipelineRun } from '../../../../utils/pipeline-augment';
+import { hasInlineTaskSpec } from '../../../../utils/pipeline-utils';
 import PipelineTopologyGraph from '../../pipeline-topology/PipelineTopologyGraph';
 import { getTopologyNodesEdges } from '../../pipeline-topology/utils';
 import { PipelineLayout } from '../../pipeline-topology/const';
@@ -17,23 +18,32 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
   pipelineRun,
 }) => {
   const { nodes, edges } = getTopologyNodesEdges(pipeline, pipelineRun);
-
-  if (nodes.length === 0 && edges.length === 0) {
+  let content: React.ReactElement;
+  if (hasInlineTaskSpec(pipeline)) {
+    // TODO: Inline taskSpec is not yet supported feature
+    content = (
+      <Alert
+        variant="info"
+        isInline
+        title="This Pipeline cannot be visualized. Pipeline taskSpec is not supported."
+      />
+    );
+  } else if (nodes.length === 0 && edges.length === 0) {
     // Nothing to render
     // TODO: Confirm wording with UX; ODC-1860
-    return <Alert variant="info" isInline title="This Pipeline has no tasks to visualize." />;
-  }
-
-  return (
-    <div className="odc-pipeline-visualization">
+    content = <Alert variant="info" isInline title="This Pipeline has no tasks to visualize." />;
+  } else {
+    content = (
       <PipelineTopologyGraph
         id={pipelineRun?.metadata?.name || pipeline.metadata.name}
         nodes={nodes}
         edges={edges}
         layout={PipelineLayout.DAGRE_VIEWER}
       />
-    </div>
-  );
+    );
+  }
+
+  return <div className="odc-pipeline-visualization">{content}</div>;
 };
 
 export default PipelineVisualization;
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
@@ -5,7 +5,7 @@ import { Link } from 'react-router-dom';
 import { Tooltip } from '@patternfly/react-core';
 import { K8sResourceKind, referenceForModel } from '@console/internal/module/k8s';
 import { Firehose, resourcePathFromModel } from '@console/internal/components/utils';
-import { runStatus } from '../../../../utils/pipeline-augment';
+import { runStatus, PipelineTaskSpec, PipelineTaskRef } from '../../../../utils/pipeline-augment';
 import { PipelineRunModel, TaskModel, ClusterTaskModel } from '../../../../models';
 import { ColoredStatusIcon } from './StatusIcon';
 import { PipelineVisualizationStepList } from './PipelineVisualizationStepList';
@@ -33,10 +33,8 @@ interface PipelineVisualizationTaskProp {
   namespace: string;
   task: {
     name?: string;
-    taskRef: {
-      name: string;
-      kind?: string;
-    };
+    taskSpec?: PipelineTaskSpec;
+    taskRef?: PipelineTaskRef;
     status?: TaskStatus;
   };
   taskRun?: string;
--- a/frontend/packages/dev-console/src/utils/__tests__/pipeline-test-data.ts
+++ b/frontend/packages/dev-console/src/utils/__tests__/pipeline-test-data.ts
@@ -1,7 +1,7 @@
-import { K8sResourceKind } from '@console/internal/module/k8s';
 import { pipelineTestData, PipelineExampleNames, DataState } from '../../test/pipeline-data';
+import { Pipeline } from '../pipeline-augment';
 
-export const mockPipelinesJSON: K8sResourceKind[] = [
+export const mockPipelinesJSON: Pipeline[] = [
   {
     apiVersion: 'tekton.dev/v1alpha1',
     kind: 'Pipeline',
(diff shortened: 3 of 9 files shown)


frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -181,7 +181,16 @@ const appendPipelineRunStatus = (pipeline, pipelineRun) => {
     return mTask;
   });
 };
-
+export const hasInlineTaskSpec = (pipeline: K8sResourceKind) => {
+  const tasks = (pipeline.spec && pipeline.spec.tasks) || [];
```
divyanshiGupta, line 185:
```suggestion
  const tasks = pipeline.spec?.tasks ?? [];
```

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -192,7 +201,7 @@ export const getPipelineTasks = (
 ): PipelineVisualizationTaskItem[][] => {
   // Each unit in 'out' array is termed as stage | out = [stage1 = [task1], stage2 = [task2,task3], stage3 = [task4]]
   const out = [];
-  if (!pipeline.spec || !pipeline.spec.tasks) {
+  if (!pipeline.spec || !pipeline.spec.tasks || hasInlineTaskSpec(pipeline)) {
```
divyanshiGupta, line 204:
```suggestion
  if (!pipeline.spec?.tasks || hasInlineTaskSpec(pipeline)) {
```

divyanshiGupta (changes requested):

sahil143 (approved):
/lgtm

verified changes locally

sahil143:
/lgtm cancel

It breaks in the Edit Pipeline page
Screenshot 2020-09-09 at 3 26 47 PM

karthikjeeyar:
Added the check in edit scenario, so that if there is an inline taskSpec, it will not throw error and will allow us to add/modify the tasks in the builder form.
image

andrewballantyne:
@karthikjeeyar I think we need UX here... I don't think it's appropriate to edit a Pipeline and clear out the tasks because they are using inline spec - this could have wildly unexpected results for the user. I think we need to take a path of least resistance here - it is an unsupported feature of the DevConsole. Maybe we showcase a message saying "Using tasks inlineSpec, which is unsupported" and removing the Edit Pipeline kebab option. 

cc @bgliwa01 / @openshift/team-devconsole-ux

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -181,7 +181,16 @@ const appendPipelineRunStatus = (pipeline, pipelineRun) => {
...
 };
-
+export const hasInlineTaskSpec = (pipeline: K8sResourceKind) => {
+  const tasks = pipeline?.spec.tasks ?? [];
+  let hasInlineSpec = false;
+  tasks.forEach((task) => {
+    if (!hasInlineSpec) {
+      hasInlineSpec = !!(task.taskSpec && !task.taskRef);
+    }
+  });
```
andrewballantyne, line 191: You're re-implementing `Array.prototype.some`... best just [use it instead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

karthikjeeyar: Updated it.

bgliwa01:
@karthikjeeyar I agree with @andrewballantyne we should remove the edit pipeline option in the kebab dropdown and then we can display the message "This Pipeline cannot be visualized. Pipeline taskSpec is not supported."

andrewballantyne:
> @karthikjeeyar I agree with @andrewballantyne we should remove the edit pipeline option in the kebab dropdown and then we can display the message "This Pipeline cannot be visualized. Pipeline taskSpec is not supported."

@karthikjeeyar When you do this, make sure there is a gap under the message before the first label.

Thanks @bgliwa01 !

karthikjeeyar:
Updated the code to hide the edit pipeline option and added the UX provided message to the visualization area.
 @bgliwa01 Please take a look at the updated screenshots.
cc: @andrewballantyne

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
```diff
@@ -17,21 +18,32 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
...
+      <Alert
+        variant="info"
+        isInline
+        title="This Pipeline cannot be visualized. Pipeline taskSpec is not supported."
+      />
+    );
+  } else if (nodes.length === 0 && edges.length === 0) {
     // Nothing to render
     // TODO: Confirm wording with UX; ODC-1860
-    return <Alert variant="info" isInline title="This Pipeline has no tasks to visualize." />;
```
andrewballantyne, line 24: Why did you remove the return? I think we can just have another return here.

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -181,7 +181,10 @@ const appendPipelineRunStatus = (pipeline, pipelineRun) => {
     return mTask;
   });
 };
-
+export const hasInlineTaskSpec = (pipeline: K8sResourceKind) => {
+  const tasks = pipeline?.spec.tasks ?? [];
```
andrewballantyne, line 185: `spec` is not optional.

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -192,7 +195,7 @@ export const getPipelineTasks = (
 ): PipelineVisualizationTaskItem[][] => {
   // Each unit in 'out' array is termed as stage | out = [stage1 = [task1], stage2 = [task2,task3], stage3 = [task4]]
   const out = [];
-  if (!pipeline.spec || !pipeline.spec.tasks) {
+  if (!pipeline?.spec.tasks) {
```
andrewballantyne, line 198: Don't think this can ever happen anymore. Tasks of 1 is required by the operator. We can keep this if you change it to look for tasks also being an empty array. 

Also `spec` is not optional.

andrewballantyne (changes requested):

frontend/packages/dev-console/src/utils/__tests__/pipeline-test-data.ts
```diff
@@ -12,7 +12,7 @@ export const mockPipelinesJSON: K8sResourceKind[] = [
       params: [
         {
           name: 'APP_NAME',
-          type: 'string',
```
andrewballantyne, line 15: `type` should be allowed 🤔  Why did you remove this?

andrewballantyne: Ah because it's not part of the type 😞 

Can you just add it to the type instead? 
It's [supported by upstream](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#specifying-parameters)
> Each declared parameter has a type field, which can be set to either array or string.

We'll want the type to probably be:

```
  type: 'string' | 'array';
```
Instead of a straight `string` type... this way we can easily see what we support.

bgliwa01 (approved):
lgtm

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
karthikjeeyar: I have moved it inside the block, so that it gets the margin-bottom style (for the gap underneath the message) from the `odc-pipeline-visualization` class.

andrewballantyne: Ah, I see... why don't we clean this up more and instead of treating errors differently, just have a variable called `content` or something and have 3 if-else statements... landing on rendering the Component as the fallback `else`.

It'll look cleaner and be easier to read.

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
karthikjeeyar: I agree, its more readable now. Updated it

andrewballantyne (approved):

frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
```diff
@@ -33,10 +33,8 @@ interface PipelineVisualizationTaskProp {
   namespace: string;
   task: {
     name?: string;
-    taskRef: {
-      name: string;
-      kind?: string;
-    };
+    taskSpec?: PipelineTaskSpec;
+    taskRef?: PipelineTaskRef;
```
jerolimov, line 37: :+1:

jerolimov (approved):
Tested it locally, works fine :+1: 

/lgtm

PR #6542, opened 2020-09-07
Bug 1876511: fixes issue with validation on pubsub modal

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4760

**Analysis / Root cause**: 
Subscription and trigger form Add button is disabled on opening the form if dropped on KSVC from channel/broker. Since required field is pre-populated the Add button should not be disabled as the user might not want to change anything and straightaway create the resource.

**Solution Description**: 
if name and subscriber name are there set dirty to true

Sep-07-2020 17-03-15

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (approved):
/lgtm

karthikjeeyar (approved):
Verified locally, works fine.

PR #6543, opened 2020-09-07
Bug 1876440: Cloud init section of common-templates is ignored

glekner:
image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
@@ -6,7 +6,7 @@ import {
   iGetRelevantTemplateSelectors,
   iGetVmSettingValue,
 } from '../../selectors/immutable/vm-settings';
-import { VMSettingsField, VMWizardProps } from '../../types';
+import { CloudInitField, VMSettingsField, VMWizardProps, VMWizardStorageType } from '../../types';
 import { InternalActionType, UpdateOptions } from '../types';
 import { asDisabled, asHidden, asRequired } from '../../utils/utils';
 import { vmWizardInternalActions } from '../internal-actions';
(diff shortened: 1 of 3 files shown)


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -115,6 +126,40 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
       { isHidden: asHidden(!isWindows, VMSettingsField.OPERATING_SYSTEM), value: isWindows },
     ),
   );
+
+  const relevantOptions = iGetRelevantTemplateSelectors(state, id);
```
suomiy, line 130: relevant template can change on relevenatOptions, not just os. IMO this needs its own updater and possibly new file called eg commonTemplatesPrefil

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -115,6 +126,40 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iTemplate =
+    iCommonTemplates && iGetRelevantTemplate(null, iCommonTemplates, relevantOptions);
+  const iCloudInitVolume = iGetCommonTemplateCloudInit(iTemplate);
+  const [data, isBase64] = CloudInitDataHelper.getUserData(
+    toShallowJS(iCloudInitVolume)?.cloudInitNoCloud,
+  );
+
```
suomiy, line 138: please take an inspiration from prefill-vm-template-state-update.ts

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -115,6 +126,40 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
...
+  if (data) {
+    dispatch(
+      vmWizardInternalActions[InternalActionType.SetCloudInitFieldValue](
+        id,
+        CloudInitField.IS_FORM,
+        false,
+      ),
+    );
+    dispatch(
```
suomiy, line 147: you need to take into account cloudinit from other templates.

suomiy: also please test this with a userTemplate

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
```diff
@@ -97,3 +98,8 @@ export const iGetRelevantTemplates = (
 export const iGetRelevantTemplate = (
   ...args: Parameters<typeof iGetRelevantTemplates>
 ): ITemplate => iGetRelevantTemplates(...args).first();
+
+export const iGetCommonTemplateCloudInit = (tmp: ITemplate) =>
```
suomiy, line 102: please make an immutable version of `selectVM`. Specifying an index is unreliable.

suomiy (changes requested):

yaacov:
Note:
When using this I get double "cloundinit" disks when I change the "OS" input, and then change it again ...

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
suomiy: as mentioned here https://bugzilla.redhat.com/show_bug.cgi?id=1874074.

We are going to need a generic approach to common template prefill soon.

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
glekner: created

glekner:
* Handling of cloudinit disk and volume now works fine when dynamically changing os,flavor,workload when using common templates.
* I noticed when using a user template, cloud init isn't imported? is this bug known? @jelkosz @suomiy @yaacov

yaacov:
> I noticed when using a user template, cloud init isn't imported? is this bug known? @jelkosz @suomiy @yaacov

you need to listen to changes in the user template, currently you only listen to common template fields.
but the user template issue can be tracked using a different bug report ( unless you want to fix it using this one ? ).

PR #6546, opened 2020-09-08
Bug 1876778: fixes issue with metrics not showing graph via topology

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4765

**Analysis / Root cause**: 
When the user clicks on the workload monitoring graph on the side panel of the topology it is redirected to the Monitoring Metrics page but the graph does not get rendered.

**Solution Description**: 
Added query as well id `useEffect` to patch if exists

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj (commented):
Verified it. The issue has been resolved. 👍 
Thanks @invincibleJai

vikram-raj (commented):
/lgtm

PR #6547, opened 2020-09-08
Bug 1876845: List and use Projects instead of Namespaces in GitOps list page

divyanshiGupta:
Jira: https://issues.redhat.com/browse/ODC-4766

**Screenshots:**
image
image

/kind bug


rohitkrai03 (approved):
/lgtm

PR #6548, opened 2020-09-08
Bug 1848957: Removes the objects count from buckets card

afreen23:
- Fixes partially https://bugzilla.redhat.com/show_bug.cgi?id=1848957
- Fix screenshot
Screenshot from 2020-09-08 15-45-22


bipuladh (changes requested):
let's make it conditional on the non-existence of the RGW flag. If RGW is present then only we will hide it. Else there is no harm in showing it (For ex: Cloud env's where RGW will not be deployed)

PR #5767, opened 2020-06-17
filter incompatible helm charts from dev catalog

debsmita1:
**JIRA Story:**
https://issues.redhat.com/browse/ODC-4062

**Solution Description:**
- filters the helm charts in the dev-catalog that are incompatible with the Kubernetes Version
- filters helm chart versions in the Chart Version dropdown that are incompatible with the Kubernetes Version
- Fixed the type of `HelmChart` (Referred the Helm docs: https://helm.sh/docs/topics/charts/ , section : `The Chart.yaml File` )
- Added tests
test

PR #6550, opened 2020-09-08
Bug 1876741: undo filtering of incompatible helm charts

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-4746
reverts changes made in https://github.com/openshift/console/pull/5767

**Root Analysis:**
- version range `>=1.10.11` is not compatible with pre-released versions like `v1.10.11-rc.2+068702d`
- version range `>=1.10.11-0` only accepts prerelease tags on the version 1.10.11 like `v1.10.11-rc.2+068702d` & not `v1.17.3-rc.2+0687785`

Before Change:
before

After Change:
after

### DIFF
--- a/frontend/__mocks__/catalogItemsMocks.ts
+++ b/frontend/__mocks__/catalogItemsMocks.ts
@@ -1,6 +1,5 @@
 export const catalogListPageProps = {
   namespace: 'default',
-  kubernetesVersion: 'v1.18.0+ui896hui',
   helmCharts: {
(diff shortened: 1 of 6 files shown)


abhinandan13jan:
just a reminder. We do need a bug for this right?

nemesis09:
tested locally. charts show up as expected.
/lgtm

rohitkrai03 (approved):
/approve

PR #6552, opened 2020-09-08
[release-4.5.z] Bug 1876461: remove the import VM from dev perspective

suomiy:


markmc:
Low risk removal to improve non-admin UX

PR #6554, opened 2020-09-08
Bug 1871104: PXE requires NADs

yaacov:
On a cluster without a newtwork attachment definition, it is not possible to use a PXE boot. Yet the user can pick it and get lost.

The ask is to provide a warning right in the first screen blocking the user from continuing to the next screen. The warning needs to explain that a new network attachment definition needs to be created in this namespace and that the administrator needs to be contacted to do it.

Screenshot:
Peek 2020-09-21 17-02

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
@@ -12,6 +12,8 @@ import {
 import { TemplateModel } from '@console/internal/models';
 import { Firehose, history } from '@console/internal/components/utils';
 import { usePrevious } from '@console/shared/src/hooks/previous';
+import { PersistentVolumeClaimKind, referenceForModel } from '@console/internal/module/k8s';
+import { NetworkAttachmentDefinitionModel } from '@console/network-attachment-definition-plugin';
 import { Location } from 'history';
 import { match as RouterMatch } from 'react-router';
 import { withReduxID } from '../../utils/redux/common';
@@ -63,7 +65,6 @@ import { ValidTabGuard } from './tabs/valid-tab-guard';
 import { FirehoseResourceEnhanced } from '../../types/custom';
 
 import './create-vm-wizard.scss';
-import { PersistentVolumeClaimKind } from '@console/internal/module/k8s';
 
 type CreateVMWizardComponentProps = {
   isSimpleView: boolean;
@@ -407,6 +408,19 @@ export const CreateVMWizardPageComponent: React.FC<CreateVMWizardPageComponentPr
         prop: VMWizardProps.dataVolumes,
       }),
     ];
+
+    if (userMode !== VMWizardMode.IMPORT) {
+      resources.push({
+        kind: referenceForModel(NetworkAttachmentDefinitionModel),
+        model: NetworkAttachmentDefinitionModel,
+        isList: true,
+        namespace: activeNamespace,
+        prop: VMWizardProps.nads,
+        errorBehaviour: {
+          ignore404: true,
+        },
+      });
+    }
 
     if (flags[FLAGS.OPENSHIFT]) {
       resources.push(
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
@@ -16,7 +16,12 @@ import {
   VIRTUAL_MACHINE_TEMPLATE_EXISTS,
 } from '../../../../utils/validations/strings';
 import { getFieldTitle } from '../../utils/renderable-field-utils';
-import { iGet, immutableListToJS } from '../../../../utils/immutable';
+import {
+  iGet,
+  iGetLoadedData,
+  iGetLoadError,
+  immutableListToJS,
+} from '../../../../utils/immutable';
 import {
   checkTabValidityChanged,
   iGetCommonData,
@@ -29,6 +34,11 @@ import { combineIntegerValidationResults } from '../../../../utils/validations/t
 import { getFieldsValidity, getValidationUpdate } from './utils';
 import { getTemplateValidations } from '../../selectors/template';
 import { BinaryUnit, convertToBytes } from '../../../form/size-unit-utils';
+import {
+  VALIDATION_PXE_NAD_ERROR_INFO,
+  VALIDATION_PXE_NAD_MISSING_INFO,
+} from '../../strings/networking';
+import { ProvisionSource } from '../../../../constants/vm/provision-source';
 import { UIValidation } from '../../../../types/ui/ui';
 
 const validateVm: Validator = (field, options) => {
@@ -110,6 +120,24 @@ const asVMSettingsFieldValidator = (
     subject: getFieldTitle(iGetFieldKey(field)),
   });
 
+const validateSource: Validator = (field, options): ValidationObject => {
+  const value = iGetFieldValue(field);
+
+  if (value === ProvisionSource.PXE.getValue()) {
+    const { getState, id } = options;
+    const state = getState();
+
+    const nads = iGetCommonData(state, id, VMWizardProps.nads);
+    return iGetLoadError(nads)
+      ? asValidationObject(VALIDATION_PXE_NAD_ERROR_INFO)
+      : iGetLoadedData(nads)?.size === 0
+      ? asValidationObject(VALIDATION_PXE_NAD_MISSING_INFO)
+      : null;
+  }
+
+  return null;
+};
+
 const validationConfig: ValidationConfig = {
   [VMSettingsField.NAME]: {
     detectValueChanges: [VMSettingsField.NAME],
@@ -141,6 +169,10 @@ const validationConfig: ValidationConfig = {
     ],
     detectCommonDataChanges: [VMWizardProps.userTemplate, VMWizardProps.commonTemplates],
     validator: memoryValidation,
+  },
+  [VMSettingsField.PROVISION_SOURCE_TYPE]: {
+    detectValueChanges: [VMSettingsField.PROVISION_SOURCE_TYPE],
+    validator: validateSource,
   },
 };
 
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/networking.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/networking.ts
@@ -1,3 +1,11 @@
 export const PXE_NIC_NOT_FOUND_ERROR = 'A PXE-capable network interface could not be found.';
 export const PXE_INFO = 'Pod network is not PXE bootable';
 export const SELECT_PXE_NIC = '--- Select PXE network interface ---';
+export const VALIDATION_PXE_NAD_ERROR_INFO =
+  'Error fetching available Network Attachment Definitions. PXE capable Network Attachment Definition is required to successfully create this virtual machine. Contact your system administrator for additional support.';
+export const VALIDATION_PXE_NAD_MISSING_INFO =
+  'No Network Attachment Definitions available. PXE capable Network Attachment Definition is required to successfully create this virtual machine. Contact your system administrator for additional support.';
+export const SELECT_PXE_NAD_ERROR_INFO =
+  'Error fetching available Network Attachment Definitions. Contact your system administrator for additional support.';
+export const SELECT_PXE_NAD_MISSING_INFO =
+  'No Network Attachment Definitions available. Contact your system administrator for additional support.';
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
@@ -5,8 +5,9 @@ import { FormFieldRow } from '../../form/form-field-row';
 import { FormField, FormFieldType } from '../../form/form-field';
 import { iGetFieldValue } from '../../selectors/immutable/field';
 import { VMSettingsField } from '../../types';
-import { iGet } from '../../../../utils/immutable';
+import { iGet, iGetIn } from '../../../../utils/immutable';
 import { FormPFSelect } from '../../../form/form-pf-select';
+import { ValidationErrorType } from '@console/shared';
 
 const ProvisionSourceDiskHelpMsg: React.FC<ProvisionSourceDiskHelpMsgProps> = ({
   provisionSourceValue,
@@ -79,6 +80,7 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
   ({ provisionSourceField, onChange, goToStorageStep, goToNetworkingStep }) => {
     const provisionSourceValue = iGetFieldValue(provisionSourceField);
     const sources = iGet(provisionSourceField, 'sources');
+    const validationType = iGetIn(provisionSourceField, ['validation', 'type']);
 
     return (
       <FormFieldRow field={provisionSourceField} fieldType={FormFieldType.PF_SELECT}>
@@ -112,9 +114,10 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
             goToStorageStep={goToStorageStep}
           />
         )}
-        {[ProvisionSource.PXE.getValue()].includes(provisionSourceValue) && (
-          <ProvisionSourceNetHelpMsg goToNetworkingStep={goToNetworkingStep} />
-        )}
+        {[ProvisionSource.PXE.getValue()].includes(provisionSourceValue) &&
+          validationType !== ValidationErrorType.Error && (
+            <ProvisionSourceNetHelpMsg goToNetworkingStep={goToNetworkingStep} />
+          )}
       </FormFieldRow>
     );
   },
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
@@ -33,6 +33,7 @@ export enum VMWizardProps {
   dataVolumes = 'dataVolumes',
   openshiftCNVBaseImages = 'openshiftCNVBaseImages',
   storageClassConfigMap = 'storageClassConfigMap',
+  nads = 'nads',
 }
 
 // order important
@@ -216,6 +217,7 @@ export type ChangedCommonDataProp =
   | VMWizardProps.dataVolumes
   | VMWizardProps.openshiftCNVBaseImages
   | VMWizardProps.storageClassConfigMap
+  | VMWizardProps.nads
   | VMWareProviderProps.deployment
   | VMWareProviderProps.deploymentPods
   | VMWareProviderProps.v2vvmware
@@ -246,6 +248,7 @@ export const DetectCommonDataChanges = new Set<ChangedCommonDataProp>([
   VMWizardProps.storageClassConfigMap,
   VMWizardProps.dataVolumes,
   VMWizardProps.openshiftCNVBaseImages,
+  VMWizardProps.nads,
   VMWareProviderProps.deployment,
   VMWareProviderProps.deploymentPods,
   VMWareProviderProps.v2vvmware,
--- a/frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
@@ -41,6 +41,10 @@ import { K8sResourceKind } from '@console/internal/module/k8s';
 import { PendingChangesAlert } from '../../Alerts/PendingChangesAlert';
 import { MODAL_RESTART_IS_REQUIRED } from '../../../strings/vm/status';
 import { FormPFSelect } from '../../form/form-pf-select';
+import {
+  SELECT_PXE_NAD_ERROR_INFO,
+  SELECT_PXE_NAD_MISSING_INFO,
+} from '../../create-vm-wizard/strings/networking';
 
 const getNetworkChoices = (
   nads: K8sResourceKind[],
@@ -96,15 +100,22 @@ export const Network: React.FC<NetworkProps> = ({
     allowPodNetwork,
     allowedMultusNetworkTypes,
   ).filter((n) => n.getType().isSupported());
+  const validationMessage = nadsLoadError
+    ? SELECT_PXE_NAD_ERROR_INFO
+    : networkChoices.length === 0
+    ? SELECT_PXE_NAD_MISSING_INFO
+    : null;
+  const validationType =
+    nadsLoadError || networkChoices.length === 0 ? ValidationErrorType.Error : null;
 
   return (
     <FormRow
       title="Network"
       fieldId={id}
       isRequired
-      isLoading={nadsLoading}
-      validationMessage={nadsLoadError}
-      validationType={nadsLoadError && ValidationErrorType.Error}
+      isLoading={nadsLoading && !nadsLoadError}
+      validationMessage={validationMessage}
+      validationType={validationType}
     >
       <FormSelect
         onChange={(net, event) => {
@@ -116,15 +127,11 @@ export const Network: React.FC<NetworkProps> = ({
         }}
         value={asFormSelectValue(network?.getReadableName())}
         id={id}
-        isDisabled={isDisabled || nadsLoading || nadsLoadError}
+        isDisabled={isDisabled || nadsLoading || nadsLoadError || networkChoices.length === 0}
       >
         <FormSelectPlaceholderOption
           isDisabled={!acceptEmptyValues}
-          placeholder={
-            networkChoices.length === 0
-              ? '--- No Network Attachment Definitions available ---'
-              : '--- Select Network Attachment Definition ---'
-          }
+          placeholder={'--- Select Network Attachment Definition ---'}
         />
         {ignoreCaseSort(networkChoices, undefined, (n) => n.getReadableName()).map(
           (networkWrapper: NetworkWrapper) => {


yaacov:
@rawagner @suomiy @glekner @irosenzw please review

yaacov:
@matthewcarleton hi,
Notes:
a - the dropdown will change once https://github.com/openshift/console/pull/6433 will be merged
     [ @glekner - I tried to minimize the conflicts ... ]
b - the error message in the bottom is different then the design because it requires re-factoring of the footer I think is bigger then a bug fix:
     1. the title is generic
     2. no link to "more info"

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -77,9 +94,18 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
...
             />
-            {(sources || []).map((source) => (
-              <FormSelectOption key={source} value={source} label={source} />
-            ))}
+            {(sources || []).map((source) => {
+              if (source === ProvisionSource.PXE.toString() && !hasNADs) {
+                return (
+                  <FormSelectOption
+                    key={source}
+                    value={SELECT_PXE_NAD_ERROR}
```
suomiy, line 102: the value should be an enum and is not for storing errors. This can get mishandled pretty easily in the redux store. Especially with changing requirements.

This will also not work well with the template flow.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -144,6 +145,14 @@ const asVMSettingsFieldValidator = (
     subject: getFieldTitle(iGetFieldKey(field)),
   });
 
+const validateSource = (value: string): ValidationObject => {
```
suomiy, line 148: it should all be validated here. This is the correct place in the wizard lifecycle where to set it. All logic depends on this behaviour.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/utils.ts
```diff
@@ -76,5 +80,12 @@ export const getFieldsValidity = (fields) => {
     error = describeFields('Please correct', invalidFields);
   }
   const isValid = hasAllRequiredFilled && invalidFields.length === 0;
+
+  // check for missing network attachment definition error
+  const invalidFieldValues = getInvalidFieldValues(invalidFields);
```
suomiy, line 85: this does not feel right

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -64,11 +70,22 @@ const ProvisionSourceNetHelpMsg: React.FC<ProvisionSourceNetHelpMsgProps> = ({
...
 };
 
-export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> = React.memo(
-  ({ provisionSourceField, onChange, goToStorageStep, goToNetworkingStep }) => {
+export const ProvisionSourceComponentWithNamespace: React.FC<ProvisionSourceComponentProps> = React.memo(
+  ({ provisionSourceField, onChange, goToStorageStep, goToNetworkingStep, activeNamespace }) => {
     const provisionSourceValue = iGetFieldValue(provisionSourceField);
     const sources = iGet(provisionSourceField, 'sources');
 
+    const networkAttachmentDefinitionModelResource = {
```
suomiy, line 78: let's fetch this wizard wide. We should also double check the permissions, so that the wizard is usable for non-admin users

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -114,4 +140,15 @@ type ProvisionSourceComponentProps = {
   onChange: (key: string, value: string | boolean) => void;
   goToStorageStep: () => void;
   goToNetworkingStep: () => void;
+  activeNamespace: string;
 };
+
+const mapStateToProps = (state: RootState): { activeNamespace: string } => {
+  return {
+    activeNamespace: state.UI.get('activeNamespace'),
```
suomiy, line 148: just a note for the future: we have our own variable in the redux which might be subjected to change

suomiy (changes requested):

matthewcarleton:
> @matthewcarleton hi,
> Notes:
> a - the dropdown will change once #6433 will be merged
> [ @glekner - I tried to minimize the conflicts ... ]
Yes we'll have to review this in the new dropdown to make sure it looks good with the PXE description
> b - the error message in the bottom is different then the design because it requires re-factoring of the footer I think is bigger then a bug fix:
That's fine
> 1. the title is generic
What is the reason for this? 
> 2. no link to "more info"
Can't link to docs?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/networking.ts
```diff
@@ -1,3 +1,9 @@
 export const PXE_NIC_NOT_FOUND_ERROR = 'A PXE-capable network interface could not be found.';
+export const PXE_NAD_NOT_FOUND_ERROR =
+  'A PXE-capable network attachment definition could not be found.';
+export const PXE_NAD_NOT_FOUND_INFO = '(Requires PXE-capable network attachment definition)';
 export const PXE_INFO = 'Pod network is not PXE bootable';
 export const SELECT_PXE_NIC = '--- Select PXE network interface ---';
+export const SELECT_PXE_NAD_ERROR = 'PXE-NAD-ERROR';
+export const SELECT_PXE_NAD_ERROR_INFO =
+  'Before you can successfuly create this virtual machine a network attachment definition must be created. Contact your system admin for aditional support';
```
jelkosz, line 9: Please replace "admin" by "administrator".

matthewcarleton, line 9:
```suggestion
  'Before you can successfully create this virtual machine a network attachment definition must be created. Contact your system admin for additional support';
```
spelling :)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
yaacov: not changing the value now

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
yaacov: logic moved to validation

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/utils.ts
yaacov: not changing the value now

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
yaacov: fetching on create wizard level

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
yaacov: using a function now

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/networking.ts
yaacov: :palm_tree: :haircut: thanks :+1: fixed.

yaacov:
@suomiy hi, back with the PXE :-) please re-review

matthewcarleton:
For the Error text:
`PXE capable Network Attachment definition could not be found. This is required to successfully create this virtual machine. Contact your system administrator for additional support.`
A bit more concise - WDYT?

yaacov:
> `PXE capable Network Attachment definition could not be found. This is required to successfully create this virtual machine. Contact your system administrator for additional support.` WDYT?

@matthewcarleton Thanks, updated the text :+1:

yaacov:
@suomiy please re-review

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -90,6 +94,21 @@ const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props)
   const [dataVolumePVCs, dataVolumePVCsLoaded, dataVolumePVCsLoadError] = usePVCBaseImages(
     commonTemplates,
   );
+  const activeNamespace = getActiveNamespace();
+  const nadsResource = React.useMemo(
+    () => ({
```
suomiy, line 99: this is not a place for listening to resources as it is inefficient and adds more complexity. Please add a nads resource to CreateVMWizardPageComponent.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -154,6 +181,19 @@ const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props)
           },
         };
       }
+      if (nadsChanged) {
```
suomiy, line 184: even for very custom data this solution is not scalable: this PR fixes https://github.com/openshift/console/pull/6704 that. But still I added a TODO there, so we can come up with more efficient solution later for the 2 resources which use this approach.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
```diff
@@ -33,6 +34,7 @@ export enum VMWizardProps {
   dataVolumes = 'dataVolumes',
   openshiftCNVBaseImages = 'openshiftCNVBaseImages',
   storageClassConfigMap = 'storageClassConfigMap',
+  nadsData = 'nadsData',
```
suomiy, line 37: why not just nads?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -108,6 +115,27 @@ const asVMSettingsFieldValidator = (
...
+  if (emptyError) {
+    return emptyError;
+  }
+
+  if (value === ProvisionSource.PXE.getValue()) {
+    const { getState, id } = options;
+    const state = getState();
+    const nads = iGetCommonData(state, id, VMWizardProps.nadsData);
+    const hasNADs =
+      !iGetLoadError(nads) && immutableListToShallowJS(iGetLoadedData(nads))?.length !== 0;
```
suomiy, line 130: you can find the length without converting the whole array to JS

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/types.ts
```diff
@@ -274,6 +278,11 @@ export type CommonData = {
       loadError: string;
       data: PersistentVolumeClaimKind[];
     };
+    [VMWizardProps.nadsData]?: {
```
suomiy, line 281: we should not overuse data portion of redux  and have just one source of data. We should deprecate this in the future.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
suomiy: does this take into account loading?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -108,6 +115,27 @@ const asVMSettingsFieldValidator = (
     subject: getFieldTitle(iGetFieldKey(field)),
   });
 
+const validateSource: Validator = (field, options): ValidationObject => {
+  const value = iGetFieldValue(field);
+  const emptyError = validateEmptyValue(value);
```
suomiy, line 120: we don't need this clause - it will get handled by setVmSettingsTabValidity

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/provision-source.tsx
```diff
@@ -112,7 +120,7 @@ export const ProvisionSourceComponent: React.FC<ProvisionSourceComponentProps> =
             goToStorageStep={goToStorageStep}
           />
         )}
-        {[ProvisionSource.PXE.getValue()].includes(provisionSourceValue) && (
+        {[ProvisionSource.PXE.getValue()].includes(provisionSourceValue) && hasNADs && (
```
suomiy, line 123: we don't need the nads here. We can just check if the field has an error

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
```diff
@@ -102,9 +102,13 @@ export const Network: React.FC<NetworkProps> = ({
       title="Network"
       fieldId={id}
       isRequired
-      isLoading={nadsLoading}
-      validationMessage={nadsLoadError}
-      validationType={nadsLoadError && ValidationErrorType.Error}
+      isLoading={nadsLoading && !nadsLoadError}
+      validationMessage={
+        nadsLoadError || networkChoices.length === 0
+          ? 'No Network Attachment Definitions available. Contact your system admin for additional support'
```
suomiy, line 108: this should check the nadsLoadError properly to display such a message.

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
```diff
@@ -102,9 +102,13 @@ export const Network: React.FC<NetworkProps> = ({
...
+          : null
+      }
+      validationType={nadsLoadError || (networkChoices.length === 0 && ValidationErrorType.Error)}
```
suomiy, line 111: nadsLoadError is not of validationType type

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
yaacov: that is the bug we are fixing, we do not want to show the raw error message, we just want to inform users that they don't have nads.
An error fetching nads may be ok here (e.g. if they don't have credentials), or not (e.g. api down), and we just want to inform them that they can't use nads, and if they think they should, they can investigate ask for admin help,

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
yaacov: > does this take into account loading?

yes, `iGetLoadedData` checks for `loading` internally so we can't get a false positive while loading.

suomiy: this will hapen when you use a template with PXE and nads have not loaded yet

aa

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
suomiy: I think it might be useful to give a small clue what is wrong.

- If it is a permission error - they know that they have to contact an admin
- if nads are empty. They can just create it themselves and they dont have to bother admin.

cc @matthewcarleton thoughts?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
yaacov: are you saying we should not validate fields field in by templates ?

yaacov:
@suomiy all comments addressed please re review

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
matthewcarleton: +1 actionable information that is accurate is the goal so we should def provide them a path to fix the issue (especially if they can do it themselves)
If they can create one we should say 
`create a Network attachment definition` is an action we should provide from the alert rather than telling them to talk to the admin if they can.

yaacov: @matthewcarleton their are two different texts now, one for error and one for no NADs:
https://github.com/openshift/console/pull/6554/files#diff-1c2a6b849bb9ba8b02f00ccb4ee98149R99

yaacov: @matthewcarleton 
p.s. even if the problem is not missing credentials and user can create a NAD they probably need admin help because they need to know the name of the network bridge to use.

matthewcarleton: Is it possible for us to provide that info to them?

yaacov: > Is it possible for us to provide that info to them?

we can change the 2 texts (for error and not no NADs), what do you want to change  ?

matthewcarleton: well if you are saying for the user who has the permissions but won't know which bridge to choose I'm asking can we tell them which network bridge to use?

yaacov: > well if you are saying for the user who has the permissions but won't know which bridge to choose I'm asking can we tell them which network bridge to use?

Sure, they should ask the cluster administrator, AFAIU this information is not available o/w

matthewcarleton: just to clarify are you talking about network type?
Screen Shot 2020-09-22 at 3 43 54 PM

yaacov: @matthewcarleton no, I'm talking about bridge name (I can't see it in your screenshot)

To create a NAD you need to know about the BM host network configuration
https://docs.openshift.com/container-platform/4.3/networking/multiple_networks/configuring-host-device.html
In all the docs it's allways "As a cluster administrator, you can ..."

matthewcarleton: ah gotcha - it sounds like we should be looking at improving that flow as well then. We should provide help to understand more about what Bridge name is if it's required. Out of scope here but relevant.

yaacov: +1 , I understand Yifat is working on QS for how to create a network

matthewcarleton: true - hopefully we can look at just generally improving that modal too and not just rely on the quick start.

matthewcarleton: We should state you may need to contact your cluster admin for the bridge name

yaacov: @matthewcarleton  We don't know what information is missing, it may be the bridge name or IP's in the network, host device names ... see: https://docs.openshift.com/container-platform/4.3/networking/multiple_networks/understanding-multiple-networks.html

The help texts say "... Contact your system admin for more information" , IMHO it informs the user, that if they don't know how to add a NAD without help, they probably need to contact their cluster administrator for help.

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
```diff
@@ -96,15 +96,22 @@ export const Network: React.FC<NetworkProps> = ({
     allowPodNetwork,
     allowedMultusNetworkTypes,
   ).filter((n) => n.getType().isSupported());
+  const validationMessage = nadsLoadError
+    ? 'Error fatching available Network Attachment Definitions. Contact your system admin for additional support'
```
matthewcarleton, line 100: I think we are saying administrator right?
```suggestion
    ? 'Error fetching available Network Attachment Definitions. Contact your system administrator for additional support'
```

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
```diff
@@ -96,15 +96,22 @@ export const Network: React.FC<NetworkProps> = ({
...
+    : networkChoices.length === 0
+    ? 'No Network Attachment Definitions available. Contact your system admin for more information'
```
matthewcarleton, line 102:
```suggestion
    ? 'No Network Attachment Definitions available. Contact your system administrator for additional  support'
```

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
matthewcarleton: good point. I like "Contact your system administrator for additional support" 
Sounds more helpful :)

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
yaacov: +1 Thanks :exclamation:

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
yaacov: Thanks, updated the two texts :+1:

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -406,6 +407,16 @@ export const CreateVMWizardPageComponent: React.FC<CreateVMWizardPageComponentPr
         namespace: activeNamespace,
         prop: VMWizardProps.dataVolumes,
       }),
+      {
```
suomiy, line 410: nads do not have to be loaded for vm import

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -108,6 +110,27 @@ const asVMSettingsFieldValidator = (
...
   });
 
+const validateSource: Validator = (field, options): ValidationObject => {
+  const value = iGetFieldValue(field);
+
+  if (value === ProvisionSource.PXE.getValue()) {
+    const { getState, id } = options;
+    const state = getState();
+    const isCreateTemplate = iGetCommonData(state, id, VMWizardProps.isCreateTemplate);
+    if (isCreateTemplate) {
```
suomiy, line 120: why is this validation not required for templates?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
```diff
@@ -108,6 +110,27 @@ const asVMSettingsFieldValidator = (
...
+      return null;
+    }
+
+    const nads = iGetCommonData(state, id, VMWizardProps.nads);
+    const missingNADs = iGetLoadError(nads) || iGetLoadedData(nads)?.size === 0;
+    if (missingNADs) {
+      return asValidationObject(SELECT_PXE_NAD_ERROR_INFO);
```
suomiy, line 127: can we use the similar messages and logic as in the nic modal? A user choosing a PXE will not be able to navigate to the nic modal once s/he sees this error. And will not get get the more detailed message.

frontend/packages/kubevirt-plugin/src/components/modals/nic-modal/nic-modal.tsx
```diff
@@ -116,15 +123,11 @@ export const Network: React.FC<NetworkProps> = ({
...
+        isDisabled={isDisabled || nadsLoading || nadsLoadError || networkChoices.length === 0}
       >
         <FormSelectPlaceholderOption
           isDisabled={!acceptEmptyValues}
-          placeholder={
-            networkChoices.length === 0
-              ? '--- No Network Attachment Definitions available ---'
-              : '--- Select Network Attachment Definition ---'
-          }
+          placeholder={'--- Select Network Attachment Definition ---'}
```
suomiy, line 130: +1 better

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
yaacov: no nads for import :-)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
yaacov: validating templates too

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/validations/vm-settings-tab-validation.ts
yaacov: sure

yaacov:
@suomiy  all comments adressed, please re-review

yaacov:
@suomiy please re-review

PR #6556, opened 2020-09-08
Bug 1873326: Adding accessibility alt and title to icons within overview status and activity sections

sg00dwin:
Currently `aria-hidden="true"` are on these icons, a label isn't required since it's not accessible to a screen reader. Including `alt` attribute to be consistent with https://github.com/openshift/console/pull/6497#issuecomment-686546767
Fix bug https://bugzilla.redhat.com/show_bug.cgi?id=1873326

### DIFF
--- a/frontend/packages/console-shared/src/components/dashboard/activity-card/EventItem.tsx
+++ b/frontend/packages/console-shared/src/components/dashboard/activity-card/EventItem.tsx
@@ -40,7 +40,11 @@ const EventItem: React.FC<EventItemProps> = React.memo(({ event, isExpanded, onT
             </div>
             <div className="co-recent-item__title-message">
               {isWarning && (
-                <YellowExclamationTriangleIcon className="co-dashboard-icon co-recent-item__icon--warning" />
+                <YellowExclamationTriangleIcon
+                  alt="Warning"
+                  title="Warning"
+                  className="co-dashboard-icon co-recent-item__icon--warning"
+                />
               )}
               {!expanded && (
                 <>
--- a/frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
+++ b/frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
@@ -14,13 +14,15 @@ import {
   getAlertTime,
 } from './alert-utils';
 
+const CriticalIcon = () => <RedExclamationCircleIcon alt="Critical" title="Critical" />;
+const WarningIcon = () => <YellowExclamationTriangleIcon alt="Warning" title="Warning" />;
 const getSeverityIcon = (severity: string) => {
   switch (severity) {
     case 'critical':
-      return RedExclamationCircleIcon;
+      return CriticalIcon;
     case 'warning':
     default:
-      return YellowExclamationTriangleIcon;
+      return WarningIcon;
   }
 };
 
--- a/frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
+++ b/frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
   [HealthState.OK]: {
     priority: 0,
     health: HealthState.OK,
-    icon: <GreenCheckCircleIcon />,
+    icon: <GreenCheckCircleIcon alt="Healthy" title="Healthy" />,
   },
   [HealthState.UNKNOWN]: {
     priority: 1,
     health: HealthState.UNKNOWN,
-    icon: <GrayUnknownIcon />,
+    icon: <GrayUnknownIcon alt="Unknown" title="Unknown" />,
     message: 'Unknown',
   },
   [HealthState.PROGRESS]: {
     priority: 2,
     health: HealthState.PROGRESS,
-    icon: <InProgressIcon />,
+    icon: <InProgressIcon alt="In progress" title="In progress" />,
     message: 'Pending',
   },
   [HealthState.UPDATING]: {
     priority: 3,
     health: HealthState.UPDATING,
-    icon: <BlueSyncIcon />,
+    icon: <BlueSyncIcon alt="Updating" title="Updating" />,
     message: 'Updating',
   },
   [HealthState.WARNING]: {
     priority: 4,
     health: HealthState.WARNING,
-    icon: <YellowExclamationTriangleIcon />,
+    icon: <YellowExclamationTriangleIcon alt="Warning" title="Warning" />,
     message: 'Degraded',
   },
   [HealthState.ERROR]: {
     priority: 5,
     health: HealthState.ERROR,
-    icon: <RedExclamationCircleIcon />,
+    icon: <RedExclamationCircleIcon alt="Error" title="Error" />,
     message: 'Degraded',
   },
   [HealthState.LOADING]: {
@@ -64,7 +64,7 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
   [HealthState.NOT_AVAILABLE]: {
     priority: 7,
     health: HealthState.NOT_AVAILABLE,
-    icon: <GrayUnknownIcon />,
+    icon: <GrayUnknownIcon alt="Not available" title="Not available" />,
     message: 'Not available',
   },
 };
--- a/frontend/packages/console-shared/src/components/status/icons.tsx
+++ b/frontend/packages/console-shared/src/components/status/icons.tsx
@@ -17,48 +17,87 @@ import { global_palette_blue_300 as blueInfoColor } from '@patternfly/react-toke
 import { global_palette_green_500 as okColor } from '@patternfly/react-tokens/dist/js/global_palette_green_500';
 import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens/dist/js/global_warning_color_100';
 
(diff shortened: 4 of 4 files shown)


spadgett (changes requested):
@sg00dwin We should only use `alt=""` if there is other text in the UI describing the icon status. If the only status is the icon, we need a real value for `alt`.

cc @jessiehuff

sg00dwin:
@spadgett included descriptive text for health states and status icons

jessiehuff (commented):
So this one took a little more thought because these icons seem to have a function. Since these icons act as a communication of the status, it seems that this information should be communicated to assistive technology. Right now, because it has `aria-hidden` on it, nothing will be read by the screen reader. The reason that gets added is in the [create icon file](https://github.com/patternfly/patternfly-react/blob/master/packages/react-icons/src/createIcon.tsx), when a title is not provided, the `aria-hidden` is added since it's assumed to be decorative. We can keep the alt, but I suggest adding a title as well. This will remove the `aria-hidden` and allow for the icon to communicate that status information. I also suggest taking off the "icon" text. It will then read out in VO as "Ok health state, image". If we keep the icon text, it would read as "Ok health state icon, image" which is a little unnecessary. However, to add this, I believe you'll need to add title to the icons.tsx file. So essentially I recommend something like this `icon: <GreenCheckCircleIcon alt="Ok health state" title="Ok health state" />` I think we could continue to investigate if there's a better way to associate these items, but I think adding that title alone would be an improvement.

sg00dwin:
Updated pr to include `title` with `svg`

frontend/packages/console-shared/src/components/dashboard/activity-card/EventItem.tsx
```diff
@@ -40,7 +40,11 @@ const EventItem: React.FC<EventItemProps> = React.memo(({ event, isExpanded, onT
             </div>
             <div className="co-recent-item__title-message">
               {isWarning && (
-                <YellowExclamationTriangleIcon className="co-dashboard-icon co-recent-item__icon--warning" />
+                <YellowExclamationTriangleIcon
+                  alt="Warning alert"
+                  title="Warning alert"
```
spadgett, line 45: These are events, not alerts. But I feel like the noun is unnecessarily verbose. Can this just be "Warning"? @jessiehuff ?

```suggestion
                  alt="Warning"
                  title="Warning"
```

If @jessiehuff disagrees, we'd need to change "alert" to "event."

frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
```diff
@@ -28,7 +28,7 @@ export const StatusItem: React.FC<StatusItemProps> = ({ Icon, timestamp, message
   return (
     <div className="co-status-card__alert-item">
       <div className="co-status-card__alert-item-icon co-dashboard-icon">
-        <Icon />
+        <Icon alt="Status alert" title="Status alert" />
```
spadgett, line 31: We need to put a real status in the alt here for whatever icon is passed. I believe you can move this into `getSeverityIcon` above.

```tsx
const CriticalIcon = <RedExclamationCircleIcon alt="Critical" title="Critical" />;
const WarningIcon = <YellowExclamationTriangleIcon alt="Warning" title="Warning" />;
const getSeverityIcon = (severity: string) => {
  switch (severity) {
    case 'critical':
      return CriticalIcon;
    case 'warning':
    default:
      return WarningIcon;
  }
};
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
...
     health: HealthState.UNKNOWN,
-    icon: <GrayUnknownIcon />,
+    icon: <GrayUnknownIcon alt="Unknown health state" title="Unknown health state" />,
     message: 'Unknown',
   },
   [HealthState.PROGRESS]: {
     priority: 2,
     health: HealthState.PROGRESS,
-    icon: <InProgressIcon />,
+    icon: <InProgressIcon alt="Progress health state" title="Progress health state" />,
```
spadgett, line 37:
```suggestion
    icon: <InProgressIcon alt="In progress" title="In progress" />,
```

spadgett:
@jessiehuff Instead of "Ok health state," would it be better to use something like "Healthy"?

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
   [HealthState.OK]: {
     priority: 0,
     health: HealthState.OK,
-    icon: <GreenCheckCircleIcon />,
+    icon: <GreenCheckCircleIcon alt="Ok health state" title="Ok health state" />,
```
spadgett, line 26:
```suggestion
    icon: <GreenCheckCircleIcon alt="Healthy" title="Healthy" />,
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
...
   },
   [HealthState.UNKNOWN]: {
     priority: 1,
     health: HealthState.UNKNOWN,
-    icon: <GrayUnknownIcon />,
+    icon: <GrayUnknownIcon alt="Unknown health state" title="Unknown health state" />,
```
spadgett, line 31:
```suggestion
    icon: <GrayUnknownIcon alt="Unknown status" title="Unknown status" />,
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
...
     health: HealthState.PROGRESS,
-    icon: <InProgressIcon />,
+    icon: <InProgressIcon alt="Progress health state" title="Progress health state" />,
     message: 'Pending',
   },
   [HealthState.UPDATING]: {
     priority: 3,
     health: HealthState.UPDATING,
-    icon: <BlueSyncIcon />,
+    icon: <BlueSyncIcon alt="Updating health state" title="Updating health state" />,
```
spadgett, line 43:
```suggestion
    icon: <BlueSyncIcon alt="Updating" title="Updating" />,
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
...
     message: 'Updating',
   },
   [HealthState.WARNING]: {
     priority: 4,
     health: HealthState.WARNING,
-    icon: <YellowExclamationTriangleIcon />,
+    icon: <YellowExclamationTriangleIcon alt="Warning health state" title="Warning health state" />,
```
spadgett, line 49:
```suggestion
    icon: <YellowExclamationTriangleIcon alt="Warning" title="Warning" />,
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -23,36 +23,36 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
...
     message: 'Degraded',
   },
   [HealthState.ERROR]: {
     priority: 5,
     health: HealthState.ERROR,
-    icon: <RedExclamationCircleIcon />,
+    icon: <RedExclamationCircleIcon alt="Error health state" title="Error health state" />,
```
spadgett, line 55:
```suggestion
    icon: <RedExclamationCircleIcon alt="Error" title="Error" />,
```

frontend/packages/console-shared/src/components/dashboard/status-card/states.tsx
```diff
@@ -64,7 +64,7 @@ export const healthStateMapping: { [key in HealthState]: HealthStateMappingValue
   [HealthState.NOT_AVAILABLE]: {
     priority: 7,
     health: HealthState.NOT_AVAILABLE,
-    icon: <GrayUnknownIcon />,
+    icon: <GrayUnknownIcon alt="Not available health state" title="Not available health state" />,
```
spadgett, line 67:
```suggestion
    icon: <GrayUnknownIcon alt="Not available" title="Not available" />,
```

jessiehuff:
> @jessiehuff Instead of "Ok health state," would it be better to use something like "Healthy"?

Yeah, I could see that working and potentially sounding better. It would essentially say "Healthy, image" then when you navigate to the next item it would read whatever that is. Picking a label can be subjective and tricky, but I agree with your comments. I think "Healthy" and "Warning" communicate what you're trying to say in a concise way so those are definite possibilities. I also wonder if adding context to the label will help make things more clear. Perhaps we could do something like "Status: Healthy", "Status: Warning", etc. to provide more meaning in this context? What do you think @spadgett? I'm ok with either approach. 🙂

spadgett (approved):
/lgtm

Thanks @sg00dwin and @jessiehuff! Let's go ahead with the fix as-is for now so that it makes 4.6.

PR #6558, opened 2020-09-08
Bug 1766287:  hello-openshift template application wrong binary format

cyril-ui-developer:
/assign @TheRealJon
/cc  @spadgett

### DIFF
--- a/frontend/public/models/yaml-templates.ts
+++ b/frontend/public/models/yaml-templates.ts
@@ -7,6 +7,10 @@ import { YAMLTemplate } from '@console/plugin-sdk';
 /**
  * Sample YAML manifests for some of the statically-defined Kubernetes models.
  */
+
+const sampleContainerImage = 'registry.redhat.io/rhel8/support-tools';
+const sampleContainerCmd = '[ "/bin/bash", "-c", "sleep infinity" ]';
+
 export const baseTemplates = ImmutableMap<GroupVersionKind, ImmutableMap<string, string>>()
   .setIn(
     ['DEFAULT', 'default'],
@@ -241,9 +245,10 @@ spec:
     spec:
       containers:
       - name: hello-openshift
-        image: openshift/hello-openshift
+        image: ${sampleContainerImage}
         ports:
         - containerPort: 8080
+        command: ${sampleContainerCmd}
 `,
   )
   .setIn(
@@ -349,9 +354,10 @@ spec:
     spec:
       containers:
       - name: hello-openshift
-        image: openshift/hello-openshift
+        image: ${sampleContainerImage}
         ports:
         - containerPort: 8080
+        command: ${sampleContainerCmd}
 `,
   )
   .setIn(
@@ -408,9 +414,10 @@ metadata:
 spec:
   containers:
     - name: hello-openshift
-      image: openshift/hello-openshift
+      image: ${sampleContainerImage}
       ports:
         - containerPort: 8080
+      command: ${sampleContainerCmd}
 `,
   )
   .setIn(
@@ -609,9 +616,10 @@ spec:
     spec:
       containers:
       - name: hello-openshift
-        image: openshift/hello-openshift
+        image: ${sampleContainerImage}
         ports:
         - containerPort: 8080
+        command: ${sampleContainerCmd}
 `,
   )
   .setIn(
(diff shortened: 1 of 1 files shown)


cyril-ui-developer:
/ bugzilla refresh

cyril-ui-developer:
@spadgett Please do you know why this PR is not updating the Bugzilla?

frontend/packages/knative-plugin/src/yaml-templates.ts
```diff
@@ -13,6 +13,7 @@ spec:
   template:
     spec:
       containers:
-      - image: openshift/hello-openshift
+      - image: registry.redhat.io/rhel8/support-tools
+        command: [ "/bin/bash", "-c", "sleep infinity" ]
```
spadgett, line 17: I'm not sure if this will cause problems for any of the Knative walkthroughs. We might want to address this one separarely.

cc @bbrowning

frontend/public/models/yaml-templates.ts
```diff
@@ -7,6 +7,10 @@ import { YAMLTemplate } from '@console/plugin-sdk';
 /**
  * Sample YAML manifests for some of the statically-defined Kubernetes models.
  */
+
+const containerImageSrc = 'registry.redhat.io/rhel8/support-tools';
```
spadgett, line 11: We probably want to reference the local tools image stream. @wking @jwforres Can you help us determine what is right?

For the node terminal, we try to get the image stream, but fall back to `registry.redhat.io/rhel8/support-tools` if we can't. That's tricky here since these samples are static today. It would be a fair amount of rework if we had to wait for an API request to resolve before we can get the sample.

frontend/public/models/yaml-templates.ts
```diff
@@ -7,6 +7,10 @@ import { YAMLTemplate } from '@console/plugin-sdk';
...
+const containerCmd = '[ "/bin/bash", "-c", "sleep infinity" ]';
```
spadgett, line 12: I'd put "sample" in the variable name to make it clear these are just samples.

```suggestion
const sampleContainerImage = 'registry.redhat.io/rhel8/support-tools';
const sampleContainerCmd = '[ "/bin/bash", "-c", "sleep infinity" ]';
```

cyril-ui-developer: > I'd put "sample" in the variable name to make it clear these are just samples.

@spadgett Addressed, thanks

frontend/packages/knative-plugin/src/yaml-templates.ts
bbrowning: I don't think this will actually work for Knative. The image needs to be one that returns a HTTP 200 in response to a HTTP GET at the root path. We are tracking this work on the Serverless side at https://issues.redhat.com/browse/SRVKS-611 but that issue has not been picked up to work on yet. I think perhaps we were missing the fact that this example is hardcoded in the console itself.

@rhuss Do you know of a better supported image we could suggest for the console to use here? Ideally this would reference an image produced by the serverless team, but I don't think we're supporting anything for this use yet and won't be until the above linked JIRA issue is solved.

rhuss: I'm not 100% sure if I grasp the context fully, but for using an image as sample for Knative serving it has to talk HTTP. Ideally such an image would also be able to consume CloudEvents so that it can be used in an Eventing context, too.

@bbrowning I'm not sure if we productise "edisplay" which is used for logging CloudEvents into the Pod's log, but this would be a good candidate IMO. @navidshaikh do you know about any productised image that we could use as an example here ?

yselkowitz:
FYI we now have someone to own hello-openshift with the intention of adding it to the release.

frontend/packages/knative-plugin/src/yaml-templates.ts
navidshaikh: @rhuss : No, we dont productise such image yet. So we may ship two such images one for hello-example and another for event-display ?

spadgett:
Closing in favor of https://github.com/openshift/origin/pull/25435

/close

frontend/packages/knative-plugin/src/yaml-templates.ts
rhuss: I wonder how suitable our test image from https://github.com/openshift/knative-client/tree/master/openshift/ci-operator/knative-test-images for this purpose ? (at least for the smoke test for serving)

rhuss: > So we may ship two such images one for hello-example and another for event-display ?

I thought about shipping a single image that we could use for serving (accepting GET requests) and eventing (accepting POST requests with cloud events as payload) ?

PR #6559, opened 2020-09-08
Bug 1870339: Sets volumeMode as block for RBD+RWX

a2batic:
- Sets volume Mode for restored claim

### DIFF
--- a/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
+++ b/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as _ from 'lodash';
 
 import { Form, FormGroup, Grid, GridItem, TextInput } from '@patternfly/react-core';
 import {
@@ -34,7 +35,10 @@ import {
 } from '@console/internal/models';
 import { getName, getNamespace, Status, isCephProvisioner } from '@console/shared';
 import { StorageClassDropdown } from '@console/internal/components/utils/storage-class-dropdown';
-import { dropdownUnits } from '@console/internal/components/storage/shared';
+import {
+  dropdownUnits,
+  cephRBDProvisionerSuffix,
+} from '@console/internal/components/storage/shared';
 import { useK8sGet } from '@console/internal/components/utils/k8s-get-hook';
 
 import './restore-pvc-modal.scss';
@@ -113,6 +117,18 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
           },
         },
       };
+
+      if (pvcResource) {
+        // should set block only for RBD + RWX
+        if (
+          _.endsWith(snapshotClassResource?.driver, cephRBDProvisionerSuffix) &&
+          accessModes.includes('ReadWriteMany')
+        ) {
+          restorePVCTemplate.spec.volumeMode = 'Block';
+        } else {
+          restorePVCTemplate.spec.volumeMode = pvcResource?.spec?.volumeMode;
+        }
+      }
 
       // eslint-disable-next-line promise/catch-or-return
       handlePromise(
@@ -227,8 +243,8 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
             </div>
           </ModalBody>
           <ModalSubmitFooter
+            submitDisabled={!pvcSC || !validSize}
             inProgress={inProgress}
-            submitDisabled={!validSize}
             errorMessage={errorMessage}
             submitText="Restore"
             cancel={cancel}


frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -108,6 +114,18 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
...
+        restorePVCTemplate.spec.storageClassName = pvcSC;
+
+        // should set block only for RBD + RWX
+        if (
+          _.endsWith(storageProvisioner, cephRBDProvisionerSuffix) &&
+          accessModes[0] === 'ReadWriteMany'
+        ) {
+          restorePVCTemplate.spec.volumeMode = 'Block';
+        }
+      }
```
cloudbehl, line 127: Is it possible to have RBD + RWX of filesystem type?

Madhu-1: No not supported check https://github.com/ceph/ceph-csi#support-matrix

a2batic: UI doesn't allow us to create a PVC of RBD + RWX in Filesystem type, this combination will always result in creating a Block PVC from UI.

cloudbehl: @a2batic just trying to understand if there is any other scenario where block can be created from CLI & while restore we create filesystem-based.

cloudbehl:
The Ideal way to figure out the weather its block or filesystem will be to get the PVC data & see the access mode. Let's not hardcode it.

frontend/packages/console-app/src/components/modals/clone/clone-pvc-modal.tsx
```diff
@@ -54,6 +66,23 @@ const ClonePVCModal = withHandlePromise((props: ClonePVCModalProps) => {
...
+  const allowedAccessModes =
+    scResourceLoaded &&
+    !scResourceLoadError &&
+    provisionerAccessModeMapping[scResource?.provisioner]
+      ? provisionerAccessModeMapping[scResource?.provisioner]
+      : getAccessModeForProvisioner(scResource?.provisioner);
+
+  const handleAccessMode: React.ReactEventHandler<HTMLInputElement> = (event) => {
+    setAccessMode(event.currentTarget.value);
+  };
```
bipuladh, line 85:
```suggestion
  const handleAccessMode: React.ReactEventHandler<HTMLInputElement> = (event) => setAccessMode(event.currentTarget.value);
```

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -76,15 +98,22 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
...
     };
 
-    const handleStorageClass = (updatedStorageClass) =>
+    const handleStorageClass = (updatedStorageClass: StorageClassResourceKind) => {
       setPVCStorageClass(updatedStorageClass?.metadata.name || '');
+      setStorageProvisioner(updatedStorageClass?.provisioner || '');
+      setAllowedAccessModes(
+        provisionerAccessModeMapping[updatedStorageClass?.provisioner]
+          ? provisionerAccessModeMapping[updatedStorageClass?.provisioner]
+          : getAccessModeForProvisioner(updatedStorageClass?.provisioner),
```
bipuladh, line 107:
```suggestion
        provisionerAccessModeMapping[updatedStorageClass?.provisioner]
        ?? getAccessModeForProvisioner(updatedStorageClass?.provisioner),
```

cloudbehl (approved):

bipuladh (changes requested):

frontend/packages/console-app/src/components/modals/clone/clone-pvc-modal.tsx
```diff
@@ -111,7 +133,40 @@ const ClonePVCModal = withHandlePromise((props: ClonePVCModalProps) => {
...
+                    {...radio}
+                    key={radio.value}
+                    onChange={handleAccessMode}
+                    inline
+                    disabled={disabled}
+                    checked={checked}
+                    name="accessMode"
+                  />
+                );
+              });
```
bipuladh, line 159:
```suggestion
              const disabled = !allowedAccessModes.includes(radio.value);
              const checked = disabled ? !disabled : radio.value === accessMode;
              return (
                  <RadioInput
                    {...radio}
                    key={radio.value}
                    onChange={handleAccessMode}
                    inline
                    disabled={disabled}
                    checked={checked}
                    name="accessMode"
                  />
                );
```
Wouldn't this work?

bipuladh (changes requested):

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -108,6 +112,18 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
         },
       };
 
+      if (pvcResource) {
+        // should set block only for RBD + RWX
+        if (
+          _.endsWith(snapshotClassResource?.driver, cephRBDProvisionerSuffix) &&
+          accessModes[0] === 'ReadWriteMany'
```
spadgett, line 119: I'm not as familiar with the PVC API, but we need to check if any of the array items are `ReadWriteMany`? Or that there is only one item in the array?

a2batic: @spadgett yes a PVC can have only 1 access mode at a time https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

spadgett: Hi, maybe I missed it, but I don't see where the link says there can only be one. (I'm also unsure why it would be an array in the API in that case.)

cloudbehl: As per my understanding: There can be multiple-access modes for a PV but when the PV is mounted using PVC there will be only one access mode that will be used. So we can assume that only one value will exist at a time. (I.e only fits for Static PV creation)

It's the same reason we have radio buttons for access mode while PVC creation.

a2batic: Hi @spadgett , the link says the `Important! A volume can only be mounted using one access mode at a time, even if it supports many.` But, PVC CR lets you create PVC with multiple access modes, so +1, we should handle it. Updated the PR to check in the array. Sorry for the confusion.

a2batic:
@spadgett addressed comments, please review.

spadgett (approved):
/lgtm

PR #6561, opened 2020-09-09
Bug 1877198: Query Browser: Improve error message when Prometheus API returns 403

kyoto:
Add `PrometheusAPIError` type to improve type checking.

Add `err.name !== 'AbortError'` check to avoid the possibility of trying
to update state after the component has unmounted.

screenshot


kyoto:
Generally, the navigation links to this page should not be visible if the user cannot access the Prometheus API. However, it is still possible to access the page (e.g. if you have bookmarked it). Also, I think it's better to handle this case with a more specific error message, to aid troubleshooting.

spadgett (approved):
/lgtm

PR #6562, opened 2020-09-09
Bug 1862564: Fix column visibility on various screen sizes

jcaianirh:
Fixes: 
https://bugzilla.redhat.com/show_bug.cgi?id=1862564

For tables supporting column management (pods, nodes, projects, namespaces), remove static column visibility classes, and add a common table function to dynamically enable/disable visibility as follows:

0 - (SM = 576) = 2 columns max 
SM - (MD = 768) = 4 columns max
MD - (LG = 992) = 4 columns max
LG - (XL = 1200) = 6 columns max
XL - (XXL = 1400) = 8 columns max
XXL - infinity = 9 columns max

Above sizes are based on PF breakpoints. 

SM - LG have the same max col values since that is the breakpoint that the nav shows/hides. Column priority is left to right.

**Before example of pods page with 9 columns selected at 1100 pixels**:
Screen Shot 2020-09-08 at 9 47 43 PM

**After example of same table config and pixel width**:
Screen Shot 2020-09-08 at 9 54 16 PM


frontend/public/components/factory/table.tsx
```diff
@@ -276,15 +276,52 @@ export type TableRowProps = {
...
+    columnID !== 'namespace' || UIActions.getActiveNamespace() === ALL_NAMESPACES_KEY;
+  if (_.isEmpty(columns) && showNamespace) {
+    return true;
+  }
+  if (!columns.has(columnID)) {
+    return false;
+  }
+  const widthInPixels = window.innerWidth;
+  if (widthInPixels < BREAKPOINT_SM) {
+    return [...columns].indexOf(columnID) < MAX_COL_SM;
```
spadgett, line 301: I'm confused by this since `columns` is a `Set`. Does `Set` have a guaranteed ordering?

nit: Calculate the index once and reuse it for each check.

```ts
const columnIndex = [...columns].indexOf(columnID);
```

frontend/public/components/factory/table.tsx
```diff
@@ -276,15 +276,52 @@ export type TableRowProps = {
...
+const BREAKPOINT_LG = 992;
+const BREAKPOINT_XL = 1200;
+const BREAKPOINT_XXL = 1400;
+const MAX_COL_SM = 2;
+const MAX_COL_MD = 4;
+const MAX_COL_LG = 4;
+const MAX_COL_XL = 6;
+const MAX_COL_XXL = 8;
+
+const isColumnVisible = (columnID: string, columns: Set<string> = new Set()) => {
```
spadgett, line 290: Is this reevaluated when the window is resized?

frontend/public/components/factory/table.tsx
```diff
@@ -463,9 +500,12 @@ const getActiveColumns = (
         }
       }),
     );
-  }
-  if (!_.isEmpty(activeColumns)) {
     columns = columns?.filter((col) => activeColumns.has(col.id) || col.title === '');
+  } else {
+    columns = columns?.filter(
+      (col) =>
+        (activeColumns.has(col.id) && isColumnVisible(col.id, activeColumns)) || col.title === '',
```
spadgett, line 507: I don't think the `activeColumns.has(col.id)` check is needed since `isColumnVisible` does that already.

spadgett (commented):
Thanks @jcaianirh

frontend/public/components/factory/table.tsx
spadgett, line 301: I think these names are confusing because if we're `< 576`, we're not at small. We're at extra small. But we're checking `MAX_COL_SM`. Should it be `MAX_COL_XS`, etc?

jcaianirh: @spadgett you are right about the set ordering...looking into that now. Fixed nit.

frontend/public/components/factory/table.tsx
jcaianirh: @spadgett this is reevaluated when the window is resized. The window resize was already causing a re render based on the resize listener.

frontend/public/components/factory/table.tsx
jcaianirh: @spadgett good point, its one size lower than named. fixed this.

frontend/public/components/factory/table.tsx
jcaianirh: @spadgett thanks...remove redundant check.

frontend/public/components/modals/column-management-modal.tsx
```diff
@@ -76,7 +76,11 @@ export const ColumnManagementModal: React.FC<ColumnManagementModalProps> = ({
 
   const submit = (event): void => {
     event.preventDefault();
-    dispatch(setTableColumns(columnLayout.id, checkedColumns));
+    const orderedCheckedColumns = new Set<string>();
+    columnLayout.columns.map(
```
spadgett, line 80:
```suggestion
    columnLayout.columns.forEach(
```

spadgett (commented):
Thanks @jcaianirh

frontend/public/components/modals/column-management-modal.tsx
jcaianirh: updated thanks

spadgett (approved):
/lgtm

PR #6563, opened 2020-09-09
Bug 1876985: Change the text message display for guest agent feature

yaacov:
BZ recommends that we use active voice, tighter wording, fewer words, and stylistic compliance with the IBM Style Guide. The text message noted above should be replaced with the following:

"A guest agent was not found for this VM. Either the guest agent was not installed or the VM has not finished booting. When a guest agent is not installed, some management features are unavailable and the metrics might be inaccurate."

Screesnshot:
screenshot-localhost_9000-2020 09 09-08_12_09


yaacov:
@matthewcarleton @pcbailey please review

yaacov:
cc:// @gouyang this text changes may affect tests

PR #6572, opened 2020-09-09
Bug 1877368: Disable Helm Chart Install form and show error alert if chart is not reachable

### LINKED ISSUE #5190: HELM - Console displays 'Install Helm Chart' screen w/o values.yaml when index.yaml points to invalid tgz path
When installing helm chart from OCP developer catalog if the helm repo index.yaml points to invalid location the install screen is displayed without values (tgz can't be found) and no error message.   

image

@pedjak

divyanshiGupta:
Jira: https://issues.redhat.com/browse/ODC-4360
Fixes #5190

**Screenshot:**
image

**Unit Tests Coverage:**
image

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/helm/HelmInstallUpgradePage.tsx
+++ b/frontend/packages/dev-console/src/components/helm/HelmInstallUpgradePage.tsx
@@ -62,6 +62,7 @@ const HelmInstallUpgradePage: React.FunctionComponent<HelmInstallUpgradePageProp
   const [appVersion, setAppVersion] = React.useState<string>('');
   const [chartReadme, setChartReadme] = React.useState<string>('');
   const [chartHasValues, setChartHasValues] = React.useState<boolean>(false);
+  const [chartError, setChartError] = React.useState<Error>(null);
 
   const [initialYamlData, setInitialYamlData] = React.useState<string>('');
   const [initialFormData, setInitialFormData] = React.useState<object>();
@@ -87,9 +88,12 @@ const HelmInstallUpgradePage: React.FunctionComponent<HelmInstallUpgradePageProp
 
     const fetchHelmRelease = async () => {
       let res;
+      let error: Error = null;
       try {
         res = await coFetchJSON(config.helmReleaseApi);
-      } catch {} // eslint-disable-line no-empty
+      } catch (e) {
+        error = e;
+      }
       if (ignore) return;
       const chart: HelmChart = res?.chart || res;
       const chartValues = getChartValuesYAML(chart);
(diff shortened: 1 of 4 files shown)


frontend/packages/dev-console/src/components/helm/HelmInstallUpgradePage.tsx
```diff
@@ -87,9 +88,12 @@ const HelmInstallUpgradePage: React.FunctionComponent<HelmInstallUpgradePageProp
...
     const fetchHelmRelease = async () => {
       let res;
+      let error: Error = null;
       try {
         res = await coFetchJSON(config.helmReleaseApi);
-      } catch {} // eslint-disable-line no-empty
+      } catch (e) {
+        error = e;
+      } // eslint-disable-line no-empty
       if (ignore) return;
```
abhinandan13jan, line 97: just a query. Why do we have this `ignore` check after the coFetch and not before it?

divyanshiGupta: Okay I will try to explain it briefly. So we do this to prevent updating a state of an unmounted component and as you can see the state is being updated only after the API call.

rohitkrai03 (changes requested):
I think chart version dropdown should be hidden if there's no chart version present. Empty dropdown feels odd here. @parvathyvr WDYT?

andrewballantyne:
> I think chart version dropdown should be hidden if there's no chart version present. Empty dropdown feels odd here. @parvathyvr WDYT?

Do we have placeholders? I wonder if "No versions available" or something akin to that is better... Just wondering if removing the input item, labels et all is going to leave more of a void in the UI.

parvathyvr (commented):
I agree with @rohitkrai03  that the empty dropdown looks odd.In the original mockup screen attached in the bug [ODC-4360](https://issues.redhat.com/browse/ODC-4360) I had the Chart Version dropdown although disabled ,showing the latest chart version/app version.But if we do not have this data to show, my opinion would be in lines with what @andrewballantyne  suggested above.I think if we do not have data to show in the ‘Chart Version’ dropdown we can show ‘Not available’ in the disabled dropdown. CC @beaumorley

rohitkrai03:
@parvathyvr Do you think we should also add the http error that we get back from backend to the alert. I tried it with a private helm chart that is not accessible and it looks something like this - 
image

parvathyvr:
@rohitkrai03 I think giving more info about why the Helm chart is unavailable is a good addition!

> @parvathyvr Do you think we should also add the http error that we get back from backend to the alert. I tried it with a private helm chart that is not accessible and it looks something like this -
> 
> Screenshot from 2020-09-10 21-23-27

divyanshiGupta:
> @rohitkrai03 I think giving more info about why the Helm chart is unavailable is a good addition!
> 
> > @parvathyvr Do you think we should also add the http error that we get back from backend to the alert. I tried it with a private helm chart that is not accessible and it looks something like this -
> > Screenshot from 2020-09-10 21-23-27

@rohitkrai03 @parvathyvr done. PTAL.

rohitkrai03 (approved):
/approve

parvathyvr (approved):
LGTM!

PR #6577, opened 2020-09-09
Use better fix for messed JSON output in yarn analyze command

vojtechszocs:
`yarn analyze` command generates webpack bundle stats (`public/dist/stats.json`) and passes it to [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) which generates visual HTML report (`public/dist/report.html`).

Unfortunately, above mentioned `stats.json` file gets polluted with logging messages:

```
Starting type checking service...
Using 1 worker with 2048MB memory limit
{
  (JSON stats here)
}
```

The current fix `awk '{if(NR>2)print}'` simply removes the first 2 lines, which isn't very robust or future-proof.

The proposed fix `sed '0,/^{/s/^[^{].*//g'` works like this:
- use a character range that starts from beginning (0) and ends with first occurence of regexp `^{` which marks the start of JSON stats object
- on this character range, replace regexp `^[^{].*` (any line that doesn't start with `{`) with empty string

Verification:

```sh
$ cat <<EOF >test
foo bar {
baz qux
{
  "errors": [],
  "warnings": [],
  "test": {}
}
EOF
# writes above to 'test' file

$ cat test | sed '0,/^{/s/^[^{].*//g'


{
  "errors": [],
  "warnings": [],
  "test": {}
}
# replaces extra logs with empty strings
```


spadgett (approved):
/lgtm

dtaylor113 (approved):

PR #6579, opened 2020-09-09
Report actual vendors bundle limit

vojtechszocs:
This removes the hard-coded vendors bundle limit text in `analyze.sh` script.


vojtechszocs:
cc @spadgett

spadgett (approved):
/lgtm

PR #6581, opened 2020-09-09
[WIP] Bug 1873348: Handle condition when no status is set on sub

zherman0:
OLM can sometimes get into a state where no install plan or CSV is made and the subscription has no error state (or error indication) whatsoever. In this case, the current install status page can get stuck in an endless loop waiting for a status.
I have added a timer that waits 10 seconds for an install plan to appear on the system.  If no plan is created, the install page shows an error and gives the user the option to switch to the installed operator view.

Note: I have not been able to fully test his as I am having trouble getting an operator into this state reliably.  

Screenshot_2020-09-09 Installing Operator · OKD


zherman0:
This will not be implemented at this time as the backend code has been updated to provide additional information when an error occurs.

PR #6583, opened 2020-09-10
Bug 1877736: fixes warning shown on close of delete revision modal

invincibleJai:
**Fixes**: 
- https://issues.redhat.com/browse/OCPBUGSM-17243

**Analysis / Root cause**: 
Warning shown on close of Delete Revision modal

**Solution Description**: 
- used cancel for close action
- make use of spec to fetch traffic as done in traffic split modal

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


invincibleJai:
> Tested it locally, works as expected.
> 
> @invincibleJai The change from status to spec is also a bugfix, right?
> 
> /lgtm

yeah updated that ticket as well https://issues.redhat.com/browse/ODC-4240

PR #6585, opened 2020-09-10
Bug 1877806: Skip updating iframe height when component is unmounted

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-4777

**Analysis / Root cause**: he code in markdown-view component uses `setTimeout` to update the height of the iframe, by quickly going through the steps sometimes the timeout function gets executed after the component has been unmounted. That's why it throws that error.

**Solution Description**: In setTimeout check if component is unmounted and skip setting of height if unmounted.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/markdown-view.tsx
+++ b/frontend/public/components/markdown-view.tsx
@@ -50,6 +50,7 @@ export class SyncMarkdownView extends React.Component<
   {}
 > {
   private frame: any;
+  private timeoutHandle: any;
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


frontend/public/components/markdown-view.tsx
```diff
@@ -67,6 +72,10 @@ export class SyncMarkdownView extends React.Component<
 
     // Let the new height take effect, then reset again once we recompute
     setTimeout(() => {
+      if (this.unmounted) {
+        return;
+      }
```
christianvogt, line 77: Prefer to clear the timeout instead.
```suggestion
    this.timeoutHandle = setTimeout(() => {
```

in `componentWillUnmount` call `clearTimeout(this.timeoutHandle)`

rohitkrai03: Done.

PR #6588, opened 2020-09-10
Bug 1877861: fix: return empty Helm repo index file when no HelmChartRepository instances are available in cluster

pedjak:
This fix enables the the requested usecase - an admin should be able to hide Helm chart browsing by removing all instances of `HelmChartRepository` in the cluster.

https://github.com/openshift/console-operator/pull/458 needs to go in first.


pkg/helm/chartproxy/repos.go
```diff
@@ -188,10 +188,8 @@ func mergeIndexFiles(files ...*repo.IndexFile) *repo.IndexFile {
 func (b *helmRepoGetter) List() ([]*helmRepo, error) {
 	var helmRepos []*helmRepo
 	repos, err := b.Client.Resource(helmChartRepositoryGVK).List(context.TODO(), v1.ListOptions{})
-	if err != nil || len(repos.Items) == 0 {
-		// In case no HelmRepoCRs configured or error reading them, use default redhat helm chart repo
-		helmRepos = append(helmRepos, &DefaultRepo)
```
jhadvig, line 193: if `DefaultRepo` is not used anywhere, it should be removed.

jhadvig (changes requested):

jhadvig (approved):
/lgtm

pkg/helm/chartproxy/repos.go
pedjak: done.

PR #6589, opened 2020-09-10
Bug 1875193: Check if resource model is loaded in firehose

jhadvig:
Prior to marking the resource as `loaded` we should check it's model is loaded. We should not set the resource as loaded if the model is missing.

/assign @spadgett @rhamilto 

@mareklibra FYI

### DIFF
--- a/frontend/public/components/utils/firehose.jsx
+++ b/frontend/public/components/utils/firehose.jsx
@@ -118,15 +118,18 @@ const stateToProps = ({ k8s }, { resources }) => {
     (models, { kind }) => models.set(kind, k8s.getIn(['RESOURCES', 'models', kind])),
     ImmutableMap(),
   );
-  const loaded = (r) =>
-    r.optional ||
-    k8s.getIn([
-      makeReduxID(
-        k8sModels.get(r.kind),
-        makeQuery(r.namespace, r.selector, r.fieldSelector, r.name),
-      ),
-      'loaded',
-    ]);
+  const loaded = (r) => {
+    return (
+      (r.optional && _.isUndefined(k8sModels.get(r.kind))) ||
+      k8s.getIn([
+        makeReduxID(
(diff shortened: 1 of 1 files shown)


frontend/public/components/utils/firehose.jsx
```diff
@@ -118,15 +118,18 @@ const stateToProps = ({ k8s }, { resources }) => {
     (models, { kind }) => models.set(kind, k8s.getIn(['RESOURCES', 'models', kind])),
     ImmutableMap(),
   );
-  const loaded = (r) =>
-    r.optional ||
-    k8s.getIn([
+  const loaded = (r) => {
+    if (_.isUndefined(k8sModels.get(r.kind))) {
+      return false;
+    }
```
spadgett, line 124: I think this could be removed and you'd get the same result since the lines below would evaluate to `undefined`.

Either way, this would be a regression of https://bugzilla.redhat.com/show_bug.cgi?id=1731480 :(

frontend/public/components/utils/firehose.jsx
```diff
@@ -118,15 +118,18 @@ const stateToProps = ({ k8s }, { resources }) => {
...
-    k8s.getIn([
-      makeReduxID(
-        k8sModels.get(r.kind),
-        makeQuery(r.namespace, r.selector, r.fieldSelector, r.name),
-      ),
-      'loaded',
-    ]);
+  const loaded = (r) => {
+    return (
+      (r.optional && _.isUndefined(k8sModels.get(r.kind))) ||
```
spadgett, line 123: I think this will still cause issues.

IMO, it would be better to switch to the `useK8sWatch` hook. We don't really need `Firehose` here since we don't have any loading or error status we want to show in the masthead.

spadgett:
Closing in favor of #6592 

/close

PR #6590, opened 2020-09-10
Bug 1877858: render ConsoleNotification BannerTop if notification.spec.…

rhamilto:
…location not specified

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1877858

### DIFF
--- a/frontend/public/components/console-notifier.tsx
+++ b/frontend/public/components/console-notifier.tsx
@@ -16,7 +16,10 @@ const ConsoleNotifier_: React.FC<ConsoleNotifierProps> = ({ obj, location }) => 
     <>
       {_.map(_.get(obj, 'data'), (notification) =>
         notification.spec.location === location ||
-        notification.spec.location === 'BannerTopBottom' ? (
+        notification.spec.location === 'BannerTopBottom' ||
+        // notification.spec.location is optional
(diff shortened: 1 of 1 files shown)


frontend/public/components/console-notifier.tsx
```diff
@@ -16,7 +16,10 @@ const ConsoleNotifier_: React.FC<ConsoleNotifierProps> = ({ obj, location }) =>
     <>
       {_.map(_.get(obj, 'data'), (notification) =>
         notification.spec.location === location ||
-        notification.spec.location === 'BannerTopBottom' ? (
+        notification.spec.location === 'BannerTopBottom' ||
+        // notification.spec.location is optional, so always output the notification
+        // if location is not specified, producing the same result as location: BannerTopBottom
```
spadgett, line 21: I'm OK with this, but I have a slight preference for defaulting to `BannerTop`. I think that's usually what people want.

rhamilto: I went with this because it was the simplest code change...

rhamilto: Updated to default location to `BannerTop` if `location` not specified.

spadgett (approved):
/lgtm

PR #6596, opened 2020-09-11
CONSOLE-2396: Add skip to content component

jessiehuff:
Currently keyboard users have to tab through the entire top nav and then the side nav in order to get to the main content. Adding a skip to content allows keyboard users to quickly and easily get to the main content of a page and helps support section [2.4.1 Bypass Blocks of WCAG](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0&showtechniques=241#qr-navigation-mechanisms-skip) and the VPAT. It's also only visible if you're using a keyboard so it shouldn't negatively affect anything visually. (When the page loads, press Tab. Focus should move to the skip to content component. Press enter and focus will shift to the main content.)

[CONSOLE-2396](https://issues.redhat.com/browse/CONSOLE-2396)

### DIFF
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -39,7 +39,7 @@ import '../vendor.scss';
 import '../style.scss';
 
 //PF4 Imports
-import { Page } from '@patternfly/react-core';
+import { Page, SkipToContent } from '@patternfly/react-core';
 
 const breakpointMD = 768;
 const NOTIFICATION_DRAWER_BREAKPOINT = 1800;
@@ -164,6 +164,13 @@ class App_ extends React.PureComponent {
                 onNavSelect={this._onNavSelect}
                 onPerspectiveSelected={this._onNavSelect}
               />
+            }
+            skipToContent={
+              <SkipToContent
+                href={`${this.props.location.pathname}${this.props.location.search}#content`}
+              >
+                Skip to Content
+              </SkipToContent>
             }
           >
             <ConnectedNotificationDrawer


jessiehuff:
Something I noticed in adding this component is that when the skip to content is used, it almost looks like a page refresh. I'm wondering if this has something to do with the way that routing is handled? I notice that this doesn't happen in any PF [page demos](https://www.patternfly.org/v4/documentation/react/demos/page) with the skip to content.


Screen Recording 2020-09-11 at 9 46 47 AM

frontend/public/components/app.jsx
```diff
@@ -144,6 +144,7 @@ class App_ extends React.PureComponent {
     const { isNavOpen, isDrawerInline } = this.state;
     const { contextProviderExtensions } = this.props;
     const { productName } = getBrandingDetails();
+    const PageSkipToContent = <SkipToContent href={'#content'}>Skip to Content</SkipToContent>;
```
spadgett, line 147: Hi @jessiehuff, can you pull this outside of render so it's a separate top-level component? If it's an anonymous component here, it will unmount/remount every render of `App`.

spadgett:
> Something I noticed in adding this component is that when the skip to content is used, it almost looks like a page refresh. I'm wondering if this has something to do with the way that routing is handled?

Yes, we use react-router. You see that when you use an `a` element instead of `Link`. You probably need to use `HashLink` in this case. If the PF component doesn't allow it -- it looks like it doesn't -- we might have to roll our own.

jessiehuff:
Nice call out on that!! I created an [issue in PF](https://github.com/patternfly/patternfly-react/issues/4809). It would make sense to me for our skip to content component to be able to accept other router components instead of always rendering as an anchor tag. I think we already do something similar in our [Nav](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Nav/NavItem.tsx#L72) so I'm hoping it shouldn't be too difficult to add that.

jcaianirh (approved):
looks good thanks @jessiehuff

jessiehuff:
I've tested this update in VO, NVDA, and JAWS, and it seems to work in all three screen readers.

spadgett:
Hi, @jessiehuff, do we still have the page refresh problem? Do we need to bump Patternfly?

jessiehuff:
> Hi, @jessiehuff, do we still have the page refresh problem? Do we need to bump Patternfly?

Hi @spadgett! I haven't seen the page refresh problem since updating the href to ```href={`${this.props.location.pathname}#content`}```. After looking into it on the patternfly side, someone from the team noticed that it likely wasn't originally working because of the `<base href="/" />` tag that OpenShift includes. I don't think we'll need to bump patternfly.

spadgett:
> > Hi, @jessiehuff, do we still have the page refresh problem? Do we need to bump Patternfly?
> 
> Hi @spadgett! I haven't seen the page refresh problem since updating the href to `` href={`${this.props.location.pathname}#content`} ``. After looking into it on the patternfly side, someone from the team noticed that it likely wasn't originally working because of the `<base href="/" />` tag that OpenShift includes. I don't think we'll need to bump patternfly.

Does `href="#content"` work?

jessiehuff:
@spadgett Unfortunately that's what I was trying previously which gave me the refresh issue. I think OpenShift sets the base href [here](https://github.com/openshift/console/blob/master/frontend/public/index.html#L5) which becomes `<base href="/" />` so that when I use the relative anchor tag like `href="#content"` on child routes like `/dashboard` it becomes `/#content` instead of `/dashboard#content`.

TheRealJon (commented):
/lgtm

frontend/public/components/app.jsx
```diff
@@ -159,6 +159,11 @@ class App_ extends React.PureComponent {
                 onPerspectiveSelected={this._onNavSelect}
               />
             }
+            skipToContent={
+              <SkipToContent href={`${this.props.location.pathname}#content`}>
```
spadgett, line 163: This will drop any query parameters in the URL I believe.

jessiehuff: Very good point, I hadn't considered that. I just updated it to add that consideration!

TheRealJon (approved):
/lgtm

PR #6611, opened 2020-09-14
Bug 1878653: fixes event source form create button enabled when form is invalid

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4302

**Analysis / Root cause**: 
event source form create button enabled when form is invalid on clicking on it again

**Solution Description**: 
If type is same as selected type then not needed to update formik values

**Screenshot/Gif**:
Sep-14-2020 15-22-54

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


nemesis09:
verified locally. Event Source form doesn't validate incorrectly on selecting the same source repeatedly.
/lgtm

PR #6612, opened 2020-09-14
Bug 1878700: Fix Add Pipeline Trigger for Pipeline Operator 1.1

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4496
https://bugzilla.redhat.com/show_bug.cgi?id=1878700

**Analysis / Root cause**:
With the OpenShift Pipeline Operator 1.1 a new version of Pipeline Trigger CRDs was released. The new version was 0.6.

In [0.5.0](https://github.com/tektoncd/triggers/releases/tag/v0.5.0) (we didn't see this version in the Operator) they deprecated `{{spec.triggers[].bindings[].name}}`.

In [0.6.0](https://github.com/tektoncd/triggers/releases/tag/v0.6.0) (TP2 - 1.1 of the Operator) it has been removed in favour of deprecated `{{spec.triggers[].bindings[].ref}}`.

So we need to change the [EventListener generation (source code)](https://github.com/openshift/console/blob/abc92b97a094343b2031d9fd0ac1b14df76a3677/frontend/packages/dev-console/src/components/pipelines/modals/triggers/resource-utils.ts#L45) in a way we support OpenShift Pipeline Operator 1.0 and 1.1.

**Solution Description**: 
The code checks the installed OpenShift Pipeline Operator version. If it finds the version 0.x or 1.0.x it uses the old EventListener format, and the new one format otherwise (also in the case the Operator version check fails).

No visual change here.

Add new tests for the new `getPipelineOperatorVersion` method.

```
 PASS  packages/dev-console/src/components/pipelines/__tests__/pipeline-operator.spec.ts
  getPipelineOperatorVersion
    ✓ should fetch the ClusterServiceVersion from the api (1ms)
    ✓ should return the active ClusterServiceVersion if multiple returns (1ms)
    ✓ should fetch the latest (highest) ClusterServiceVersion from the api
    ✓ should return null if there is no ClusterServiceVersion available (1ms)
    ✓ should return null if there is no matching ClusterServiceVersion available
```

**Test setup:**
* Add the OpenShift Pipeline Operator in the Version 1.0 and 1.1.
* Create a new Pipeline
* Add a trigger

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/resource-types/triggers.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/resource-types/triggers.ts
@@ -27,9 +27,16 @@ export type TriggerTemplateKind = K8sResourceCommon & {
 };
 
 export type EventListenerKindBindingReference = {
-  // TriggerBinding / ClusterTriggerBinding name reference
-  name: string;
-  kind?: string;
+  // TriggerBinding / ClusterTriggerBinding reference
+  kind: string;
+  // Ref is used since Tekton Triggers 0.5 (part of OpenShift Pipeline Operator 1.1)
+  ref: string;
+  // We also support older operators, so need to show & save the old field as well.
+  // https://github.com/tektoncd/triggers/pull/603/files
+  // https://github.com/tektoncd/triggers/releases/tag/v0.5.0 and
+  // https://github.com/tektoncd/triggers/releases/tag/v0.6.0
+  /** @deprecated use ref instead */
+  name?: string;
 };
 
 export type EventListenerKindTrigger = {
--- a/frontend/packages/dev-console/src/components/pipelines/utils/__tests__/pipeline-operator.spec.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/utils/__tests__/pipeline-operator.spec.ts
@@ -1,0 +1,124 @@
+import { k8sList } from '@console/internal/module/k8s';
+import { ClusterServiceVersionKind } from '@console/operator-lifecycle-manager';
+import { getPipelineOperatorVersion } from '../pipeline-operator';
+
+jest.mock('@console/internal/module/k8s', () => ({
+  k8sList: jest.fn(),
+}));
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+const k8sListMock = k8sList as jest.Mock;
+
+describe('getPipelineOperatorVersion', () => {
+  it('should fetch the ClusterServiceVersion from the api', async () => {
+    const csvs = [
+      {
+        metadata: { name: 'openshift-pipelines-operator.v1.0.1' },
+        spec: { version: '1.0.1' },
+        status: { phase: 'Succeeded' },
+      } as ClusterServiceVersionKind,
+    ];
+    k8sListMock.mockReturnValueOnce(Promise.resolve(csvs));
+
+    const version = await getPipelineOperatorVersion('unit-test');
+
+    expect(version.raw).toBe('1.0.1');
+    expect(version.major).toBe(1);
+    expect(version.minor).toBe(0);
+    expect(version.patch).toBe(1);
+    expect(k8sList).toHaveBeenCalledTimes(1);
+  });
+
+  it('should return the active ClusterServiceVersion if multiple returns', async () => {
+    const csvs = [
+      {
+        metadata: { name: 'openshift-pipelines-operator.v1.0.1' },
+        spec: { version: '1.0.1' },
+        status: { phase: 'Succeeded' },
+      } as ClusterServiceVersionKind,
+      {
+        metadata: { name: 'openshift-pipelines-operator.v1.1.1' },
+        spec: { version: '1.1.1' },
+        status: { phase: 'Pending' },
+      } as ClusterServiceVersionKind,
+    ];
+    k8sListMock.mockReturnValueOnce(Promise.resolve(csvs));
+
+    const version = await getPipelineOperatorVersion('unit-test');
+
+    expect(version.raw).toBe('1.0.1');
+    expect(version.major).toBe(1);
+    expect(version.minor).toBe(0);
+    expect(version.patch).toBe(1);
+    expect(k8sList).toHaveBeenCalledTimes(1);
+  });
+
+  it('should fetch the latest (highest) ClusterServiceVersion from the api', async () => {
+    const csvs = [
+      {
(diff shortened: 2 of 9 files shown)


andrewballantyne:
@jerolimov You have TS errors, I think it's causing the whole CI stack to fail:

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/pipelines/utils/triggers.ts
(171,19): Property 'name' does not exist on type 'EventListenerKindBindingReference'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/pipelines/utils/triggers.ts
(199,20): Property 'name' does not exist on type 'EventListenerKindBindingReference'.
```

jerolimov:
> You have TS errors, I think it's causing the whole CI stack to fail:

@andrewballantyne Thanks for that hint. :) Fixed the TS error. This fixes also an issue that the `ClusterTriggerBinding` name (or now ref) was not shown on the EventListener detail page. We talked about this issue already.

Previously it does not show the name for Pipeline Triggers which was created with 1.1.

I noticed that Pipeline Triggers which was created with 1.0 does not run in 1.1. But this is true for all Pipeline objects I think. The pipelines itself also doesn't work after the update. Since everything is Tech Preview and Alpha here I think this is "okayish"

frontend/packages/dev-console/src/components/pipelines/resource-types/triggers.ts
```diff
@@ -27,9 +27,16 @@ export type TriggerTemplateKind = K8sResourceCommon & {
...
+  // TriggerBinding / ClusterTriggerBinding reference
+  kind: string;
+  // Ref is used since Tekton Triggers 0.5 (part of OpenShift Pipeline Operator 1.1)
+  ref: string;
+  // We also support older operators, so need to show & save the old field as well.
+  // https://github.com/tektoncd/triggers/pull/603/files
+  // https://github.com/tektoncd/triggers/releases/tag/v0.5.0 and
+  // https://github.com/tektoncd/triggers/releases/tag/v0.6.0
+  /** @deprecated use ref instead */
+  name?: string;
```
rottencandy, line 39: Wondering if we might be able to use `keyof` here, or would it become too verbose? Especially since `name` is deprecated.

jerolimov: Don't know what the solution looks like then. But the usage in triggers.ts works fine with name and at least VScode shows these deprecated hint. But I'm open to see and try an example here. /me open the TS keyof definition :)

jerolimov:
The current solution uses the Tekton CRD version labels. Unfortunately these are not readable for all users / developers. Will change the solution and remove the WIP label after that.

jerolimov:
@andrewballantyne @rottencandy I changed the implementation to check the OpenShift Pipeline Operator version, which is part of the operator name. I extracted the code into an utils function.

frontend/packages/dev-console/src/components/pipelines/utils/pipeline-operator.ts
```diff
@@ -0,0 +1,27 @@
...
+import { k8sList } from '@console/internal/module/k8s';
+
+export const getPipelineOperatorVersion = async (namespace: string): Promise<SemVer | null> => {
+  const allCSVs: ClusterServiceVersionKind[] = await k8sList(ClusterServiceVersionModel, {
+    ns: namespace,
+  });
+  const matchingCSVs = allCSVs.filter(
+    (csv) =>
+      csv.metadata?.name?.startsWith('openshift-pipelines-operator') &&
+      csv.status?.phase === ClusterServiceVersionPhase.CSVPhaseSucceeded,
```
jerolimov, line 16: @andrewballantyne @christianvogt Does it makes sense to check the status.phase of the Pipeline Operator to get the right version? If this is not required I could use `k8sListPartialMetadata` insteadof `k8sList` which makes the API response much smaller.

frontend/packages/dev-console/src/components/pipelines/utils/__tests__/pipeline-operator.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(k8sList).toHaveBeenCalledTimes(1);
+  });
+
+  it('should return null if there is no ClusterServiceVersion available', async () => {
+    k8sListMock.mockReturnValueOnce(Promise.resolve([]));
+    await expect(getPipelineOperatorVersion('unit-test')).resolves.toBe(null);
+    expect(k8sList).toHaveBeenCalledTimes(1);
+  });
+
+  it('should return null if there is no matching ClusterServiceVersion available', async () => {
```
andrewballantyne, line 94: Could you write one more test for valid items but not including a Pipeline Operator? I think you covered both sides (none and mixed), but I think there is still valid test possible here for just invalid ones.

jerolimov: added new test, it should return null if there is no pipeline operator at all

frontend/packages/dev-console/src/components/pipelines/utils/pipeline-operator.ts
```diff
@@ -0,0 +1,25 @@
...
+  );
+  const versions = matchingCSVs.map((csv) => parse(csv.spec.version)).filter(Boolean);
+  // Orders from small (oldest) to highest (newest) version
```
andrewballantyne, line 19: I'm confused by this comment vs the tests.

```suggestion
  // Orders versions from latest to earliest
```
You're grabbing the last one, the earliest version based on your tests.

jerolimov: The order after this `versions.sort(compare)` is 1.0, 1.1, 2.0, ... And then I selected the last one (to get the newest version). The comment was only for the `versions.sort` line. I will update that comment so that it matches both lines or do some code change. don't know yet.

andrewballantyne: Ah I was confused by your test... I didn't read the `phase` value. You're right. Apologizes -- ignore this thread.

andrewballantyne (approved):

jerolimov:
Because automatically check picking fails I created PR #6675 manually.

PR #6570, opened 2020-09-09
Bug 1877351: Add fallback to builder image detection

karthikjeeyar:
**Fixes**: https://issues.redhat.com/browse/ODC-4205

**Problem**:
If there are multiple builder image recommendations, then builder detection should pick the releavant builder image based on the list of supported builderImages.

**Solution**:
Builder image detection needs to match with the first supported builder images in the console.

**Screenshots**:
builder-image-issue

To Test: 

1. go to +Add -> From git
2. enter git repo url https://github.com/nodeshift-starters/react-web-app

Builder image detection should select `nodeJs` since `modern-web-app` is no longer supported by openshift imagestreams.

PR #6617, opened 2020-09-14
Bug 1878713: Add fallback to builder image detection

karthikjeeyar:
This is a manual cherry pick of https://github.com/openshift/console/pull/6570

Fixes: https://issues.redhat.com/browse/ODC-4205

Problem:
If there are multiple builder image recommendations, then builder detection should pick the releavant builder image based on the list of supported builderImages.

Solution:
Builder image detection needs to match with the first supported builder images in the console.

cc: @divyanshiGupta


andrewballantyne (approved):
/lgtm

Over cautious cherry-pick bot. Same diff as 4.6. Nearby lines are different but shouldn't impact.

markmc:
(patch manager) relatively low risk UX improvement

PR #6621, opened 2020-09-15
Bug 1866337: Changes placeholder to 'Search by node name' in install page

vbnrh:

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
@@ -182,6 +182,7 @@ export const CreateInternalCluster = withHandlePromise<
           customData={{
             onRowSelected: setNodes,
           }}
+          filterPlaceholder="Search by node name..."
         >
           <>
             <div>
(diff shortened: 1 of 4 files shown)


cloudbehl:
@vbnrh you are changing the generic component to fix the problem for OCS. You need to pass the placeholder from the OCS filter component.

bipuladh (commented):
@andybraren can we replace `Search by name...` with `Search by "resource" name...`

andybraren:
Reading through the [Bugzilla discussion](https://bugzilla.redhat.com/show_bug.cgi?id=1866337), I think Shiri's latest suggestion of `Search by node name` sounds good.

Ankush is correct that `Search` is the word we use elsewhere in the Console at the moment (with a few outliers), and if we ever decide to switch to `Filter` (which other products and [PatternFly's examples](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters) use) we can switch all of our fields at the same time.

I do wonder if the incorrect position of the filter contributed to (multiple?) users being confused during the study, but `Search by [resource] name` seems like a harmless adjustment to make it even clearer for now.

FYI @yuvalgalanti @shirimordechay

spadgett (approved):
/lgtm

PR #6627, opened 2020-09-15
Bug 1879089: Use new parameter prefix (tt.param.) for Tekton Trigger version 0.6+

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4452
https://bugzilla.redhat.com/show_bug.cgi?id=1879089

**Analysis / Root cause**: 
Part of Tekton Triggers v0.6.0 comes a deprecation notice that changes the way parameters are referenced. The UI makes a note to inform the user how to approach using variables and this will need to be updated.

We have until v0.7.0 (v1.2 of the Operator, aka TP3) to implement this change before we start misleading the user. There shouldn't be any code break as it's entirely user driven input.

**Solution Description**: 
Check the installed Tekton TriggerTemplate CRD `version` label. The version label is missing for Tekton CRDs before 0.6. So we just verify if the version is available or not. We use the new param syntax ("tt.params.parameter") as default and switch to the old one ("params.parameter") if we find a CRD without a version.

image

Unchanged

**Test setup:**
* Install the OpenShift Pipeline Operator 1.0.x and 1.1.x
* Create a Deployment and enable the option "Add Pipeline"
* Open the Pipeline and select the action "Add Trigger"
* Select a Git Type Provider and press "Show variables" 

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/modals/triggers/TriggerBindingSection.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/triggers/TriggerBindingSection.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { useFormikContext } from 'formik';
 import { Badge, ExpandableSection, FormHelperText } from '@patternfly/react-core';
 import FormSection from '../../../import/section/FormSection';
+import { usePipelineOperatorVersion } from '../../utils/pipeline-operator';
 import { TriggerBindingKind, TriggerBindingParam } from '../../resource-types';
 import TriggerBindingSelector from './TriggerBindingSelector';
 import { AddTriggerFormValues } from './types';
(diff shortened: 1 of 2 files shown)


jerolimov:
The current solution uses the Tekton CRD version labels. Unfortunately these are not readable for all users / developers. Will change the solution and remove the WIP label after that.

andrewballantyne (approved):
This appears stacked on your other one... so we'll put this on hold until that merges.

/hold
Until #6112 merges and this is rebased to remove the 1st commit.

jerolimov:
> This appears stacked on your other one... so we'll put this on hold until that merges.
> 
> Until #6112 merges and this is rebased to remove the 1st commit.

Exactly that was my plan. #6612 (not #6112) is now merged. This one is rebased and have now an acceptable size for this kind of issue / backward compatibility.

andrewballantyne (approved):

rottencandy (approved):
/lgtm

jerolimov:
Remove hold since it was a request for "pls rebase this PR". This was now done and Andrew already approved this PR after that.

/hold cancel

jerolimov:
Ups..

/cherry-pick release-4.5

PR #6628, opened 2020-09-15
Bug 1880118: Cypress:  Disable eslint rule, remove `return cy...`'s to fix flake

dtaylor113:
"promise/catch-or-return" was actually disabled on 9/2, this PR removes `return cy...`'s which were in place to get around this rule. Noticed some flakyness around using `return cy..`; noticeably Monitoring 'Expire Silence' -even though reported 'pass' looking at improved headless logging (PR #6629) and watching movie noticed that 'Expire Silence' modal never actually called due to returning too early!

Disabled `promise/no-nesting` rule due to yamlEditor using `cy.window` which is async, so we do need to nest promises to handle set/get window-yaml-editor content appropriately.

### DIFF
--- a/frontend/packages/integration-tests-cypress/.eslintrc
+++ b/frontend/packages/integration-tests-cypress/.eslintrc
@@ -9,6 +9,7 @@
     "no-console": "off",
     "no-namespace": "off",
     "no-redeclare": "off",
-    "promise/catch-or-return": "off"
+    "promise/catch-or-return": "off",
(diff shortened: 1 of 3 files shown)


spadgett (approved):
/lgtm

Thanks @dtaylor113. This will need a BZ to merge.

spadgett:
@dtaylor113 This will need a Bugzilla bug before we can merge.

PR #6642, opened 2020-09-16
Bug 1847845: Sending requested storage size as min PV Size for Bare Metal

gnehapk:
- Sending requested storage size as of the minimum PV size for create OCS and add capacity on BM/Attached Devices
- Allowed the manually partition disks too for discovery results

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
@@ -8,7 +8,8 @@ import {
   ModalTitle,
 } from '@console/internal/components/factory';
 import { usePrometheusPoll } from '@console/internal/components/graphs/prometheus-poll-hook';
-import { k8sPatch, StorageClassResourceKind } from '@console/internal/module/k8s';
+import { k8sPatch, StorageClassResourceKind, K8sResourceKind } from '@console/internal/module/k8s';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
 import { getName, getRequestedPVCSize } from '@console/shared';
 import { OCSServiceModel } from '../../../models';
 import { getCurrentDeviceSetIndex } from '../../../utils/add-capacity';
@@ -22,7 +23,9 @@ import {
 import { OCSStorageClassDropdown } from '../storage-class-dropdown';
 import { PVsAvailableCapacity } from '../../ocs-install/pvs-available-capacity';
 import { createDeviceSet, DeviceSet } from '../../ocs-install/ocs-request-data';
-import { cephCapacityResource } from '../../../constants/resources';
+import { cephCapacityResource, pvResource } from '../../../constants/resources';
+import { getSCAvailablePVs, getMinSizePVCapacity } from '../../../selectors';
+
 import './_add-capacity-modal.scss';
 
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
@@ -47,6 +50,7 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const selectedSCName: string = getName(storageClass);
   const deviceSetIndex: number = getCurrentDeviceSetIndex(deviceSets, selectedSCName);
   const replica = OCS_DEVICE_SET_REPLICA;
+  const pvsList = useK8sWatchResource<K8sResourceKind[]>(pvResource);
 
   let currentCapacity: React.ReactNode;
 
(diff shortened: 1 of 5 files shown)


frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -72,3 +72,10 @@ export const calcPVsCapacity = (pvs: K8sResourceKind[]): number =>
 
 export const getSCAvailablePVs = (pvsData: K8sResourceKind[], sc: string): K8sResourceKind[] =>
   pvsData.filter((pv) => getPVStorageClass(pv) === sc && pv.status.phase === status.AVAILABLE);
+
+export const getMinSizePVCapacity = (pvs: K8sResourceKind[]): number => {
```
cloudbehl, line 76: We need min size for the PV's that are not bound.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -104,6 +107,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
     }
   };
 
+  const submit = (event: React.FormEvent<EventTarget>) => {
+    event.preventDefault();
+    setProgress(true);
+
+    if (isNoProvionerSC) {
+      const [pvsData, pvsLoaded, pvsLoadError] = pvsList;
+      if ((pvsLoadError || pvsData.length === 0 || !storageClass) && pvsLoaded) {
```
cloudbehl, line 116: when the storage class is not set why do we want to make a request?

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -104,6 +107,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+        // if pv list fails, pass the size as 1
+        makeAddCapacityRequest(defaultRequestSize.BAREMETAL);
+      } else if (pvsLoaded) {
+        const pvs = getSCAvailablePVs(pvsData, getName(storageClass));
+        // empty pvs list will return infinity, hence added this condition
+        const size = pvs.length
+          ? getMinSizePVCapacity(pvs).toString()
+          : defaultRequestSize.BAREMETAL;
```
cloudbehl, line 124: PvsLoaded and pvs.length ===0 will be caught in first block itself. Why we need to check again?

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
cloudbehl: If it's loaded & we know the pvs are present we should get getMinSizePVCapacity otherwise we set default size. This might reduce the complexity of statement.

gnehapk: That is what I am doing. Refer https://github.com/openshift/console/pull/6642/files#diff-ecff7031320d48d9e801ba99a2fb7188R120. 
Request will not go, as there is check for storage class.

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
gnehapk: Yup, pvs with available state only will come in the list

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
gnehapk: This is applicable when the pvs with available state are not there. There can be other pvs with bound state and hence will come to this block.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
cloudbehl: We are checking storage class over here itself. Why need to check again?

https://github.com/openshift/console/pull/6642/files#diff-ecff7031320d48d9e801ba99a2fb7188R114

cloudbehl: ALso do we need to update the value while doing add capacity & replace the existing values?

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
cloudbehl: ack

PR #6643, opened 2020-09-16
Bug 1879054: UI should display password in cloud-init

glekner:
Added label with the password to the right of the new window button.
The label will only be present if a password is found in the YAML.
thoughts @jelkosz @matthewcarleton ?

image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vms/vm-console.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vms/vm-console.tsx
@@ -25,6 +25,7 @@ import {
   asVM,
   getIsGraphicsConsoleAttached,
   getIsSerialConsoleAttached,
+  getCloudInitVolume,
 } from '../../selectors/vm';
 import { isVM, isVMI } from '../../selectors/check-type';
 import { VMIKind, VMKind } from '../../types/vm';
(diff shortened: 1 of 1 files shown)


jelkosz:
2 things: 
1: we need to provide both username and password. If the "user" is not set, the username and the password is the same. But if it looks e.g. like this:
              #cloud-config
              user: admin
              password: q1w2e3
              chpasswd: {expire: False}
              ssh_pwauth: True
than the user is admin

2: you have to show the password only after the user chooses to. @matthewcarleton @yfrimanm WDYT?

glekner:
Added user, and to show the password you now have to click Show Password

image

glekner:
image

@jelkosz @matthewcarleton

jelkosz:
@glekner 
1: the "if it is not successful" needs to have a capital "I"
2: once you hit "show password", you need to have also a "hide password" to hide it again

other than that work very nice!

matthewcarleton:
> image
> 
> @jelkosz @matthewcarleton

sorry small text tweak:


`The following credentials for this operating system were created via Cloud-init. If unsuccessful the cloud-init could be improperly configured. Please contact the image provider for more information. `

glekner:
fixed text and added `Hide password` button

PR #6646, opened 2020-09-16
Bug 1879485: fix application selector incontext for add forms

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-2386

**Analysis / Root cause**: when the application is selected using the ApplicationSelector in the secondary masthead . Context for the selected application is not kept in the import forms

**Solution Description**: Keep the context of application

**Screen shots / Gifs for design review**: Coming soon

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/dropdown/ApplicationSelector.tsx
+++ b/frontend/packages/dev-console/src/components/dropdown/ApplicationSelector.tsx
@@ -27,16 +27,21 @@ interface DispatchProps {
 type Props = ApplicationSelectorProps & StateProps & DispatchProps;
 
 const ApplicationSelector: React.FC<Props> = ({ namespace, application, onChange, disabled }) => {
-  if (namespace === ALL_NAMESPACES_KEY) return null;
-
   const allApplicationsTitle = 'all applications';
   const noApplicationsTitle = 'unassigned';
-  const title: string =
+  const dropdownTitle: string =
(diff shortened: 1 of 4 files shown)


invincibleJai:
Looks good, Verified locally seems to work as expected

christianvogt:
The selector is still showing `all applications` when the last application is deleted in another browser.

invincibleJai:
@sahil143 have verified again after fix , i see as below i.e even if user deletes app name from another browser still show's that application. That's the expected behaviour right?


Sep-24-2020 14-19-35

sahil143:
> @sahil143 have verified again after fix , i see as below i.e even if user deletes app name from another browser still show's that application. That's the expected behaviour right?

Yes, That is expected behaviour

cc @christianvogt

jerolimov:
Unfortunately this change means also that a deep link with the `?application` parameter was not displayed correctly. But the new Deployment was created with the correct application anyway!

Screencast_24 09 2020_14:35:27 webm

christianvogt:
@jerolimov I don't see a difference in your scenario with this change or without. It seems like it's something we can handle in a separate issue. Do you mind creating a issue for this?

christianvogt:
This improves the issue.
UX had some further suggestions that we can consider after further review.

PR #6649, opened 2020-09-16
Bug 1877822: The OS which already has a golden image can not be selected

glekner:
Os with a golden image selected from a URL param or from the dropdown will have the following alert:

image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
@@ -27,6 +27,7 @@ import {
   history,
   resourcePath,
   ExternalLink,
+  ResourceLink,
 } from '@console/internal/components/utils';
 import { StorageClassDropdown } from '@console/internal/components/utils/storage-class-dropdown';
 import { RadioInput } from '@console/internal/components/radio';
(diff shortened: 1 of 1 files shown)


glekner:
@yaacov can you review

jelkosz:
@glekner you need to disable the controls on the screen when you pick an os which already has an image. Now I do see the error message but I can ignore it and attempt to do an upload and get 
image

matthewcarleton:
> Os with a golden image selected from a URL param or from the dropdown will have the following alert:
> 
> image

So does that text within the dropdown change to default data image already exists when they select it? I don't think we need to change it if we are showing the alert.

glekner:
@matthewcarleton @jelkosz 
Tomas wrote in the BZ to keep the `default data image already exists` inside the dropdown. can we settle?

matthewcarleton:
My opinion hasn’t changed :) Tomas WDYT?

jelkosz:
@matthewcarleton it does not change - it is there the whole time so I would keep it there. The "source available" feels like "everything is fine, source is available, feel free to pick me" while the "default data image already exists" feels like "dont touch this, it has already been configured" :)

glekner:
Upload button is now disabled when the alert is shown @jelkosz 
image

matthewcarleton:
> @matthewcarleton it does not change - it is there the whole time so I would keep it there. The "source available" feels like "everything is fine, source is available, feel free to pick me" while the "default data image already exists" feels like "dont touch this, it has already been configured" :)

ok I see your point. LGTM :)

PR #6652, opened 2020-09-16
Bug 1874028: Add sum to query

rebeccaalpert:
Added sum to query per Yadan's comment in https://bugzilla.redhat.com/show_bug.cgi?id=1874028.


spadgett (approved):
/lgtm

PR #6655, opened 2020-09-16
Add useResolvedExtensions hook and improve related code

vojtechszocs:
This PR adds new `useResolvedExtensions` hook as an alternative to existing `useExtensions` hook.

The purpose of `useResolvedExtensions` hook is to resolve *all* code references within each extension's `properties` object for *all* matching extensions (narrowed by `typeGuards` and currently in use).

```ts
const fooResolvedExtensions = useResolvedExtensions<Foo>(isFoo);
```

Here's a runtime value comparison (`fooExtensions` vs. `fooResolvedExtensions`):

useExtensions

useResolvedExtensions

Internally, `useResolvedExtensions` calls `useExtensions` and adapts its result, preserving the order of elements in the array.

Other code improvements:
- `useExtensions` and `useResolvedExtensions` trigger re-render only for mounted components
- more reliable way of tracking `CodeRef<T>` functions via [`Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) marker property
- added `pluginID: string` to `LoadedExtension` interface
- removed `pluginEntryCallbackRegistered` flag in `plugin-loader.ts`

Verified that plugin-specific exposed module chunks (e.g. `utils_bar_ts-chunk.js`) are fetched over the network just once; subsequent `CodeRef<T>` resolutions use the cached module.

---

cc @spadgett @christianvogt

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
@@ -1,0 +1,76 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import { useExtensions } from '@console/plugin-sdk/src/api/useExtensions';
+import {
+  Extension,
+  ExtensionTypeGuard,
+  LoadedExtension,
+} from '@console/plugin-sdk/src/typings/base';
+import { resolveCodeRefProperties } from '../coderefs/coderef-resolver';
+import {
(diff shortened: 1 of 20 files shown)


vojtechszocs:
/hold

Needs some small updates.

vojtechszocs:
`e2e-gcp-console` failed on infra/container setup, retesting.

/retest

christianvogt:
2nd commit changes lgtm

PR #6660, opened 2020-09-17
Added insights-plugin

bond95:
This PR adds new insights-plugin, which is used to show insights information about cluster, such as number of issues that was found and Total Risks of issues and link to cluster's page in OCM.

More info about integration insights into OCP WebConsole could be found here: https://docs.google.com/document/d/1d4KdH3DpUrUTFIrFN6tnItQPFV2NRFYnm_-yMJ6gNN0/edit#

image
image


frontend/packages/insights-plugin/src/models.ts
```diff
@@ -0,0 +1,13 @@
+import { K8sKind } from '@console/internal/module/k8s';
+
+export const InsightsModel: K8sKind = {
```
iNecas, line 3: I guess this can be removed now, right?

bond95: Fixed

bond95:
@spadgett @smarterclayton Could you, please, also take a look?

doruskova:
@bond95 Can you add a better quality image for the card? Thanks

bond95:
@doruskova done

doruskova:
@bond95 It looks great, but is possible to align the icons with the title "Total risk"?

frontend/packages/insights-plugin/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@console/container-security": "0.0.0-fixed",
+    "@console/dev-console": "0.0.0-fixed",
+    "@console/internal": "0.0.0-fixed",
+    "@console/knative-plugin": "0.0.0-fixed",
+    "@console/kubevirt-plugin": "0.0.0-fixed",
+    "@console/local-storage-operator-plugin": "0.0.0-fixed",
+    "@console/metal3-plugin": "0.0.0-fixed",
+    "@console/network-attachment-definition-plugin": "0.0.0-fixed",
+    "@console/noobaa-storage-plugin": "0.0.0-fixed",
+    "@console/operator-lifecycle-manager": "0.0.0-fixed",
```
spadgett, line 21: I assume you don't really depend on all of these? You should only need to specify dependencies specific to your plugin here.

frontend/packages/insights-plugin/OWNERS
```diff
@@ -0,0 +1,8 @@
+reviewers:
+  - bond95
+  - tisnik
+approvers:
+  - bond95
+  - tisnik
+labels:
+  - component/insights
```
spadgett, line 8: You'll need a PR to the openshift/release repo to add this label. See https://github.com/openshift/release/pull/4526

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+import './style.scss';
+
+const DataComponent: React.FC<DataComponentProps> = ({ x, y, datum }) => {
+  const Icon = riskIcons[datum.id];
+  return <Icon x={x} y={y - 5} fill={legendColorScale[datum.id]} />;
+};
+
+export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses, k8sResult }) => {
+  const resource = mapMetrics(responses[0].response);
+  const clusterId = _.get(k8sResult, 'data.spec.clusterID', '');
```
spadgett, line 15: nit: For new code, we prefer optional chaining. Console naming conventions would make this uppercase `ID`

```suggestion
  const clusterID = k8sResult?.data?.spec?.clusterID || '';
```

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+        {!isThereIssues && (
+          <div className="co-insights__no-rules">No Insights data to display.</div>
+        )}
+      </div>
+      <div className="co-status-popup__section">
+        {isThereIssues && (
+          <>
+            <div style={{ fontWeight: 'bold' }}>Fixable issues</div>
+            <div>
+              <a href={`https://cloud.redhat.com/openshift/details/${clusterId}`}>
```
spadgett, line 79: Use the `ExternalLink` component. Here's an example:

https://github.com/openshift/console/blob/24912acc536dd84b8e9e3317fc3a324a2cbf8bfd/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-community-provider-modal.tsx#L52-L55

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+                View all in OpenShift Cluster Manager
+              </a>
+            </div>
+          </>
+        )}
+        {!isThereIssues && (
+          <a href="https://docs.openshift.com/container-platform/latest/support/getting-support.html">
```
spadgett, line 86: `ExternalLink` here as well.

frontend/packages/insights-plugin/src/components/InsightsPopup/style.scss
```diff
@@ -0,0 +1,5 @@
+.co-insights__box {
+	.co-insights__no-rules {
+		color: var(--pf-global--Color--200);
+	}
+}
```
spadgett, line 5: nit: missing newline at the end of the file.

frontend/packages/insights-plugin/src/components/InsightsPopup/style.scss
```diff
@@ -0,0 +1,5 @@
+.co-insights__box {
+	.co-insights__no-rules {
```
spadgett, line 2: We typically avoid nesting of CSS styles and find it's not often necessary with BEM. Is `.co-insights__no-rules` by itself specific enough here?

spadgett (commented):
A few comments, but looks good overall. Nice work. Thanks!

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+type Metrics = {
+  critical?: number;
+  important?: number;
+  low?: number;
+  moderate?: number;
+};
+
+export const mapMetrics = (response) => {
```
rawagner, line 49: missing types

frontend/packages/insights-plugin/src/plugin.tsx
```diff
@@ -0,0 +1,35 @@
...
+        name: 'version',
+        isList: false,
+        prop: 'cluster',
+      },
+      popupComponent: () =>
+        import('./components/InsightsPopup/index' /* webpackChunkName: "insights-plugin" */).then(
+          (m) => m.InsightsPopup,
+        ),
+      popupTitle: 'Insights status',
+    },
```
rawagner, line 31: You need to gate the extension by some model contributed by Insights operator. To do so:

1. create ModelDefinition extension
```js
{
    type: 'ModelDefinition',
    properties: {
      models: [YourModel],
    },
  },
```

2. Add FeatureFlag extension
```js
const INSIGHTS_FLAG = 'INSIGHTS_FLAG';
```
```js
{
    type: 'FeatureFlag/Model',
    properties: {
      model: YourModel,
      flag: INSIGHTS_FLAG,
    },
  },
```
 
3. gate Dashboard Health with the above flag

```suggestion
   flags: {
      required: [InsightsFlag],
    },
    },
```

rawagner (changes requested):

bond95:
@doruskova I updated screenshot, please, take a look

frontend/packages/insights-plugin/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@console/insights-plugin",
+  "version": "0.0.0-fixed",
+  "description": "Plugin for Insights operator",
```
vojtechszocs, line 4: Nit: I'd put a more descriptive message here, like:

```
Insights - provide cluster health data and integrate with OpenShift Cluster Manager
```

frontend/packages/insights-plugin/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "main": "src/index.ts",
+  "scripts": {
+    "test": "yarn --cwd ../.. run test packages/insights-plugin"
```
vojtechszocs, line 8:
```suggestion
    "test": "yarn --cwd ../.. test packages/insights-plugin"
```

frontend/packages/insights-plugin/src/plugin.tsx
vojtechszocs: @bond95 As @rawagner wrote, we should enable the plugin's extensions only when the cluster has the right capability (i.e. Insights operator is installed).

Detecting Insights operator can be done via CRD detection. Use `ModelDefinition` extension to make Console aware of your CRD, then use `FeatureFlag/Model` extension to introduce a feature flag reflecting the presence of such CRD.

Every other extension can then use the `flags` object to declare which feature flags are required and/or disallowed in order for that particular extension to be in use. Typically, you'd want all Insights extensions to be gated by `INSIGHTS_FLAG`.

vojtechszocs (changes requested):

doruskova:
@bond95 Thank you for the update. I noticed 2 things: 

- Insights status title looks smaller than the Total risk title. Insights status should use 16px and Total risk 14px (If it's too small > use 16px). 
- Could you align Insights status, Total risk, and Fixable issues? Currently, the total risk is more indented to the right.

frontend/packages/insights-plugin/OWNERS
bond95: I created PR for that https://github.com/openshift/release/pull/12398

frontend/packages/insights-plugin/src/plugin.tsx
bond95: @rawagner @vojtechszocs What if we don't have any CRD?

rawagner: Another option is to implement your own detection logic and use `CustomFeatureFlag` . Example https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/plugin.ts#L76

bond95: @rawagner We had discussion with @smarterclayton and @spadgett, they suggested not to hide tab even if insights-operator is disabled.

bond95:
@doruskova I updated screenshot. About headers, `Insights` header is default one, it size is `14px`, so I made `Total risk` size same as `Fixable issues` (13px).

doruskova:
@bond95 The alignment looks great now, but you should keep 14px for every title because 13px is not an official variable and it's too small for a user. Then it should be fine :)

bond95:
@doruskova Now it's all 14px

doruskova:
@bond95 looks good to me!

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable @typescript-eslint/camelcase */
+import * as _ from 'lodash';
+
+import {
+  global_palette_blue_50,
+  global_palette_blue_300,
+  global_palette_gold_400,
+  global_palette_orange_300,
+  global_palette_red_200,
+} from '@patternfly/react-tokens';
```
vojtechszocs, line 10: Please disable `@typescript-eslint/camelcase` rule only for the relevant imports:

```ts
import * as _ from 'lodash';
/* eslint-disable @typescript-eslint/camelcase */
import {
  global_palette_blue_50,
  global_palette_blue_300,
  global_palette_gold_400,
  global_palette_orange_300,
  global_palette_red_200,
} from '@patternfly/react-tokens';
/* eslint-enable @typescript-eslint/camelcase */
```

frontend/packages/insights-plugin/src/plugin.tsx
vojtechszocs: @bond95 Assuming that Insights integration is implemented via cluster-level operator (i.e. not via OLM operator), it makes sense to have Insights plugin extensions active by default :+1:

frontend/packages/insights-plugin/src/plugin.tsx
```diff
@@ -0,0 +1,35 @@
+import {
+  Plugin,
+  DashboardsOverviewHealthPrometheusSubsystem,
+  ModelFeatureFlag,
```
vojtechszocs, line 4: Nit: unnecessary type import, since we don't need to add any feature flags.

Please remove it here and update the `ConsumedExtensions` type alias.

vojtechszocs (changes requested):

vojtechszocs:
Please import from `@patternfly/react-tokens` like

```ts
import { global_palette_blue_300 as blueInfoColor } from '@patternfly/react-tokens/dist/js/global_palette_blue_300';
```

PR #6661, opened 2020-09-17
Bug 1860985: Updated Minimum and Recommended Requirements

gnehapk:
Screenshot from 2020-09-17 15-03-13

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
+++ b/frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
@@ -12,6 +12,8 @@ export const storageClassTooltip =
   'The Storage Class will be used to request storage from the underlying infrastructure to create the backing persistent volumes that will be used to provide the OpenShift Container Storage (OCS) service.';
 export const labelTooltip =
   'The backing storage requested will be higher as it will factor in the requested capacity, replica factor, and fault tolerant costs associated with the requested capacity.';
+export const recommendedReqTooltip =
+  'The recommended configuration will allow you to expand up to 3 storage device set. The minimum configuration will require you to add additional nodes or migrate to larger nodes.';
 
 export enum defaultRequestSize {
   BAREMETAL = '1',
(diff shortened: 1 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
```diff
@@ -12,6 +12,8 @@ export const storageClassTooltip =
   'The Storage Class will be used to request storage from the underlying infrastructure to create the backing persistent volumes that will be used to provide the OpenShift Container Storage (OCS) service.';
 export const labelTooltip =
   'The backing storage requested will be higher as it will factor in the requested capacity, replica factor, and fault tolerant costs associated with the requested capacity.';
+export const recommendedReqTooltip =
+  'The recommended configuration will allow you to expand up to 3 storage device set. The minimum configuration will require you to add additional nodes or migrate to larger nodes.';
```
cloudbehl, line 16: Does the user understand device set? Can we use some laymen language to explain

cloudbehl: @shirimordechay can you look into this. ^^

shirimordechay: The terminology of "Device set" will be used in the documentation and should be aligned in the UI to avoid confusion. It will be used in other places in the console as well. I think we should keep it as it is and the user will be able to learn more in the external link documentation if necessary.

cloudbehl: Looks like we are trying to explain the internals of storage cluster YAML to the user which we don't even expose to the user from the UI.  

IMO the documentation should be aligned to UI, not the other way around.

Example: While expanding we don't say `Expand Cluster via Device set` We just say Expand Cluster and device set is what we temper internally to expand.

shirimordechay: This terminology was chosen for the UI and then used for the documentation by @tamireran. I'm not sure it's up to us to change it. I still don't see any reason to change it, maybe add "(OSD set)".

cloudbehl: Ack!

cloudbehl:
As per UX suggestion, the PR looks good to me.

/lgtm

PR #6666, opened 2020-09-17
Bug 1879940: Non admin users should load the create vm wizard

yaacov:
Don't fail to load vm wizard for user without permission to load base images

Screenshot:
screenshot-localhost_9000-2020 09 17-15_21_42


yaacov:
@rawagner @glekner @irosenzw @pcbailey please review

@matthewcarleton please review

matthewcarleton:
Talked with @jelkosz and we came up with this:
`Could not access default operating system images. Contact your administrator to gain access to these images. Otherwise provide a manual boot source below.`

yaacov:
@matthewcarleton Thanks ! updated the text.

PR #6667, opened 2020-09-17
Bug 1850569: Fix incorrect oVirt connection status message

pcbailey:
Currently, when logging in with invalid credentials for the RHV provider, the UI will display a connection successful message and then an error stating that it failed to load the VMs list. The ideal solution would be to provide a detailed error for why the connection fails, but at this time it is not possible to determine the specific cause of failures to connect to the oVirt backend. This PR changes the messages to be more generic and to solely inform the user that the connection failed.

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
+++ b/frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
@@ -7,7 +7,7 @@ type V2VProviderStatusFlags = {
   requestsVM?: boolean;
 };
 
-const hasV2vVMWareStatus = (providerCR: K8sResourceKind, flags: V2VProviderStatusFlags) => {
+const hasV2VStatus = (providerCR: K8sResourceKind, flags: V2VProviderStatusFlags) => {
   if (providerCR && flags.requestsVM) {
     return { status: V2VProviderStatus.LOADING_VM_DETAIL };
   }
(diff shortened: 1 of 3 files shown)


frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
```diff
@@ -1,6 +1,7 @@
 import { K8sResourceKind } from '@console/internal/module/k8s';
 import { getStatusPhase } from '../../selectors/selectors';
 import { V2VProviderStatus } from './constants';
+import { OVirtProviderModel } from '../../models';
 
 type V2VProviderStatusFlags = {
   hasConnectionFailed?: boolean;
```
suomiy, line 7: can we rename the `hasV2vVMWareStatus` method?

frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
```diff
@@ -14,6 +15,13 @@ const hasV2vVMWareStatus = (providerCR: K8sResourceKind, flags: V2VProviderStatu
 
   const status = V2VProviderStatus.fromPhase(getStatusPhase(providerCR));
 
+  if (
+    providerCR?.kind === OVirtProviderModel.kind &&
```
suomiy, line 19: IMO it would be preferred to use this in a place where we render it. Not here where we resolve the status

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
pcbailey: Yes! Done. =)

frontend/packages/kubevirt-plugin/src/statuses/v2v/v2vvmware-status.ts
pcbailey: Done!

PR #6669, opened 2020-09-17
Bug 1880004: Disable submit button of import forms while submitting

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-3903

**Analysis / Root cause**: Formik expects the submit handler to return a promise based on which it sets `isSubmitting` to `false`. The import forms were not returning promise from submit handler.

**Solution Description**: Return promise from submit handler of all the import forms.

Peek 2020-09-17 20-25

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


jerolimov (approved):

/lgtm

PR #6670, opened 2020-09-17
Bug 1880099: fix the console error logged while importing from dockerfile

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-4865

**Root Analysis:**
`builderImages` is undefined in the dockerfile flow and there isn't any check for that before performing `hasOwnProperty` on it

**Solution Description:**
added a check for `builderImages`

### DIFF
--- a/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
+++ b/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
@@ -61,18 +61,21 @@ const GitSection: React.FC<GitSectionProps> = ({ showSample, builderImages }) =>
           !values.application.name &&
           values.application.selectedKey !== UNASSIGNED_KEY &&
           setFieldValue('application.name', `${gitRepoName}-app`);
-        setFieldValue('image.isRecommending', true);
-        const buildTools: BuildType[] = await gitService.detectBuildTypes();
-        setFieldValue('image.isRecommending', false);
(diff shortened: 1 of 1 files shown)


rottencandy (approved):
Tested.
/lgtm

frontend/packages/dev-console/src/components/import/git/GitSection.tsx
```diff
@@ -65,7 +65,8 @@ const GitSection: React.FC<GitSectionProps> = ({ showSample, builderImages }) =>
         const buildTools: BuildType[] = await gitService.detectBuildTypes();
         setFieldValue('image.isRecommending', false);
         const buildTool = buildTools.find(
-          ({ buildType: recommended }) => recommended && builderImages.hasOwnProperty(recommended),
+          ({ buildType: recommended }) =>
+            builderImages && recommended && builderImages.hasOwnProperty(recommended),
```
rohitkrai03, line 69: If there's no `builderImages` why even run the detection?

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/git/GitSection.tsx
debsmita1: Done

rohitkrai03 (approved):
/approve

PR #6671, opened 2020-09-18
Bug 1784330: Query Browser: Fix tooltip position when there are disabled series

kyoto:
After disabling one or more data series (lines on the graph), sometimes
the tooltip would hover over a point on the Y-axis, caused by setting
the disabled series data to `{}`.

Also fixes the type of `Series` and adds a couple of missing type
definitions.


spadgett (approved):
/lgtm

PR #6674, opened 2020-09-18
Bug 1880379: Clicking on a project link in gitops application details page should navigate to topology view with that project selected

divyanshiGupta:
Jira: https://issues.redhat.com/browse/ODC-4870

Peek 2020-09-23 21-50

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/gitops/details/GitOpsDetails.tsx
+++ b/frontend/packages/dev-console/src/components/gitops/details/GitOpsDetails.tsx
@@ -1,9 +1,11 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import { Stack, StackItem, Card, CardTitle, SplitItem, Split, Label } from '@patternfly/react-core';
-import { ResourceLink, ExternalLink } from '@console/internal/components/utils';
+import { ResourceIcon } from '@console/internal/components/utils';
 import GitOpsServiceDetailsSection from './GitOpsServiceDetailsSection';
 import { GitOpsEnvironment } from '../utils/gitops-types';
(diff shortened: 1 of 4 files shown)


sahil143 (commented):
@divyanshiGupta Please add screenshots or gif for the change. It would be better to follow the PR template.

frontend/packages/dev-console/src/components/gitops/details/GitOpsDetails.tsx
```diff
@@ -56,7 +57,21 @@ const GitOpsDetails: React.FC<GitOpsDetailsProps> = ({ envs }) => {
                         )}
                       </StackItem>
                       <StackItem className="co-truncate co-nowrap">
-                        <ResourceLink kind="Project" name={env.environment} />
+                        <span className="co-resource-item">
+                          <ResourceIcon kind="Project" />
+                          {env.cluster ? (
+                            <Link
+                              className="co-resource-item__resource-name"
+                              to={`${env.cluster}/topology/ns/${env.environment}`}
```
rohitkrai03, line 65: I think there should be a check to see if the cluster URL is of the current cluster. If yes, then use relative links if not then absolute links. You can check with `route` of `console` deployment in `openshift-console` namespace.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/gitops/details/GitOpsDetails.tsx
divyanshiGupta: I think a normal user will not have access to openshift-console ns :thinking:

rohitkrai03 (approved):
/approve

PR #6676, opened 2020-09-18
Bug 1880410: convert pipeline visualization node to SVG

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3661

**Analysis / Root cause**: 
There are several rendering issues with the Pipeline Visualization Nodes (aka Tasks). This needs to be handled better and the whole thing is currently rendered through a foriegnObject inside the SVG visualization.

**Solution Description**: 
Convert pipeline visualization node to SVG.

(Should ideally be unchanged)
tmp
tmp0
tmp
tmp2

Unchanged.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


rottencandy:
/hold
Will be worked on in the next sprint.

jerolimov:
Code change looks valid, but I find some UI difference.

**In view mode:**
Before:
pr-6676-view-before
Now:
pr-6676-view-after

1. The vertical alignment is not centered anymore, can you center it by increasing the (line) height or the y-offset here?
2. The old version uses the complete width of the boxes for longer task names (first box, I repeated the task name multiple times).
3. There is a little bit "more shadow" in the old version which is shown in the new one only on hover. I like it! I would prefer if we accept the new version.

**In edit mode:**
Before:
pr-6676-edit-before
Now:
pr-6676-edit-after

1. I think the vertical alignment could also be one pixel (sorry :P) lower.
2. I don't know why, but the "really long" task name overlaps the box. Only in the edit case when it does not find the task reference!
3. The bottom border is now bigger than in the old version also when the box is not selected. Not bad, but different.

rottencandy:
Thanks for the review @jerolimov !
>1. The vertical alignment is not centered anymore, can you center it by increasing the (line) height or the y-offset here?

Updated.
tmp
 
>2. The old version uses the complete width of the boxes for longer task names (first box, I repeated the task name multiple times).

Updated the max length. I think there will be some differences here. The old version used a CSS solution(using `text-overflow` considers font width) but the new one uses a JS solution(does not consider font width), there seems to be no way to truncate text in SVG.

>3. There is a little bit "more shadow" in the old version which is shown in the new one only on hover. I like it! I would prefer if we accept the new version.

I tried to replicate the old version which also had hover shadow, albeit was very subtle. I can add the non-hover shadow as well if you think it's better.

> **In edit mode:**
>1. I think the vertical alignment could also be one pixel (sorry :P) lower.

Updated.

>2. I don't know why, but the "really long" task name overlaps the box. Only in the edit case when it does not find the task reference!

Thank you for catching this, I seem to have missed it. Updated.
tmp0

>3. The bottom border is now bigger than in the old version also when the box is not selected. Not bad, but different.
Not sure why it appears bigger, I've set it to the smallest possible width of `1px`(under `__trigger-underline` in `TaskListNode.scss`). I'll go through SVG attrs again to see if I find something.

rottencandy:
Also about the bottom border, I think there is a bit of inconsistency between browsers.
Didn't notice this earlier since I was testing mostly on Firefox, the button appears slightly different on Chromium.

Firefox vs. Chromium:
inconsistency

Should I follow the Firefox appearance, or the Chromium one?
cc: @andrewballantyne

jerolimov (approved):
Updated version looks fine for me. :+1: 

/lgtm

rottencandy:
Rebased.

jerolimov (approved):
Tested it again locally, works fine.
/lgtm

andrewballantyne:
@rottencandy The pipeline-plugin exists now, you'll need to rebase your code over there. Should be a straightforward endeavour as the folder structure should be relatively unchanged.

andrewballantyne:
/retitle [WIP] Bug 1880410: convert pipeline visualization node to SVG
/hold

Discussion with Saud has lead me to putting a WIP on this PR. There are some technical issues (beyond the rebase on the new pipeline-plugin package). Currently not at the top of our priority list. Putting on the back burner to see if we can get some time near the end of this sprint.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/StatusIcon.tsx
```diff
@@ -11,17 +12,25 @@ import {
...
-export const StatusIcon: React.FC<StatusIconProps> = ({ status, ...props }) => {
+export const StatusIcon: React.FC<StatusIconProps> = ({ status, shiftOrigin, ...props }) => {
   switch (status) {
     case runStatus['In Progress']:
     case runStatus.Running:
-      return <SyncAltIcon {...props} className="fa-spin" />;
+      return (
+        <SyncAltIcon
+          {...props}
+          className={cx('fa-spin', { 'pipeline-status-icon__spin': shiftOrigin })}
```
andrewballantyne, line 31:
```suggestion
          className={cx('fa-spin', { 'pipeline-status-icon--spin': shiftOrigin })}
```

BEM is 
 - Block (`pipeline-status-icon`)
 - Element (`pipeline-status-icon__someWrapper`) -- kinda like a sub Block
 - Modifier (`pipeline-status-icon--spin`) -- when you want it to be "modified" based on a prop

andrewballantyne (changes requested):
Having some hovering issues.

image

When I hover over an item and add a parallel item, and then mouse up over the node and slowly to the dropdown, it removes and reappears the "+" icons on the above node.

andrewballantyne:
Ah, it's the tooltip capturing the mouse-over and I think it persists at that point.

andrewballantyne:
FWIW, frontend test failure is lack of `yarn i18n`

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/StatusIcon.tsx
rottencandy: Updated.

rottencandy:
> Ah, it's the tooltip capturing the mouse-over and I think it persists at that point.

Replacing the `onMouse` handlers with `useHover` seems to fix it.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/TaskListNode.scss
```diff
@@ -5,18 +5,26 @@
...
   &__trigger-background {
-    background: var(--pf-global--BackgroundColor--light-100);
+    fill: var(--pf-global--BackgroundColor--light-100);
+    &.is-disabled {
+      fill: var(--pf-global--disabled-color--300);
+    }
+  }
+  &__trigger-underline {
+    fill: var(--pf-global--Color--100);
+    &__hover {
```
andrewballantyne, line 16:
```suggestion
    &--hover {
```

andrewballantyne (changes requested):
Tiny nit, will test it out once I finish compiling 🙂

andrewballantyne (approved):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/TaskListNode.scss
rottencandy: Updated.

andrewballantyne:
/cc @karthikjeeyar @jerolimov 

Please one of you toss a review on this as well. If I don't see any review by EOD tomorrow, I'll LGTM it.

karthikjeeyar:
Verified locally, works fine. Thanks @rottencandy 
/lgtm

PR #6679, opened 2020-09-18
Bug 1880629: Fixes support for pre-provisioned snapshots

bipuladh:
Screenshots after fix: 
Screenshot from 2020-09-18 21-14-32
Screenshot from 2020-09-18 21-13-34

cc @andybraren

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
@@ -16,7 +16,7 @@ import {
 import { referenceForModel, VolumeSnapshotKind } from '@console/internal/module/k8s';
 import { ResourceEventStream } from '@console/internal/components/events';
 import { DetailsPage, DetailsPageProps } from '@console/internal/components/factory';
-import { Status } from '@console/shared';
+import { Status, snapshotSource } from '@console/shared';
 import { volumeSnapshotStatus } from '../../status';
 
 const { editYaml, events } = navFactory;
@@ -25,11 +25,15 @@ const menuActions = [RestorePVC, ...common];
 
 const Details: React.FC<DetailsProps> = ({ obj }) => {
   const { namespace } = obj.metadata || {};
-  const { persistentVolumeClaimName: pvcName } = obj.spec?.source || {};
+  const sourceModel = obj?.spec?.source?.persistentVolumeClaimName
+    ? PersistentVolumeClaimModel
+    : VolumeSnapshotContentModel;
+  const sourceName = snapshotSource(obj);
   const size = obj.status?.restoreSize;
   const sizeBase = convertToBaseValue(size);
   const sizeMetrics = size ? humanizeBinaryBytes(sizeBase).string : '-';
   const snapshotContent = obj?.status?.boundVolumeSnapshotContentName;
+  const snapshotClass = obj?.spec?.volumeSnapshotClassName;
   return (
     <div className="co-m-pane__body">
       <SectionHeading text="Volume Snapshot Details" />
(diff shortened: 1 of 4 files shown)


andybraren:
WRT renaming the `PVC` column to `Source` in the table, I think that makes sense since the column can have both PVC's and VSC's within it. I believe OpenShift Virtualization uses the word `Source` for a similar reason in their Templates area. FYI @yuvalgalanti @shirimordechay

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot.tsx
```diff
@@ -148,10 +149,14 @@ const Row: RowFunction<VolumeSnapshotKind> = ({ key, obj, style, index, customDa
...
       </TableData>
       <TableData className={tableColumnClasses[6]}>
-        <ResourceLink
-          kind={referenceForModel(VolumeSnapshotClassModel)}
-          name={obj?.spec?.volumeSnapshotClassName}
-        />
+        {snapshotClass ? (
+          <ResourceLink
+            kind={referenceForModel(VolumeSnapshotClassModel)}
+            name={obj?.spec?.volumeSnapshotClassName}
```
cloudbehl, line 155:
```suggestion
            name={snapshotClass}
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -25,11 +25,16 @@ const menuActions = [RestorePVC, ...common];
 
 const Details: React.FC<DetailsProps> = ({ obj }) => {
   const { namespace } = obj.metadata || {};
-  const { persistentVolumeClaimName: pvcName } = obj.spec?.source || {};
+  const sourceKind = obj?.spec?.source?.persistentVolumeClaimName
+    ? PersistentVolumeClaimModel.kind
+    : referenceForModel(VolumeSnapshotContentModel);
```
TheRealJon, line 30:
```suggestion
  const sourceModel = obj?.spec?.source?.persistentVolumeClaimName
    ? PersistentVolumeClaimModel
    : VolumeSnapshotContentModel;
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot.tsx
```diff
@@ -110,8 +110,13 @@ const Row: RowFunction<VolumeSnapshotKind> = ({ key, obj, style, index, customDa
   const size = obj?.status?.restoreSize;
   const sizeBase = convertToBaseValue(size);
   const sizeMetrics = size ? humanizeBinaryBytes(sizeBase).string : '-';
-  const pvcName = obj?.spec?.source?.persistentVolumeClaimName;
+  const sourceKind = obj?.spec?.source?.persistentVolumeClaimName
+    ? PersistentVolumeClaimModel.kind
+    : referenceForModel(VolumeSnapshotContentModel);
+  const sourceName =
```
cloudbehl, line 116:
```suggestion
  const sourceName = snapshotSource(obj);
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -25,11 +25,16 @@ const menuActions = [RestorePVC, ...common];
...
+  const sourceName =
```
cloudbehl, line 31:
```suggestion
  const sourceName =snapshotSource(obj);
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -50,17 +54,17 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
...
-                namespace={namespace}
-              />
+            <dt>
+              {sourceKind === PersistentVolumeClaimModel.kind
+                ? PersistentVolumeClaimModel.label
+                : VolumeSnapshotContentModel.label}
+            </dt>
+            <dd>
+              <ResourceLink kind={sourceKind} name={sourceName} namespace={namespace} />
             </dd>
```
TheRealJon, line 64:
```suggestion
            <dt>{sourceModel.label}</dt>
            <dd>
              <ResourceLink kind={referenceForModel(sourceModel)} name={sourceName} namespace={namespace} />
            </dd>
```

frontend/packages/console-app/src/components/volume-snapshot/volume-snapshot-details.tsx
```diff
@@ -69,13 +73,17 @@ const Details: React.FC<DetailsProps> = ({ obj }) => {
...
               </>
             )}
-            <dt>Volume Snapshot Class</dt>
-            <dd data-test="details-item-value__SC">
-              <ResourceLink
-                kind={referenceForModel(VolumeSnapshotClassModel)}
-                name={obj?.spec?.volumeSnapshotClassName}
-              />
-            </dd>
+            {snapshotClass && (
```
TheRealJon, line 76: Are we sure we want to omit the entire Volume Snapshot Class detail item? Should we use an empty state value 'None' or '-'? That way the user can see that this concept exists even if it's not defined for this instance.

TheRealJon (changes requested):
Not new, but for consistency, this details page should use the `DetailsItem` component. Recommend opening a tech debt story for that.

TheRealJon (approved):
/lgtm

spadgett (approved):
/lgtm

PR #6685, opened 2020-09-19
Bug 1880698: Improve graph render performance

kyoto:
In my testing, using `groupComponent={<g />}` reduced render time by about 6% and using `VictoryTooltip` directly reduced it by a full 45%, so this PR cuts the render time in half.

These performance gains will apply to the Alerting, Metrics and Dashboards pages for both Dev and Admin perspectives.

Using `groupComponent={<g />}` means these charts will no longer support animation when data changes, but we are not using that functionality anyway.


kyoto:
FYI @spadgett, @dtaylor113, @cshinn, @sichvoge, @dlabrecq

spadgett (approved):
/lgtm

PR #6690, opened 2020-09-21
Bug 1879911: (security) dont use the hardcoded passwords

glekner:
when changing a common template, cloud-init password is initiated with a `xxxx-xxxx-xxxx` pattern

image


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -300,7 +303,9 @@ const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOpt
   const [data, isBase64] = CloudInitDataHelper.getUserData(
     toShallowJS(iGetCommonTemplateCloudInit(iTemplate))?.cloudInitNoCloud,
   );
-  const typeData = CloudInitDataHelper.toCloudInitNoCloudSource(data, isBase64);
+  // generate password
+  const updatedData = data?.replace(/(?<=password:\s+).*/, generateCloudInitPassword());
```
suomiy, line 307: you can create an instance of the helper and then rather use has/set/get methods instead

also this will not work with base64 content

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -300,7 +303,9 @@ const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOpt
...
+  const typeData = CloudInitDataHelper.toCloudInitNoCloudSource(updatedData, isBase64);
   const hasCloudInitData = !!typeData?.userData;
 
   if (hasCloudInitData) {
```
suomiy, line 311: isEmpty method can be used here instead.

and as asCloudInitNoCloudSource down bellow

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -300,7 +303,9 @@ const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOpt
   const [data, isBase64] = CloudInitDataHelper.getUserData(
```
suomiy, line 303: hasVMSettingsValueChanged can take multiple arguments

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -300,7 +303,9 @@ const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOpt
...
     toShallowJS(iGetCommonTemplateCloudInit(iTemplate))?.cloudInitNoCloud,
   );
-  const typeData = CloudInitDataHelper.toCloudInitNoCloudSource(data, isBase64);
+  // generate password
+  const updatedData = data?.replace(/(?<=password:\s+).*/, generateCloudInitPassword());
+  const typeData = CloudInitDataHelper.toCloudInitNoCloudSource(updatedData, isBase64);
   const hasCloudInitData = !!typeData?.userData;
```
suomiy, line 310: this will remove custom cloudinit storage on cloudinit change even if the templates do not include it. Also the type should not be VMWizardStorageType.UI_INPUT.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
suomiy: it is not iCloudInitStorage, but cloudInitStorage. Please use `iGetStorages` to get `iCloudInitStorage`.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/cloud-init-data-helper.tsx
```diff
@@ -23,6 +23,15 @@ export const formAllowedKeys = new Set([
   CloudInitDataFormKeys.SSH_AUTHORIZED_KEYS,
 ]);
 
+export const generateCloudInitPassword = (length = 12) => {
+  const charset = 'abcdefghijklmnopqrstuvwxyz0123456789';
+  let password = '';
+  for (let i = 0, n = charset.length; i < length; ++i) {
```
suomiy, line 29: there is already a method called `getRandomChars`

glekner:
added `VMWizardStorageType.TEMPLATE_CLOUD_INIT` to know if to delete the cloud init volume when needed.
@suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -313,29 +327,38 @@ const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOpt
...
+            updatedCloudInitHelper.asCloudInitNoCloudSource(),
+          )
           .asResource(),
       }),
     );
-  } else if (iCloudInitStorage && !hasCloudInitData) {
+  } else if (
+    cloudInitStorage &&
+    iGet(cloudInitStorage, 'type') === VMWizardStorageType.TEMPLATE_CLOUD_INIT &&
+    updatedCloudInitHelper.isEmpty()
```
suomiy, line 350: this is already checked in the first if

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -286,24 +289,35 @@ const flavorUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) =>
 const commonTemplatesUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
   const state = getState();
```
suomiy, line 290: this updater should be ignored when using a user template

suomiy (changes requested):

glekner:
* fixed suggestions and moved to a new file @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
```diff
@@ -0,0 +1,93 @@
...
+        id,
+        CloudInitField.IS_FORM,
+        false,
+      ),
+    );
+    dispatch(
+      vmWizardInternalActions[InternalActionType.UpdateStorage](id, {
+        id: iGet(iCloudInitStorage, 'id'),
+        type: iGet(iCloudInitStorage, 'type') || VMWizardStorageType.TEMPLATE_CLOUD_INIT,
+        disk: new DiskWrapper()
```
suomiy, line 66: you can use the disk and volume you already have and `setType` This way user changes in bus/etc will be preserved.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
```diff
@@ -0,0 +1,93 @@
...
+  );
+
+  const relevantOptions = iGetRelevantTemplateSelectors(state, id);
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iTemplate = iCommonTemplates && iGetRelevantTemplate(iCommonTemplates, relevantOptions);
+  const cloudInitHelper = new CloudInitDataHelper(
+    toShallowJS(iGetCommonTemplateCloudInit(iTemplate)),
+  );
+  // generate new password if the common template has a password for cloud ionit
+  if (cloudInitHelper.get(CloudInitDataFormKeys.PASSWORD)) {
```
suomiy, line 52: I thought we want to use the cloud config even if the common template does not have the password specified.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
```diff
@@ -0,0 +1,93 @@
...
+      id,
+      VMSettingsField.OPERATING_SYSTEM,
+      VMSettingsField.FLAVOR,
+      VMSettingsField.WORKLOAD_PROFILE,
+    ) ||
+    iUserTemplate
+  ) {
+    return;
+  }
+  const iCloudInitStorage = iGetStorages(state, id).find((stor) =>
```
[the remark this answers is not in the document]
jelkosz: I dont think we should preserve it. If the user changes the template under it, that changes everything...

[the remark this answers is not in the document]
suomiy: after some discussion, let's just override it with the template one on change for simplicity

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
glekner: but this change will only occur in the first step of the wizard, its only right that the disk and volume will reset.

glekner:
fixed @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const relevantOptions = iGetRelevantTemplateSelectors(state, id);
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iTemplate = iCommonTemplates && iGetRelevantTemplate(iCommonTemplates, relevantOptions);
+  const cloudInitHelper = new CloudInitDataHelper(
+    toShallowJS(iGetCommonTemplateCloudInit(iTemplate)),
+  );
+  // generate new password if the common template has a password for cloud ionit
+  if (!cloudInitHelper.isEmpty()) {
+    cloudInitHelper.set(CloudInitDataFormKeys.PASSWORD, generateCloudInitPassword());
```
suomiy, line 54: should set password only if it includes a password

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-common-templates-updater.ts
suomiy: ok then, let's reset everything

PR #6692, opened 2020-09-21
Bug 1881898: fix Quickstart subtitle block

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-4779

# Problem
The sub title of the main and step headers are misaligned when the header text is wrapping to another line

# Solution
- Put Step no inside h3 and within inline-block wrapper
- Put Duration inside Header and within inline-block wrapper

# Screenshots
Screenshot from 2020-10-01 13-43-29
Screenshot from 2020-10-01 13-42-57
Screenshot from 2020-10-01 13-42-47
Screenshot from 2020-10-01 13-42-34

# Tests
Altered tests QuickStartHeader.spec.scss

# Browser conformance

Chrome, Firefox

### DIFF
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
@@ -78,10 +78,10 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
               size="xl"
               style={{ marginRight: 'var(--pf-global--spacer--md)' }}
             >
-              {quickStart?.spec.displayName}
-            </Title>
-            <Title headingLevel="h6" size="md" className="text-secondary">
-              {`${quickStart?.spec.duration} minutes`}
+              {quickStart?.spec.displayName}{' '}
+              <small className="co-quick-start-drawer__duration text-secondary">
+                {`${quickStart?.spec.duration} minutes`}
+              </small>
             </Title>
           </div>
           <DrawerActions>
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import cx from 'classnames';
-import { Title, WizardNavItem, Text, TextVariants } from '@patternfly/react-core';
+import { Title, WizardNavItem } from '@patternfly/react-core';
 import { CheckCircleIcon, TimesCircleIcon } from '@patternfly/react-icons';
 import { QuickStartTaskStatus } from '../utils/quick-start-types';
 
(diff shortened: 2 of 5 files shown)


rohitkrai03 (changes requested):
The bug also mentions that the duration shown after the quick start title should not wrap around to next line. In your screenshots I only see the issue resolved for task titles.

abhinandan13jan:
> The bug also mentions that the duration shown after the quick start title should not wrap around to next line. In your screenshots I only see the issue resolved for task titles.

@rohitkrai03 I've added the Header part changes and also updated the screenshots. However, it looks a little odd in some cases.

rohitkrai03:
> > The bug also mentions that the duration shown after the quick start title should not wrap around to next line. In your screenshots I only see the issue resolved for task titles.
> 
> @rohitkrai03 I've added the Header part changes and also updated the screenshots. However, it looks a little odd in some cases.

@abhinandan13jan Yeah, looking at the screenshots it seems the text is wrapping at wrong places. Not sure if using `<Text>` is a good idea here. @christianvogt WDYT?

christianvogt:
image

Wrap the duration in an element with `display: inline-block`

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
```diff
@@ -78,10 +80,12 @@ const QuickStartDrawer: React.FC<QuickStartDrawerProps> = ({
...
-              {quickStart?.spec.displayName}
-            </Title>
-            <Title headingLevel="h6" size="md" className="text-secondary">
-              {`${quickStart?.spec.duration} minutes`}
+              {quickStart?.spec.displayName}{' '}
+              <span className="co-quick-start-drawer__duration">
+                <Text component={TextVariants.small} className="text-secondary">
+                  {`${quickStart?.spec.duration} minutes`}
+                </Text>
+              </span>
```
christianvogt, line 88: Why are you using `Text` ? Fairly certain `Text` is only useful inside of `TextContent`.

```suggestion
              <small className="co-quick-start-drawer__duration text-secondary">
                  {`${quickStart?.spec.duration} minutes`}
              </span>
```

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
```diff
@@ -71,12 +71,13 @@ const QuickStartTaskHeader: React.FC<QuickStartTaskHeaderProps> = ({
       <Title headingLevel="h3" size={size} className={classNames}>
         <TaskIcon taskIndex={taskIndex} taskStatus={taskStatus} isActiveTask={isActiveTask} />
         {title}
+        {isActiveTask && subtitle && (
+          <Text component={TextVariants.a} className="text-secondary">
+            {' '}
+            {subtitle}
+          </Text>
```
christianvogt, line 78: You created a class name never used it. Also you need to remove the space from the inline-block element.
```suggestion
          <>
            {' '}
            <small className="co-quick-start-task-header__subtitle text-secondary">
              {subtitle}
            </small>
          </>
```

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartTaskHeader.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawer.tsx
abhinandan13jan: I was using Text as I simply moved a pre-existing block inside the Header without modifying it

rohitkrai03 (approved):
/lgtm

PR #6699, opened 2020-09-21
Bug 1881314: fix application selector in eventsource form

karthikjeeyar:
**Fixes**: https://issues.redhat.com/browse/ODC-3960

**Problem**:
In event source page, the application name is not auto populated when a event source is selected for the first time.
**Solution**:
Prevent resetting the application name if it already has a recommended name.

**Screenshots**:

**Scenario 1:**
Given: Namespace has application groups.
If User chooses unassigned option from the dropdown.
unassigned-key

**Scenario 2:**
**Given:** Namespace has **no** application groups.
If User manually enters the application name
user-application

**Scenario 3:**
**Given:** Namespace has **no** application groups.
If User removes the recommended application name
user-removing-recommendation

**Tests**:
image

cc: @invincibleJai

### DIFF
--- a/frontend/packages/knative-plugin/src/components/add/event-sources/EventSourcesSelector.tsx
+++ b/frontend/packages/knative-plugin/src/components/add/event-sources/EventSourcesSelector.tsx
@@ -3,11 +3,10 @@ import * as _ from 'lodash';
 import { useFormikContext, FormikValues } from 'formik';
 import { ItemSelectorField } from '@console/shared';
 import FormSection from '@console/dev-console/src/components/import/section/FormSection';
-import { NormalizedEventSources } from '../import-types';
+import { NormalizedEventSources, EVENT_SOURCES_APP } from '../import-types';
 import { KNATIVE_EVENT_SOURCE_APIGROUP } from '../../../const';
 import { getEventSourceModels } from '../../../utils/fetch-dynamic-eventsources-utils';
 import { isKnownEventSource, getEventSourceData } from '../../../utils/create-eventsources-utils';
-import { CREATE_APPLICATION_KEY } from '@console/dev-console/src/const';
 
 interface EventSourcesSelectorProps {
   eventSourceList: NormalizedEventSources;
@@ -17,7 +16,7 @@ const EventSourcesSelector: React.FC<EventSourcesSelectorProps> = ({ eventSource
   const eventSourceItems = Object.keys(eventSourceList).length;
   const {
     values: {
-      application: { selectedKey },
+      application: { name: applicationName, selectedKey },
       type,
     },
     setFieldValue,
@@ -26,6 +25,18 @@ const EventSourcesSelector: React.FC<EventSourcesSelectorProps> = ({ eventSource
     setErrors,
     setStatus,
   } = useFormikContext<FormikValues>();
+
+  const [recommended, setRecommended] = React.useState(false);
+
+  React.useEffect(() => {
+    if (!applicationName && !selectedKey && !recommended) {
+      setFieldValue('application.name', EVENT_SOURCES_APP);
+      setFieldTouched('application.name', true);
+      setRecommended(true);
+      validateForm();
+    }
+  }, [selectedKey, applicationName, validateForm, recommended, setFieldValue, setFieldTouched]);
+
   const handleItemChange = React.useCallback(
     (item: string) => {
       if (item !== type) {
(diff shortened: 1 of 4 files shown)


frontend/packages/dev-console/src/components/import/app/ApplicationSelector.tsx
```diff
@@ -30,7 +30,7 @@ const ApplicationSelector: React.FC<ApplicationSelectorProps> = ({
   const onDropdownChange = (key: string, application: string) => {
     setFieldValue('application.selectedKey', key);
     setFieldTouched('application.selectedKey', true);
-    setFieldValue('application.name', sanitizeApplicationValue(application, key));
+    application && setFieldValue('application.name', sanitizeApplicationValue(application, key));
```
invincibleJai, line 33: Shall we add check for `setFieldTouched` as well as we have here?

rottencandy:
If we select "Create application" and change the application name, it gets reset when we select a different source.
tmp

karthikjeeyar:
@invincibleJai @rottencandy Thanks for the review, i have addressed the comments.

rottencandy (approved):
/lgtm

divyanshiGupta:
@karthikjeeyar selecting `unassigned` from application-dropdown doesn't work as expected anymore
Peek 2020-09-22 23-38

divyanshiGupta:
Also I see that in your gif app-name changes wrt to event source type but I dont see that happening as you are setting `event-sources-app` for all.
Peek 2020-09-22 23-44

karthikjeeyar:
@divyanshiGupta I have fixed the unassigned option selection and updated the screenshots with various scenarios.

divyanshiGupta:
@karthikjeeyar I dont see the updated screenshots I still see the previous gif.

karthikjeeyar:
@divyanshiGupta Oops, forget to hit the update button :facepalm: , I have updated now. Please take a look.

divyanshiGupta:
Peek 2020-09-23 18-48
@karthikjeeyar it doesnt seem like a big issue but seems weird as after selecting the create application option again it should be populated with the the recommended app name.

divyanshiGupta:
@karthikjeeyar I still see this issue
Peek 2020-09-23 19-04

karthikjeeyar:
@divyanshiGupta Previously, we were recommending name based on the type of the selection of the event, so whenever you change the type a new name will be generated, but what we need is a  generic app name throughout the event sources, Thats the reason i have added `event-sources-app`. 
cc: @invincibleJai

karthikjeeyar:
@divyanshiGupta When the user chooses `unassigned` and then switches back to `Create Application` then i gues we need to clear the input box for the user to type his own input instead of recommending. so updated the code to match the current behaviour
Prompting_user

divyanshiGupta:
@karthikjeeyar now I dont see app-name getting populated on when an event source is selected for the first time :(
Peek 2020-09-24 14-31

karthikjeeyar:
@divyanshiGupta  It seems like, the appllication selector is always loading the dropdown component first (onLoad function is used to see if the application exists), when the application dropdown is loaded, then onDropdownChange handler resets the application.name, but the recomendation happens way before this in event selector component. So the solution i am thinking here is to, use useEffect hook in the EventSourceSelector to recommend a name when the selectedKey is set to CREATE_APPLICATION or if there isn't any name in the input field.
cc: @invincibleJai

divyanshiGupta:
@karthikjeeyar ideally we should refactor the `ApplicationSelector` to handle this edge case as it might occur elsewhere also but since we are near to the code freeze it will not be a good idea to do so since we will be at risk of introducing new bugs so for now we can go with the solution that you have suggested above.

divyanshiGupta:
@karthikjeeyar the changes to `EventSourcesSelector` solves this issue, I tested the event-source flows with just the changes in `EventSourcesSelector` and it seemed to work fine. Just wanted to know why do you still have the changes in `ApplicationSelector`? Are those still needed?

karthikjeeyar:
@divyanshiGupta Those were just over cautious checks before resetting the fields, Not needed to be honest. I have removed it.

frontend/packages/knative-plugin/src/components/add/event-sources/EventSourcesSelector.tsx
```diff
@@ -26,6 +26,19 @@ const EventSourcesSelector: React.FC<EventSourcesSelectorProps> = ({ eventSource
...
     setStatus,
   } = useFormikContext<FormikValues>();
+
+  const recommendApplicationName = React.useCallback(() => {
+    setFieldValue('application.name', 'event-sources-app');
+    setFieldTouched('application.name', true);
+  }, [setFieldTouched, setFieldValue]);
+
+  React.useEffect(() => {
+    if (!applicationName && !selectedKey) {
```
divyanshiGupta, line 36:
```suggestion
    if (!applicationName && (!selectedKey || selectedKey === CREATE_APPLICATION_KEY) ) {
```

invincibleJai:
@karthikjeeyar it works well overall in below scenario i see issue

1. Created a ns
2. Add -> Event Source
3. App name is auto-populated , however i can't clear it

Sep-29-2020 21-16-28

karthikjeeyar:
@invincibleJai Fixed the issue.

frontend/packages/knative-plugin/src/components/add/event-sources/EventSourcesSelector.tsx
```diff
@@ -26,6 +25,21 @@ const EventSourcesSelector: React.FC<EventSourcesSelectorProps> = ({ eventSource
     setErrors,
     setStatus,
   } = useFormikContext<FormikValues>();
+
+  const [recommended, setRecommended] = React.useState(false);
+  const recommendApplicationName = React.useCallback(() => {
+    setFieldValue('application.name', EVENT_SOURCES_APP);
+    setFieldTouched('application.name', true);
+  }, [setFieldTouched, setFieldValue]);
```
invincibleJai, line 33: nit: do we need it as used only in one place now and can be used inside `useEffect`

invincibleJai:
/lgtm
/approve

Verified the changes locally , works as expected

PR #6701, opened 2020-09-22
Bug 1881097: Fix kubevirt-plugin gating job

irosenzw:
Change to 'kubevirt' namespace

### DIFF
--- a/frontend/packages/kubevirt-plugin/integration-tests/deploy-kubevirt-gating.sh
+++ b/frontend/packages/kubevirt-plugin/integration-tests/deploy-kubevirt-gating.sh
@@ -41,7 +41,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: kubevirt-config
-  namespace: kubevirt-hyperconverged
+  namespace: kubevirt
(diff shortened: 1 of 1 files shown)


yaacov:
/lgtm

p.s.
this will only work on upstream env created by hyperconverge operator 
it will not work on downstream env (using different namespace IIRC "openshift-cnv" ? )

irosenzw:
This is for upstream purpose for now. Once it will be stable we can consider moving it downstream

yaacov:
p.p.s.
we have a constant for the openshift virtualization namespace:
https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/integration-tests/tests/utils/constants/common.ts#L63
and we should be able to override it from env variable like we do for the storage class:
https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/integration-tests/tests/utils/constants/common.ts#L3

e.g.
`export const { KUBEVIRT_PROJECT_NAME = 'openshift-cnv' } = process.env;`

so we can run the tests on different envs

PR #6704, opened 2020-09-22
Bug 1881347: fix createVMWizard onClose destructor

suomiy:
- run it on unMount so all providers get a chance for a cleanup
- make onClose idempotent, so multiple calls are not stacked
- use useRef instead so the value is propagated to the destructor
- remove code duplication for checking checking the common data changes
- disable relevant fields until cnvBaseImages are loaded plus allow for non admin user as well

please take a look 
@rawagner @yaacov @irosenzw @glekner @pcbailey

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
@@ -83,7 +83,29 @@ type CreateVMWizardComponentProps = {
 } & { [k in ChangedCommonDataProp]: any };
 
 const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props) => {
-  const [closed, setClosed] = React.useState(false);
+  const closed = React.useRef(false);
+
+  const onClose = React.useCallback(
+    (disposeOnly?: boolean) => {
+      if (closed.current) {
+        return;
+      }
+      closed.current = true;
+      props.onClose(disposeOnly);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  );
+  React.useEffect(() => {
+    props.onInitialize();
+
+    return () => {
+      onClose(true);
+    };
+    // constuctor/desctructor logic - should run only once
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   const commonTemplates = React.useMemo(
     () => immutableListToShallowJS(iGetLoadedData(props.commonTemplates)),
     [props.commonTemplates],
@@ -91,17 +113,6 @@ const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props) 
   const [dataVolumePVCs, dataVolumePVCsLoaded, dataVolumePVCsLoadError] = usePVCBaseImages(
     commonTemplates,
   );
-
-  React.useEffect(() => {
-    if (!(props[VMWizardProps.isProviderImport] && props[VMWizardProps.isCreateTemplate])) {
-      props.onInitialize();
-    } else {
-      console.error('It is not possible to make an import VM template'); // eslint-disable-line no-console
-      setClosed(true);
-    }
-    // No need to run the effect multiple times
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
 
   // Store previuse props
   const prevProps = usePrevious<CreateVMWizardComponentProps>(props);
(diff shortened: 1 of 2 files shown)


yaacov:
@suomiy hi,
This PR is about the BZ "VM Import Wizard does not call Import provider cleanup"
But is does much more, can you make a new PR that only fixes the BZ issue (and wait with this PR for 4.7), or remove the unneeded parts from this PR ?

https://bugzilla.redhat.com/show_bug.cgi?id=1881347

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -444,6 +429,16 @@ export const CreateVMWizardPageComponent: React.FC<CreateVMWizardPageComponentPr
...
 
+  const loadedCommonTemplates = React.useMemo(
+    () =>
+      userMode === VMWizardMode.IMPORT // make empty for import so no PVCBaseImages get loaded
+        ? []
+        : immutableListToShallowJS(iGetLoadedData(commonTemplates)),
+    [commonTemplates, userMode],
+  );
+  // TODO integrate the list of watches into the redux store to prevent unnecessary copying of data
+  const openshiftCNVBaseImages = useWatchedResultAsObject(usePVCBaseImages(loadedCommonTemplates));
```
rawagner, line 440: `usePVCBaseImages` already memoizes the output object

rawagner: ah, this is used as converter to object. But why hook form ?

suomiy:
@yaacov these are very small changes which fix the same part of the code and are part of the same regression. I updated BZ to include these changes as well, so we can proceed with these changes atomically.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
suomiy: because they use `useMemo` hook inside and thus should be called as a hook unconditionally.

rawagner: sure, but why are you creating a hook which only wraps `useMemo` ? Just use `useMemo` directly.

suomiy: this is just to simplify things and have one call which does what is required. Having two calls beets the point of extracting this functionality a bit IMO. 

We can also inline the functionality here, if you think such small optimization doesn't make sense.

yaacov:
> @yaacov these are very small changes which fix the same part of the code and are part of the same regression. I updated BZ to include these changes as well, so we can proceed with these changes atomically.

We are 2 days from code freeze, we should not insert so much code because if
we insert a new bug we will not have time to fix it, we can do this after the code freeze.

I understand that BZ is about k8s import objects not cleaning up after the Import Wizard closes, this PR (or a new one for BZ 1881347) shroud fix the Import issue. Re-factoring stuff can be done in a separate PR (or in this one, and open a new one for the BZ)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
yaacov: > this is just to simplify things 

No, it is adding cyclomatic complexity (because we use useMemo twice) and (119-65) code lines and one file.
It is also harder for a newcomers to the code because they will have one more object to learn.

rawagner: I would prefer inlining :+1:

suomiy: > No, it is adding cyclomatic complexity (because we use useMemo twice) and (119-65) code lines and one file.

we need to use `useMemo` twice either way, so the cyclomatic complexity will not change in total, unless I modify the `usePVCBaseImages` function.

> It is also harder for a newcomers to the code because they will have one more object to learn.

this is not so easy topic and for longer discussion :)

anyway, this was just a small refactoring - np, we can inline it

yaacov: > this is not so easy topic and for longer discussion :)

:smile:

suomiy:
moving the rest to 6.1, left the destructor logic

PR #6705, opened 2020-09-22
Bug 1880008: general step text enhancements

glekner:


matthewcarleton:
Can we also cover this from the bz in this PR?
```
Update text when no image is available

Default operating system image not available. Define a boot source manually below or request your administrator to define one for this operating system via the PVC upload form.

When user is told to go to storage step to edit disk we tell them the name of the disk rather than say "mounted disk" 

example:

Enter URL below or edit the rootdisk in the storage step
```

glekner:
added texts @matthewcarleton

matthewcarleton:
Lgtm!

jelkosz:
good, 
/hold cancel

jelkosz:
come on Kobi, why are you always faster than me?!

PR #6709, opened 2020-09-22
Bug 1879054: UI should display password in cloud-init

glekner:


jelkosz:
only a text change and the text change is good so
/lgtm

PR #6711, opened 2020-09-22
Bug 1881541: Add pipeline workspace section also to Add Trigger modal

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4186
https://bugzilla.redhat.com/show_bug.cgi?id=1881541

**Analysis / Root cause**: 
Workspaces are used in most (all?) predefined Pipelines of the Pipeline Operators 1.1.1. These are automatically used when import a Deployment and enable "Add Pipelines". Pipeline workspaces are already support in the "Start Pipeline" modal but not in the "Add Trigger" modal.

**Solution Description**: 
Add the workspaces section also to the "Add Trigger" modal and move some type informations and helpers from the Start pipeline specific part to the more generic implementation.

odc-4186 webm

Small changes, coverage not changed.

**Test setup:**
Small unit test changes and type changes. To reproduce this fix create a new pipeline and compare Start pipeline and Add trigger modals.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
+++ b/frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
@@ -60,7 +60,7 @@ export const getLogSnippet = (pipelineRun: PipelineRun): PipelineRunErrorDetails
   }
 
   const containerName = failedTaskRun.status.steps?.find(
-    (step: PLRTaskRunStep) => step.terminated.exitCode !== 0,
+    (step: PLRTaskRunStep) => step.terminated?.exitCode !== 0,
   )?.container;
 
   if (!failedTaskRun.status.podName || !containerName) {
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
@@ -1,5 +1,10 @@
 import { FormikValues } from 'formik';
-import { PipelineParam } from '../../../../utils/pipeline-augment';
+import {
+  PipelineParam,
+  VolumeTypeConfigMaps,
+  VolumeTypePVC,
+  VolumeTypeSecret,
+} from '../../../../utils/pipeline-augment';
 
 export type PipelineModalFormResource = {
   name: string;
@@ -11,8 +16,27 @@ export type PipelineModalFormResource = {
   };
 };
 
+export type PipelineModalFormWorkspace = {
+  name: string;
+  type: string;
+  data:
+    | {
+        emptyDir: {};
+      }
+    | {
+        secret: VolumeTypeSecret;
+      }
+    | {
+        configMap: VolumeTypeConfigMaps;
+      }
+    | {
+        persistentVolumeClaim: VolumeTypePVC;
+      };
+};
+
 export type CommonPipelineModalFormikValues = FormikValues & {
   namespace: string;
   parameters: PipelineParam[];
   resources: PipelineModalFormResource[];
+  workspaces: PipelineModalFormWorkspace[];
 };
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
@@ -4,9 +4,9 @@ import { SecretModel, ConfigMapModel } from '@console/internal/models';
 import { DropdownField } from '@console/shared';
 import FormSection from '../../../import/section/FormSection';
 import { VolumeTypes } from '../../const';
-import { PipelineRunWorkspaceFormEntry } from '../start-pipeline/types';
 import PVCDropdown from './PVCDropdown';
 import MultipleResourceKeySelector from './MultipleResourceKeySelector';
+import { PipelineModalFormWorkspace } from './types';
 
 const getVolumeTypeFields = (volumeType: VolumeTypes, index: number) => {
   switch (VolumeTypes[volumeType]) {
@@ -44,7 +44,7 @@ const getVolumeTypeFields = (volumeType: VolumeTypes, index: number) => {
 
 const PipelineWorkspacesSection: React.FC = () => {
   const { setFieldValue } = useFormikContext<FormikValues>();
-  const [{ value: workspaces }] = useField<PipelineRunWorkspaceFormEntry[]>('workspaces');
+  const [{ value: workspaces }] = useField<PipelineModalFormWorkspace[]>('workspaces');
   return (
     workspaces.length > 0 && (
       <FormSection title="Workspaces" fullWidth>
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/__tests__/utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/__tests__/utils.spec.ts
@@ -96,6 +96,7 @@ describe('PipelineAction testing getPipelineRunFromForm', () => {
       namespace: 'corazon',
       parameters: [],
       resources: [],
+      workspaces: [],
     };
     const labels: { [key: string]: string } = {
       anotherlabel: 'another-label-value',
@@ -119,6 +120,7 @@ describe('PipelineAction testing getPipelineRunFromForm', () => {
       namespace: 'corazon',
       parameters: [],
       resources: [],
+      workspaces: [],
     };
     const labels: { [key: string]: string } = {
       anotherlabel: 'another-label-value',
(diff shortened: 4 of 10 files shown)


frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
```diff
@@ -60,7 +60,7 @@ export const getLogSnippet = (pipelineRun: PipelineRun): PipelineRunErrorDetails
   }
 
   const containerName = failedTaskRun.status.steps?.find(
-    (step: PLRTaskRunStep) => step.terminated.exitCode !== 0,
+    (step: PLRTaskRunStep) => step.terminated?.exitCode !== 0,
```
andrewballantyne, line 63: Good catch... Wish TS warned that this failed the type 😕

jerolimov: We need to finetune eslint for this, but will have too much errors then.. :thinking: 

It crashs locally and adding this here is a workaround to creating a separate Jira, BZ and PR for this one char :smile:

jerolimov:
Only rebased

andrewballantyne (approved):

karthikjeeyar:
@jerolimov can we disable the Add button untils we add a valid workspace entry ?

image

jerolimov:
@karthikjeeyar good catch, will check this..

divyanshiGupta:
@jerolimov since we now have the workspace section in both start-pipeline and add-trigger forms you can add 
```
workspaces: yup.array().of(
    yup.object().shape({
      type: yup.string().required('Required'),
      data: volumeTypeSchema,
    }),
  ),
```
in `commonPipelineSchema` as both the forms require these validations now.

jerolimov:
@karthikjeeyar @divyanshiGupta You're right. :+1: Moved the `workspaces` yup validation to `commonPipelineSchema` (and rebased the PR).

I tested it locally also with the changes from #6673, then it works fine. Please take a look again, thanks.

andrewballantyne:
@jerolimov Need a rebase.

jerolimov:
@andrewballantyne @divyanshiGupta Rebased and tested locally. Needed to solve some type conflicts.

divyanshiGupta:
@jerolimov I see there are some inconsistencies with the types, though it doesn't seem to break anything but it will be good to clean it up
So we have three types
```
export type PipelineWorkspace = Param;
```
Type of pipeline workspace.
```
export type PipelineModalFormWorkspace = {
  name: string;
  type: string;
  data: { [volumeType: string]: VolumeTypeSecret | VolumeTypeConfigMaps | VolumeTypePVC | {} };
};
```
Type of workspace data stored in the form state, so here you will see the inconsistency, as in your PR you dont have the data attribute but this is how the workspace values are being stored in the form.

```
export interface PipelineRunWorkspace extends Param {
  [volumeType: string]: VolumeTypeSecret | VolumeTypeConfigMaps | VolumeTypePVC | {};
}
```
Type of pipeline-run workspace.

I also see that you have changed PipelineRunWorkspaceFormEntry ->  PipelineModalFormWorkspace in this PR but you havent cleaned the code for PipelineRunWorkspaceFormEntry which is in start-pipeline/types added by you in https://github.com/openshift/console/pull/6614.

Please fix these issues and also just go through the flow once to see if there is some redundant code still left to be cleaned up.Thanks!

jerolimov:
@divyanshiGupta You are absolute right here, and explained all the types already. Thanks for the long response. :+1: 

Anyway I want to give you my perspective on the different types. Yes we have currently three types

* PipelineWorkspace contains only a name, see Tekton [Pipeline#workspaces spec], (https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#specifying-workspaces), so Param is correct here.
* PipelineRunWorkspace contains also the linked volume, see Tekton [PipelineRun#workspaces](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#specifying-workspaces spec), so we could not reuse the same type as for Pipelines.

I'm would prefer here also if we would change this type (which accepts currently all keys) to a more specific type like this:

Instead of:

```tsx
export interface PipelineRunWorkspace extends Param {
  [volumeType: string]: VolumeTypeSecret | VolumeTypeConfigMaps | VolumeTypePVC | {};
}
```

This (untested proof of concept):

```tsx
export type VolumeSource = {
  emptyDir: {},
} | {
  configmap: VolumeTypeConfigMaps,
} | {
  secret: VolumeTypeSecret,
} | {
  persistentVolumeClaim: VolumeTypePVC,
};

export interface PipelineRunWorkspace extends Param & VolumeSource;
```

* What do you think? Should we change this as part of this PR or should I create new PR for 4.7? I would wait for the next version (and if you want I can prepare such a PR now). For 4.7, I expect that we need also workspace types for Tasks and TaskRuns which are not matching the existing one.

And finally we have the type `PipelineModalFormWorkspace` (which I only moved and yes, renamed). Yes, I introduced the type in #6614, but it only describes the code which already exist and is still used to select a workspace in this modal. 

The form data are saved before both PRs and after them in a structure like this (`{ name: '..', type: 'configmap', data: { configmap: { ... } } }`). The mapping from this `PipelineModalFormWorkspace` to a `PipelineRunWorkspace` was done in `getPipelineRunWorkspaces`

I see here two options:

* Remove the type `PipelineModalFormWorkspace` and try to save the form data directly in an `PipelineRunWorkspace` instance. But then I need to change the form as well which saves the data within this structure. See [PiplelineWorkspacesSection.tsx](https://github.com/openshift/console/blob/5b1ce934a188137c9c07f9723fd6c6c93c42680e/frontend/packages/dev-console/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx). Changing this will need to change more than expected for this PR. I can do this, but would also prefer if we do a cleanup in 4.7 if you like.
* The benefit (which we currently do not support here! :-/) of an additional form type is: It allows the user to select ConfigMap, then select a specific ConfigMap, switch the type to Secrets, go back to ConfigMap and his selection is still available. For such features we need a Form type which is not matching the final API.

I would prefer a type like this here:

```tsx
export type PipelineModalFormWorkspace = {
  name: string;
  selectedType: string;
  emptyDir: {};
  configmap: VolumeTypeConfigMaps;
  secret: VolumeTypeSecret;
  persistentVolumeClaim: VolumeTypePVC;
};
```

But this requires also a refactoring of the form itself, where I thought this is too much now. While #6614 just introduced or fixed types, this PR only moves types (and added the form and form validation for the second modal).

What's your opinion? How much should I refactor in this PR and what should we keep back (and solve together?) in 4.7? :smile:

divyanshiGupta:
@jerolimov I will suggest to make only absolute necessary changes in this PR and we can do the refactor work in 4.7 as the code freeze is this Friday.

frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
```diff
@@ -11,8 +16,15 @@ export type PipelineModalFormResource = {
   };
 };
 
+export type PipelineModalFormWorkspace = {
+  name: string;
+  type: string;
+  [volumeType: string]: VolumeTypeSecret | VolumeTypeConfigMaps | VolumeTypePVC | {};
```
divyanshiGupta, line 22:
```suggestion
 data: { [volumeType: string]: VolumeTypeSecret | VolumeTypeConfigMaps | VolumeTypePVC | {}; };
```

frontend/packages/dev-console/src/components/pipelines/modals/common/types.ts
```diff
@@ -11,8 +16,27 @@ export type PipelineModalFormResource = {
...
+  type: string;
+  data:
+    | {
+        emptyDir: {};
+      }
+    | {
+        secret: VolumeTypeSecret;
+      }
+    | {
+        config: VolumeTypeConfigMaps;
```
divyanshiGupta, line 30:
```suggestion
        configMap: VolumeTypeConfigMaps;
```

andrewballantyne:
> /hold

I'm a little annoyed the type inconsistency didn't cause a TS error 😕

jerolimov:
> I'm a little annoyed the type inconsistency didn't cause a TS error

Yep, me too. Thats a big downside of `resourceNameField` in [PiplelineWorkspacesSection.tsx](https://github.com/openshift/console/blob/2d6c745f538c80b15a8c87a681382aef5263405a/frontend/packages/dev-console/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx)

Maybe we should prefer useField or some other more type-safe ways to handle forms with formik.. We should take some time for such ideas / things.

PR #6714, opened 2020-09-22
Bug 1881659: fix guided tour alert to not show up if skipped once

nemesis09:
**Fixes:**
https://issues.redhat.com/browse/ODC-4799

**Analysis / Root cause:**
Even though the user chooses to skip the guided tour of the dev perspective, the system will nag them again if they happen to follow these steps.

**Solution Description:**
User should not see guided tour if they have previously skipped or completed it.

**Screens:**
_Before_
guided-tour-before

_After_
guided-tour

**Test Coverage**
Screenshot from 2020-09-23 01-55-18

**Browser Conformance**
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/tour/tour-context.ts
+++ b/frontend/packages/console-app/src/components/tour/tour-context.ts
@@ -1,4 +1,12 @@
-import { createContext, useReducer, Reducer, Dispatch, ReducerAction } from 'react';
+import {
+  createContext,
+  useReducer,
+  useState,
+  Reducer,
+  Dispatch,
+  ReducerAction,
+  useEffect,
+} from 'react';
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
 import { useSelector } from 'react-redux';
@@ -16,9 +24,16 @@ import {
 } from './utils';
 import { TourDataType, Step } from './type';
 
-export const tourReducer = (state: TourState, action: TourActions) => {
+type TourStateAction = { type: TourActions; payload?: { completed?: boolean } };
+export const tourReducer = (state: TourState, action: TourStateAction) => {
   const { stepNumber } = state;
-  switch (action) {
+  switch (action.type) {
+    case TourActions.initialize:
+      return {
+        completedTour: action.payload.completed,
+        stepNumber: 0,
+        startTour: !action.payload.completed,
+      };
     case TourActions.start:
       return { startTour: true, completedTour: false, stepNumber: 0 };
     case TourActions.next:
@@ -39,7 +54,7 @@ export const tourReducer = (state: TourState, action: TourActions) => {
   }
 };
 
-type TourReducer = Reducer<TourState, TourActions>;
+type TourReducer = Reducer<TourState, TourStateAction>;
 
 type TourContextType = {
   tourState?: TourState;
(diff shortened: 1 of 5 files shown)


frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -39,7 +46,7 @@ export const tourReducer = (state: TourState, action: TourActions) => {
   }
 };
 
-type TourReducer = Reducer<TourState, TourActions>;
+type TourReducer = Reducer<TourState, { type: TourActions; payload?: boolean }>;
```
sahil143, line 49: Probably introduce a new for action say
```
type TourStateAction = { type: TourActions; payload?: { completed?: boolean }}
```
```suggestion
type TourReducer = Reducer<TourState, TourStateAction>;
```
would be better to make payload an object and with a `completed` prop. So it's more readable and we know what we are sending in the payload. In future we might need to send some other data inside payload.

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -16,9 +16,16 @@ import {
...
+  completedTour: completed,
+  stepNumber: 0,
+  startTour: !completed,
+});
+export const tourReducer = (state: TourState, action: { type: TourActions; payload?: boolean }) => {
   const { stepNumber } = state;
-  switch (action) {
+  switch (action.type) {
+    case TourActions.initialize:
+      return initializeState(action.payload);
```
sahil143, line 28: no need to declare a method here. just set the state directly here
```suggestion
      return {
  completedTour: completed,
  stepNumber: 0,
  startTour: !completed,
};
```

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -84,11 +91,16 @@ export const useTourValuesForContext = (): TourContextType => {
       setTourCompletionLocalStorageDataForPerspective(activePerspective, true);
     }
   };
+  const activePerspectiveRef = useRef(activePerspective);
```
sahil143, line 94:
```suggestion
  const activePerspectiveRef = useRef<string>(activePerspective);
```

sahil143 (changes requested):

frontend/packages/console-app/src/components/tour/tour-context.ts
nemesis09: done

frontend/packages/console-app/src/components/tour/tour-context.ts
nemesis09: done

frontend/packages/console-app/src/components/tour/tour-context.ts
nemesis09: done

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -84,11 +91,16 @@ export const useTourValuesForContext = (): TourContextType => {
...
+  const activePerspectiveRef = useRef<string>(activePerspective);
   const [tourState, tourDispatch] = useReducer<TourReducer>(tourReducer, {
     completedTour: completed,
     stepNumber: 0,
     startTour: !completed,
   });
+  if (activePerspective !== activePerspectiveRef.current) {
+    tourDispatch({ type: TourActions.initialize, payload: { completed } });
+    activePerspectiveRef.current = activePerspective;
+  }
```
christianvogt, line 103: You should execute side effects inside `useEffect`
```suggestion
React.useEffect(() => {
    tourDispatch({ type: TourActions.initialize, payload: { completed } });
    setPerspective(activePerspective);
  }, [activePerspective, completed]);
```

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -84,11 +91,16 @@ export const useTourValuesForContext = (): TourContextType => {
       setTourCompletionLocalStorageDataForPerspective(activePerspective, true);
     }
   };
+  const activePerspectiveRef = useRef<string>(activePerspective);
```
christianvogt, line 94:
```suggestion
  const perspective = React.useState(activePerspective);
```

frontend/packages/console-app/src/components/tour/tour-context.ts
nemesis09: used `useEffect` for initializing state on perspective change

frontend/packages/console-app/src/components/tour/tour-context.ts
nemesis09: done

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -78,17 +93,24 @@ export const useTourValuesForContext = (): TourContextType => {
...
   const [tourState, tourDispatch] = useReducer<TourReducer>(tourReducer, {
     completedTour: completed,
     stepNumber: 0,
     startTour: !completed,
   });
+
+  useEffect(() => {
+    tourDispatch({ type: TourActions.initialize, payload: { completed } });
+    setPerspective(activePerspective);
+  }, [activePerspective, completed]);
```
sahil143, line 112: I don't think we need to run the effect on completed. We want to only run the effect when the perspective change so that the tour state can be initialized.

```suggestion
  }, [activePerspective]);
```

nemesis09: updated dependency and disabled es lint rules for the line

sahil143 (approved):
/lgtm

christianvogt:
Going from admin -> dev perspective when the dev perspective tour has already been completed still shows the tour.
/lgtm cancel

nemesis09:
> Going from admin -> dev perspective when the dev perspective tour has already been completed still shows the tour.
> /lgtm cancel

@christianvogt  pushed changes addressing the issue. please take a look

PR #6716, opened 2020-09-23
Bug 1881820: Fix external mode queries for RGW Performance Card

bipuladh:
- Fix the queries for External Mode
- Show an empty graph for empty data frames. 
- Add dependency array for Infrastructure Detection Logic in Ceph
After Fix:
Screenshot from 2020-09-23 19-57-14

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
@@ -76,7 +76,7 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
           setIndepModeSupportedPlatform(true);
         }
       });
-  });
+  }, []);
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -76,7 +76,7 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
           setIndepModeSupportedPlatform(true);
         }
       });
-  });
+  }, []);
```
cloudbehl, line 79: Do we need useEffect if this is not going refreshed base on some parameters?

anmolsachan:
metrics look good to me

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
bipuladh: Discussed in DMs

cloudbehl:
/lgtm

nit: We might need to rethink calling `bandwidth` as `Throughput`.

PR #6719, opened 2020-09-23
Bug 1881881: Replace route dropdown input with typeahead select menu in import/edit flow

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4869

**Analysis / Root cause**: 
The dropdown only allows for the selection of the exposed ports from the builder image.

**Solution Description**: 
The user should have the ability to specify their target port to ensure the route works.
Relevant Slack thread: https://coreos.slack.com/archives/CHC2R6AGG/p1600757882002000

tmp
tmp

Unchanged.

**Test setup:**
- Navigate to +Add -> From git
- Fill form, select Routing in advanced options
- Notice the port input field provides typeahead for default ports, and also allows adding a custom port.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
+++ b/frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
@@ -1,26 +1,19 @@
 import * as React from 'react';
-import * as _ from 'lodash';
-import { useFormikContext, FormikValues } from 'formik';
+import { useFormikContext } from 'formik';
 import { TextInputTypes } from '@patternfly/react-core';
-import { InputField, DropdownField } from '@console/shared';
-import { makePortName } from '../../../utils/imagestream-utils';
+import { InputField } from '@console/shared';
+import { DeployImageFormData, GitImportFormData } from '../import-types';
+import PortInputField from './PortInputField';
 
 const CreateRoute: React.FC = () => {
   const {
     values: {
       image: { ports },
-      route: { defaultUnknownPort, targetPort },
+      route: { defaultUnknownPort },
     },
-  } = useFormikContext<FormikValues>();
-  const portOptions = ports.reduce((acc, port) => {
-    const name = makePortName(port);
-    acc[name] = (
-      <>
-        {port.containerPort} &rarr; {port.containerPort} ({port.protocol})
-      </>
-    );
-    return acc;
-  }, {});
+  } = useFormikContext<DeployImageFormData | GitImportFormData>();
+  const portOptions = ports.map((port) => port.containerPort.toString());
+  const placeholderPort = ports[0]?.containerPort || defaultUnknownPort;
 
   return (
     <>
@@ -37,24 +30,13 @@ const CreateRoute: React.FC = () => {
         placeholder="/"
         helpText="Path that the router watches to route traffic to the service."
       />
-      {_.isEmpty(ports) ? (
-        <InputField
-          type={TextInputTypes.text}
-          name="route.unknownTargetPort"
-          label="Target Port"
-          placeholder={defaultUnknownPort}
-          helpText="Target port for traffic."
-        />
-      ) : (
-        <DropdownField
-          name="route.targetPort"
-          label="Target Port"
-          items={portOptions}
-          title={portOptions[targetPort] || 'Select target port'}
-          helpText="Target port for traffic."
-          fullWidth
-        />
-      )}
+      <PortInputField
+        name="route.unknownTargetPort"
+        label="Target Port"
+        placeholderText={placeholderPort.toString()}
+        helpText="Target port for traffic."
+        options={portOptions}
+      />
     </>
   );
 };
--- a/frontend/packages/dev-console/src/components/import/route/PortInputField.tsx
+++ b/frontend/packages/dev-console/src/components/import/route/PortInputField.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+import { useField, useFormikContext, FormikValues } from 'formik';
+import { FormGroup, Select, SelectVariant, SelectOption } from '@patternfly/react-core';
+import { useFormikValidationFix, getFieldId } from '@console/shared';
+
+interface RouteInputFieldProps {
+  name: string;
+  label: string;
+  options: string[];
+  placeholderText: string;
+  helpText: string;
+}
+
+const PortInputField: React.FC<RouteInputFieldProps> = ({
+  name,
+  label,
+  options,
+  placeholderText,
+  helpText,
+}) => {
+  const [field, { touched, error }] = useField<string>(name);
+  const { setFieldValue, setFieldTouched } = useFormikContext<FormikValues>();
+  const [isOpen, setIsOpen] = React.useState<boolean>(false);
+  const fieldId = getFieldId(name, 'select-input');
+  const isValid = !(touched && error);
+  const errorMessage = !isValid ? error : '';
+
+  useFormikValidationFix(field.value);
+
+  const onToggle = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const onSelect = (_event, selection: string) => {
+    setFieldValue(name, selection);
+    setFieldTouched(name);
+    onToggle();
+  };
+
+  const onClearSelection = () => {
+    setFieldValue(name, '');
+    setFieldTouched(name);
+  };
+
+  return (
+    <FormGroup
+      fieldId={fieldId}
+      validated={isValid ? 'default' : 'error'}
+      label={label}
+      helperText={helpText}
+      helperTextInvalid={errorMessage}
+    >
+      <Select
+        variant={SelectVariant.typeahead}
+        onToggle={onToggle}
+        onSelect={onSelect}
+        onClear={onClearSelection}
+        isOpen={isOpen}
+        selections={field.value}
+        placeholderText={placeholderText}
+        isCreatable
+      >
+        {options.map((val) => (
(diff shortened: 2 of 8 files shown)


frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
```diff
@@ -1,26 +1,17 @@
...
-  const portOptions = ports.reduce((acc, port) => {
-    const name = makePortName(port);
-    acc[name] = (
-      <>
-        {port.containerPort} &rarr; {port.containerPort} ({port.protocol})
-      </>
-    );
-    return acc;
-  }, {});
+  const portOptions = ports.map((port) => port.containerPort);
```
jerolimov, line 14: When you convert the ports here to strings the `Select` automatically matches the user input and do not show "Create "8080"" if the option is already available. Wdyt?

```suggestion
  const portOptions: string[] = ports.map((port) => port.containerPort.toString());
```

image

frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
```diff
@@ -1,26 +1,17 @@
...
+import RouteInputField from './RouteInputField';
 
 const CreateRoute: React.FC = () => {
   const {
     values: {
       image: { ports },
-      route: { defaultUnknownPort, targetPort },
+      route: { defaultUnknownPort },
     },
   } = useFormikContext<FormikValues>();
```
jerolimov, line 13:
```suggestion
  } = useFormikContext<DeployImageFormData | GitImportFormData>();
```

Adding this types so that the ports and defaultUnknownPort vars are typed.

jerolimov:
@rottencandy Technical this works fine, I have added two comments to the code. I also want ask you if you want rename your new component from `import/route/RouteInputField` to `import/route/PortInputField`. For me this matches more the special case it handles here.

In this case a port validation 1-65535 would be also a good idea. It's not required, but I missed this when entering some invalid numbers.

Wdyt?

frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
rottencandy: Updated. Not sure if there is something we can do about the "Create ..." option, even the Patternfly demos behave the same way: https://www.patternfly.org/v4/documentation/react/components/select#typeahead

frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
rottencandy: Updated, didn't know we could do that. Now I'm wondering why we even use `FormikValues` type :P.

rottencandy:
> @rottencandy Technical this works fine, I have added two comments to the code. I also want ask you if you want rename your new component from `import/route/RouteInputField` to `import/route/PortInputField`. For me this matches more the special case it handles here.

Thanks for the review @jerolimov , Renamed to `PortInputField`. 
> In this case a port validation 1-65535 would be also a good idea. It's not required, but I missed this when entering some invalid numbers.

Currently we do the validation using a regex(`/^\d+$/`), I wasn't able to find a way to check for a maximum number using regex.

frontend/packages/dev-console/src/components/import/route/CreateRoute.tsx
jerolimov: Okay, thanks for revalidating this. I though that the .toString() solves this issue.

jerolimov:
> Currently we do the validation using a regex(`/^\d+$/`)

We could use a number check similar to this instead of regex for the `unknownTargetPort` validation scheme:

```tsx
const portValidationSchema = yup
  .number()
  .typeError('Port must be an Integer.')
  .integer('Port must be an Integer.')
  .min(1, 'Port must be between 1 and 65535.')
  .max(65535, 'Port must be between 1 and 65535.');
```

Didn't know the typeError check before. It was the message if you enter "asd". Integer ensures that user could not enter "3.1415". Wdyt? Is it worth to update the PR again? :)

rottencandy:
> We could use a number check similar to this instead of regex for the `unknownTargetPort` validation scheme:
> 
> ```tsx
> const portValidationSchema = yup
>   .number()
>   .typeError('Port must be an Integer.')
>   .integer('Port must be an Integer.')
>   .min(1, 'Port must be between 1 and 65535.')
>   .max(65535, 'Port must be between 1 and 65535.');
> ```
> 
> Didn't know the typeError check before. It was the message if you enter "asd". Integer ensures that user could not enter "3.1415". Wdyt? Is it worth to update the PR again? :)

Updated.

rottencandy:
Updated to behave the same way when selecting Knative Service.

jerolimov:
Hmm, I'm really sorry but could not verify this issue. I tried to set a custom port for the route and select "Knative Service" as Deployment. The created Route has not set the port anywhere.

Then I tried this on master, and changed `ServerlessRouteSection.tsx` so that the `<InputField name="route.unknownTargetPort" ... ` was always used. And the Route does not contain the entered port in this case as well. Only when selecting a port from the Dropdown (on master) the value was set in the Route correctly. So maybe this is already an issue before? @invincibleJai  what do you think?

@rottencandy Unfortunately, I'm now thinking about the unifying between two different cases. Before we have `<InputField name="route.unknownTargetPort" ... ` OR `<DropdownField name="route.targetPort" ...`. But with this PR we always use the `unknownTargetPort`. If we have no other reason to keep `targetPort` which I maybe oversee at the moment, we should maybe simplify the complete `createRoute` algorithm and should always use only one variable / form field here? :thinking:

rottencandy:
@jerolimov The knative service function prioritized `targetPort` over `unknownTargetPort`, which is why it was never using the entered port, I've updated it to use `unknownTargetPort` when available.

AFAIK, `targetPort` is mainly being used in two places:
- In the edit flow, when user does not intend to change port, `targetPort` will be set as the existing port value(in full format eg.`8080-tcp`).
- In the image stream search fields, both internal & external fields set `targetPort` to the first of available ports, I'm still looking into why it is being set here.

invincibleJai:
Verified scenarios with Deployment / DeploymentConfig / KSVC and edit flows , works as expected 

/cc @christianvogt

jerolimov:
Discussed with Saud, thanks for that, in case of Knative Service the port was set in the KSVC correctly. Not in the Route.

If @invincibleJai and @christianvogt are happy with that, I'm happy as well. :+1: :)

serenamarie125 (approved):
This looks great, thanks @rottencandy !

bgliwa01 (approved):
lgtm 
cc @beaumorley @parvathyvr

christianvogt:
There are 2 issues here I found while testing.

1. This may be a patternfly issue, but when the select menu is open and I press `Enter` to confirm my selection, the form submits. I actually tested the application selector and it exhibits the same behavior. When selecting an option while the dropdown is opened, the enter key should prevent default. Not as important to fix this since we already the same issue present.

2. While the route gets updated to `3000-tcp` the service does not have a matching port. Instead it gets created with this:
```
  ports:
    - name: 8080-tcp
      protocol: TCP
      port: 8080
      targetPort: 8080
```

This should have also been updated to use port 3000:
```
  ports:
    - name: 3000-tcp
      protocol: TCP
      port: 3000
      targetPort: 3000
```

Once the build completes, click the route on topology and it should open up the web page. If it doesn't, there's still an issue.
Using `https://github.com/nodeshift-starters/react-web-app` as a test repo, port 3000 should work.

cc @invincibleJai

invincibleJai:
@rottencandy one issue am seeing after the fix with servcie in deployimge flow, i think before appending the exposed port for service we need to check if that exists

image

rottencandy:
Added duplicate check before appending port.

invincibleJai:
Thanks @rottencandy works fine now 

cc @christianvogt

christianvogt:
One more small issue:
- import from git `https://github.com/nodeshift-starters/react-web-ap` and set port to 3000
- from topology right click the deployment and select `edit <mydeployment>`
- open the routing advanced section
- click the `x` in the target port text field
- observe the placeholder shows `8080`
- complete the form
- if you navigate to the route associated with this deployment, the port remains as `3000-tcp`

rottencandy:
> One more small issue:
> 
>     * import from git `https://github.com/nodeshift-starters/react-web-ap` and set port to 3000
> 
>     * from topology right click the deployment and select `edit <mydeployment>`
> 
>     * open the routing advanced section
> 
>     * click the `x` in the target port text field
> 
>     * observe the placeholder shows `8080`
> 
>     * complete the form
> 
>     * if you navigate to the route associated with this deployment, the port remains as `3000-tcp`

Fixed.

debsmita1:
> > One more small issue:
> > ```
> > * import from git `https://github.com/nodeshift-starters/react-web-ap` and set port to 3000
> > 
> > * from topology right click the deployment and select `edit <mydeployment>`
> > 
> > * open the routing advanced section
> > 
> > * click the `x` in the target port text field
> > 
> > * observe the placeholder shows `8080`
> > 
> > * complete the form
> > 
> > * if you navigate to the route associated with this deployment, the port remains as `3000-tcp`
> > ```
> 
> Fixed.

@rottencandy I could reproduce the above issue for knative service workload. Have a look at the gif below:
port

rottencandy:
Updated edit flow to use `defaultUnknownPort` when port is input is cleared.

debsmita1:
verified this locally
/lgtm

PR #6725, opened 2020-09-23
Bug 1881386: hide Import with Wizard button when missing permissions

suomiy:


frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
```diff
@@ -248,6 +269,27 @@ const VirtualMachinesPage: React.FC<VirtualMachinesPageProps> = (props) => {
...
+      checkAccess({
+        group: ConfigMapModel.apiGroup,
+        verb: 'get',
+        resource: ConfigMapModel.plural,
+        name: VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME,
+        namespace: configMapNamespace,
+      })
+        .then((result: SelfSubjectAccessReviewKind) => {
+          if (result?.status?.allowed) {
+            setImportAllowed(result?.status?.allowed);
```
yaacov, line 285: can you replace `result?.status?.allowed` with `true` ?
we don't really using the `result` here, for example we can't set it to `false` in case the result is false

frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
```diff
@@ -248,6 +269,27 @@ const VirtualMachinesPage: React.FC<VirtualMachinesPageProps> = (props) => {
...
+          }
+        })
+        .catch(() => setImportAllowed(true));
```
yaacov, line 288: why do we return `true` on fail, what am I missing ?

frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
suomiy: it would be keeping the origin of that value. Nevertheless it doesn't really matter - changed to true

frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
suomiy: this is the default behaviour of rbac in console. If you fail checking permisions for something, then do it and fail/succeed a bit later

yaacov: can you point to an example ?

I found 2 places where we default to `false` and did not not find places where we default to `true`
[1] https://github.com/openshift/console/blob/e7565ea11aa80c80f8e0b34259f3be8fe56f8d57/frontend/packages/dev-console/src/components/import/image-search/ImageStreamNsDropdown.tsx#L38
[2] https://github.com/openshift/console/blob/aa4d040806db418f5348a998d5139fedaf27ae6e/frontend/packages/dev-console/src/utils/application-utils.ts#L351

suomiy: https://github.com/openshift/console/blob/master/frontend/public/components/utils/rbac.tsx#L119

yaacov: a - if you decide we should use this use case, you should also add the comment explaining the odd behaviour:
https://github.com/openshift/console/blob/master/frontend/public/components/utils/rbac.tsx#L114
b - this is not the default behaviour, `useAccessReview2` is used far less then `useAccessReview` that defaults to `false`
https://github.com/openshift/console/blob/master/frontend/public/components/utils/rbac.tsx#L127

yaacov:
/hold 

we need a comment to explain why we default to allow access when access check is failing (or default to dissallow)

frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
suomiy: `useAccessReview` uses `useAccessReview2` under the hood, so it has the same behaviour and defaults to `true`

ok, copied the explanation there

PR #6726, opened 2020-09-23
Bug 1868551: Fixed filtered node list view

gnehapk:
- Fixed filtered node list view and its alignment
- Fixed display of units for disk size
- Fixed fetching of pvs on create storage cluster step

Screenshot from 2020-09-24 18-22-40

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
@@ -71,7 +71,20 @@
 
 .ceph-ocs-install__filtered-modal {
   max-height: 75%;
-  width: 50%;
+  width: 55%;
+
+  .ceph-ocs-install__node-list {
+    max-width: 100%;
+  }
+
+  .co-m-nav-title {
+    margin-top: 0;
+  }
+
+  .co-m-pane__body {
+    padding-left: 0;
+    padding-right: 0;
+  }
 }
 
 // inner class of modal for modifying margin
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
@@ -137,8 +137,12 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
   React.useEffect(() => {
     if ((pvLoadError || pvData.length === 0) && pvLoaded) {
       setFilteredNodes([]);
+    } else if (pvLoaded) {
+      const pvs = getSCAvailablePVs(pvData, getName(storageClass));
+      const scNodes = getAssociatedNodes(pvs);
+      setFilteredNodes(scNodes);
     }
-  }, [pvData, pvLoaded, pvLoadError]);
+  }, [pvData, pvLoaded, pvLoadError, storageClass]);
 
   const submit = (event: React.MouseEvent<HTMLButtonElement>) => {
     event.preventDefault();
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
@@ -9,7 +9,7 @@ import {
 } from '@console/shared';
 import { humanizeCpuCores, ResourceLink } from '@console/internal/components/utils/';
 import { Table } from '@console/internal/components/factory';
-import { createMapForHostNames } from '@console/local-storage-operator-plugin/src/utils';
+import { NodeKind } from '@console/internal/module/k8s';
 import { getConvertedUnits } from '../../../utils/install';
 import { getColumns } from '../node-list';
 import { GetRows, NodeTableProps } from '../types';
@@ -18,10 +18,15 @@ import '../ocs-install.scss';
 const getRows: GetRows = ({ componentProps, customData }) => {
   const { data } = componentProps;
   const { nodes, setNodes, filteredNodes } = customData;
+  let filteredData: NodeKind[] = data;
 
-  const nodeList = filteredNodes ?? data.map(getName);
-  const hostNames = createMapForHostNames(data);
-  const filteredData = data.filter((node) => nodeList.includes(hostNames[getName(node)]));
+  if (filteredNodes) {
+    filteredData = data.filter(
+      (node: NodeKind) =>
+        filteredNodes.includes(getName(node)) ||
+        filteredNodes.includes(node.metadata.labels?.['kubernetes.io/hostname']),
+    );
+  }
 
   const rows = filteredData.map((node) => {
     const roles = getNodeRoles(node).sort();
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/donut-chart.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/donut-chart.tsx
@@ -9,7 +9,6 @@ import { ListPage } from '@console/internal/components/factory';
 import { getNodes } from '@console/local-storage-operator-plugin/src/utils';
 import { DiskListModal } from './disk-list';
 import { State, Action, Discoveries } from '../state';
-import { dropdownUnits } from '../../../../../constants';
 import { getTotalDeviceCapacity } from '../../../../../utils/install';
 import AttachedDevicesNodeTable from '../../sc-node-list';
 import '../../attached-devices.scss';
@@ -29,10 +28,10 @@ export const DiscoveryDonutChart: React.FC<DiscoveryDonutChartProps> = ({ state,
   React.useEffect(() => {
     const filterDisks = () => {
       const minSize = state.minDiskSize
-        ? Number(convertToBaseValue(`${state.minDiskSize} ${dropdownUnits[state.diskSizeUnit]}`))
+        ? Number(convertToBaseValue(`${state.minDiskSize} ${state.diskSizeUnit}`))
         : 0;
       const maxSize = state.maxDiskSize
-        ? Number(convertToBaseValue(`${state.maxDiskSize} ${dropdownUnits[state.diskSizeUnit]}`))
+        ? Number(convertToBaseValue(`${state.maxDiskSize} ${state.diskSizeUnit}`))
         : '';
       const filteredDiscoveries: Discoveries[] = state.nodesDiscoveries.filter((disk) => {
         if (nodes.includes(disk.node)) {
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
@@ -184,7 +184,7 @@ export const LocalVolumeSetInner: React.FC<LocalVolumeSetInnerProps> = ({
             <Dropdown
               id="create-lvs-disk-size-unit-dropdown"
               items={diskSizeUnitOptions}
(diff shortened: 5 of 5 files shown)


cloudbehl:
few  UX questions:
1. The nodes list looks extra padded to be from left and right
2. When we have [X] on top why we need a close button at the bottom left doesn't make sense.
3. Can we increase the width of the model so that in XL screen zone can come in one line?

gnehapk:
> few UX questions:
> 
> 1. The nodes list looks extra padded to be from left and right
> 2. When we have [X] on top why we need a close button at the bottom left doesn't make sense.
This is same behavior with all other modals we have in console
> 3. Can we increase the width of the model so that in XL screen zone can come in one line?

cloudbehl:
> This is same behavior with all other modals we have in console
RIght but this is an informational modal that doesn't have any option to do next or create. We can check with UX & work on this in the next release.

gnehapk:
> > This is same behavior with all other modals we have in console
> 
> RIght but this is an informational modal that doesn't have any option to do next or create. We can check with UX & work on this in the next release.
> 
> @shirimordechay @yuvalgalanti thoughts?

Even as per UXD, we have the cancel button - https://marvelapp.com/prototype/ai19e9g/screen/69554638

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
```diff
@@ -17,11 +18,16 @@ import '../ocs-install.scss';
...
 
   const nodeList = filteredNodes ?? data.map(getName);
-  const hostNames = createMapForHostNames(data);
-  const filteredData = data.filter((node) => nodeList.includes(hostNames[getName(node)]));
+  if (isHostNameSame) {
+    filteredData = data.filter((node) => nodeList.includes(getName(node)));
+  } else {
+    const hostNames = createMapForHostNames(data);
+    filteredData = data.filter((node) => nodeList.includes(hostNames[getName(node)]));
+  }
```
cloudbehl, line 30: When the hostname & node name is the same then the`hostNames` map can have the same value in key value pair? Why do we need if condition looking at line 23 & 24 it should have worked. 

Am I missing something?

shirimordechay:
Looks good to me, we don't have any modals without a button in the console. Even if it's just informative the user should have another exit point except for the X icon.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
```diff
@@ -71,7 +71,21 @@
...
+    max-width: 100%;
+  }
+
+  .co-m-nav-title {
+    margin-top: 0;
+  }
+
+  .co-m-pane__body {
+    padding-left: 0;
+    padding-right: 0;;
```
cloudbehl, line 87: Nit: I think `co-m-nav-title` & `co-m-pane__body` should be inside the `ceph-ocs-install__filtered-modal` & we should not override the default class. 
```suggestion
    padding-right: 0;
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
gnehapk: Else part will be executed only when the hostname and node names are not same. Check at line 25

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
gnehapk: I agree, but there is no other way to over write the styles. Here I have over-written the styles for just under `.ceph-ocs-install__node-list `, so it will not impact other views i.e. a > b

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
cloudbehl: Right we can create the map for both cases(if they are diff or not) and remove the if the condition itself. Rather than passing a https://github.com/openshift/console/pull/6726/files#diff-99347a73d43d6e64cba499ad4c8093edR147 hardcoded true value.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
cloudbehl: ack!

cloudbehl:
> Looks good to me, we don't have any modals without a button in the console. Even if it's just informative the user should have another exit point except for the X icon.

I don't see any informational modal as well right now in UI. One example of an informational modal is about a page that doesn't have close as a separate button.

shirimordechay:
The about modal has a very different design from any other designs in the console, I don't think it's comparable. I don't see any reason to remove the button, an additional escape hatch is basic in overlays designs.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
gnehapk: For i3, `filteredNodes` from `create storage cluster step` will have `hostNames` as this is fetched from pv's hostname label. Hence `filteredData` will be empty as this condition will not work - `nodeList.includes(getName(node))` as its comparing with node name. That's why the else block is required. Hope that clarifies.
Else you are welcome to debug if you want to understand further.

afreen23: Why dont you just maintain hostname, discovery requires hostname only

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
```diff
@@ -71,7 +71,21 @@
 
 .ceph-ocs-install__filtered-modal {
   max-height: 75%;
-  width: 50%;
+  width: 55%;
+
+  // for this page, 100% width is required
```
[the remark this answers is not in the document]
gnehapk: This means here `the filtered node list page here`. For all other node lists,  as part of alignment changes, the width is set to 70%. And just for this page, 100% is required. Hence this comment.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
gnehapk: Just maintaining hostname will not work for all other lists and will increase the complexity. Hence, I prefer not to.
Anyways this comment seems like more of refactoring change, which I prefer not to do as a part of this PR.

cloudbehl (approved):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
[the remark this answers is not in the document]
afreen23: I thought it is  for `ceph-ocs-install__node-list` . Because I see 100% width set for that. Please place comments close to the object in reference.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
```diff
@@ -137,8 +137,12 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
   React.useEffect(() => {
     if ((pvLoadError || pvData.length === 0) && pvLoaded) {
       setFilteredNodes([]);
```
afreen23, line 139: One thing to point out is that we are not handling anywhere the fallback UX when we have no `filterednodes`.
It will show up the list page with table headers and no rows.
I am not sure why this is so, because if we get empty data in list page, we see "No Nodes Found" like message.
Screenshot from 2020-09-28 14-12-22

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
afreen23, line 30: Wont any of these work ?
```suggestion
  const filteredData = data.filter(
    (node: NodeKind) =>
      nodeList.includes(getName(node)) ||
      nodeList.includes(node.metadata.labels?.['kubernetes.io/hostname']),
  );
```
OR

Pass the maintained state from pre-step about the "filteredNodes"

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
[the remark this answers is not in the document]
afreen23: Making comments specific will help you in future in understanding, if the name is sufficient then dont put comment.
Thats the rule :)

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
gnehapk: Ack. This change needs to be done at multiple places and hence, I will take this change as a part of separate PR. 
This PR is urgent and needs to be get in asap.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
gnehapk: We already have the function - `createMapForHostNames`, which checks for the hostname and hence no need to do the similar check here.
Also this component is reused at multiple places, hence it make sense to do the check here, instead of calculating the same from all different components.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/attached-devices.scss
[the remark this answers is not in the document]
gnehapk: I will remove the comment as I think its self explanatory.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
afreen23: I am okay If you want to pass prop but I think in any scenario, you dont need to `createMapForHostNames`, you can directly do,
```
nodeList.includes(node.metadata.labels?.['kubernetes.io/hostname'])
```
 Its just extra computation.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/sc-node-list.tsx
```diff
@@ -17,11 +18,16 @@ import '../ocs-install.scss';
 
 const getRows: GetRows = ({ componentProps, customData }) => {
   const { data } = componentProps;
-  const { nodes, setNodes, filteredNodes } = customData;
+  const { nodes, setNodes, filteredNodes, isHostNameSame } = customData;
+  let filteredData: NodeKind[] = [];
 
   const nodeList = filteredNodes ?? data.map(getName);
```
afreen23, line 24: Instead of setting nodeList -> data.map(getName) in case of no `filteredNodes` , You want to directly set `filteredData` to the `data.map(getName)` and avoid redundant filtering below.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
afreen23: I am okay to see it in another PR as well.

afreen23 (changes requested):

afreen23 (approved):
/lgtm

PR #6727, opened 2020-09-23
Bug 1882026: update notifications message on Topology and removes from Add

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4907

**Analysis / Root cause**: 
- In Topology view Notification is shown to user "No workloads found" if there are no resources exists but currently topology shows more then regular workloads i.e channels/broker/sources (some may not have underlying deployments)
- In Add view Notification is shown to user "No workloads found" if there are no resources exists , for this we fetch lot of resources to figure that out which may not be needed as if user directly comes to Add, they would be interested to Add a resource

**Solution Description**: 
- if user goes to Topology view then show notification and add option as we show now and update message to "No resources found" from "No workloads found" 
- Don't show notification in Add view

@openshift/team-devconsole-ux 
- Topology view with no resources
image

- Add view with no resources
image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/AddPage.tsx
+++ b/frontend/packages/dev-console/src/components/AddPage.tsx
@@ -1,22 +1,7 @@
 import * as React from 'react';
-import * as _ from 'lodash';
 import { Helmet } from 'react-helmet';
 import { match as RMatch } from 'react-router';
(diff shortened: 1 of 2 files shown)


bgliwa01 (approved):
lgtm! Thanks Jai

PR #6731, opened 2020-09-23
Bug 1882097: Fix layout issue which appears only on Safari

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4910
https://bugzilla.redhat.com/show_bug.cgi?id=1882097

**Analysis / Root cause**: 
There are `position:` issues which only appears in Safari when viewing and editing a pipeline.

**Solution Description**: 
Remove and add a position css rule so that the pipeline visualization looks similar in Safari and the others browsers. Tested this in latest Chrome, Firefox and Safari.

Chrome 85, Firefox 80 and Safari 14 on macOS:
odc-4910 mov

Unchanged

**Test setup:**
* Test a Pipeline visualization in read only mode (open an existing pipeline)
* Create a new pipeline and check if the Select task was also shown at the right position

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.scss
+++ b/frontend/packages/dev-console/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.scss
@@ -2,7 +2,6 @@
 
 $border-color: var(--pf-global--BorderColor--light-100);
 .odc-pipeline-vis-task {
-  position: relative;
   width: 10em;
   cursor: default;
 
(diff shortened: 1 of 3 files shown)


christianvogt:
Small side effect of using `position: fixed` is that the popover now overlaps the entire button because the node no longer has a height.
image

jerolimov:
> Small side effect of using `position: fixed` is that the popover now overlaps the entire button because the node no longer has a height.

Thanks for finding this. I need to add an additional height which we already use for the svg part. Updated the video above again.

I noticed that there are some more small issues when using the browser zoom, esp. on Safari.

andrewballantyne (approved):
Tested it on Safari, FF, Chrome, and Edge (Chromium) and it works as expected.

PR #6732, opened 2020-09-23
Bug 1882142: Automatically select correct PipelineResource when starting a PipelineRun (or create a trigger)

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3327
https://bugzilla.redhat.com/show_bug.cgi?id=1882142

**Analysis / Root cause**: 
For pipeline templates with input and output resources the console automatically creates PipelineResources (for known resource names). This PipelineResources has no link to the Pipeline instance.

When starting a PipelineRun the user needs to select the correct PipelineResources which is hard when there is more than one PipelineResource available.

**Solution Description**:
This change adds a new label to the auto-generated pipeline resources. `tekton.dev/pipeline: $pipelineName`.

When starting a pipeline (create a PipelineRun) the modal automatically select resources based on the pipeline name and this resource label.

odc-3327 webm

Unchanged

**Test setup:**
* Requires OpenShift Pipelines Operator 1.0.1 (or earlier)
* Import a Deployment from git and enable the "Add Pipeline" option
* Check the created PipelineResources
* Open "Start Pipeline" or "Add Trigger" modal and check if the correct resource is preselected

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/PipelineResourceDropdown.tsx
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/PipelineResourceDropdown.tsx
@@ -16,6 +16,7 @@ import './PipelineResourceDropdown.scss';
 
 type PipelineResourceDropdownProps = {
   autoSelect?: boolean;
+  pipelineName?: string;
   filterType: string;
   name: string;
   namespace: string;
@@ -23,7 +24,7 @@ type PipelineResourceDropdownProps = {
 };
 
 const PipelineResourceDropdown: React.FC<PipelineResourceDropdownProps> = (props) => {
-  const { autoSelect, filterType, name, namespace, selectedKey } = props;
+  const { autoSelect, pipelineName, filterType, name, namespace, selectedKey } = props;
 
   const [isExpanded, setExpanded] = React.useState(false);
   const { setFieldValue, setFieldTouched } = useFormikContext<FormikValues>();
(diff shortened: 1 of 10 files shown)


frontend/packages/dev-console/src/components/pipelines/modals/common/PipelineResourceDropdown.tsx
```diff
@@ -41,21 +42,27 @@ const PipelineResourceDropdown: React.FC<PipelineResourceDropdownProps> = (props
...
     if (canAutoSelect) {
-      if (availableResources.length === 0) {
-        setFieldValue(name, CREATE_PIPELINE_RESOURCE);
+      const matchingResources = availableResources.filter(
+        (resource) => resource.metadata?.labels?.['tekton.dev/pipeline'] === pipelineName,
+      );
+      if (matchingResources.length > 0) {
+        setFieldValue(name, matchingResources[0].metadata.name, false);
+      } else if (availableResources.length > 0) {
+        setFieldValue(name, availableResources[0].metadata.name, false);
```
jerolimov, line 59: @andrewballantyne I keep the "select the first resource" fallback here. If we drop this here the user needs to select a resource if there is no matching resource found. I would prefer this over selecting "a random one". Because of this question and that we are close to the release I marked this PR as WIP.

frontend/packages/dev-console/src/components/pipelines/modals/common/PipelineResourceDropdown.tsx
```diff
@@ -41,21 +42,27 @@ const PipelineResourceDropdown: React.FC<PipelineResourceDropdownProps> = (props
...
       } else {
-        setFieldValue(name, availableResources[0].metadata.name);
+        setFieldValue(name, CREATE_PIPELINE_RESOURCE, false);
```
jerolimov, line 61: If we drop this else-case as well the user will show a "Select Resource" instead of seeing the option "Create resource" with an additional input field. So my recommendation is to set the value only if we found a matching resource.

andrewballantyne:
Given the shift to Workspaces, not sure this will ever get use. Perhaps we just close this off and focus on Workspace support.

jerolimov:
> Given the shift to Workspaces, not sure this will ever get use. Perhaps we just close this off and focus on Workspace support.

Yeah, focusing on Workspaces should be the goal. But I expect we keep this code at least for 4.7 so I prefer that we fix this in 4.6 and backport this to 4.5 for all users which will now or in the near future update to 4.5.

andrewballantyne:
/close

I don't think we will be doing this _ever_ given the shift in priorities upstream. See the Jira for more context on my reasoning. Closing this to prevent noise for devs during the last days of Code Freeze when we are trying to close PRs.

PR #6733, opened 2020-09-23
Bug 1882152: Make "Add to Navigation" accessible for tabs

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3796
https://bugzilla.redhat.com/show_bug.cgi?id=1882152

**Analysis / Root cause**: 
A link without href is not "tab-able"

**Solution Description**: 
Change to button and remove the css which remove that the outline was not shown.

@openshift/team-devconsole-ux button shows now an outline:
odc-3796 webm

Not changed

**Test setup:**
* Open the search in developer perspective

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/public/components/search.tsx
+++ b/frontend/public/components/search.tsx
@@ -7,6 +7,8 @@ import {
   AccordionContent,
   AccordionItem,
   AccordionToggle,
+  Button,
+  ButtonVariant,
   Toolbar,
   ToolbarChip,
   ToolbarContent,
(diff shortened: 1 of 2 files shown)


frontend/public/components/search.tsx
```diff
@@ -269,7 +269,8 @@ const SearchPage_: React.FC<SearchProps & StateProps & DispatchProps> = (props)
                 >
                   {getToggleText(resource)}
                   {props.perspective !== 'admin' && (
-                    <a
+                    <button
```
sahil143, line 272: Can we use the Patternfly button here? Then we wouldn't have to use the `pf-c-button`, `pf-m-link` classes here.

jerolimov: Good idea, changed to PatternFly Button

sahil143 (approved):
/lgtm

PR #6736, opened 2020-09-24
Bug 1881125: Disable 'Snapshots' tab on VM details page

irosenzw:


suomiy:
I see that objective of the bug has changed

/lgtm

irosenzw:
> I see that objective of the bug has changed
We'll consider the issue the bug was originally opened for when we restore the `Snapshots` tab.

PR #6739, opened 2020-09-24
Bug 1853352: update CDI version to v1beta1

suomiy:


yaacov:
/hold

kubevirt-plugin test is failing, holding until we know what is the reason

@irosenzw @omril-redhat is the test failing because of a problem in the code ?

cc/ @metalice @rawagner

irosenzw:
@yaacov it looks like a problem in the test infra, I'll dig into this.

yaacov:
/close
will continue in #7640

PR #6743, opened 2020-09-24
Bug 1874890: Include accessibility text for screenreaders only, applied to dashboard status sections when displaying loading status skeleton

sg00dwin:
@spadgett Is this an acceptable approach?

Bug https://bugzilla.redhat.com/show_bug.cgi?id=1874890

Screen Shot 2020-09-24 at 2 27 00 PM

### DIFF
--- a/frontend/packages/console-shared/src/components/dashboard/status-card/HealthItem.tsx
+++ b/frontend/packages/console-shared/src/components/dashboard/status-card/HealthItem.tsx
@@ -16,7 +16,9 @@ const HealthItem: React.FC<HealthItemProps> = React.memo(
     return (
       <div className={classNames('co-status-card__health-item', className)}>
         {state === HealthState.LOADING ? (
-          <div className="skeleton-health" />
+          <div className="skeleton-health">
+            <span className="pf-u-screen-reader">Loading {title} Status</span>
+          </div>
         ) : (
           !noIcon && <HealthItemIcon state={state} />
         )}


frontend/packages/console-shared/src/components/dashboard/status-card/HealthItem.tsx
```diff
@@ -16,7 +16,7 @@ const HealthItem: React.FC<HealthItemProps> = React.memo(
     return (
       <div className={classNames('co-status-card__health-item', className)}>
         {state === HealthState.LOADING ? (
-          <div className="skeleton-health" />
+          <div className="skeleton-health" aria-label={`Loading ${title} Status`} />
```
spadgett, line 19: I'm not sure about using `aria-label` on a `div`. @jessiehuff Do you know if this works?

sg00dwin: Based on https://www.w3.org/WAI/GL/wiki/Using_aria-label_to_provide_labels_for_objects it looks that it can, though it should also include a `role`. I updated the pr to include `role="status` since it is a identifying a transient state - more info https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22.html

`<div className="skeleton-health" aria-label={`Loading ${title} Status`} role="status" />`

jessiehuff: I don't really recommend using an `aria-label` on a `div`. Leonie Watson, who is a prominent thought leader in the accessibility community, wrote up this [note](https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/) on labeling usage with different elements, and using aria labels on a div are not likely to work across all browser/assistive technology combinations. What we could do instead is use the [screen reader utility](https://www.patternfly.org/v4/documentation/core/utilities/accessibility#screen-reader-only) to add text specifically for screen reader users.

sg00dwin: I have updated the pr to use `.pf-u-screen-reader` with accessibility text associated.

spadgett (approved):
/lgtm

PR #6639, opened 2020-09-15
Bug 1868653: Convert Pipeline StartedBy Label to Annotation

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4862

**Analysis / Root cause**: 
StartedBy was added in 4.5, it was added as a label. Labels have limitations, it was unclear the full usage of the name and it was incorrectly assigned to a label, whereas we need to support a full range of characters beyond what labels can have.

**Solution Description**: 
Annotations do not have this limitation, moving the started by content from labels to annnotations.

**Screen shots / Gifs for design review**: 

kube:admin
image

Email username:
Screen Shot 2020-09-15 at 7 06 47 PM

Trigger:
image

**Unit test coverage report**: 
No change

**Test setup:**
* Setup a Pipeline
* Start said Pipeline, notice the started by information that is your user
    * Try `kube:admin`
    * Try creating a user with an email address
* Setup a trigger and invoke it to see it still shows the trigger details

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #6744, opened 2020-09-24
Bug 1882529: Removes the kubectl annotation from the created PLRs

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4917

**Analysis / Root cause**: 
Blindly copying all the annotations over seems to cause an issue with the TriggerTemplate creation if there is an issue in the `kubectl.kubernetes.io/last-applied-configuration` annotation

**[Temp] Solution Description**: 
Remove the annotation as it's a hold over from the previous configuration.

There is a growing need to refactor this method and subsequently the callers to properly define a clean structure... however this shouldn't be done in 4.6 and definitely not this close to Code Freeze.

No visual change.

**Test setup:**
See the ticket.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/pipelines/modals/common/utils.ts
+++ b/frontend/packages/dev-console/src/components/pipelines/modals/common/utils.ts
@@ -58,6 +58,14 @@ export const getPipelineRunData = (
   const latestRunParams = latestRun?.spec.params;
   const pipelineParams = pipeline?.spec.params;
   const params = latestRunParams || getPipelineRunParams(pipelineParams);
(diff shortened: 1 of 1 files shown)


andrewballantyne:
Cherry-pick to 4.5 should be done manually with https://github.com/openshift/console/pull/6639

jerolimov (approved):
Tested it locally, works fine with the given resources.yaml.

/lgtm

PR #6745, opened 2020-09-25
Bug 1860985: Updated minimum requirement for Internal And Internal-Attached Devices

gnehapk:
- Updated minimum requirement text for internal and attached devices mode
- Updated text and conditions for Minimal cluster deployment Alert for internal and attached devices mode. For both, it should be same.

[Ignore the error message as this is just for making the changes visible)
Screenshot from 2020-09-25 17-08-22

Screenshot from 2020-09-25 17-08-10

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
@@ -51,7 +51,7 @@ import {
 import {
   filterSCWithNoProv,
   getAssociatedNodes,
-  shouldDeployAttachedAsMinimal,
+  shouldDeployAsMinimal,
 } from '../../../utils/install';
 import { getSCAvailablePVs } from '../../../selectors';
 import '../ocs-install.scss';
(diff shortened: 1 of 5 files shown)


frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
```diff
@@ -22,9 +23,7 @@ type MinimalDeploymentAlertProps = {
   isInternalMode?: boolean;
 };
 
-export const MinimalDeploymentAlert: React.FC<MinimalDeploymentAlertProps> = ({
-  isInternalMode,
-}) => (
+export const MinimalDeploymentAlert: React.FC<MinimalDeploymentAlertProps> = () => (
```
bipuladh, line 26:
```suggestion
export const MinimalDeploymentAlert: React.FC = () => (
```
Lets remove the type declaration as well.

bipuladh (commented):
nit

frontend/packages/ceph-storage-plugin/src/utils/common-ocs-install-el.tsx
gnehapk: ack. Missed it

bipuladh (approved):
/lgtm

gnehapk:
/hold
Waiting for confirmation for Alert text

bipuladh (approved):
/lgtm

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
```diff
@@ -100,7 +100,7 @@ export const CreateOCS = withHandlePromise<CreateOCSProps & HandlePromiseProps>(
   const isMinimalSupported = useFlag(OCS_SUPPORT_FLAGS.MINIMAL_DEPLOYMENT);
   const isEncryptionSupported = useFlag(OCS_SUPPORT_FLAGS.ENCRPYTION);
 
-  const isMinimal = shouldDeployAttachedAsMinimal(nodes);
+  const isMinimal = shouldDeployAsMinimal(nodes);
```
bipuladh, line 103:
```suggestion
  const isMinimal = nodes.length >= 3 ? shouldDeployAsMinimal(nodes) : false;
```

bipuladh (changes requested):
For LSO this is required because we shouldn't be showing the error and warning at the same time.
/lgtm cancel

bipuladh (approved):
/lgtm

PR #6746, opened 2020-09-25
Bug 1866337: Changes placeholder to 'Search by node name/label' in install page

vbnrh:
This PR add fixes to previous https://github.com/openshift/console/pull/6621. Now we can pass a resource name to display the placeholder appropriately and it will change the placeholder according to the type (Name, Label etc) selected 

### DIFF
--- a/frontend/public/components/filter-toolbar.tsx
+++ b/frontend/public/components/filter-toolbar.tsx
@@ -91,7 +91,8 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
     hideLabelFilter,
     hideNameLabelFilters,
     columnLayout,
-    filterPlaceholder,
+    nameFilterPlaceholder,
+    labelFilterPlaceholder,
     location,
     textFilter = filterTypeMap[FilterType.NAME],
     labelFilter = filterTypeMap[FilterType.LABEL],
@@ -100,8 +101,9 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
   const [inputText, setInputText] = React.useState('');
   const [filterType, setFilterType] = React.useState(FilterType.NAME);
   const [isOpen, setOpen] = React.useState(false);
-  const placeholder =
-    filterPlaceholder || (FilterType.NAME === filterType ? 'Search by name...' : 'app=frontend');
+  const [placeholder, setPlaceholder] = React.useState(
+    nameFilterPlaceholder || 'Search by name...',
+  );
 
   // (rowFilters) => {'rowFilterTypeA': ['staA', 'staB'], 'rowFilterTypeB': ['stbA'] }
   const filters: Filter = rowFilters.reduce((acc, curr) => {
@@ -258,6 +260,16 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
 
   const switchFilter = (type: FilterType) => {
     setFilterType(FilterType[type]);
+    switch (FilterType[type]) {
+      case 'Name':
+        setPlaceholder(nameFilterPlaceholder || 'Search by name...');
+        break;
+      case 'Label':
+        setPlaceholder(labelFilterPlaceholder || 'Search by label...');
+        break;
+      default:
+        setPlaceholder('app=frontend');
+    }
     setInputText(nameFilter && FilterType[type] === FilterType.NAME ? nameFilter : '');
   };
 
@@ -377,7 +389,8 @@ type FilterToolbarProps = {
   kinds?: any;
   labelPath?: string;
   columnLayout?: ColumnLayout;
-  filterPlaceholder?: string;
+  nameFilterPlaceholder?: string;
+  labelFilterPlaceholder?: string;
 };
 
 export type RowFilter<R = any> = {
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/create-form.tsx
@@ -188,7 +188,8 @@ export const CreateInternalCluster = withHandlePromise<
           customData={{
             onRowSelected: setNodes,
           }}
-          filterPlaceholder="Search by node name..."
+          nameFilterPlaceholder="Search by node name..."
+          labelFilterPlaceholder="Search by node label..."
         >
           <>
             <div>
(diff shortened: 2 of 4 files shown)


frontend/public/components/filter-toolbar.tsx
```diff
@@ -100,8 +100,11 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
   const [inputText, setInputText] = React.useState('');
   const [filterType, setFilterType] = React.useState(FilterType.NAME);
   const [isOpen, setOpen] = React.useState(false);
-  const placeholder =
-    filterPlaceholder || (FilterType.NAME === filterType ? 'Search by name...' : 'app=frontend');
+  const [placeholder, setPlaceholder] = React.useState(
+    FilterType.NAME === filterType
+      ? `Search by ${kind !== undefined ? kind.toLowerCase() : ''} name...`
```
spadgett, line 105: I would just lowercase `kind`. It is awkward for longer names since the word boundaries are no longer apparent.

spadgett (commented):
What bug are we fixing specifically? Is it just that the placeholder is wrong when "Label" is selected? If so, I'd just change the prop to `nameFilterPlaceholder` and only use it for "Name."

frontend/public/components/filter-toolbar.tsx
vbnrh: The good reasons for using this approach and not just using nameFIlterPlaceholder  :- 

1. We won't  have to hardcode the resource ( Node ) and the type of filter (Name, Label)
1. filter-toolbar being a generic component can accommodate multiple types of filter in the future. This could handle those scenarios too.
1. We just can pass the type of resource by kind prop and let the filter automatically handle any resource types.

vbnrh: > I would just lowercase kind. It is awkward for longer names since the word boundaries are no longer apparent.

Sorry i didn't get this point. Could you explain a bit more ?

spadgett: You shouldn't just manipulate kind with `toLowercase`. Things like `SecurityContextContraint` become `securitycontextconstraint` which is hard to read. You would have to use the model display name.

Won't this change all the list pages instead of the one we want to change?

spadgett:
Hi, we're only a few days away from code freeze for 4.6, and this change affects all list pages in the console. Can we go with a simpler fix like

> Is it just that the placeholder is wrong when "Label" is selected? If so, I'd just change the prop to nameFilterPlaceholder and only use it for "Name."

/hold

frontend/public/components/filter-toolbar.tsx
```diff
@@ -258,6 +261,20 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
+      setPlaceholder(
+        nameFilterPlaceholder ||
+          (FilterType[type] === 'Name' ? 'Search by name...' : 'app=frontend'),
+      );
+    }
+
+    if (FilterType[type] === 'Label') {
+      setPlaceholder(
+        labelFilterPlaceholder ||
+          (FilterType[type] === 'Label' ? 'Search by label...' : 'app=frontend'),
```
spadgett, line 274:
```suggestion
        labelFilterPlaceholder || 'app=frontend',
```

frontend/public/components/filter-toolbar.tsx
```diff
@@ -258,6 +261,20 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
 
   const switchFilter = (type: FilterType) => {
     setFilterType(FilterType[type]);
+    if (FilterType[type] === 'Name') {
+      setPlaceholder(
+        nameFilterPlaceholder ||
+          (FilterType[type] === 'Name' ? 'Search by name...' : 'app=frontend'),
```
spadgett, line 267:
```suggestion
        nameFilterPlaceholder || 'Search by name...',
```

frontend/public/components/filter-toolbar.tsx
```diff
@@ -258,6 +261,20 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
 
   const switchFilter = (type: FilterType) => {
     setFilterType(FilterType[type]);
+    if (FilterType[type] === 'Name') {
```
spadgett, line 264: This is probably better as a switch statement.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -100,8 +101,10 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
   const [inputText, setInputText] = React.useState('');
   const [filterType, setFilterType] = React.useState(FilterType.NAME);
   const [isOpen, setOpen] = React.useState(false);
-  const placeholder =
-    filterPlaceholder || (FilterType.NAME === filterType ? 'Search by name...' : 'app=frontend');
+  const [placeholder, setPlaceholder] = React.useState(
+    nameFilterPlaceholder ||
+      (FilterType.NAME === filterType ? 'Search by name...' : 'app=frontend'),
+  );
```
spadgett, line 107: Filter type is initialized to name, so you don't need to check it here.

```suggestion
  const [placeholder, setPlaceholder] = React.useState(nameFilterPlaceholder || 'Search by name...');
```

spadgett (changes requested):

frontend/public/components/filter-toolbar.tsx
vbnrh: Instead of 'app=frontend' i could keep it as 'Search by label' so that if `labelFilterPlaceholder` is not set then switching the filter keeps the placeholder appropriate. In any case QA is looking for this as stated in the bug

frontend/public/components/filter-toolbar.tsx
vbnrh: done

frontend/public/components/filter-toolbar.tsx
vbnrh: done

frontend/public/components/filter-toolbar.tsx
vbnrh: changed

frontend/public/components/filter-toolbar.tsx
vbnrh: I have added `app=frontend` as default

frontend/public/components/filter-toolbar.tsx
vbnrh: Yes i understand your concern on readability in which case we may need to keep non lower cased name. 
> Won't this change all the list pages instead of the one we want to change?

Yes that was the initial idea that if we are not passing the kind prop, it will just show 'Search by name' instead of 'Search by {kind} name'

spadgett (approved):
/lgtm

PR #6747, opened 2020-09-25
Bug 1879911: (security) dont use the hardcoded passwords

suomiy:
this PR supeseedes https://github.com/openshift/console/pull/6690


jelkosz:
/hold
looks like it breaks the chpasswd section of the cloud init

jelkosz:
/hold cancel
sorry, false alarm, it is still a valid yaml and it does work with cloudinit

PR #6748, opened 2020-09-25
Bug 1881211: use alertmanager-tenancy api end point to create silence in devconsole

vikram-raj:
Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1881211

/cc @kyoto @s-urbaniak


spadgett (approved):
/lgtm

PR #6749, opened 2020-09-25
Bug 1880698: Query Browser: Improve graph render speed by not using VictoryTooltip

kyoto:
We are using almost none of the functionality of `VictoryTooltip`
because we have a completely customized tooltip. We can therefore use
our own simple component instead of `VictoryTooltip` with just the
functionality we need.

This takes https://github.com/openshift/console/pull/6685 one step further by not using `VictoryTooltip` either.

In my testing, I'm seeing a big performance boost with render times typically reduced by more than half.


spadgett (approved):
/lgtm

Nice! Thank you 👍

kyoto:
FYI @dtaylor113, @cshinn, @sichvoge, @dlabrecq, @christianvogt

PR #6753, opened 2020-09-26
Bug 1880246:Entry keys in aggregated Helm index are appended with the name of source Helm repo

pedjak:
The introduced convention enables that all charts with the same name, but
in different source Helm repo, appear in the aggregated index file under
the name following the pattern: `foo-chart--my-repo`, where `my-repo` is the name
of `HelmChartRepository` available in the cluster.

### DIFF
--- a/pkg/helm/chartproxy/proxy.go
+++ b/pkg/helm/chartproxy/proxy.go
@@ -69,13 +69,15 @@ func (p *proxy) IndexFile() (*repo.IndexFile, error) {
 		return nil, err
 	}
 
-	var indexFiles []*repo.IndexFile
+	indexFile := repo.NewIndexFile()
 	for _, helmRepo := range helmRepos {
 		idxFile, err := helmRepo.IndexFile()
 		if err != nil {
 			return nil, err
 		}
-		indexFiles = append(indexFiles, idxFile)
+		for key, entry := range idxFile.Entries {
+			indexFile.Entries[key+"--"+helmRepo.Name] = entry
+		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/helm/chartproxy/proxy.go
```diff
@@ -69,13 +69,15 @@ func (p *proxy) IndexFile() (*repo.IndexFile, error) {
...
-	var indexFiles []*repo.IndexFile
+	indexFile := repo.NewIndexFile()
 	for _, helmRepo := range helmRepos {
 		idxFile, err := helmRepo.IndexFile()
 		if err != nil {
 			return nil, err
 		}
-		indexFiles = append(indexFiles, idxFile)
+		for key, entry := range idxFile.Entries {
+			indexFile.Entries[key+"--"+helmRepo.Name] = entry
```
rohitkrai03, line 79: Maybe we add '~~' as a delimiter between the two names to make it more unique?
```suggestion
			indexFile.Entries[key+"~~"+helmRepo.Name] = entry
```

rohitkrai03 (changes requested):

pkg/helm/chartproxy/proxy.go
pedjak: this char is not alowed - only letters, digits and dash

rohitkrai03: Okay, no problem.

rohitkrai03 (approved):
Looks good to me. Tested the backend changes with UI changes in https://github.com/openshift/console/pull/6761. Works well.

PR #6756, opened 2020-09-28
Bug 1883200: fix silence toggle in devconsole monitoring

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4423

**GIF**
Kapture 2020-10-01 at 12 43 25

**Test Setup**
1. `oc apply -f https://gist.githubusercontent.com/vikram-raj/d801e9ef6400cc1bd5af5a336bbedfa0/raw/61d90b705617ba8070da80f8c97ac16376a9ca58/prometheus-example-app.yaml`
2. `oc apply -f https://gist.githubusercontent.com/vikram-raj/86300d923a7a3a3a3348e359327e2307/raw/535d889923141a3f3e26e355d6bccec4650a0fcd/user-workload-config.yaml`
3. After applying the above 2 YAML it will create a rule and alert in `ns1` namespace. But in local setup all alerts and rules will be visible so, filter by name `CriticalAlert` and to test this PR.

### DIFF
--- a/frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
@@ -1,36 +1,75 @@
 import * as React from 'react';
 import * as _ from 'lodash';
+// FIXME upgrading redux types is causing many errors at this time
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useDispatch } from 'react-redux';
 import { Switch } from '@patternfly/react-core';
 import { Rule, RuleStates } from '@console/internal/components/monitoring/types';
 import { StateTimestamp } from '@console/internal/components/monitoring/alerting';
 import { coFetchJSON } from '@console/internal/co-fetch';
 import SilenceDurationDropDown from './SilenceDurationDropdown';
+import { monitoringSetRules } from '@console/internal/actions/ui';
+import {
+  ALERT_MANAGER_TENANCY_BASE_PATH,
+  PROMETHEUS_TENANCY_BASE_PATH,
+} from '@console/internal/components/graphs';
+import { getAlertsAndRules } from '@console/internal/components/monitoring/utils';
+import { alertingRuleStateOrder } from '@console/internal/reducers/monitoring';
+import { refreshNotificationPollers } from '@console/internal/components/notification-drawer';
 
 type SilenceAlertProps = {
   rule: Rule;
+  namespace: string;
 };
-
-const { alertManagerBaseURL } = window.SERVER_FLAGS;
 
 const SilenceUntil = ({ rule }) => {
   if (!_.isEmpty(rule.silencedBy)) {
-    return <StateTimestamp text="Until" timestamp={_.max(_.map(rule.silencedBy, 'endsAt'))} />;
+    return (
+      <div onClick={(e) => e.preventDefault()} role="presentation">
+        <StateTimestamp text="Until" timestamp={_.max(_.map(rule.silencedBy, 'endsAt'))} />
+      </div>
+    );
   }
   return null;
 };
 
-const SilenceAlert: React.FC<SilenceAlertProps> = ({ rule }) => {
-  const [isChecked, setIsChecked] = React.useState(rule.state !== RuleStates.Silenced);
-  const [ruleState] = React.useState(rule);
-  React.useEffect(() => setIsChecked(ruleState.state !== RuleStates.Silenced), [ruleState]);
+const SilenceAlert: React.FC<SilenceAlertProps> = ({ rule, namespace }) => {
+  const [isChecked, setIsChecked] = React.useState(true);
+  const [isInprogress, setIsInprogress] = React.useState(false);
+  const dispatch = useDispatch();
+
+  React.useEffect(() => {
+    if (rule.state === RuleStates.Silenced) {
+      setIsChecked(false);
+    }
+  }, [rule]);
 
   const handleChange = (checked: boolean) => {
+    setIsChecked(checked);
     if (checked) {
       _.each(rule.silencedBy, (silence) => {
-        coFetchJSON.delete(`${alertManagerBaseURL}/api/v2/silence/${silence.id}`);
+        coFetchJSON
+          .delete(`${ALERT_MANAGER_TENANCY_BASE_PATH}/api/v2/silence/${silence.id}`)
+          .then(() => {
+            refreshNotificationPollers();
+            // eslint-disable-next-line promise/no-nesting
+            return coFetchJSON(
+              `${PROMETHEUS_TENANCY_BASE_PATH}/api/v1/rules?namespace=${namespace}`,
+            ).then((response) => {
+              const thanosAlertsAndRules = getAlertsAndRules(response?.data);
+              const sortThanosRules = _.sortBy(thanosAlertsAndRules.rules, alertingRuleStateOrder);
+              dispatch(monitoringSetRules('devRules', sortThanosRules, 'dev'));
+              setIsChecked(true);
+            });
+          })
+          .catch((err) => {
+            // eslint-disable-next-line no-console
+            console.warn('Could not expire silence:', err);
+            setIsChecked(false);
+          });
       });
     }
-    setIsChecked(checked);
   };
 
   return (
@@ -39,12 +78,16 @@ const SilenceAlert: React.FC<SilenceAlertProps> = ({ rule }) => {
       className="odc-silence-alert"
       label={null}
       labelOff={
-        rule.state === RuleStates.Silenced ? (
+        rule.state === RuleStates.Silenced && !isChecked ? (
           <SilenceUntil rule={rule} />
         ) : (
-          <SilenceDurationDropDown rule={rule} />
+          <SilenceDurationDropDown
+            silenceInProgress={(progress) => setIsInprogress(progress)}
+            rule={rule}
+          />
         )
       }
+      isDisabled={isInprogress}
       isChecked={isChecked}
       onChange={handleChange}
     />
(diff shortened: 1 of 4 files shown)


frontend/packages/dev-console/src/components/monitoring/alerts/monitoring-alerts-utils.tsx
```diff
@@ -157,13 +152,15 @@ const setOrderBy = (orderBy: SortByDirection, data: Rule[]): Rule[] => {
   return orderBy === SortByDirection.asc ? data : data.reverse();
 };
 
-const alertingRuleNotificationsOrder = (rule: Rule) => [
-  rule.state === RuleStates.Silenced ? 1 : 0,
-  rule.state,
-];
+export const alertingRuleNotificationsOrder = (rule: Rule) => {
```
nemesis09, line 155: verified locally. silencing works as expected.
just one nit: could you please add the return type for the function
`export const alertingRuleNotificationsOrder = (rule: Rule): number[] => {`

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
```diff
@@ -48,18 +60,42 @@ const SilenceDurationDropDown: React.FC<SilenceDurationDropDownProps> = ({ rule
...
-
-      coFetchJSON.post(`${ALERT_MANAGER_TENANCY_BASE_PATH}/api/v2/silences`, payload);
+      setSilencing(true);
+      coFetchJSON
+        .post(`${ALERT_MANAGER_TENANCY_BASE_PATH}/api/v2/silences`, payload)
+        .then(() => {
+          // eslint-disable-next-line promise/no-nesting
+          coFetchJSON(`${ALERT_MANAGER_TENANCY_BASE_PATH}/api/v2/silences`)
+            .then((silences) => {
+              rule.silencedBy = _.filter(
```
nemesis09, line 70: `rule` is a prop received by the component. ideally we should avoid modification of props. we could use a deep copy of the prop instead. WDYT?

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
```diff
@@ -48,18 +61,43 @@ const SilenceDurationDropDown: React.FC<SilenceDurationDropDownProps> = ({ rule
...
+                (s) =>
+                  s.status.state === SilenceStates.Active && _.some(clonedRule.alerts, isSilenced),
+              );
+              if (!_.isEmpty(clonedRule.silencedBy)) {
+                _.each(clonedRule.alerts, (a) => (a.state = AlertStates.Silenced));
+                clonedRule.state = RuleStates.Silenced;
+              }
+              const updatedRules = [...rules, ...[clonedRule]];
+              dispatch(monitoringSetRules('devRules', updatedRules, 'dev'));
+              setSilencing(false);
```
karthikjeeyar, line 82: ` setSilencing(false);  
`
This needs to be set on the catch block as well, or else loader is shown there indefinitely even after the error modal is closed.
eg: 
image

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.scss
```diff
@@ -0,0 +1,3 @@
+.silence-duration-dropdown {
```
karthikjeeyar, line 1:
```suggestion
.odc-silence-duration-dropdown {
```

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
```diff
@@ -48,18 +61,43 @@ const SilenceDurationDropDown: React.FC<SilenceDurationDropDownProps> = ({ rule
...
-    <Dropdown
-      dropDownClassName="dropdown--full-width"
-      items={durations}
-      onChange={(v: string) => setDuration(v)}
-      selectedKey={'silenceFor'}
-    />
+    <>
+      <Dropdown
+        dropDownClassName="dropdown--full-width"
+        className="silence-duration-dropdown"
```
karthikjeeyar, line 94:
```suggestion
        className="odc-silence-duration-dropdown"
```

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
```diff
@@ -20,14 +27,35 @@ const SilenceUntil = ({ rule }) => {
...
+    }
+  }, [rule]);
 
   const handleChange = (checked: boolean) => {
     if (checked) {
       _.each(rule.silencedBy, (silence) => {
-        coFetchJSON.delete(`${alertManagerBaseURL}/api/v2/silence/${silence.id}`);
+        coFetchJSON
+          .delete(`${alertManagerBaseURL}/api/v2/silence/${silence.id}`)
+          .then(() => {
```
kyoto, line 44: In the admin Console, after deleting a silence, we call `refreshNotificationPollers`, which immediately fetches the latest silences from Alertmanager and then triggers the evaluation of which alerts are silenced. I think that would achieve the same thing you are doing here in a simpler way.

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
```diff
@@ -20,14 +27,35 @@ const SilenceUntil = ({ rule }) => {
...
+            _.each(rule.alerts, (a) => {
+              _.omit(a, 'silencedBy');
```
kyoto, line 46: `_.omit` returns a new object rather than modifying `a`, so it looks like this doesn't do anything useful.

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
```diff
@@ -20,14 +27,35 @@ const SilenceUntil = ({ rule }) => {
...
+              a.state = AlertStates.NotFiring;
+            });
+            _.omit(rule, 'silencedBy');
```
kyoto, line 49: Same use of `_.omit` as above

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
```diff
@@ -37,29 +48,58 @@ const SilenceDurationDropDown: React.FC<SilenceDurationDropDownProps> = ({ rule
...
+    const payload = {
+      createdBy,
+      endsAt: endsAt.toISOString(),
+      startsAt: startsAt.toISOString(),
+      matchers,
+      comment: '',
+    };
+    setSilencing(true);
+    coFetchJSON
+      .post(`${ALERT_MANAGER_TENANCY_BASE_PATH}/api/v2/silences`, payload)
```
kyoto, line 63: Same comment as for the `delete`. In the admin Console, after creating a silence, we call `refreshNotificationPollers` to trigger a refresh of the alert's `state`, so I think we could do the same here.

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
andrewballantyne: @vikram-raj This definitely isn't doing anything but wasting part of a cycle to remove a field from an object that is immediately tossed up to the garbage collector. I think what you want to do is either change this to a map (instead of a for-each) or the less desirable choice is to use `delete a.silencedBy`. 

However, given the context and that `rules` is not your object (it comes from redux) you should not even be doing `a.state = <something>` as you're mutating Redux state when you do this.

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
vikram-raj: Our scenario is a little different. In our case, we need to trigger the refresh for rules and silences. But in `refreshNotificationPollers` it is refreshing only silences.

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceAlert.tsx
```diff
@@ -14,23 +20,51 @@ const { alertManagerBaseURL } = window.SERVER_FLAGS;
...
+    }
+  }, [rule]);
 
   const handleChange = (checked: boolean) => {
+    setIsChecked(checked);
     if (checked) {
       _.each(rule.silencedBy, (silence) => {
-        coFetchJSON.delete(`${alertManagerBaseURL}/api/v2/silence/${silence.id}`);
+        coFetchJSON
+          .delete(`${alertManagerBaseURL}/api/v2/silence/${silence.id}`)
```
kyoto, line 49: I think this should be using `ALERT_MANAGER_TENANCY_BASE_PATH` instead of `alertManagerBaseURL`

invincibleJai:
@vikram-raj there is improvement with handling toggle component based on issue description, but i see some delay still in silences getting updated and user can toggle again meanwhile

Oct-02-2020 14-53-17

frontend/packages/dev-console/src/components/monitoring/alerts/SilenceDurationDropdown.tsx
kyoto: I looked into this some more and I still think we can simplify the code here by using `refreshNotificationPollers` or a similar approach. However, I'm OK with addressing that with a follow up PR.

vikram-raj:
@invincibleJai I tested it locally and in the cluster as well. And did not notice any delay.

Kapture 2020-10-02 at 16 26 47

invincibleJai:
/approve

Thanks @vikram-raj looks good now after the changes, verified it again

jerolimov (approved):

/lgtm

PR #6765, opened 2020-09-28
CONSOLE-2297: Added Create Namespace option during operator install

harishsurf:
This change adds an option `Create Namespace` to the namespace dropdown list in the install operators page.
Jira: https://issues.redhat.com/browse/CONSOLE-2297

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -27,7 +27,10 @@
     "test-gui": "yarn run test-suite --suite all",
     "test-suite": "ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/protractor integration-tests/protractor.conf.ts",
     "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
-    "test-cypress-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-olm": "cd packages/operator-lifecycle-manager/integration-tests-cypress && ../../../node_modules/.bin/cypress open --config-file cypress-olm.json --env openshift=true",
+    "test-cypress-console-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-olm-headless": "cd packages/operator-lifecycle-manager/integration-tests-cypress && node --max-old-space-size=4096 ../../../node_modules/.bin/cypress run --config-file cypress-olm.json --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-headless": "yarn test-cypress-console-headless && yarn test-cypress-olm-headless",
     "cypress-merge": "mochawesome-merge ./gui_test_screenshots/cypress_report*.json > ./gui_test_screenshots/cypress.json",
     "cypress-generate": "marge -o ./gui_test_screenshots/ -f cypress-report -t 'OpenShift Console Cypress Test Results' -p 'OpenShift Cypress Test Results' --showPassed false --assetsDir ./gui_test_screenshots/cypress/assets ./gui_test_screenshots/cypress.json",
     "cypress-a11y-report": "echo '\nA11y Test Results:' && mv packages/integration-tests-cypress/cypress-a11y-report.json ./gui_test_screenshots/ && node -e \"console.table(JSON.parse(require('fs').readFileSync(process.argv[1])));\" ./gui_test_screenshots/cypress-a11y-report.json",
@@ -66,7 +69,7 @@
     ],
     "testPathIgnorePatterns": [
       "<rootDir>/node_modules/",
-      "<rootDir>/packages/integration-tests-cypress"
+      "<rootDir>/.*/integration-tests-cypress"
     ],
     "testRegex": ".*\\.spec\\.(ts|tsx|js|jsx)$",
     "testURL": "http://localhost",
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
@@ -1,0 +1,57 @@
+import { checkErrors, testName } from '../../../integration-tests-cypress/support';
+import { modal } from '../../../integration-tests-cypress/views/modal';
+
+describe('Create namespace from install operators', () => {
+  before(() => {
+    cy.login();
+    cy.createProject(testName);
+  });
+
+  afterEach(() => {
+    checkErrors();
+  });
+
+  after(() => {
+    cy.deleteProject(testName);
+    cy.logout();
+  });
+
+  const nsName = `${testName}-ns`;
+
+  it('creates namespace from operator install page', () => {
+    const operatorSelector = 'advanced-cluster-management-redhat-operators-openshift-marketplace';
+    cy.log('test namespace creation from dropdown');
+    cy.visit(`/operatorhub/ns/${testName}`);
+    cy.byTestID(operatorSelector).click();
+    cy.byLegacyTestID('operator-install-btn').click({ force: true });
+
+    // configure operator install
+    cy.byTestID('Select a namespace-radio-input').check();
+    cy.byTestID('dropdown-selectbox')
+      .click()
+      .get('[data-test-dropdown-menu="Create_Namespace"]')
+      .click();
+
+    // verify namespace modal is opened
+    modal.shouldBeOpened();
+    cy.byTestID('input-name').type(nsName);
+    modal.submit();
+    modal.shouldBeClosed();
+
+    // verify the dropdown selection shows the newly created namespace
+    cy.byTestID('dropdown-selectbox').should('contain', `${nsName}`);
+
+    cy.get('button')
+      .contains('Install')
+      .click();
+
+    // verify operator began installation
+    cy.byTestID('view-installed-operators-btn').should(
+      'contain',
+      `View Installed Operators in namespace ${nsName}`,
+    );
+
+    // Verify namespace was created successfully
+    cy.deleteProject(nsName);
+  });
+});
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
@@ -506,7 +506,8 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
     <NsDropdown
       id="dropdown-selectbox"
       selectedKey={selectedTargetNamespace}
-      onChange={setTargetNamespace}
+      onChange={(ns) => setTargetNamespace(ns)}
+      dataTest="dropdown-selectbox"
     />
   ) : (
     <div className="form-group">
@@ -531,13 +532,14 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
         }}
         value={suggestedNamespace}
         checked={!useSuggestedNSForSingleInstallMode}
-        title="Pick an existing namespace"
+        title="Select a namespace"
       />
       {!useSuggestedNSForSingleInstallMode && (
         <NsDropdown
           id="dropdown-selectbox"
           selectedKey={selectedTargetNamespace}
-          onChange={setTargetNamespace}
+          onChange={(ns) => setTargetNamespace(ns)}
+          dataTest="dropdown-selectbox"
         />
       )}
     </div>
--- a/frontend/public/components/utils/list-dropdown.jsx
+++ b/frontend/public/components/utils/list-dropdown.jsx
@@ -3,13 +3,20 @@ import * as React from 'react';
 import * as fuzzy from 'fuzzysearch';
 import * as PropTypes from 'prop-types';
 import { Alert } from '@patternfly/react-core';
-
+import { useFlag } from '@console/shared/src/hooks/flag';
 import { FLAGS } from '@console/shared/src/constants';
 import { Dropdown } from './dropdown';
 import { Firehose } from './firehose';
 import { LoadingInline } from './status-box';
 import { ResourceName } from './resource-icon';
-import { connectToFlags, flagPending } from '../../reducers/features';
+import { flagPending } from '../../reducers/features';
+import { useAccessReview } from '@console/internal/components/utils';
+import { createNamespaceModal, createProjectModal } from '../modals';
+import { NamespaceModel, ProjectModel } from '@console/internal/models';
+
+const getKey = (key, keyKind) => {
+  return keyKind ? `${key}-${keyKind}` : key;
+};
 
 class ListDropdown_ extends React.Component {
   constructor(props) {
@@ -19,9 +26,7 @@ class ListDropdown_ extends React.Component {
     };
 
     if (props.selectedKey) {
-      this.state.selectedKey = props.selectedKeyKind
-        ? `${props.selectedKey}-${props.selectedKeyKind}`
-        : props.selectedKey;
+      this.state.selectedKey = getKey(props.selectedKey, props.selectedKeyKind);
     }
 
     this.state.title = props.loaded ? props.placeholder : <LoadingInline />;
@@ -29,9 +34,12 @@ class ListDropdown_ extends React.Component {
     this.autocompleteFilter = (text, item) => fuzzy(text, item.props.name);
     // Pass both the resource name and the resource kind to onChange()
     this.onChange = (key) => {
+      if (_.find(this.props.actionItems, { actionKey: key })) {
+        this.props.onChange(key);
+      }
       const { name, kindLabel, resource } = _.get(this.state, ['items', key], {});
       this.setState({ selectedKey: key, title: <ResourceName kind={kindLabel} name={name} /> });
-      this.props.onChange(name, kindLabel, resource);
+      this.props.onChange?.(name, kindLabel, resource);
     };
   }
 
@@ -49,12 +57,13 @@ class ListDropdown_ extends React.Component {
     loadError,
     resources,
     dataFilter,
+    ...nextProps
   }) {
     if (!loaded) {
       return;
     }
 
-    this.setState(({ selectedKey }) => {
+    this.setState((currentState) => {
       if (loadError) {
         return {
           title: <div className="cos-error-title">Error Loading {desc}</div>,
@@ -85,8 +94,17 @@ class ListDropdown_ extends React.Component {
         .forEach((key) => {
           sortedList[key] = unsortedList[key];
         });
+
+      const keyChanged = currentState.selectedKey !== nextProps.selectedKey;
+      const keyKindChanged = currentState.selectedKeyKind !== nextProps.selectedKeyKind;
+      const selectedKey =
+        keyChanged || keyKindChanged
+          ? getKey(nextProps.selectedKey, nextProps.selectedKeyKind)
+          : currentState.selectedKey;
+
       const selectedItem = sortedList[selectedKey];
       return {
+        selectedKey,
         items: sortedList,
         title: selectedItem ? (
           <ResourceName kind={selectedItem.kindLabel} name={selectedItem.name} />
@@ -116,6 +134,7 @@ class ListDropdown_ extends React.Component {
       items[selectedKey]
     ) : (
       <Dropdown
+        actionItems={this.props.actionItems}
         autocompleteFilter={this.autocompleteFilter}
         autocompletePlaceholder={placeholder}
         items={items}
@@ -175,22 +194,77 @@ ListDropdown.propTypes = {
   id: PropTypes.string,
 };
 
-const NsDropdown_ = (props) => {
-  const openshiftFlag = props.flags[FLAGS.OPENSHIFT];
-  if (flagPending(openshiftFlag)) {
-    return null;
+const useProjectOrNamespaceModel = () => {
+  const canCreateNamespace = useAccessReview({
+    group: NamespaceModel.apiGroup,
+    resource: NamespaceModel.plural,
+    verb: 'create',
+  });
+
+  const canCreateProject = useFlag(FLAGS.CAN_CREATE_PROJECT);
+  const openshiftFlag = useFlag(FLAGS.OPENSHIFT);
+
+  if (flagPending(openshiftFlag) || flagPending(canCreateProject)) {
+    return [];
   }
-  const kind = openshiftFlag ? 'Project' : 'Namespace';
-  const resources = [{ kind }];
-  return (
+
+  // NamespaceModal is used when not on an openshift cluster
+  const model = canCreateNamespace || !openshiftFlag ? NamespaceModel : ProjectModel;
+  const canCreate = canCreateNamespace || (openshiftFlag && canCreateProject);
+  return [model, canCreate];
+};
+
+/** @type {React.FC<{dataFilter?: (ns: any) => boolean, desc?: string, selectedKey?: string, fixed?: boolean, placeholder?: string, onChange?: (selectedKey: string, selectedKeyKind: string,  selectedResource?: K8sResourceKind) => void, id?: string, dataTest?: string}}>} */
+export const NsDropdown = (props) => {
+  const [selectedKey, setSelectedKey] = React.useState(props.selectedKey);
+  const [model, canCreate] = useProjectOrNamespaceModel();
+
+  const actionItems =
+    model && canCreate
+      ? [
+          {
+            actionTitle: `Create ${model.label}`,
+            actionKey: `Create_${model.label}`,
+          },
+        ]
+      : [];
+
+  const onChange = (actionKey, kindLabel, resource) => {
+    switch (actionKey) {
+      case 'Create_Namespace':
+        createNamespaceModal({
+          blocking: true,
+          onSubmit: (newNamespace) => {
+            setSelectedKey(newNamespace.metadata.name);
+            props.onChange?.(newNamespace.metadata.name, newNamespace.kind, newNamespace);
+          },
+        });
+        break;
+      case 'Create_Project':
+        createProjectModal({
+          blocking: true,
+          onSubmit: (newProject) => {
+            setSelectedKey(newProject.metadata.name);
+            props.onChange?.(newProject.metadata.name, newProject.kind, newProject);
+          },
+        });
+        break;
+      default:
+        props.onChange?.(actionKey, kindLabel, resource);
+        break;
+    }
+  };
+
+  return model ? (
     <ListDropdown
       {...props}
-      desc="Namespaces"
-      resources={resources}
-      selectedKeyKind={kind}
-      placeholder="Select namespace"
+      actionItems={actionItems}
+      desc={model.plural}
+      onChange={onChange}
+      placeholder={`Select ${model.kind}`}
+      resources={[{ kind: `${model.kind}` }]}
+      selectedKeyKind={model.kind}
+      selectedKey={selectedKey}
     />
-  );
+  ) : null;
 };
-/** @type {React.FC<{dataFilter?: (ns: any) => boolean, desc?: string, selectedKey?: string, selectedKeyKind?: string, fixed?: boolean, placeholder?: string, onChange?: (selectedKey: string, event: React.Event) => void, id?: string}}>} */
-export const NsDropdown = connectToFlags(FLAGS.OPENSHIFT)(NsDropdown_);
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/cypress-olm.json
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/cypress-olm.json
@@ -1,0 +1,18 @@
+{
+  "integrationFolder": "tests",
+  "screenshotsFolder": "../../../gui_test_screenshots/cypress/screenshots",
+  "videosFolder": "../../../gui_test_screenshots/cypress/videos",
+  "video": true,
+  "reporter": "../../../node_modules/cypress-multi-reporters",
+  "reporterOptions": {
+    "configFile": "reporter-config.json"
+  },
+  "supportFile": "../../integration-tests-cypress/support/index.ts",
+  "pluginsFile": "../../integration-tests-cypress/plugins/index.js",
+  "fixturesFolder": "fixtures",
+  "defaultCommandTimeout": 30000,
+  "retries": {
+    "runMode": 1,
+    "openMode": 0
+  }
+}
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/reporter-config.json
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/reporter-config.json
@@ -1,0 +1,14 @@
+{
+  "reporterEnabled": "mocha-junit-reporter, mochawesome",
+  "mochaJunitReporterReporterOptions": {
+    "mochaFile": "../../../gui_test_screenshots/junit_cypress-[hash].xml",
+    "toConsole": false
+  },
+  "mochawesomeReporterOptions": {
+    "reportDir": "../../../gui_test_screenshots/",
+    "reportFilename": "cypress_report_olm",
+    "overwrite": false,
+    "html": false,
+    "json": true
+  }
+}
--- a/frontend/public/components/radio.tsx
+++ b/frontend/public/components/radio.tsx
@@ -15,7 +15,7 @@ export const RadioInput: React.SFC<RadioInputProps> = (props) => {
       <label
         className={classNames({ 'radio-inline': props.inline, 'co-disabled': props.disabled })}
       >
-        <input type="radio" {...inputProps} />
+        <input type="radio" {...inputProps} data-test={`${props.title}-radio-input`} />
         {props.title} {props.subTitle && <span className="co-no-bold">{props.subTitle}</span>}
       </label>
       {props.desc && <p className="co-m-radio-desc text-muted">{props.desc}</p>}


frontend/public/components/utils/dropdown.jsx
```diff
@@ -388,6 +388,8 @@ export class Dropdown extends DropdownMixin {
               onclick={this.onClick}
               selected={ai.actionKey === selectedKey && !noSelection}
               hover={ai.actionKey === keyboardHoverKey}
+              actionItems={actionItems}
+              onChange={this.props.onChange}
```
TheRealJon, line 392: These changes shouldn't be necessary. `this.onClick` will be called whenever an action item is clicked, and I don't see a need to pass all action items into each action item row.
```suggestion
              actionItems={actionItems}
              onChange={this.props.onChange}
```

harishsurf: Makes sense. Removed it.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  const nsName = `${testName}-ns`;
+
+  it('creates namespace from operator install page', () => {
+    const operatorSelector = '3scale-community-operator-community-operators-openshift-marketplace';
+    cy.log('test namespace creation from dropdown');
+    cy.visit(`/operatorhub/ns/${testName}`);
+    cy.byTestID(operatorSelector).click();
+    cy.byLegacyTestID('operator-install-btn').click({ force: true });
```
harishsurf, line 27: I had to force the click (as a workaround) since cypress wasn't able to identify the `Install` button from the pop-up that opens after selecting an operator on the install operators page. I tried different selectors but had no luck. Not sure why cypress isn't able to recognize this (maybe the focus shifts or that it is not a button element?). Please suggest if there is any way of achieving this.

frontend/package.json
```diff
@@ -28,6 +28,8 @@
     "test-suite": "ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/protractor integration-tests/protractor.conf.ts",
     "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
     "test-cypress-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-olm": "cd packages/operator-lifecycle-manager/integration-tests-cypress && cypress open --config-file cypress-olm.json --env openshift=true",
+    "test-cypress-olm-headless": "cd packages/operator-lifecycle-manager/integration-tests-cypress && node --max-old-space-size=4096 ../../../node_modules/.bin/cypress run --config-file cypress-olm.json --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
```
spadgett, line 32: We want this to run with you run `test-cypress` instead of having a separate yarn command. Then we can pass an arg to `test-cypress` if you want to run any particular scenario.

spadgett: We'll want to make sure this runs in CI when `test-gui.sh` is called.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
spadgett: What message does Cypress give when this happens?

cc @dtaylor113

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
```diff
@@ -0,0 +1,55 @@
...
+    cy.visit('/k8s/cluster/namespaces');
+    listPage.filter.byName(nsName);
+    listPage.rows.shouldExist(nsName);
+    listPage.rows.clickKebabAction(nsName, 'Delete Namespace');
+    modal.shouldBeOpened();
+    cy.byTestID('project-name-input').type(nsName);
+    cy.testA11y('Delete Namespace modal');
+    modal.submit();
+    modal.shouldBeClosed();
+    cy.resourceShouldBeDeleted(testName, 'namespaces', nsName);
```
spadgett, line 53: Can we call `cy.deleteProject(nsName)` instead? I believe this should work.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+
+  const prModel = modelFor('Project');
+  const useProjectAccessReview = () => {
+    const canCreateProject = useAccessReview({
+      group: prModel.apiGroup,
+      resource: prModel.plural,
+      verb: 'create',
+    });
+    return canCreateProject;
+  };
```
spadgett, line 213: Use the `CAN_CREATE_PROJECT` flag instead (`useFlag` hook). Projects are a special API with a different way to determine the user's permission.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+
+  const onChange = (actionKey) => {
+    if (actionKey === CREATE_NEW_NAMESPACE) {
+      createNamespaceModal({ blocking: true });
+    } else if (actionKey === CREATE_NEW_PROJECT) {
+      createProjectModal({ blocking: true });
+    }
+  };
+
+  const kind = useNamespaceAccessReview ? 'Namespace' : useProjectAccessReview ? 'Project' : '';
```
spadgett, line 239: Nesting ternary expressions gets confusing. I'd avoid that.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+
+  const nsModel = modelFor('Namespace');
+  const useNamespaceAccessReview = () => {
+    const canCreateNamespace = useAccessReview({
+      group: nsModel.apiGroup,
+      resource: nsModel.plural,
+      verb: 'create',
+    });
+    return canCreateNamespace;
+  };
```
spadgett, line 203: It's unclear to me why this is wrapped in a function. It will always be truthy in the checks below. Shouldn't this just be

```suggestion
  const canCreateNamespace = useAccessReview({
    group: nsModel.apiGroup,
    resource: nsModel.plural,
    verb: 'create',
  });
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+    ? [
+        {
+          actionTitle: `Create ${prModel.label}`,
+          actionKey: CREATE_NEW_PROJECT,
+        },
+      ]
+    : [];
+
+  const onChange = (actionKey) => {
+    if (actionKey === CREATE_NEW_NAMESPACE) {
```
spadgett, line 232: nit: this is better as a `switch` statement

spadgett (commented):
Thanks @harishsurf

frontend/public/components/utils/list-dropdown.jsx
harishsurf: The eslint was complaining: 
```
error  React Hook "useAccessReview" is called conditionally. React Hooks must be called in the exact same order in every component render 
```

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
harishsurf: Cypress waits for 7-10 seconds to locate the UI element and says the element is not visible. However, when I cycle through the snapshot previews of cypress, I see that it highlights the button, but the `cy.get()` just doesn't work.
install_button

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
```diff
@@ -0,0 +1,55 @@
+import { checkErrors, testName } from '../../../integration-tests-cypress/support';
```
dtaylor113, line 1: Pls rename to `create-namespace.spec.ts` -thanks

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
```diff
@@ -0,0 +1,55 @@
...
+  const nsName = `${testName}-ns`;
+
+  it('creates namespace from operator install page', () => {
+    const operatorSelector = '3scale-community-operator-community-operators-openshift-marketplace';
+    cy.log('test namespace creation from dropdown');
+    cy.visit(`/operatorhub/ns/${testName}`);
+    cy.byTestID(operatorSelector).click();
+    cy.byLegacyTestID('operator-install-btn').click({ force: true });
+
+    cy.get('#dropdown-selectbox').click();
```
dtaylor113, line 29: Hi, we are trying to stay away from using css id (#) selectors and using [Cypress best practices](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/README.md#best-practices) for selecting elements.   
A notable exception is if the elements is a third-party dependency where we cannot add a `data-test` attribute.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
```diff
@@ -0,0 +1,55 @@
...
+    cy.byLegacyTestID('resource-title').should('have.text', nsName);
+
+    cy.log('delete the Namespace');
+    cy.visit('/k8s/cluster/namespaces');
+    listPage.filter.byName(nsName);
+    listPage.rows.shouldExist(nsName);
+    listPage.rows.clickKebabAction(nsName, 'Delete Namespace');
+    modal.shouldBeOpened();
+    cy.byTestID('project-name-input').type(nsName);
+    cy.testA11y('Delete Namespace modal');
```
dtaylor113, line 50: :+1: :1st_place_medal:

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
TheRealJon: I'm wondering if this is caused by the `#modal-container` element, which has a calculated height of 0. I was reading over https://github.com/cypress-io/cypress/issues/695 and decided to inspect that button and it's parent elements to make sure none of them had an effective height/width of zero, and found this.

TheRealJon: Screen Shot 2020-10-05 at 11 27 23 AM

frontend/public/components/utils/list-dropdown.jsx
spadgett: > The eslint was complaining:
> 
> ```
> error  React Hook "useAccessReview" is called conditionally. React Hooks must be called in the exact same order in every component render 
> ```

I think your solution to wrap it in a function has the same problem. You're just hiding it from eslint :)

You need to move the hook above the OpenShift flag check so it always run.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+      createNamespaceModal({ blocking: true });
+    } else if (actionKey === CREATE_NEW_PROJECT) {
+      createProjectModal({ blocking: true });
+    }
+  };
+
+  const kind = useNamespaceAccessReview ? 'Namespace' : useProjectAccessReview ? 'Project' : '';
   const resources = [{ kind }];
   return (
```
TheRealJon, line 241:
```suggestion
  return flagPending(openshiftFlag) ? null : (
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
 
 const NsDropdown_ = (props) => {
   const openshiftFlag = props.flags[FLAGS.OPENSHIFT];
+
   if (flagPending(openshiftFlag)) {
     return null;
   }
-  const kind = openshiftFlag ? 'Project' : 'Namespace';
+
+  const nsModel = modelFor('Namespace');
```
TheRealJon, line 195: I would suggest writing a simple hook or helper function that will give us the correct model. This will eliminate the need for conditionals later in the component. Within that hook, You can use the `useFlag` hook to retrieve the `CAN_CREATE_PROJECTS` flag that @spadgett suggested, and `useAccessReview` for namespaces. If user can't create either, then the result will be undefined and we can handle that in the logic below.

```tsx
 const model = useProjectOrNamespaceModel(/* args */);
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
...
+        },
+      ]
+    : useProjectAccessReview
+    ? [
+        {
+          actionTitle: `Create ${prModel.label}`,
+          actionKey: CREATE_NEW_PROJECT,
+        },
+      ]
+    : [];
```
TheRealJon, line 229: If we establish the model for the resource the user can create as suggested above, we shouldn't need any condition here. We can just build the action key from the model.label as well. So the end result would be something like:

```tsx
  const actionItems = model ? [
    {
      actionTitle: `Create ${model.label}`,
      actionKey: `CREATE_${model.label}`,
    }
  ]
```

TheRealJon (commented):
Just a couple of suggestions to simplify the conditional logic here and avoid the eslint errors you were seeing with hooks.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -177,10 +187,56 @@ ListDropdown.propTypes = {
 
 const NsDropdown_ = (props) => {
   const openshiftFlag = props.flags[FLAGS.OPENSHIFT];
+
   if (flagPending(openshiftFlag)) {
     return null;
   }
```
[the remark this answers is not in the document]
TheRealJon: See https://github.com/openshift/console/pull/6765#discussion_r499706293 for why I would suggest removing this.

frontend/public/components/utils/list-dropdown.jsx
TheRealJon: You would also have to update your onChange function to use the new actionKeys.

TheRealJon:
I just tested this and it redirects to Home > Projects when you successfully create a new project from that dropdown. We will need to add a callback to the the modal to handle that. We should just set the dropdown's selected item to the newly created project/namespace instead of redirecting.

frontend/package.json
dtaylor113: @harishsurf, you'll need to follow what `dev-console` has done irt. updating the [yarn scripts](https://github.com/openshift/console/pull/6776/files#diff-0a6694dd0d6b93a2aa14f32e3d873b4c).  
You'll also need to include the Reporters [here](https://github.com/openshift/console/pull/6776/files#diff-6f2ef64ef74ee3c4091b2665adbb373dR20) and [here](https://github.com/openshift/console/pull/6776/files#diff-c42dd9cc6e3aad4af77fe8d29faad3ff).  As well as [copying](https://github.com/openshift/console/pull/6776/files#diff-6f2ef64ef74ee3c4091b2665adbb373dR24) cypress screenshots and recordings to `frontend/gui_test_screenshots`.

dtaylor113:
@TheRealJon, hope people don't get confused between our profile icons; although your doggie does look better in the fedora :-)

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+        ]
+      : [];
+
+  const onChange = (actionKey) => {
+    switch (actionKey) {
+      case 'Create_Namespace':
+        createNamespaceModal({
+          blocking: true,
+          onSubmit: (newNamespace) => {
+            setSelectedKey(newNamespace);
```
TheRealJon, line 228:
```suggestion
            setSelectedKey(newNamespace.metadata.name);
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+          },
+        });
+        break;
+      case 'Create_Project':
+        createProjectModal({
+          blocking: true,
+          onSubmit: (newProject) => {
+            setSelectedKey(newProject);
```
TheRealJon, line 236:
```suggestion
            setSelectedKey(newProject.metadata.name);
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+          },
+        });
+        break;
+      default:
+        break;
+    }
+  };
+
+  const kind = model.label;
   const resources = [{ kind }];
```
TheRealJon, line 246: If `model` is undefined here, which it can be, `model.label` will throw a runtime error. Even if we use optional chaining to prevent that, these variables will never get used when model is undefined, because we return null in that case. Better to just reference the properties of `model` below as you need them.

```suggestion
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
-  return (
+
+  return model ? (
     <ListDropdown
       {...props}
+      actionItems={actionItems}
       desc="Namespaces"
+      onChange={onChange}
+      placeholder={`Select ${kind}`}
```
TheRealJon, line 254:
```suggestion
      placeholder={`Select ${model.kind}`}
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
       resources={resources}
       selectedKeyKind={kind}
```
TheRealJon, line 256:
```suggestion
      selectedKeyKind={model.kind}
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
-  return (
+
+  return model ? (
     <ListDropdown
       {...props}
+      actionItems={actionItems}
       desc="Namespaces"
+      onChange={onChange}
+      placeholder={`Select ${kind}`}
       resources={resources}
```
TheRealJon, line 255:
```suggestion
      resources={[ kind: model.kind ]}
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+
+  const kind = model.label;
   const resources = [{ kind }];
-  return (
+
+  return model ? (
     <ListDropdown
       {...props}
+      actionItems={actionItems}
       desc="Namespaces"
```
TheRealJon, line 252: If we are going to update the rest of the the labels to use the accurate resource label/kind, we should do it here too:

```suggestion
      desc={model.plural}
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
   id: PropTypes.string,
 };
 
-const NsDropdown_ = (props) => {
-  const openshiftFlag = props.flags[FLAGS.OPENSHIFT];
-  if (flagPending(openshiftFlag)) {
-    return null;
+const useProjectOrNamespaceModel = () => {
+  const nsModel = modelFor('Namespace');
```
TheRealJon, line 186: Just import the static NamespaceModel directly. No need to retrieve it dynamically.

```suggestion
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+  const canCreateNamespace = useAccessReview({
+    group: nsModel.apiGroup,
+    resource: nsModel.plural,
```
TheRealJon, line 189: As stated, just import the static NamespaceModel, then use it here:

```suggestion
    group: NamespaceModel.apiGroup,
    resource: NamespaceModel.plural,
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +182,79 @@ ListDropdown.propTypes = {
...
+
+  if (flagPending(openshiftFlag) || flagPending(canCreateProject)) {
+    return [];
+  }
+
+  if (canCreateNamespace) {
+    return [modelFor('Namespace'), canCreateNamespace];
+  } else if (canCreateProject) {
+    return [modelFor('Project'), canCreateProject];
   }
```
TheRealJon, line 204: We always need to return `[model, canCreate]` from this function once we have the flags. The model is used to list the resources whether or not the user can create them. The `openshiftFlag` tells us whether or not we are running against an OpenShift cluster. If it's true, we need to use the OpenShift-specific `ProjectModel`, otherwise default to the vanilla K8s`NamespaceModel`. Also, import and use static models instead of retrieving them dynamically:

```suggestion
  return openshiftFlag 
    ? [ProjectModel, canCreateProject] 
    : [NamespaceModel, canCreateNamespace];
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -507,6 +507,7 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
       id="dropdown-selectbox"
       selectedKey={selectedTargetNamespace}
       onChange={setTargetNamespace}
```
TheRealJon, line 509: We need to handle this onChange prop in the NsDropdown component. We added our own onChange function to that component and no longer pass this down to the ListDropdown via the spread operator.

TheRealJon (commented):
Much improved,  a few other tweaks to make. In addition to my comments, we still need to address the selected key being updated correctly when creating a new project. As we discussed the `onChange` handler in ListDropdown is overriding the `selectedKey` prop that was passed in. My suggestion would be to make ListDropdown an "optionally controlled" component. Just add a new `isControlled` prop that defaults to false. In the onChange handler, only update the components internal state if the component is not controlled (i.e. the parent is not managing the state).

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
harishsurf: The `onChange()` handler in `NSDropdown` component now calls `props.onChange()`

frontend/public/components/utils/list-dropdown.jsx
harishsurf: handled this

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -85,8 +94,17 @@ class ListDropdown_ extends React.Component {
         .forEach((key) => {
           sortedList[key] = unsortedList[key];
         });
+
+      const keyChanged = this.props.selectedKey !== nextProps.selectedKey;
+      const keyKindChanged = this.props.selectedKey !== nextProps.selectedKeyKind;
```
TheRealJon, line 99:
```suggestion
      const keyKindChanged = this.props.selectedKeyKind !== nextProps.selectedKeyKind;
```

TheRealJon (changes requested):
Just a typo and a small adjustment to the `ListDropDown.UNSAFE_componentWIllRecieveProps` function.

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -175,22 +194,77 @@ ListDropdown.propTypes = {
...
+
+  // NamespaceModal is used when not on an openshift cluster
+  const model = canCreateNamespace || !openshiftFlag ? NamespaceModel : ProjectModel;
+  const canCreate = canCreateNamespace || (openshiftFlag && canCreateProject);
+  return [model, canCreate];
+};
+
+/** @type {React.FC<{dataFilter?: (ns: any) => boolean, desc?: string, selectedKey?: string, selectedKeyKind?: string, fixed?: boolean, placeholder?: string, onChange?: (selectedKey: string, event: React.Event) => void, id?: string, dataTest?: string}}>} */
+export const NsDropdown = (props) => {
+  const [selectedKey, setSelectedKey] = React.useState('');
```
TheRealJon, line 219: One more small change. This ensures the initial selection is correct:
```suggestion
  const [selectedKey, setSelectedKey] = React.useState(props.selectedKey);
```

frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -85,8 +94,17 @@ class ListDropdown_ extends React.Component {
         .forEach((key) => {
           sortedList[key] = unsortedList[key];
         });
+
+      const keyChanged = this.props.selectedKey !== nextProps.selectedKey;
+      const keyKindChanged = this.props.selectedKey !== nextProps.selectedKeyKind;
+      selectedKey =
+        keyChanged || keyKindChanged
+          ? getKey(nextProps.selectedKey, nextProps.selectedKeyKind)
+          : getKey(this.props.selectedKey, this.props.selectedKeyKind);
```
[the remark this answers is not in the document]
harishsurf: gotcha! this makes sense

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
TheRealJon: This also needs to be updated to prevent a React warning. The call signature of `NsDropdown`'s `props.onChange` function accepts three args. If you pass more than one arg to the `React.useState` setter callback, it will trigger a console warning.  
```suggestion
      onChange={(ns) => setTargetNamespace(ns)}
```

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -531,13 +532,14 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
         value={suggestedNamespace}
         checked={!useSuggestedNSForSingleInstallMode}
-        title="Pick an existing namespace"
+        title="Select a namespace"
       />
       {!useSuggestedNSForSingleInstallMode && (
         <NsDropdown
           id="dropdown-selectbox"
           selectedKey={selectedTargetNamespace}
           onChange={setTargetNamespace}
```
TheRealJon, line 541:
```suggestion
          onChange={(ns) => setTargetNamespace(ns)}
```

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/createNamespace.ts
harishsurf: Added `data-test` attribute as suggested

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    cy.log('test namespace creation from dropdown');
+    cy.visit(`/operatorhub/ns/${testName}`);
+    cy.byTestID(operatorSelector).click();
+    cy.byLegacyTestID('operator-install-btn').click({ force: true });
+
+    // configure operator install
+    cy.byTestID('Select a namespace-radio-input').check();
+    cy.byTestID('dropdown-selectbox')
+      .click()
+      .wait(2000);
```
dtaylor113, line 32: We should not be waiting, pls see: https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting

frontend/package.json
dtaylor113: @harishsurf, pls follow how `dev-console` has integrated with our CI tests.
https://github.com/openshift/console/pull/6776/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R30

dtaylor113 (changes requested):
Please address Cypress and CI issues.  -thanks

frontend/package.json
TheRealJon: Couldn't we just write a simple automation script that searches for a cypress test folder in each package, and runs `cypress open` in that dir if it exists?

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    cy.byLegacyTestID('dropdown-menu')
+      .contains('Create Namespace')
+      .click();
+
+    // verify namespace modal is opened
+    modal.shouldBeOpened();
+    cy.byTestID('input-name').type(nsName);
+    modal.submit();
+    modal.shouldBeClosed();
```
TheRealJon, line 41: Check to ensure that the dropdown selection updates to the newly created namespace before moving on to the next test step.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    cy.get('button')
+      .contains('Install')
+      .click();
+
+    // wait till the operator was successfully installed in the new namespace
+    cy.contains('Installed operator - ready for use', { timeout: 60000 });
+    cy.byTestID('view-installed-operators-btn').should(
+      'contain',
+      `View Installed Operators in namespace ${nsName}`,
+    );
```
TheRealJon, line 52: We really just need to make sure that the new ns was successfully created. No need to complete the full operator installation. We can just navigate to the project list page and make sure that the newly created project exists there, then delete it.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
harishsurf: removed it

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
harishsurf: added the assertion

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
harishsurf: removed waiting for operator to complete install.  Validating creation of new `ns` & `project` is already ensured by `cy.deleteProject(nsName)` on #L54 and #L15 respectively.

spadgett:
for QE / docs / CEE review

This is ready to test.

frontend/package.json
```diff
@@ -26,8 +26,12 @@
     "test-gui-openshift": "yarn run test-suite --suite crud --params.openshift true",
     "test-gui": "yarn run test-suite --suite all",
     "test-suite": "ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/protractor integration-tests/protractor.conf.ts",
-    "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
-    "test-cypress-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress": "yarn run test-cypress-common && yarn run test-cypress-olm",
```
dtaylor113, line 29: Pls see [dev-console cypress PR](https://github.com/openshift/console/pull/6776/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R32) changes.

Pls revert changes to `test-cypress`.

frontend/package.json
```diff
@@ -26,8 +26,12 @@
...
+    "test-cypress-headless": "yarn run test-cypress-common-headless && yarn run test-cypress-olm-headless",
+    "test-cypress-common": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
+    "test-cypress-common-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-olm": "cd packages/operator-lifecycle-manager/integration-tests-cypress && cypress open --config-file cypress-olm.json --env openshift=true",
```
dtaylor113, line 33: Pls see [dev-console cypress PR](https://github.com/openshift/console/pull/6776/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R32) changes.

We are using `test-cypress-console...` instead of `..common`, but great minds do think alike :-)

frontend/package.json
```diff
@@ -26,8 +26,12 @@
     "test-gui-openshift": "yarn run test-suite --suite crud --params.openshift true",
     "test-gui": "yarn run test-suite --suite all",
     "test-suite": "ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/protractor integration-tests/protractor.conf.ts",
-    "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
-    "test-cypress-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress": "yarn run test-cypress-common && yarn run test-cypress-olm",
+    "test-cypress-headless": "yarn run test-cypress-common-headless && yarn run test-cypress-olm-headless",
+    "test-cypress-common": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
+    "test-cypress-common-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
```
TheRealJon, line 32:
```suggestion
    "test-cypress-console": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
    "test-cypress-console-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
```

frontend/package.json
```diff
@@ -26,8 +26,12 @@
     "test-gui-openshift": "yarn run test-suite --suite crud --params.openshift true",
     "test-gui": "yarn run test-suite --suite all",
     "test-suite": "ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/protractor integration-tests/protractor.conf.ts",
-    "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
-    "test-cypress-headless": "cd packages/integration-tests-cypress && node --max-old-space-size=4096 ../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress": "yarn run test-cypress-common && yarn run test-cypress-olm",
+    "test-cypress-headless": "yarn run test-cypress-common-headless && yarn run test-cypress-olm-headless",
```
TheRealJon, line 30:
```suggestion
    "test-cypress-headless": "yarn test-cypress-console-headless && yarn test-cypress-olm-headless",
```

frontend/package.json
TheRealJon, line 29:
```suggestion
    "test-cypress": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
```

frontend/package.json
```diff
@@ -26,8 +26,12 @@
...
+    "test-cypress-common": "cd packages/integration-tests-cypress && cypress open --env openshift=true",
```
TheRealJon, line 31:
```suggestion
```

TheRealJon:
/hold

Adress yarn script changes.

frontend/package.json
dtaylor113: This might help clarify what we're after:  https://issues.redhat.com/browse/CONSOLE-2426

frontend/package.json
dtaylor113: Thanks, this might clarify what we're eventually going to do:  https://issues.redhat.com/browse/CONSOLE-2426

ahardin-rh:
:+1: from the docs side

yapei:
@schituku  Please help test it

schituku:
/test @harishsurf @spadgett could able to create new project as part of the operator install but the following are also mentioned in the design document 
- the "Namespace" should be "namespace" to be consistent through out the application
- the create project should have one line description and also help icon and a popup box per the design

schituku:
@spadgett thanks for the clarifications on slack.

QE /lgtm

yapei:
Based on above comment, adding correct label

spadgett:
> 👍 from the docs side

/label docs-approved

dtaylor113:
Hi, olm isn't being included in the [cypress-report.html](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/6765/pull-ci-openshift-console-master-e2e-gcp-console/1318584239955906560/artifacts/e2e-gcp-console/gui_test_screenshots/cypress-report.html) which only contains the `console` cypress test results.  This is confusing to me because olm [overwrites](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/6765/pull-ci-openshift-console-master-e2e-gcp-console/1318584239955906560/artifacts/e2e-gcp-console/container-logs/test.log) the `/frontend/gui_test_screenshots/cypress_report.json` last, so I would expect to only see "_Create namespace from install operators_" in the report, instead I'm seeing only the console cypress test results(?).

I belive you are going to need your own [reporter-config.json](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/reporter-config.json), pls use `"reportFilename": "cypress_report_olm",`.  Point to new file [here](https://github.com/openshift/console/pull/6765/files#diff-488790c57b6d5fe5f827544a05f1a72cd79b61cc8edf186471206de353fc577dR8). 

cc @makambalaji

dtaylor113:
Hi @harishsurf, I see you added a `reporter-config.json`, however I still don't see a `cypress_report_olm` [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/6765/pull-ci-openshift-console-master-e2e-gcp-console/1320808027669925888/artifacts/e2e-gcp-console/gui_test_screenshots/)

I believe your relative path is incorrect (/frontend/**_packages_**/gui_test_screenshots/):

```
 Running:  create-namespace.spec.ts                                                        (1 of 1)
 ...
  1 passing (24s)
[mochawesome] Report JSON saved to /go/src/github.com/openshift/console/frontend/packages/gui_test_screenshots/cypress_report_olm.json
```
...compared to `Auth test` in console (/frontend/gui_test_screenshots/):
```
Report JSON saved to /go/src/github.com/openshift/console/frontend/gui_test_screenshots/cypress_report.json
```

dtaylor113 (approved):
Good to see olm cypress tests being included in [cypress-report.html](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/6765/pull-ci-openshift-console-master-e2e-gcp-console/1321434059380887552/artifacts/e2e-gcp-console/gui_test_screenshots/cypress-report.html).  "Create namespace from install operators" last item in report.

TheRealJon (approved):
/lgtm
/approve

PR #6766, opened 2020-09-28
Bug 1878676: Add resource badges to application groups in topology list view

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4877

**Description**
Adds a resource badge to the application groups in the topology list view.
Change the name from 'unassigned' to 'no application group' in the application selectors and the list view heading.
Do not show the 'no application group' heading when there are no application groups.

**Screenshots**
image

image

image

image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

/cc @bgliwa01 @beaumorley @serenamarie125

### DIFF
--- a/frontend/packages/dev-console/integration-tests/features/Knative/broker-ODC-4004.feature
+++ b/frontend/packages/dev-console/integration-tests/features/Knative/broker-ODC-4004.feature
@@ -2,7 +2,7 @@ Feature: Knative Eventing Broker Support
     User should be able to experience the Knative Eventing Broker and associated features
 
 
-Background: 
+Background:
     Given user has installed the OpenShift Serverless Operator
     And user has created knative serving
     And user has create knative eventing
@@ -21,7 +21,7 @@ Scenario: Create default Broker
 @regression
 Scenario: Sink event source to Broker
     Given user is at Developer Perspective
-    And user has created the Broker 
+    And user has created the Broker
     When user goes to +Add page
     And user clicks on the Event Source card
     And user selects Ping Source card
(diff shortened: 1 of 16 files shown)


bgliwa01 (approved):
lgtm

debsmita1:
tried this out locally
/lgtm

andrewballantyne (approved):
/lgtm

spadgett:
Everywhere else we use the badges, it's for a Kubernetes resource. Is it confusing to add a badge to something that's not a resource? (I assume here that the application groups are controlled by labels on the resources.)

andrewballantyne:
> Everywhere else we use the badges, it's for a Kubernetes resource. Is it confusing to add a badge to something that's not a resource? (I assume here that the application groups are controlled by labels on the resources.)

We already use it on the Topology screen, this is just carrying it over to the list view (which is also the workloads view in Admin => Projects => Workloads)

spadgett:
I guess I have the same concern about that context :) But it looks like that is something we've already added in a previous release.

andrewballantyne:
@spadgett We absolutely welcome the feedback :) I was merely stating it wasn't "new" for this situation... it was a mirror of content we already have. This was decided back in 4.3(?) to add it to the Topology graph view.

cc @christianvogt for awareness.

christianvogt:
> I guess I have the same concern about that context :) But it looks like that is something we've already added in a previous release.

Feedback goes to @serenamarie125  :)

PR #6767, opened 2020-09-28
Bug 1881953: Generate new namespace with empty node selector for node…

rhamilto:
… debug pods

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1881953


spadgett (approved):
nice work 👍 

/lgtm

spadgett:
I think we can target 4.6 here, too.

/hold cancel
/bugzilla refresh

PR #6768, opened 2020-09-28
Bug 1877891: Aggregated Helm index contains only entries from healthy repos

pedjak:
A repo does not contribute to the aggregated index in the following cases:
* failure at reading Helm repo configuration in the cluster
* networking issues accessing the repo index
* repo returns response that is not Helm index
* repo returns response with http status code different from 200


go.mod
```diff
@@ -28,6 +28,9 @@ require (
 )
 
 replace (
+	// ww/goautoneg repo does not exist on butbucket anymore, hence we
+	// need to use a mirror. It can be removed when we update to newer libary-go
+	bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822
```
spadgett, line 33: @jhadvig FYI

I think we have another PR to bump library-go

spadgett (approved):
/lgtm

PR #6769, opened 2020-09-28
Bug 1883342: [Release 4.5] Convert Pipeline StartedBy Label to Annotation

andrewballantyne:
Manual cherry-pick of https://github.com/openshift/console/pull/6639 and then https://github.com/openshift/console/pull/6744 to fix an edge case issue with Triggers.

There is an inconsistency in implementation as the lines around this changed because of the work done in the `master` branch for https://github.com/openshift/console/pull/6615 so it is a prerequisite for this PR and will cause merge conflicts which I'll have to rebase.

/hold


abhinandan13jan:
verified locally
 /lgtm

abhinandan13jan:
/lgtm
verified locally

andrewballantyne:
/hold cancel

This is now sitting on top of the full code that is in master. Should be good for the next available z-stream.

PR #6771, opened 2020-09-28
Bug 1847074: PF fix for filter bar layout issues at some screen widths on search page

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1847074

After:

Screen Shot 2020-09-28 at 5 23 57 PM
Screen Shot 2020-09-28 at 5 23 45 PM


frontend/packages/console-shared/src/components/alerts/AlertSeverityIcon.tsx
```diff
@@ -19,7 +19,7 @@ const AlertSeverityIcon: React.FC<AlertSeverityIconProps> = ({
             fontSize,
             color: 'var(--pf-global--danger-color--100)',
           }}
-          alt="Monitoring Alert"
+          aria-label="Monitoring Alert"
```
spadgett, line 22: I'm not sure this is right. Are these SVGs? I believe they should be `alt`.

cc @jessiehuff

spadgett (commented):
Thanks @jcaianirh. I think we should hold this one for 4.7 due to the PF bump.

frontend/packages/console-shared/src/components/alerts/AlertSeverityIcon.tsx
jessiehuff: Truthfully, svgs are tough, and I've never seen a consensus on that debate between documentation and experts. One of the [best comparisons](https://www.deque.com/blog/creating-accessible-svgs/) I've seen is from deque (who created Axe) which essentially ends up stating that they found that one solution was not necessarily superior over the other. They cover all combinations of alt, roles, title/desc, and aria, and conclude: 

"On the flip side, there was no clear frontrunner for “best in show” for the different patterns when relaying basic alternative content. This is actually good news! Since many of the patterns were able to render the alt content, that equates to more pattern choices for designers and developers. For SVGs needing more descriptive alternative content, pattern 11 `<svg> + role="img" + <title> + <desc> + aria-labelledby="[ID]"` was the most reliable choice for the different browser and screen readers that were tested."

I've often still recommended alt because it's not just for screen readers, it's frequently used for low bandwidth users as well (when an image can't load, it will show the alt instead). I have read in some articles that svgs have limited support with alt or that it needs `role="img"` for it to be valid, but it looks like there's a lot of different ways for us to make this accessible. I think if it sounds good in testing, I'm ok with whatever solution we choose.

Some other notable documentation/articles that are interesting to include is this [comparison](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html) by Scott O'Hara, the more old school [recommendations](https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/) by Leonie Watson (both O'Hara and Watson are considered a11y experts), and the [Appendix H ](https://www.w3.org/TR/SVG11/access.html)on svg accessibility by W3.

spadgett: Thanks @jessiehuff. @jcaianirh I'd stick with the existing `alt` unless there is a specific reason we need to change. Does this cause problems with the new Patternfly version?

jcaianirh: @spadgett pf has been updated to support alt tags again. I will put this on hold and update the pr.

frontend/packages/container-security/src/components/summary.tsx
```diff
@@ -74,7 +74,7 @@ export const SecurityBreakdownPopup: React.FC<SecurityBreakdownPopupProps> = ({
                         <div>
                           <ExclamationTriangleIcon
                             color={priority.color.value}
-                            alt={priority.title}
+                            aria-label={priority.title}
```
spadgett, line 77: title?

frontend/public/components/custom-resource-definition.tsx
```diff
@@ -115,11 +115,11 @@ const namespaced = (crd: CustomResourceDefinitionKind) => crd.spec.scope === 'Na
 const Established: React.FC<{ crd: CustomResourceDefinitionKind }> = ({ crd }) => {
   return crd.status && isEstablished(crd.status.conditions) ? (
     <span>
-      <GreenCheckCircleIcon alt="true" />
+      <GreenCheckCircleIcon aria-label="true" />
```
spadgett, line 118: title?

frontend/public/components/custom-resource-definition.tsx
```diff
@@ -115,11 +115,11 @@ const namespaced = (crd: CustomResourceDefinitionKind) => crd.spec.scope === 'Na
...
     </span>
   ) : (
     <span>
-      <BanIcon alt="false" />
+      <BanIcon aria-label="false" />
```
spadgett, line 122: title?

spadgett (approved):
/lgtm

PR #6774, opened 2020-09-29
Bug 1881458: Prefer templates with data volumes

yaacov:


yaacov:
Closing:
After off-line discussion with @omeryahud we decided to remove:
https://github.com/kubevirt/common-templates/blob/master/templates/win2k12r2-deprecated.tpl.yaml#L28-L31

This will prevent choosing deprecated templates as options in UI drop-downs if they don't have OS name.

cc:// @jelkosz

PR #6778, opened 2020-09-29
Bug 1885343: Fallback to GQL HTTP on iOS devices

rawagner:
iOS does not allow connecting to wss with self signed certificate. There's also no way to know why the connection wasnt successful (given by WebSocket specification). 

To overcome the issue I've added code specific for iOS(iPhone/iPad):
 - if graphql ws connection fails 5 times in a row, fallback to http connection
 - delay console App rendering until we know which type of connection we can use (show Loading in the meantime)

@spadgett WDYT ?

### DIFF
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -14,7 +14,7 @@ import { getBrandingDetails, Masthead } from './masthead';
 import { ConsoleNotifier } from './console-notifier';
 import { ConnectedNotificationDrawer } from './notification-drawer';
 import { Navigation } from './nav';
-import { history, AsyncComponent } from './utils';
+import { history, AsyncComponent, LoadingBox } from './utils';
 import * as UIActions from '../actions/ui';
 import { fetchSwagger, getCachedResources } from '../module/k8s';
 import { receivedResources, watchAPIServices } from '../actions/k8s';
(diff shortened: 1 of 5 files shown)


frontend/public/components/app.jsx
```diff
@@ -192,80 +193,94 @@ class App_ extends React.PureComponent {
...
+    window.windowError = promiseRejectionEvent;
+  };
+
+  // Used by GUI tests to check for unhandled exceptions
+  window.windowError = false;
+  window.onerror = window.onunhandledrejection = (e) => {
+    // eslint-disable-next-line no-console
+    console.error('Uncaught error', e);
+    window.windowError = e || true;
+  };
```
spadgett, line 239: Looks like a mistake rebasing here. You can remove these lines.

frontend/public/graphql/client.ts
```diff
@@ -1,13 +1,47 @@
...
+  setReady() {
+    this.ready = true;
+    if (!this.wasCalled && this.callback) {
+      this.wasCalled = true;
+      this.callback();
+    }
+  }
+
+  onReady(cb: VoidFunction) {
+    if (!isIOS) {
```
spadgett, line 31: Is there any performance penalty if we skip the iOS check and always fallback on WebSocket errors?

spadgett (commented):
Thanks @rawagner !

frontend/public/graphql/client.ts
rawagner: There will be some difference but pretty small. And we will also gain the fallback for other cases.
I've removed the check, makes the code simpler and more secure which I think is worth the tiny perf downgrade.

frontend/public/components/app.jsx
rawagner: Sorry, should be fixed now.

spadgett (approved):
/lgtm

PR #6779, opened 2020-09-29
Bug 1883440: The default yaml is using hard code password: fedora

glekner:


frontend/packages/kubevirt-plugin/src/k8s/requests/vm/create/default-template.ts
```diff
@@ -55,6 +61,27 @@ export const resolveDefaultVMTemplate = (params: DefaultVMLikeEntityParams): Tem
         .asResource(),
     });
 
+  const cloudInitVolume = vm.getCloudInitVolume();
+
+  if (cloudInitVolume) {
+    const cloudInitHelper = new CloudInitDataHelper(cloudInitVolume.cloudInitNoCloud);
+    if (cloudInitHelper.hasKey(CloudInitDataFormKeys.PASSWORD)) {
+      cloudInitHelper.set(CloudInitDataFormKeys.PASSWORD, generateCloudInitPassword());
+      vm.removeStorage(CLOUDINIT_DISK).prependStorage({
```
suomiy, line 70: we are using the original cloud init storage with all the attributes. We don't have to replace it and we can just modify the password.

frontend/packages/kubevirt-plugin/src/k8s/requests/vm/create/default-vm.ts
```diff
@@ -26,6 +37,26 @@ export const resolveDefaultVM = async (params: DefaultVMLikeEntityParams): Promi
   const processedTemplate = await k8sCreate(ProcessedTemplatesModel, template.asResource(), null); // temporary
 
   const vm = new VMWrapper(selectVM(processedTemplate));
+  const cloudInitVolume = vm.getCloudInitVolume();
```
suomiy, line 40: this function already called resolveDefaultVMTemplate. So there is no need to repeat the logic here.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/vm-wrapper.ts
```diff
@@ -202,6 +211,26 @@ export class VMWrapper extends K8sResourceWrapper<VMKind, VMWrapper> implements
     return this;
   };
 
+  generateCloudInitPassword = () => {
```
suomiy, line 214: IMO, it is not a good idea to overload this wrapper with too over specific methods. 
We should not let this class to become huge and thus unmaintainable.

Can we rather?
1. keep the logic outside and call it only in default-template.ts
2. or move the logic into the VolumeWrapper where it makes more sense from responsibility standpoint

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/k8s/requests/vm/create/default-template.ts
```diff
@@ -55,6 +61,18 @@ export const resolveDefaultVMTemplate = (params: DefaultVMLikeEntityParams): Tem
         .asResource(),
     });
 
+  const cloudInitVolume = vm.getCloudInitVolume();
+  const cloudInitHelper = new CloudInitDataHelper(cloudInitVolume?.cloudInitNoCloud);
+  if (!cloudInitHelper.isEmpty() && cloudInitHelper.hasKey(CloudInitDataFormKeys.PASSWORD)) {
+    cloudInitHelper.set(CloudInitDataFormKeys.PASSWORD, generateCloudInitPassword());
+    vm.updateVolume(
+      new VolumeWrapper()
```
suomiy, line 69: we should use the original volume and not create a new one

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/k8s/requests/vm/create/default-template.ts
glekner: fixed @suomiy

PR #6782, opened 2020-09-29
Bug 1883566: adds check for kind before accessing

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4344

**Analysis / Root cause**: 
this scenario occurs if cluster is low on resources for knative service UI shows underlying Deployment and thus now the sidebar is trying to open up a Knative sidebar when it needs a Deployment sidebar. It fails to match up the model used for this particular workload and hits a NPE failing to get it with error `can not read property kind of undefined`.

**Solution Description**: 
adds check for kind before accessing

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/components/overview/KnativeResourceOverviewPage.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/KnativeResourceOverviewPage.tsx
@@ -55,13 +55,15 @@ export const KnativeResourceOverviewPage: React.ComponentType<KnativeResourceOve
   if (kindsInFlight) {
     return !knativeModels ? null : <LoadingBox />;
   }
-  const apiInfo = groupVersionFor(item.obj.apiVersion);
-  const resourceModel = knativeModels.find(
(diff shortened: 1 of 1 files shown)


andrewballantyne:
Code looks fine for safety checks against missing items. I haven't checked it out and chased types... but I have a question @invincibleJai... Do you know of a scenario that we can do to reproduce this? If I recall correctly, we never figured out why this happens.

invincibleJai:
> Code looks fine for safety checks against missing items. I haven't checked it out and chased types... but I have a question @invincibleJai... Do you know of a scenario that we can do to reproduce this? If I recall correctly, we never figured out why this happens.

I couldn't reproduce it as this happens if `webhooks` aren't reachable as cluster is low on resources thus for firehose query we won't get ksvc resource but we do get underlying deployment.

andrewballantyne (approved):

PR #6786, opened 2020-09-29
Bug 1883918: Proposed fix for duplicate UpdateAvailable notifications in the notification drawer

jcaianirh:
Eliminate duplicate notification entries in the notification drawer to update a cluster. There is currently an ota alert named UpdateAvailalble and also a recommendation created from data in clusterversion. They provide duplicate info, and the solution proposed is to filter out the ota alert. This pr addresses that solution.

Before:
Screen Shot 2020-09-29 at 10 56 07 PM

After:
Screen Shot 2020-09-29 at 10 56 20 PM


spadgett (approved):
/lgtm

dtaylor113 (approved):
/lgtm

PR #6787, opened 2020-09-30
Bug 1883764: fix issue in topology when channel with different kind have same name

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4931

**Analysis / Root cause**: 
if user created a channel of kind `IMC` name `my-channel` and sink to ksvc and then create another channel for kind other then IMC say `NATS` with name `my-channel`. Can see below issues

- user will see sidebar with subscriber KSVC for NATSS channel as well along with connetor
- In sink for eventSources user see only one channel

**Solution Description**: 
- Added check for kind as well as channel with different name can have same kind
- Key for resource dropdown was channel name , have made use of custom key `kind-channelname`

- Sidebar and topology

image

**Test setup:**
1. Install Serverless operator

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified locally, works fine
/lgtm

sahil143 (approved):
/lgtm

tested locally, works as expected.

PR #6788, opened 2020-09-30
Bug 1883766: Adjust tests for UI changes

gouyang:


frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
```diff
@@ -26,15 +26,15 @@ import {
...
     source: 'https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img',
   },
   [ProvisionSource.CONTAINER]: {
-    method: ProvisionSource.CONTAINER,
+    method: ProvisionSourceItem.CONTAINER,
     source: 'kubevirt/fedora-cloud-registry-disk-demo:latest',
   },
-  [ProvisionSource.PXE]: { method: ProvisionSource.PXE },
-  [ProvisionSource.DISK]: { method: ProvisionSource.DISK },
+  [ProvisionSource.PXE]: { method: ProvisionSourceItem.PXE },
```
suomiy, line 36: what benefit do we get from splitting ProvisionSource into 2 constants? This will result in worse maintainability.

Can we either

1. rename the original to what we need: `URL (adds disk)`, etc.
2. refactor the ProvisionSource to become ObjectEnum so it has all the required information in one place (name, label and source)

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
```diff
@@ -63,7 +63,7 @@ deepFreeze(multusNAD);
 
 export const dataVolumeManifest = ({ name, namespace, sourceURL, accessMode, volumeMode }) => {
   return {
-    apiVersion: 'cdi.kubevirt.io/v1alpha1',
+    apiVersion: 'cdi.kubevirt.io/v1beta1',
```
suomiy, line 66: I have already done this here for all places (also tests) and we decide to wait for 4.7. https://github.com/openshift/console/pull/6739

Maybe it would be better to merge it atomically? I know there are some bugs because of that atm.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
```diff
@@ -97,13 +93,21 @@ export class Wizard {
...
   async selectWorkloadProfile(workloadProfile: string) {
-    await selectOptionByText(view.workloadProfileSelect, workloadProfile);
+    await selectItemFromDropdown(view.workloadProfileSelect, view.dropDownItem(workloadProfile));
   }
 
   async selectProvisionSource(provisionOptions) {
-    await selectOptionByText(view.provisionSourceSelect, provisionOptions.method);
+    await selectItemFromDropdown(
+      view.provisionSourceSelect,
+      view.bootSourceDropDownItem(provisionOptions.method),
```
suomiy, line 102: can we call it provisionSourceDropDownItem for consistency reasons? I know we renamed it in the UI. but we should be consistent here IMO.

Or we should rename all the occurrences here

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
```diff
@@ -97,13 +93,21 @@ export class Wizard {
...
+    );
     if (Object.prototype.hasOwnProperty.call(provisionOptions, 'source')) {
-      await fillInput(view.provisionSourceInputs[provisionOptions.method], provisionOptions.source);
+      if (provisionOptions.method === ProvisionSourceItem.URL) {
```
suomiy, line 105: is this necessary? can we keep the original form?

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
```diff
@@ -208,27 +212,20 @@ export class Wizard {
     if (description) {
       await this.fillDescription(description);
     }
-    if ((await browser.getCurrentUrl()).match(/\?template=.+$/)) {
-      // We are creating a VM from template via its action button
-      // ProvisionSource, OS and workload are prefilled and disabled - ignoring them
-    } else if (template) {
-      await this.selectTemplate(template);
+    if (os) {
```
suomiy, line 215: please can you show me, where is the logic for the template now? We are still offering this option.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
```diff
@@ -239,11 +236,11 @@ export class Wizard {
   }
 
   async processNetworkStep(data: VMBuilderData) {
-    const { networks, provisionSource, template } = data;
+    const { networks, provisionSource } = data;
     for (const resource of networks) {
       await this.addNIC(resource);
     }
-    if (provisionSource?.method === ProvisionSource.PXE && template === undefined) {
+    if (provisionSource?.method === ProvisionSourceItem.PXE) {
```
suomiy, line 243: plus here

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
```diff
@@ -15,10 +15,7 @@ describe('Test adding/removing discs/nics to/from a VM template', () => {
     .setProvisionSource(provisionSources.Container)
     .build();
 
-  const vm = new VMBuilder(getBasicVMBuilder())
-    .setTemplate(vmt.name)
-    .setName('from-template')
-    .build();
+  const vm = new VMBuilder(getBasicVMBuilder()).setName('from-template').build();
```
suomiy, line 18: there is a new flow for creating vm's from templates: through `/k8s/ns/default/virtualization/templates` or `k8s/ns/default/vmtemplates/vm-template-example`. 

We should test the new way of creating vms from templates in all our scenarios

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.wizard.scenario.ts
```diff
@@ -55,7 +55,6 @@ describe('Create VM from Template using wizard', () => {
             .setNamespace(testName)
             .setDescription(`VM from template ${vmt.name}`)
             .setFlavor(flavorConfigs.Tiny)
```
suomiy, line 57: we will still need to know the name of the template to create the VM from

suomiy (changes requested):

frontend/packages/kubevirt-plugin/integration-tests/tests/utils/constants/vm.ts
```diff
@@ -71,16 +70,16 @@ Object.freeze(networkTabCol);
 
 // Storage
 export enum DISK_SOURCE {
```
irosenzw, line 72: I don't think its a good idea maintaining these `enum` s here...
Every time something changes in `src/` and it isn't updated here the tests break.

I propose another solution - to have these `enum` s declared in `src/` and imported from there:
I created a small PR for the disks as an example of what I think should be done: https://github.com/openshift/console/commit/7d8179cc5fec93e2ad5396ac80ae1949e25024b5

It might be out of scope for this PR but I think it is worth doing it soon.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
irosenzw: like this: 
```
await selectItemFromDropdown(view.diskSource, view.diskDropDownItem(disk.source || DISK_SOURCE.Blank))
```

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
ibragins: I used the code listed here:
```
async selectOperatingSystem(operatingSystem: string) {
    await selectItemFromDropdown(view.operatingSystemSelect, view.dropDownItem(operatingSystem));
  }
```
and it couldn't select OS when running VMWare wizard

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
ibragins: I used the code listed here:
```
async selectOperatingSystem(operatingSystem: string) {
    await selectItemFromDropdown(view.operatingSystemSelect, view.dropDownItem(operatingSystem));
  }
```
and it couldn't select OS when running VMWare wizard

frontend/packages/kubevirt-plugin/integration-tests/tests/utils/constants/vm.ts
suomiy: Please let's not mix src and tests code right now. This needs a proper discussion and evaluation of pros/cons.

The main issue with this is that the tests should not depend and be bound on the implementation. This can hurt when refactoring things and eg adding i18n.

If you bring bugs into the code it can bring bugs to the tests and it might be hard to discover them. For example we had a discussion about bringing VMWrappers into the tests, but decided against it for this exact reason.

The main hardship is that we are not commiting code with the tests at the same. Once we have the CI, such renaming issue would get picked up immediately and such code would not get merged in the first place. So there is less pressure for sharing the constants IMO.

suomiy: I know it would be easier for now to do it like you suggest. But will get harder once you start introducing i18n.
Also, it would start a precedence with sharing the code, which might be confusing for newcomers.

Anyway, we can discuss this in a meeting next week if you want.

frontend/packages/kubevirt-plugin/integration-tests/tests/dialogs/diskDialog.ts
```diff
@@ -97,7 +102,12 @@ export class DiskDialog {
 
   async create(disk: Disk) {
     await waitForNoLoaders();
-    await selectOptionByText(view.diskSource, disk.source || DISK_SOURCE.Blank);
+    if (disk.source) {
```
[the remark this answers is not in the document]
gouyang: ok, will try this way.

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
gouyang: I added `ProvisionSourceItem` because string like `URL (adds disk)` cannot be used as the key.
Will look into `ObjectEnum`.

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
gouyang: If we wait for 4.7, then the 4.6 and 4.6.z tests cannot get PASS. My opinion is let's fixing tests firstly here.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: > can we call it provisionSourceDropDownItem for consistency reasons? I know we renamed it in the UI. but we should be consistent here IMO.
ok.
> and it couldn't select OS when running VMWare wizard
I guess the style of OS list in VMWare wizard is different from vm Wizard, if so, I think the best way is making the style consistent.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: ok. will do

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: The wizard field is removed by [1871103](https://bugzilla.redhat.com/show_bug.cgi?id=1871103), so remove it here as well.

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
gouyang: we have the test `ID(CNV-4290) Creates VM using VM Template create virtual machine link` in vmtemplate.wizard.scenario.ts

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.wizard.scenario.ts
gouyang: Will check this late.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
```diff
@@ -199,7 +203,7 @@ export class Wizard {
   }
 
   async processGeneralStep(data: VMBuilderData) {
-    const { name, description, template, provisionSource, os, flavor, workload } = data;
```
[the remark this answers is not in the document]
gouyang: because `template` is not a wizard field any more.
Will check `vm.template` scenario.

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
suomiy: Why not? Is there a specific reason why it cannot be used as a key?

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
suomiy: ok, so please replace all the occurences in the tests

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
suomiy: It is using the same component in the code. @ibragins can you check if it renders different html/styles for you?

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
suomiy: yes, but this test tests resources (disks/nics) and should also do that with template flow in mind

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
suomiy: the field was removed, but the wizard is still used for creating VMs from the template. So it should still not be possible to select os/workload/etc for these VMs. This applies for other logic as well. Only the selection should be removed.

Still we could automate this and add the template field tot the build and it would just go to the template page and select the create VM action

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
gouyang: If using `URL (adds disk)` , it throws errors very quickly:
```
[10:33:24] E/launcher - Error: TypeError: Cannot read property 'source' of undefined
    at Object.exports.getTestDataVolume (/home/gouyang/console/frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts:102:37)
```

frontend/packages/kubevirt-plugin/integration-tests/tests/dialogs/diskDialog.ts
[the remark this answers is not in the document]
gouyang: done

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
gouyang: @suomiy Could you give me an example how to use ObjectEnum?

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
gouyang: done

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: done

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: The reason is the value of `provisionOptions.method` for `URL` is `URL (adds disk)`, provisionSourceInputs only has `URL` and `Container`, so do it this way.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: Removed field template works well for create vm from template or not from template based on the new changes and the tests are PASS for vm.wizard and vmtemplate.wizard:
- template is removed in vm wizard
- the new link/button to create vm from template

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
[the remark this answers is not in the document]
gouyang: I checked vmtemplate test scenario, it works well with this change.

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.wizard.scenario.ts
gouyang: No need to use template here based on the actual test results.

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
gouyang: No need to use template here based on the actual test results.

frontend/packages/kubevirt-plugin/integration-tests/tests/models/wizard.ts
gouyang: No need to use template here based on the actual test results.

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
suomiy: can you please specify, why is there no need to test templates?

suomiy:
@gouyang btw @irosenzw started working on this patch as well in https://github.com/openshift/console/pull/6850 . Can you guys please sync, so you don't do the same work twice?

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
gouyang: It does not skip/ignore template test, I mean no need to use `template` field.
```
} else if (template) {
      await this.selectTemplate(template);
    } else {
```

`await this.selectTemplate(template);` will never be executed as we don't have this field. If remove this line and leave this block empty does not make sense and is not accepted by eslint.

To create vm from template should use:
```
      await vmt.createVMFromRowLink();
      await wizard.processWizard(vm.getData());
```
It does not need to have `template` at all.

gouyang:
> @gouyang btw @irosenzw started working on this patch as well in #6850 . Can you guys please sync, so you don't do the same work twice?

Thanks for the reminder, I will review #6850 and close this.

frontend/packages/kubevirt-plugin/integration-tests/tests/vmtemplate.resources.scenario.ts
suomiy: I get that the template option is not in the wizard, but it is still possible to create a VM from the template

gouyang: We have tests to create vm from template and it does not require `template` field to create vm from template.

suomiy: you need to know the name of the template to select it. The implementation (wizard field or clicking on the template row) is less important. 
But having one config in one place is always better IMO.

PR #6792, opened 2020-09-30
Bug 1883591: Support ServiceBinding breaking api changes

divyanshiGupta:
**Jira:** https://issues.redhat.com/browse/ODC-4928

**Analysis / Root cause:**
Service Binding Operator (starting with v0.2.0) has changed API (CRD) https://github.com/redhat-developer/service-binding-operator/pull/608 that is not supported in OCP 4.6 and hence service-binding is not working.

**Solution Description:**
Refactor code based on the latest API changes. Also fix failing unit tests accordingly

**Gifs:**
Create service-binding using:

1. Binding connector
Peek 2020-09-30 16-32

2. Import YAML
Peek 2020-09-30 16-35

**PR Test setup:**
Install community version of Service Binding from OperatorHub using the beta channel.

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/__tests__/service-binding-test-data.ts
+++ b/frontend/packages/dev-console/src/components/topology/__tests__/service-binding-test-data.ts
@@ -1,27 +1,30 @@
-import { DeploymentKind, K8sResourceKind } from '@console/internal/module/k8s';
+import { apiVersionForModel, DeploymentKind, K8sResourceKind } from '@console/internal/module/k8s';
+import { ServiceBindingModel } from '../../../models';
 import { TopologyDataResources } from '../topology-types';
 
 export const serviceBindingRequest: K8sResourceKind = {
   data: {
-    apiVersion: 'apps.openshift.io/v1alpha1',
-    kind: 'ServiceBindingRequest',
+    apiVersion: apiVersionForModel(ServiceBindingModel),
+    kind: ServiceBindingModel.kind,
     metadata: {
       name: 'analytics-deployment-D-wit-deployment-D',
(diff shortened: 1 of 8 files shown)


christianvogt:
@Divyansh42 can we add support for both SBR and SB CRDs?
Keep old and new models.
Create a new feature flag for the SB.
Fetch both SBR and SB for topology and visualize them in the same way.
On creation, if the SB feature flag is set, create the SB, otherwise create the SBR.

How much additional work is needed to do the above?

divyanshiGupta:
> @Divyansh42 can we add support for both SBR and SB CRDs?
> Keep old and new models.
> Create a new feature flag for the SB.
> Fetch both SBR and SB for topology and visualize them in the same way.
> On creation, if the SB feature flag is set, create the SB, otherwise create the SBR.
> 
> How much additional work is needed to do the above?

Shouldnt it be that the user can only install the operator from particular channels on particular OCP versions. For example version < = 4.5 operator will be installed using alpha channel and for version > 4.5 operator will be installed using beta channel. Here https://issues.redhat.com/browse/ODC-4928 there are some comments regarding having a single channel so I assume for particular versions we will have respective single channels. 

And say if in future we have a new CRD so then we will support three CRDs for a resource?

christianvogt:
After discussing this further we will go with updating support for the new SB and dropping SBR in 4.6.

christianvogt:
I've tested the the changes and creation and visualization of the SB work as expected.
/lgtm
/approve

christianvogt:
/cherry-pick release-4.5

New SBO has been promoted to be the default for 4.5 as well.

PR #6799, opened 2020-09-30
Treat console-extensions.json file as JSON with Comments (jsonc)

vojtechszocs:
Plugin developers can now use comments in their `console-extensions.json` file.

Some other improvements included in this PR:

- Updated typings in extension consuming APIs and `PluginStore` to work with `LoadedExtension` (runtime extension interface exposing additional metadata) where appropriate.

- In `dynamic-demo-plugin`, use `link` instead of `file` protocol when referencing dynamic plugin SDK package. This means we can modify the referenced package and re-build demo plugin without having to re-run `yarn` again.

cc @spadgett @christianvogt

### DIFF
--- a/frontend/.vscode/settings.json
+++ b/frontend/.vscode/settings.json
@@ -48,6 +48,9 @@
     "**/.cache-loader": true,
     "**/public/dist": true
   },
+  "files.associations": {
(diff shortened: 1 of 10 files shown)


christianvogt:
last commit looks good

PR #6800, opened 2020-09-30
Bug 1881844: Fix OperatorHub item badges and alerts

TheRealJon:
Drive operator hub item Marketplace and Community badges and alerts with catalog source name instead
of `opsrc-provider` label, which was deprecated in 4.6.


spadgett (approved):
/lgtm

👍

PR #6801, opened 2020-10-01
Bug 1884664: Start the install status page as soon as subscription exist

zherman0:
Currently, the install status page does not show until either a CSV or an install plan has been created; instead we show an "Installing..." message.  This was to ensure we could get the needed info and CSV logo from the csv object or indirectly from the installPlan object.  However, the information is also available on the package manifest (which we now use) as soon as we have the subscription. We can therefore show the install page immediately which look more finished.
(I do realize the description here is longer than the code change)

no-loading


zherman0:
@spadgett @tlwu2013 @itsptk - Take a look, although it is tough to see without using it.  I immediately show the install status page as soon as we have the subscription instead of waiting for IP/CSV.  Let me know if I should open a bug to get this into 4.6.

spadgett (approved):
/lgtm

PR #6804, opened 2020-10-01
Bug 1871742: It is not possible to delete a PVC created using a data upload

glekner:


jelkosz:
It does not seem to work on a pvc which is in error state
/hold

jelkosz:
/hold cancel
sorry, apparently I should drink more coffee, it does work. As a token of apology, accept this 
/joke

PR #6807, opened 2020-10-01
Re-enable Snapshots tab

glekner:


frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
```diff
@@ -41,6 +49,8 @@ export const breadcrumbsForVMPage = (match: any) => () => [
 
 export const VirtualMachinesDetailsPage: React.FC<VirtualMachinesDetailsPageProps> = (props) => {
   const { name, ns: namespace } = props.match.params;
+  const [snapData, snapLoaded, snapErr] = useK8sGet(VirtualMachineSnapshotModel);
+  const isSnapshotSupported = snapData && snapLoaded && !snapErr;
```
suomiy, line 53: can you please look at the original description of this bug https://bugzilla.redhat.com/show_bug.cgi?id=1881125 ? And fix the loading when reintroducing the snapshots again?

suomiy (changes requested):

glekner:
@suomiy we know look for snapshots only in the relevant namespace. Is this solution sufficient?

suomiy:
@glekner we should check if the CRD is present

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
```diff
@@ -22,10 +23,21 @@ import { VMConsoleFirehose } from './vm-console';
...
+  TEMPLATE_TYPE_LABEL,
+  TEMPLATE_TYPE_VM,
+  VM_DETAIL_ENVIRONMENT,
+  VM_DETAIL_SNAPSHOTS,
+} from '../../constants/vm';
 import { VMEnvironmentFirehose } from './vm-environment/vm-environment-page';
 import { VMNics } from '../vm-nics';
 import { PendingChangesWarningFirehose } from './pending-changes-warning';
+import { VMSnapshotsPage } from '../vm-snapshots/vm-snapshots';
+import {
```
pcbailey, line 36: nitpick: This should be above the kubevirt package imports. Of course, we can catch it in a follow-up if you don't have any other changes to make.

glekner:
added `useModelLoaded` https://github.com/openshift/console/pull/6807/files#diff-1c59bec23f195454582776c62b6085baR64
@suomiy @rawagner

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
```diff
@@ -41,6 +49,7 @@ export const breadcrumbsForVMPage = (match: any) => () => [
 
 export const VirtualMachinesDetailsPage: React.FC<VirtualMachinesDetailsPageProps> = (props) => {
   const { name, ns: namespace } = props.match.params;
+  const isSnapshotsAvailable = useModelLoaded(VirtualMachineSnapshotModel.kind);
```
suomiy, line 52: shouldn't `referenceForModel` be used here?

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
```diff
@@ -95,6 +102,7 @@ export const VirtualMachinesDetailsPage: React.FC<VirtualMachinesDetailsPageProp
     consolePage,
     nicsPage,
     disksPage,
+    ...(isSnapshotsAvailable ? [snapshotsPage] : []),
```
suomiy, line 105: are we expecting more pages? IMO simple if would be more readable.

frontend/public/components/utils/k8s-watch-hook.ts
```diff
@@ -61,6 +61,9 @@ const useModelsLoaded = (): boolean => {
   return ref.current;
 };
 
+export const useModelLoaded = (model: string): boolean =>
+  !!useSelector<RootState, K8sKind>(({ k8s }) => k8s.getIn(['RESOURCES', 'models', model]));
```
suomiy, line 65: we could check for empty `model` first

frontend/public/components/utils/k8s-watch-hook.ts
```diff
@@ -61,6 +61,9 @@ const useModelsLoaded = (): boolean => {
   return ref.current;
 };
 
+export const useModelLoaded = (model: string): boolean =>
```
suomiy, line 64: +1 hook might be useful

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
glekner: depends on the use case, some would use `referenceForModel(model)` but in our case the resource in store is called `VMSnapshotModel` so we need `VMSnapshotModel.kind`

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
glekner: this is a one line and it is a convention.
check https://github.com/openshift/console/blob/master/frontend/webpack.config.ts#L192 for example

frontend/public/components/utils/k8s-watch-hook.ts
glekner: react hooks shouldn't be called conditionally. eslint will shout at you

frontend/public/components/utils/k8s-watch-hook.ts
rawagner: You can use https://github.com/openshift/console/blob/master/frontend/packages/console-shared/src/hooks/useK8sModel.ts instead

suomiy: yeah that sound much better

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
suomiy: I wouldn't cite webpack.config.ts as a convention. I know there are both use cases in the code base.

This one is less elegant. 
Do as you want - I can live with both.

frontend/packages/kubevirt-plugin/src/components/vms/vm-details-page.tsx
suomiy: it might get loaded only with the group+version to the store so we should use `referenceForModel` to be safe

frontend/public/components/utils/k8s-watch-hook.ts
suomiy: yes, unless the condition is inside

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
```diff
@@ -91,7 +98,22 @@ export const VMSnapshotsPage: React.FC<VMLikeEntityTabProps> = ({ obj: vmLikeEnt
     [namespace],
   );
 
-  const [snapshots, snapshotsLoaded, snapshotsError] = useK8sWatchResource<VMSnapshot[]>(resource);
+  const restoreResource: WatchK8sResource = React.useMemo(
```
rawagner, line 101: useK8sWatchResource doesnt require memoized resource anymore :)

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-row.tsx
```diff
@@ -66,12 +56,14 @@ const getActions = (
     return [];
   }
 
-  const actions = [menuActionRestore, menuActionDelete];
+  const actions = [menuActionDelete];
```
suomiy, line 59: what about the snapshot detail later?

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/utils.ts
```diff
@@ -4,6 +4,8 @@ import { Kebab } from '@console/internal/components/utils';
 export const snapshotsTableColumnClasses = [
```
suomiy, line 4: the total is 12

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
```diff
@@ -64,6 +64,13 @@ export const VMSnapshotsTable: React.FC<VMSnapshotsTableProps> = ({
             sortField: 'status.readyToUse',
             transforms: [sortable],
           },
+          {
+            title: 'Last restored',
```
glekner, line 68: @suomiy 
since the latest restore date is computed inside the row component (because one snapshot can have multiple restore objects) we can't sort the table by it. do you have any suggestions?

https://github.com/openshift/console/pull/6807/files#diff-d4bb250ed498bbf43eeee21385193b71R146

rawagner: @glekner you can use sortFunc prop

glekner: done, hope it won't cause performance issues
https://github.com/openshift/console/pull/6807/files#diff-b2d4cf524cc534d555f71cccff72c6ecR172

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-row.tsx
glekner: snapshot detail?
the restore is now a dedicated button for each row, not a kebab option

glekner:
Updated to match latest design

Restore failed popover:
Screen Shot 2020-10-12 at 13 03 24

Restore in progress popover:
Screen Shot 2020-10-12 at 15 18 12

can you review @suomiy @matthewcarleton @yfrimanm

yfrimanm:
Thanks @glekner 
Nit:  For the restoring popover, maybe we can make it clearer with: ''Virtual-machine-name 2019-8-28 is being restored from this snapshot''.

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/types.ts
```diff
@@ -6,4 +7,6 @@ export type VMSnapshotRowCustomData = {
   vmLikeEntity: VMLikeEntityKind;
   columnClasses: string[];
   isDisabled: boolean;
+  restores: VMRestore[];
+  withProgress: (promise: Promise<any>) => void;
```
suomiy, line 11: why do we need withProgress in snapshots?

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
suomiy: I am not a big fan of expecting that the core code should handle our CRDs. Wouldn't it be more straightforward to just resolve the time here and pass it as a data like we do in our other tables?

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
```diff
@@ -81,17 +89,33 @@ export const VMSnapshotsPage: React.FC<VMLikeEntityTabProps> = ({ obj: vmLikeEnt
...
+  };
 
-  const [snapshots, snapshotsLoaded, snapshotsError] = useK8sWatchResource<VMSnapshot[]>(resource);
+  const [snapshots, snapshotsLoaded, snapshotsError] = useK8sWatchResource<VMSnapshot[]>(
+    snapshotResource,
+  );
+  const [restores, restoresLoaded, restoresError] = useK8sWatchResource<VMRestore[]>(
+    restoreResource,
+  );
+  const sortedRestores = React.useMemo(
```
suomiy, line 112: it might be easier to create a map with `getVmRestoreSnapshotName` key here for better consumption. 
Then it wouldn't need to be even sorted as you could just save the relevant restore.

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-row.tsx
```diff
@@ -66,32 +62,103 @@ const getActions = (
...
-    <Status status={readyToUse ? 'Ready' : 'Not Ready'} />
-  );
+export const VMSnapshotStatus: React.FC<VMSnapshotStatusProps> = ({ snapshot, restore }) => {
+  const snapshotError = getVMSnapshotError(snapshot);
+  const restoreError = getVMRestoreError(restore);
+  const isRestoreProgressing = isVmRestoreProgressing(restore);
+  const snapshotReady = isVMSnapshotReady(snapshot);
+
+  if (snapshotError) {
+    return <ErrorStatus title={'Create Failed'}>{snapshotError?.message}</ErrorStatus>;
```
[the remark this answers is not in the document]
glekner: user can access the snapshot details by clicking on its name in the table

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/types.ts
glekner: why not? its a resource creation like any other modal?

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
glekner: which snapshotName? one vm can have multiple snapshots with different names? a row for each.

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/types.ts
suomiy: because this feature is used only by the wizard

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
suomiy: yes, but relevant restore shoul be only one , right?

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/types.ts
suomiy: ah you are right, It is used by the detail pages as well

glekner:
- mapped restores by key
- changed kebab options to: 
1. Last Restore details (if available)
2. Delete Snapshot
image

@suomiy can you review?
the common kebab options can be confusing here because of the "Dual" resource page of snapshots/kebab. if a user wants to play with them, they should visit a specific resource page

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshots.tsx
```diff
@@ -81,17 +93,46 @@ export const VMSnapshotsPage: React.FC<VMLikeEntityTabProps> = ({ obj: vmLikeEnt
...
+  const [snapshots, snapshotsLoaded, snapshotsError] = useK8sWatchResource<VMSnapshot[]>(
+    snapshotResource,
+  );
+  const [restores, restoresLoaded, restoresError] = useK8sWatchResource<VMRestore[]>(
+    restoreResource,
+  );
+
+  const mappedRelevantRestores = React.useMemo(
+    () =>
+      restores.reduce((restoreMap, currentRestore) => {
```
suomiy, line 119: this would be simpler

```tsx
      restores.reduce((restoreMap, currentRestore) => {
        const relevantRestore = restoreMap[getVmRestoreSnapshotName(currentRestore)];
        if (!relevantRestore || new Date(getVmRestoreTime(relevantRestore)).getTime() < new Date(getVmRestoreTime(currentRestore)).getTime()) {
            restoreMap[getVmRestoreSnapshotName(currentRestore)] = currentRestore
        }
        return restoreMap
      }, {}),
```

suomiy:
> the common kebab options can be confusing here because of the "Dual" resource page of snapshots/kebab. if a user wants to play with them, they should visit a specific resource page

I was under an impression that it would make more sense to show snapshots as one logical resources (ie as an aggregation of resources like vm/vmi does). 
Do you think it makes sense to introduce VMRestore resource to the users? And what advantages does it have?

glekner:
* Created custom Details Page for VM Snapshot with Description if available
* Added Restore action, Status and Last Restore(same new component like the table)
* Removed Last Restore Details kebab option in a Snapshot row 
* Added breadcrumbs for VM context

Screen Shot 2020-10-20 at 21 03 00

@suomiy @matthewcarleton @yfrimanm can you take a look

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-details.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const restoreResource: WatchK8sResource = {
+    isList: true,
+    kind: VirtualMachineRestoreModel.kind,
+    namespaced: true,
+    namespace,
+  };
+
+  const [restores, restoresLoaded] = useK8sWatchResource<VMRestore[]>(restoreResource);
+
+  const mappedRelevantRestores = React.useMemo(
```
suomiy, line 104: could we make a hook for this, since we are using ti twice?

suomiy (changes requested):
@glekner we probably should ask for delete confirmation of snapshots when we delete the VM. The same we do with disks. To prevent stale snapshots being stored in the system. 
Thoughts?

Also not sure if we want an owner reference on them.

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-details.tsx
rawagner: you could include `useK8sWatchResource` into the new hook and it would return `[mappedRestores, loaded, error]`

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/snapshot-row.tsx
```diff
@@ -15,38 +18,30 @@ import {
...
-  isDisabled: isVMRunningOrExpectedRunning(asVM(vmLikeEntity)),
-  callback: () =>
-    withProgress(
-      snapshotRestoreModal({
-        snapshot,
-      }).result,
-    ),
-});
 
 const menuActionDelete = (
```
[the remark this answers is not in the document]
glekner: We don't need any of it actually, the only kebab option as of now is Delete.so now just using 
`const { Delete } = Kebab.factory`

glekner:
@suomiy @rawagner 
* Implemented suggestions
* Created `useMappedVMRestores` hook
* Made `DescriptionModal` generic to reuse in the snapshot details.

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/use-mapped-vm-restores.ts
```diff
@@ -0,0 +1,41 @@
...
+            new Date(getVmRestoreTime(currentRestore)).getTime()
+        ) {
+          restoreMap[getVmRestoreSnapshotName(currentRestore)] = currentRestore;
+        }
+        return restoreMap;
+      }, {}),
+    [restores],
+  );
+
+  return [mappedRelevantRestores, restoresLoaded, restoresError];
```
rawagner, line 40: you can return directly from React.useMemo so you wont create a new array every time

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshot-details.tsx
```diff
@@ -0,0 +1,157 @@
...
+        name: 'Virtualization',
+        path: `/k8s/ns/${vmNamespace}/virtualization`,
+      },
+      {
+        name: 'Virtual Machines',
+        path: `/k8s/ns/${vmNamespace}/virtualization`,
+      },
+      {
+        name: `${vmName} Snapshots`,
+        path: `/k8s/ns/${match.params.ns ||
```
suomiy, line 127: can use the variable here as well

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/use-mapped-vm-restores.ts
glekner: fixed

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshot-details.tsx
glekner: fixed

frontend/packages/kubevirt-plugin/src/components/modals/description-modal/description-modal.tsx
```diff
@@ -8,28 +8,26 @@ import {
...
-import { getVMLikeModel } from '../../../selectors/vm';
 import { getUpdateDescriptionPatches } from '../../../k8s/patches/vm/vm-patches';
 
 // TODO: should be moved under kubevirt-plugin/src/style.scss
-import './_vm-description-modal.scss';
+import './_description-modal.scss';
 
-export const VMDescriptionModal = withHandlePromise((props: VMDescriptionModalProps) => {
-  const { vmLikeEntity, inProgress, errorMessage, handlePromise, close, cancel } = props;
+const DescriptionModal = withHandlePromise((props: DescriptionModalProps) => {
```
rawagner, line 18:
```suggestion
const DescriptionModal = withHandlePromise<DescriptionModalProps>((props) => {
```

frontend/packages/kubevirt-plugin/src/components/modals/description-modal/index.ts
```diff
@@ -0,0 +1 @@
+export * from './description-modal';
```
rawagner, line 1: lets avoid creating new index files, they cause cyclic deps and we will be removing them in the future

frontend/packages/kubevirt-plugin/src/components/modals/snapshot-modal/snapshot-modal.tsx
```diff
@@ -58,6 +60,13 @@ const SnapshotsModal = withHandlePromise((props: SnapshotsModalProps) => {
               onChange={(v) => setName(v)}
             />
           </FormRow>
+          <FormRow title="Description" fieldId={asId('desc')}>
+            <TextArea
+              value={description}
+              onChange={(d) => setDescription(d)}
```
rawagner, line 66:
```suggestion
              onChange={setDescription}
```

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshot-details.tsx
```diff
@@ -0,0 +1,156 @@
...
+      {...props}
+      match={match}
+      name={snapshotName}
+      namespace={namespace}
+      kind={VirtualMachineSnapshotModel.kind}
+      kindObj={VirtualMachineSnapshotModel}
+      resources={[resource]}
+      menuActions={menuActions}
+      pages={pages}
+      breadcrumbsFor={(obj) => breadcrumbsForSnapshots(obj)}
```
rawagner, line 149:
```suggestion
      breadcrumbsFor={breadcrumbsForSnapshots}
```

frontend/packages/kubevirt-plugin/src/selectors/selectors.ts
```diff
@@ -26,7 +25,7 @@ export const getLabelValue = (entity: K8sResourceKind, label: string): string =>
 
 // Annotations
 export const getAnnotations = (
-  vm: VMGenericLikeEntityKind,
+  vm: K8sResourceKind,
```
rawagner, line 28:
```suggestion
  vm: K8sResourceCommon,
```

PR #6808, opened 2020-10-01
Bug 1883679: Fix runtime error in legacy operand form

TheRealJon:
Use optional chaining to prevent undefined references in resource requirements input widget in
legacy operand form.

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operand/DEPRECATED_operand-form.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operand/DEPRECATED_operand-form.tsx
@@ -779,9 +779,9 @@ export const DEPRECATED_CreateOperandForm: React.FC<OperandFormProps> = ({
           <dt>Limits</dt>
           <dd>
             <ResourceRequirements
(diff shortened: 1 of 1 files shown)


frontend/packages/operator-lifecycle-manager/src/components/operand/DEPRECATED_operand-form.tsx
```diff
@@ -779,9 +779,9 @@ export const DEPRECATED_CreateOperandForm: React.FC<OperandFormProps> = ({
           <dt>Limits</dt>
           <dd>
             <ResourceRequirements
-              cpu={currentValue.getIn(_.toPath(cpuLimitsPath))}
-              memory={currentValue.getIn(_.toPath(memoryLimitsPath))}
-              storage={currentValue.getIn(_.toPath(storageLimitsPath))}
+              cpu={currentValue?.getIn?.(_.toPath(cpuLimitsPath))}
```
spadgett, line 782: Do we need the `?` after `getIn`? I'd expect that to always work if `currentValue` isn't null.

spadgett (approved):
/lgtm

PR #6816, opened 2020-10-02
Bug 1852743: Show total cores in node list

spadgett:
/assign @rhamilto 

Screenshot_2020-10-02 Nodes · OKD


frontend/packages/console-app/src/components/nodes/NodesPage.tsx
```diff
@@ -443,6 +448,10 @@ const fetchNodeMetrics = (): Promise<NodeMetrics> => {
       key: 'cpu',
       query: 'sum by(instance) (instance:node_cpu:rate:sum)',
     },
+    {
+      key: 'totalCPU',
+      query: 'sum by(instance) (instance:node_num_cpu:sum)',
```
spadgett, line 453: This is the same query we use on the node dashboard.

dtaylor113 (approved):
/lgtm

PR #6817, opened 2020-10-02
Add "clean-deps" and "clean-install" to yarn scripts.

TheRealJon:
- "clean-deps": Recursively remove all node_modules directories in the project
- "clean-install": Remove and reinstall all dependencies

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,6 +10,8 @@
   ],
   "scripts": {
     "clean": "rm -rf ./public/dist",
+    "clean-deps": "rm -rf ./**/**/node_modules",
+    "clean-install": "yarn clean-deps && yarn",
     "dev": "yarn clean && yarn gql-generate && NODE_OPTIONS=--max-old-space-size=4096 ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/webpack-dev-server",
     "dev-once": "yarn clean && yarn gql-generate && NODE_OPTIONS=--max-old-space-size=4096 ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/webpack --mode=development",
     "build": "yarn clean && yarn gql-generate && NODE_ENV=production NODE_OPTIONS=--max-old-space-size=4096 ts-node -O '{\"module\":\"commonjs\"}' ./node_modules/.bin/webpack --mode=production",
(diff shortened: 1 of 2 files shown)


dtaylor113 (approved):
/lgtm

frontend/package.json
```diff
@@ -10,6 +10,8 @@
   ],
   "scripts": {
     "clean": "rm -rf ./public/dist",
+    "clean-deps": "rm -rf ./**/**/node_modules",
+    "clean-install": "yarn clean-deps && yarn",
```
spadgett, line 14: I don't mind this, but we do have `clean-frontend.sh`. Do we need both?

TheRealJon: It's more just for convenience than anything. We could maybe just make these scripts reference the `clean-frontend.sh` script?

TheRealJon: It saves a couple of extra commands and is consistent with other helper yarn scripts. Currently, if you want to do a clean reinstall of frontend deps, and your current working dir is frontend (which at least for me, is almost always the case), you have to:

```
cd .. 
./clean-frontend.sh 
cd frontend 
yarn
``` 
whereas with these yarn scripts, you can do the same with just one command in your current working dir.

TheRealJon: Also, the `clean-frontend.sh` script does more than just remove frontend deps. These are focused specifically on just doing a clean reinstall of the frontend deps.

dtaylor113: I'm fine either way, but wouldn't the flow be:
```
console>./clean-frontend.sh
console>./build-frontend.sh    <--- does yarn install
```

TheRealJon: Yeah, and that is a production build as well, which might not be needed. These were really meant to just be used for dev purposes. Whenever I switch to an older release branch or I'm troubleshooting unusual build errors, I'll do a clean install of the frontend deps. These scripts were just meant to make that more convenient.

dtaylor113 (approved):
/lgtm

frontend/package.json
spadgett: I'd think it would be very rare that you need to do this. When I switch branches, `yarn install` almost always works. If you're having strange build errors, you probably want to clean everything. The reinstall of all the dependencies takes much longer than the rest of the build, so I don't really see a reason not to clean everything to build fresh.

If there is a common problem you're seeing switching branches, should we look at fixing that so this isn't needed?

PR #6820, opened 2020-10-02
Bug 1879282: Update doc links in installed operators page

harishsurf:
Update doc links in installed operators page


spadgett (approved):
/lgtm

Let's fix in 4.6 since the link was wrong. Thanks @harishsurf

PR #6824, opened 2020-10-02
Bug 1876701: Fix some spec descriptor widgets inline edit behavior

TheRealJon:
Fix incorrectly formatted patch path in configureSizeModal and configureUpdateStrategyModal.

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/configure-size.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/configure-size.tsx
@@ -1,6 +1,7 @@
 import { configureCountModal } from '@console/internal/components/modals';
 import { K8sResourceKind, K8sKind } from '@console/internal/module/k8s';
 import { Descriptor } from '../types';
+import { getPatchPathFromDescriptor } from '../utils';
 
 export const configureSizeModal = ({
   kindObj,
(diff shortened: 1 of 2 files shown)


frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/configure-size.tsx
```diff
@@ -14,7 +14,7 @@ export const configureSizeModal = ({
     defaultValue: specValue || 0,
     title: `Modify ${specDescriptor.displayName}`,
     message: specDescriptor.description,
-    path: `/spec/${specDescriptor.path}`,
+    path: `/spec/${specDescriptor.path.replace(/\]/g, '').replace(/[.[]/g, '/')}`,
```
spadgett, line 17: I'd suggest using `_.toPath` and then building up the patch path from the array. Something like (untested)

```suggestion
    path: ['spec', _.toPath(specDescriptor.path)].join('/'),
```

spadgett:
Are the other descriptors that could have this problem?

frontend/packages/operator-lifecycle-manager/src/components/descriptors/spec/configure-size.tsx
TheRealJon: @spadgett fixed

TheRealJon:
Yes, already found another one. Let me scour through them and fix them all.

TheRealJon:
@spadgett Turns out there was only one other instance of this problem in the descriptors. Should be fixed now.

spadgett (approved):
/lgtm

👍

PR #6831, opened 2020-10-05
Query Browser: Keep tooltip constrained to graph frame

kyoto:
The tooltip could stick out from the left or right side of the graph
frame and be partially obscured by the edge of the window.

This commit keeps the tooltip from going too far out from the graph
frame by limiting its X position, then shifting the tooltip arrow left
or right to keep it pointing at the selected data point.

screenshot


spadgett (approved):
/lgtm

PR #6835, opened 2020-10-05
Bug 1885309:  Fix bug where <Timestamp> for 12 hour has incorrect suffix

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1883779

Before:
Screen Shot 2020-10-05 at 9 50 40 AM

After:
Screen Shot 2020-10-05 at 9 49 08 AM

This is a pretty bad bug, so we're gonna want to back port to 4.5?

### DIFF
--- a/frontend/public/components/utils/timestamp.tsx
+++ b/frontend/public/components/utils/timestamp.tsx
@@ -38,8 +38,10 @@ const timestampFor = (mdate: Date, now: Date, omitSuffix: boolean) => {
 
   let a = 'am';
   let hours = mdate.getHours();
-  if (hours > 12) {
-    hours -= 12;
+  if (hours >= 12) {
+    if (hours > 12) {
+      hours -= 12;
+    }
     a = 'pm';
   }
 


rhamilto:
/hold

I think this breaks 12 am.  Investigating...

spadgett (approved):
/lgtm

rhamilto:
/hold cancel

There is an existing bug where `12 am` displays as `0 am`, but after a conversation with @spadgett, we're opting to leave the `am` bug as it is relatively minor, and we're switching to moment.js for timestamps in 4.7, which will address both bugs.

rhamilto:
/hold

Revisiting with @spadgett.

rhamilto:
/hold cancel

@rebeccaalpert will fix https://bugzilla.redhat.com/show_bug.cgi?id=1883779 with https://github.com/openshift/console/pull/6194 and we'll just fix the `12 pm` case in 4.6.z with this PR.

spadgett (approved):
/lgtm

PR #6839, opened 2020-10-05
Use Pod resource data when necessary in topology

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4966

**Description**
Currently, we determine pod status data for each workload upon receiving resource data and carry that data around for each workload node. This PR updates to retrieve pod resources to determine status and list pods in the components that show the pods' status/list.

Non-functional change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup


frontend/packages/console-shared/src/hooks/usePodsWatcher.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const resources = useK8sWatchResources(watchedResources);
+
+  React.useEffect(() => {
+    const errorKey = Object.keys(resources).find((key) => resources[key].loadError);
+    if (errorKey) {
+      setLoadError(resources[errorKey].loadError);
+      return;
+    }
+    setLoadError(null);
```
jerolimov, line 30: The initial value is `""` and typed as string. If the effect does not find an error this sets the error from `""` to `null` which force a re-rendering. Can we use `null` or `""` as initial state and here as well?

frontend/packages/console-shared/src/utils/pod-utils.ts
```diff
@@ -173,18 +173,14 @@ export const podDataInProgress = (
...
+  podRCData: PodRCData,
 ): { inProgressDeploymentData: ExtPodKind[] | null; completedDeploymentData: ExtPodKind[] } => {
-  const strategy: DeploymentStrategy = _.get(dc, ['spec', 'strategy', 'type'], null);
-  const currentDeploymentphase = current && current.phase;
-  const currentPods = current && current.pods;
-  const previousPods = previous && previous.pods;
+  const strategy: DeploymentStrategy = _.get(podRCData.obj, ['spec', 'strategy', 'type'], null);
+  const currentDeploymentphase = podRCData.current && podRCData.current.phase;
+  const currentPods = podRCData.current && podRCData.current.pods;
+  const previousPods = podRCData.previous && podRCData.previous.pods;
```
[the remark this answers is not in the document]
jeff-phillips-18: It has been this way. I think it better for constructors to explicitly set them to null when need be.

frontend/packages/console-shared/src/hooks/usePodsWatcher.ts
jerolimov: @jeff-phillips-18 Thanks for removing the initial value "". But the default value is now undefined, and setting null will still re-render the component unnecessary. Using the exact same time in both cases or wrapping setLoadError(null) in an if would help.

jeff-phillips-18:
Obviously this PR is too large. Closing and will submit smaller chunks in separate PRs.

PR #6841, opened 2020-10-05
Bug 1885315: specify full regex for jest transform

suomiy:
fixes unit tests on my machine

I don't know what is the real cause, but specifying the full regex fixed my issue. 
I looked into jest docs and the code, but could not figure out why it happens.

@spadgett @rawagner please take a look


spadgett (approved):
/lgtm

PR #6840, opened 2020-10-05
Update TaskRun List Page Columns & PipelineRun Tab

debsmita1:
**JIRA story:**
https://issues.redhat.com/browse/ODC-4800

**Solution Description:**
- added `Task Runs` tab in the pipelineRun details page which lists all the Task Runs associated with the Pipeline Run
- Added columns `pipeline` , `task` & `pod`  & removed `duration`
- The pipeline column is not shown in the TaskRun Tab under Pipeline Run details page
- fixed the TaskRun list in the Search page
- added tests for `TaskRunsRow`

**GIF/Screenshots:**
PDtaskruns
taskruns1
TaskRuns
all-pro
search

**Test Coverage:**
test1

PR #6851, opened 2020-10-06
TaskRun details and log page

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-4801

# Screenshot
task-run-det-multi

# Tests
Added two test suites, TaskRunDetails and TaskRunLogs

# Browser conformance
chrome


frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return <StatusBox loaded={podLoaded} loadError={podError} label="TaskRun Log" />;
+};
+
+// to safeguard against cases where Pod is yet to be assigned or is removed
+export default (props: TaskRunLogProps) =>
+  props.obj.status?.podName ? (
+    <TaskRunLog {...props} />
+  ) : (
+    <StatusBox loadError="Pod not found" label="TaskRun Log" />
```
karthikjeeyar, line 29: Why can't we have this check in the above function and early exit if podName is not available :thinking:

abhinandan13jan: It works, but it fails lint check when there is an `if` block before the `useK8sGet` Hook

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
```diff
@@ -0,0 +1,30 @@
...
+  obj: K8sResourceKind;
+};
+
+const TaskRunLog: React.FC<TaskRunLogProps> = ({ obj }) => {
+  const [pod, podLoaded, podError] = useK8sGet<K8sResourceKind>(
+    PodModel,
+    obj.status.podName,
+    obj.metadata.namespace,
+  );
+  if (podLoaded) {
```
sahil143, line 18:
```suggestion
  if (podLoaded && !podError) {
```

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
sahil143: Best to name this wrapper component and export that as default.

sahil143 (changes requested):

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
abhinandan13jan: updated this as well as named to `TaskRunLogWrapper`

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
abhinandan13jan: named as wrapper

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
```diff
@@ -0,0 +1,32 @@
...
+    obj.metadata.namespace,
+  );
+  if (podLoaded && !podError) {
+    return <PodLogs obj={pod} />;
+  }
+  return <StatusBox loaded={podLoaded} loadError={podError} label="TaskRun Log" />;
+};
+
+// to safeguard against cases where Pod is yet to be assigned or is removed
+const TaskRunLogWrapper = (props: TaskRunLogProps) =>
```
sahil143, line 25: This is a Functional Component. Should type it properly
```suggestion
const TaskRunLogWrapper: React.FC<TaskRunLogProps> = (props) =>
```

sahil143 (commented):
@abhinandan13jan Logs doesn't seem to be streaming. It just says loading and when step is changed from the dropdown. There is an error
Screenshot 2020-10-09 at 10 23 47 AM
Screenshot 2020-10-09 at 10 24 00 AM
But it streams fine in the Pipeline run logs
Screenshot 2020-10-09 at 10 24 13 AM

frontend/packages/dev-console/src/components/taskruns/TaskRunDetails.tsx
```diff
@@ -15,6 +19,15 @@ const TaskRunDetails: React.FC<TaskRunDetailsProps> = ({ obj: taskRun }) => {
...
           <ResourceSummary resource={taskRun} />
         </div>
+        <div className="col-sm-6 odc-taskrun-details__status">
+          <dl>
+            <dt>Status</dt>
+            <dd>
+              <Status status={pipelineRunFilterReducer(taskRun)} />
+            </dd>
+          </dl>
+          <PipelineRunDetailsErrorLog pipelineRun={taskRun} />
```
karthikjeeyar, line 29: These two pipelineRun specific functions called in the context of `taskRun` looks odd to me. If taskRun has more specific type definition, then this will cause lint errors and confusions in future.  It will be good to refactor these to a generic function calls. WDYT?

abhinandan13jan:
> @abhinandan13jan Logs doesn't seem to be streaming. It just says loading and when step is changed from the dropdown. There is an error
> Screenshot 2020-10-09 at 10 23 47 AM
> Screenshot 2020-10-09 at 10 24 00 AM
> But it streams fine in the Pipeline run logs
> Screenshot 2020-10-09 at 10 24 13 AM

Strange, I find it to work the same as Pod log

trlog2
trlog1

frontend/packages/dev-console/src/components/taskruns/TaskRunDetails.tsx
abhinandan13jan: I've updated such that PLR and TR uses separate utils and point to a common component

frontend/packages/dev-console/src/components/pipelineruns/logs/LogSnippetBlock.tsx
```diff
@@ -1,19 +1,30 @@
...
 import { PipelineRun } from '../../../utils/pipeline-augment';
 import LogSnippetFromPod from './LogSnippetFromPod';
 import { PipelineRunErrorDetails } from './pipelineRunLogSnippet';
 
 type LogSnippetBlockProps = {
   children: (logSnippet: string) => React.ReactNode;
   logDetails: PipelineRunErrorDetails;
-  pipelineRun: PipelineRun;
+  pipelineRun?: PipelineRun;
+  taskRun?: K8sResourceKind;
```
andrewballantyne, line 11: Maybe we just simplify this to a single `K8sResourceKind` or even better down to just `namespace`?

frontend/packages/dev-console/src/components/taskruns/__tests__/TaskRunDetails.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+  it('Renders a Resource Summary', () => {
+    expect(pipelineRunWrapper.find(ResourceSummary).exists());
+  });
+  it('Renders a Status', () => {
+    expect(pipelineRunWrapper.find(Status).exists());
+  });
+  it('Renders a Task Run Log snippet', () => {
+    expect(pipelineRunWrapper.find(TaskRunDetailsErrorLog).exists());
+  });
+});
```
andrewballantyne, line 42: Testing against static content won't save us from anything. The best it can do is fail a test when we change the layout... there isn't any conditional here that we would be concerned about. Ie, no logic breaks can cause these tests to fail.

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { useK8sGet } from '@console/internal/components/utils/k8s-get-hook';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { PodLogs } from '@console/internal/components/pod-logs';
+import { StatusBox } from '@console/internal/components/utils/status-box';
+import { PodModel } from '@console/internal/models';
+
+export type TaskRunLogProps = {
+  obj: K8sResourceKind;
```
andrewballantyne, line 9:
```suggestion
  obj: TaskRunKind;
```

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/taskruns/__tests__/TaskRunDetails.spec.tsx
abhinandan13jan: This is just based on current acceptance criteria. Just to make sure, the code adheres to the AC

frontend/packages/dev-console/src/components/taskruns/TaskRunLog.tsx
abhinandan13jan: updated

frontend/packages/dev-console/src/components/pipelineruns/logs/LogSnippetBlock.tsx
abhinandan13jan: simplified to namespace for all cases

karthikjeeyar (commented):
I observe the same issue as sahil mentioned above.When you create a taskRun, logs are not streaming correctly.

frontend/packages/dev-console/src/components/taskruns/__tests__/TaskRunLog.spec.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { PodLogs } from '@console/internal/components/pod-logs';
+import { StatusBox } from '@console/internal/components/utils/status-box';
+import { shallow } from 'enzyme';
```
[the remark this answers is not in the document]
abhinandan13jan: updated

frontend/packages/dev-console/src/components/taskruns/__tests__/TaskRunDetails.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+          lastTransitionTime: '2019-10-29T11:57:53Z',
+          reason: 'Failed',
+          status: 'True',
+          type: 'Failed',
+        },
+      ],
+    },
+  },
+};
+
```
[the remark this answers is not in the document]
abhinandan13jan: moved

frontend/packages/dev-console/src/utils/pipeline-filter-reducer.ts
```diff
@@ -58,3 +58,8 @@ export const pipelineResourceTypeFilter = (filters, pipelineResource): boolean =
   const type = pipelineResourceFilterReducer(pipelineResource);
   return filters.selected.has(type) || !_.includes(filters.all, type);
 };
+
+export const taskRunFilterReducer = (taskRun): string => {
+  const status = pipelineRunStatus(taskRun);
```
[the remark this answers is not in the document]
abhinandan13jan: There's one issue with that. It's a cyclic dependence and pipeline-filter-reducer doesn't have types. It should be a separate issue to refactor `Types` out of pipeline-augment. cc: @andrewballantyne

frontend/packages/dev-console/src/utils/pipeline-filter-reducer.ts
```diff
@@ -58,3 +58,8 @@ export const pipelineResourceTypeFilter = (filters, pipelineResource): boolean =
   const type = pipelineResourceFilterReducer(pipelineResource);
   return filters.selected.has(type) || !_.includes(filters.all, type);
 };
+
+export const taskRunFilterReducer = (taskRun): string => {
```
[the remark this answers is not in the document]
abhinandan13jan: cyclic dependence will be there

abhinandan13jan:
> I observe the same issue as sahil mentioned above.When you create a taskRun, logs are not streaming correctly.

Fixed it with useK8sWatchResource hook

bgliwa01:
Will all of the columns be added on the list view? Right now I only see namespace and created, but there should be others included as seen in this PR - https://github.com/openshift/console/pull/6840#event-3844836192 

I'm not sure if these designs will automatically update based on Deb's PR or not...

serenamarie125 (changes requested):
Based on @siamaksade comment ... 
> Strange, I find it to work the same as Pod log

@abhinandan13jan If possible, I'd suggest that we use the same component from the Log tab in PipelineRun here, but just not show the left nav piece with the task list

I think this was a miscommunication because we discussed "using conventions" :)

abhinandan13jan:
> Will all of the columns be added on the list view? Right now I only see namespace and created, but there should be others included as seen in this PR - [#6840 (comment)](https://github.com/openshift/console/pull/6840#event-3844836192)
> 
> I'm not sure if these designs will automatically update based on Deb's PR or not...

@bgliwa01 Yes, the list view will be updated as per @debsmita1 's PR. You can ignore the list view shown in my screenshots :)

abhinandan13jan:
> Based on @siamaksade comment ...
> 
> > Strange, I find it to work the same as Pod log
> 
> @abhinandan13jan If possible, I'd suggest that we use the same component from the Log tab in PipelineRun here, but just not show the left nav piece with the task list
> 
> I think this was a miscommunication because we discussed "using conventions" :)

I've updated to use the Multistreaming Log

task-run-det-multi

karthikjeeyar:
Verified locally, works fine
/lgtm

bgliwa01 (approved):
lgtm

invincibleJai:
@abhinandan13jan looks like, it needs rebase

frontend/packages/dev-console/src/components/taskruns/TaskRunDetailsErrorLog.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { getLogSnippet } from '../pipelineruns/logs/pipelineRunLogSnippet';
+import LogSnippetBlock from '../pipelineruns/logs/LogSnippetBlock';
+import { TaskRunKind } from '../../utils/pipeline-augment';
+
+type TaskStatusLogProps = {
+  taskRun: TaskRunKind;
+};
+
+const TaskRunDetailsErrorLog: React.FC<TaskStatusLogProps> = ({ taskRun }) => {
+  const logDetails = getLogSnippet(taskRun);
```
jerolimov, line 11: `getLogSnippet` only accepts `PipelineRun` instances and no `TaskRun` types here. The getLogSnippet method converts the given PipelineRun to a logDetails object which will be used by `LogSnippetBlock` to fetch the correct log.

For a PipelineRun this method return an object like this one:

```tsx
{
  containerName: "step-build",
  podName: "docker-bs1qu0-build-qmsfx-pod-5hpp4",
  title: "Failure on task build - check logs for details.",
}
```

But for the TaskRun it returns this message (extracted from yaml):

```tsx
{
  staticMessage: "\"step-build\" exited with code 1 (image: \"registry.redhat.io/rhel8/buildah@sha256:180c4d9849b6ab0e5465d30d4f3a77765cf0d852ca1cb1efb59d6e8c9f90d467\"); for logs run: kubectl -n christoph-test logs docker-bs1qu0-build-qmsfx-pod-5hpp4 -c step-build\n",
  title: "Failure - check logs for details.",
}
```

This is the reason why we see currently a different error log does on the PipelineRun Detail page and on the TaskRun detail page.

The PipelineRun detail page fetches the log from the Pod:

image

But the TaskRun detail page shows an error message from the yaml:

image

I think we should write a similar getLogSnippet (or getLogDetails) for TaskRuns which also extracts the right podName and containerName (and shows the current message also as fallback).

The status steps contains all required infos for this:

```yaml
status:
  completionTime: '2020-10-21T00:24:58Z'
  conditions:
    - message: >
        "step-build" exited with code 1.... # FALLBACK
      reason: Failed
      status: 'False'
      type: Succeeded
  podName: docker-bs1qu0-build-qmsfx-pod-5hpp4
  steps:
    - container: step-generate
      name: generate
      terminated:
        exitCode: 0
        reason: Completed
    - container: step-build
      name: build
      terminated:
        exitCode: 1
        reason: Error    # I expect that we want see the log snippet of this container
```

jerolimov (changes requested):
/hold

abhinandan13jan:
Screenshot from 2020-10-22 15-26-08

@jerolimov Thanks for pointing it out. I've separated out and refactored to use a new logsnippet for TaskRuns
@invincibleJai Rebased
@karthikjeeyar could you please re-review once the tests are done.

frontend/packages/dev-console/src/utils/pipeline-filter-reducer.ts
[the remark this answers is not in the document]
jerolimov: I would prefer here to not reuse `pipelineRunStatus` at all and to not mix PipelineRun and TaskRun objects. I personally would prefer in this case "duplicated" code for both types instead of methods which supports both.

frontend/packages/dev-console/src/components/taskruns/TaskRunDetailsErrorLog.tsx
jerolimov: Could verify that this is solved now :+1:

jerolimov:
Tested this locally and it works fine for me. There are an option question from Karthik about an a missing Type. Andrew was cc-ed and I added my opinion as well, otherwise I would give a lgtm here.

/hold cancel

jerolimov:
/lgtm

Let us see what @andrewballantyne says to the type definitions in pipeline-filter-reducer.ts

jerolimov (approved):

frontend/packages/dev-console/src/components/taskruns/__tests__/TaskRunDetails.spec.tsx
andrewballantyne: Unit tests guard against accidental breaks in logic.

Eg. A unit test for `add5()` would be `add5(5).toBe(10)`... you could test positive numbers, negative numbers, strings (should return NaN), objects, etc, etc...

If you wrap your `add5()`:

```
function getOffsetNumber() {
  return add5(37);
}
```
This function doesn't get you much but it allows the caller not to worry about providing this static `37` (this is an overly simple example). A testing of this function doesn't mean much. `getOffsetNumber().toBe(42)`... how will this test break? If we choose `37` is not a valid constant anymore? So we just change the test and the code at the same time and nothing super shocking here happens. No assistance is really provided by our test, no assurances we don't accidentally break.

Tests are so you can go into `add5` and change the fact that it doesn't add 5 anymore, it subtracts 8 or does something else (uses math.random for instance). Zero inputs or inputs that don't change the output are not test worthy scenarios. They just explicitly tie our tests to our implementation and thus makes us update one more place when we need to change something. It doesn't guard against breaks.

With this mental model, React Components have a similar structure with 1 extra condition. We only shallowly look at the structure of a component. So if your React Component does not take any props or the props are not conditional rendering props but just pass-thru props, then there are no tests needing to be written. 

So in summary, because I ranted a bit here. Don't test a component that does not use conditional rendering as you're not really writing a test, you're writing just 1 more spot to update if we need to change anything.

frontend/packages/dev-console/src/utils/pipeline-filter-reducer.ts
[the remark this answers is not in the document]
andrewballantyne: Yeah, types are annoying in this file. So it should be handled by the work to take Pipelines out of DevConsole package

frontend/packages/dev-console/src/components/taskruns/events/__tests__/TaskRunDetailsPage.spec.tsx
```diff
@@ -36,7 +36,7 @@ describe('TaskRunDetailsPage:', () => {
   it('Should contain three tabs in the Details Page', () => {
     const wrapper = shallow(<TaskRunDetailsPage {...taskRunDetailsPageProps} />);
     const { pages } = wrapper.props();
-    expect(pages).toHaveLength(3);
+    expect(pages).toHaveLength(4);
```
andrewballantyne, line 39: Point in case to a bad test. Did we worry about the number of pages dynamically changing on us? You just added a new tab, and then went an updated a test, likely after it failed on you and not as a proactive measure to drive a TDD scenario... so, what value does this test get us?

andrewballantyne (changes requested):
I'm seeing a lot of DRY problems here.

1. `getLogSnippet` for PipelineRun and TaskRun have similar logic, we _might_ be able to improve this a bit
2. `joinConditions` is repeated in both `taskRunLogSnippet.ts` and `pipelineRUnLogSnippet.ts`
3. The types are duplicated between both log file... we should have a generic type -- this also leads to `LogSnippetBlock` having the type `PipelineRunErrorDetails` for `logDetails` of which you pass a TaskRunErrorDetails value to

There are also issues with the tests...

- We have static tests tied making sure static content is rendered
- We have tests looking into the contents of another component to see if they are getting the right values
- We have `pipelineRunWrapper` in TaskRun tests

frontend/packages/dev-console/src/components/taskruns/events/__tests__/TaskRunDetailsPage.spec.tsx
abhinandan13jan: This was added as part of some separate PR and I felt simply to update it. I've removed the block. @andrewballantyne

abhinandan13jan:
> I'm seeing a lot of DRY problems here.
> 
> 1. `getLogSnippet` for PipelineRun and TaskRun have similar logic, we _might_ be able to improve this a bit
> 2. `joinConditions` is repeated in both `taskRunLogSnippet.ts` and `pipelineRUnLogSnippet.ts`
> 3. The types are duplicated between both log file... we should have a generic type -- this also leads to `LogSnippetBlock` having the type `PipelineRunErrorDetails` for `logDetails` of which you pass a TaskRunErrorDetails value to
> 
> There are also issues with the tests...
> 
> * We have static tests tied making sure static content is rendered
> * We have tests looking into the contents of another component to see if they are getting the right values
> * We have `pipelineRunWrapper` in TaskRun tests

@andrewballantyne I've refactored and moved common `utils` from the `logSnippets` into a separate file and the common `types` as well. 
I've removed the `static tests from the DetailsPage component` and have added tests for the logSnippet utils.
I've also removed the number of tabs tests which was part of another PR.

frontend/packages/dev-console/src/components/taskruns/events/__tests__/TaskRunDetailsPage.spec.tsx
andrewballantyne: I understand this was written in another PR, but it was an example to help you understand what happens when you write a "static verification" test. There isn't really any gain here, nothing is protected from accidental breaks, and you just end up modifying the test as you change the content.

frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
```diff
@@ -4,26 +4,11 @@ import {
   PLRTaskRunData,
   PLRTaskRunStep,
 } from '../../../utils/pipeline-augment';
+import { CombinedErrorDetails } from './log-snippet-types';
 import { pipelineRunStatus } from '../../../utils/pipeline-filter-reducer';
+import { taskRunSnippetMessage } from './log-snippet-utils';
```
andrewballantyne, line 9: You created a generic `log-snippet-utils`... perhaps we just move this content to that file instead of having a PLR one and a "generic" one that only contains TaskRun functionality. Just merge the two and we don't need to worry about it.

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
```diff
@@ -0,0 +1,24 @@
...
+  containerName: string;
+  podName: string;
+};
+export type ErrorDetailsWithStaticLog = ErrorDetails & {
+  staticMessage: string;
+};
+
+export type CombinedErrorDetails = ErrorDetailsWithLogName | ErrorDetailsWithStaticLog;
+
+export type LogSnippetTask = K8sResourceKind & {
```
andrewballantyne, line 18: Nit... My only reasoning for this would be so that it's metadata the reader of the code can gleam and know it's not all that K8s has to offer. 
```suggestion
type LoggablePipelineKinds = PipelineRun | TaskRunKind;
export type LogSnippetTask = LoggablePipelineKinds & {
```

frontend/packages/dev-console/src/components/pipelineruns/logs/__tests__/log-snippet-utils.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { LogSnippetTaskData } from './log-snippet-test-data';
+import { taskRunSnippetMessage } from '../log-snippet-utils';
+import { ErrorDetailsWithLogName, ErrorDetailsWithStaticLog } from '../log-snippet-types';
+
+const testConatiner = 'container-B';
```
andrewballantyne, line 5: Typo...
```suggestion
const testContainer = 'container-B';
```

frontend/packages/dev-console/src/components/taskruns/logs/taskRunLogSnippet.ts
```diff
@@ -0,0 +1,58 @@
...
+type ErrorDetailsWithStaticLog = ErrorDetails & {
+  staticMessage: string;
+};
+
+export type TaskRunErrorDetails = ErrorDetailsWithLogName | ErrorDetailsWithStaticLog;
+
+const joinConditions = (conditions: Condition[]) =>
+  conditions.map((condition) => condition.message).join('\n') || 'Unknown failure condition';
+
+export const getLogSnippet = (taskRun: TaskRunKind): TaskRunErrorDetails => {
```
andrewballantyne, line 21: I'm confused... you refactored the PipelineRunLogSnippet to be more able to work with both, why is this still the old (almost)duplicated format?

It appears this is also used by `TaskRunDetailsErrorLog`. My point in my last review was to merge these two, which I thought you were doing with the changes to `pipelineRunLogSnippet.ts`.

What am I missing?

frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
```diff
@@ -4,26 +4,11 @@ import {
...
-  staticMessage: string;
-};
-
-export type PipelineRunErrorDetails = ErrorDetailsWithLogName | ErrorDetailsWithStaticLog;
-
-const joinConditions = (conditions: Condition[]) =>
-  conditions.map((condition) => condition.message).join('\n') || 'Unknown failure condition';
-
-export const getLogSnippet = (pipelineRun: PipelineRun): PipelineRunErrorDetails => {
+export const getLogSnippet = (pipelineRun: PipelineRun): CombinedErrorDetails => {
```
andrewballantyne, line 11: Would appear I failed to write tests for this method. Would you mind doing that as part of this PR?

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-utils.ts
```diff
@@ -0,0 +1,25 @@
...
+import { CombinedErrorDetails, LogSnippetTask } from './log-snippet-types';
+
+const joinConditions = (conditions: Condition[]) =>
+  conditions.map((condition) => condition.message).join('\n') || 'Unknown failure condition';
+
+export const taskRunSnippetMessage = (
+  taskRun: LogSnippetTask,
+  containerName: string,
+): CombinedErrorDetails => {
+  const taskName = taskRun.pipelineTaskName || taskRun.metadata.name;
```
andrewballantyne, line 11: Hmm, these resources are so alike, but just different enough to dirty the implementation here.

I wonder if instead of passing a `LogSnippetTask` we could look to pass the status object and the name separately, so this internal method is unaware of the relationship between these two separate but same data structures.

Ie:
```suggestion
export const taskRunSnippetMessage = (
  taskName: string,
   // we may want to keep `taskRunStatus` as a stand alone type, but I expanded inline to show what I am referring to
  taskRunStatus: { podName: string, conditions: ConditionType[] },
  containerName: string,
): CombinedErrorDetails => {
  const taskName = taskRun.pipelineTaskName || taskRun.metadata.name;
```

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
andrewballantyne: Pretty sure this is moot now based on one of my comments, but I want to leave it for clarity and if I am wrong, I still want to report the Nit 😄

andrewballantyne (changes requested):
Closer, but I'm confused a bit on the existence of `taskRunLogSnippet.ts`

frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
abhinandan13jan: well, I think it is elaborate enough. `taskRunSnippetMessage` which is fundamental message block for both the taskSnippets. So, if you look close, the final output of both the LogSnippet a PipelineRun/ TaskRun is to return a message as per the relevant taskRun container/pod or a static one.

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
abhinandan13jan: It won't work as the earlier implementation of TaskRunKind is all messy

frontend/packages/dev-console/src/components/taskruns/logs/taskRunLogSnippet.ts
abhinandan13jan: For PLR -> Check the PLR status -> search for the relevant TaskRun which failed -> return correct message based on TaskRun
For TaskRun -> Check TaskRun status -> return correct message based on the TaskRun

The last step in both is the same. Hence, I have written that code in the `log-util` file and rest is separate.

frontend/packages/dev-console/src/components/pipelineruns/logs/pipelineRunLogSnippet.ts
abhinandan13jan: This PR is dragging for a long time. Maybe it'll be better to write under later tech debts, as there are lots of other pending stories

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-utils.ts
abhinandan13jan: That was the case in the very beginning of the PR. And there would have been no need to keep a separate util for rendering out of  strings.

abhinandan13jan: updated

frontend/packages/dev-console/src/components/pipelineruns/logs/__tests__/log-snippet-utils.spec.ts
abhinandan13jan: updated

frontend/packages/dev-console/src/components/taskruns/logs/taskRunLogSnippet.ts
abhinandan13jan: For `PipelineRunDetailsErrorLog` vs `TaskRunDetailsErrorLog`, if you go by the previous comments, it was decided that the two be kept separate. Earlier we were using the same component.

frontend/packages/dev-console/src/components/pipelineruns/logs/RunDetailsErrorLog.tsx
```diff
@@ -1,15 +1,12 @@
 import * as React from 'react';
-import { PipelineRun } from '../../../utils/pipeline-augment';
-import { getLogSnippet } from './pipelineRunLogSnippet';
 import LogSnippetBlock from './LogSnippetBlock';
 
 type PipelineStatusLogProps = {
-  pipelineRun: PipelineRun;
+  logDetails;
```
andrewballantyne, line 5: Need a type

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
```diff
@@ -0,0 +1,25 @@
...
+export type CombinedErrorDetails = ErrorDetailsWithLogName | ErrorDetailsWithStaticLog;
+
+export type TaskRunStatus = {
+  podName?: string;
+  conditions: Condition[];
+};
+
+export type TaskRunData = K8sResourceKind & {
+  status: TaskRunStatus;
+};
```
andrewballantyne, line 25: Should just add these to the `TaskRunKind`...

frontend/packages/dev-console/src/components/taskruns/TaskRunDetails.tsx
```diff
@@ -1,10 +1,17 @@
...
+import { taskRunFilterReducer } from '../../utils/pipeline-filter-reducer';
+import RunDetailsErrorLog from '../pipelineruns/logs/RunDetailsErrorLog';
 import { TaskRunModel } from '../../models';
+import { getTRLogSnippet } from './logs/taskRunLogSnippet';
+import { TaskRunStatus } from '../pipelineruns/logs/log-snippet-types';
+import './TaskRunDetails.scss';
 
 export interface TaskRunDetailsProps {
-  obj: K8sResourceKind;
+  obj: K8sResourceKind & { status: TaskRunStatus };
```
andrewballantyne, line 14:
```suggestion
  obj: TaskRunKind;
```

frontend/packages/dev-console/src/components/taskruns/logs/taskRunLogSnippet.ts
```diff
@@ -0,0 +1,30 @@
+import { Condition, PLRTaskRunStep } from '../../../utils/pipeline-augment';
+import { taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
+import { CombinedErrorDetails, TaskRunData } from '../../pipelineruns/logs/log-snippet-types';
+import { taskRunSnippetMessage } from '../../pipelineruns/logs/log-snippet-utils';
+
+export const getTRLogSnippet = (taskRun: TaskRunData): CombinedErrorDetails => {
```
andrewballantyne, line 6:
```suggestion
export const getTRLogSnippet = (taskRun: TaskRunKind): CombinedErrorDetails => {
```

andrewballantyne (approved):
Looks good, just need to use the `TaskRunKind`

karthikjeeyar:
Verified locally, works fine
/lgtm

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
```diff
@@ -0,0 +1,28 @@
...
+  completionTime?: string;
+  conditions: Condition[];
+  podName?: string;
+  startTime?: string;
+  steps?: PLRTaskRunStep[];
+};
+
+export type TaskRunData = K8sResourceKind & {
+  status: TaskRunStatus;
+};
```
andrewballantyne, line 28: Failed to remove this code 😄  Please trim it out so we don't have confusion on which to use.

andrewballantyne:
/approve

Remove the unused code and we should be good to go.

frontend/packages/dev-console/src/components/pipelineruns/logs/log-snippet-types.ts
abhinandan13jan: yup.. updated..thanks

PR #6854, opened 2020-10-06
Add Vulnerabilities tab in project details page

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4882

**Description**: 
1. Show Vulnerabilities tab for each Project
2. Vulnerabilities tab should show list of IMVs (Image Manifest Vulnerability) for each Project
3. Click on IMV should redirect to details view for IMVs

Screenshot 2020-11-03 at 7 06 10 PM
Screenshot 2020-11-03 at 7 06 48 PM

Screenshot 2020-11-03 at 11 25 02 AM

Kapture 2020-11-03 at 19 07 52

Screenshot 2020-10-08 at 6 26 54 PM

**Test setup:**
Install Quay Container Security operator to see the Vulnerabilities.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/badges/EmptyStateResourceBadge.tsx
+++ b/frontend/packages/console-shared/src/components/badges/EmptyStateResourceBadge.tsx
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import { Badge } from '@patternfly/react-core';
+import { K8sKind, kindToAbbr } from '@console/internal/module/k8s';
+import './EmptyStateResourceBadge.scss';
+
+type EmptyStateResourceBadgeProps = {
+  model: K8sKind;
+};
+
+const EmptyStateResourceBadge: React.FC<EmptyStateResourceBadgeProps> = ({ model }) => (
+  <div className="ocs-empty-state-resource-badge">
+    <Badge className="ocs-empty-state-resource-badge--badge" isRead>
+      {model.abbr || kindToAbbr(model.kind)}
+    </Badge>
+  </div>
+);
+
+export default EmptyStateResourceBadge;
--- a/frontend/packages/container-security/src/components/image-manifest-vuln.tsx
+++ b/frontend/packages/container-security/src/components/image-manifest-vuln.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import * as classNames from 'classnames';
-import { Tooltip } from '@patternfly/react-core';
+import { useTranslation } from 'react-i18next';
+import { EmptyState, EmptyStateVariant, Title, Tooltip } from '@patternfly/react-core';
 import { sortable } from '@patternfly/react-table';
 import { SecurityIcon } from '@patternfly/react-icons';
 import {
@@ -13,7 +14,7 @@ import {
   ListPage,
   RowFunction,
 } from '@console/internal/components/factory';
-import { GreenCheckCircleIcon } from '@console/shared/';
+import { EmptyStateResourceBadge, GreenCheckCircleIcon } from '@console/shared/';
 import { referenceForModel, PodKind, ContainerStatus } from '@console/internal/module/k8s';
 import { match } from 'react-router';
 import {
@@ -26,7 +27,6 @@ import {
   Firehose,
   FirehoseResult,
   Loading,
-  MsgBox,
 } from '@console/internal/components/utils';
 import { ChartDonut } from '@patternfly/react-charts';
 import { DefaultList } from '@console/internal/components/default-resource';
@@ -44,6 +44,13 @@ const shortenImage = (img: string) =>
     .slice(1, 3)
     .join('/');
 const shortenHash = (hash: string) => hash.slice(7, 18);
+export const totalCount = (obj: ImageManifestVuln) => {
+  if (!obj.status) return 0;
+  const { highCount = 0, mediumCount = 0, lowCount = 0, unknownCount = 0 } = obj.status;
+  return highCount + mediumCount + lowCount + unknownCount;
+};
+export const affectedPodsCount = (obj: ImageManifestVuln) =>
+  Object.keys(obj.status.affectedPods).length;
 
 export const ImageVulnerabilityRow: React.FC<ImageVulnerabilityRowProps> = (props) => {
   return (
@@ -65,6 +72,7 @@ export const ImageVulnerabilityRow: React.FC<ImageVulnerabilityRowProps> = (prop
 };
 
 export const ImageVulnerabilitiesTable: React.FC<ImageVulnerabilitiesTableProps> = (props) => {
+  const { t } = useTranslation();
   const vulnerabilites = _.sortBy(
     _.flatten(
       props.features.map((feature) =>
@@ -79,11 +87,21 @@ export const ImageVulnerabilitiesTable: React.FC<ImageVulnerabilitiesTableProps>
       <SectionHeading text="Vulnerabilities" />
       <div className="co-m-table-grid co-m-table-grid--bordered">
         <div className="row co-m-table-grid__head">
-          <div className="col-lg-3 col-md-3 col-sm-4 col-xs-6">Vulnerability</div>
-          <div className="col-lg-2 col-md-2 col-sm-5 col-xs-6">Severity</div>
-          <div className="col-lg-2 col-md-2 col-sm-3 hidden-xs">Package</div>
-          <div className="col-lg-2 col-md-2 hidden-sm hidden-xs">Current Version</div>
-          <div className="col-lg-3 col-md-3 hidden-sm hidden-xs">Fixed in Version</div>
+          <div className="col-lg-3 col-md-3 col-sm-4 col-xs-6">
+            {t('container-security~Vulnerability')}
+          </div>
+          <div className="col-lg-2 col-md-2 col-sm-5 col-xs-6">
+            {t('container-security~Severity')}
+          </div>
+          <div className="col-lg-2 col-md-2 col-sm-3 hidden-xs">
+            {t('container-security~Package')}
+          </div>
+          <div className="col-lg-2 col-md-2 hidden-sm hidden-xs">
+            {t('container-security~Current Version')}
+          </div>
+          <div className="col-lg-3 col-md-3 hidden-sm hidden-xs">
+            {t('container-security~Fixed in Version')}
+          </div>
         </div>
         <div className="co-m-table-grid__body">
           {vulnerabilites.map(({ feature, vulnerability }) => (
@@ -101,6 +119,7 @@ export const ImageVulnerabilitiesTable: React.FC<ImageVulnerabilitiesTableProps>
 };
 
 export const ImageManifestVulnDetails: React.FC<ImageManifestVulnDetailsProps> = (props) => {
+  const { t } = useTranslation();
   const total = props.obj.spec.features.reduce((sum, f) => sum + f.vulnerabilities.length, 0);
 
   return (
@@ -122,15 +141,28 @@ export const ImageManifestVulnDetails: React.FC<ImageManifestVulnDetailsProps> =
             />
           </div>
           <div className="imagemanifestvuln-details__summary">
-            <h3>Quay Security Scanner has detected {total} vulnerabilities.</h3>
-            <h4>Patches are available for {props.obj.status.fixableCount} vulnerabilities.</h4>
+            <h3>
+              {t(
+                'container-security~Quay Security Scanner has detected {{total, number}} vulnerabilities.',
+                { total },
+              )}
+            </h3>
+            <h4>
+              {t(
+                'container-security~Patches are available for {{fixableCount, number}} vulnerabilities.',
+                {
+                  fixableCount: props.obj.status.fixableCount,
+                },
+              )}
+            </h4>
             <div className="imagemanifestvuln-details__summary-list">
               {vulnPriority
                 .map((v, k) =>
                   totalFor(k)(props.obj) > 0 ? (
                     <span style={{ margin: '5px' }} key={v.index}>
                       <SecurityIcon color={v.color.value} />
-                      &nbsp;<strong>{totalFor(k)(props.obj)}</strong> {v.title} vulnerabilities.
+                      &nbsp;<strong>{totalFor(k)(props.obj)}</strong>{' '}
+                      {t('container-security~{{title}} vulnerabilities', { title: v.title })}.
                     </span>
                   ) : null,
                 )
@@ -182,6 +214,7 @@ export const AffectedPods: React.FC<AffectedPodsProps> = (props) => {
 export const ImageManifestVulnDetailsPage: React.FC<ImageManifestVulnDetailsPageProps> = (
   props,
 ) => {
+  const { t } = useTranslation();
   return (
     <DetailsPage
       match={props.match}
@@ -198,7 +231,7 @@ export const ImageManifestVulnDetailsPage: React.FC<ImageManifestVulnDetailsPage
         navFactory.editYaml(),
         {
           href: 'pods',
-          name: 'Affected Pods',
+          name: t('container-security~Affected Pods'),
           component: AffectedPods,
         },
       ]}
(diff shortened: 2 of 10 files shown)


vikram-raj:
cc: @openshift/team-devconsole-ux @itsptk

itsptk:
@vikram-raj Looks good. One minor change is that navigating back on the breadcrumb should return the user to the full list of 'Image Manifest Vulnerabilities' (probably with the project selector set to the current project) and not the sub-tab within the Project resource. The breadcrumb is intended to convey resource hierarchy and not necessarily navigational structure. 

http://openshift.github.io/openshift-origin-design/conventions/documentation/interactions.html#breadcrumbs

vikram-raj:
> @vikram-raj Looks good. One minor change is that navigating back on the breadcrumb should return the user to the full list of 'Image Manifest Vulnerabilities' (probably with the project selector set to the current project) and not the sub-tab within the Project resource. The breadcrumb is intended to convey resource hierarchy and not necessarily navigational structure.
> 
> http://openshift.github.io/openshift-origin-design/conventions/documentation/interactions.html#breadcrumbs

@itsptk Isn't it confusing for the user when the user does not go back to the same page by clicking on breadcrumb from where the user came.

frontend/public/components/namespace.jsx
```diff
@@ -70,6 +71,7 @@ import {
 } from './dashboard/project-dashboard/project-dashboard';
 import { removeQueryArgument } from './utils/router';
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import { ContainerSecurityFlag } from '@console/container-security/src/const';
```
invincibleJai, line 74: this import looks bad , we shouldn't import from `container-security` in public

itsptk:
> @itsptk Isn't it confusing for the user when the user does not go back to the same page by clicking on breadcrumb from where the user came from?

The current console convention is that the breadcrumb generally serves the purpose of conveying the hierarchy of a list view to an individual resource, and not necessarily the navigation order.

> In other instances, like when navigating to a pod from a deployment, the breadcrumb changes to reflect the parent of the resource as the resource list (pod list) rather than the originating resource (a deployment).

http://openshift.github.io/openshift-origin-design/conventions/documentation/interactions.html#breadcrumbs

vikram-raj:
@itsptk updated the PR. Now, on clicking the breadcrumbs user are redirected back to the full list page of Vulnerabilities. Updated the GIF. PTAL.

frontend/packages/console-shared/src/components/container-security/ImageManifestVulnListPage.tsx
```diff
@@ -0,0 +1,38 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+import { match as RMatch } from 'react-router';
+import { MultiListPage } from '@console/internal/components/factory';
+import { referenceForModel } from '@console/internal/module/k8s';
+import { ImageManifestVulnModel } from '@console/container-security/src/models';
+import { ImageManifestVulnList } from '@console/container-security/src/components/image-manifest-vuln';
```
invincibleJai, line 7: we shouldn't be importing from `container-security` in `console-shared`

frontend/public/components/namespace.jsx
```diff
@@ -1118,6 +1120,15 @@ export const ProjectsDetailsPage = (props) => (
...
       navFactory.roles(RolesPage),
+      ...(props.flags[ContainerSecurityFlag]
+        ? [
+            {
+              href: 'vulnerabilities',
+              name: 'Vulnerabilities',
+              component: ImageManifestVulnListPage,
+            },
+          ]
+        : []),
```
invincibleJai, line 1131: Do we need this here , can we contribute this via plugin

frontend/public/components/factory/table-filters.ts
```diff
@@ -25,6 +25,7 @@ import {
   silenceState,
 } from '../../reducers/monitoring';
 import { Alert, Rule, Silence } from '../monitoring/types';
+import { ImageManifestVuln } from '@console/container-security';
```
invincibleJai, line 28: should not import from `container-security`

vikram-raj:
@invincibleJai Addressed the review comments. PTAL again.

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -302,28 +302,34 @@ export const ImageManifestVulnList: React.FC<ImageManifestVulnListProps> = (prop
 };
 
 export const ImageManifestVulnPage: React.FC<ImageManifestVulnPageProps> = (props) => {
+  const { showTitle = true, hideNameLabelFilters = true } = props;
+  const namespace = props.namespace || props.match.params.ns || props.match.params.name;
```
sahil143, line 306: @vikram-raj Why are we looking for `props.match.params.name` here?

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -302,28 +302,34 @@ export const ImageManifestVulnList: React.FC<ImageManifestVulnListProps> = (prop
...
   return (
     <MultiListPage
```
sahil143, line 308: We can easily use `ListPage` component here. Since it's only taking one type of resource `ImageManifestVulnModel`. Namespace condition through props and params is handled internally in the `ListPage` component.

sahil143 (commented):
@vikram-raj filter by name doesn't seem to be working 
IV

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
vikram-raj: In the project details page of admin perspective, we get the namespace `props.match.params.name`, in dev perspective, we get the namespace in `props.match.params.ns` whereas in list page we get the namespace in `props.namespace`.

vikram-raj:
> @vikram-raj filter by name doesn't seem to be working

Fixed this.

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
vikram-raj: Even in `ListPage` component, we have to check the namespace condition and pass it. As this component is used in both the perspective and params name is not the same in both the perspective.

sahil143 (commented):
/lgtm

verified changes locally

invincibleJai:
@vikram-raj Thanks looks good , i think we need column with total count as well based on design

image

frontend/packages/container-security/src/plugin.ts
```diff
@@ -148,6 +148,23 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+    type: 'HorizontalNavTab',
+    properties: {
+      model: ProjectModel,
+      page: {
+        name: 'Vulnerabilities',
+        href: 'vulnerabilities',
+      },
+      loader: () =>
+        import(
+          './components/image-manifest-vuln' /* webpackChunkName: "container-security" */
```
invincibleJai, line 161: update `webpackChunkName`

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -244,6 +243,14 @@ export const ImageManifestVulnTableRow: RowFunction<ImageManifestVuln> = ({
         {Object.keys(obj.status.affectedPods).length}
       </TableData>
       <TableData className={tableColumnClasses[4]}>{obj.status.fixableCount || 0}</TableData>
+      <TableData className={tableColumnClasses[4]}>
+        {_.sum([
+          obj.status.highCount,
+          obj.status.mediumCount,
+          obj.status.lowCount,
+          obj.status.unknownCount,
```
invincibleJai, line 251: can destructure it(`highCount` , `mediumCount` etc) above and good to have check for status and check unknownCount and other count is defined as we are doing sum.

and maybe we can avoid lodash

frontend/public/components/factory/table.tsx
```diff
@@ -171,6 +171,13 @@ const sorts = {
   volumeSnapshotSource: (snapshot: VolumeSnapshotKind): string => snapshotSource(snapshot),
   snapshotLastRestore: (snapshot: K8sResourceKind, { restores }) =>
     restores[getName(snapshot)]?.status?.restoreTime,
+  vulnTotalOrder: (imv) =>
+    _.sum([
```
invincibleJai, line 175: same as above

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -44,6 +44,10 @@ const shortenImage = (img: string) =>
     .slice(1, 3)
     .join('/');
 const shortenHash = (hash: string) => hash.slice(7, 18);
+const totalCount = (obj: ImageManifestVuln) => {
+  const { highCount, mediumCount, lowCount, unknownCount } = obj.status;
+  return (highCount || 0) + (mediumCount || 0) + (lowCount || 0) + (unknownCount || 0);
```
invincibleJai, line 49: shall we have check for status as well, see if it make sense

```suggestion
  if(!obj.status) return 0;
  const { highCount =0, mediumCount =0, lowCount =0 , unknownCount =0 } = obj.status;
  return highCount + mediumCount + lowCount + unknownCount;
```

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -244,6 +248,7 @@ export const ImageManifestVulnTableRow: RowFunction<ImageManifestVuln> = ({
         {Object.keys(obj.status.affectedPods).length}
       </TableData>
       <TableData className={tableColumnClasses[4]}>{obj.status.fixableCount || 0}</TableData>
+      <TableData className={tableColumnClasses[5]}>{totalCount(obj)}</TableData>
       <TableData className={tableColumnClasses[4]}>
```
jerolimov, line 252:
```suggestion
      <TableData className={tableColumnClasses[6]}>
```

and can you fill the empty tableColumnClasses with some `tableColumnClasses` content. We have a lot of `const tableColumnClasses` in our code which works better than the current one?! Also if this is not perfect this would be a good step into the right direction, or wdyt?

invincibleJai:
@vikram-raj one general comment , we should start support for i18n wrt changes done as part of this PR.

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
vikram-raj: Added tableColumnClasses.

vikram-raj:
> @vikram-raj one general comment , we should start support for i18n wrt changes done as part of this PR.

Added support for i18n.

jerolimov (approved):
Tested it locally. i18n and responsive design works fine.

/lgtm

vikram-raj:
Resolved merge conflict removed the lgtm label.

invincibleJai:
@vikram-raj i think we need sort ability even for affected pods 

image

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -253,46 +289,68 @@ export const ImageManifestVulnTableRow: RowFunction<ImageManifestVuln> = ({
...
+  const { t } = useTranslation();
+  const EmptyMsg = () => (
+    <EmptyState variant={EmptyStateVariant.large}>
+      <Title headingLevel="h4" size="lg">
+        <div className="imagemanifestvuln-empty-state__badge-wrapper">
+          <Badge className="imagemanifestvuln-empty-state__badge" isRead>
+            {t('container-security~IMV')}
+          </Badge>
+        </div>
+        {t('container-security~No Image Vulnerabilities Found')}
```
invincibleJai, line 342:
```suggestion
        {t('container-security~No Image Vulnerabilities found')}
```

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -122,15 +140,25 @@ export const ImageManifestVulnDetails: React.FC<ImageManifestVulnDetailsProps> =
             />
           </div>
           <div className="imagemanifestvuln-details__summary">
-            <h3>Quay Security Scanner has detected {total} vulnerabilities.</h3>
-            <h4>Patches are available for {props.obj.status.fixableCount} vulnerabilities.</h4>
+            <h3>
+              {t(
+                'container-security~Quay Security Scanner has detected {{total}} vulnerabilities.',
+                { total },
+              )}
```
invincibleJai, line 147: Where we have number can translate that as well, WDYT?
```suggestion
              {t(
                'container-security~Quay Security Scanner has detected {{total, number}} vulnerabilities.',
                { total },
              )}
```

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -253,46 +289,68 @@ export const ImageManifestVulnTableRow: RowFunction<ImageManifestVuln> = ({
...
-  const EmptyMsg = () => <MsgBox title="No Image Vulnerabilities Found" detail="" />;
+  const { t } = useTranslation();
+  const EmptyMsg = () => (
+    <EmptyState variant={EmptyStateVariant.large}>
+      <Title headingLevel="h4" size="lg">
+        <div className="imagemanifestvuln-empty-state__badge-wrapper">
+          <Badge className="imagemanifestvuln-empty-state__badge" isRead>
+            {t('container-security~IMV')}
+          </Badge>
+        </div>
```
invincibleJai, line 341: Shall we extract it and make a generic component, which can be used across can take model or  abbr as prop and i don't think we need to translate this, WDYT?

frontend/packages/console-shared/src/components/badges/EmptyStateResourceBadge.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Badge } from '@patternfly/react-core';
+import { K8sKind, kindToAbbr } from '@console/internal/module/k8s';
+import './EmptyStateResourceBadge.scss';
+
+type EmptyStateResourceBadgeProps = {
+  model: K8sKind;
+};
+
+const EmptyStateResourceBadge: React.FC<EmptyStateResourceBadgeProps> = ({ model }) => {
+  return (
```
invincibleJai, line 11: nit: can do implicit return

frontend/packages/container-security/src/components/image-manifest-vuln.scss
```diff
@@ -11,3 +11,14 @@
...
+.imagemanifestvuln-empty-state {
+  &__badge-wrapper {
+    margin-bottom: var(--pf-global--spacer--md);
+  }
+  &__badge {
+    color: var(--pf-global--palette--white);
+    font-size: var(--pf-global--icon--FontSize--md);
+    background-color: var(--pf-global--Color--400);
+  }
+}
```
invincibleJai, line 24: we may not need this right?

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -122,15 +142,25 @@ export const ImageManifestVulnDetails: React.FC<ImageManifestVulnDetailsProps> =
...
-            <h3>Quay Security Scanner has detected {total} vulnerabilities.</h3>
-            <h4>Patches are available for {props.obj.status.fixableCount} vulnerabilities.</h4>
+            <h3>
+              {t(
+                'container-security~Quay Security Scanner has detected {{total, number}} vulnerabilities.',
+                { total },
+              )}
+            </h3>
+            <h4>
+              {t('container-security~Patches are available for {{fixableCount}} vulnerabilities.', {
```
invincibleJai, line 152: here  can use `number` formatter?

frontend/packages/container-security/src/components/image-manifest-vuln.tsx
```diff
@@ -253,46 +289,67 @@ export const ImageManifestVulnTableRow: RowFunction<ImageManifestVuln> = ({
 
 export const ImageManifestVulnTableHeader = () => [
   {
-    title: 'Image Name',
+    title: i18n.t('container-security~Image Name'),
```
invincibleJai, line 292: global use of `i18n.t` , we should avoid

invincibleJai:
/lgtm
/assign @christianvogt 

Verified the changes looks good

PR #6856, opened 2020-10-06
Bug 1908598: Fix events filter for persitent dashboard

a2batic:
- Activity card component cleanup
- Filters the PVC event only for `volume.beta.kubernetes.io/storage-provisioner: (ceph provisioner)` PVCs

Jira: https://issues.redhat.com/browse/RHSTOR-1106

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
@@ -5,66 +5,85 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
 import DashboardCardBody from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardBody';
 import DashboardCardHeader from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardHeader';
 import DashboardCardTitle from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardTitle';
-import { EventKind, K8sResourceKind } from '@console/internal/module/k8s';
-import { FirehoseResource, FirehoseResult } from '@console/internal/components/utils';
-import { EventModel, PersistentVolumeClaimModel } from '@console/internal/models';
+import {
+  EventKind,
+  K8sResourceKind,
+  PersistentVolumeClaimKind,
+} from '@console/internal/module/k8s';
+import { FirehoseResult } from '@console/internal/components/utils';
+import { PersistentVolumeClaimModel } from '@console/internal/models';
 import ActivityBody, {
   RecentEventsBody,
   OngoingActivityBody,
 } from '@console/shared/src/components/dashboard/activity-card/ActivityBody';
 import { PrometheusResponse } from '@console/internal/components/graphs';
-import { getNamespace } from '@console/shared';
+import {
+  getNamespace,
+  getAnnotations,
+  isCephProvisioner,
+  getName,
+  useDeepCompareMemoize,
+} from '@console/shared';
 import {
   DashboardItemProps,
   withDashboardResources,
 } from '@console/internal/components/dashboard/with-dashboard-resources';
-import { referenceForModel } from '@console/internal/module/k8s/k8s';
-import { SubscriptionModel, SubscriptionKind } from '@console/operator-lifecycle-manager';
-import { CEPH_STORAGE_NAMESPACE, OCS_OPERATOR } from '../../../../constants/index';
+import { SubscriptionKind } from '@console/operator-lifecycle-manager';
+import {
+  CEPH_STORAGE_NAMESPACE,
+  OCS_OPERATOR,
+  PVC_PROVISIONER_ANNOTATION,
+} from '../../../../constants/index';
 import { DATA_RESILIENCY_QUERY, StorageDashboardQuery } from '../../../../constants/queries';
+import {
+  pvcResource,
+  subscriptionResource,
+  storageClusterResource,
+  eventsResource,
+} from '../../../../constants/resources';
 import { getResiliencyProgress } from '../../../../utils';
-import { OCSServiceModel } from '../../../../models';
 import { isClusterExpandActivity, ClusterExpandActivity } from './cluster-expand-activity';
 import { isOCSUpgradeActivity, OCSUpgradeActivity } from './ocs-upgrade-activity';
 import './activity-card.scss';
-
-const eventsResource: FirehoseResource = { isList: true, kind: EventModel.kind, prop: 'events' };
-const subscriptionResource: FirehoseResource = {
-  isList: true,
-  kind: referenceForModel(SubscriptionModel),
-  namespaced: false,
-  prop: 'subs',
-};
-
-const storageClusterResource: FirehoseResource = {
-  isList: true,
-  kind: referenceForModel(OCSServiceModel),
-  namespaced: false,
-  prop: 'storage-cluster',
-};
 
 export const getOCSSubscription = (subscriptions: FirehoseResult): SubscriptionKind => {
   const itemsData: K8sResourceKind[] = subscriptions?.data;
   return _.find(itemsData, (item) => item?.spec?.name === OCS_OPERATOR) as SubscriptionKind;
 };
 
-const ocsEventNamespaceKindFilter = (event: EventKind): boolean => {
-  const eventKind = event?.involvedObject?.kind;
-  const eventNamespace = getNamespace(event);
-  return eventNamespace === CEPH_STORAGE_NAMESPACE || eventKind === PersistentVolumeClaimModel.kind;
-};
-
 const RecentEvent = withDashboardResources(
   ({ watchK8sResource, stopWatchK8sResource, resources }: DashboardItemProps) => {
     React.useEffect(() => {
       watchK8sResource(eventsResource);
+      watchK8sResource(pvcResource);
       return () => {
         stopWatchK8sResource(eventsResource);
+        stopWatchK8sResource(pvcResource);
       };
     }, [watchK8sResource, stopWatchK8sResource]);
+
+    const validPVC = ((resources.pvcs?.data || []) as PersistentVolumeClaimKind[])
+      .filter((obj) => isCephProvisioner(getAnnotations(obj)?.[PVC_PROVISIONER_ANNOTATION]))
+      .map(getName);
+    const memoizedPVCNames = useDeepCompareMemoize(validPVC, true);
+
+    const ocsEventNamespaceKindFilter = (event: EventKind): boolean => {
+      const eventKind = event?.involvedObject?.kind;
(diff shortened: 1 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/constants/index.ts
```diff
@@ -19,6 +19,7 @@ export const ATTACHED_DEVICES_ANNOTATION = 'cluster.ocs.openshift.io/local-devic
 export const DASHBOARD_LINK = '/dashboards/persistent-storage';
 export const AVAILABLE = 'Available';
 export const OSD_REMOVAL_TEMPLATE = 'ocs-osd-removal';
+export const PVC_FILTER_ANNOTATION = 'volume.beta.kubernetes.io/storage-provisioner';
```
afreen23, line 22:
```suggestion
export const PVC_PROVISIONER_ANNOTATION = 'volume.beta.kubernetes.io/storage-provisioner';
```

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -4,68 +4,91 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+        const annotations = getAnnotations(pvcObj);
+        return (
+          annotations &&
+          annotations[PVC_FILTER_ANNOTATION] &&
+          isCephProvisioner(annotations[PVC_FILTER_ANNOTATION])
+        );
+      }
+      return eventNamespace === CEPH_STORAGE_NAMESPACE;
+    };
+
```
afreen23, line 76: It would be better to create a map for pvcName-annotation prior and then utilize in this filter function. Both events and pvc can go high in numbers and you will end up with n^2

afreen23 (changes requested):

cloudbehl:
@a2batic Probably not just for PVC but also the pods that in this namespace.(Pods events in openshift-storage+PVC events)

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -4,68 +4,91 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+    ) : (
+      <div className="co-status-card__alerts-body">
+        <div className="co-status-card__alert-item co-status-card__alert-item--loading">
+          <div className="skeleton-activity__dashboard" />
+          <div className="skeleton-activity__dashboard" />
+          <div className="skeleton-activity__dashboard" />
+          <div className="skeleton-activity__dashboard" />
+          <div className="skeleton-activity__dashboard" />
+        </div>
+      </div>
```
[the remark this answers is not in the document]
rawagner: you could merge `pvcs.loaded` prop with `events.loaded`. Something like

```js
const events = {...resources.events, loaded: resources.events.loaded && resources.pvcs.loaded}
return (
  <RecentEventsBody
        events={events}
        filter={ocsEventNamespaceKindFilter}
      />
)

```

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
rawagner: FYI the filter will show only events for existing PVCs. Its a tradeoff between this or showing events for all PVCs - non ceph ones but includes deleted ones too and only existing ceph PVCs. Is that really what you prefer ?

afreen23: Yes, thats what we want to do here as per [UX](https://issues.redhat.com/browse/RHSTOR-1106?focusedCommentId=14173091&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14173091)

afreen23: The reason being , OCS supports ceph based storage classes only. PV/PVCs created by other storage classes should be filtered out.

afreen23:
> @a2batic Probably not just for PVC but also the pods that in this namespace.(Pods events in openshift-storage+PVC events)

We return the events for openshift-storage namespace here which  covers pods
https://github.com/openshift/console/blob/fdfe2e7a8bb42789b75bb54d610e115b92b7d726/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx#L74

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
a2batic: @afreen23, as per @rawagner's comment, if we keep this filter, the events of deleted PVC will no longer be listed (being Ceph or non-ceph), the Ceph one as well which were appearing before will also disappear as the events won't match any PVC in pvc list, doesn't that take away the purpose of showing events?

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
afreen23: We sorted this offline, commenting for reference here:  Its a tradeoff but we need to have ocs based events only in the dashboard. All events can be looked anytime in console.

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -4,66 +4,98 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+    }, [resources]);
+
+    const ocsEventNamespaceKindFilter = (event: EventKind): boolean => {
+      const eventKind = event?.involvedObject?.kind;
+      const eventNamespace = getNamespace(event);
+      const eventObjectName = event?.involvedObject?.name;
+      if (eventKind === PersistentVolumeClaimModel.kind) {
+        return validPVC().includes(eventObjectName);
+      }
+      return eventNamespace === CEPH_STORAGE_NAMESPACE;
```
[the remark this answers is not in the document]
a2batic: We need to check for `validPVC()` array only for PVC name, we need not check it for every resource.
Also, for `validPVC()`, `useCallback` is used, so number of calls are reduced, not sure we should maintenance a map, as we are already maintaining an array.

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -5,66 +5,99 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+
+    const validPVC = React.useMemo((): string[] => {
+      const pvcList = resources.pvcs
+        ? (resources?.pvcs?.data as PersistentVolumeClaimKind[]).filter((obj) => {
+            const annotations = getAnnotations(obj);
+            return (
+              annotations &&
+              annotations[PVC_PROVISIONER_ANNOTATION] &&
+              isCephProvisioner(annotations[PVC_PROVISIONER_ANNOTATION])
+            );
```
bipuladh, line 67:
```suggestion
        ? (resources?.pvcs?.data as PersistentVolumeClaimKind[]).filter((obj) =>
                         isCephProvisioner(getAnnotations(obj)?.[PVC_PROVISIONER_ANNOTATION])
```

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -5,66 +5,99 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+        pvcList,
+        (res, pvcObj) => {
+          const name = getName(pvcObj);
+          res.push(name);
+          return res;
+        },
+        [],
+      );
+      return pvcNames;
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
bipuladh, line 80:
```suggestion
  return pvcList?.map(getName);
```
I also don't see the need to disable the eslint rule.

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -5,66 +5,99 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+    }, [resources.pvcs]);
```
bipuladh, line 81: `resources.pvcs` is an array of objects, I would suggest you to use `deepCompareMemoize` and use the memoized value of resources. pvcs.  
You could completely eliminate this useMemo block by parsing the pvcList normally and then memoizing the list of PVC names. That should be more efficient. 
```suggestion
    const validPVC = (resources.pvcs?.data || [] as PersistentVolumeClaimKind[]).filter(
    (obj) => isCephProvisioner(getAnnotations(obj)?.[PVC_PROVISIONER_ANNOTATION]))
    .map(getName);
    const memoizedPVCNames = useDeepCompareMemoize(validPVC, true);
```
Use `memoizedPVCNames` everywhere.

frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/activity-card/activity-card.tsx
```diff
@@ -5,66 +5,99 @@ import DashboardCard from '@console/shared/src/components/dashboard/dashboard-ca
...
+
+    const ocsEventNamespaceKindFilter = (event: EventKind): boolean => {
+      const eventKind = event?.involvedObject?.kind;
+      const eventNamespace = getNamespace(event);
+      const eventObjectName = event?.involvedObject?.name;
+      if (eventKind === PersistentVolumeClaimModel.kind) {
+        return validPVC.includes(eventObjectName);
+      }
+      return eventNamespace === CEPH_STORAGE_NAMESPACE;
```
bipuladh, line 90:
```suggestion
      return (eventKind === PersistentVolumeClaimModel.kind) ? validPVC.includes(eventObjectName) : eventNamespace === CEPH_STORAGE_NAMESPACE;
```

bipuladh (changes requested):

bipuladh (approved):
/lgtm

PR #6857, opened 2020-10-06
Bug 1856354: Keep swagger definitions up to date

TheRealJon:
Poll k8s openapi/v2 endpoint every 5 minutes and every time API discovery occurs to ensure swagger definitions stay up to date.

### DIFF
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -222,8 +222,10 @@ graphQLReady.onReady(() => {
   // Global timer to ensure all <Timestamp> components update in sync
   setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000);
 
-  // Fetch swagger on load if it's stale.
+  // Fetch swagger definitions immediately upon application start
   fetchSwagger();
+  // then poll swagger definitions every 5 minutes to ensure they stay up to date
+  setInterval(fetchSwagger, 5 * 60 * 1000);
 
   // Used by GUI tests to check for unhandled exceptions
   window.windowError = null;
--- a/frontend/public/actions/k8s.ts
+++ b/frontend/public/actions/k8s.ts
@@ -11,7 +11,7 @@ import { k8sList, k8sWatch, k8sGet } from '../module/k8s/resource';
 import { makeReduxID } from '../components/utils/k8s-watcher';
 import { APIServiceModel } from '../models';
 import { coFetchJSON } from '../co-fetch';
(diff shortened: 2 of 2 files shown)


frontend/public/components/app.jsx
```diff
@@ -214,8 +214,8 @@ graphQLReady.onReady(() => {
   // Global timer to ensure all <Timestamp> components update in sync
   setInterval(() => store.dispatch(UIActions.updateTimestamps(Date.now())), 10000);
 
-  // Fetch swagger on load if it's stale.
-  fetchSwagger();
+  // Poll swagger definitions to ensure they stay up to date
+  setInterval(fetchSwagger, 30000);
```
spadgett, line 218: Looks like the commit message doesn't match the timeout in the code.

We might scale this back to 1 minute so we're not hitting the server quite as often. Another try would be to wait 15s or 30s after API services change.

TheRealJon: Okay, I updated to do both. Definitions get updated by a poll every minute and every time API discovery occurs.

spadgett: If 10s is enough of a delay to have the doc update, we probably want to increase the poll timeout since it's not so important.

TheRealJon: Yeah I tested 10s with a smaller CRD, but I might want to try creating one that has a heftier OpenAPI definition and see if it still works.

TheRealJon: Okay, tested wit Portworx which has a pretty large OpenAPI definition, and that took 15 to 25 seconds to update. I should probably set this to something higher than 10, but not so high that it won't have a chance to update before users go to create a CRD instance. What sounds reasonable to you @spadgett?

TheRealJon: Actually tested a bit more, and even with Portworx, a 10 second delay seems to work consistently.

TheRealJon:
@spadgett Updated the poll interval to 5 minutes. I kept the delay after API discovery at 10 seconds. It worked consistently with my testing, even with large definitions.

spadgett (approved):
/lgtm

👍

spadgett:
/hold

This is failing CI consistently. We should investigate.

TheRealJon:
/retest

@spadgett  I wanted to let this run again and see if it's still failing. The failures look unusual to me so I just wanted to make sure it wasn't a CI issue over the weekend.

TheRealJon:
/hold cancel

@spadgett Figured out the CI failure, this should be ready

spadgett (approved):
/lgtm

Issue #6858, opened 2020-10-06
Create Secret component allows saving SSH key without trailing newline


willhaines:
If you create an SSH key secret by pasting the contents without an empty line at the end, then git clone will fail to authenticate.

![okd-ssh-key-bug](https://user-images.githubusercontent.com/4226764/95257708-f38be600-07e1-11eb-8b57-3c9708a63ce9.gif)

Simply adding an empty line to the secret contents fixes the problem. IMHO, this would probably be best fixed upstream in SSH, however there is precedent in other projects to simply work around this behaviour by making sure SSH private keys end with a newline character:

https://github.com/jenkinsci/ssh-credentials-plugin/pull/33
https://github.com/ansible-semaphore/semaphore/issues/183

PR #6859, opened 2020-10-06
Append '\n' to ssh private key, if it ends without one.

willhaines:
Source code checkout will fail if an ssh private key is created
without a newline as the last character.

Fixes: #6858


spadgett (approved):
/lgtm

Thanks @willhaines 👍

PR #6862, opened 2020-10-07
Bug 1854567: Dont list Subscrition when matching CSV exists on the Installed Operators page

jhadvig:
The logic here is vice versa, meaning the if the tested **obj**ect is CSV it should be filtered out **OR** if the Subscription 's `currentCSV` and `startingCSV` is in any CSV's `metadata.name`.

/assign @TheRealJon


TheRealJon (approved):
/approve
/lgtm

PR #6865, opened 2020-10-08
Add Node qualifier to Affinity table

glekner:
* Fixed a bug where new affinities didn't have `id`
* Fixed Node count in Affinity Edit, by creating `getIntersectedQualifiedNodes`
* Added ability to `NodeChecker` which can display qualified preferred nodes aside from qualified required nodes. Preferred nodes are always taken out of the required qualified nodes pool.

image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/affinity-modal/affinity-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/affinity-modal/affinity-modal.tsx
@@ -5,6 +5,7 @@ import {
   HandlePromiseProps,
   FirehoseResult,
 } from '@console/internal/components/utils';
+import { getName } from '@console/shared';
 import {
   Button,
   ButtonVariant,
@@ -16,6 +17,8 @@ import {
   EmptyStateVariant,
   Title,
   EmptyStateBody,
+  Stack,
+  StackItem,
 } from '@patternfly/react-core';
 import { ModalTitle, ModalBody, ModalComponentProps } from '@console/internal/components/factory';
 import { NodeModel } from '@console/internal/models';
@@ -30,7 +33,7 @@ import { AFFINITY_MODAL_TITLE, AFFINITY_CREATE, AFFINITY_EDITING } from '../shar
 import { AffinityTable } from './components/affinity-table/affinity-table';
 import { AffinityRow } from './components/affinity-table/affinity-row';
 import { AffinityEdit } from './components/affinity-edit/affinity-edit';
-import { AffinityRowData } from './types';
+import { AffinityCondition, AffinityRowData, AffinityType } from './types';
 import {
   getRowsDataFromAffinity,
   getAffinityFromRowsData,
@@ -38,8 +41,9 @@ import {
   columnClasses,
   getAvailableAffinityID,
 } from './helpers';
+import { useAffinitiesQualifiedNodes } from '../shared/hooks';
+import { NodeChecker } from '../shared/NodeChecker/node-checker';
 import { getAffinityPatch } from '../../../../k8s/patches/vm/vm-scheduling-patches';
-
 import '../shared/scheduling-modals.scss';
 
 export const AffinityModal = withHandlePromise<AffinityModalProps>(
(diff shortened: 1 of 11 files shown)


glekner:
@matthewcarleton can you approve the design above? it is only shown in the affinity table

matthewcarleton:
What if we had a single action that shows the list and then flag the preferred one in the list? Rather than having separate actions for each.

glekner:
image

thoughts @matthewcarleton

matthewcarleton:
What about something like this? Can we show the preferred first and make it a bit more prominent?
Screen Shot 2020-10-08 at 11 57 37 AM

glekner:
Screen Shot 2020-10-11 at 14 41 48
@matthewcarleton

matthewcarleton:
@glekner that's looking nice. Can we order them based on preferred? The only thing I'm wondering is the term "preferred" do we define that anywhere?

glekner:
@matthewcarleton the nodes are now sorted so preferred are first.
preferred is a known term, in scheduling

matthewcarleton:
LGTM!

glekner:
@suomiy can you review

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/hooks.ts
```diff
@@ -70,15 +71,54 @@ export const useNodeQualifier = <T extends IDLabel = IDLabel>(
...
+): NodeKind[] => {
+  const loadedNodes = getLoadedData(nodes, []);
+  const isNodesLoaded = isLoaded(nodes);
+
+  const [qualifiedNodes, setQualifiedNodes] = React.useState([]);
+
+  React.useEffect(() => {
+    if (isNodesLoaded) {
+      const suitableNodes = [];
+      affinities.forEach((aff) => {
```
suomiy, line 98: can you please rather use map here?

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/hooks.ts
```diff
@@ -70,15 +71,54 @@ export const useNodeQualifier = <T extends IDLabel = IDLabel>(
...
+        const expressions = aff?.expressions;
+        const fields = aff?.fields;
+        const affNodes = [];
+        loadedNodes.forEach((node) => {
```
suomiy, line 102: and filter+map here?

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/affinity-modal/affinity-modal.tsx
```diff
@@ -183,17 +229,33 @@ export const AffinityModal = withHandlePromise<AffinityModalProps>(
...
+                      customData={{
+                        isDisabled: false,
+                        vmLikeFinal,
+                        onEdit: onAffinityClickEdit,
+                        onDelete: onAffinityDelete,
+                      }}
+                      row={AffinityRow}
+                    />
+                  </StackItem>
+                  {requiredNodeAffinities?.length > 0 && (
```
suomiy, line 246: so if you have empty required nodes we don't show the matching node table. That is a bit confusing since the VM will schedule alright. Shouldn't we default to all nodes in that case?

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/NodeChecker/node-checker.tsx
```diff
@@ -25,48 +30,67 @@ import './node-checker.scss';
...
 export const NodeChecker: React.FC<NodeCheckerProps> = ({
   qualifiedNodes,
+  qualifiedPerferredNodes,
   wariningTitle,
   warningMessage,
 }) => {
   const size = qualifiedNodes.length;
   const buttonText = pluralize(size, 'Node');
+
+  const isNodePreferred = (node: NodeKind) =>
```
suomiy, line 40: can we rather create a Set here?

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/NodeChecker/node-checker.tsx
```diff
@@ -25,48 +30,67 @@ import './node-checker.scss';
...
-        </Button>
-      </Popover>
+      <Stack>
+        <StackItem>
+          <Popover
+            headerContent={<div>{buttonText} found</div>}
+            position={PopoverPosition.right}
+            className="kv-node-checker__popover"
+            bodyContent={qualifiedNodes
+              .sort((a) => (isNodePreferred(a) ? -1 : 1))
```
suomiy, line 63: we could also sort it by the weight or by the name

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/NodeChecker/node-checker.tsx
glekner: weight is an affinity field, not accessible from the generic node checker.

glekner:
@suomiy applied suggestions.
- created missing types
- node checker shows preferred nodes even if required affinities don't exist

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/shared/NodeChecker/node-checker.tsx
suomiy: ok

frontend/packages/kubevirt-plugin/src/components/modals/scheduling-modals/affinity-modal/affinity-modal.tsx
```diff
@@ -71,6 +75,45 @@ export const AffinityModal = withHandlePromise<AffinityModalProps>(
...
 
+    const requiredNodeAffinities = React.useMemo(
+      () =>
+        affinities?.filter(
+          (aff) => aff?.type === AffinityType.node && aff?.condition === AffinityCondition.required,
+        ),
+      [affinities],
+    );
+
+    const preferredNodeAffinities = React.useMemo(
```
rawagner, line 86: you can have this as a part of first `useMemo`

```
const [requiredNodeAffinities, preferredNodeAffinities] = React.useMemo(() => [affinities?.filter(
          (aff) => aff?.type === AffinityType.node && aff?.condition === AffinityCondition.required,
        ), affinities?.filter(
          (aff) =>
            aff?.type === AffinityType.node && aff?.condition === AffinityCondition.preferred,
        )], [affinities]
```

rawagner (changes requested):

PR #6872, opened 2020-10-08
Bug 1886452: Fix for topology layout options style on select

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4980

**Description**
Update the current layout variable on layout change so the buttons draw selection appropriately.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
+++ b/frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
@@ -4,6 +4,7 @@ import { Button, PageHeaderToolsItem, Tooltip } from '@patternfly/react-core';
 import { TopologyIcon } from '@patternfly/react-icons';
 import {
   TopologyControlBar as PfTopologyControlBar,
+  observer,
   action,
   createTopologyControlButtons,
   defaultControlButtonsOptions,
@@ -15,8 +16,8 @@ interface TopologyControlBarProps {
   visualization: Visualization;
 }
 
-const TopologyControlBar: React.FC<TopologyControlBarProps> = ({ visualization }) => {
-  const layout = visualization.getGraph()?.getLayout() ?? 'COLA_LAYOUT';
+const TopologyControlBar: React.FC<TopologyControlBarProps> = observer(({ visualization }) => {
+  const layout = visualization.getGraph()?.getLayout() ?? COLA_LAYOUT;
   return (
     <span className="pf-topology-control-bar">
       <PfTopologyControlBar
@@ -77,6 +78,6 @@ const TopologyControlBar: React.FC<TopologyControlBarProps> = ({ visualization }
       </PfTopologyControlBar>
     </span>
   );
-};
+});
 
-export default React.memo(TopologyControlBar);
+export default TopologyControlBar;


frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
```diff
@@ -16,7 +16,14 @@ interface TopologyControlBarProps {
 }
 
 const TopologyControlBar: React.FC<TopologyControlBarProps> = ({ visualization }) => {
-  const layout = visualization.getGraph()?.getLayout() ?? 'COLA_LAYOUT';
+  const [layout, setLayout] = React.useState<string>(
+    visualization?.getGraph()?.getLayout() ?? 'COLA_LAYOUT',
```
jerolimov, line 20: I expect you want to use the const here:

```suggestion
    visualization?.getGraph()?.getLayout() ?? COLA_LAYOUT,
```

jerolimov: Oh see now it was previously also this string.. But I expect we should fix this anyway :smirk:

jerolimov:
Tested it locally, works fine and solves the issue. If you can use the const above I would add a lgtm.

jerolimov (changes requested):
Tested it locally, works fine and solves the issue. If you can use the const above I would add a lgtm.

invincibleJai:
Looks good and works as expected!!

jeff-phillips-18:
Updated. Thanks @jerolimov

frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
```diff
@@ -16,7 +16,14 @@ interface TopologyControlBarProps {
...
 
 const TopologyControlBar: React.FC<TopologyControlBarProps> = ({ visualization }) => {
-  const layout = visualization.getGraph()?.getLayout() ?? 'COLA_LAYOUT';
+  const [layout, setLayout] = React.useState<string>(
+    visualization?.getGraph()?.getLayout() ?? COLA_LAYOUT,
+  );
+  React.useEffect(() => {
+    visualization.getGraph().setLayout(layout);
+    visualization.getGraph().layout();
+  }, [layout, visualization]);
```
jerolimov, line 25: Sorry that I missed this in the first round.. This effect was also called after the first rendering and it calls setLayout and layout right after the initial rendering. When seeing this I checked that setLayout [do nothing if the layoutType does not changed](https://github.com/patternfly/patternfly-react/blob/2c4b25ce8aa64a08de38c0a1884c96df643f348e/packages/react-topology/src/elements/BaseGraph.ts#L109-L111).

If the initial layout call here required? Should we ignore this or do we want do this only when the user pressed one of the buttons? :thinking:

christianvogt: Ya we should avoid a call to `layout` here on first render.

Alternative to `useState` would be to just make the control bar an `observer` which will force re-render when layout changes using the old implementation.

jerolimov: Or useState/useRef and call setLayout and layout again in an onClick handler.

jeff-phillips-18: Thanks. Just added the observer and updated to use the constant.

frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
```diff
@@ -15,8 +16,8 @@ interface TopologyControlBarProps {
   visualization: Visualization;
 }
 
-const TopologyControlBar: React.FC<TopologyControlBarProps> = ({ visualization }) => {
-  const layout = visualization.getGraph()?.getLayout() ?? 'COLA_LAYOUT';
+const TopologyControlBar: React.FC<TopologyControlBarProps> = observer(({ visualization }) => {
```
christianvogt, line 19: `observer` makes the component a `memo`. You can remove your call to memo below.

jeff-phillips-18: Fixed. Thanks.

rohitkrai03 (approved):
/lgtm

frontend/packages/dev-console/src/components/topology/TopologyControlBar.tsx
jerolimov: Hmm interesting.. reading the docs.. :)

PR #6873, opened 2020-10-08
Improve secret hostname hint

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4906

**Analysis / Root cause**: 
When starting a pipeline there is an option to add a `Secret` to fetch data from a private git repository. The hint here suggest a server name like "https://github.com", but the pipeline/secret combination works only when the server name is something like "github.com".

**Solution Description**: 
Update the hint from static `The base server url (e.g. https://github.com)` to a dynamic version which is now depending on the selected auth mechanism:

For docker config json: `The base server url (e.g. https://quay.io/)` (new example server)
For http basic auth: `The base server url (e.g. https://github.com)` (unchanged)
For SSH: `Server hostname without schema or path (e.g. github.com)`

Edit: After getting some feedback in this PR, this PR now also changes the order of the input fields to show depending fields below the more generic options.

It also hides some useless "Access To" and "Authentication Type" combinations:

* For "Access To: Git" we only show the auth types "Basic Authentication" and "SSH Key" and dropped the option "Image Registry Credentials"
* For "Access To: Image" we only show the auth types "Basic Authentication" and "Image Registry Credentials" and dropeed the option "SSH Key".

See https://github.com/openshift/console/pull/6873#issuecomment-717912374 and https://github.com/openshift/console/pull/6873#issuecomment-718436148

@openshift/team-devconsole-ux

Before:
old-layout

After:
updated-pr

And some more detail screens:
<details>
<summary>Access to options in now above the auth type</summary>

</details>
<details>
<summary>Docker registry auth type options:</summary>

</details>
<details>
<summary>Docker registry with basic auth selected:</summary>

</details>
<details>
<summary>Docker registry with Image Registry selected:</summary>

</details>
<details>
<summary>Git server auth type options:</summary>

</details>
<details>
<summary>Git server with basic auth selected:</summary>

</details>
<details>
<summary>Git server with SSH Key selected:</summary>

</details>

No unit test changed.

**Test setup:**
* Install OpenShift Pipelines Operator
* Create a new pipeline (e.g. add a deployment and enable the option "Add Pipeline")
* Open the "Start Pipeline" dialog

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
@jerolimov , The Server URL annotation's value can be `https://github.com` (for: https://github.com/test/testRepo) or `github.com` (for ssh url: git@github.com/test/testRepo), It depends on the git url that we use in the pipelines.

jerolimov:
@karthikjeeyar You're absolute right here. We need to make a difference here between http and ssh. I updated this PR (incl the PR description and gif). Wdyt about this change now?

(I didn't tested the dockerconfig way and are not sure if https:// is correct in that case.)

karthikjeeyar:
@jerolimov Thank you for improving the hint based on the Authentication type, this is very useful to the user, but however i have a suggestion to reorder the `Authentication Type` and the `Annotation section`, because i feel changing the previous field's help text will not help the user to identify the value upfront, they might have to revisit the previous field to adjust the value. I think we need UX input here.
Auth_Type_reorder

 @bgliwa01 Let us know your thought on this.

jerolimov:
@karthikjeeyar Yes, I see this exactly like you and had this idea as well, from the code perspective it looks a little bit strange, but I think also this would improve the UX. I reordered the components to get a better feeling for the result:

image
(Not part of this PR yet!)

@karthikjeeyar @bgliwa01 Wdyt? I removed the subtitle "Designate provider to be authenticated" because I was unsure where we want keep it. But I can restore it again if you want this.

jerolimov:
Btw, if the tekton [Authentication at Run Time](https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#configuring-ssh-auth-authentication-for-git) here contains all cases we also should (in this ticket, or in a new one..) hide some combinations here:

image
(screenshot of the doc toc)

* When "Access to" "Git Server" is selected we should hide the "Auth Type" option "Image Registry Credentials"
* When "Access to" "Docker Server" is selected we should hide the "Auth Type" option "SSH auth"

/cc @andrewballantyne @vdemeester wdyt? Should I create a new ticket for this or solve this here as well?

frontend/packages/dev-console/src/components/pipelines/modals/common/SecretAnnotation.tsx
```diff
@@ -1,16 +1,24 @@
...
 type SecretAnnotationParam = {
   fieldName: string;
+  secretType: SecretType;
   isReadOnly?: boolean;
 };
 
+const helpText = {
+  [SecretType.dockerconfigjson]: 'The base server url (e.g. https://quay.io/)',
+  [SecretType.basicAuth]: 'The base server url (e.g. https://github.com)',
+  [SecretType.sshAuth]: 'Server hostname without schema or path (e.g. github.com)',
```
christianvogt, line 17: Localize these.

vdemeester:
> * When "Access to" "Git Server" is selected we should hide the "Auth Type" option "Image Registry Credentials"
> * When "Access to" "Docker Server" is selected we should hide the "Auth Type" option "SSH auth"
> 
> /cc @andrewballantyne @vdemeester wdyt? Should I create a new ticket for this or solve this here as well?

Right, I think it make sense to hide thoses

frontend/packages/dev-console/src/components/pipelines/modals/common/SecretAnnotation.tsx
jerolimov: We have PR https://github.com/openshift/console/pull/7143 by @vikram-raj  for pipeline i18n. I would like to keep it there. Depending on whats merged first we should update the other PR, I'm fine to rebase this one if the other is merged first.

jerolimov:
@christianvogt @vdemeester @karthikjeeyar @andrewballantyne Updated the PR with your feedback,

* it nows also reorders the fields so that you need to select the "Access to" and "Auth type" fields before entering the "Server URL".
* it hides the "Auth type" "Image Registry Credentials" for Git servers and "SSH auth" for Docker registry.

Btw, should we rename Docker registry to Image registry?!?

jerolimov:
@bgliwa01 Thanks for bringing this up in the weekly pipeline sync. Added the following changes now:

* Changed section title from "Create Source Secret" to "Create Secret"
* Restored "Designate provider to be authenticated" as hint for "Access to:"
* Renamed "Docker Registry" to "Image Registry"

updated-pr

karthikjeeyar:
Thanks @jerolimov
/lgtm

bgliwa01 (approved):
lgtm

andrewballantyne:
Perhaps the fields should be emptied (reset to defaults) when you switch the various controls. Anything that is under the control should return to the defaults. I can set a Git URL, and go "oh no wait, I have an SSH key instead!" and switching that will not clear the "https://github.com" url, and thus subtlety not work. I, for one, would not look at a hint for a field that I already entered a value for.

andrewballantyne:
This will need to be rebased on the new pipeline-plugin package. Hopefully it's straight forward as the folder structure should be relatively untouched.

bgliwa01:
> Perhaps the fields should be emptied (reset to defaults) when you switch the various controls. Anything that is under the control should return to the defaults. I can set a Git URL, and go "oh no wait, I have an SSH key instead!" and switching that will not clear the "https://github.com" url, and thus subtlety not work. I, for one, would not look at a hint for a field that I already entered a value for.

+1

jerolimov:
Rebased to the new file structure of the pipeline-plugin

karthikjeeyar:
@jerolimov  needs rebase again :(

jerolimov:
Translation does not contain the components under public/components/secrets and looks now like this:

image

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/SecretForm.tsx
```diff
@@ -48,20 +42,73 @@ const renderSecretForm = (
...
   const [stringData, setStringData] = React.useState({
     [SecretType.basicAuth]: {},
     [SecretType.sshAuth]: {},
     [SecretType.dockerconfigjson]: {},
   });
 
+  const secretTypes = React.useMemo<Record<string, string>>(
+    () => ({
+      [SecretAnnotationId.Git]: t('pipelines-Git Server'),
+      [SecretAnnotationId.Image]: t('pipelines-Image Registry'),
```
vikram-raj, line 73:
```suggestion
      [SecretAnnotationId.Git]: t('pipelines-plugin~pipelines-Git Server'),
      [SecretAnnotationId.Image]: t('pipelines-plugin~pipelines-Image Registry'),
```

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/SecretForm.tsx
```diff
@@ -48,20 +42,73 @@ const renderSecretForm = (
...
+      setFieldValue('type', SecretType.basicAuth);
+    }
+  }, [authTypes, values.type, setFieldValue]);
+
+  // Uses a memo instead of const outside of the function so that we can add i18n right here
+  const helpText = React.useMemo(
+    () => ({
+      [SecretType.dockerconfigjson]: t('pipelines-The base server url (e.g. https://quay.io/)'),
+      [SecretType.basicAuth]: t('pipelines-The base server url (e.g. https://github.com)'),
+      [SecretType.sshAuth]: t('pipelines-Server hostname without schema or path (e.g. github.com)'),
```
vikram-raj, line 107:
```suggestion
      [SecretType.dockerconfigjson]: t('pipelines-plugin~The base server url (e.g. https://quay.io/)'),
      [SecretType.basicAuth]: t('pipelines-plugin~The base server url (e.g. https://github.com)'),
      [SecretType.sshAuth]: t('pipelines-plugin~Server hostname without schema or path (e.g. github.com)'),
```

frontend/packages/dev-console/src/components/pipelines/modals/common/SecretAnnotation.tsx
jerolimov: Now localized

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/SecretForm.tsx
jerolimov: Thanks :+1:

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/SecretForm.tsx
jerolimov: Thanks :+1:

jerolimov:
Thanks @vikram-raj, I added the missing translation namespace.

vikram-raj (commented):
verified it. Thanks @jerolimov 

/lgtm

jerolimov:
> Forgot to hit a review on Friday. @jerolimov You still have the same issue that I noted in my description to Brigid. If you switch the type or access-to it does not reset the value of the Server URL... which could be easily overlooked that the hint changed.

So you want that we clear the server URL or change them as well? Yeah its maybe the safest way.

We could also apply some additional form validation rules here. But let us start with the easiest solution (resetting it)?! Will do that.

andrewballantyne:
> > Forgot to hit a review on Friday. @jerolimov You still have the same issue that I noted in my description to Brigid. If you switch the type or access-to it does not reset the value of the Server URL... which could be easily overlooked that the hint changed.
> 
> So you want that we clear the server URL or change them as well? Yeah its maybe the safest way.
> 
> We could also apply some additional form validation rules here. But let us start with the easiest solution (resetting it)?! Will do that.

Right, the easiest thing is for us to clear (reset back to default) items below the changed field... that way the user is mentally required to re-address those values. I think you'll probably also want to reset the field back to untouched(?) because Formik won't understand what we are doing and will flag the field against the validation schema -- which says "string not empty"... and errors it because we just emptied it haha. I don't think this noise is needed for the user.

jerolimov:
@andrewballantyne @vikram-raj @bgliwa01 I added the following change so that the "Server URL" was now cleared when

* the user switches the "Access to" and this automatically switch the the "Authentication Type"
* the user manually switches "Authentication Type"

image

jerolimov:
> JavaScript heap out of memory

/retest

karthikjeeyar:
Thanks @jerolimov, Verified locally, works fine
/lgtm

andrewballantyne (approved):

PR #6874, opened 2020-10-08
Create buildConfig or pipeline based on user input

divyanshiGupta:
**Story:** https://issues.redhat.com/browse/ODC-3779

- This PR adds the functionality to create pipeline or buildConfig based on user input. If add-pipeline is selected then build configuration advanced option will not be shown. 
- Also if a user adds a pipeline during app creation then on submit the pipeline will now start automatically (pipelineRun gets created on submit) using the created resources. 
- Builds section will not be shown in the sidebar if buildConfig is not present ~~and the app was created using import flows. The latter check is added to make sure that if D/DC was created using CLI/import from yaml then we show the builds section at all times to let the user know that they havent added a buildConfig. Also showing waiting for build alert similar to what is shown in case of buildConfig.~~
- Show pipeline not started icon in build-decorator when pipeline didnt get started on submit.
- Refactored the edit application utilities to get the git and docker specific values from associated pipeline so that the edit flow doesn't break. Also getting the buildStrategy (source/docker) from the associated pipeline since now there can either be a buildConfig or pipeline.

**Gif:**
Peek 2020-10-14 16-45

**Unit test coverage:**
image
image
image

### DIFF
--- a/frontend/packages/console-shared/src/components/status/Status.tsx
+++ b/frontend/packages/console-shared/src/components/status/Status.tsx
@@ -7,6 +7,7 @@ import {
   BanIcon,
   ExclamationTriangleIcon,
   UnknownIcon,
+  OutlinedHourglassIcon,
 } from '@patternfly/react-icons';
 import { DASH } from '../../constants';
 import { YellowExclamationTriangleIcon } from './icons';
@@ -92,6 +93,9 @@ export const Status: React.FC<StatusProps> = ({
 
     case 'Unknown':
       return <StatusIconAndText {...statusProps} icon={<UnknownIcon />} />;
+
+    case 'PipelineNotStarted':
+      return <StatusIconAndText {...statusProps} icon={<OutlinedHourglassIcon />} />;
 
     default:
       return <>{status || DASH}</>;
--- a/frontend/packages/dev-console/src/components/edit-application/EditApplicationPage.tsx
+++ b/frontend/packages/dev-console/src/components/edit-application/EditApplicationPage.tsx
@@ -8,6 +8,7 @@ import { referenceForModel } from '@console/internal/module/k8s';
 import NamespacedPage, { NamespacedPageVariants } from '../NamespacedPage';
 import EditApplication from './EditApplication';
 import { EditApplicationProps } from './edit-application-types';
+import { PipelineModel } from '../../models';
 
 const INSTANCE_LABEL = 'app.kubernetes.io/instance';
 const EditApplicationComponentLoader: React.FunctionComponent<EditApplicationProps> = (
@@ -35,6 +36,13 @@ const EditApplicationPage: React.FunctionComponent<ImportPageProps> = ({ match, 
     {
       kind: 'BuildConfig',
       prop: 'buildConfig',
+      name: appName,
+      namespace,
+      optional: true,
+    },
+    {
+      kind: referenceForModel(PipelineModel),
+      prop: 'pipeline',
       name: appName,
       namespace,
       optional: true,
--- a/frontend/packages/dev-console/src/components/edit-application/EditApplicationPage.tsx
+++ b/frontend/packages/dev-console/src/components/edit-application/EditApplicationPage.tsx
@@ -42,7 +42,7 @@ const EditApplicationPage: React.FunctionComponent<ImportPageProps> = ({ match, 
     },
     {
       kind: referenceForModel(PipelineModel),
-      prop: 'pipeline',
+      prop: PipelineModel.id,
       name: appName,
       namespace,
       optional: true,
--- a/frontend/packages/dev-console/src/components/edit-application/__tests__/edit-application-utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/edit-application/__tests__/edit-application-utils.spec.ts
@@ -7,7 +7,7 @@ import {
   getInitialValues,
   getExternalImagelValues,
 } from '../edit-application-utils';
-import { Resources } from '../../import/import-types';
+import { GitImportFormData, Resources } from '../../import/import-types';
 import {
   knativeService,
   knAppResources,
@@ -29,7 +29,31 @@ describe('Edit Application Utils', () => {
   });
 
   it('getInitialValues should return values based on the resources and the create flow used to create the application', () => {
-    const { route, editAppResource, buildConfig, imageStream } = appResources;
+    const { route, editAppResource, buildConfig, pipeline, imageStream } = appResources;
+    const gitImportValues: GitImportFormData = {
+      ...gitImportInitialValues,
+      git: {
+        ...gitImportInitialValues.git,
+        ref: 'master',
+      },
+      pipeline: {
+        enabled: true,
+      },
+      build: {
+        ...gitImportInitialValues.build,
+        triggers: { config: false, image: false, webhook: false },
+      },
+      image: {
+        ...gitImportInitialValues.image,
+        tag: '',
+      },
+    };
+    expect(
+      getInitialValues({ pipeline, editAppResource, route }, 'nationalparks-py', 'div'),
+    ).toEqual(gitImportValues);
+    expect(
+      getInitialValues({ buildConfig, editAppResource, route }, 'nationalparks-py', 'div'),
+    ).toEqual(gitImportInitialValues);
     expect(
       getInitialValues({ buildConfig, editAppResource, route }, 'nationalparks-py', 'div'),
     ).toEqual(gitImportInitialValues);
--- a/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
+++ b/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
@@ -33,7 +33,10 @@ import {
   GitReadableTypes,
   Resources,
 } from './import-types';
-import { createPipelineRunForImportFlow } from './pipeline/pipeline-template-utils';
+import {
+  createPipelineForImportFlow,
+  createPipelineRunForImportFlow,
+} from './pipeline/pipeline-template-utils';
 import { Perspective } from '@console/plugin-sdk';
 
 export const generateSecret = () => {
@@ -462,7 +465,8 @@ export const createOrUpdateResources = async (
       ),
     );
   } else if (pipeline.template && !dryRun) {
-    requests.push(createPipelineRunForImportFlow(formData));
+    const newPipeline = await createPipelineForImportFlow(formData);
+    requests.push(createPipelineRunForImportFlow(formData, newPipeline));
   }
 
   verb === 'create' && requests.push(createWebhookSecret(formData, 'generic', dryRun));
--- a/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
+++ b/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
@@ -33,7 +33,7 @@ import {
   GitReadableTypes,
   Resources,
 } from './import-types';
-import { createPipelineForImportFlow } from './pipeline/pipeline-template-utils';
+import { createPipelineRunForImportFlow } from './pipeline/pipeline-template-utils';
 import { Perspective } from '@console/plugin-sdk';
 
 export const generateSecret = () => {
@@ -439,6 +439,7 @@ export const createOrUpdateResources = async (
   ) {
     generatedImageStreamName = `${name}-${getRandomChars()}`;
   }
+
   requests.push(
     createOrUpdateImageStream(
       formData,
@@ -448,23 +449,25 @@ export const createOrUpdateResources = async (
       generatedImageStreamName ? 'create' : verb,
       generatedImageStreamName,
     ),
-    createOrUpdateBuildConfig(
-      formData,
-      imageStream,
-      dryRun,
-      _.get(appResources, 'buildConfig.data'),
-      verb,
-      generatedImageStreamName,
-    ),
   );
+  if (!pipeline.enabled) {
+    requests.push(
+      createOrUpdateBuildConfig(
+        formData,
+        imageStream,
+        dryRun,
+        appResources?.buildConfig?.data,
+        verb,
+        generatedImageStreamName,
+      ),
+    );
+  } else if (pipeline.template && !dryRun) {
+    requests.push(createPipelineRunForImportFlow(formData));
+  }
 
   verb === 'create' && requests.push(createWebhookSecret(formData, 'generic', dryRun));
 
   const defaultAnnotations = getGitAnnotations(repository, ref);
-
-  if (pipeline.enabled && pipeline.template && !dryRun) {
-    requests.push(createPipelineForImportFlow(formData));
-  }
 
   if (formData.resources === Resources.KnativeService) {
     // knative service doesn't have dry run capability so returning the promises.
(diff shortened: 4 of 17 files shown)


divyanshiGupta:
@karthikjeeyar we will also need another story for adding the functionality to edit pipeline using the edit app flow because as I mentioned above previously edit flow used to get git data from buildConfig and if the git data was changed then the new values were added to the buildConfig and it was updated but since now if a pipeline is added there will be no buildConfig and hence I am getting git data from the pipeline for the edit flow but as we do not have the functionality for editing the pipeline the changed data is not propagated back to the pipeline.

karthikjeeyar:
@divyanshiGupta When you say git data, is it the url and branch name? If Yes, can we use the Deployemnt/DeploymentConfig/KSVC  resource's metadata annotations like app.openshift.io/vcs-ref (branch Name) , app.openshift.io/vcs-uri (URL) to pull the data. WDYT?

divyanshiGupta:
> @divyanshiGupta When you say git data, is it the url and branch name? If Yes, can we use the Deployemnt/DeploymentConfig/KSVC  resource's metadata annotations like app.openshift.io/vcs-ref (branch Name) , app.openshift.io/vcs-uri (URL) to pull the data. WDYT?

@karthikjeeyar it's not about from where to get the data but if that is changed it should be propagated to all the resources that use that data in someway. For example the associated pipeline has a git URL param and we can get that data from this param but if git URL changes we need the functionality to update the git URL param of the pipeline. Currently we don't have the functionality to update the pipeline using edit flows since we were using buildConfig to build the image but now since we are using pipeline to the build the image we need to have this functionality. It's not just about updating the git params but also we need to decide upon if we want to show the pipeline section and stuff like that and that's where UX input is needed.

frontend/public/components/overview/build-overview.tsx
```diff
@@ -122,18 +122,27 @@ const BuildOverviewList: React.SFC<BuildOverviewListProps> = ({ buildConfig }) =
...
+  if (isCreatedUsingAddFlows && _.isEmpty(buildConfigs)) {
+    return null;
+  }
+
+  return (
+    <div className="build-overview">
+      <SidebarSectionHeading text="Builds" />
+      {_.isEmpty(buildConfigs) ? (
+        <span className="text-muted">No Build Configs found for this resource.</span>
+      ) : (
```
karthikjeeyar, line 138: can this be removed now?, because if we create an app from `cli`  then it will always have buildconfig created and if we create via `webconsole` and associate pipeline then this component will exit with `null` based on the new condition.

frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -49,17 +78,35 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
...
+export const createPipelineRunForImportFlow = async (
+  formData: GitImportFormData,
+  pipeline: Pipeline,
+) => {
+  const {
+    name,
+    project: { name: namespace },
+    git,
+  } = formData;
+  const pipelineInitialValues = {
```
karthikjeeyar, line 100:
```suggestion
  const pipelineInitialValues: StartPipelineFormValues = {
```

frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -49,17 +78,35 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
...
-    await createImageResource(name, namespace);
-  }
-
   return k8sCreate(PipelineModel, template, { ns: namespace });
 };
+
+export const createPipelineRunForImportFlow = async (
+  formData: GitImportFormData,
+  pipeline: Pipeline,
+) => {
```
karthikjeeyar, line 94:
```suggestion
): Promise<PipelineRun> => {
```

frontend/packages/dev-console/src/components/edit-application/EditApplicationPage.tsx
```diff
@@ -39,6 +40,13 @@ const EditApplicationPage: React.FunctionComponent<ImportPageProps> = ({ match,
       namespace,
       optional: true,
     },
+    {
+      kind: referenceForModel(PipelineModel),
+      prop: 'pipeline',
```
karthikjeeyar, line 45: nit:
```suggestion
      prop: PipelineModel.id,
```

frontend/packages/dev-console/src/components/edit-application/__tests__/edit-application-utils.spec.ts
```diff
@@ -29,7 +29,31 @@ describe('Edit Application Utils', () => {
   });
 
   it('getInitialValues should return values based on the resources and the create flow used to create the application', () => {
-    const { route, editAppResource, buildConfig, imageStream } = appResources;
+    const { route, editAppResource, buildConfig, pipeline, imageStream } = appResources;
+    const gitImportValues = {
```
karthikjeeyar, line 33:
```suggestion
    const gitImportValues: GitImportFormData = {
```

karthikjeeyar:
@divyanshiGupta I have created a story [ODC-5018](https://issues.redhat.com/browse/ODC-5018) to capture the improvements around the edit flows, you have mentioned above.

frontend/public/components/overview/build-overview.tsx
divyanshiGupta: > Builds section will not be shown in the sidebar if buildConfig is not present and the app was created using import flows. The latter check is added to make sure that if D/DC was created using CLI/import from yaml then we show the builds section at all times to let the user know that they haven't added a buildConfig.

As I have mentioned in the PR description I have kept this condition since if only a D/DC is created via CLI/Import from yaml we should show the build section so we should not remove this condition. I can do one thing though, since a pipeline can be added to an app using CLI also I can change the above condition `isCreatedUsingAddFlows && _.isEmpty(buildConfigs)` to `pipeline && _.isEmpty(buildConfigs)` so the build section will not be shown only if a pipeline is associated with an app and no buildConfig was found for that app.

karthikjeeyar: Yes, It makes sense to conditionally show BC based on the pipeline associated with or not.

karthikjeeyar:
/lgtm

frontend/public/components/overview/resource-overview-page.tsx
```diff
@@ -17,14 +17,21 @@ const { common } = Kebab.factory;
 export const OverviewDetailsResourcesTab: React.SFC<OverviewDetailsResourcesTabProps> = ({
   item,
 }) => {
-  const { buildConfigs, hpas, routes, services, pods, obj } = item;
+  const { buildConfigs, hpas, routes, services, pods, obj, pipelines } = item;
   const hasBuildConfig = buildConfigs?.length > 0;
+  const hasPipeline = pipelines?.length > 0;
```
christianvogt, line 22: Having pipelines as an indirect dependency here in core is bad because all pipeline knowledge should be controlled by the pipeline UI package (eventually separated to the operator).

Perhaps to simplify the code we can ask UX if we really need to _ever_ show the BuildOverview if there are no associated build configs? I'd much prefer controlling the visibility of the section based on whether or not there is any data to show.

divyanshiGupta: I added the check for pipeline since I didnt wanted to change the existing behavior in console where if a buildConfig is not present the buildSection is shown with an empty state until and unless a pipeline is associated and we do not need the buildSection any more as per our usecase. But if its okay to change this behavior I can simply just check for the buildConfig and show the section based on its availability.

frontend/public/components/overview/pods-overview.tsx
```diff
@@ -128,14 +128,17 @@ export const PodsOverview: React.SFC<PodsOverviewProps> = ({
   allPodsLink,
   emptyText,
   hasBuildConfig,
+  hasPipeline,
```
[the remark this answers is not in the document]
divyanshiGupta: Do we want any UX input on this?

frontend/public/components/overview/build-overview.tsx
```diff
@@ -122,18 +122,27 @@ const BuildOverviewList: React.SFC<BuildOverviewListProps> = ({ buildConfig }) =
...
-    )}
-  </div>
-);
+export const BuildOverview: React.SFC<BuildConfigsOverviewProps> = ({
+  buildConfigs,
+  hasPipeline,
+}) => {
+  if (hasPipeline && _.isEmpty(buildConfigs)) {
+    return null;
+  }
```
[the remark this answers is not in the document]
andrewballantyne: Agreed, this is addressed in the [comments of the Story](https://issues.redhat.com/browse/ODC-3779?focusedCommentId=15105759&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15105759).

We should definitely not be doing Pipeline logic inside the Build. It heavily muddies the water.

frontend/public/components/overview/resource-overview-page.tsx
andrewballantyne: Already have their confirmation in the comments of the story.

[See my other comment on your PR](https://github.com/openshift/console/pull/6874/files#r510293235) for the link.

divyanshiGupta: Oh okay I thought it was just about devconsole. Will remove the check for pipeline.

frontend/public/components/overview/pods-overview.tsx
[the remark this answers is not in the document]
divyanshiGupta: I have removed pipeline check from here as well and we can add the alert as per UX input in the PR for updating the UI with the final UX https://issues.redhat.com/browse/DTUX-736.

frontend/public/components/overview/resource-overview-page.tsx
divyanshiGupta: Done.

bgliwa01 (approved):
lgtm

frontend/packages/console-shared/src/components/status/Status.tsx
```diff
@@ -92,6 +93,9 @@ export const Status: React.FC<StatusProps> = ({
     case 'Unknown':
       return <StatusIconAndText {...statusProps} icon={<UnknownIcon />} />;
 
+    case 'PipelineNotStarted':
+      return <StatusIconAndText {...statusProps} icon={<OutlinedHourglassIcon />} />;
```
andrewballantyne, line 97: Created a Story to update this icon now they have it figured out. https://issues.redhat.com/browse/ODC-5047

frontend/packages/dev-console/src/utils/pipeline-utils.ts
```diff
@@ -306,18 +306,21 @@ export const constructCurrentPipeline = (
   }
 
   const latestRun = getLatestRun({ data: pipelineRuns }, 'creationTimestamp');
-
+  let status: string;
   if (!latestRun) {
-    // Without the latestRun we will not have progress to show
-    return null;
+    status = runStatus.PipelineNotStarted;
```
andrewballantyne, line 311: Nit: You kept the exit early, this could have just been added to the return directly and no other changes needed to be done.

andrewballantyne (changes requested):
Few nits and a change I am unsure I understand. Looks like you were removing support for resources but then added support for resources elsewhere.

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
```diff
@@ -448,24 +452,27 @@ export const createOrUpdateResources = async (
...
+        imageStream,
+        dryRun,
+        appResources?.buildConfig?.data,
+        verb,
+        generatedImageStreamName,
+      ),
+    );
+  } else if (pipeline.template && !dryRun) {
+    const newPipeline = await createPipelineForImportFlow(formData);
+    requests.push(createPipelineRunForImportFlow(formData, newPipeline));
```
andrewballantyne, line 469: Also, if this fails (which I don't think is possible today) we'll need to not cancel the submit flow. I've created another story (https://issues.redhat.com/browse/ODC-5048) to chase this down as it may require hacking up the resource in order to test it. No sense in holding up this PR for potentially a non-starter flow.

frontend/packages/dev-console/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -49,17 +79,35 @@ export const createPipelineForImportFlow = async (formData: GitImportFormData) =
...
+    ...convertPipelineToModalData(pipeline, true),
+    secretOpen: false,
+  };
+  pipelineInitialValues.resources = (pipelineInitialValues.resources || []).map((r) => ({
+    ...r,
+    data: {
+      ...r.data,
+      ...getResourceParams(r.data.type, r.name, name, namespace, git),
+    },
+  }));
```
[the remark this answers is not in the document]
divyanshiGupta: I removed the creation of resources while creating a pipeline and added it to when creating a pipelineRun since we are now automatically starting the pipeline and we need to create the pipelineRun with the required resources and this way I get the resourceRef for the resources easily added to the pipelineRun. Also I think we still need to create the required resources to successfully run the pipeline, isnt it?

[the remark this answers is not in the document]
andrewballantyne: Right, that is a correct assumption to the requirement for the PipelineRun. However, PipelineResources are being moved way from in upstream and all our catalog options use Workspaces and Params to facilitate this area/gap.

So the code you removed above:
```
  if (template.spec.resources?.find((r) => r.type === 'git' && r.name === 'app-source')) {
    await createGitResource(git.url, namespace, git.ref);
  }
  if (template.spec.resources?.find((r) => r.type === 'image' && r.name === 'app-image')) {
    await createImageResource(name, namespace);
  }
```
Was for us to create the PipelineResources based off of the agreed upon names with the Operator resources.

Your `getResourceParams` only manages these said agreed upon PipelineResources. So thus it's a bit confusing removing our creation of the resources (which likely won't be used anymore) and then adding support for said resources.

I think we can just not support PipelineResources anymore in our auto start flow and if it fails to start the PipelineRun, they can go manually add a PipelineResource with the Start Modal.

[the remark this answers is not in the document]
divyanshiGupta: >Your getResourceParams only manages these said agreed upon PipelineResources. So thus it's a bit confusing removing our creation of the resources (which likely won't be used anymore) and then adding support for said resources.

I removed the creation from createPipelineForImportFlow and added it to createPipelineRunForImportFlow since it is required to add either the resourceSpec or resourceRef of the created resources to the PipelineRun so if I created the resources in createPipelineForImportFlow then I wouldnt know the resourceRef of the resources as the resource name is randomly generated and adding resourceSpec is also easier with submitStartPipeline as it handles everything from creating the resource to adding the correct data to the pipelineRun. 

> However, PipelineResources are being moved way from in upstream and all our catalog options use Workspaces and Params to facilitate this area/gap.

I wasnt aware of this so I will remove the creation of resources from our  our auto start flow.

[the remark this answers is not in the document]
andrewballantyne: Ah, that makes sense; not the only way to solve it but definitely the most logical solution. Unfortunately, even if we had still support for PipelineResources upstream, we do not have a lot of support for inline specs anywhere in our UI. So that likely would have run into its own set of issues.

andrewballantyne (approved):
/lgtm
/assign @rohitkrai03 @christianvogt

rohitkrai03 (approved):
/approve

PR #6879, opened 2020-10-08
Bug 1856351: Fix build details page charts

TheRealJon:
Update build details page charts to only show utilization data from when the build ran.

- Create new global time related constants
- Update build details page area charts to only show data from build start to completion
- Minimum queried timepan to one minute
- Show seconds on build charts if timespan is less than 5 minutes
- Force x domain on build charts to match the queried timespan so that chart will not show less than
  one minute on the x-axis.
- Add new xMutator callback arg to graph util `mapLimitsRequests` function to allow for post-processing of
  utilization chart x values
- Add new xMutator and yMutator callback args to graph util `getRangeVectorStats` function to allow
  for post-processing of prometheus response data x and y values
- Remove logic from getRangeVectorStats function that removed second and millisecond data from all x
  values
- Update PrometheusMultilineUtilizationItem to use new xMutator argument to trim seconds and
  milliseconds off of multiline area chart data points.
- Create new global constants for prometheus query default samples and timespan
- Update Area and AreaChart components to use new global prometheus constants and accept domain prop
- Update twentyFourHourTime util function to accept `showSeconds` arg and get rid of
  twentyFourHourTimeWithSeconds function.
- Update usage of twentyFourHourTime/twentyFourHourTimeWithSeconds function where needed

### DIFF
--- a/frontend/public/components/build.tsx
+++ b/frontend/public/components/build.tsx
@@ -5,7 +5,7 @@ import * as classNames from 'classnames';
 import { sortable } from '@patternfly/react-table';
 import { Alert } from '@patternfly/react-core';
 
-import { Status } from '@console/shared';
+import { ONE_HOUR, ONE_MINUTE, Status } from '@console/shared';
 import { ByteDataTypes } from '@console/shared/src/graph-helper/data-utils';
 import {
   K8sResourceKindReference,
@@ -42,6 +42,7 @@ import { BuildLogs } from './build-logs';
 import { ResourceEventStream } from './events';
 import { Area, requirePrometheus } from './graphs';
 import { BuildModel } from '../models';
+import { twentyFourHourTime } from './utils/datetime';
 
 const BuildsReference: K8sResourceKindReference = 'Build';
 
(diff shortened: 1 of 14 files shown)


TheRealJon:
This is likely failing the the analyze test because I am importing moment into the datetime helper file. I did this to simplify the implementation, but it's probably overkill. I'll update to remove this import.

TheRealJon:
I think this is ready for review.

/assign @rohitkrai03 

cc @spadgett

TheRealJon:
/hold

need to fix some unit tests

TheRealJon:
cc @rawagner

frontend/public/components/build.tsx
```diff
@@ -127,40 +128,52 @@ export const BuildNumberLink = ({ build }) => {
 };
 
 const BuildGraphs = requirePrometheus(({ build }) => {
-  const podName = _.get(build, ['metadata', 'annotations', 'openshift.io/build.pod-name']);
+  const podName = build.metadata.annotations?.['openshift.io/build.pod-name'];
   if (!podName) {
     return null;
   }
-
+  const endTime = build?.status?.completionTimestamp
```
kyoto, line 135: We're assuming that `build` is set elsewhere in this function, so I guess we can use `build.` instead of `build.?` everywhere.

frontend/public/components/build.tsx
```diff
@@ -127,40 +128,52 @@ export const BuildNumberLink = ({ build }) => {
...
+    : ONE_HOUR;
+  const timespan = Math.max(runTime, ONE_MINUTE); // Minimum timespan of one minute
   const namespace = build.metadata.namespace;
-
+  const areaProps = {
+    namespace,
+    endTime,
+    timespan,
+    formatDate: (d) => twentyFourHourTime(d, timespan < 5 * ONE_MINUTE),
+    domain: { x: [endTime - timespan, endTime] }, // force domain to match queried timespan
```
kyoto, line 148: I think we're OK in this case, but in general I've had trouble with triggering unnecessary graph re-renders by having function / object / array props that are recreated in the render function like this.

If the graph does not seem to be re-rendering too frequently, I think this is fine, but in general I'm trying to avoid these situations. For example, passing `startTime` and `endTime` integer props instead of a `domain` object might avoid some re-renders, depending on how `BuildGraphs` is used.

frontend/public/components/graphs/area.tsx
```diff
@@ -130,9 +135,12 @@ export const AreaChart: React.FC<AreaChartProps> = ({
             theme={theme}
             scale={{ x: 'time', y: 'linear' }}
             padding={padding}
-            {...(allZero && { domain: { y: [0, 1] } })}
+            domain={{
+              ...(allZero && { y: [0, 1] }),
+              ...domain,
+            }}
```
kyoto, line 141: The could be triggering unnecessary re-renders. If so, `domain` could be memoized.

frontend/public/components/graphs/area.tsx
```diff
@@ -130,9 +135,12 @@ export const AreaChart: React.FC<AreaChartProps> = ({
...
           >
-            {xAxis && <ChartAxis tickCount={tickCount} tickFormat={formatDate} />}
+            {xAxis && <ChartAxis tickCount={tickCount} tickFormat={(tick) => formatDate(tick)} />}
```
kyoto, line 143: Using `(tick) => formatDate(tick)` instead of just `formatDate` might cause `ChartAxis` to re-render unnecessarily too. An easy solution to avoid that possibility would be to define `const xAxisTickFormat = (tick) => formatDate(tick);` outside the component.

kyoto (commented):
Thanks @TheRealJon!

A few comments on the smaller details, but generally LGTM 👍

frontend/public/components/utils/datetime.ts
```diff
@@ -83,15 +83,9 @@ export const parsePrometheusDuration = (duration: string): number => {
...
-};
-
-export const twentyFourHourTimeWithSeconds = (date: Date): string => {
-  const hours = zeroPad(date.getHours());
-  const minutes = zeroPad(date.getMinutes());
-  const seconds = zeroPad(date.getSeconds());
-  return `${hours}:${minutes}:${seconds}`;
+export const twentyFourHourTime = (date: Date, showSeconds?: boolean): string => {
+  const hours = zeroPad(date?.getHours() ?? 0);
+  const minutes = `:${zeroPad(date?.getMinutes() ?? 0)}`;
```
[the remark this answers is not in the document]
TheRealJon: To keep it consistent with `seconds`, I think we should keep the colon as part of `minute`.

frontend/public/components/build.tsx
TheRealJon: I'll pull out `domain` into a separate declaration and memoize it so it will only get re-initialized if `timespan` or `endTime` change.

TheRealJon: I'll also memoize `areaProps` so that it won't cause unnecessary re-renders.

frontend/public/components/graphs/area.tsx
TheRealJon: We can't define this outside the component since `formatDate` is a prop. I'll use the React callback hook to make sure the function definition only changes when necessary.

TheRealJon:
@kyoto Addressed your feedback

frontend/public/components/utils/datetime.ts
```diff
@@ -83,15 +83,9 @@ export const parsePrometheusDuration = (duration: string): number => {
...
-  const minutes = zeroPad(date.getMinutes());
-  return `${hours}:${minutes}`;
-};
-
-export const twentyFourHourTimeWithSeconds = (date: Date): string => {
-  const hours = zeroPad(date.getHours());
-  const minutes = zeroPad(date.getMinutes());
-  const seconds = zeroPad(date.getSeconds());
-  return `${hours}:${minutes}:${seconds}`;
+export const twentyFourHourTime = (date: Date, showSeconds?: boolean): string => {
```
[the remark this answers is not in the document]
TheRealJon: IMO, letting an undefined optional parameter evaluate to false when undefined is okay.  We have been leaning more towards optional args + optional chaining or null coalescence where needed. Mainly because optional chaining and null coalescence catch `null` **and** `undefined`, where  default parameter assignment only works for `undefined`.

kyoto:
Thanks @TheRealJon!

/lgtm

PR #6881, opened 2020-10-09
Monitoring: Fix graph refresh on alert and rule details pages

kyoto:
Pass `query` and `ruleDuration` to `Graph` instead of the whole `rule`
object to prevent some unnecessary re-rendering. This also removes the
need to memoize `queries`.

Memoize `labels` in `AlertsDetailsPage` to avoid triggering unnecessary
re-renders of `Graph`.

Initialize `pollInterval` to `undefined` instead of `null`. `null`
indicates that polling was disabled, whereas `undefined` indicates that
it is not specified and the `QueryBrowser` should use its default poll
interval.

### DIFF
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -382,12 +382,11 @@ const Graph_: React.FC<GraphProps> = ({
   deleteAll,
   filterLabels = undefined,
   formatLegendLabel,
-  patchQuery,
-  rule,
   namespace,
+  patchQuery,
+  query,
+  ruleDuration,
 }) => {
-  const { duration = 0, query = '' } = rule || {};
-
   // Set the query in Redux so that other components like the graph tooltip can access it
   React.useEffect(() => {
     patchQuery(0, { query });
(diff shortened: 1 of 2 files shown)


frontend/public/reducers/ui.ts
```diff
@@ -120,7 +120,7 @@ export default (state: UIState, action: UIAction): UIState => {
       }),
       queryBrowser: ImmutableMap({
         metrics: [],
-        pollInterval: null,
+        pollInterval: undefined,
```
spadgett, line 123: I try to avoid code that differentiates between `undefined` and `null`, which isn't very obvious and can be a source of bugs. I think this at least deserves a comment explaining.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -596,10 +593,15 @@ const alertStateToProps = (state: RootState, { match }): AlertsDetailsPageProps
 export const AlertsDetailsPage = withFallback(
   connect(alertStateToProps)((props: AlertsDetailsPageProps) => {
     const { alert, loaded, loadError, namespace, rule, silencesLoaded } = props;
-    const { annotations = {}, labels = {}, silencedBy = [] } = alert || {};
-    const { alertname, severity } = labels as any;
+    const annotations = alert?.annotations ?? {};
```
spadgett, line 596: You might want `||` instead of `??` in most of these cases. If you ever get `annotations: null`, could that cause an error?

kyoto: Thanks @spadgett. After looking at this again, I ended up refactoring in a way that removes the need for these 2 `??` anyway.

frontend/public/reducers/ui.ts
kyoto: Thanks, the `null` / `undefined` distinction was not a great solution 😬

I'm now thinking that a better approach is to force the polling interval (`pollInterval={Math.round(timespan / 120)}`) instead of using `QueryBrowser`'s default polling interval logic. This removes the need for the `null` / `undefined` change.

kyoto:
@spadgett Rebased and addressed your feedback

spadgett (approved):
/lgtm

PR #6888, opened 2020-10-09
CONSOLE-2439: i18n common components: list view/details components

rebeccaalpert:
Handled list page component, filter toolbar, details page breadcrumbs, resource summary/details item, common details page headings, managed by operator link, and the conditions table.

I'm handling shared tabs (environment, YAML, etc.) in separate PRs.

This PR excludes action menus and kebabs since I'm still debugging a dropdown-related TypeScript issue.

CONSOLE-2439


rebeccaalpert:
Adding a hold label because this will need to merge after the first IBM PR.

frontend/public/components/utils/details-item.tsx
```diff
@@ -96,7 +98,7 @@ export const DetailsItem: React.FC<DetailsItemProps> = ({
             <>
               <SplitItem isFilled />
               <SplitItem>
-                <EditButton onClick={onEdit}>Edit</EditButton>
+                <EditButton onClick={onEdit}>{t('details-page~Edit')}</EditButton>
```
spadgett, line 101: `Edit` should probably go in public

frontend/public/components/filter-toolbar.tsx
```diff
@@ -259,16 +280,16 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
-        setPlaceholder(nameFilterPlaceholder || 'Search by name...');
+        setPlaceholder(nameFilterPlaceholder || t('filter-toolbar~Search by name...'));
         break;
       case 'Label':
-        setPlaceholder(labelFilterPlaceholder || 'Search by label...');
+        setPlaceholder(labelFilterPlaceholder || t('filter-toolbar~Search by label...'));
         break;
       default:
-        setPlaceholder('app=frontend');
+        setPlaceholder(t('filter-toolbar~app=frontend'));
```
spadgett, line 292: This shouldn't be externalized since it's an actual label selector.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -259,16 +280,16 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
   }, []);
 
   const switchFilter = (type: FilterType) => {
-    setFilterType(FilterType[type]);
+    setFilterType(translateFilterType(FilterType[type]));
```
spadgett, line 283: Aren't these internal identifiers that shouldn't be translated?

frontend/public/components/factory/list-page.jsx
```diff
@@ -459,12 +458,12 @@ export const MultiListPage = (props) => {
       autoFocus={autoFocus}
       canCreate={canCreate}
       createAccessReview={createAccessReview}
-      createButtonText={createButtonText || 'Create'}
+      createButtonText={createButtonText || t('list-page~Create')}
```
spadgett, line 461: I'd put this in public.

frontend/public/locales/en/filter-toolbar.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Name": "Name",
+  "Label": "Label",
+  "labels": "labels",
+  "name": "name",
+  "{{value}}": "{{value}}",
```
spadgett, line 6: If the only thing here is the variable, this doesn't need to be externalized.

spadgett (commented):
Thanks @rebeccaalpert

frontend/public/locales/en/filter-toolbar.json
rebeccaalpert: Good catch!

frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -12,10 +12,10 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
   return `/k8s/cluster/${customPlural || match.params.plural}`;
 };
 
-export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () => [
+export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any, t: any) => () => [
```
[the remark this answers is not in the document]
rebeccaalpert: `TFunction` seems to work great! Thanks.

frontend/public/components/filter-toolbar.tsx
rebeccaalpert: If you remove this bit, you wind up with untranslated values in the dropdown (i.e. "Name") when the selection changes.

rebeccaalpert:
Addressed PR feedback; let me know if you have other comments!

frontend/public/components/filter-toolbar.tsx
spadgett: > If you remove this bit, you wind up with untranslated values in the dropdown (i.e. "Name") when the selection changes.

The problem is that we're using the internal identifier as the label in the dropdown. We need two values: an internal identifier and then a mapping to a label that we should in the UI. With this implementation, it will probably break when you switch languages since the value for the wrong language is saved in state.

rebeccaalpert: I'll look into adding those this week.

rebeccaalpert: All set!

rebeccaalpert:
Rebased.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -98,11 +99,30 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
+        return t('filter-toolbar~Label');
+      case 'labels':
+        return t('filter-toolbar~labels');
+      case 'name':
+        return t('filter-toolbar~name');
+      default:
+        return value;
+    }
+  };
+  const translatedFilterType = {
```
spadgett, line 117: I'd call this something like `filterDropdownItems` to be more specific.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -98,11 +99,30 @@ const FilterToolbar_: React.FC<FilterToolbarProps & RouteComponentProps> = (prop
...
+  const translateFilterType = (value: string) => {
+    switch (value) {
+      case 'Name':
+        return t('filter-toolbar~Name');
+      case 'Label':
+        return t('filter-toolbar~Label');
+      case 'labels':
+        return t('filter-toolbar~labels');
+      case 'name':
+        return t('filter-toolbar~name');
```
spadgett, line 112: I'm a little confused why we need both `Name` and `name` and `Label` and `label`. Shouldn't it always be one or the other?

frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -12,10 +13,10 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
   return `/k8s/cluster/${customPlural || match.params.plural}`;
 };
 
-export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () => [
+export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any, t: TFunction) => () => [
```
spadgett, line 16: It's better to not pass `t` around since `useTranslation` can be initialized with different values. I'd just use the global i18n instance in this case.

frontend/public/components/utils/details-page.tsx
```diff
@@ -118,11 +118,11 @@ export const ResourceSummary: React.SFC<ResourceSummaryProps> = ({
               onClick={Kebab.factory.ModifyTolerations(model, resource).callback}
               variant="link"
             >
-              {pluralize(_.size(tolerations), 'Toleration')}
+              {t('details-page~{{count}} toleration', { count: _.size(tolerations) })}
               <PencilAltIcon className="co-icon-space-l pf-c-button-icon--plain" />
             </Button>
           ) : (
-            pluralize(_.size(tolerations), 'Toleration')
+            t('details-page~Toleration', { count: _.size(tolerations) })
```
spadgett, line 125:
```suggestion
            t('details-page~{{count}} Toleration', { count: _.size(tolerations) })
```

frontend/public/components/utils/details-page.tsx
```diff
@@ -118,11 +118,11 @@ export const ResourceSummary: React.SFC<ResourceSummaryProps> = ({
               onClick={Kebab.factory.ModifyTolerations(model, resource).callback}
               variant="link"
             >
-              {pluralize(_.size(tolerations), 'Toleration')}
+              {t('details-page~{{count}} toleration', { count: _.size(tolerations) })}
```
spadgett, line 121: It looks like this uses inconsistent case with the one just below.

frontend/public/components/utils/details-page.tsx
```diff
@@ -140,11 +140,11 @@ export const ResourceSummary: React.SFC<ResourceSummaryProps> = ({
               onClick={Kebab.factory.ModifyAnnotations(model, resource).callback}
               variant="link"
             >
-              {pluralize(_.size(metadata.annotations), 'Annotation')}
+              {t('details-page~{{count}} annotation', { count: _.size(metadata.annotations) })}
               <PencilAltIcon className="co-icon-space-l pf-c-button-icon--plain" />
             </Button>
           ) : (
-            pluralize(_.size(metadata.annotations), 'Annotation')
+            t('details-page~Annotation', { count: _.size(metadata.annotations) })
```
spadgett, line 147: Missing `{{count}}`?

frontend/public/components/utils/details-page.tsx
rebeccaalpert: Mine should output "Toleration" or "Tolerations." If I do `{{count}} Toleration`, it'll be "2 Tolerations," etc.

rebeccaalpert: Good catch; thanks!

frontend/public/components/utils/details-page.tsx
spadgett: This still has inconsistent case with the one below. I assume they should be the same.

rebeccaalpert: Whoops, sorry. Should be all set now @spadgett.

rebeccaalpert:
This is ready for review again @spadgett.

spadgett (approved):
/lgtm
/hold for approvals

yanpzhan:
@rebeccaalpert  There is still no "Operators" nav menu, so code about OLM in the pr could not be tested from console, correct me if I understand wrong.
Screenshot from 2020-11-12 17-10-06

rebeccaalpert:
Hey @yanpzhan - Correct, there's no Operators nav menu. If you want to get around it and test those pages, you can turn pseudolocalization off, navigate to that section, and then turn pseudolocalization back on. To do that, just remove the query parameter, do a hard refresh, and then add the query parameter again when you're in the right place.

rebeccaalpert:
Rebased so I lost the lgtm.

spadgett:
```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/conditions.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/conditions.tsx(2,10):
TS2300: Duplicate identifier 'useTranslation'.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/conditions.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/conditions.tsx(6,10):
TS2300: Duplicate identifier 'useTranslation'.
```

rebeccaalpert:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/victory-shared-events/-/victory-shared-events-34.3.8.tgz: getaddrinfo EAI_AGAIN registry.yarnpkg.com registry.yarnpkg.com:443".
info If you think this is a bug, please open a bug report with the information provided in "/go/src/github.com/openshift/console/frontend/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
+ copyArtifacts
+ '[' -d /tmp/artifacts ']'
+ '[' -d frontend/gui_test_screenshots ']'
error: failed to execute wrapped command: exit status 1
```

rebeccaalpert:
Rebased again.

spadgett (approved):
/lgtm

PR #6891, opened 2020-10-10
CONSOLE-2438: i18n common components: environment tab

rebeccaalpert:
Addressed components associated with environment details page tab.

Need to address jest and integration tests.

[CONSOLE-2438](https://issues.redhat.com/browse/CONSOLE-2438)


frontend/public/components/environment.jsx
```diff
@@ -285,371 +286,384 @@ const stateToProps = ({ k8s, UI }, { obj }) => ({
...
+        {readOnly && !_.isEmpty(owners) && (
+          <div className="co-toolbar__group co-toolbar__group--left">
+            <Alert
+              isInline
+              className="co-alert col-md-11 col-xs-10"
+              variant="info"
+              title={t('environment~Environment variables set from parent')}
+            >
+              {t('environment~View environment for resource')}{' '}
+              {owners.length > 1 ? <>owners: {owners}</> : owners}
```
spadgett, line 531: "owners" should be localized

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -89,17 +88,21 @@ export const NameValueEditor = withDragDropContext(
...
         return (
           <PairElement
             onChange={this._change}
             index={i}
-            nameString={nameString}
-            valueString={valueString}
+            nameString={this.props.nameString}
+            nameKey={!this.props.nameString && t('environment~key')}
+            valueString={this.props.valueString}
+            valueKey={!this.props.valueString && t('environment~value')}
```
spadgett, line 105: I'm unclear why we need both `nameString` and `nameKey`. Can you explain a bit? It would be better to have only one if we can.

`nameKey` is kind of misleading here since it's not the translation key, it's the actual message.

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -294,8 +298,12 @@ export const EnvFromEditor = withDragDropContext(
         <>
           <div className="row pairs-list__heading">
             {!readOnly && <div className="col-xs-1 co-empty__header" />}
-            <div className="col-xs-5 text-secondary text-uppercase">{firstTitle}</div>
-            <div className="col-xs-5 text-secondary text-uppercase">{secondTitle}</div>
+            <div className="col-xs-5 text-secondary text-uppercase">
+              {firstTitle ? firstTitle : t('environment~Config map/secret')}
```
spadgett, line 302: Why do we need defaulting now when we didn't before?

This is shorter as

```suggestion
              {firstTitle || t('environment~Config map/secret')}
```

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -309,7 +317,10 @@ export const EnvFromEditor = withDragDropContext(
                   variant="link"
                   isDisabled={addButtonDisabled}
                 >
-                  <PlusCircleIcon /> {addButtonLabel}
+                  <PlusCircleIcon />{' '}
+                  {addButtonLabel
+                    ? addButtonLabel
+                    : t('environment~Add all from Config Map or Secret')}
```
spadgett, line 323: Why do we need defaulting now when we didn't before?

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -504,7 +510,7 @@ const PairElement = DragSource(
                 <input
                   type="text"
                   className="pf-c-form-control"
-                  placeholder={nameString.toLowerCase()}
+                  placeholder={nameKey ? nameKey : nameString.toLowerCase()}
```
spadgett, line 513:
```suggestion
                  placeholder={nameKey || nameString.toLowerCase()}
```

But it would be better to not have two different ways to pass this string if possible.

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -670,15 +673,15 @@ const EnvFromPairElement = DragSource(
                   data-test-id="env-prefix"
                   type="text"
                   className="pf-c-form-control"
-                  placeholder={valueString.toLowerCase()}
+                  placeholder={valueKey ? valueKey : valueString.toLowerCase()}
```
spadgett, line 676:
```suggestion
                  placeholder={valueKey || valueString.toLowerCase()}
```

frontend/public/components/utils/name-value-editor.jsx
rebeccaalpert: That's a good point; I didn't spend a lot of time on naming this prop. There's some minor string manipulation (lowercasing) that goes on in PairElement, and I was using this prop to avoid the string manipulation for i18n strings. If we avoid doing the string manipulation inside that component, we don't need `nameKey`.

frontend/public/components/utils/name-value-editor.jsx
rebeccaalpert: We had defaulting on line 348. I wasn't sure how to i18n .defaultProps objects, so I went this route.

frontend/public/components/utils/name-value-editor.jsx
rebeccaalpert: We had defaulting previously on 345; I wasn't sure how to i18n in that context so I moved the defaulting into the component.

frontend/public/components/utils/name-value-editor.jsx
spadgett: @rebeccaalpert Can you share what values are being lowercased? Maybe with a screenshot of the context where it's used?

rebeccaalpert: It's in the placeholders (some of which you've commented on). `placeholder={nameKey ? nameKey : nameString.toLowerCase()}` is an example. The other placeholder I looked at as part of i18n was also lowercase. Maybe it's just better to make it a `placeholder` prop?

Screen Shot 2020-10-27 at 3 32 47 PM

Screen Shot 2020-10-27 at 3 33 12 PM

spadgett: If it's only used for a placeholder, yeah, I think it's better to have "placeholder" somewhere in the name. I'm not sure we need to lowercase the values, and probably don't want to since we don't know what the translated string will be.

frontend/public/components/environment.jsx
spadgett: It looks like this comment is still relevant?

frontend/public/components/environment.jsx
```diff
@@ -285,371 +286,386 @@ const stateToProps = ({ k8s, UI }, { obj }) => ({
...
+            </Alert>
+          </div>
+        )}
+        {currentEnvVars.isContainerArray && (
+          <div className="co-toolbar__group co-toolbar__group--left">
+            <div className="co-toolbar__item">
+              {containerType === 'containers'
+                ? t('environment~Container:')
+                : t('environment~Init Container:')}
```
spadgett, line 540:
```suggestion
                : t('environment~Init container:')}
```

frontend/public/components/environment.jsx
```diff
@@ -285,371 +286,386 @@ const stateToProps = ({ k8s, UI }, { obj }) => ({
...
+              )}
+            </h3>
+          )}
+          <NameValueEditorComponent
+            nameValueId={containerIndex}
+            nameValuePairs={envVar[EnvType.ENV]}
+            updateParentData={this.updateEnvVars}
+            nameString={t('environment~Name')}
+            namePlaceholder={t('environment~name')}
+            test="pizza"
```
spadgett, line 568: pizza? :)

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -153,7 +154,7 @@ export const NameValueEditor = withDragDropContext(
                           data-test-id="pairs-list__add-icon"
                           className="co-icon-space-r"
                         />
-                        Add from Config Map or Secret
+                        {t('environment~Add from Config Map or Secret')}
```
spadgett, line 157:
```suggestion
                        {t('environment~Add from ConfigMap or Secret')}
```

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -275,7 +277,8 @@ export const EnvFromEditor = withDragDropContext(
           <EnvFromPairElement
             onChange={this._change}
             index={i}
-            nameString="config map/secret"
+            nameString={t('environment~Config Map/Secret')}
```
spadgett, line 280:
```suggestion
            nameString={t('environment~ConfigMap/Secret')}
```

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -275,7 +277,8 @@ export const EnvFromEditor = withDragDropContext(
...
+            namePlaceholder={t('environment~config map/secret')}
```
spadgett, line 281: I might just skip the placeholder property and use `nameString` as-is without lowercasing. In this instance, we still want `ConfigMap/Secret`.

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -294,8 +297,12 @@ export const EnvFromEditor = withDragDropContext(
         <>
           <div className="row pairs-list__heading">
             {!readOnly && <div className="col-xs-1 co-empty__header" />}
-            <div className="col-xs-5 text-secondary text-uppercase">{firstTitle}</div>
-            <div className="col-xs-5 text-secondary text-uppercase">{secondTitle}</div>
+            <div className="col-xs-5 text-secondary text-uppercase">
+              {firstTitle || t('environment~Config map/secret')}
```
spadgett, line 301:
```suggestion
              {firstTitle || t('environment~ConfigMap/secret')}
```

frontend/public/components/utils/name-value-editor.jsx
```diff
@@ -309,7 +316,8 @@ export const EnvFromEditor = withDragDropContext(
                   variant="link"
                   isDisabled={addButtonDisabled}
                 >
-                  <PlusCircleIcon /> {addButtonLabel}
+                  <PlusCircleIcon />{' '}
+                  {addButtonLabel || t('environment~Add all from Config Map or Secret')}
```
spadgett, line 320:
```suggestion
                  {addButtonLabel || t('environment~Add all from ConfigMap or Secret')}
```

frontend/public/components/environment.jsx
rebeccaalpert: Whoops, sorry. I remember making this change, but must not have committed it.

frontend/public/components/utils/name-value-editor.jsx
rebeccaalpert: Okay, I'll go through and remove it/update the integration tests.

rebeccaalpert:
This is ready for review again @spadgett.

frontend/public/components/utils/value-from-pair.jsx
```diff
@@ -64,14 +50,31 @@ export const NameKeyDropdownPair = ({
...
+  const getHeaders = (configMap, secret, serviceAccount) => {
+    const headers = {};
+    if (configMap && !_.isEmpty(configMap)) {
+      headers[configMap] = t('environment~ConfigMaps');
+    }
+    if (secret && !_.isEmpty(secret)) {
+      headers[secret] = t('environment~Secrets');
+    }
+    if (serviceAccount && !_.isEmpty(serviceAccount)) {
+      headers[serviceAccount] = t('environment~Service Accounts');
```
spadgett, line 64:
```suggestion
      headers[serviceAccount] = t('environment~ServiceAccounts');
```

spadgett (commented):
for approvals

frontend/public/components/utils/value-from-pair.jsx
rebeccaalpert: Updated the string @spadgett.

rebeccaalpert:
`Add annotation without key` is passing locally for me. /retest

spadgett (approved):
/lgtm
/hold

rebeccaalpert:
Rebased!

rebeccaalpert:
Had to rebase i18n.js. Lost the lgtm @spadgett.

spadgett (approved):
/lgtm

PR #6893, opened 2020-10-12
Bug 1872124: Adds a dropdown for device type selection.

vbnrh:
Screenshot from 2020-10-21 17-17-12

### DIFF
--- a/frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
+++ b/frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
+  onChange,
+  placeholder,
+  id,
+  options,
+  defaultSelected = [],
+}) => {
+  const [isOpen, setOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState<string[]>(defaultSelected || []);
+  const onSelect = (event: React.MouseEvent | React.ChangeEvent, selection: string) => {
+    let cSelected: string[] = selected;
+    if (selected.includes(selection)) {
+      cSelected = selected.filter((item) => item !== selection);
+    } else {
+      cSelected = [...selected, selection];
+    }
+    setSelected(cSelected);
+    onChange(cSelected);
+  };
+  const items: JSX.Element[] = options.map((item) => {
+    return <SelectOption key={item} value={item} />;
+  });
+  return (
+    <div>
+      <Select
+        variant={SelectVariant.typeaheadMulti}
+        aria-label="Select Input"
+        onToggle={setOpen}
+        onSelect={onSelect}
+        selections={selected}
+        isOpen={isOpen}
+        placeholderText={placeholder || 'Select options'}
+        aria-labelledby={id}
+      >
+        {items}
+      </Select>
+    </div>
+  );
+};
+
+export type MultiSelectDropdownProps = {
+  id?: string;
+  options?: string[];
+  defaultSelected?: string[];
+  placeholder?: string;
+  onChange: (selected: string[]) => void;
+};
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-request-data.ts
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-request-data.ts
@@ -3,8 +3,24 @@ import { apiVersionForModel, Toleration } from '@console/internal/module/k8s';
 import { LocalVolumeSetModel } from '../../models';
 import { LocalVolumeSetKind, DiskType } from './types';
 import { State } from './state';
-import { DISK_TYPES, HOSTNAME_LABEL_KEY, LABEL_OPERATOR } from '../../constants';
 import { getNodes, getHostNames } from '../../utils';
+import {
+  DISK_TYPES,
+  HOSTNAME_LABEL_KEY,
+  LABEL_OPERATOR,
+  deviceTypeDropdownItems,
+} from '../../constants';
+
+const getDeviceTypes = (deviceType: string[]) => {
+  const { DISK, PART } = deviceTypeDropdownItems;
+  if ((deviceType.includes(DISK) && deviceType.includes(PART)) || deviceType.length === 0) {
+    return [DiskType.RawDisk, DiskType.Partition];
+  }
+  if (deviceType.includes(PART)) {
+    return [DiskType.Partition];
+  }
+  return [DiskType.RawDisk];
+};
 
 export const getLocalVolumeSetRequestData = (
   state: State,
@@ -12,6 +28,7 @@ export const getLocalVolumeSetRequestData = (
   toleration?: Toleration,
 ): LocalVolumeSetKind => {
   const nodes = getNodes(state.showNodesListOnLVS, state.nodeNamesForLVS, state.nodeNames);
+  const deviceTypes = getDeviceTypes(state.deviceType);
   const requestData = {
     apiVersion: apiVersionForModel(LocalVolumeSetModel),
     kind: LocalVolumeSetModel.kind,
@@ -20,7 +37,7 @@ export const getLocalVolumeSetRequestData = (
       storageClassName: state.storageClassName || state.volumeSetName,
       volumeMode: state.diskMode,
       deviceInclusionSpec: {
-        deviceTypes: [DiskType.RawDisk, DiskType.Partition],
+        deviceTypes,
       },
       nodeSelector: {
         nodeSelectorTerms: [
--- a/frontend/packages/local-storage-operator-plugin/src/constants/index.ts
+++ b/frontend/packages/local-storage-operator-plugin/src/constants/index.ts
@@ -18,6 +18,10 @@ export const DISK_TYPES: {
     property: 'Rotational',
   },
 };
+export const deviceTypeDropdownItems = Object.freeze({
+  DISK: 'Disk',
+  PART: 'Part',
+});
 
 export const diskTypeDropdownItems = (t: TFunction) =>
   Object.freeze({
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/state.ts
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/state.ts
@@ -16,6 +16,7 @@ export const initialState: State = {
   showNodesListOnLVS: false,
   diskType: 'All',
   diskMode: diskModeDropdownItems.BLOCK,
+  deviceType: [],
   maxDiskLimit: '',
   nodeNames: [], // nodes selected on the LVS step
   minDiskSize: '0',
@@ -110,6 +111,7 @@ export type State = {
   showNodesListOnLVS: boolean;
   diskType: string;
   diskMode: string;
+  deviceType: string[];
   maxDiskLimit: string;
   nodeNames: string[];
   minDiskSize: string;
@@ -152,6 +154,7 @@ export type Action =
   | { type: 'setStorageClassName'; name: string }
   | { type: 'setShowNodesListOnLVS'; value: boolean }
   | { type: 'setDiskType'; value: string }
+  | { type: 'setDeviceType'; value: string[] }
   | { type: 'setDiskMode'; value: string }
   | { type: 'setMaxDiskLimit'; value: string }
   | { type: 'setNodeNames'; value: string[] }
@@ -203,6 +206,8 @@ export const reducer = (state: State, action: Action) => {
       return Object.assign({}, state, { diskType: action.value });
     case 'setDiskMode':
       return Object.assign({}, state, { diskMode: action.value });
+    case 'setDeviceType':
+      return Object.assign({}, state, { deviceType: action.value });
     case 'setMaxDiskLimit':
       return Object.assign({}, state, { maxDiskLimit: action.value });
     case 'setNodeNames':
--- a/frontend/packages/console-shared/src/utils/index.ts
+++ b/frontend/packages/console-shared/src/utils/index.ts
@@ -16,3 +16,4 @@ export * from './operator-utils';
 export * from './helm-utils';
 export * from './hpa-utils';
 export * from './sample-utils';
+export * from './multiselectdropdown';
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
@@ -18,6 +18,10 @@
   margin-bottom: var(--pf-global--spacer--md);
 }
 
+.lso-create-lvs__device-type-dropdown--margin {
+  margin-bottom: var(--pf-global--spacer--md);
+}
+
 .lso-create-lvs__disk-size-form-group--margin {
   margin: var(--pf-global--spacer--lg) 0;
 }
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
@@ -12,11 +12,16 @@ import {
 import { Dropdown } from '@console/internal/components/utils';
 import { ListPage } from '@console/internal/components/factory';
 import { NodeKind } from '@console/internal/module/k8s';
-import { getName } from '@console/shared';
+import { getName, MultiSelectDropdown } from '@console/shared';
 import { NodeModel } from '@console/internal/models';
 import { NodesSelectionList } from './nodes-selection-list';
 import { State, Action } from './state';
-import { diskModeDropdownItems, diskTypeDropdownItems, diskSizeUnitOptions } from '../../constants';
+import {
+  diskModeDropdownItems,
+  diskTypeDropdownItems,
+  diskSizeUnitOptions,
+  deviceTypeDropdownItems,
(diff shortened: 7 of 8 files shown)


bipuladh (changes requested):

frontend/packages/local-storage-operator-plugin/src/constants/index.ts
```diff
@@ -2,6 +2,13 @@ export const diskModeDropdownItems = Object.freeze({
   BLOCK: 'Block',
   FILESYSTEM: 'Filesystem',
 });
+
+export const deviceTypeDropdownItems = Object.freeze({
+  DISK: 'Disk',
+  PART: 'Part',
+  BOTH: 'Both',
```
cloudbehl, line 9: I'll not have `both` option rather we have had multiple selections dropdown or first value as non selected which means both will be allowed. 

Can you check if deviceType is not provided what happens?

gnehapk: Yes, there should be multi select dropdown instead of having "Both" as option

yuvalgalanti: yap, I agree, multi-selection works

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-request-data.ts
```diff
@@ -5,8 +5,20 @@ import { State } from './state';
 import { LOCAL_STORAGE_NAMESPACE, HOSTNAME_LABEL_KEY, LABEL_OPERATOR } from '../../constants';
 import { getNodes, getHostNames } from '../../utils';
 
+const getDeviceTypes = (deviceType: string) => {
+  if (deviceType === 'Disk') {
```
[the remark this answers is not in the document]
vbnrh: addressed

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-request-data.ts
```diff
@@ -5,8 +5,20 @@ import { State } from './state';
...
+    return [DiskType.RawDisk];
+  }
+
+  if (deviceType === 'Part') {
+    return [DiskType.Partition];
+  }
+
+  return [DiskType.RawDisk, DiskType.Partition];
```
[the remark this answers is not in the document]
vbnrh: changed the code

cloudbehl:
Why do you need `All` checkboxes when you have the option to select both of them?

vbnrh:
> Why do you need `All` checkboxes when you have the option to select both of them?

Initial designs from Yuval suggested to add 'ALL'. I'll be removing it now

yuvalgalanti:
thanks @vbnrh

frontend/public/components/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
```
bipuladh, line 1: I don't see the point in making a generic component out of a patternfly variant. This way we would end up with a lot of generic components in the public folder. I would prefer we use `Select` wherever we need it.

frontend/public/components/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = (props) => {
+  const { onChange } = props;
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState([]);
+  const [placeholder, setPlaceholder] = React.useState(props.placeholder);
```
bipuladh, line 8: I don't see why we need state for placeholder.

frontend/public/components/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+  };
+  const onSelect = (event, selection) => {
+    let cSelected: string[] = selected;
+    if (selected.includes(selection)) {
+      cSelected = selected.filter((item) => item !== selection);
+    } else {
+      cSelected = [...selected, selection];
+    }
+    setSelected(cSelected);
+    setPlaceholder(cSelected.length > 0 ? cSelected.join(', ') : props.placeholder);
```
afreen23, line 20: I think you dont need a separate state for placeholder, you can do the same calcs below and render the placeholder.

afreen23: Having `selected` in state would be enough.

frontend/public/components/utils/multiselectdropdown.tsx
vbnrh: made the changes... please review

frontend/public/components/utils/multiselectdropdown.tsx
vbnrh: Updated.. Do we not need isOpen for the dropdown open/close state ?

yuvalgalanti:
looks great @vbnrh. thanks

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+    setSelected(cSelected);
+    onChange(cSelected);
+  };
+  const items: JSX.Element[] = props.options.map((item) => {
+    return <SelectOption key={0} value={item} />;
+  });
+  return (
+    <div>
+      <span id={props.id} hidden>
```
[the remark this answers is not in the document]
vbnrh: addressed

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = (props) => {
+  const { onChange, placeholder } = props;
```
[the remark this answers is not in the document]
vbnrh: addressed

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState([]);
+  const onToggle = (open) => {
+    setIsOpen(open);
+  };
+  const onSelect = (event, selection) => {
```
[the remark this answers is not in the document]
vbnrh: addressed

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = (props) => {
+  const { onChange, placeholder } = props;
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState([]);
+  const onToggle = (open) => {
```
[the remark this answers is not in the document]
vbnrh: addressed

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = (props) => {
```
a2batic, line 4: Nit
```suggestion
export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({ onChange, placeholder, id }) => {
```

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { onChange, placeholder, id } = props;
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState([]);
```
a2batic, line 7:
```suggestion
  const [selected, setSelected] = React.useState<string[]>();
```

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div>
+      <span id={id} hidden>
+        Checkbox Title
+      </span>
+
+      <Select
+        variant={SelectVariant.typeaheadMulti}
+        aria-label="Select Input"
+        onToggle={onToggle}
```
a2batic, line 33: If we are just setting state, we don't need to define another function.
```suggestion
        onToggle={setIsOpen}
```

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-request-data.ts
```diff
@@ -2,11 +2,27 @@ import { apiVersionForModel } from '@console/internal/module/k8s';
...
+  LOCAL_STORAGE_NAMESPACE,
+  HOSTNAME_LABEL_KEY,
+  LABEL_OPERATOR,
+  deviceTypeDropdownItems,
+} from '../../constants';
 import { getNodes, getHostNames } from '../../utils';
 
+const getDeviceTypes = (deviceType: string[]) => {
+  const { DISK, PART } = deviceTypeDropdownItems;
+  if ((deviceType.includes(DISK) && deviceType.includes(PART)) || deviceType.length === 0) {
```
a2batic, line 15: Since at `deviceType.length === 0` we are taking both `Disk` and `Part`, in multi-select input bar, keep both selected as default, instead of keeping it empty.

vbnrh: fixed. now we can pass the pre default values

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: done

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: done

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: done

vbnrh:
@a2batic @cloudbehl please review

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
+  onChange,
+  placeholder,
+  id,
+  options,
+  defaultSelected,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
```
bipuladh, line 11:
```suggestion
  const [isOpen, setOpen] = React.useState(false);
```

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [selected, setSelected] = React.useState<string[]>(defaultSelected || []);
```
bipuladh, line 12:
```suggestion
  const [selectedItems, setSelectedItems] = React.useState<string[]>(defaultSelected);
```
 `defaultSelected` will not work properly if it's not already set when this component is being rendered if this is something that is gotten from network calls or is evaluated after this component gets rendered problems can arise with this approach. I would suggest moving it to a `useEffect` block and setting the state there.

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
+  onChange,
+  placeholder,
+  id,
+  options,
+  defaultSelected,
```
bipuladh, line 9:
```suggestion
  defaultSelected = [],
```

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+    setSelected(cSelected);
+    onChange(cSelected);
+  };
+  const items: JSX.Element[] = options.map((item) => {
+    return <SelectOption key={0} value={item} />;
+  });
+  return (
+    <div>
+      <span id={id} hidden>
```
bipuladh, line 28:
```suggestion
      <span id={id}>
```
Not sure about hidden being used here. If we are trying to hide this element then why render it? if hiding is necessary we should be doing it via CSS files and not through here.

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onChange,
+  placeholder,
+  id,
+  options,
+  defaultSelected,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState<string[]>(defaultSelected || []);
+  const onSelect = (event: React.MouseEvent | React.ChangeEvent, selection: string) => {
+    let cSelected: string[] = selected;
```
bipuladh, line 14: This closure would cause this function to revaluated every time the value of selected is changed. And since `selected` is an array, I think we should create a memoized function using `useCallback` with `useDeepCompareMemoize`

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return <SelectOption key={0} value={item} />;
+  });
+  return (
+    <div>
+      <span id={id} hidden>
+        Checkbox Title
+      </span>
+
+      <Select
+        variant={SelectVariant.typeaheadMulti}
```
bipuladh, line 33: I would assume most of the different variants would follow a similar logic. Maybe this would be an enhancement but it would be better to have a common component for all types of variants of Select.

bipuladh (changes requested):

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,54 @@
...
+      cSelected = selected.filter((item) => item !== selection);
+    } else {
+      cSelected = [...selected, selection];
+    }
+    setSelected(cSelected);
+    onChange(cSelected);
+  };
+  const items: JSX.Element[] = options.map((item) => {
+    return <SelectOption key={0} value={item} />;
+  });
```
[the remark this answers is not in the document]
cloudbehl: or probably. 
```suggestion
  const items: JSX.Element[] = options.map((item, i) => {
    return <SelectOption key={i} value={item} />;
  });
```

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+
+export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
+  onChange,
```
bipuladh, line 5:
```suggestion
 onSelect
```
Let's keep the names consistent,

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+      cSelected = selected.filter((item) => item !== selection);
+    } else {
+      cSelected = [...selected, selection];
+    }
+    setSelected(cSelected);
+    onChange(cSelected);
+  };
+  const items: JSX.Element[] = options.map((item) => {
+    return <SelectOption key={item} value={item} />;
+  });
```
bipuladh, line 25:
```suggestion
  const items: JSX.Element[] = options.map((item) => <SelectOption key={item} value={item} />;
```

bipuladh (changes requested):

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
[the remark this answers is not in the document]
bipuladh: Indexes should not be used as keys.

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: done

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: done

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: will follow up with this

frontend/packages/console-shared/src/utils/multiselectdropdown.tsx
vbnrh: addressed

bipuladh (approved):
/lgtm
@rawagner tagging you for the final review!

PR #6895, opened 2020-10-12
Query Browser: Remove SeriesButton's key prop

kyoto:
It appears that we no longer need this hack. (I can no longer recreate
the bug that it was fixing.)


spadgett (approved):
/lgtm

PR #6897, opened 2020-10-12
Cleanup unused topology resources

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4970

**Description**
With the change to use the topology list view in the projects workload tab, the OverviewCRD definitions are no longer being used. Removed any of those definitions that still existed.
With the change to use the data reconciler to determine which topology nodes belong to the operator groupings we no longer use the operator workload node. Removed the operator workload node and references to it.

Non-functional change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup


rottencandy (approved):
Tested locally, code changes look good.
/lgtm

rohitkrai03 (approved):
/approve

PR #6900, opened 2020-10-12
Bug 1875516: - disabled scheduling is easy to miss in node page of OCP console

cyril-ui-developer:
/assign @TheRealJon
/cc @spadgett

Overrode `text-muted` style property with  PF `--pf-global--Color--200` and change  `co-status-card__health-item-subtitle`   color to `--pf-global--Color--200` See screenshot attached:

Screen Shot 2020-10-20 at 9 54 23 AM
Screen Shot 2020-10-20 at 9 55 06 AM

### DIFF
--- a/frontend/public/style/_overrides.scss
+++ b/frontend/public/style/_overrides.scss
@@ -497,3 +497,7 @@ button.pf-c-dropdown__menu-item.pf-m-disabled {
 #modal-container .pf-c-backdrop {
   position: absolute !important;
 }
+
+.text-muted{
+  color: var(--pf-global--Color--200);
+}


spadgett:
/assign @rhamilto 
cc @andybraren 

I'd think we either want to use a different style or change `text-muted` everywhere instead of one place. `text-muted` doesn't have enough contrast to meet accessibility guidelines, but it would impact a lot of places in console.

spadgett:
I would probably just use a totally different style for this one instance. @andybraren any recommendation?

andybraren:
I imagine we've only been using `text-muted` against white backgrounds so far - any darker and the text would truly be invisible. 🙂

If that's true then switching from `#9c9c9c` (which fails [WCAG AA](https://webaim.org/resources/contrastchecker/) compliance) to `--pf-global--Color--200` seems relatively low-risk to me, but I agree that checking a few other places would be a good idea.

FYI @mceledonia

mceledonia:
`--pf-global--Color--200` will pass accessibility over the light gray background color as well, but nothing darker than that. It is intended to act as a secondary text color - a lighter text for hierarchical purposes such as this. I think it would work well here!

Our disabled styling colors will not meet accessibility guidelines, and they aren't required to, however I'd like to figure out a way to still meet them while also providing that "disabled" feel.

spadgett:
Sounds like we can override `text-muted` with `--pf-global--Color--200`. @cyril-ui-developer Do you mind trying that and attaching a screenshot?

cyril-ui-developer:
> Sounds like we can override `text-muted` with `--pf-global--Color--200`. @cyril-ui-developer Do you mind trying that and attaching a screenshot?

@spadgett Sure, will do!

cyril-ui-developer:
> This isn't new, but we should try to align the styles of the "Scheduling disabled" text and Health Checks "Not configured" text. They seem to be different sizes and colors.

@spadgett I saw this but thought that Health Checks "Not configured" text style might be unique to it. 
`co-status-card__health-item-subtitle {
    color: #8a8d90;
    font-size: 0.7rem;
}`

Will change it to `text-muted` to align properly. There may be other secondary text in console that this might impact.

spadgett:
We should get input from @andybraren on specific styles... Both of these components are used in different contexts, but we should try to align styles if they show up side by side. Thanks @cyril-ui-developer

andybraren:
Hm, I'm not sure what's cleanest or the most conventional code-wise, but `.co-status-card__health-item-subtitle`'s smaller font size is deliberate and should apply to any substatus within any dashboard's Status card. `Scheduling disabled` being `14px` instead makes it an outlier.

`.text-muted` should still be `--pf-global--Color--200` to improve the legibility of list views, but in dashboard cards maybe tweaking `.co-status-card__health-item-subtitle` to the below is enough... unless we want to make changing the color `.text-muted`'s job and remove the color declaration from here:

```
.co-status-card__health-item-subtitle {
    color: var(--pf-global--Color--200);
    font-size: var(--pf-global--FontSize--xs); // equates to 12px, smaller than the main status
}
```

cyril-ui-developer:
> Hm, I'm not sure what's cleanest or the most conventional code-wise, but `.co-status-card__health-item-subtitle`'s smaller font size is deliberate and should apply to any substatus within any dashboard's Status card. `Scheduling disabled` being `14px` instead makes it an outlier.
> 
> `.text-muted` should still be `--pf-global--Color--200` to improve the legibility of list views, but in dashboard cards maybe tweaking `.co-status-card__health-item-subtitle` to the below is enough... unless we want to make changing the color `.text-muted`'s job and remove the color declaration from here:
> 
> ```
> .co-status-card__health-item-subtitle {
>     color: var(--pf-global--Color--200);
>     font-size: var(--pf-global--FontSize--xs); // equates to 12px, smaller than the main status
> }
> ```

@spadgett I added `font-size: var(--pf-global--FontSize--xs)` to  `text- muted`  to align with health check secondary text.

frontend/public/style/_overrides.scss
```diff
@@ -497,3 +497,8 @@ button.pf-c-dropdown__menu-item.pf-m-disabled {
 #modal-container .pf-c-backdrop {
   position: absolute !important;
 }
+
+.text-muted{
+  color: var(--pf-global--Color--200);
+  font-size: var(--pf-global--FontSize--xs);
```
spadgett, line 503: We don't want to change the font-size all of muted text. We should this style on a lot of text in the console that is different sizes.

spadgett:
> Hm, I'm not sure what's cleanest or the most conventional code-wise, but `.co-status-card__health-item-subtitle`'s smaller font size is deliberate and should apply to any substatus within any dashboard's Status card. `Scheduling disabled` being `14px` instead makes it an outlier.

The issue is that node status with "Scheduling disabled" appears in at least one other context, the node list. I believe we use that status style in other places in console, too.

frontend/public/style/_overrides.scss
```diff
@@ -497,3 +497,8 @@ button.pf-c-dropdown__menu-item.pf-m-disabled {
 #modal-container .pf-c-backdrop {
   position: absolute !important;
 }
+
+.text-muted{
+  color: var(--pf-global--Color--200);
```
spadgett, line 502: We also want to apply this color to `co-status-card__health-item-subtitle`

cyril-ui-developer: > We also want to apply this color to `co-status-card__health-item-subtitle`

@spadgett I have changed the color for `co-status-card__health-item-subtitle` to `var(--pf-global--Color--200)`, and left `text-muted` to inherit parent size. Updated screenshot is attached above. 
Please take a look if this looks good.

spadgett (approved):
/lgtm

PR #6901, opened 2020-10-12
Add graph view to project overview workloads tab

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4925

**Description**
Adds the ability to switch to the graph view in the workloads tab for the project overview page.

**Screen Shots**: 
image

image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

@bgliwa01 @bmignano @serenamarie125

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
+++ b/frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
@@ -1,49 +1,84 @@
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
-import { matchPath, match as RMatch, Link, Redirect } from 'react-router-dom';
-import { Tooltip, Popover, Button } from '@patternfly/react-core';
-import { ListIcon, TopologyIcon, QuestionCircleIcon } from '@patternfly/react-icons';
-import { observer } from '@patternfly/react-topology';
+import { matchPath, match as RMatch } from 'react-router-dom';
 import { useQueryParams } from '@console/shared/src';
-import { Firehose, removeQueryArgument } from '@console/internal/components/utils';
+import { removeQueryArgument, setQueryArgument } from '@console/internal/components/utils';
+import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
 import NamespacedPage, { NamespacedPageVariants } from '../NamespacedPage';
 import ProjectsExistWrapper from '../ProjectsExistWrapper';
 import CreateProjectListPage from '../projects/CreateProjectListPage';
 import { TopologyDataRenderer } from './TopologyDataRenderer';
-import TopologyShortcuts from './TopologyShortcuts';
-import { LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY } from './components/const';
+import { LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY } from './components';
 import { TOPOLOGY_SEARCH_FILTER_KEY } from './filters';
 import DataModelProvider from './data-transforms/DataModelProvider';
-import ModelContext, { ExtensibleModel } from './data-transforms/ModelContext';
+import { TopologyPageToolbar } from './TopologyPageToolbar';
 
 import './TopologyPage.scss';
+import { K8sResourceKind } from '@console/internal/module/k8s';
 
 export interface TopologyPageProps {
   match: RMatch<{
     name?: string;
   }>;
+  activeViewStorageKey?: string;
+  title?: string;
+  hideProjects?: boolean;
 }
 
-const setTopologyActiveView = (id: string) => {
-  localStorage.setItem(LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY, id);
+const setTopologyActiveView = (key: string, id: string) => {
+  localStorage.setItem(key, id);
 };
 
-const getTopologyActiveView = () => {
-  return localStorage.getItem(LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY);
+const getTopologyActiveView = (key: string) => {
+  return localStorage.getItem(key);
 };
 
-export const TopologyPageContext: React.FC<TopologyPageProps> = observer(({ match }) => {
-  const queryParams = useQueryParams();
+export const TopologyPage: React.FC<TopologyPageProps> = ({
+  match,
+  activeViewStorageKey = LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY,
+  title = 'Topology',
+  hideProjects = false,
+}) => {
   const namespace = match.params.name;
-  const dataModelContext = React.useContext<ExtensibleModel>(ModelContext);
-  const showListView = !!matchPath(match.path, {
+  const queryParams = useQueryParams();
+  let view = queryParams.get('view');
+  const { projects } = useK8sWatchResources<{ [key: string]: K8sResourceKind[] }>({
+    projects: { kind: 'Project', isList: true },
+  });
+
+  // Backwards Compatibility
+  const urlView = matchPath(match.path, {
     path: '*/list',
     exact: true,
-  });
-  const showGraphView = !!matchPath(match.path, {
-    path: '*/graph',
-    exact: true,
-  });
+  })
+    ? 'list'
+    : matchPath(match.path, {
+        path: '*/graph',
+        exact: true,
+      })
+    ? 'graph'
+    : null;
+
+  if (urlView && !view) {
+    setQueryArgument('view', urlView);
+    view = urlView;
+  }
+
+  if (!view) {
+    view = getTopologyActiveView(activeViewStorageKey);
+    setQueryArgument('view', view);
+  }
+
+  const showGraphView = view === 'graph';
+
+  const onViewChange = React.useCallback(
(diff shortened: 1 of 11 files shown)


serenamarie125:
FYI @alimobrem @beanh66

serenamarie125:
@jeff-phillips-18 can you verify what is happening with in-context "adds" ?

IMO, if you are in the Project->Workloads(graph) tab and do an in-context add, you would be brought to a form, and then brought back to the Project->Workloads(graph) tab to see the results.  I'm guessing right now it may navigate directly to the Topology page we use in Dev Perspective.

bgliwa01 (approved):
lgtm

serenamarie125 (changes requested):
@jeff-phillips-18 just to be sure, I'm going to include this as a Request for Change

If you are in the Project->Workloads(graph) tab and do an in-context add, you should be brought to a form, and then brought back to the Project->Workloads(graph) tab to see the results. I'm guessing right now it may navigate directly to the Topology page we use in Dev Perspective.

jeff-phillips-18:
@serenamarie125 currently it takes you back to the workloads tab but with the list view selected :(

jeff-phillips-18:
@serenamarie125 fixed to redirect to the previous view for the workloads tab.

debsmita1:
@jeff-phillips-18 In the Project workloads page, I tried adding a database using the Add to Project context menu, on creation it took me to the Template Instance Details page. I was expecting that it would bring me back to the view from where I started

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -125,11 +113,52 @@ export const TopologyPageContext: React.FC<TopologyPageProps> = observer(({ matc
   );
 });
 
-export const TopologyPage: React.FC<TopologyPageProps> = ({ match }) => {
+type TopologyWorkloadsPageProps = {
+  match: any;
```
vikram-raj, line 117:
```suggestion
  match: RMatch<{
    name?: string;
  }>;
```

frontend/public/components/overview/OverviewListPage.tsx
```diff
@@ -1,13 +1,10 @@
 import * as React from 'react';
-import { useParams } from 'react-router-dom';
-import DataModelProvider from '@console/dev-console/src/components/topology/data-transforms/DataModelProvider';
-import { TopologyDataRenderer } from '@console/dev-console/src/components/topology/TopologyDataRenderer';
+import TopologyPage from '@console/dev-console/src/components/topology/TopologyPage';
```
vikram-raj, line 2: This import looks bad. We shouldn't import the dev-console package in public.

frontend/public/components/overview/OverviewListPage.tsx
```diff
@@ -1,13 +1,10 @@
...
 
-export const OverviewListPage: React.FC = () => {
-  const namespace = useParams().name;
-  return (
-    <DataModelProvider namespace={namespace}>
-      <TopologyDataRenderer showGraphView={false} />
-    </DataModelProvider>
-  );
+type OverviewListPageProps = {
+  match: any;
```
vikram-raj, line 5:
```suggestion
  match: RMatch<{
    ns?: string;
  }>;
```

vikram-raj:
There is no padding around the container in the loading state. 
Screenshot 2020-10-21 at 2 56 02 PM

frontend/public/components/overview/OverviewListPage.tsx
jeff-phillips-18: This is known and already existing. I believe the plan is move Topology out of dev-console and into its own mono repo.

christianvogt: If we are to share topology with admin console, we need to move it to a new package as a standalone tool.

jeff-phillips-18:
@vikram-raj Updated PTAL.

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -32,10 +38,17 @@ const getTopologyActiveView = () => {
   return localStorage.getItem(LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY);
 };
 
+const setTopologyWorkloadsActiveView = (id: string) => {
+  localStorage.setItem(LAST_TOPOLOGY_WORKLOADS_VIEW_LOCAL_STORAGE_KEY, id);
+};
+
+const getTopologyWorkloadsActiveView = () => {
+  return localStorage.getItem(LAST_TOPOLOGY_WORKLOADS_VIEW_LOCAL_STORAGE_KEY);
+};
```
christianvogt, line 47: Does UX want a separate setting for the topology page and for the workload page?

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -125,11 +113,54 @@ export const TopologyPageContext: React.FC<TopologyPageProps> = observer(({ matc
...
+export const TopologyPage: React.FC<TopologyPageProps> = ({ match, workloadsPage = false }) => {
   const namespace = match.params.name;
   return (
     <DataModelProvider namespace={namespace}>
-      <TopologyPageContext match={match} />
+      {workloadsPage ? (
+        <TopologyWorkloadsPage match={match} />
+      ) : (
+        <TopologyPageContext match={match} />
+      )}
```
christianvogt, line 163: Both views are exactly the same minus the project selector.
I think they should share a lot more in common between the two pages and come up with a common strategy for the URL handling.

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
```diff
@@ -58,15 +59,9 @@ export const CreateOperand: React.FC<CreateOperandProps> = ({
...
   const { data: crd } = customResourceDefinition;
   const [helpText, setHelpText] = React.useState(FORM_HELP_TEXT);
-  const next =
-    activePerspective === 'dev'
-      ? '/topology'
-      : `${resourcePathFromModel(
-          ClusterServiceVersionModel,
-          match.params.appName,
-          match.params.ns,
-        )}/${match.params.plural}`;
```
christianvogt, line 68: I don't think we can change the location we go to after creating an operand. The user who was in the `installed operators` nav item will now be redirected to the workloads tab even though an operand may not result in a workload being created.

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
jeff-phillips-18: UX and PM agreed to using a single setting.

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
jeff-phillips-18: Removed this change

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
jeff-phillips-18: Updated.

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
jeff-phillips-18: Sorry, missed the thought behind this. We are keeping view preference separate per sections (Workloads vs Topology). UX and PM agreed to that. The saved layout of the graph view is shared between the sections.

serenamarie125:
@alimobrem @beanh66 FYI

jerolimov (approved):
I tested this a while and it works great for me.

/lgtm

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -1,136 +1,133 @@
...
-              <CreateProjectListPage title="Topology">
-                Select a project to view the topology
-              </CreateProjectListPage>
-            )}
-          </ProjectsExistWrapper>
-        </Firehose>
-      </NamespacedPage>
-    </>
+      <div className="odc-topology-page">
+        <NamespaceBar onNamespaceChange={handleNamespaceChange} hideProjects={hideProjects}>
```
christianvogt, line 96: Why the move away from `NamespacedPage`?

jeff-phillips-18: Changed it back. I had tried going in another direction that is no longer relevant.

jerolimov (approved):
Tested it locally.
/lgtm
Happy if @christianvogt will revisit it again :)

rohitkrai03 (changes requested):
@jeff-phillips-18 Going to the workloads tab in a project without any workloads renders the Add page from developer perspective in both list and topology view. Not sure if this is the expected behavior. cc: @serenamarie125 

Peek 2020-11-04 01-15

christianvogt:
@rohitkrai03 yes this is something we need to address with @openshift/team-devconsole-ux as the behavior exists in master already.

jeff-phillips-18:
@jerolimov @christianvogt Updated `TopologyPage` to fix tests

PR #6904, opened 2020-10-12
Remove notification drawer header override fix

rhamilto:
We previously added this override fix to correct for an upstream bug in PatternFly.  The upstream bug was fixed, but the override did not get removed.

Before:
Screen Shot 2020-10-12 at 2 06 07 PM

After:
Screen Shot 2020-10-12 at 2 04 07 PM


spadgett (approved):
/lgtm

Thank you

PR #6906, opened 2020-10-12
Bug 1886524: Change default terminal command for Windows Pods

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1886524

WaNxdzBb9T

Thanks to @PratikMahajan for setting up a Windows pod!

### DIFF
--- a/frontend/public/components/pod-exec.jsx
+++ b/frontend/public/components/pod-exec.jsx
@@ -47,10 +47,16 @@ export const PodExec = connectToFlags(FLAGS.OPENSHIFT)(
     }
 
     connect_() {
-      const { metadata } = this.props.obj;
+      const {
+        metadata,
+        spec: { tolerations },
+      } = this.props.obj;
       const { activeContainer } = this.state;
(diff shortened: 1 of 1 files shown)


frontend/public/components/pod-exec.jsx
```diff
@@ -47,10 +47,16 @@ export const PodExec = connectToFlags(FLAGS.OPENSHIFT)(
...
+      const {
+        metadata,
+        spec: { tolerations },
+      } = this.props.obj;
       const { activeContainer } = this.state;
       const usedClient = this.props.flags[FLAGS.OPENSHIFT] ? 'oc' : 'kubectl';
-
+      const isWindows = _.some(tolerations, (t) => {
+        return t.key === 'os' && t.value === 'Windows';
+      });
```
spadgett, line 58: nit: Prefer the native `Array.some`
```suggestion
      const isWindows = tolerations?.some((t) => {
        return t.key === 'os' && t.value === 'Windows';
      });
```

spadgett (approved):
/lgtm

PR #6907, opened 2020-10-12
Bug 1874968: Pressing escape key doesn't exit the dropdown list

harishsurf:
Bug 1874968: Keyboard shortcut to exit dropdown doesn't work

### DIFF
--- a/frontend/public/components/utils/dropdown.jsx
+++ b/frontend/public/components/utils/dropdown.jsx
@@ -233,7 +233,13 @@ export class Dropdown extends DropdownMixin {
     this.onKeyDown = (e) => this.onKeyDown_(e);
     this.changeTextFilter = (e) => this.applyTextFilter_(e.target.value, this.props.items);
     const { shortCut } = this.props;
+
     this.globalKeyDown = (e) => {
+      if (e.key === 'Escape' && this.state.active) {
+        this.hide(e);
+        return;
+      }
+
       const { nodeName } = e.target;
 
       if (nodeName === 'INPUT' || nodeName === 'TEXTAREA') {
@@ -300,10 +306,6 @@ export class Dropdown extends DropdownMixin {
 
   onKeyDown_(e) {
     const { key } = e;
-    if (key === 'Escape') {
-      this.hide(e);
-      return;
-    }
 
     if (key !== 'ArrowDown' && key !== 'ArrowUp' && key !== 'Enter') {
       return;


frontend/public/components/utils/dropdown.jsx
```diff
@@ -186,6 +187,7 @@ class DropDownRow extends React.PureComponent {
             focus: selected,
           })}
           onClick={(e) => onclick(itemKey, e)}
+          onKeyDown={(e) => onKeyDown(e)}
```
spadgett, line 190: nit: this can be

```suggestion
          onKeyDown={onKeyDown}
```

frontend/public/components/utils/dropdown.jsx
```diff
@@ -233,7 +233,13 @@ export class Dropdown extends DropdownMixin {
     this.onKeyDown = (e) => this.onKeyDown_(e);
     this.changeTextFilter = (e) => this.applyTextFilter_(e.target.value, this.props.items);
     const { shortCut } = this.props;
+
     this.globalKeyDown = (e) => {
+      if (e.key === 'Escape' && this.state.active) {
```
spadgett, line 238: This fixes one problem, but the accessibility issues seem to run a little deeper. You can tab outside of the project dropdown with it still open. We should really be using a focus trap and the tab key should cycle through the items in the dropdown, with escape closing. I shouldn't be able to tab outside the dropdown to other items on the page when it's open.

cc @jessiehuff

harishsurf: @jessiehuff, in the link you shared above, using the up/down arrow keys don't seem to be traversing through the dropdown list. Only the tab key does that. Do we want to have similar behavior for this dropdown?

jessiehuff: I would say that the ideal keyboard interaction would be to be able to tab to the search input and the create project button, but once focus leaves “Create Project” it should land on the options menu as a whole. Hitting the Tab key again would take the user back to the beginning of the panel, to the search input. While focus is on the options menu, the recommendation is to use arrow keys to manage focus within the menu: down/up arrow keys to traverse through the list items, right/left arrow keys to get between the item options. The reason for this is based on the [WAI fundamental keyboard navigation conventions](https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_generalnav): 

> “A primary keyboard navigation convention common across all platforms is that the tab and shift+tab keys move focus from one UI component to another while other keys, primarily the arrow keys, move focus inside of components that include multiple focusable elements.“

The menu itself could be considered a UI component on its own and then within that UI component is when you would use the arrow keys. At least according to the WAI ideal. However, being able to tab through all items instead of using arrow keys is still accessible so what you have here with just Tab is an ok alternative. We’ve had intentions to update our context selector to follow that pattern and become more ideal as well, but given our recent commitments, it hasn’t been as high of a priority. The [typeahead select](https://www.patternfly.org/v4/components/select#favorites) is likely the most similar to the previous interaction mentioned specifically with the favoriting/bookmarking feature. For that component, you’re able to cycle through the items with the up/down arrow keys, then you press tab to switch focus into the select before you can use the up/down and right/left arrow key interaction that I’m referring to here. So it is a little bit different in how to get to the arrow key functionality, but that was necessary given the typeahead functionality of being able to select from the options as you’re typing.

harishsurf: Alright. I have created this [Jira](https://issues.redhat.com/browse/CONSOLE-2468) to work on this. For now, this PR will be a temporary fix

spadgett (approved):
/lgtm

PR #6914, opened 2020-10-13
Query Browser: All pressing Escape key to cancel zoom

kyoto:
Once you started a zoom action (by selecting an area of the graph with
the mouse), there was no way to cancel the action. This makes it
possible to cancel by pressing the Escape key.

### DIFF
--- a/frontend/public/components/monitoring/_monitoring.scss
+++ b/frontend/public/components/monitoring/_monitoring.scss
@@ -497,6 +497,9 @@ $monitoring-line-height: 18px;
 .query-browser__zoom {
   cursor: ew-resize;
   position: relative;
+  &:focus {
+    outline: none;
+  }
 }
 
 .query-browser__zoom-overlay {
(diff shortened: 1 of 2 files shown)


frontend/public/components/monitoring/_monitoring.scss
```diff
@@ -497,6 +497,9 @@ $monitoring-line-height: 18px;
 .query-browser__zoom {
   cursor: ew-resize;
   position: relative;
+  &:focus {
+    outline: none;
```
spadgett, line 501: I'm not sure we want to remove the focus indicator. That's generally bad for accessibility. How does it look without this style?

kyoto: I do think removing the focus outline looks better and I'm not sure that the focus styling is actually beneficial for accessibility in this case?

With focus outline:
screenshot

Without focus outline:
screenshot

spadgett: The problem is, if you're tabbing through elements on the page, the focus ring will disappear when you get once you get to the graph, and you won't know where you are on the page.

kyoto: I think we're OK here. Setting `tabIndex` to `-1` prevents selecting the element with the keyboard.

From https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex:
> A negative value (usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation, but could be focused with Javascript or visually by clicking with the mouse.

spadgett (approved):
/lgtm

PR #6917, opened 2020-10-13
Changes to list page to support multiple table filters in the same page.

bipuladh:
Make ResourceLink to default title to Name if none provided

This feature is required for pages that have two tables in a single view. 

/assign @spadgett


bipuladh:
/close
Handled as part of a different PR

PR #6918, opened 2020-10-13
Changes to Select List to support previously selected items

bipuladh:
/assign @rawagner

### DIFF
--- a/frontend/packages/console-shared/src/hooks/select-list.ts
+++ b/frontend/packages/console-shared/src/hooks/select-list.ts
@@ -6,12 +6,19 @@ export const useSelectList = <R extends K8sResourceCommon>(
   data: R[],
   visibleRows: Set<string>,
   onRowSelected: (rows: R[]) => void,
+  preSelected: string[],
 ): {
   onSelect: OnSelect;
   selectedRows: Set<string>;
   updateSelectedRows: (rows: R[]) => void;
 } => {
   const [selectedRows, setSelectedRows] = React.useState<Set<string>>(new Set());
+
+  React.useEffect(() => {
+    if (preSelected?.length > 0 && selectedRows.size === 0) {
+      setSelectedRows(new Set(preSelected));
+    }
+  }, [preSelected, selectedRows.size]);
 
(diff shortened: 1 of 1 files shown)


frontend/packages/console-shared/src/hooks/select-list.ts
```diff
@@ -6,12 +6,21 @@ export const useSelectList = <R extends K8sResourceCommon>(
   data: R[],
   visibleRows: Set<string>,
   onRowSelected: (rows: R[]) => void,
+  preSelected: string[] = [],
```
rawagner, line 9: I'd prefer not to specify default value, because empty array as default will retrigger useEffect on every rerender (its always a new array instance).

frontend/packages/console-shared/src/hooks/select-list.ts
```diff
@@ -6,12 +6,21 @@ export const useSelectList = <R extends K8sResourceCommon>(
...
 ): {
   onSelect: OnSelect;
   selectedRows: Set<string>;
   updateSelectedRows: (rows: R[]) => void;
 } => {
   const [selectedRows, setSelectedRows] = React.useState<Set<string>>(new Set());
+  const prevProps = React.useRef<string[]>(null);
+
+  React.useEffect(() => {
+    if (preSelected.length > 0 && prevProps.current !== preSelected) {
```
rawagner, line 19: needs optional chaining since we wont have default value
also this is supposed to be only initial selection we should make sure it runs only once, right ?
```suggestion
    if (preSelected?.length > 0 && prevProps.current !== preSelected && selectedRows.length === 0) {
```

Do you need to store the `prevProps` and check it against `preSelected` ?

rawagner (changes requested):

frontend/packages/console-shared/src/hooks/select-list.ts
bipuladh: yes that's the reason we are checking with prevProps

bipuladh:
/close
This PR is now obsolete

PR #6919, opened 2020-10-13
Save/Restore topology graph layout during current session

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-2344

**Description**
Stores the topology graph view's node locations and graph settings during the current session and restores them upon navigation back to the topology view.

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/Topology.tsx
+++ b/frontend/packages/dev-console/src/components/topology/Topology.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
+import { connect } from 'react-redux';
 import { action } from 'mobx';
+import { debounce } from '@patternfly/react-core';
 import {
   ComponentFactory,
   Visualization,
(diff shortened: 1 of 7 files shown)


invincibleJai:
@jeff-phillips-18 looks like this needs rebase

frontend/packages/dev-console/src/components/topology/Topology.tsx
```diff
@@ -9,24 +11,64 @@ import {
...
+  try {
+    const storedLayouts = JSON.parse(storedLayoutsString);
+    return namespace ? storedLayouts[namespace] : storedLayouts;
+  } catch (e) {
+    return undefined;
+  }
+};
+
+const setTopologyLayout = (namespace: string, nodes: NodeModel[], layout: string) => {
+  const currentStore = getTopologyLayout() || {};
```
jerolimov, line 57: `getTopologyLayout` returns based on the namespace parameter the StoredLayout or ALL StoredLayouts. I would prefer here two functions. Maybe something like this:

```suggestion
const getStoredData = (): Record<string, StoredLayout> => {
  const storedLayoutsString = localStorage.getItem(TOPOLOGY_LAYOUT_LOCAL_STORAGE_KEY);
  try {
    return JSON.parse(storedLayoutsString) || {};
  } catch (e) {
    return {};
  }
};

const getTopologyLayout = (namespace: string): StoredLayout => {
  return getStoredData()[namespace];
};

const setTopologyLayout = (namespace: string, nodes: NodeModel[], layout: string) => {
  const currentStore = getStoredData();
```

jerolimov (changes requested):
Tested it locally and it works great. Cool new topology feature.

Added one recommendation about reusing a function for two different usecases.

frontend/packages/dev-console/src/components/topology/Topology.tsx
jeff-phillips-18: @jerolimov Updated, PTAL.

frontend/packages/dev-console/src/components/topology/Topology.tsx
```diff
@@ -9,24 +11,65 @@ import {
...
+
+type StoredLayout = {
+  nodes: NodeModel[];
+  layout: string;
+};
+
+const getStoredLayouts = (): { [namespace: string]: StoredLayout } => {
+  const storedLayoutsString = localStorage.getItem(TOPOLOGY_LAYOUT_LOCAL_STORAGE_KEY);
+  try {
+    return JSON.parse(storedLayoutsString);
```
jerolimov, line 49: @jeff-phillips-18 `localStorage.getItem()` returns null for a key which is not existing yet. And JSON.parse(null) returns null. So I think you want to add a fallback here, or? ` || {}`

jerolimov (approved):
Works great. Missed that feature since I started. :+1:

New topology version is required to save viewport zoom & position. Everything else works also with the old version.

/lgtm

jeff-phillips-18:
/hold

Found an issue with stored layouts when switching to list view

PR #6922, opened 2020-10-14
Bug 1887783: PVC upload cannot continue after approve the certificate

glekner:
We now check for a valid upload proxy certificate before allocating resources, so the user can quickly resume the upload after approving it. The invalid error will now be shown in the form it self:
image


glekner:
@rawagner can you review please

PR #6883, opened 2020-10-09
CONSOLE-2394: i18n administration pages

rhamilto:
Note:  this PR does not include the `Administration > Namespaces` page as the code for it is shared with `Home > Projects`, and @jcaianirh has agreed to cover it as part of his story.

[CONSOLE-2394](https://issues.redhat.com/browse/CONSOLE-2394)

PR #6928, opened 2020-10-14
CONSOLE-2372: i18n - Networking pages

dtaylor113:
Administrator -> Networking:
- Services
- Routes
- Ingresses
- Network  Policies


frontend/public/components/routes.tsx
```diff
@@ -233,25 +195,30 @@ const RouteTableRow: RowFunction<RouteKind> = ({ obj: route, index, key, style }
...
-      <DetailsItem label="Certificate" obj={route} path="spec.tls.certificate">
+      <DetailsItem label={t('network-route~Certificate')} obj={route} path="spec.tls.certificate">
         {tls.certificate ? <CopyToClipboard value={tls.certificate} /> : '-'}
       </DetailsItem>
       <dt className="co-m-route-tls-reveal__title">
-        Key{' '}
+        {t('network-route~Key')}{' '}
         {tls.key && (
           <Button
             className="pf-m-link--align-left"
```
spadgett, line 224: Hide and Reveal text below needs to be externalized.

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,82 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
+            >
+              {ingress.wildcardPolicy}
+            </DetailsItem>
+            <DetailsItem
+              label={t('network-route~Router canonical hostname:')}
+              obj={route}
+              path="status.ingress.routerCanonicalHostname"
+            >
+              <div>{ingress.routerCanonicalHostname || '-'}</div>
+              {true && (
```
spadgett, line 376: Don't forget to revert this change for testing :)

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,139 @@ const ServiceAddress = ({ s }) => {
     switch (type) {
       case 'NodePort':
         return ServiceIPsRow(
-          'Node Port',
-          'Accessible outside the cluster',
+          t('network-service~Node Port'),
```
spadgett, line 106:
```suggestion
          t('network-service~Node port'),
```

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,139 @@ const ServiceAddress = ({ s }) => {
...
+          t('network-service~Accessible outside the cluster'),
           _.map(s.spec.ports, 'nodePort'),
-          '(all nodes): ',
+          t('network-service~(all nodes): '),
         );
       case 'LoadBalancer':
         return ServiceIPsRow(
-          'External Load Balancer',
-          'Ingress point(s) of load balancer',
+          t('network-service~External Load Balancer'),
```
spadgett, line 113:
```suggestion
          t('network-service~External load balancer'),
```

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,139 @@ const ServiceAddress = ({ s }) => {
...
+          t('network-service~Ingress point(s) of load balancer'),
           _.map(s.status.loadBalancer.ingress, (i) => i.hostname || i.ip || '-'),
         );
       case 'ExternalName':
         return ServiceIPsRow(
-          'External Service Name',
-          'Location of the resource that backs the service',
+          t('network-service~External Service Name'),
```
spadgett, line 119:
```suggestion
          t('network-service~External service name'),
```

spadgett (commented):
Looks good overall, just a few comments. Thanks @dtaylor113 👍

frontend/public/components/ingress.jsx
```diff
@@ -199,47 +169,87 @@ const RulesRows = (props) => {
...
-    </div>
-  </>
-);
-
+      <div className="co-m-pane__body">
+        <SectionHeading text={t('ingress~Ingress rules')} />
+        <p className="co-m-pane__explanation">
+          {t(
+            'ingress~rules_description',
+            'These rules are handled by a routing layer (Ingress Controller) which is updated as the rules are modified. The Ingress controller implementation defines how headers and other metadata are forwarded or manipulated',
```
rebeccaalpert, line 190: I'm concerned using keys and default values will make it difficult to pass files off to translation. (They pretty much only get the keys.) Do you have thoughts on that @spadgett? Thanks.

frontend/public/components/network-policy.jsx
```diff
@@ -178,26 +184,28 @@ const IngressRow = ({ ingress, namespace, podSelector }) => {
...
-        <SectionHeading text="Ingress Rules" />
+        <SectionHeading text={t('network-policy~Ingress rules')} />
         <p className="co-m-pane__explanation">
-          Pods accept all traffic by default. They can be isolated via Network Policies which
-          specify a whitelist of ingress rules. When a Pod is selected by a Network Policy, it will
-          reject all traffic not explicitly allowed via a Network Policy. See more details in{' '}
+          {t(
+            'network-policy~ingress_rules_description',
+            'Pods accept all traffic by default. They can be isolated via Network Policies which specify a whitelist of ingress rules. When a Pod is selected by a Network Policy, it will reject all traffic not explicitly allowed via a Network Policy. See more details in:',
+          )}
```
rebeccaalpert, line 200: I'm concerned about defined keys here as well.

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,82 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
+      headerContent={<>{t('network-route~Custom route')}</>}
+      bodyContent={
+        <div>
+          <p>
+            <Trans i18nKey="network-route~custom_route">
+              To use a custom route, you must update your DNS provider by creating a canonical name
+              (CNAME) record. Your CNAME record should point to your custom domain{' '}
+              <strong>{{ host }}</strong>, to the OpenShift canonical router hostname,{' '}
+              <strong>{{ routerCanonicalHostname }}</strong>, as the alias.
+            </Trans>
```
rebeccaalpert, line 340: Worried about keys here too. If you do something like `<Trans t={t} ns="network-route">` you can use the text as the key.

rebeccaalpert (approved):
Looks good to me aside from the capitalization bits Sam noted. I'm just concerned about the impact of using keys instead of text on translation given that the translators only really see the keys. @spadgett really makes the final decision on that though.

frontend/public/components/routes.tsx
dtaylor113: yikes!   -thanks :-)

frontend/public/components/ingress.jsx
dtaylor113: I thought this was neccesary due to some character limit of the keys?

rebeccaalpert: I believe that was actually an issue with i18n not being configured to allow periods. We had the parser configured, but not i18n itself. I think Robb said it was resolved.

rebeccaalpert: Let me know if you run into something weird like that though.

frontend/public/components/network-policy.jsx
dtaylor113: @spadgett & @rebeccaalpert, I think we need a ruling here :-)  I'm concerned other devs might be using keys.  We had a thread about their usage, and I thought it was best practice to use the key for long descriptions.

dtaylor113: I see some keys being used in @rhamilto 's PR [here](https://github.com/openshift/console/pull/6883/files#diff-938bd1f94fe7531ac8316dbfa825e7947b180102cdbd171e53c8758af149030fR20).

rebeccaalpert: Yeah, I haven't reviewed Robb's yet. It's up to Sam ultimately; I'm just concerned about the impact on translation.

frontend/public/components/ingress.jsx
spadgett: Originally we had a problem with periods, but I thought that was resolved. @rhamilto, did you switch to using the full string as a key?

The translators should have access to the English file, so I think they could figure it out. But I might just use the full message as the key everywhere so we're consistent and there's no confusion.

dtaylor113:
Hi @spadgett & @rebeccaalpert, I believe I fixed all the capitalization issues.

There is still a question about [defined keys](https://github.com/openshift/console/pull/6928#discussion_r504879052).

And I am also seeing this issue using this simple parameter subsitution:
```
t('network-policy~All traffic is allowed to Pods in {{namespace}}', {
  namespace: np.metadata.namespace,
})
```
which produces:
"All traffic is allowed to Pods in {{namespace}}": "All traffic is allowed to Pods in {{namespace}}"

English output seems fine:
"All traffic is allowed to Pods in daves-project"

but pseudo-translation is translating the variable values!?
[ÀÀḽḽ ţṛααϝϝḭḭͼ ḭḭṡ ααḽḽṓṓẁḛḛḍ ţṓṓ Ṕṓṓḍṡ ḭḭṇ ḍααṽḛḛṡ-ṗṛṓṓĵḛḛͼţ]
Shouldn't this be:
[ÀÀḽḽ ţṛααϝϝḭḭͼ ḭḭṡ ααḽḽṓṓẁḛḛḍ ţṓṓ Ṕṓṓḍṡ ḭḭṇ daves-project]  ?

Other parameter subs are working, but they are using the `<Trans />` component though

rebeccaalpert:
@dtaylor113 - Sam made a ruling here: https://github.com/openshift/console/pull/6928#discussion_r504948558

Unfortunately, that's an issue with the pseudolocalization library we're using. It's pretty bare-bones as far as code goes. Maybe when this project is over we can look into contributing and fixing that bug.

frontend/public/components/routes.tsx
```diff
@@ -512,9 +500,56 @@ export const RoutesDetailsPage: React.FC<RoutesDetailsPageProps> = (props) => (
...
+      {
+        title: '',
+        props: { className: tableColumnClasses[5] },
+      },
+    ];
+  };
+  return (
+    <Table
+      {...props}
+      aria-label="Routes"
```
rebeccaalpert, line 546: This aria label should be marked for translation.

rebeccaalpert (commented):
Capitalization looks good to me!

frontend/public/components/ingress.jsx
dtaylor113: Changed to use full message as the key.

frontend/public/components/network-policy.jsx
dtaylor113: Changed to use full message as the key.

frontend/public/components/routes.tsx
dtaylor113: Changed to use `<Trans />` element as suggested, thanks!

frontend/public/components/routes.tsx
dtaylor113: fixed, good catch, thought I did a search on those files, just redid one, no untranslated `aria-label`'s found in these files.

dtaylor113:
Hi @spadgett & @rebeccaalpert, I believe I have addressed all comments in this PR.  PTAL  -thanks :-)

rebeccaalpert (approved):
Looks good to me! Thank you.

frontend/public/components/routes.tsx
rebeccaalpert: I missed it my first pass! :) I swear they hide.

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,82 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
-);
+const RouteIngressStatus: React.FC<RouteIngressStatusProps> = ({ route }) => {
+  const { t } = useTranslation();
+  return (
+    <>
+      {_.map(route.status.ingress, (ingress: RouteIngress) => (
+        <div key={ingress.routerName} className="co-m-route-ingress-status">
+          <SectionHeading text={`${t('network-route~Router:')} ${ingress.routerName}`} />
+          <dl>
+            <DetailsItem label={t('network-route~Host:')} obj={route} path="status.ingress.host">
```
spadgett, line 360: We shouldn't include a `:` in the property names passed to `DetailsItem`

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,82 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
-    ))}
-  </>
-);
+const RouteIngressStatus: React.FC<RouteIngressStatusProps> = ({ route }) => {
+  const { t } = useTranslation();
+  return (
+    <>
+      {_.map(route.status.ingress, (ingress: RouteIngress) => (
+        <div key={ingress.routerName} className="co-m-route-ingress-status">
+          <SectionHeading text={`${t('network-route~Router:')} ${ingress.routerName}`} />
```
spadgett, line 358: Ideally `routerName` is a parameter in the message

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,139 @@ const ServiceAddress = ({ s }) => {
...
           _.map(s.spec.ports, 'nodePort'),
-          '(all nodes): ',
+          t('network-service~(all nodes): '),
         );
       case 'LoadBalancer':
         return ServiceIPsRow(
-          'External Load Balancer',
-          'Ingress point(s) of load balancer',
+          t('network-service~External load balancer'),
+          t('network-service~Ingress point(s) of load balancer'),
```
spadgett, line 114: nit: Not new, but I would remove the parentheses around the `s`

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,139 @@ const ServiceAddress = ({ s }) => {
...
+        <div className="col-xs-6">{t('network-service~Location')}</div>
       </div>
       <div className="rows">
         {ServiceType(s.spec.type)}
         {s.spec.externalIPs &&
           ServiceIPsRow(
-            'External IP',
-            'IP Address(es) accepting traffic for service',
+            t('network-service~External IP'),
+            t('network-service~IP Address(es) accepting traffic for service'),
```
spadgett, line 143: I'd remove the parentheses here too

frontend/public/components/routes.tsx
dtaylor113: fixed, and also the two below this one :-)

frontend/public/components/routes.tsx
dtaylor113: fixed

frontend/public/components/service.jsx
dtaylor113: fixed

frontend/public/components/service.jsx
dtaylor113: fixed

frontend/public/components/routes.tsx
spadgett: `:` is not a problem in the key. It's simply that we don't use `:` in the `DetailsItem` labels. It's fine here in the header.

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,90 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
+      {_.map(route.status.ingress, (ingress: RouteIngress) => (
+        <div key={ingress.routerName} className="co-m-route-ingress-status">
+          <SectionHeading
+            text={`${t('network-route~Router: {{routerName}}', {
+              routerName: ingress.routerName,
+            })}`}
+          />
+          <dl>
+            <DetailsItem
+              label={`${t('network-route~Host')}:`}
```
spadgett, line 365:
```suggestion
              label={t('network-route~Host')}
```

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,90 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
+              obj={route}
+              path="status.ingress.host"
+            >
+              {ingress.host}
+            </DetailsItem>
+            <DetailsItem
+              label={`${t('network-route~Wildcard policy')}:`}
```
spadgett, line 372:
```suggestion
              label={t('network-route~Wildcard policy')}
```

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,90 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
+              obj={route}
+              path="status.ingress.wildcardPolicy"
+            >
+              {ingress.wildcardPolicy}
+            </DetailsItem>
+            <DetailsItem
+              label={`${t('network-route~Router canonical hostname')}:`}
```
spadgett, line 379:
```suggestion
              label={t('network-route~Router canonical hostname')}
```

frontend/public/components/routes.tsx
dtaylor113: @spadgett What do we do with the `:` ?

frontend/public/components/routes.tsx
dtaylor113: Are we dropping the `:`?  I guess so, I don't see them on other detail pages, wonder why there are in this one?

frontend/public/components/routes.tsx
dtaylor113: ah, k , got it -thanks

dtaylor113:
@spadgett, I believe I have addressed all review comments -thanks

sg00dwin:
I noticed that the Routes status filter dropdown label and selections are not translated. Should they be? https://github.com/openshift/console/blob/988aa8f24a9e7e2c223d8fecf29be3a91c0e458e/frontend/public/components/routes.tsx#L558-L569

I have a similar status filter on the VolumeSnapshots list page.

dtaylor113:
> I noticed that the Routes status filter dropdown label and selections are not translated...I have a similar status filter on the VolumeSnapshots list page

@sg00dwin & @rebeccaalpert, this [commit](https://github.com/openshift/console/pull/6928/commits/ddecfc84b26e7aef2fc57b780609d5c38fbab6a8) seemed to work for me:
image

frontend/public/components/routes.tsx
```diff
@@ -353,69 +324,86 @@ const RouteTargetRow: React.FC<RouteTargetRowProps> = ({ route, target }) => (
...
     route.spec.host,
     route.status.ingress,
   );
   return (
     <>
       <div className="co-m-pane__body">
-        <SectionHeading text="Route Details" />
-        <RouteCharts namespace={route.metadata.namespace} route={route.metadata.name} />
+        <SectionHeading text={t('network-route~Route details')} />
+        <RouteCharts namespace={route.metadata.namespace} route={route.metadata.name} t={t} />
```
spadgett, line 406: We shouldn't be passing `t` to components. They should use their own `useTranslation`.

We'll eventually be passing a namespace when we call `useTranslation` so not all `t` are the same.

dtaylor113: ok, fixed

frontend/public/components/nav/admin-nav.tsx
```diff
@@ -197,7 +197,7 @@ const AdminNav = () => {
           activePath="/brokermanagement/"
           startsWith={brokerManagementStartsWith}
         />
-      </NavSection>
+      </NavSection> conf
```
spadgett, line 200: Typo?

dtaylor113: yup, fixed, thanks

spadgett (approved):
/lgtm
/hold

spadgett:
for approvals

frontend/public/components/network-policy.jsx
```diff
@@ -178,26 +184,29 @@ const IngressRow = ({ ingress, namespace, podSelector }) => {
 };
 
 const Details_ = ({ obj: np, flags }) => {
+  const { t } = useTranslation();
   return (
     <>
       <div className="co-m-pane__body">
-        <SectionHeading text="Namespace Details" />
+        <SectionHeading text={t('network-policy~Namespace details')} />
```
yanpzhan, line 191: Hi, when checked on network policy detail page, found this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1891419 , so the translation text also need update accordingly.

dtaylor113: Hi @yanpzhan, PTAL and retest with new 'Network policy details' title (per UXD and I18n aggrement on sentance-case headings).  -thanks

spadgett: Hi, @yanpzhan can you retest? The heading should be fixed now.

yanpzhan: Sorry for late response, I'm on leave yesterday. I checked on console against the pr,  the network policy detail page has been fixed.

spadgett (approved):
/lgtm

Thanks @dtaylor113 👍

PR #6930, opened 2020-10-14
i18n: update namespace consolidator

rebeccaalpert:
Fixed bug related to asynchronous file editing. Also updated the parser config so it won't create _old files.


spadgett (approved):
/lgtm

PR #6936, opened 2020-10-15
[release-4.6.z] Bug 1887813: assign masquerade to pod interface in RHV import

suomiy:


suomiy:
added the vmware part as well

PR #6948, opened 2020-10-15
CONSOLE-2428: i18n for search section of home nav

jcaianirh:
For: https://issues.redhat.com/browse/CONSOLE-2428

Search section only


spadgett (approved):
lgtm, we'll need QE review.

Note that I think there is still some work to do to properly lazy load the translations, but we need to investigate. That can be handled separately.

spadgett:
@yapei This is ready for QE review. This PR only covers basic elements of the Home -> Search page like the title and toolbar. It does not cover the list contents itself, which will be handled separately for each resource type.

for QE/docs/CEE approval

ahardin-rh:
LGTM from the docs side

spadgett:
> LGTM from the docs side

/label docs-approved

jcaianirh:
> /label px-approved
/label px-approved

spadgett (approved):
/lgtm

PR #6954, opened 2020-10-19
Bug 1886168: Check if the label exists before comparing its value

jhadvig:
We need to check if the label exists before comparing to a value. If it does not exists then the statement result will true. Additional PR to https://github.com/openshift/console/pull/6876
/assign @spadgett

### DIFF
--- a/frontend/packages/console-app/src/components/nodes/NodeDetailsPage.tsx
+++ b/frontend/packages/console-app/src/components/nodes/NodeDetailsPage.tsx
@@ -31,11 +31,11 @@ const NodeDetailsPage: React.FC<React.ComponentProps<typeof DetailsPage>> = (pro
         <PodsPage showTitle={false} fieldSelector={`spec.nodeName=${obj.metadata.name}`} />
       )),
       events(ResourceEventStream),
-      ...(_.find(
+      ...(!_.some(
         node?.metadata?.labels,
-        (label) =>
-          label['corev1.LabelOSStable'] !== 'windows' ||
-          label['node.openshift.io/os_id'] !== 'Windows',
+        (v, k) =>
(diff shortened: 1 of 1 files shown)


frontend/packages/console-app/src/components/nodes/NodeDetailsPage.tsx
```diff
@@ -34,8 +34,8 @@ const NodeDetailsPage: React.FC<React.ComponentProps<typeof DetailsPage>> = (pro
       ...(_.find(
         node?.metadata?.labels,
         (label) =>
-          label['corev1.LabelOSStable'] !== 'windows' ||
-          label['node.openshift.io/os_id'] !== 'Windows',
+          (label['corev1.LabelOSStable'] && label['corev1.LabelOSStable'] !== 'windows') ||
+          (label['node.openshift.io/os_id'] && label['node.openshift.io/os_id'] !== 'Windows'),
       )
```
spadgett, line 39: This is simpler as

```ts
!_.some(node?.metadata?.labels, (label) => {
  return label['corev1.LabelOSStable'] === 'windows' || label['node.openshift.io/os_id'] === 'Windows';
})
```

jhadvig: Since the `labels` is not a nested object the iterated `label` will be the just the value of label. Needed to update it to:
```js
!_.some(node?.metadata?.labels, (v, k) => (k === 'node.openshift.io/os_id' && v === 'Windows') || (k === 'corev1.LabelOSStable' && v === 'windows'))
```

jhadvig:
@spadgett PR update. Added additional comment.

spadgett (approved):
/lgtm

PR #6957, opened 2020-10-19
Bug 1872923: Fix cancel button to return to operator details

zherman0:
Cancel button uses goBack which in the case of the new operator install status screen, would go back to the install subscription page instead of the operator details page (which is where you go when you come from operator details page initially).

### DIFF
--- a/frontend/packages/console-shared/src/components/dynamic-form/index.tsx
+++ b/frontend/packages/console-shared/src/components/dynamic-form/index.tsx
@@ -28,6 +28,7 @@ export const DynamicForm: React.FC<DynamicFormProps> = ({
   onChange = _.noop,
   onError = _.noop,
   onSubmit = _.noop,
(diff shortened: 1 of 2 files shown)


frontend/packages/console-shared/src/components/dynamic-form/index.tsx
```diff
@@ -104,6 +105,7 @@ type DynamicFormProps = FormProps<any> & {
   ErrorTemplate?: React.FC<{ errors: string[] }>;
   noActions?: boolean;
   customUISchema?: boolean;
+  onCancel?: () => void;
```
spadgett, line 108: The TypeScript type should match the signature of `onClick` which takes an event

spadgett (approved):
/lgtm

Just one small nit on the TypeScript type, but I don't want to hold the PR on that.

PR #6959, opened 2020-10-20
Query Browser: Fix tooltips next to the edge of the graph

kyoto:
By default, `ChartVoronoiContainer` has 5px padding, which prevents the
tooltip appearing for any data points within 5px of the edge of the
graph. This fixes that by setting the padding to zero.


spadgett (approved):
/lgtm

PR #6960, opened 2020-10-20
update pipelinerun graph to show past pipeline structure

karthikjeeyar:
**Fixes**: https://issues.redhat.com/browse/ODC-2302

**Description**: 
As a user, I would like to be able to view my past Pipeline Run visualization graphs without them being impacted by my current Pipeline status.

**Problem**:
Pipeline visualization component was always fetching the tasks from the current pipeline.

**Solution**:
Visualization component now uses the tasks from the `pipelinerun.status.pipelineSpec` field, which holds the reference of the pipeline at the time of execution. As a fallback, we still read from the pipeline if this  `pipelinerun.status.pipelineSpec` field is not available.

**Screenshots:**
pipelinerun-graph

**Test cases:**
image

cc: @siamaksade @andrewballantyne 
/kind feature


jerolimov (approved):
Tested this locally, works as expected

/lgtm

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -396,9 +397,21 @@ export const getRunStatusColor = (status: string): StatusMessage => {
...
+ * Returns the pipeline with the spec at the time of pipeline run execution.
+ * @param pipeline
+ * @param pipelineRun
+ */
+export const getExecutedPipeline = (pipeline: Pipeline, pipelineRun: PipelineRun): Pipeline => {
+  return {
+    ...pipeline,
+    ...(pipelineRun?.status?.pipelineSpec && { spec: pipelineRun?.status?.pipelineSpec }),
+  };
+};
```
andrewballantyne, line 410: I think we want to do two things here:

1. Remove the dependancy on Pipeline and trim it out along all the call paths (I just did this and it's quite a bit of changes and adjustments, no references to Pipeline should exist in the call paths)
2. Rename it to something like (don't have to take this if you have something better) `getPipelineFromPipelineRun`

Tests will need to be updated naturally when this happens, since `getTaskStatus` heavily uses Pipeline. We may even need to update the test data to add `pipelineSpec` (which should be a mirror - maybe break that out to a var? - of the pipeline.spec used in the same data structure)

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -178,6 +178,7 @@ export interface PipelineRun extends K8sResourceKind {
     startTime?: string;
     completionTime?: string;
     taskRuns?: PLRTaskRuns;
+    pipelineSpec?: PipelineSpec;
```
andrewballantyne, line 181:
```suggestion
    pipelineSpec: PipelineSpec;
```
Based on my review comment.

andrewballantyne (changes requested):
hmm, I don't think we need to keep the dependancy on `Pipeline`. It is true that 1.0.1 of the Operator does still require it but based on the [Compatibility Matrix](https://docs.jboss.org/display/SRVKP/Pipeline+Component+Matrix) 4.7 is going to be a far cry away from upstream Pipelines not having this feature (which was [added in 0.12](https://github.com/tektoncd/pipeline/releases/tag/v0.12.0)).

For sanity for the customer's cluster, I think we could go back 1 (1.2) or even more safety 2 (1.1) versions and still give that support so they are not getting a broken UI immediately after upgrading but not changing their channel. If memory serves correctly, when the Operator updates to give us an `ocp-4.7` channel, they'll remove the `ocp-4.5` channel. Which will remove any chance for unsupported code to exist.

The easiest approach to solving this ask I think is to just remove the requirement of Pipeline from `getExecutedPipeline` (see my comment below to get you started).

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -396,9 +397,21 @@ export const getRunStatusColor = (status: string): StatusMessage => {
 export const truncateName = (name: string, length: number): string =>
   name.length < length ? name : `${name.slice(0, length - 1)}...`;
 
+/**
+ * Returns the pipeline with the spec at the time of pipeline run execution.
+ * @param pipeline
+ * @param pipelineRun
```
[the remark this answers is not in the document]
karthikjeeyar: Removed it.

frontend/packages/dev-console/src/utils/pipeline-augment.ts
karthikjeeyar: Removed the pipeline dependency in the call paths, updated test data and tests as well.

frontend/packages/dev-console/src/utils/__tests__/pipeline-actions.spec.ts
```diff
@@ -36,14 +36,22 @@ export const actionPipelineRuns: PipelineRun[] = [
...
     apiVersion: 'abhiapi/v1',
     kind: 'Pipeline',
     metadata: { name: 'dragonstone', namespace: 'corazon' },
     spec: { pipelineRef: { name: 'danaerys-targaeryen' } },
-    status: { creationTimestamp: '31', conditions: [{ type: 'Succeeded', status: 'Unknown' }] },
+    status: {
+      pipelineSpec: { tasks: [] },
+      creationTimestamp: '31',
+      conditions: [{ type: 'Succeeded', status: 'Unknown' }],
+    },
```
andrewballantyne, line 54: Well this is duplicate data from `frontend/packages/dev-console/src/components/pipelines/modals/common/__tests__/utils.spec.ts` 😞 

Think we could move this data into `pipeline-data.ts` and then import it back?

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -421,9 +424,24 @@ export const getRunStatusColor = (status: string): StatusMessage => {
...
 
-export const getTaskStatus = (pipelinerun: PipelineRun, pipeline: Pipeline): TaskStatus => {
-  const totalTasks =
-    pipeline && pipeline.spec && pipeline.spec.tasks ? pipeline.spec.tasks.length : 0;
+export const getPipelineFromPipelineRun = (pipelineRun: PipelineRun): Pipeline => {
+  if (!pipelineRun?.status?.pipelineSpec) {
+    return null;
+  }
+  return {
+    apiVersion: PipelineModel.apiVersion,
```
andrewballantyne, line 432:
```suggestion
    apiVersion: apiVersionForModel(PipelineModel),
```
Awkward, I know... but `apiVersion` in the Model is actually the version, whereas the K8s CR is the "group + version".

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -421,9 +424,24 @@ export const getRunStatusColor = (status: string): StatusMessage => {
 export const truncateName = (name: string, length: number): string =>
   name.length < length ? name : `${name.slice(0, length - 1)}...`;
 
-export const getTaskStatus = (pipelinerun: PipelineRun, pipeline: Pipeline): TaskStatus => {
-  const totalTasks =
-    pipeline && pipeline.spec && pipeline.spec.tasks ? pipeline.spec.tasks.length : 0;
+export const getPipelineFromPipelineRun = (pipelineRun: PipelineRun): Pipeline => {
+  if (!pipelineRun?.status?.pipelineSpec) {
```
andrewballantyne, line 428: It's probably possible to pass in a inlineSpec for Pipeline, can we also check to see if the label for Pipeline is here on the PipelineRun?

```suggestion
  const pipelineName = pipelineRun.metadata.labels[TektonResourceLabel.pipeline];
  if (!pipelineName || !pipelineRun?.status?.pipelineSpec) {
```

frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -421,9 +424,24 @@ export const getRunStatusColor = (status: string): StatusMessage => {
...
+    return null;
+  }
+  return {
+    apiVersion: PipelineModel.apiVersion,
+    kind: PipelineModel.kind,
+    metadata: {
+      name: pipelineRun.metadata[TektonResourceLabel.pipeline],
```
andrewballantyne, line 435: ❓ Shouldn't this be `pipelineRun.metadata.labels`?
```suggestion
      name: pipelineRun.metadata.labels[TektonResourceLabel.pipeline],
```
Also, perhaps make use of the variable based on my last comment.

frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts
```diff
@@ -328,3 +296,38 @@ describe('Pipeline exists test to determine whether a pipeline is linked to a pi
...
+      ...testPipelineRun,
+      status: {
+        pipelineSpec: {
+          ...currentPipeline.spec,
+          tasks: [currentPipeline.spec.tasks[0]],
+        },
+      },
+    };
+    const executedPipeline = getPipelineFromPipelineRun(plrWithOneTask);
+    expect(currentPipeline.spec.tasks).toHaveLength(2);
```
andrewballantyne, line 329: This is a weird `expect`... you're not testing data you used to deliver the test of `getPipelineFromPipelineRun` you trimmed it out on 324 to being only the first one. What is this expect doing other than testing you have data that has 2 points (which is a bit odd of a test)?

frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts
```diff
@@ -328,3 +296,38 @@ describe('Pipeline exists test to determine whether a pipeline is linked to a pi
...
+    expect(executedPipeline.spec.tasks).toHaveLength(1);
+    expect(executedPipeline).not.toEqual(currentPipeline);
```
andrewballantyne, line 331: Confused about this too... this follows my previous comment's confusion.

frontend/packages/dev-console/src/utils/__tests__/pipeline-test-data.ts
```diff
@@ -272,79 +272,8 @@ export const constructPipelineData = {
...
-          status: 'True',
-          type: 'Succeeded',
-        },
-      ],
-      startTime: '2019-10-29T11:56:40Z',
-    },
-  },
+  pipelineTestData[PipelineExampleNames.SIMPLE_PIPELINE].pipelineRuns[DataState.SUCCESS],
+  pipelineTestData[PipelineExampleNames.WORKSPACE_PIPELINE].pipelineRuns[DataState.SUCCESS],
+  pipelineTestData[PipelineExampleNames.CLUSTER_PIPELINE].pipelineRuns[DataState.SUCCESS],
```
andrewballantyne, line 278: Hard approve of this change. Very nicely done, consolidating test data is awesome!

andrewballantyne (changes requested):
Few more touch ups please.

frontend/packages/dev-console/src/components/topology/components/nodes/build-decorators/PipelineBuildDecoratorTooltip.tsx
```diff
@@ -22,7 +22,7 @@ const PipelineBuildDecoratorTooltip: React.FC<PipelineBuildDecoratorTooltipProps
     return null;
   }
 
-  const taskStatus = getTaskStatus(pipeline.latestRun, pipeline);
+  const taskStatus = getTaskStatus(pipeline.latestRun);
```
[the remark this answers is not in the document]
karthikjeeyar: Removed the pipeline dependency here.

frontend/packages/dev-console/src/utils/__tests__/pipeline-actions.spec.ts
karthikjeeyar: Removed this mock data and used the data from `pipeline-data.ts`  and also refactored `utils.spec.ts` to avoid duplication of mock data.

frontend/packages/dev-console/src/utils/pipeline-augment.ts
karthikjeeyar: Yes, it should be. updated.

frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts
karthikjeeyar: Yeah I Agree, after the pipeline dependancy removal this test wasn't serving the purpose. So i updated it now.

frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts
karthikjeeyar: Updated the test

frontend/packages/dev-console/src/test/pipeline-data.ts
```diff
@@ -75,13 +277,15 @@ export const pipelineTestData: PipelineTestData = {
...
+          pipelineSpec: pipelineSpec[PipelineExampleNames.SIMPLE_PIPELINE],
           conditions: [
             {
               lastTransitionTime: '2019-09-12T20:38:01Z',
               message: 'All Tasks have completed executing',
               reason: 'Succeeded',
-              status: '',
-              type: 'Running',
+              status: 'Unknown',
+              type: 'Succeeded',
```
andrewballantyne, line 288: Interesting, this is `DataState.IN_PROGRESS`, you have changed it to completed... I think you're mixing data points here.

frontend/packages/dev-console/src/test/pipeline-data.ts
```diff
@@ -1266,12 +1395,12 @@ export const pipelineTestData: PipelineTestData = {
...
         },
         status: {
+          pipelineSpec: pipelineSpec[PipelineExampleNames.INVALID_PIPELINE_MISSING_TASK],
           completionTime: '2020-07-13T17:21:05Z',
           conditions: [
             {
               lastTransitionTime: '2020-07-13T17:21:05Z',
-              message:
-                'Pipeline andrew/task-ref-error can\'t be Run; it contains Tasks that don\'t exist: Couldn\'t retrieve Task "not-a-task": task.tekton.dev "not-a-task" not found',
+              message: '',
```
andrewballantyne, line 1403: Any reason you got rid of this data text? We could trim out my name for sure haha, it was log dumps when I created. We could make the message something like `Test error text`, but I think it shouldn't be empty... as our tests could send that data in and it be mistaken as no error text rather than just an empty error text for testing.

frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts
```diff
@@ -328,3 +296,34 @@ describe('Pipeline exists test to determine whether a pipeline is linked to a pi
     expect(pipelineFound).toBeFalsy();
   });
 });
+
+describe('Pipelinerun graph to show the executed pipeline structure', () => {
```
andrewballantyne, line 300: We are missing a test for the actual result. You definitely have tested the paths to our current usage, but the method is, in isolation, just creating a Pipeline. We don't have a test to make sure it did do that. Goals of Unit Tests for functions should be to cover edge cases, any conditional paths, intended uses today and the function is performing the action it prescribes to - which in this case is making a PL CR out of the PLR CR.

andrewballantyne (changes requested):
Great test changes, thanks for cleaning that up. Couple odd changes in the test data and one minor test needed.

frontend/packages/dev-console/src/test/pipeline-data.ts
karthikjeeyar: While consolidating the test data, there was a test case, where it returns a static message `Unknown failure condition` if the message field is emtpy. For that reason i added this. I have reverted back  the above one and handled the mock data in the test case itself.

frontend/packages/dev-console/src/test/pipeline-data.ts
karthikjeeyar: reverted back.

andrewballantyne (approved):
/lgtm

andrewballantyne:
/retest

Looks to be a connection issue

karthikjeeyar:
I rebased due to merge conflicts.

PR #6968, opened 2020-10-21
Bug 1889698: Fix form re-render on cancelling

afreen23:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1889698
- dispatch of an action that do not exists causing a reset of redux store.
- removed the stale dispatch action


bipuladh (commented):
https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/state.ts#L184
We should return the current state instead here too. 
This fix is sufficient but let's do this along with it.

PR #6973, opened 2020-10-21
Bug 1890145: Add classname to decrease font size for Status Ready

harishsurf:
Bug 1890145: The mismatched of font size for Status Ready
and Health Check secondary text

The image shows the applied fix:

font_size

### DIFF
--- a/frontend/packages/console-shared/src/components/status/SecondaryStatus.tsx
+++ b/frontend/packages/console-shared/src/components/status/SecondaryStatus.tsx
@@ -3,14 +3,16 @@ import * as _ from 'lodash';
 
 type SecondaryStatusProps = {
   status?: string | string[];
+  className?: string;
 };
 
-const SecondaryStatus: React.FC<SecondaryStatusProps> = ({ status }) => {
+const SecondaryStatus: React.FC<SecondaryStatusProps> = ({ status, className }) => {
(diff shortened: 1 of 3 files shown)


frontend/packages/console-shared/src/components/status/SecondaryStatus.tsx
```diff
@@ -10,7 +10,7 @@ const SecondaryStatus: React.FC<SecondaryStatusProps> = ({ status }) => {
   if (statusLabel) {
     return (
       <div>
-        <small className="text-muted">{statusLabel}</small>
+        <small className="co-status-card__health-item-subtitle text-muted">{statusLabel}</small>
```
spadgett, line 13: Per BEM, this class should only be applied in the health card. I believe `SecondaryStatus` is used in other contexts.

Is it possible to take in a className prop?

spadgett: http://getbem.com/introduction/

harishsurf: Modified the code to use `SecondaryStatus` component to render subtitle for `Health Checks`

spadgett (approved):
/lgtm

Thanks!

PR #6974, opened 2020-10-21
Add support for Multus(Network) in storage cluster creation under configuration

bipuladh:
Screenshot: 
Screenshot from 2020-10-21 23-54-18
Validations: 
Screenshot from 2020-11-08 21-26-19
Screenshot from 2020-11-08 21-23-26

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
@@ -1,19 +1,23 @@
 import * as React from 'react';
-
-import { FormGroup, Checkbox } from '@patternfly/react-core';
-import { FieldLevelHelp } from '@console/internal/components/utils';
-import { encryptionTooltip } from '../../../constants/ocs-install';
-import { TechPreviewBadge } from '@console/shared';
+import { FormGroup, Checkbox, Radio } from '@patternfly/react-core';
+import { FieldLevelHelp, Firehose } from '@console/internal/components/utils';
+import { TechPreviewBadge, getName, ResourceDropdown } from '@console/shared';
+import { NetworkAttachmentDefinitionKind } from '@console/network-attachment-definition-plugin/src/types';
+import { NetworkAttachmentDefinitionModel } from '@console/network-attachment-definition-plugin';
+import { referenceForModel } from '@console/internal/module/k8s';
 import { InternalClusterState, InternalClusterAction, ActionType } from '../internal-mode/reducer';
 import { State, Action } from '../attached-devices/create-sc/state';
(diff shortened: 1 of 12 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
```diff
@@ -40,6 +51,9 @@ export const CreateInternalCluster: React.FC<CreateInternalClusterProps> = ({ ma
   const title = 'create internal mode storage cluster wizard';
   const scName = getName(state.storageClass);
   const hasEnabledCreateStep = !!(state.nodes.length >= MINIMUM_NODES && scName);
+  // User can't have empty public NAD when using multus
+  const hasConfiguredNetwork =
+    state.networkType === NetworkType.MULTUS ? !!state.publicNetwork : true;
```
afreen23, line 56:
```suggestion
  // User can't have empty public NAD when using multus
 We want to disable the cluster creation step as well.
   const hasConfiguredNetwork =
    state.networkType === NetworkType.MULTUS ? !!state.publicNetwork : true;
      const hasEnabledCreateStep = !!(state.nodes.length >= MINIMUM_NODES && scName && hasConfiguredNetwork );
```

afreen23 (changes requested):
Add the changes in review step as well and the baermetal wizard flows https://github.com/openshift/console/pull/6976 are in place, so we need to update multus there as well.

yuvalgalanti:
The UI looks good! Thanks

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
```diff
@@ -1,14 +1,30 @@
...
+import { NetworkAttachmentDefinitionKind } from '@console/network-attachment-definition-plugin/src/types/types';
+import { Firehose } from '@console/internal/components/utils';
+import { NetworkType } from '../types';
+import './_configure.scss';
+
+const resources = [
+  {
+    isList: true,
+    kind: referenceForModel(NetworkAttachmentDefinitionModel),
+    namespace: 'openshift-storage',
```
rohan47, line 15: If the NAD is in `openshift-storage` namespace, then it can't be used in other namespaces.
For example, an application using OCS as storage backend which is in namespace other than openshift-storage won't be able to utilize multus.

Only the NADs in the `default` namespace can be used accross namespaces. So we need to list NADs from the `default` namespace as well.
The NADs from the `default` namespace is referenced as `default/<name-of-NAD>` in the network spec
e.g., 
```yaml
public: "default/ocs-public-nw"
cluster: "default/ocs-cluster-nw"
```

rohan47 (changes requested):

PR #6975, opened 2020-10-21
Bug 1887465: Fall back to "all namespaces" when currently active namespace is deleted

TheRealJon:
If a user deletes their currently active namespace from the UI, fall back to selecting
"all namespaces" as the active namespace. Only effects active namespace within the browser
session where the project was deleted. Update to typescript and function component using hooks to simplify the logic needed for this fix.

- create promise handler hook
- create active namespace selector hook
- migrate DeleteNamepsaceModal to TypeScript and FunctionComponent
- Add logic in DeleteNamespaceModal to change active namespace state when current namespace is deleted

### DIFF
--- a/frontend/packages/console-shared/src/hooks/promise-handler.ts
+++ b/frontend/packages/console-shared/src/hooks/promise-handler.ts
@@ -1,17 +1,24 @@
 import * as React from 'react';
 
-export const usePromiseHandler: PromisHandlerHook = () => {
+export const usePromiseHandler: PromisHandlerHook = <T extends unknown = any>() => {
   const [inProgress, setInProgress] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState('');
-  const handlePromise = (promise) => {
+  const handlePromise: PromiseHandlerCallback<T> = (promise) => {
     setInProgress(true);
     return promise
-      .then(() => setErrorMessage(''))
-      .catch((reason) => setErrorMessage(reason || 'An error occurred. Please try again.'))
+      .then((res) => {
+        setErrorMessage('');
+        return res;
+      })
+      .catch((reason) => {
+        const error = reason || 'An error occurred. Please try again.';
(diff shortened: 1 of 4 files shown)


frontend/packages/console-shared/src/hooks/promise-handler.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+export const usePromiseHandler: PromisHandlerHook = () => {
+  const [inProgress, setInProgress] = React.useState(false);
+  const [errorMessage, setErrorMessage] = React.useState('');
+  const handlePromise = (promise) => {
+    setInProgress(true);
+    return promise
+      .then(() => setErrorMessage(''))
+      .catch((reason) => setErrorMessage(reason || 'An error occurred. Please try again.'))
```
spadgett, line 10: Even though we're handling the failed promise here, I think this will cause an unhandled promise rejection event (and an error overlay in dev mode) because we're calling `then` on the returned promise. We should test.

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
```
spadgett, line 1: @jcaianirh FYI, this will conflict with the i18n work you're doing for the Home nav section.

spadgett (commented):
Thanks for update the dialog to use hooks/TypeScript 👍

frontend/packages/console-shared/src/hooks/promise-handler.ts
TheRealJon: @spadgett  This wasn't causing an uncaught Promise rejection, but it was "swallowing" the rejection since I wasn't returning a rejected promise or throwing. Should be fixed now.

spadgett (approved):
/lgtm

PR #6983, opened 2020-10-22
[WIP] i18n:  Automate Cypress pseudolocalization test

dtaylor113:
### Overview
Updated **pseudolocalization.spec.ts** to allow devs to specify a url to test for i18n translation, along with an array of testIDs and/or selectors for those elements/strings which have been i18n translated.

### List pages
Resource list pages test that all table column headers have been translated.  You can also indicate if an 'example' resource should be created and deleted, as well as any other `beforeTest` or `afterTest` steps requird (such as clicking a filter dropdown):
```
    .set('Ingresses list', {
      url: '/k8s/all-namespaces/ingresses',
      resourcesListPage: true,
      beforeTest: () => createExampleResource(),
      afterTest: () => deleteExampleResource('Ingress'),
    })
```
image
**_Note:_** Routes list page also tests that the filter dropdown has been translated.

### Detail pages
Resource detail pages test that the all following have been translated:
- Section Headings
- Detail Item Labels (`<dt>`'s)
- Any element which has `data-i18n-translated` attribute
- `timestamp`'s 
```
    .set('Ingress details', {
      url: '/k8s/all-namespaces/ingresses',
      resourceDetailsPage: true,
      beforeTest: () => {
        createExampleResource();
        listPage.rows.clickRowByName('example');
      },
      selectors: [i18nTranslated],
      afterTest: () => deleteExampleResource('Ingress'),
    })
```
**_Note:_** For test independance `resourceDetailsPage`'s which need an `example` resource created actually start with the resource list page URL, create the example resource, then drilldown into the `example` details page (see beforeTest() above).
image

**data-i18n-translated** example:
```
  <div data-i18n-translated className="col-xs-4">
    {t('network-policy~Target pods')}
  </div>
```

### Headless 
```
Running:  i18n/pseudolocalization.spec.ts                                                 (1 of 1)
  Localization
  skipping login, no BRIDGE_KUBEADMIN_PASSWORD set
    ✓ pseudolocalizes Masthead (8996ms)
    ✓ pseudolocalizes Sidebar navigation (1972ms)
    ✓ pseudolocalizes Cluster dashboard (5564ms)
    ✓ pseudolocalizes Events (2749ms)
    ✓ pseudolocalizes Services list (3508ms)
    ✓ pseudolocalizes Service details (3579ms)
    ✓ pseudolocalizes Routes list (2808ms)
    ✓ pseudolocalizes Route details (3908ms)
    ✓ pseudolocalizes Ingresses list (7345ms)
    ✓ pseudolocalizes Ingress details (4095ms)
    ✓ pseudolocalizes Policies list (6109ms)
  skipping logout
    ✓ pseudolocalizes Policy details (3741ms)
  12 passing (56s)
```


frontend/public/components/ingress.jsx
```diff
@@ -199,47 +177,86 @@ const RulesRows = (props) => {
...
       </div>
-    </div>
-  </>
-);
-
+      <div className="co-m-pane__body">
+        <SectionHeading text={t('ingress~Ingress rules')} />
+        <p data-test="ingress-rules-desc" className="co-m-pane__explanation">
+          {t(
+            'ingress~These rules are handled by a routing layer (Ingress Controller) which is updated as the rules are modified. The Ingress controller implementation defines how headers and other metadata are forwarded or manipulated',
```
rebeccaalpert, line 197: Ingress controller capitalization is inconsistent here - I'm not sure what's right.

frontend/public/locales/en/network-policy.json
```diff
@@ -0,0 +1,15 @@
...
+  "Pod selector": "Pod selector",
+  "Network policies": "Network policies",
+  "Target pods": "Target pods",
+  "From": "From",
+  "To ports": "To ports",
+  "NS selector": "NS selector",
+  "Namespace details": "Namespace details",
+  "Ingress rules": "Ingress rules",
+  "Pods accept all traffic by default. They can be isolated via Network Policies which specify a whitelist of ingress rules. When a Pod is selected by a Network Policy, it will reject all traffic not explicitly allowed via a Network Policy. See more details in:": "Pods accept all traffic by default. They can be isolated via Network Policies which specify a whitelist of ingress rules. When a Pod is selected by a Network Policy, it will reject all traffic not explicitly allowed via a Network Policy. See more details in:",
+  "Network Policies documentation": "Network Policies documentation",
```
rebeccaalpert, line 13: Is "Network Policies" supposed to be capitalized? It's not on line 5.

frontend/public/components/service.jsx
```diff
@@ -142,121 +103,153 @@ const ServiceAddress = ({ s }) => {
...
+        </div>
       </div>
       <div className="rows">
         {ServiceType(s.spec.type)}
         {s.spec.externalIPs &&
           ServiceIPsRow(
-            'External IP',
-            'IP Address(es) accepting traffic for service',
+            t('network-service~External IP'),
+            t('network-service~IP Addresses accepting traffic for service'),
```
rebeccaalpert, line 147: Is "addresses" supposed to be capitalized?

rebeccaalpert (commented):
The new testing stuff looks so much better than what I had. Thank you!

I just had a couple of questions about capitalization, which I'm sure Sam will know the answers to. The actual internationalization stuff looked good to me.

frontend/public/components/ingress.jsx
dtaylor113: Hi @rebeccaalpert, these capitilization issues should be filed in #6928!  -thanks

frontend/public/locales/en/network-policy.json
dtaylor113: Hi, if these are capitilization issues, pls add them to #6928! -thanks
cc @spadgett

rebeccaalpert: Thanks!

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
-  });
+  const testObjs = OrderedMap<
+    string,
+    {
+      url: string;
+      createResource?: boolean;
+      deleteResource?: string;
+      testIds?: string[];
+      selectors?: string[];
+    }
```
spadgett, line 51: I'd define a type for this instead of doing it inline

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
-    cy.visit('/dashboards?pseudolocalization=true&lng=en');
-    cy.byTestID('utilization-card-item-text').isPseudoLocalized();
-  });
+  const testObjs = OrderedMap<
+    string,
+    {
+      url: string;
+      createResource?: boolean;
+      deleteResource?: string;
+      testIds?: string[];
```
spadgett, line 49:
```suggestion
      testIDs?: string[];
```

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
+      selectors?: string[];
+    }
+  >()
+    .set('masthead', {
+      url: '/dashboards',
```
spadgett, line 54: I would pick a URL that isn't as heavy as `/dashboards` for these common components. Maybe the cluster settings page.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
+        'activity-recent-title',
+        'ongoing-title',
+        'events-view-all-link',
+        'events-pause-button',
+      ],
+    })
+    .set('utilization card', {
+      url: '/dashboards',
+      testIds: ['utilization-card-item-text'],
+    })
```
spadgett, line 74: I'd combine the elements on one page as a single entry

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
+    .set('timestamps', {
```
spadgett, line 75: Change this to cover all strings on the events page.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,171 @@ describe('Localization', () => {
...
+      url: '/k8s/all-namespaces/events',
+      testIds: ['timestamp'],
+    })
+    .set('networking services', {
```
spadgett, line 79: I wouldn't prefix each with the nav section since not all pages have a nav section. I'd use names like `Service list` and `Service details`.

frontend/public/components/routes.tsx
```diff
@@ -208,16 +208,23 @@ const TLSSettings: React.FC<TLSSettingsProps> = ({ route }) => {
         label={t('network-route~Termination type')}
         obj={route}
         path="spec.tls.termination"
+        dataTest="termination-type"
```
spadgett, line 211: Instead of adding an identifier for every details item, I'm wondering if we can check something like `co-m-pane__details th` to make sure *all* the labels are translated. It's less code, and we cover new items that aren't manually added to the test. Or we can add a common `data-` attribute into `DetailsItem` for the label.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: Unfortunately, at this time `timestamps` are the only things translated on the Events page

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/public/components/routes.tsx
dtaylor113: At this time, not all labels are translated on all detail pages.  

_"...Or we can add a common data- attribute into DetailsItem for the label"_
That's what I did [here](https://github.com/openshift/console/pull/6983/files#diff-631a7e4e0f7e94fa6334e256bfae9262ddff281897e4b386eca899e30bd0afaeR72), unless I'm misunderstanding what you are asking about?

dtaylor113: Fixed, testing for all Detail Item Labels and introduced 'data-i18n-translated` attribute.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: Although, I see #7130 was just created.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,134 @@ describe('Localization', () => {
...
 
-  it('pseudolocalizes timestamps', () => {
-    cy.log('test timestamps');
-    cy.visit('/k8s/all-namespaces/events?pseudolocalization=true&lng=en');
-    cy.byTestID('timestamp').isPseudoLocalized();
-  });
+  testObjs.forEach(
+    ({ url, createResource = false, deleteResource, testIDs = [], selectors = [] }, resource) => {
+      it(`pseudolocalizes ${resource}`, () => {
+        const pseudoLocUrl = `${url}?pseudolocalization=true&lng=en`;
```
spadgett, line 118: It's better to use `URLSearchParams` (or the Node equivalent if needed) to build the URL so we correctly handle URLs that already have params.

spadgett:
```suggestion
        const pseudoLocURL = `${url}?pseudolocalization=true&lng=en`;
```

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,134 @@ describe('Localization', () => {
...
+
+        if (createResource) {
+          createExampleResource(pseudoLocUrl);
+        }
+
+        if (resource === 'Masthead') {
+          masthead.clickMastheadLink('help-dropdown-toggle');
+        } else if (resource === 'Routes list') {
+          listPage.filter.clickFilterDropdown();
+        }
```
spadgett, line 129: We shouldn't hard-code this here. Maybe `LocalizationTestProps` has some kind of setup function we can call to do whatever it needs to do.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,134 @@ describe('Localization', () => {
...
+      selectors: [sectionHeadings, detailLabels, i18nTranslated, '.graph-title'],
+      testIDs: ['timestamp'],
+    })
+    .set('Ingresses list', {
+      url: '/k8s/all-namespaces/ingresses',
+      createResource: true,
+      selectors: [tableColumnHeaders],
+    })
+    .set('Ingress details', {
+      url: '/k8s/all-namespaces/ingresses/example',
```
spadgett, line 98: It would be better if each test was independent so it doesn't rely on the previous test creating the resource (or a different test deleting the resource that was created).

We could look at having two groups, one for general pages, and one for resource lists that always create the resource (like the crud tests). Or we can maybe hook this into the existing crud tests.

frontend/public/components/ingress.jsx
```diff
@@ -103,10 +103,18 @@ const RulesHeader = () => {
   const { t } = useTranslation();
   return (
     <div className="row co-m-table-grid__head">
-      <div className="col-xs-3">{t('ingress~Host')}</div>
-      <div className="col-xs-3">{t('ingress~Path')}</div>
-      <div className="col-xs-3">{t('ingress~Service')}</div>
-      <div className="col-xs-2">{t('ingress~Service port')}</div>
+      <div data-i18n-translated className="col-xs-3">
```
spadgett, line 106: It would be better to match all table headers IMO instead of using `data-i18n-translated`. We can consider adding a data attribute for things we expect *not* to be translated, but we should assume by default everything is translated. You should have to opt out of the tests, not opt in.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed.  Added `beforeTest()` and `afterTest()` hooks

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
dtaylor113: fixed

frontend/public/components/ingress.jsx
dtaylor113: I was able to reduce the number of `data-i18n-translated` by using the `dt` selector, however there are still a few examples of translated text in various `div`'s, `span`'s, and `p`'s;  I'm not sure how to programatically select, or omit, those to only isolate those examples.

dtaylor113: In other words, how can you tell if a `div` is showing a transalted label or data?

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,173 @@ describe('Localization', () => {
...
-  it('pseudolocalizes navigation', () => {
-    cy.log('test navigation');
-    cy.visit('/dashboards?pseudolocalization=true&lng=en');
-    cy.byTestID('nav').isPseudoLocalized();
-  });
+  const createExampleResource = () => {
+    let initialUrl;
+    cy.url().then((url) => {
+      initialUrl = url;
+    });
```
spadgett, line 33: Will this block so we don't have a race?

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,173 @@ describe('Localization', () => {
...
+    listPage.clickCreateYAMLbutton();
+    cy.byTestID('resource-sidebar').should('exist');
+    yamlEditor.isLoaded();
+    yamlEditor.clickSaveCreateButton();
+    cy.get(errorMessage).should('not.exist');
+    cy.then(() => cy.visit(initialUrl));
```
spadgett, line 39: It would be better to have this utility navigate to the list page then create the resource, then stop. Then the caller can just the visit the URL it needs when done.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,173 @@ describe('Localization', () => {
...
-    cy.byTestID('events-view-all-link').isPseudoLocalized();
-    cy.byTestID('events-pause-button').isPseudoLocalized();
-  });
+  const deleteExampleResource = (kind: string) => {
+    cy.url().then((url) => {
+      if (url.includes('example')) {
+        detailsPage.clickPageActionFromDropdown(`Delete ${kind}`);
+      } else {
+        listPage.rows.clickKebabAction('example', `Delete ${kind}`);
+      }
```
spadgett, line 48: We shouldn't assume that all sample resources have name `example`.

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,173 @@ describe('Localization', () => {
...
-    cy.visit('/dashboards?pseudolocalization=true&lng=en');
-    cy.byTestID('activity').isPseudoLocalized();
-    cy.byTestID('activity-recent-title').isPseudoLocalized();
-    cy.byTestID('ongoing-title').isPseudoLocalized();
-    cy.byTestID('events-view-all-link').isPseudoLocalized();
-    cy.byTestID('events-pause-button').isPseudoLocalized();
-  });
+  const deleteExampleResource = (kind: string) => {
+    cy.url().then((url) => {
+      if (url.includes('example')) {
```
spadgett, line 44: Let's not do URL sniffing. Why not just pass the name and kind to this utility and run an oc command to delete it?

frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts
```diff
@@ -14,38 +20,173 @@ describe('Localization', () => {
...
+  const sectionHeadings = '[data-test-section-heading]';
+  const detailLabels = 'dt';
+  const i18nTranslated = '[data-i18n-translated]';
 
-  it('pseudolocalizes navigation', () => {
-    cy.log('test navigation');
-    cy.visit('/dashboards?pseudolocalization=true&lng=en');
-    cy.byTestID('nav').isPseudoLocalized();
-  });
+  const createExampleResource = () => {
```
spadgett, line 29: I feel like this would be much easier to integrate resource-specific pseudo-localization checks into the CRUD tests so that we don't have to duplicate the logic of creating and removing resources.

frontend/public/components/ingress.jsx
spadgett: I would avoid trying to test every string on the page and focus on the common elements that should be translated for all pages. These tests I think are most valuable to catch new strings that weren't properly externalized.

dtaylor113:
Hi @spadgett, added WIP label to this PR as #7213 maybe an alternative.  PTAL  -thanks

dtaylor113:
Closed in favor of #7213

PR #6986, opened 2020-10-22
Bug 1889724: When LocalVolumeDiscovery CR is created form the LSO page , user must Disk tab

afreen23:
- The flag  is set only by OCS and gets undefined when OCS is not installed. The flag must be set by LSO plugin and imported in OCS.
- Created an action flag in LSO package to be set
- Looking for discovery CR rather than discoveryresult CR
- minor fix for empty state too

### DIFF
--- a/frontend/packages/local-storage-operator-plugin/src/features.ts
+++ b/frontend/packages/local-storage-operator-plugin/src/features.ts
@@ -1,0 +1,33 @@
+import { FeatureDetector } from '@console/plugin-sdk';
+import { K8sKind, k8sList } from '@console/internal/module/k8s';
+import { getAnnotations } from '@console/shared/src/selectors/common';
+import { setFlag, handleError } from '@console/internal/actions/features';
+
+export const OCSServiceModel: K8sKind = {
+  label: 'Storage Cluster',
+  labelPlural: 'Storage Clusters',
+  apiVersion: 'v1',
+  apiGroup: 'ocs.openshift.io',
+  plural: 'storageclusters',
+  abbr: 'OCS',
+  namespaced: true,
+  kind: 'StorageCluster',
+  id: 'ocscluster',
+  crd: true,
+};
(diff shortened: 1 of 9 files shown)


bipuladh:
Please attach screenshots of the fix.

frontend/packages/local-storage-operator-plugin/src/features.ts
```diff
@@ -0,0 +1,33 @@
...
+    const storageCluster = storageClusters.find((sc) => sc.status.phase !== 'Ignored');
+    const isAttachedDevicesCluster =
+      getAnnotations(storageCluster)?.[ATTACHED_DEVICES_ANNOTATION] === 'true';
+    dispatch(setFlag(OCS_ATTACHED_DEVICES_FLAG, isAttachedDevicesCluster));
+  } catch (err) {
+    err?.response?.status === 404
+      ? dispatch(setFlag(OCS_ATTACHED_DEVICES_FLAG, false))
+      : handleError(err, OCS_ATTACHED_DEVICES_FLAG, dispatch, detectOCSAttachedDeviceMode);
+  }
+};
```
bipuladh, line 33: LSO should not be dependent on OCS. If this is indeed the case, then the plugins that are passed via LSO that use above flag should be passed via the `ceph` plguin. And since ceph is dependent on LSO we can in turn import their flag which are being used for the `AlertActions` below.

afreen23: Sorry, I failed to understand, whats the dependency here ?

afreen23: > LSO should not be dependent on OCS. 
Sorry, I failed to understand, whats the dependency here ?
> If this is indeed the case, then the plugins that are passed via LSO that use above flag should be passed via the `ceph` plguin.
It was like that only and we need to do this change, since it was causing issues, if you check the PR.
> And since ceph is dependent on LSO we can in turn import their flag which are being used for the `AlertActions` below.
Not quite understood.

PR #6990, opened 2020-10-23
Bug 1887864: Update library-go and replace runc module for v1.0.0-rc8 version

jhadvig:
/assign @spadgett 

cc'ing @pedjak for additional review


go.mod
```diff
@@ -30,7 +30,8 @@ require (
 replace (
 	// ww/goautoneg repo does not exist on butbucket anymore, hence we
 	// need to use a mirror. It can be removed when we update to newer libary-go
-	bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822
+	// bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822
```
jhadvig, line 33: @pedjak double checking if its ok to remove

pedjak: I think it is.. otherwise, you would not be able to do perform `go mod vendor`

go.mod
```diff
@@ -33,4 +34,5 @@ replace (
 	bitbucket.org/ww/goautoneg => github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822
```
pedjak, line 34: this line (and the comment) can be removed now, given that library-go is getting updated in this PR

jhadvig: @pedjak comment addressed

spadgett (approved):
/lgtm

PR #6991, opened 2020-10-23
Made singulars for Insights plugin

bond95:
Fixed `issues` label for single issue. PTAL @vojtechszocs @spadgett @doruskova

image
image

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -36,7 +36,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
                 y: v,
               }))}
               title={`${numberOfIssues}`}
-              subTitle="Total issues"
+              subTitle={`Total ${numberOfIssues === 1 ? 'issue' : 'issues'}`}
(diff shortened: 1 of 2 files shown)


frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -36,7 +36,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
                 y: v,
               }))}
               title={`${numberOfIssues}`}
-              subTitle="Total issues"
+              subTitle={`Total ${numberOfIssues > 1 ? 'issues' : 'issue'}`}
```
spadgett, line 39:
```suggestion
              subTitle={`Total ${numberOfIssues === 1 ? 'issue' : 'issues'}`}
```

This is better in general since you want "0 issues" if it's ever 0. I realize that `hasIssues` should be false here, but I'd rather code defensively in case that ever changes.

spadgett (commented):
lgtm, just one nit

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
bond95: Done

spadgett (approved):
/lgtm

PR #6995, opened 2020-10-26
Add Owner Reference to VM Snapshots

glekner:


glekner:
@suomiy can you lgtm

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
```diff
@@ -79,4 +80,22 @@ export abstract class K8sResourceWrapper<
     }
     return (this as any) as SELF;
   };
+
+  addOwnerReferences = (...additionalOwnerReferences) => {
+    if (!getOwnerReferences(this.data)) {
```
suomiy, line 85: you can use `ensurePath` function instead

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
```diff
@@ -79,4 +80,22 @@ export abstract class K8sResourceWrapper<
     }
     return (this as any) as SELF;
   };
+
+  addOwnerReferences = (...additionalOwnerReferences) => {
```
suomiy, line 84: can you please add a type?

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
```diff
@@ -79,4 +80,22 @@ export abstract class K8sResourceWrapper<
     }
     return (this as any) as SELF;
   };
+
```
suomiy, line 83: we can add this function to K8sResourceObjectWithTypePropertyWrapper as well for consistency

suomiy (commented):
Are you planning to add the message to the VM delete modal in another PR?

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
glekner: `ensurePath` doesn't initialize `ownerReferences` as an array. `getOwnerReferences(this.data)` returns `{}` if there are no owners, resulting in errors inside the array functions

glekner:
@suomiy applied.
> Are you planning to add the message to the VM delete modal in another PR?
this should go through designers yes

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
suomiy: please look at the implementation, it is possible to pass the desired structure

suomiy:
> > Are you planning to add the message to the VM delete modal in another PR?
> 
> this should go through designers yes

I think we already got response from designers in the snapshot thread. Let's just add a simple message to the delete vm modal. And we can ping designers in that PR and ask them what they think.

glekner:
image

Added message to delete vm modal
@suomiy @matthewcarleton

frontend/packages/kubevirt-plugin/src/k8s/wrapper/common/k8s-resource-wrapper.ts
glekner: updated

matthewcarleton:
Looks good but can you adjust the text to be
**Warning**: All snapshots of this virtual machine will be deleted as well.

matthewcarleton (approved):
LGTM!

PR #6996, opened 2020-10-26
Bug 1891419: Wrong detail head on network policy detail page

jhadvig:
/assign @spadgett


spadgett (approved):
/lgtm

dtaylor113 (approved):
I was going to just update this in my i18n networking pages open PR, but this is already in place and approved.

spadgett:
> I was going to just update this in my i18n networking pages open PR, but this is already in place and approved.

We probably want to backport, so I'd keep it as a separate PR.

PR #6878, opened 2020-10-08
Bug 1885706:  Cypress: Fix 'link-name' accesibility violation

dtaylor113:
Added `aria-label`'s to Cluster Dashboard Utilization chart's links.

**Before:**
```
 Running:  crud/other-routes.spec.ts                                                       (1 of 1)
1 accessibility violation was detected for / page
┌─────────┬─────────────┬───────────┬───────────────────────────────────────┬───────┐
│ (index) │     id      │  impact   │              description              │ nodes │
├─────────┼─────────────┼───────────┼───────────────────────────────────────┼───────┤
│    0    │ 'link-name' │ 'serious' │ 'Ensures links have discernible text' │   5   │
└─────────┴─────────────┴───────────┴───────────────────────────────────────┴───────┘
    ✓ successfully displays view for route: / (7398ms)
```
**After:**
```
  Running:  crud/other-routes.spec.ts                                                       (1 of 1)
    ✓ successfully displays view for route: / (7400ms)
    ✓ successfully displays view for route: /k8s/cluster/clusterroles/view (6159ms)
    ✓ successfully displays view for route: /k8s/cluster/nodes (5539ms)
    ✓ successfully displays view for route: /k8s/all-namespaces/events (5554ms)
```

PR #6997, opened 2020-10-26
Bug 1891376: Remove title passed to the PrometheusGraph component in the AreaChart

jhadvig:
Was introduced in https://github.com/openshift/console/pull/6878/files

/assign @spadgett 
cc'ing @dtaylor113


spadgett:
@jhadvig Does this still pass the accessibility check?

spadgett:
@jhadvig You'll need to revert the unit test change as well.

@dtaylor113 It looks like this is passing the e2e tests... Was your original change to fix a problem found during the e2e tests, or was it an accessibility issue found with other testing?

jhadvig:
Closing the PR per discussion with @dtaylor113

PR #7003, opened 2020-10-27
Bug 1889529: Check for any CR name versus a specific name

zherman0:
Current code looks for specific name from initialization resource metadata:name value however any named custom resource of that specific type will fulfil the requirements.  In the case of this bug, we were looking for `storagecluster-example` but OCS names the storage cluster `ocs-storagecluster.`
We now have a watch that looks for any specific custom resource to be created in order to trigger us to turn off the alert saying it needs to be created.


spadgett (approved):
/lgtm

PR #7004, opened 2020-10-27
Bug 1812412: Monitoring: Increase Prometheus query_range timeouts to 30s

kyoto:
The current timeout of 5 seconds is probably much too strict.

Prometheus has a maximum concurrent queries limit, which is currently
set to 20. This should protect against overloading Prometheus, even if
we increase the timeout significantly.

Also, if there are more than 20 queries, some will be queued and the
time they spend in the queue will count towards the time limit, which is
another reason to make the timeout less strict.


spadgett (approved):
/lgtm

PR #7005, opened 2020-10-27
check for taskRef kind in the TR spec

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5031

**Root Analysis:**
The Task col in the TaskRun list is always considering the taskRef to be of kind `Task`. So, in a scenario where TaskRun references ClusterTask, the linked task name gives a 404 error

**Solution Description:**
added a check for the taskRef kind

**GIF:**
clkind

### DIFF
--- a/frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
+++ b/frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
@@ -2,14 +2,13 @@ import * as React from 'react';
 import { TableRow, TableData, RowFunction } from '@console/internal/components/factory';
 import { ResourceLink, Timestamp, Kebab, ResourceKebab } from '@console/internal/components/utils';
 import { referenceForModel } from '@console/internal/module/k8s';
-import { TaskRunKind } from '../../../utils/pipeline-augment';
-import { TaskRunModel, PipelineModel, TaskModel } from '../../../models';
+import { TaskRunKind, getModelReferenceFromTaskKind } from '../../../utils/pipeline-augment';
+import { TaskRunModel, PipelineModel } from '../../../models';
 import { tableColumnClasses } from './taskruns-table';
 import { Status } from '@console/shared';
 import { pipelineRunFilterReducer as taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
 
 const taskRunsReference = referenceForModel(TaskRunModel);
-const taskReference = referenceForModel(TaskModel);
 const pipelineReference = referenceForModel(PipelineModel);
 
 const TaskRunsRow: RowFunction<TaskRunKind> = ({ obj, index, key, style, ...props }) => (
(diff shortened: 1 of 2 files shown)


frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
```diff
@@ -3,70 +3,75 @@ import { TableRow, TableData, RowFunction } from '@console/internal/components/f
...
-      <ResourceLink kind="Namespace" name={obj.metadata.namespace} />
-    </TableData>
-    {props.customData?.showPipelineColumn && (
-      <TableData className={tableColumnClasses[2]}>
-        {obj.metadata.labels['tekton.dev/pipeline'] ? (
+const TaskRunsRow: RowFunction<TaskRunKind> = ({ obj, index, key, style, ...props }) => {
+  const taskReference =
+    obj.spec.taskRef?.kind === ClusterTaskModel.kind
+      ? referenceForModel(ClusterTaskModel)
+      : referenceForModel(TaskModel);
```
jerolimov, line 18: Can we keep the result of referenceForModel outside of this Row function? So that it's not required to recalculate the kind for each row? It keeps the references consts also at one place.

jerolimov (changes requested):
Tested it locally, works fine. But I have one small change request.

frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
debsmita1: done. PTAL

jerolimov (approved):
Tested this locally, works fine :+1: 

/lgtm

frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
```diff
@@ -42,7 +43,9 @@ const TaskRunsRow: RowFunction<TaskRunKind> = ({ obj, index, key, style, ...prop
     <TableData className={tableColumnClasses[3]}>
       {obj.spec.taskRef?.name ? (
         <ResourceLink
-          kind={taskReference}
+          kind={
+            obj.spec.taskRef?.kind === ClusterTaskModel.kind ? clusterTaskReference : taskReference
```
andrewballantyne, line 47: So we already have this kinda check ... but not quite this _exact_ check in the [pipeline-augment.ts](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/utils/pipeline-augment.ts#L478-L490) file. Perhaps we just expand this with another wrapper function:

Ie, add this:
```
const getModelReferenceFromTaskKind = (kind: string): GroupVersionKind => {
  const model = getResourceModelFromTaskKind(kind);
  return referenceForModel(model);
}
```

andrewballantyne (changes requested):
This looks like it will work fine, but I think we can go one step further and place a new utility with the others like it.

frontend/packages/dev-console/src/components/taskruns/list-page/TaskRunsRow.tsx
andrewballantyne: So sorry for the confusion... please place this utility with the others in the `pipeline-augment.ts` file noted above.

andrewballantyne:
Can we [co-locate the utility](https://github.com/openshift/console/pull/7005#discussion_r516967670)? Sorry for the confusion.

andrewballantyne (approved):
/lgtm

PR #7006, opened 2020-10-27
Add descriptions to flavor, workload dropdowns

glekner:
Flavor:

flavor

Workload:
workload

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
@@ -16,7 +16,12 @@ import {
   iGetName,
   iGetNamespace,
 } from '../../selectors/immutable/selectors';
-import { iGetRelevantTemplate } from '../../../../selectors/immutable/template/combined';
+import {
+  iGetDefaultTemplate,
+  iGetRelevantTemplate,
+  getITemplateDefaultFlavor,
+  getITemplateDefaultWorkload,
+} from '../../../../selectors/immutable/template/combined';
 import {
   CUSTOM_FLAVOR,
   TEMPLATE_DATAVOLUME_NAME_PARAMETER,
@@ -85,6 +90,31 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
       { isHidden: asHidden(!isWindows, VMSettingsField.OPERATING_SYSTEM), value: isWindows },
     ),
   );
+
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iDefaultTemplate = iCommonTemplates && iGetDefaultTemplate(iCommonTemplates, os);
+  const defaultFlavor = getITemplateDefaultFlavor(iDefaultTemplate);
+  const defaultWorkload = getITemplateDefaultWorkload(iDefaultTemplate);
+
+  if (defaultFlavor) {
+    dispatch(
+      vmWizardInternalActions[InternalActionType.UpdateVmSettingsField](
+        id,
+        VMSettingsField.FLAVOR,
+        { value: defaultFlavor.getValue() },
+      ),
+    );
+  }
+
+  if (defaultWorkload) {
+    dispatch(
+      vmWizardInternalActions[InternalActionType.UpdateVmSettingsField](
+        id,
+        VMSettingsField.WORKLOAD_PROFILE,
+        { value: defaultWorkload.getValue() },
+      ),
+    );
+  }
 };
 
 const baseImageUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
@@ -1,4 +1,3 @@
-import * as _ from 'lodash';
 import * as React from 'react';
 import { SelectOption } from '@patternfly/react-core';
 import { ValidationErrorType, asValidationObject } from '@console/shared/src/utils/validation';
@@ -12,7 +11,9 @@ import {
 import { FormFieldRow } from '../../form/form-field-row';
 import { FormField, FormFieldType } from '../../form/form-field';
 import {
+  getFlavorLabel,
   getFlavors,
+  getOsDefaultTemplate,
   getWorkloadProfiles,
 } from '../../../../selectors/vm-template/combined-dependent';
 import { flavorSort } from '../../../../utils/sort';
@@ -20,8 +21,13 @@ import { VMSettingsField } from '../../types';
 import { iGetFieldValue } from '../../selectors/immutable/field';
 import { nullOnEmptyChange } from '../../utils/utils';
 import { FormPFSelect } from '../../../form/form-pf-select';
+import { Flavor } from '../../../../constants/vm/flavor';
+import { getCPU, getFlavor, getMemory } from '../../../../selectors/vm';
+import { getFlavorText } from '../../../../selectors/vm/flavor-text';
+import { selectVM } from '../../../../selectors/vm-template/basic';
+import { getLabelValue } from '../../../../selectors/selectors';
 
-export const Flavor: React.FC<FlavorProps> = React.memo(
+export const FlavorSelect: React.FC<FlavorProps> = React.memo(
   ({
     iUserTemplate,
     cnvBaseImages,
@@ -47,7 +53,16 @@ export const Flavor: React.FC<FlavorProps> = React.memo(
         : []
       : immutableListToShallowJS(iGetLoadedData(commonTemplates));
 
+    const defaultTemplate = getOsDefaultTemplate(templates, os);
+
     const flavors = flavorSort(getFlavors(templates, params));
+    const flavorDescriptions = templates.reduce((acc, t) => {
+      acc[getFlavor(t)] = getFlavorText({
+        cpu: getCPU(selectVM(t)),
+        memory: getMemory(selectVM(t)),
+      });
+      return acc;
+    }, {});
 
     const workloadProfiles = getWorkloadProfiles(templates, params);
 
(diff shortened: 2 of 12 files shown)


frontend/packages/kubevirt-plugin/src/constants/vm/workload-profile.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class WorkloadProfile extends SelectDropdownObjectEnum<string> {
+  static readonly DESKTOP = new WorkloadProfile('desktop', {
+    label: 'Desktop',
+    description: WORKLOAD_PROFILE_DESKTOP_DESC,
+    order: 1,
+  });
+  static readonly SERVER = new WorkloadProfile('server', {
+    label: 'Server',
+    isdefault: true,
```
suomiy, line 18: the defaults depend on the exact template (os) * and should not behard coded like this

* https://github.com/kubevirt/common-templates/blob/master/generate-templates.yaml

frontend/packages/kubevirt-plugin/src/constants/vm/flavor.ts
```diff
@@ -0,0 +1,52 @@
...
+    order: 1,
+  });
+  static readonly SMALL = new Flavor('small', {
+    label: 'Small',
+    description: FLAVOR_SMALL_DESC,
+    order: 2,
+  });
+  static readonly MEDIUM = new Flavor('medium', {
+    label: 'Medium',
+    isdefault: true,
```
suomiy, line 25: same here with the default

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -15,8 +15,9 @@ import { VMSettingsField } from '../../types';
 import { nullOnEmptyChange } from '../../utils/utils';
 import { iGetFieldValue } from '../../selectors/immutable/field';
 import { FormPFSelect } from '../../../form/form-pf-select';
+import { WorkloadProfile } from '../../../../constants/vm/workload-profile';
 
-export const WorkloadProfile: React.FC<WorkloadProps> = React.memo(
+export const WorkloadProfileComponent: React.FC<WorkloadProps> = React.memo(
```
suomiy, line 20: let's use some better name :) What about `WorkloadSelect` ?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -20,8 +19,11 @@ import { VMSettingsField } from '../../types';
 import { iGetFieldValue } from '../../selectors/immutable/field';
 import { nullOnEmptyChange } from '../../utils/utils';
 import { FormPFSelect } from '../../../form/form-pf-select';
+import { Flavor } from '../../../../constants/vm/flavor';
+import { getFlavor } from '../../../../selectors/vm';
+import { getTemplateFlavorDesc } from '../../../../selectors/vm-template/selectors';
 
-export const Flavor: React.FC<FlavorProps> = React.memo(
+export const FlavorComponent: React.FC<FlavorProps> = React.memo(
```
suomiy, line 26: name

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -87,13 +93,21 @@ export const Flavor: React.FC<FlavorProps> = React.memo(
...
+              .sort((a, b) => a.getOrder() - b.getOrder())
+              .map((f) => {
+                return (
+                  <SelectOption
+                    key={f.getValue()}
+                    value={f.getValue()}
+                    description={f.getDescription()}
+                  >
+                    {f.toString()}
+                    {f !== Flavor.CUSTOM && ` - ${flavorDescriptions?.[f.getValue()]}`}
```
suomiy, line 107: let's resolve the value first and then ask if we found it, so we don't end up with dangling ` - `

frontend/packages/kubevirt-plugin/src/selectors/vm-template/selectors.ts
```diff
@@ -41,3 +45,11 @@ export const getTemplateValidationsFromTemplate = (
     return new TemplateValidations();
   }
 };
+
+export const getTemplateMemory = (template: TemplateKind): string => {
+  const baseMemoryValue = convertToBaseValue(getMemory(selectVM(template)));
+  return humanizeBinaryBytes(baseMemoryValue).string;
+};
+
+export const getTemplateFlavorDesc = (template: TemplateKind): string =>
```
suomiy, line 54: there is a function called getFlavorText. Can you please modify it so the flavor would be optional there?. 
Also it is not a good idea to add such specific functions like getTemplateMemory to the template since template is a generic concept and this is a domain of a VM. Eg it is easier to create a VMWrapper and all the functions are there

suomiy (changes requested):

glekner:
@suomiy fixed suggestions

suomiy:
looks good, but still needs the correct default behaviour

glekner:
@suomiy from what i'm seeing there is no default workload/flavor in the common templates, only supported ones. please correct me otherwise

suomiy:
@glekner  this is what I see when I generate the templates from master

```
apiVersion: template.openshift.io/v1
kind: Template
metadata:
  name: centos6-server-small-v0.11.3
  ...
  labels:
    ...
    template.kubevirt.io/default-os-variant: "true"

```

it depends on the os which flavor and workload has it (should be just one per os)

glekner:
* Now detecting default Workload/Flavor based on the default template per OS.
* Added the detection to `osUpdater`, now Workload and Flavor are preselected on OS change if a default value is available.

@suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -92,13 +111,25 @@ export const Flavor: React.FC<FlavorProps> = React.memo(
...
-                </SelectOption>
-              );
-            })}
+            {flavors
+              .map(Flavor.fromString)
+              .sort((a, b) => a.getOrder() - b.getOrder())
+              .map((f) => {
+                const flavorDesc = flavorDescriptions?.[f.getValue()];
+                const isDefault =
+                  getLabelValue(defaultTemplate, `${TEMPLATE_FLAVOR_LABEL}/${f.getValue()}`) ===
```
suomiy, line 120: we can use getFlavorLabel here

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/workload-profile.tsx
```diff
@@ -66,9 +79,25 @@ export const WorkloadProfile: React.FC<WorkloadProps> = React.memo(
...
-                return <SelectOption key={workloadProfile} value={workloadProfile} />;
-              })}
+              {(workloadProfiles || [])
+                .map(WorkloadProfile.fromString)
+                .sort((a, b) => a.getOrder() - b.getOrder())
+                .map((workload) => {
+                  const isDefault =
+                    getLabelValue(
+                      defaultTemplate,
+                      `${TEMPLATE_WORKLOAD_LABEL}/${workload.getValue()}`,
```
suomiy, line 89: we can use getWorkloadLabel

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
suomiy: maybe we could create similar functions like we have for immutable

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
```diff
@@ -99,3 +104,24 @@ export const iGetCommonTemplateCloudInit = (tmp: ITemplate) => {
...
+  os: string,
+): ITemplate =>
+  (iCommonTemplates || ImmutableMap()).valueSeq().find((iTemplate) => {
+    const labels = iGetLabels(iTemplate);
+    return labels && labels.get(getOsLabel(os)) && labels.get(TEMPLATE_DEFAULT_LABEL) === 'true';
+  });
+
+export const getITemplateDefaultFlavor = (template: ITemplate): Flavor =>
+  template &&
+  Flavor.getAll().find(
```
suomiy, line 119: there is no need to go all over them - there is just one flavor per template

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
```diff
@@ -99,3 +104,24 @@ export const iGetCommonTemplateCloudInit = (tmp: ITemplate) => {
...
+    (f) => iGetLabels(template).get(`${TEMPLATE_FLAVOR_LABEL}/${f.getValue()}`) === 'true',
+  );
+
+export const getITemplateDefaultWorkload = (template: ITemplate): WorkloadProfile =>
```
suomiy, line 123: the same here. Also this can be generic - not only applying for default

glekner:
@suomiy added safety check for outdated templates

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/flavor.tsx
```diff
@@ -47,7 +54,19 @@ export const Flavor: React.FC<FlavorProps> = React.memo(
         : []
       : immutableListToShallowJS(iGetLoadedData(commonTemplates));
 
+    const defaultTemplate = templates.find(
```
suomiy, line 57: i suppose we could make a function of this as well

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -85,6 +90,31 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
       { isHidden: asHidden(!isWindows, VMSettingsField.OPERATING_SYSTEM), value: isWindows },
     ),
   );
+
+  const iCommonTemplates = iGetLoadedCommonData(state, id, VMWizardProps.commonTemplates);
+  const iDefaultTemplate = iCommonTemplates && iGetDefaultTemplate(iCommonTemplates, os);
+  const defaultFlavor = getITemplateDefaultFlavor(iDefaultTemplate);
+  const defaultWorkload = getITemplateDefaultWorkload(iDefaultTemplate);
+
+  if (defaultFlavor) {
```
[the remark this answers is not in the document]
glekner: what do you mean at the beginning? the os field is blank in the beginning. so the only way to update the flavour/workload is when it changes? and changing it only once is not a good option IMO.
There is no good reason for the user to pick a specific flavor and then randomly change the OS. If he does, we can live with the fact he needs to refill it again.

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
glekner: and how would you detect the right one?

the label is `flavor.template.kubevirt.io/FLAVOR`
we need to iterate through all of them and find the one who matches.

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
glekner: https://github.com/openshift/console/pull/7006#discussion_r516639243

glekner:
fixed + commented @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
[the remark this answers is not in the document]
suomiy: ok, I don't mind - it might be just annoying in some instances

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
suomiy: it is a map with well defined key so it should be easy to detect

suomiy: sorry, you are right, this is the correct solution. I mistakenly had in my mind that we know the flavor/workload

PR #7013, opened 2020-10-28
Create an extension to support inContext for import Flows

sahil143:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5050

**Analysis / Root cause**: Need a generic way to support the InContext for all import flows	

**Solution Description**: Create an extension to support inContext for import Flows	

**Unit test coverage report**: done

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/post-form-submit.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/post-form-submit.ts
@@ -1,0 +1,19 @@
+import { Extension } from './base';
+
+namespace ExtensionProperties {
+  export interface PostFormSubmissionAction<T, P> {
+    /** action type */
+    type: string;
+    /** callback for the related action */
+    callback: (arg: T, payload?: P) => void;
+  }
+}
+
+export interface PostFormSubmissionAction<R = any, P = string>
+  extends Extension<ExtensionProperties.PostFormSubmissionAction<R, P>> {
+  type: 'PostFormSubmissionAction';
+}
+
+export const isPostFormSubmissionAction = (e: Extension): e is PostFormSubmissionAction => {
+  return e.type === 'PostFormSubmissionAction';
+};
--- a/frontend/packages/console-shared/src/hooks/post-form-submit-action.ts
+++ b/frontend/packages/console-shared/src/hooks/post-form-submit-action.ts
@@ -1,0 +1,58 @@
+import { useCallback, useMemo } from 'react';
+import {
+  isPostFormSubmissionAction,
+  PostFormSubmissionAction,
+  useExtensions,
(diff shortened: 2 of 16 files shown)


frontend/packages/console-plugin-sdk/src/typings/post-form-submit.ts
```diff
@@ -0,0 +1,19 @@
+import { Extension } from './base';
+
+namespace ExtensionProperties {
+  export interface PostFormSubmissionAction {
+    /** action type */
+    type: string;
+    /** callback for the related action */
+    callback: (...args) => void;
```
christianvogt, line 8: This should probably be typed as an array of resources since the caller must know what to provide the actions.

sahil143: Fixed, added proper type for extension and hook.

frontend/packages/console-shared/src/hooks/post-form-submit-action.ts
```diff
@@ -0,0 +1,53 @@
...
+import { useQueryParams } from './useQueryParams';
+
+export const usePostFormSubmitAction = <R = K8sResourceCommon[]>(
+  key: string = 'action',
+): ((arg: R) => void) => {
+  const params = useQueryParams();
+  const actionParams = params.get(key);
+
+  const [actionTypes, actionPayload] = useMemo(() => {
+    const parsedAction = JSON.parse(actionParams);
```
invincibleJai, line 17: should we have `JSON.parse`  in try catch block or it's always safe?

sahil143: I think it's safe to use it like this as any action that is passed will be stringified.

christianvogt: Since the data is coming from a query string, I think it's a good idea to catch the error.

christianvogt (commented):
@sahil143 i couldn't test this as it seems in context isn't working to show the menu item even on master.

christianvogt:
@jeff-phillips-18 fixed the in context issue with https://github.com/openshift/console/pull/7093 which allowed me to test your PR

christianvogt:
/approve

@sahil143 please see the comment about json.parse

frontend/packages/console-shared/src/hooks/post-form-submit-action.ts
sahil143: fixed!

invincibleJai:
/lgtm

Verified the changes

PR #7014, opened 2020-10-28
use FocusTrap for popover and remount StepComponent to remove the focus from Next button

sahil143:
**Fixes**: 

https://issues.redhat.com/browse/ODC-4722
https://issues.redhat.com/browse/ODC-4919

**Analysis / Root cause**: Color of Next button was changed because once next button is clicked it remains focused.

**Solution Description**: Remount the component so focus is not on the Next Button

focus-trap

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/popover/Popover.tsx
+++ b/frontend/packages/console-shared/src/components/popover/Popover.tsx
@@ -1,13 +1,14 @@
 import * as React from 'react';
-import { Popper } from '../popper';
-import styles from '@patternfly/react-styles/css/components/Popover/popover';
 import { css } from '@patternfly/react-styles';
+import { FocusTrap } from '@patternfly/react-core';
+import styles from '@patternfly/react-styles/css/components/Popover/popover';
 import { PopoverArrow } from '@patternfly/react-core/dist/js/components/Popover/PopoverArrow';
 import { PopoverBody } from '@patternfly/react-core/dist/js/components/Popover/PopoverBody';
 import { PopoverCloseButton } from '@patternfly/react-core/dist/js/components/Popover/PopoverCloseButton';
 import { PopoverContent } from '@patternfly/react-core/dist/js/components/Popover/PopoverContent';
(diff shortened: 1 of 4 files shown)


lwrigh:
This looks good from UX perspective!

rohitkrai03 (approved):
/approve

frontend/packages/console-shared/src/components/popover/Popover.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Popper } from '../popper';
+import * as FocusTrap from 'focus-trap-react';
```
christianvogt, line 2: Use focus trap from PF:
```suggestion
import * as FocusTrap from '@patternfly/react-core';
```

andrewballantyne: Agree, there are only a few other places that use FocusTrap from this external lib, we should get rid of them and drop the dependency.

andrewballantyne:
@sahil143 Could you squash your commits?

andrewballantyne:
LGTM, retested it with the PF lib and operates the same. Just need the commits to be squashed and then we get this merged.

frontend/packages/console-shared/src/components/popover/Popover.tsx
sahil143: fixed in other places as well

sahil143: Not sure if we want to remove the dependency as well from package.json

andrewballantyne: @christianvogt Did you want to consult with the Admin team?

@sahil143 Did you test the other areas? Because you've just created an extra step before I can lgtm this lol... Will have to wait until I can test the other areas.

PR #7016, opened 2020-10-28
Fix po-to-i18n script

rebeccaalpert:
I didn't realize recent file structure changes broke the script. I fixed that and added better handling for nonexistent directories and bad file names. I also updated .eslintignore so we could remove "ignore me linter!" comments.

### DIFF
--- a/frontend/i18n-scripts/.eslintrc.js
+++ b/frontend/i18n-scripts/.eslintrc.js
@@ -1,0 +1,68 @@
+var fs = require('fs');
+
+module.exports = {
+  root: true,
+  env: {
+    browser: true,
+    es6: true,
+    jasmine: true,
+    jest: true,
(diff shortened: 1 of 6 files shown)


frontend/.eslintignore
```diff
@@ -7,3 +7,4 @@ public/lib
 Godeps
 @types
 dynamic-demo-plugin
+i18n-scripts
```
spadgett, line 10: We probably want to keep linting, just use a different eslintrc in the i18n-scripts folder with different rules.

rebeccaalpert: I moved in an eslintrc file; I'm still going through there and removing things that aren't necessary for this folder.

spadgett (approved):
/lgtm

PR #7019, opened 2020-10-28
Bug 1890235: update Protractor's checkErrors logging

dtaylor113:
Want to display better info than `omg js error: [object Object]`:

**Before:**
```
Failed Specs:
1) Alertmanager: Configuration : launches Alert Routing modal, edits and saves correctly
   Error: Failed: omg js error: [object Object]
```
**After:**
```
Protractor:

yarn run test-protractor-suite --suite dashboards --params.openshift true

Failed Specs:

1) Cluster Dashboard Details Card : has all fields populated
   Error: Failed: Missing i18n key "Activity" in namespace "dashboard" and language "en."
       at protractor_1.browser.executeScript.then /home/dtaylor/repos/openshift/origin/src/github.com/openshift/console/frontend/integration-tests/protractor.conf.ts:261:13
       ...
       ...

Cypress

./test-cypress.sh -p console -s 'tests/crud/other*' -h true

 1) Visiting other routes
       "after each" hook for "successfully displays view for route: /":
     AssertionError: Missing i18n key "Activity" in namespace "dashboard" and language "en.": expected false to be true


```

### DIFF
--- a/frontend/integration-tests/protractor.conf.ts
+++ b/frontend/integration-tests/protractor.conf.ts
@@ -270,7 +270,7 @@ function hasError() {
 export const checkErrors = async () =>
   await browser.executeScript(hasError).then((err) => {
     if (err) {
-      fail(`omg js error: ${err}`);
+      fail(err);
     }
   });
 
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -230,14 +230,17 @@ graphQLReady.onReady(() => {
   // Used by GUI tests to check for unhandled exceptions
   window.windowError = null;
   window.onerror = (message, source, lineno, colno, error) => {
+    const { stack } = error;
+    const formattedStack = stack?.replace(/\\n/g, '\n');
+    window.windowError = `unhandled error: ${message} ${formattedStack || ''}`;
     // eslint-disable-next-line no-console
-    console.error('Uncaught error', error);
-    window.windowError = error;
+    console.error(window.windowError);
   };
   window.onunhandledrejection = (promiseRejectionEvent) => {
+    const { reason } = promiseRejectionEvent;
+    window.windowError = `unhandled promise rejection: ${reason}`;
     // eslint-disable-next-line no-console
-    console.error('Unhandled promise rejection', promiseRejectionEvent);
-    window.windowError = promiseRejectionEvent;
+    console.error(window.windowError);
   };
 
   if ('serviceWorker' in navigator) {
--- a/frontend/public/i18n.js
+++ b/frontend/public/i18n.js
@@ -137,11 +137,9 @@ i18n
       },
       saveMissing: true,
       missingKeyHandler: function(lng, ns, key) {
-        window.windowError = new Error(
-          `Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`,
-        );
+        window.windowError = `Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`;
         // eslint-disable-next-line no-console
-        console.error(`Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`);
+        console.error(window.windowError);
       },
     },
     () => {
--- a/frontend/@types/console/declarations.d.ts
+++ b/frontend/@types/console/declarations.d.ts
@@ -41,7 +41,7 @@ declare interface Window {
     GOOS: string;
     graphqlBaseURL: string;
   };
-  windowError?: Error | PromiseRejectionEvent;
+  windowError?: string;
   __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: Function;
   store?: {}; // Redux store
   pluginStore?: {}; // Console plugin store
--- a/frontend/packages/integration-tests-cypress/support/index.ts
+++ b/frontend/packages/integration-tests-cypress/support/index.ts
@@ -39,22 +39,9 @@ after(() => {
   });
 });
 
-const formatWindowError = (windowError: Error | PromiseRejectionEvent) => {
-  if (!windowError) {
-    return 'no window error detected';
-  }
-  const { reason } = windowError as PromiseRejectionEvent;
-  if (reason) {
-    return reason;
-  }
-  const { message, stack } = windowError as Error;
-  const formattedStack = stack?.replace(/\\n/g, '\n');
-  return `window error detected: ${message} ${formattedStack}`;
-};
-
 export const checkErrors = () =>
   cy.window().then((win) => {
-    assert.isTrue(!win.windowError, formatWindowError(win.windowError));
+    assert.isTrue(!win.windowError, win.windowError);
   });
 
 export const testName = `test-${Math.random()


frontend/public/i18n.js
```diff
@@ -112,11 +112,11 @@ i18n
       },
       saveMissing: true,
       missingKeyHandler: function(lng, ns, key) {
-        window.windowError = new Error(
-          `Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`,
-        );
+        const error = new Error();
+        error.message = `Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`;
+        window.windowError = error;
```
dtaylor113, line 117: I'm tempted to make this PR a WIP because I don't understand why the original way doesn't work.  Using the original `new Error(...)` constructor resulted in a defined, but empty `window.windowError` object when processed in updated `protractor.conf.ts`'s `formatWindowError`.
What's even weirder is I could add `error.name, error.fooBar` and those props would be there in `formatWindowError`, but `error.stack` would disappear!
I suspect it has something to do with `Error` typescripting..maybe..?  In my IDE when I click on `Error` it comes up with two different defs, one in `node`:
```
interface Error {
    stack?: string;
}
```
..and one in `typescript.es5`:
```
interface Error {
    name: string;
    message: string;
    stack?: string;
}
```
`stack` is in both defs, so really confused as to why it doesn't 'stick'!?
```

File: frontend/public/i18n.js [hunk 2]
```diff
@@ -112,11 +112,11 @@ i18n
...
         // eslint-disable-next-line no-console
-        console.error(`Missing i18n key "${key}" in namespace "${ns}" and language "${lng}."`);
+        console.error('Missing i18n key', error);
```
dtaylor113, line 119: I tried to `throw error` / `throw window.windowError`, but it doesn't get handled in [app.jsx](https://github.com/openshift/console/blob/ed2eb277b3d5bec45c2afda444f9644f2b38b189/frontend/public/components/app.jsx#L230) like Cypress does.
When I attempt to `throw` an error from `i18n.js`, protractor timesout with: 
```
(node:187007) UnhandledPromiseRejectionWarning: JavascriptError: javascript error: circular reference
  ...
  at Object.exports.checkErrors (/home/dtaylor/repos/openshift/origin/src/github.com/openshift/console/frontend/integration-tests/protractor.conf.ts:283:39)
(node:187007) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```
I would've thought the `UnhandledPromiseRejection` would've been caught [here](https://github.com/openshift/console/blob/ed2eb277b3d5bec45c2afda444f9644f2b38b189/frontend/public/components/app.jsx#L235) in app.jsx.   Is `i18n.js` "out-of-scope" of `app.jsx`?   `window.windowError` existed in `protractor.conf.ts` so I assumed other window errors and unhandledrejections work/worked.

dtaylor113:
Hi @TheRealJon & @jcaianirh,  wondering if you guys have any insight?   This was supposibly a simple enhancement, but I ran into issues I'm struggling with.  Any help appreciated :-)

frontend/integration-tests/protractor.conf.ts
```diff
@@ -264,13 +264,26 @@ export const checkLogs = async () => {
...
+};
+
 function hasError() {
   return window.windowError;
 }
 export const checkErrors = async () =>
   await browser.executeScript(hasError).then((err) => {
     if (err) {
-      fail(`omg js error: ${err}`);
+      fail(`${formatWindowError(err)}`);
```
spadgett, line 286:
```suggestion
      fail(formatWindowError(err));
```

spadgett (commented):
@dtaylor113 I'm wondering if it's not simpler to make `window.windowError` a string always and format it in the app itself (with stack trace). That way we know exactly what type we're dealing with in each event handler, and we don't have to duplicate the logic across protractor and cypress tests. And maybe it avoids the weirdness you're seeing in your comment below.

dtaylor113:
> @dtaylor113 I'm wondering if it's not simpler to make `window.windowError` a string always and format it in the app itself (with stack trace).

Great suggestion @spadgett!  So much simpler, better :-)

frontend/integration-tests/protractor.conf.ts
```diff
@@ -270,7 +270,7 @@ function hasError() {
 export const checkErrors = async () =>
   await browser.executeScript(hasError).then((err) => {
     if (err) {
-      fail(`omg js error: ${err}`);
+      fail(`${err}`);
```
spadgett, line 273:
```suggestion
      fail(err);
```

frontend/packages/integration-tests-cypress/views/modal.ts
```diff
@@ -5,7 +5,7 @@ export const modal = {
   shouldBeClosed: () => cy.byLegacyTestID('modal-cancel-action').should('not.be.visible'),
   submitShouldBeDisabled: () => cy.get(submitButton).should('be', 'disabled'),
   submitShouldBeEnabled: () => cy.get(submitButton).should('not.be', 'disabled'),
-  submit: () => cy.get(submitButton).click(),
+  submit: () => cy.get(submitButton).click({ force: true }),
```
spadgett, line 8: Why is force needed here?

frontend/public/components/app.jsx
```diff
@@ -230,14 +230,17 @@ graphQLReady.onReady(() => {
...
+    const formattedStack = stack?.replace(/\\n/g, '\n');
+    window.windowError = `window error detected: ${message} ${formattedStack || ''}`;
     // eslint-disable-next-line no-console
-    console.error('Uncaught error', error);
-    window.windowError = error;
+    console.error(window.windowError);
   };
   window.onunhandledrejection = (promiseRejectionEvent) => {
+    const { reason } = promiseRejectionEvent;
+    window.windowError = reason;
```
spadgett, line 241: Should this be something like

```suggestion
    window.windowError = `unhandled promise rejection: ${reason}`;
```

frontend/public/components/app.jsx
```diff
@@ -230,14 +230,17 @@ graphQLReady.onReady(() => {
   // Used by GUI tests to check for unhandled exceptions
   window.windowError = null;
   window.onerror = (message, source, lineno, colno, error) => {
+    const { stack } = error;
+    const formattedStack = stack?.replace(/\\n/g, '\n');
+    window.windowError = `window error detected: ${message} ${formattedStack || ''}`;
```
spadgett, line 235:
```suggestion
    window.windowError = `unhandled error: ${message} ${formattedStack || ''}`;
```

frontend/integration-tests/protractor.conf.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/views/modal.ts
dtaylor113: removed & retested, not needed.

frontend/public/components/app.jsx
dtaylor113: fixed

frontend/public/components/app.jsx
dtaylor113: fixed

spadgett (approved):
/lgtm

PR #7024, opened 2020-10-28
Bug 1884629: pf update to fix dropdown accessibility bug

jcaianirh:
Fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1884629

Contains fix for pf issue: https://github.com/patternfly/patternfly-react/issues/4946

During a call with the customer and bug submitter, we verified (via the pf examples site) the pf dropdown fix (which landed in the above pf release) has fixed the issue the user had with the openshift dev/admin context selector.


spadgett (approved):
/lgtm

jcaianirh:
fyi: @jessiehuff

PR #7030, opened 2020-10-29
Update i18n scripts

rebeccaalpert:
I updated the i18n-to-pot script to export pos rather than pot files based on the latest info from translation. I also added a bash script so we can pull out all pos programmatically.

i18n to PO script additions:
* Appended the language to the header 
* Added merging so existing translations are included in po file
* Updated the version of the converter library since they accepted my PR to fix their known wrapping issue: https://github.com/i18next/i18next-gettext-converter/pull/97
* Added foldLength: 0 option to converter library call so our text doesn't wrap
* File names from packages now have `_package=${package}` at the end so I can sort them programmatically
* Files are exported into frontend/po-files so I don't have to hunt through multiple folders (happy to put these somewhere else if this doesn't make sense)
* Can now export keys that contain underscores correctly

PO to i18n script additions:
* Now takes arguments (directory, language) instead of making assumptions
* User supplies directory they want parsed and the language they want to send files to
* File names no longer require the language
* File names from packages must include `_package=${package}` so it can sort them into the correct locale folders


frontend/i18n-scripts/i18n-to-po.js
```diff
@@ -1,15 +1,26 @@
...
 
-function save(target) {
-  return (result) => {
-    fs.writeFileSync(target, result);
-  };
+function addLanguageToFile(filePath, language) {
+  const data = fs.readFileSync(filePath, 'utf-8');
+  const arr = data.split('\n');
+  const index = arr.findIndex((i) => i.includes('Plural-Form'));
+  arr.splice(index, 0, `"Language: ${language}\\n"`);
```
spadgett, line 11: Set `options.language` instead of doing it this way:

https://github.com/i18next/i18next-gettext-converter/blob/master/lib/json2gettext.js#L106-L108

spadgett (changes requested):

frontend/i18n-scripts/i18n-to-po.js
rebeccaalpert: I think there's a bug. I'm already setting the language option here: https://github.com/openshift/console/pull/7030/files#diff-d6aa690af9a93d13d9031bfc763234fdf51c5083c5ed52db1cf376b9c4c3d154R59, but the header isn't generated.

spadgett: @rebeccaalpert But you aren't passing an `options` parameter to that call. You need a third parameter with `{ language: language }`

rebeccaalpert: Perfect! Missed that, thanks.

frontend/i18n-scripts/po-to-i18n.js
```diff
@@ -39,40 +24,59 @@ function processFile(fileName, package) {
...
+const options = {
+  string: ['language', 'file', 'directory'],
+  boolean: ['help'],
+  alias: {
+    h: 'help',
+    d: 'directory',
+    l: 'language',
+  },
+  default: {
+    language: 'ja',
```
spadgett, line 67: Why is the default `ja`?

frontend/i18n-scripts/i18n-to-po.js
```diff
@@ -0,0 +1,138 @@
+const fs = require('fs');
```
spadgett, line 1: It's hard to review this file since we've lost the diff due to the rename. Can you let us know what's changed?

rebeccaalpert: I outlined it in the description: https://github.com/openshift/console/pull/7030#issue-512414140

frontend/i18n-scripts/po-to-i18n.js
rebeccaalpert: All set!

rebeccaalpert:
Rebased package.json.

frontend/i18n-scripts/i18n-to-po.js
```diff
@@ -0,0 +1,146 @@
...
+    fs.mkdirSync(path.join(__dirname, `./../packages/${package}/locales/tmp`), {
+      recursive: true,
+    });
+
+    tmpFile = path.join(__dirname, `./../packages/${package}/locales/tmp/${fileName}.json`);
+
+    removeValues(i18nFile, tmpFile);
+    consolidateWithExistingTranslations(tmpFile, fileName, language, package);
+
+    fs.mkdirSync(path.join(__dirname, `./../public/locales/pos/${language}`), {
```
spadgett, line 70: The generated files should go into a different folder that has a gitignore

spadgett: A sibling folder to `frontend/public`. `public` is everything that we ship as part of the app itself.

rebeccaalpert: Thanks. Do you care what the new folder is named?

rebeccaalpert: @spadgett - I changed it to `frontend/po-files`; let me know if you want a different name. I added a `.gitignore`.

rebeccaalpert:
This is ready for review again @spadgett. It blocks the Korean work for next sprint and would make importing files from Terry easier.

frontend/po-files/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+*/
+!.gitignore
```
spadgett, line 3: I would move this to the top-level gitignore file

rebeccaalpert: All set @spadgett.

spadgett (approved):
/lgtm

PR #6905, opened 2020-10-12
add i18n in dev-console package

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4890

**Description**: 
i18n support for dev-console package

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #7034, opened 2020-10-29
Fix application dropdown to fetch resources based on extensions.

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3613

**Solution Description**: 
Use the topology plugin extensions to determine which resource kinds to fetch to then search for existing application groupings.

**Browser conformance**: 
- [x ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/dropdown/ApplicationDropdown.tsx
+++ b/frontend/packages/dev-console/src/components/dropdown/ApplicationDropdown.tsx
@@ -1,16 +1,14 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { Firehose } from '@console/internal/components/utils';
-import { referenceForModel } from '@console/internal/module/k8s';
-import { ServiceModel } from '@console/knative-plugin';
-import { VirtualMachineModel } from '@console/kubevirt-plugin/src/models';
-import { PodModel, JobModel, CronJobModel } from '@console/internal/models';
+import { FirehoseResult } from '@console/internal/components/utils';
 import { ResourceDropdown } from '@console/shared';
+import { useExtensions } from '@console/plugin-sdk/src';
 import {
-  getDynamicChannelResourceList,
-  getDynamicEventSourcesResourceList,
-} from '@console/knative-plugin/src/utils/fetch-dynamic-eventsources-utils';
-import { knativeEventingResourcesBroker } from '@console/knative-plugin/src/utils/get-knative-resources';
+  useK8sWatchResources,
+  WatchK8sResources,
+} from '@console/internal/components/utils/k8s-watch-hook';
+import { getBaseWatchedResources } from '../topology/data-transforms';
+import { isTopologyDataModelFactory, TopologyDataModelFactory } from '../../extensions/topology';
 
 interface ApplicationDropdownProps {
   id?: string;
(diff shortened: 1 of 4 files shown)


frontend/packages/dev-console/src/components/dropdown/ApplicationDropdown.tsx
```diff
@@ -41,86 +46,73 @@ interface ApplicationDropdownProps {
...
-        placeholder="Select an Application"
-        dataSelector={['metadata', 'labels', 'app.kubernetes.io/part-of']}
-      />
-    </Firehose>
+    <ResourceDropdown
+      {...props}
+      resources={resources}
+      loaded={loaded}
+      loadError={loadError}
+      placeholder={t('devconsole~Select application')}
```
debsmita1, line 106:
```suggestion
      placeholder={t('devconsole~Select an application')}
```
unsure if we need to take care of localization in this PR 'cause PR https://github.com/openshift/console/pull/6905 takes care of all the files under dev-console

jeff-phillips-18: afaik, the ask was to make sure any PRs do the i18n work as well in order to help ensure things are not missed.

christianvogt:
This works fine. However we are severely overfetching resources now.
There is no efficient way to properly identify all the applications within a namespace.
cc @serenamarie125 as this problem will continue to get worse.

frontend/packages/dev-console/src/components/dropdown/ApplicationDropdown.tsx
```diff
@@ -41,86 +46,73 @@ interface ApplicationDropdownProps {
...
+    }, watchedBaseResources);
+  }, [modelFactories, namespace]);
+
+  const watchResults = useK8sWatchResources(watchedResources);
+  React.useEffect(() => {
+    const resourcesLoaded =
+      !kindsInFlight &&
+      Object.keys(watchResults).length > 0 &&
+      Object.keys(watchResults).every(
+        (key) => watchResults[key].loaded || watchResults[key].loadError,
```
christianvogt, line 78: Do you need `kindsInFlight`? Seems like you can rely on `loaded` of the watched resources. `kindsInFlight` will include every resource that's currently in flight from anywhere on the page.

jeff-phillips-18: Initially loaded can return true with load errors for every resource fetched when `kindsInFlight` is true. This results in an error being shown briefly.

jeff-phillips-18: I've put a fix into `useK8sWatchResources` to not set the loaded flag to true when the model is not loaded and kinds are still in flight. @rawagner PTAL

christianvogt:
/approve

@rawagner please have a look at the change to `k8s-watch-hook.ts`

frontend/public/components/utils/k8s-watch-hook.ts
```diff
@@ -212,7 +212,7 @@ export const useK8sWatchResources = <R extends ResourcesObject>(
         } else {
           acc[key] = {
             data: resources[key].isList ? [] : {},
-            loaded: true,
+            loaded: modelsLoaded,
```
rawagner, line 215: I think that everytime we get here we can say that either all models are loaded or all requested models are loaded - https://github.com/openshift/console/blob/6fc7a2df6d190b1274196166f34d484748c9bfa7/frontend/public/components/utils/k8s-watch-hook.ts#L144 . That means its safe to say resource is not being loaded anymore and the model is missing. Or I missed something, do you have a case which I can try that causes the issue ?

`modelsLoaded` is also missing in dependency array https://github.com/openshift/console/blob/6fc7a2df6d190b1274196166f34d484748c9bfa7/frontend/public/components/utils/k8s-watch-hook.ts#L230

rawagner: @jeff-phillips-18 I guess I see the problem. Can you change `hasAllModelsLoaded` to

```
const hasAllModelsLoaded = React.useMemo(
    () => modelsLoaded || Object.values(resources).every((r) => k8sModels.includes(r.kind)),
    [k8sModels, modelsLoaded, resources],
  );
```

and revert other changes ? That might help

rawagner: @jeff-phillips-18 I've updated my PR https://github.com/openshift/console/pull/6464 with some more changes which should resolve your issues. Can you try ?

jeff-phillips-18: @rawagner I'll give that a try. I definitely saw instances where all models returned `loaded: true` here while `modelsLoaded` was false.

rawagner: @jeff-phillips-18 makes sense, `hasAllModelsLoaded` was broken because second part (after or operator) was just wrong.

jeff-phillips-18: Yes, that fixes the issue I was seeing. Thanks, I will remove my changes to `k8s-watch-hook.ts`

christianvogt:
@jeff-phillips-18 do we wait for the related bug to be fixed or are we sticking with this small fix to the k8s-watch-hook?

jeff-phillips-18:
> @jeff-phillips-18 do we wait for the related bug to be fixed or are we sticking with this small fix to the k8s-watch-hook?

No, we can go in without the change. Just can't update `TopologyDataRetriever` until the fix to the hook goes in.

jeff-phillips-18:
@christianvogt https://github.com/openshift/console/pull/6464 has merged. This PR should be ready now.

PR #7035, opened 2020-10-29
feature flag Helm based on HelmRepository CRs

debsmita1:
**JIRA Story:**
https://issues.redhat.com/browse/ODC-5024

**Root Analysis:**
Need a way to enable/disable helm specific navigation items from the console if there are no helm repositories configured 

**Solution Description:**
- disables the FLAG_OPENSHIFT_HELM flag when there are no HelmChartRepository configured &
- hides the Helm nav item
- hides the Helm catalog item from the Add page
- the helm URLs shows 404

**GIF/Sceenshot:**
helm_ff

### DIFF
--- a/frontend/packages/dev-console/src/components/catalog/catalog-plugin.ts
+++ b/frontend/packages/dev-console/src/components/catalog/catalog-plugin.ts
@@ -1,5 +1,6 @@
 import { CatalogItemProvider, CatalogItemType, Plugin } from '@console/plugin-sdk';
 import { builderImageProvider, helmChartProvider, templateProvider } from './providers';
+import { FLAG_OPENSHIFT_HELM } from '../../const';
 
 export type CatalogConsumedExtensions = CatalogItemProvider | CatalogItemType;
 
@@ -77,12 +78,18 @@ export const catalogPlugin: Plugin<CatalogConsumedExtensions> = [
         },
       ],
     },
+    flags: {
+      required: [FLAG_OPENSHIFT_HELM],
+    },
   },
   {
     type: 'Catalog/ItemProvider',
     properties: {
       type: 'HelmChart',
       provider: helmChartProvider,
+    },
+    flags: {
+      required: [FLAG_OPENSHIFT_HELM],
     },
   },
 ];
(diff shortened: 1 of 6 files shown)


rohitkrai03 (changes requested):
Even after helm flag is disabled catalog has a type filter for Helm Charts. Its probably better if we remove that filter item as well when Helm is disabled. @serenamarie125 thoughts on this?

Screenshot from 2020-11-04 00-59-59

frontend/packages/dev-console/src/components/topology/helm/helmTopologyPlugin.tsx
```diff
@@ -32,6 +33,9 @@ export const helmTopologyPlugin: Plugin<HelmTopologyConsumedExtensions> = [
       getDataModel: getHelmTopologyDataModel,
       isResourceDepicted: getIsHelmResource,
     },
+    flags: {
+      required: [FLAG_OPENSHIFT_HELM],
```
rohitkrai03, line 37: No need to remove topology visualization if helm is disabled.

rohitkrai03 (changes requested):

jeff-phillips-18:
We should also add the flag to the Helm Topology plugin: https://github.com/openshift/console/blob/6062e7e96baff020679764676655a8b54d46eb59/frontend/packages/topology/src/helm/helmTopologyPlugin.tsx

rohitkrai03 (changes requested):
@debsmita1 Tested the PR locally. Everything works as expected except the catalog. You need to feature flag the catalog extensions for Helm as well. Update the file `/catalog/catalog-plugin.ts` in components folder and add the Helm feature flag in helm extensions.

rohitkrai03:
@jeff-phillips-18 Based on the chats from this [thread](https://coreos.slack.com/archives/CRDPF7YS2/p1604943507073800?thread_ts=1604934392.072500&cid=CRDPF7YS2) it seems we would want to show the topology visualization for helm charts regardless of helm flag. Maybe @serenamarie125 can confirm it here as well.

rohitkrai03:
> We should also add the flag to the Helm Topology plugin: https://github.com/openshift/console/blob/6062e7e96baff020679764676655a8b54d46eb59/frontend/packages/topology/src/helm/helmTopologyPlugin.tsx

@jeff-phillips-18 Based on the chats from this [thread](https://coreos.slack.com/archives/CRDPF7YS2/p1604943507073800?thread_ts=1604934392.072500&cid=CRDPF7YS2) it seems we would want to show the topology visualization for helm charts regardless of helm flag. Maybe @serenamarie125 can confirm it here as well.

rohitkrai03 (approved):
/approve
/lgtm

PR #7038, opened 2020-10-29
Bug 1892198: Fix runtime error on operand details page.

TheRealJon:
Update logic in OperandList component to memoize post-processed data instead of calling `ensureKind` function on every render. Add optional chaining to prevent runtime error if `props.data` is undefined.


spadgett (approved):
/lgtm

PR #6971, opened 2020-10-21
Bug 1889515: - Accessibility - The checkmark in the Node > overview page has no text description, label, or other accessible information

cyril-ui-developer:
/assign @jcaianirh 
cc: @spadgett

PR #7041, opened 2020-10-29
Fix status icon size in topology and alignments

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5058

**Analysis / Root cause**: 
Changes made in https://github.com/openshift/console/pull/6971 caused the necessary classNames not to be added since the icon passed was changed to a function rather than a component.

**Solution Description**: 
Change the icon parameters back to a component and add an sr-only span for the title.

image

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/status/GenericStatus.tsx
+++ b/frontend/packages/console-shared/src/components/status/GenericStatus.tsx
@@ -4,23 +4,30 @@ import PopoverStatus from './PopoverStatus';
 import StatusIconAndText from './StatusIconAndText';
 
 const GenericStatus: React.FC<GenericStatusProps> = (props) => {
-  const { Icon, children, popoverTitle, title, ...restProps } = props;
-  return React.Children.toArray(children).length ? (
-    <PopoverStatus
-      title={popoverTitle || title}
+  const { Icon, children, popoverTitle, title, noTooltip, iconOnly, ...restProps } = props;
+  const renderIcon = iconOnly && !noTooltip ? <Icon title={title} /> : <Icon />;
+  const statusBody = (
+    <StatusIconAndText
       {...restProps}
-      statusBody={<StatusIconAndText {...restProps} title={title} icon={<Icon />} />}
-    >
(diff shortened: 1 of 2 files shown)


zherman0 (approved):

jerolimov:
@jeff-phillips-18 Maybe too many review questions today, sorry :smirk: 

`<GenericStatus` has also an optional title, in most cases status title is passed through. As an result this change duplicates the message for screen reader:

image

If I understand the origin PR #6971 correctly, we can maybe use your strings as fallback here if there is no title provided?

```tsx
export const ErrorStatus: React.FC<StatusComponentProps> = (props) => (
  <GenericStatus {...props} title={props.title || 'Error'} Icon={RedExclamationCircleIcon} />
);
ErrorStatus.displayName = 'ErrorStatus';
```

To add a screen reader label in this case we could maybe extend `<StatusIconAndText` with an `{iconOnly ? <span className="sr-only">{title}</span> : ...` case? Wdyt?

jeff-phillips-18:
@jerolimov Thought of that but it would cause a `Popover` around all the status icons which may not be desirable in all cases. However, we should add the sr-only span if there is no title passed. Thanks.

(and review comments/questions are always appreciated)

jeff-phillips-18:
@jerolimov Makes sense. Thanks!

jerolimov (approved):
Cool :+1:

/lgtm

jerolimov:
With the latest version we have now 2 tooltips if you hover the icon. Not perfect, but it still valid...
/lgtm
/cc christianvogt wdyt?

jerolimov:
With the latest version we have now 2 tooltips if you hover the icon. Not perfect, but it still valid...
/lgtm
/cc @christianvogt wdyt?

jeff-phillips-18:
@jerolimov Fixed the double tooltip. thanks!

rohitkrai03 (approved):
/approve

PR #7051, opened 2020-10-30
Use User settings hook for topology layout data, topology view state (graph/list) and cloudshell terminal

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5082

**Analysis / Root cause**: 
User preferences were saved in localStorage so not persisted across browsers 

Implemented this hook for 
- Topology layout Data
- Topology view state (graph/list)
- Cloudshell terminal

Depends on : https://github.com/openshift/console/pull/7301/files

**Solution Description**: 
Implemented useUserSettings for three features

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -28,7 +27,11 @@ type Props = {
 type CloudShellTerminalProps = StateProps & Props;
 
 const CloudShellTerminal: React.FC<CloudShellTerminalProps> = ({ user, onCancel }) => {
-  const [namespace, setNamespace] = React.useState(getCloudShellNamespace());
+  const [namespace, setNamespace] = useUserSettings(
+    CLOUD_SHELL_NAMESPACE_CONFIG,
+    '',
+    CLOUD_SHELL_NAMESPACE,
```
karthikjeeyar, line 33: Just a suggestion, If we rearrange the second and third positions of the arguments (which has default assignments) in this `useUserSettings` hook, then the caller function signature will be improved.

eg: useUserSetting(CLOUD_SHELL_NAMESPACE_CONFIG, CLOUD_SHELL_NAMESPACE) Third argument will always have empty string if the argument isn't passed. WDYT?

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
+import { ConfigMapModel, ProjectRequestModel, ProjectModel } from '@console/internal/models';
+import { K8sResourceKind, k8sGet, k8sCreate, k8sPatch } from '@console/internal/module/k8s';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+// FIXME upgrading redux types is causing many errors at this time
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useSelector } from 'react-redux';
+import { RootState } from '@console/internal/redux';
```
karthikjeeyar, line 9: nit: sort imports

frontend/packages/dev-console/src/components/topology/components/const.ts
```diff
@@ -7,6 +7,10 @@ export const TYPE_SERVICE_BINDING = 'service-binding';
 export const TYPE_APPLICATION_GROUP = 'part-of';
 export const TYPE_TRAFFIC_CONNECTOR = 'traffic-connector';
 export const LAST_TOPOLOGY_VIEW_LOCAL_STORAGE_KEY = `${STORAGE_PREFIX}/last-topology-view`;
+export const TOPOLOGY_LAYOUT_LOCAL_STORAGE_KEY = `${STORAGE_PREFIX}/topology-layout`;
+
+export const LAST_TOPOLOGY_VIEW_CONFIG_STORAGE_KEY = 'devconsole.topology.last-topology-view';
+export const TOPOLOGY_LAYOUT_CONFIG_STORAGE_KEY = 'devconsole.topology.topology-layout';
```
karthikjeeyar, line 13: It will be good to put `devconsole.topology` in the separate consts to avoid string repetition.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
invincibleJai: Updated to make use of just default value

```
const [topologyLayoutData, setTopologyLayoutData] = useUserSettings(
    TOPOLOGY_LAYOUT_CONFIG_STORAGE_KEY,
    localStorage.getItem(TOPOLOGY_LAYOUT_LOCAL_STORAGE_KEY) || '',
  );
```

frontend/packages/dev-console/src/components/topology/Topology.tsx
```diff
@@ -205,7 +213,8 @@ const Topology: React.FC<TopologyProps & StateProps & DispatchProps> = ({
             scaleExtent: storedGraphModel.scaleExtent,
           };
         }
-        const storedLayout = getTopologyLayout(namespace);
+        // const storedLayout = getTopologyLayout(namespace);
```
sahil143, line 216: remove comment

christianvogt:
/hold

We cannot merge this until the backend support is ready. I suggest splitting this PR into one just for the utility function and tests and another for the subsequent updates to the existing code to move from localStorage to user settings.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -46,9 +46,9 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps> = ({ user, onCancel
   React.useEffect(() => {
     if (loaded && !loadError) {
       // workspace may be undefined which is ok
-      setCloudShellNamespace(workspaceNamespace);
+      setNamespace(workspaceNamespace);
```
christianvogt, line 49: I would have to look into this further but looks like this is a change in behavior because `setCloudShellNamespace` only updated what was stored in localStorage. Now `setNamespace` also updates the state.

Unless absolutely sure, I suggest keeping the state namespace separate from the user settings.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,112 @@
...
+        data: {
+          [key]: defaultValue,
+        },
+      },
+    ],
+    [userUid, key, defaultValue],
+  );
+
+  const [cfData, cfLoaded, cfLoadError] = useK8sWatchResource<K8sResourceKind>(configMapResource);
+  const [settings, setSettings] = React.useState('');
```
christianvogt, line 43: Should this be undefined?
It cannot be the default value because we don't want a flash of the default value while we're still loading the settings.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,112 @@
...
+        (async () => {
+          await patchConfigMap(patch);
+        })();
+      }
+    }
+    // This effect should only be run on change of settings state.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [settings]);
+
+  return [settings, setSettings] as const;
```
christianvogt, line 111: Since settings can be lazy loaded, I think we need another return value which indicates whether the settings have been fetched yet or not.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -28,7 +25,10 @@ type Props = {
 type CloudShellTerminalProps = StateProps & Props;
 
 const CloudShellTerminal: React.FC<CloudShellTerminalProps> = ({ user, onCancel }) => {
-  const [namespace, setNamespace] = React.useState(getCloudShellNamespace());
+  const [namespace, setNamespace] = useUserSettings(
+    CLOUD_SHELL_NAMESPACE_CONFIG_STORAGE_KEY,
+    localStorage.getItem(CLOUD_SHELL_NAMESPACE) || '',
```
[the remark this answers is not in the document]
invincibleJai: That's what i thought we can pass `localstorage_key` as an argument to `useUserSetting` and on successful patch remove local storage that way this we be handled just in hook and as suggested can be removed later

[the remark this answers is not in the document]
jerolimov: I think also we should avoid any usage of localStorage in the components and also find a way to drop the duplicated storage key consts here

frontend/packages/console-app/src/components/cloud-shell/const.ts
```diff
@@ -0,0 +1,6 @@
+import { STORAGE_PREFIX, STORAGE_PERSPECTIVE } from '@console/shared';
+
+const STORAGE_CLOUDSHELL = 'cloudshell';
```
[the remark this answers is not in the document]
jerolimov: Esp. when we want change keys I recommend that we need some migration here

frontend/packages/console-shared/src/hooks/useUserSettings.ts
jerolimov: Valid point, esp when we mean defaultValue is returned != data fetched

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+    // This effect should only be run once per key.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [key, cfLoaded]);
+
+  React.useEffect(() => {
+    if (cfData && cfLoaded && settings) {
+      const patch = [
+        {
+          op: cfData.data?.[key] ? 'replace' : 'add',
```
jerolimov, line 92:
```suggestion
          op: 'replace',
```

I though a little bit about parallel create user-settings and parallel patch calls. The `replace` operation works here for adding or updating while adding works just when the data key is not available.

frontend/packages/console-app/src/components/cloud-shell/const.ts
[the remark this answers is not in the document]
invincibleJai: have made use of `terminal` and Sorry any suggestion on how we can do migration here 

i.e here LocalStorage used `bridge/command-line-terminal-namespace`
and based on proposal for configmap key naming convention is `console.terminal.command-line-terminal-namespace`?

and similar will be case with other storage keys

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
invincibleJai: i tried verifying it and we had `setCloudShellNamespace` to set in localStorage and `getCloudShellNamespace` to get from LocalStorage so with `setNamespace` it'll update the ns in ConfigMap, don't we need that.

@sahil143 can you as well verify once?

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -27,28 +30,25 @@ export interface TopologyPageProps {
...
+    isTopologyViewStateLoaded,
+  ] = useUserSettingsCompatibility(TOPOLOGY_VIEW_CONFIG_STORAGE_KEY, 'graph', activeViewStorageKey);
   const namespace = match.params.name;
   const queryParams = useQueryParams();
   let viewType = queryParams.get('view') as TopologyViewType;
   const { projects } = useK8sWatchResources<{ [key: string]: K8sResourceKind[] }>({
     projects: { kind: 'Project', isList: true },
   });
-  if (!viewType) {
+  if (!viewType && isTopologyViewStateLoaded) {
```
jeff-phillips-18, line 51: We could still use the URL path if `isTopologyViewStateLoaded` is false.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,108 @@
...
+const createConfigMap = async (configMapData: K8sResourceKind) => {
+  try {
+    await k8sCreate(ConfigMapModel, configMapData);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+};
+
+const patchConfigMap = async (patch, cfData) => {
```
jerolimov, line 58:
```suggestion
const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: string) => {
 // create the patch here inline
```

This would simplify the useEffect which triggers an update

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { useUserSettings } from './useUserSettings';
+
+export const useUserSettingsCompatibility = (
+  key: string,
+  defaultValue: string = '',
+  storageKey?: string,
```
jerolimov, line 7: Also when this is @christianvogt recommended order, we should use `useUserSettingsCompatibility` only with a storageKey. So I still recommend `key: string, storageKey: string, defaultValue?: T`

If we replace this later with `useUserSettings` it doesn't matter if you need to drop the 2nd or 3rd argument.

jerolimov: I also would recommend to return JSON parsed data if possible?!

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,108 @@
...
+  try {
+    await k8sPatch(ConfigMapModel, cfData, patch);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+};
+
+export const useUserSettings = (key: string, defaultValue: string = '') => {
```
jerolimov, line 67: I would like if we have a generic type here and return JSON parsed data here.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+    (state: RootState) => state.UI.get('user')?.metadata?.uid ?? 'kubeadmin',
+  );
+  const [configMapResource, configMapData] = getConfigResourceData(
+    userUid,
+    key,
+    seralizeData(defaultValue ?? ''),
+  );
+  const [cfData, cfLoaded, cfLoadError] = useK8sWatchResource<K8sResourceKind>(configMapResource);
+  const [settings, setSettings] = React.useState();
+  const [loading, setLoading] = React.useState(false);
```
jeff-phillips-18, line 110: should this var be `loaded` rather than `loading`?

jeff-phillips-18 (approved):

frontend/packages/console-app/src/components/cloud-shell/__tests__/CloudShellTerminal.spec.tsx
```diff
@@ -19,6 +19,14 @@ jest.mock('react-i18next', () => {
   };
 });
 
+jest.mock('@console/shared', () => {
+  const ActualShared = require.requireActual('@console/shared');
+  return {
+    ...ActualShared,
+    useUserSettingsCompatibility: () => ['', () => {}],
+  };
+});
```
sahil143, line 28: I don't think we need requireActual here.
```suggestion
jest.mock('@console/shared/src/hooks/useUserSettingsCompatibility', () => {
  return {
    useUserSettingsCompatibility: () => ['', () => {}],
  };
});
```

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
sahil143: I tested it locally cloud shell seems to work as expected.

sahil143 (approved):
/lgtm

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -28,11 +25,17 @@ type Props = {
 type CloudShellTerminalProps = StateProps & Props;
 
 const CloudShellTerminal: React.FC<CloudShellTerminalProps> = ({ user, onCancel }) => {
-  const [namespace, setNamespace] = React.useState(getCloudShellNamespace());
+  const [namespace, setNamespace, isNamespaceLoaded] = useUserSettingsCompatibility(
+    CLOUD_SHELL_NAMESPACE_CONFIG_STORAGE_KEY,
+    CLOUD_SHELL_NAMESPACE,
+  );
```
christianvogt, line 31: The way the default namespace works, we must have it loaded before we initialize the `useCloudShellWorkspace` hook. Which means this must go in a separate component like an HOC.

We can either create an HOC now for others to use or make a one off case and create a parent component which doesn't render the child component until loaded.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -28,11 +25,17 @@ type Props = {
...
   const [initData, setInitData] = React.useState<TerminalInitData>();
   const [initError, setInitError] = React.useState<string>();
 
-  const [workspace, loaded, loadError] = useCloudShellWorkspace(user, namespace);
+  const [workspace, loaded, loadError] = useCloudShellWorkspace(
+    user,
+    isNamespaceLoaded ? namespace : '',
+  );
```
christianvogt, line 38: This needs to remain as is. See other comment about loading.

frontend/packages/console-shared/src/constants/common.ts
```diff
@@ -78,3 +78,5 @@ export enum FLAGS {
   CONSOLE_EXTERNAL_LOG_LINK = 'CONSOLE_EXTERNAL_LOG_LINK',
   CONSOLE_YAML_SAMPLE = 'CONSOLE_YAML_SAMPLE',
 }
+
+export const CONFIG_STORAGE_PERSPECTIVE_CONSOLE = 'console';
```
christianvogt, line 82: Why is this named `CONFIG_STORAGE_PERSPECTIVE_CONSOLE` with `PERSPECTIVE` in it?

frontend/packages/dev-console/src/components/topology/Topology.tsx
```diff
@@ -133,6 +120,16 @@ const Topology: React.FC<TopologyProps & StateProps & DispatchProps> = ({
...
+  const [
+    topologyLayoutData,
+    setTopologyLayoutData,
+    isTopologyLayoutDataLoaded,
+  ] = useUserSettingsCompatibility(
+    TOPOLOGY_LAYOUT_CONFIG_STORAGE_KEY,
+    TOPOLOGY_LAYOUT_LOCAL_STORAGE_KEY,
+    {},
+  );
+  const topologyLayoutDataJson = isTopologyLayoutDataLoaded ? topologyLayoutData : {};
```
christianvogt, line 132: Same comment about loaded here. I think we _need_ the config loaded before we render this component since we require the data on first render.

frontend/packages/dev-console/src/components/topology/TopologyPage.tsx
```diff
@@ -60,16 +64,16 @@ export const TopologyPage: React.FC<TopologyPageProps> = ({
           exact: true,
         })
       ? TopologyViewType.graph
-      : getTopologyActiveView(activeViewStorageKey) || defaultViewType;
-    setQueryArgument('view', viewType);
+      : isTopologyViewStateLoaded && ((topologyViewState as TopologyViewType) || defaultViewType);
```
christianvogt, line 67: Same comment about loaded.

invincibleJai: have added HOC for other two implementations and for this i think we are supporting backward compatibility so we  check path and proceed if exists otherwise use any stored preference. cc @jeff-phillips-18 

https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/topology/TopologyPage.tsx#L52-L63

invincibleJai:
Have added another PR just with Hooks https://github.com/openshift/console/pull/7242

sahil143 (commented):
tested the fallback local storage. Looks good to me.

jerolimov (approved):
Tested the persistence of data and this works fine for me.

- Topology layout Data
- Topology view state (graph/list)
- Cloudshell terminal

/lgtm

sahil143 (approved):
/lgtm

jerolimov:
Tested it again together with api change, works fine.

@invincibleJai needs rebase after #7199

jerolimov:
Still works fine :smile:

/lgtm

jerolimov:
/retitle Use User settings hook for topology layout data, topology view state (graph/list) and cloudshell terminal

jerolimov:
/title Use User settings hook for topology layout data, topology view state (graph/list) and cloudshell terminal

christianvogt:
Looks like `replace` doesn't work if the entry isn't present.
I'm seeing the following error trying to patch the configmap:

request url:
```
PATCH
/api/kubernetes/api/v1/namespaces/console-user-settings/configmaps/user-settings-d15fe175-17a2-463a-bfbc-d6f7da23e31c
```

request payload:
```
[{op: "replace", path: "/data/console.terminal.namespace", value: "cd-dev2"}]
```

422 error response:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "the server rejected our request due to an error in our request",
  "reason": "Invalid",
  "details": {
    
  },
  "code": 422
}
```

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
christianvogt: Verified as well that setting this, while it does trigger an effect, will not cause any issues.

jerolimov:
> 422 error response:

@christianvogt @invincibleJai  The reason is that the ConfigMap contains no data. I fixed this in my PR #7327 by using a different patch format (merge-patch+json instead of patch+json). See:

https://github.com/openshift/console/blob/b06c1a73b9c46c84cbb57b8b931b4bdcb57cab43/frontend/packages/console-shared/src/utils/user-settings.ts#L18-L49

We can merge that as part of my PR or Jai can also integrate this into this PR.

invincibleJai:
> > 422 error response:
> 
> @christianvogt @invincibleJai The reason is that the ConfigMap contains no data. I fixed this in my PR #7327 by using a different patch format (merge-patch+json instead of patch+json). See:
> 
> https://github.com/openshift/console/blob/b06c1a73b9c46c84cbb57b8b931b4bdcb57cab43/frontend/packages/console-shared/src/utils/user-settings.ts#L18-L49
> 
> We can merge that as part of my PR or Jai can also integrate this into this PR.
> 
> As workaround you need to add some fake `data: aaa: bbb` to your ConfigMap.

Thanks @jerolimov , as in current flow am updating Configmap with data in creation flow itself so i don't think we need workaround now and as part of your PR (actual proxy endpoint) we have creation just to handle create and then updated which handles this scenario. So let me know if need to handle anything here ?

cc @christianvogt @jerolimov

invincibleJai:
> > 422 error response:
> 
> @christianvogt @invincibleJai The reason is that the ConfigMap contains no data. I fixed this in my PR #7327 by using a different patch format (merge-patch+json instead of patch+json). See:
> 
> https://github.com/openshift/console/blob/b06c1a73b9c46c84cbb57b8b931b4bdcb57cab43/frontend/packages/console-shared/src/utils/user-settings.ts#L18-L49
> 
> We can merge that as part of my PR or Jai can also integrate this into this PR.
> 
> As workaround you need to add some fake `data: aaa: bbb` to your ConfigMap.

Thanks @jerolimov @christianvogt the problem is as you described if data is not there in Configmap and tries operator `replace` with patch request. what we can do is in `op` make a check if key doesn't exists instead of always doing `replace` perform `add` and we should good.

Let me know if you see issue.

PR #7061, opened 2020-11-01
Bug 1843045: Adds a hook Data loader that pre loads data related to Subscriptions, InstallPlan and CatalogSource.

bipuladh:


frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx
```diff
@@ -341,7 +341,9 @@ export const SubscriptionDetails: React.FC<SubscriptionDetailsProps> = ({
   const installFailedMessage =
     installFailedCondition?.message || installFailedCondition?.reason || 'Install plan failed';
 
-  const pkg = packageForSubscription(packageManifests, obj);
+  const pkg = _.isArray(packageManifests)
+    ? packageForSubscription(packageManifests, obj)
+    : packageManifests;
```
bipuladh, line 346: We can remove this block if we add a similar wrapper for `SubscriptionDetailsPage` at https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx#L594

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1038,28 +1038,54 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+/**
+ * Give the name of the operator from subscription to the CSV Details Page
+ */
+export const CSVDetailsPageWrapper: React.FC<CSVDetailsPageWrapper> = (props) => {
+  const { match } = props;
+  const subscriptionResource = React.useMemo(
+    () => ({
+      isList: true,
+      kind: referenceForModel(SubscriptionModel),
+      namespace: match.params.ns,
```
spadgett, line 1064: Is this right for global operators? If it's a copied CSV, I'd expect the Subscription to be in a different namespace.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1038,28 +1038,54 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
     </StatusBox>
   );
 };
 
+/**
+ * Give the name of the operator from subscription to the CSV Details Page
+ */
+export const CSVDetailsPageWrapper: React.FC<CSVDetailsPageWrapper> = (props) => {
+  const { match } = props;
+  const subscriptionResource = React.useMemo(
```
spadgett, line 1060: I wouldn't expect us to need a new watch. We already have a Subscription tab, so I'd expect us to already have the data somewhere.

frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx
spadgett: I don't think it's a good idea to overload the prop to have two different types. It would be better to do it one way in all contexts.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
bipuladh: The whole idea of this PR is to get the subscription before getting package manifests so that we can narrow down the scope of our request for it. I have removed the block for subsequent fetching of Subscription and supplying this instance.
 https://github.com/openshift/console/pull/7061/files/a52ebe7ec2f3dd0c51bba749c16b8d100241665d#diff-c7de8d95faa26753bd35817bb04d94993b4f37ecd076de5edd8690488d8c8ecfL1141.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
bipuladh: In that case, we first need to fetch the CSV then the subscription, and based on `olm.operatorNamespace` annotation adjust the above namespace value.

TheRealJon (commented):
A couple of things:

1.  I think the logic for retrieving the subscription should be inside the ClusterServiceVersionDetailsPage component instead of adding an extra wrapper component.
2. If we are going to update the way SubscriptionDetails gets PackageManifest, we might as well update the logic for CatalogSource, ClusterServiceVersion, and InstallPlan as well. Right now, in the ClusterServiceVersionDetailsPage and the SubscriptionDetailsPage components, we watch a list of all of these resources, and ultimately just pass them down to the SubscriptionDetails component, where we do some business logic to find the correct associated resource from each list. IMO, it would make more sense to use the `useK8sWatchResources` hook in the SubscriptionDetails component and retrieve only the individual resources we need, at the time we actually need them. Even though this requires two or three subsequent API calls, I think it's better than watching 4 lists of resources and manually searching each one.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1034,35 +1038,177 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+    installPlan: InstallPlanKind;
+    catalogSource: CatalogSourceKind;
+    subscription: SubscriptionKind;
+    csv: ClusterServiceVersionKind;
+  },
+  boolean,
+  any,
+];
+
+const useCSVPageDataLoader = (dataLoaderProps: DataLoaderProps): DataLoaderResults => {
```
TheRealJon, line 1083: If we move the k8sWatch hooks for the Subscription-related resources to the SubscriptionDetails component, I don't think this custom hook will be necessary.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1034,35 +1038,177 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+
+  const memoizedCSVData = useDeepCompareMemoize(csvData, true);
+
+  const operatorNamespace = memoizedCSVData?.metadata?.annotations?.['olm.operatorNamesapce'];
+
+  const subscriptionResource = {
+    isList: true,
+    kind: referenceForModel(SubscriptionModel),
+    namespace: operatorNamespace,
+  };
```
TheRealJon, line 1104: We can use fieldSelector here to get only the subscription we need:

```suggestion
  const subscriptionResource = {
    isList: false,
    kind: referenceForModel(SubscriptionModel),
    namespace: operatorNamespace,
    fieldSelector: `status.installedCSV=${name}`
  };
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1034,35 +1038,177 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+
+  const currSubscription = React.useMemo(
+    () =>
+      _.find(memoizedSubscriptionData, {
+        status: {
+          installedCSV: name,
+        },
+      }),
+    [memoizedSubscriptionData, name],
+  );
```
TheRealJon, line 1120: Adjust the k8s watch hook to only retrieve the subscription we want and we can remove this logic.

```suggestion
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1034,35 +1038,177 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+      loadError: catalogSourceLoadError,
+    },
+  } = data;
+
+  const loaded =
+    pkgManifestLoaded ||
+    installPlanLoaded ||
+    catalogSourceLoaded ||
+    csvLoaded ||
+    subscriptionLoaded;
```
TheRealJon, line 1176: This will be obsolete with the other changes I'm requesting, but this expression should use `&&` instead of `||`

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1034,35 +1038,177 @@ export const ClusterServiceVersionDetails: React.SFC<ClusterServiceVersionDetail
...
+  const loadError =
+    pkgManifestLoadError ||
+    installPlanLoadError ||
+    catalogSourceLoadError ||
+    csvLoadError ||
+    subscriptionLoadError;
```
TheRealJon, line 1182: Again, obsolete, but this would only show one loadError even if more than one failed.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -1096,19 +1242,11 @@ export const ClusterServiceVersionsDetailsPage: React.FC<ClusterServiceVersionsD
...
-      ...(_.isEmpty(subscription)
-        ? [Kebab.factory.Delete(model, obj)]
-        : [editSubscription(subscription), uninstall(subscription, obj)]),
-    ];
-  };
+  const menuActions = (model, obj: ClusterServiceVersionKind) => [
+    ...(_.isEmpty(subscription)
+      ? [Kebab.factory.Delete(model, obj)]
+      : [editSubscription(subscription), uninstall(subscription, obj)]),
+  ];
```
TheRealJon, line 1249:
```suggestion
  const menuActions = React.useCallback((model, obj: ClusterServiceVersionKind, _unused, { subscription }) => [
    ...(_.isEmpty(subscription)
      ? [Kebab.factory.Delete(model, obj)]
      : [editSubscription(subscription), uninstall(subscription, obj)]),
  ], []);
```

TheRealJon (commented):
Since this is a performance bug fix, I'd like to address the performance issue all the way through. These changes address performance issues on the CSV details page, but the SubscriptionDetailsPage probably has similar issues. See my comments about moving some of these k8s watch hooks into SubscriptionDetails.

spadgett:

PR #7070, opened 2020-11-02
Fetch pods when displaying pods resources in overview pages

jeff-phillips-18:
**Fixes**
https://issues.redhat.com/browse/ODC-4966 (portions of it)

**Description**
Update the pod resources section of overview pages to retrieve pods for display rather than depending upon the passed in item's resources to be kept up to date.

Non-functional change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup

### DIFF
--- a/frontend/public/components/overview/pods-overview.tsx
+++ b/frontend/public/components/overview/pods-overview.tsx
@@ -3,9 +3,15 @@ import * as React from 'react';
 import { Link } from 'react-router-dom';
 
 import { Alert, AlertActionLink } from '@patternfly/react-core';
-import { Status } from '@console/shared';
-import { ResourceLink, resourcePath, SidebarSectionHeading } from '../utils';
+import {
+  Status,
+  usePodsWatcher,
+  getPodsForResource,
+  getResourcesToWatchForPods,
+} from '@console/shared';
+import { ResourceLink, resourcePath, SidebarSectionHeading, LoadingBox } from '../utils';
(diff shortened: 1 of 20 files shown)


frontend/public/components/overview/pods-overview.tsx
```diff
@@ -173,22 +181,122 @@ export const PodsOverview: React.SFC<PodsOverviewProps> = ({
         </Alert>
       ) : null}
       {_.isEmpty(filteredPods) ? (
-        <span className="text-muted">{emptyMessage}</span>
+        <span className="text-muted">{loaded ? loadError || emptyMessage : <LoadingBox />}</span>
```
andrewballantyne, line 184: Hmm... looks like if we get a gateway error because the cluster going down, this doesn't really render properly; it's stuck not loaded but with an error. Did we want to check loaded and NOT an error?

React Dev Tools & the Response (Error) Object:
Screen Shot 2020-11-03 at 7 10 27 PM
Screen Shot 2020-11-03 at 7 10 34 PM

andrewballantyne (changes requested):

jeff-phillips-18:
@andrewballantyne Added tests and updated to display loadError if there is one.

andrewballantyne (approved):
/lgtm

andrewballantyne:
/retest

Looks like connection issues

PR #7071, opened 2020-11-02
Bug 1893351: Determine Secret type upon the save

jhadvig:
We have been passing "Opaque" type when updating secret via Generic Secret form, even though  `kubernets/tls` and `kubernets/service-account-token` Secret types are being updated with this form type.
Since with Generic form any Secret type can be created, determine the type when the Secret is being created.

/assign @spadgett

### DIFF
--- a/frontend/public/components/secrets/create-secret.tsx
+++ b/frontend/public/components/secrets/create-secret.tsx
@@ -101,6 +101,24 @@ const generateSecret = () => {
   return s4() + s4() + s4() + s4();
 };
 
+const determineSecretType = (stringData): SecretType => {
+  const dataKeys = _.keys(stringData).sort();
+  if (_.isEqual(dataKeys, ['tls.crt', 'tls.key'])) {
+    return SecretType.tls;
+  } else if (_.isEqual(dataKeys, ['ca.crt', 'namespace', 'service-ca.crt', 'token'])) {
+    return SecretType.serviceAccountToken;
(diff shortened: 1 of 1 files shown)


frontend/public/components/secrets/create-secret.tsx
```diff
@@ -101,6 +101,24 @@ const generateSecret = () => {
   return s4() + s4() + s4() + s4();
 };
 
+const determineSecretType = (stringData): SecretType => {
```
spadgett, line 104: `stringData` makes it sounds like the parameter is a string, but it appears to be an object.

spadgett (changes requested):
We shouldn't try to infer the type on edit. We should simply preserve whatever was previously in the secret. The field is immutable, so it shouldn't change. It also works for any secret type, including unknown types.

frontend/public/components/secrets/create-secret.tsx
jhadvig: @spadgett so the `stringData` is name of the [field](https://kubernetes.io/docs/concepts/configuration/secret/#overview-of-secrets). Similar to `data` but this take an arbitrary string value.

jhadvig:
> We shouldn't try to infer the type on edit. We should simply preserve whatever was previously in the secret. The field is immutable, so it shouldn't change. It also works for any secret type, including unknown types.

So this fix is will take the `type` of the edited Secret. Wont edit anything.

spadgett (approved):
/lgtm

PR #7075, opened 2020-11-02
E2E test: Use KUBEADM_PASSWORD_FILE by default

alvaroaleman:
We want to move the current template-based tests to the more modern
step-registry based tests, which provide the kubeadm password file
location via an env var.


spadgett (approved):
/lgtm

PR #7080, opened 2020-11-03
not all disk types can be selected as the boot source

glekner:


glekner:
image

@matthewcarleton @jelkosz can you approve the validation message

glekner:
@suomiy can you review

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/storage-tab/storage-boot-source.tsx
```diff
@@ -56,11 +65,9 @@ export const StorageBootSource: React.FC<StorageBootOrderProps> = ({
           isDisabled={isDisabled}
         >
           <FormSelectPlaceholderOption isDisabled placeholder={SELECT_BOOTABLE_DISK} />
-          {ignoreCaseSort(filteredStorages, null, (storage) => storage.disk?.name).map(
-            (storage) => (
-              <FormSelectOption key={storage.id} value={storage.id} label={storage.disk?.name} />
-            ),
-          )}
+          {ignoreCaseSort(storages, null, (storage) => storage.disk?.name).map((storage) => (
```
suomiy, line 68: why are we including the disks they cannot select?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/storage.ts
```diff
@@ -1,3 +1,5 @@
 export const ATTACH_CD = 'Attach CD-ROM';
-export const NO_BOOTABLE_ATTACHED_DISK_ERROR = 'A bootable attached disk could not be found';
+export const BOOTABLE_ATTACHED_DISK_MESSAGE = 'Boot disks can not have a blank source';
```
suomiy, line 2: blank source has a special meaning. Shouldn't we say something like that `selecting a boot source is required`?

suomiy (changes requested):

glekner:
added `CONTAINER_DISK` as a viable boot source @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/storage-tab/storage-boot-source.tsx
suomiy: ok let's show all disks

suomiy:
+1, just this left https://github.com/openshift/console/pull/7080#discussion_r516660247

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/storage.ts
glekner: @matthewcarleton can you jump in?

matthewcarleton: What we want to communicate is that the disk they are selecting must have a source defined and can not be blank. 
Could we say:
"A boot source disk must have a source and can not be blank"

glekner:
@suomiy updated string

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/strings/storage.ts
suomiy: not sure if we want to mention other invalid sources (which is not just blank type), but ok let's put this message

PR #7082, opened 2020-11-03
Bug 1893963: Dont use lookbehinds regexp

yaacov:
On Firefox 68.12.0esr (64-bit) and  66.0.2 (64-bit) when going to the Virtualization tab in the UI nothing is loading on the right side 
FireFox 68 does not support lookbehinds regular expresions [1], this PR introduce a workaround supported by older browsers.

[1] https://kangax.github.io/compat-table/es2016plus/#test-RegExp_Lookbehind_Assertions

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vms/vm-console.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vms/vm-console.tsx
@@ -82,9 +82,8 @@ const VMConsoles: React.FC<VMConsolesProps> = ({
   const cloudInitVolume = getCloudInitVolume(vm);
   const data = new VolumeWrapper(cloudInitVolume).getCloudInitNoCloud();
   const cloudInitHelper = new CloudInitDataHelper(data);
-  const cloudInitUserData = cloudInitHelper.getUserData();
-  const cloudInitUsername = cloudInitUserData?.match(/(?<=user:\s+).*/);
-  const cloudInitPassword = cloudInitUserData?.match(/(?<=password:\s+).*/);
+  const cloudInitUsername = cloudInitHelper.get('user');
+  const cloudInitPassword = cloudInitHelper.get('password');
 
   if (!isVMIRunning(vmi)) {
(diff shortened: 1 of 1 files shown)


yaacov:
@glekner please review

suomiy:
/lgtm cancel

there is no need for regexes at all here, you can do just:

```tsx
cloudInitHelper.get(CloudInitDataFormKeys.PASSWORD)
```

yaacov:
@suomiy 

> there is no need for regexes at all here, you can do just:
> `cloudInitHelper.get(CloudInitDataFormKeys.PASSWORD)`

does the NAME works for you (we need name and password) ?
`cloudInitHelper.get(CloudInitDataFormKeys.NAME)`

suomiy:
I suppose you meant a key `user` which can be easily added to `CloudInitDataFormKeys`

yaacov:
> I suppose you meant a key user which can be easily added to CloudInitDataFormKeys

@suomiy this is a bug we need to fix and backport, lets merge it as is, and then make a bigger PR to add the missing field and re-factor the regexps only for 4.7, WDYT ?

suomiy:
there is no refactoring needed, just adding one key is enough. Plus the regex approach is not working with base64 cloud init

yaacov:
> there is no refactoring needed, 

The helper is broken  we need to fix it to support user, and then we will need to use the new feature we added to this helper class here. it's adding a fix to the the helper class bug in this bug fix. I would rather fix the helper class in a different PR, WDYT ?

suomiy:
can you tell me how is the helper class broken?

yaacov:
> can you tell me how is the helper class broken?

@suomiy it does not support `user`, we need both `user` and `password` for this fix

suomiy:
if it is, then it should be different PR indeed

suomiy:
> @suomiy it does not support user, we need both user and password for this fix

that is not broken - that just means it is not possible to define all constants in existence which might appear in cloud init

you can use a plain string `"user"` if you do not want to touch the helper class

suomiy:
although, having constants is more useful IMO

PR #7084, opened 2020-11-03
add support for Operator Backed SBR through topology inContext

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5051

**Analysis / Root cause**: SBR for Operator Backed through incontext was missing

**Solution Description**: added support for Operator Backed SBR through topology inContext

OB-incontext

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/plugin.tsx
+++ b/frontend/packages/dev-console/src/plugin.tsx
@@ -46,7 +46,7 @@ import {
   FLAG_OPENSHIFT_PIPELINE,
   ALLOW_SERVICE_BINDING,
   FLAG_OPENSHIFT_GITOPS,
-  CONNECTOR_INCONTEXT_ACTIONS,
+  INCONTEXT_ACTIONS_CONNECTS_TO,
 } from './const';
 import {
   newPipelineTemplate,
@@ -1178,7 +1178,7 @@ const plugin: Plugin<ConsumedExtensions> = [
   {
     type: 'PostFormSubmissionAction',
     properties: {
-      type: CONNECTOR_INCONTEXT_ACTIONS.connectsTo,
+      type: INCONTEXT_ACTIONS_CONNECTS_TO,
       callback: doConnectsToBinding,
     },
   },
--- a/frontend/packages/dev-console/src/utils/useAddToProjectAccess.ts
+++ b/frontend/packages/dev-console/src/utils/useAddToProjectAccess.ts
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useFlag } from '@console/shared';
 import { AccessReviewResourceAttributes, K8sKind } from '@console/internal/module/k8s';
 import { useAccessReview } from '@console/internal/components/utils';
 import {
@@ -10,7 +11,12 @@ import {
   SecretModel,
   ServiceModel,
 } from '@console/internal/models';
-import { allCatalogImageResourceAccess, allImportResourceAccess } from '../actions/add-resources';
+import {
+  allCatalogImageResourceAccess,
+  allImportResourceAccess,
+  serviceBindingAvailable,
+} from '../actions/add-resources';
+import { ALLOW_SERVICE_BINDING } from '../const';
 
 const resourceAttributes = (model: K8sKind, namespace: string): AccessReviewResourceAttributes => {
   return {
@@ -22,8 +28,6 @@ const resourceAttributes = (model: K8sKind, namespace: string): AccessReviewReso
 };
 
 export const useAddToProjectAccess = (activeNamespace: string): string[] => {
-  const [addAccess, setAddAccess] = React.useState<string[]>([]);
-
   const buildConfigsAccess = useAccessReview(resourceAttributes(BuildConfigModel, activeNamespace));
   const imageStreamAccess = useAccessReview(resourceAttributes(ImageStreamModel, activeNamespace));
   const deploymentConfigAccess = useAccessReview(
@@ -36,7 +40,9 @@ export const useAddToProjectAccess = (activeNamespace: string): string[] => {
   const routeAccess = useAccessReview(resourceAttributes(RouteModel, activeNamespace));
   const serviceAccess = useAccessReview(resourceAttributes(ServiceModel, activeNamespace));
 
-  React.useEffect(() => {
+  const serviceBindingEnabled = useFlag(ALLOW_SERVICE_BINDING);
+
+  return React.useMemo(() => {
     const createResourceAccess: string[] = [];
     if (
       buildConfigsAccess &&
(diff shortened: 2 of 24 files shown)


frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -1166,6 +1166,13 @@ const plugin: Plugin<ConsumedExtensions> = [
       callback: doConnectsToBinding,
     },
   },
+  {
+    type: 'PostFormSubmissionAction',
+    properties: {
+      type: CONNECTOR_INCONTEXT_ACTIONS.serviceBinding,
+      callback: doContextualBinding,
+    },
+  },
```
jeff-phillips-18, line 1175: I'd rather this be in `operatorsTopologyPlugin` to keep it with other service binding extension support (eventually moved out of dev-console and into another mono repo).
Should this be dependent on the `ALLOW_SERVICE_BINDING` flag?

frontend/public/components/catalog/catalog-page.tsx
```diff
@@ -47,328 +47,335 @@ import { Item } from './types';
 export const CatalogListPage = withExtensions<CatalogListPageExtensionProps>({
   devCatalogExtensions: isDevCatalogModel,
 })(
-  class CatalogListPage extends React.Component<CatalogListPageProps, CatalogListPageState> {
-    constructor(props: CatalogListPageProps) {
-      super(props);
+  withQueryParams<CatalogListPageProps>(
+    class CatalogListPage extends React.Component<CatalogListPageProps, CatalogListPageState> {
+      constructor(props: CatalogListPageProps) {
+        super(props);
```
jeff-phillips-18, line 53: Would be better just to convert this to a Functional Component.

jeff-phillips-18 (commented):
At some point the options for resource menu should come from the plugin as well rather than hardcoding the check and adding the operator backed option.

frontend/public/components/catalog/catalog-page.tsx
sahil143: Catalog page is already getting a big refactor in other epic. So didn't wanted to make any big changes here.

frontend/packages/dev-console/src/plugin.tsx
sahil143: This is just a callback. the actual action is dispatched from the context menu through URL params. Not sure if we should add the flag here.

frontend/packages/dev-console/src/utils/useAddToProjectAccess.ts
```diff
@@ -36,6 +45,10 @@ export const useAddToProjectAccess = (activeNamespace: string): string[] => {
   const routeAccess = useAccessReview(resourceAttributes(RouteModel, activeNamespace));
   const serviceAccess = useAccessReview(resourceAttributes(ServiceModel, activeNamespace));
 
+  const serviceBindingEnabled = useSelector((state: RootState) =>
+    state.FLAGS.get(ALLOW_SERVICE_BINDING),
+  );
```
christianvogt, line 50: I'm surprised we don't have a `useFeatureFlag` hook. We should create one inside `features.ts`

frontend/packages/dev-console/src/utils/useAddToProjectAccess.ts
```diff
@@ -36,6 +45,10 @@ export const useAddToProjectAccess = (activeNamespace: string): string[] => {
...
+
   React.useEffect(() => {
```
christianvogt, line 52: This actually shouldn't be an effect. There's no reason for this hook to use `useState`.
`createResourceAccess` can easily be derived inside the hook function and at most we can replace `useEffect` with `useMemo` which returns `createResourceAccess` instead of setting state.

frontend/packages/operator-lifecycle-manager/src/components/operand/DEPRECATED_operand-form.tsx
```diff
@@ -716,7 +718,10 @@ export const DEPRECATED_CreateOperandForm: React.FC<OperandFormProps> = ({
           ? immutableFormData.setIn(['metadata', 'namespace'], match.params.ns).toJS()
           : immutableFormData.toJS(),
       )
-        .then(() => history.push(next))
+        .then((res) => {
+          postFormCallback(res);
```
christianvogt, line 722: I realize now that we probably want the `postFormCallback` to be async so that we can ensure it's complete before executing `history.push`

frontend/packages/operator-lifecycle-manager/src/plugin.tsx
```diff
@@ -184,19 +184,6 @@ const plugin: Plugin<ConsumedExtensions> = [
...
-      exact: true,
-      path: `/k8s/ns/:ns/${models.ClusterServiceVersionModel.plural}/:appName/:plural/~new`,
-      loader: async () =>
-        (
-          await import(
-            './components/operand/create-operand' /* webpackChunkName: "create-operand" */
-          )
-        ).CreateOperandPage,
-    },
-  },
```
christianvogt, line 199: I don't think you can remove this route as it seems there are other references in the code.

sahil143: This is a duplicate route.

christianvogt: Ah i see where the other is now :)

frontend/packages/operator-lifecycle-manager/src/components/operand/DEPRECATED_operand-form.tsx
```diff
@@ -716,7 +718,10 @@ export const DEPRECATED_CreateOperandForm: React.FC<OperandFormProps> = ({
...
+          return history.push(next);
+        })
```
christianvogt, line 724: Any reason to not wait for the form actions to complete first?
```suggestion
        .then((res) => postFormCallback(res))
        .then(() => history.push(next))
```

frontend/packages/operator-lifecycle-manager/src/components/operand/operand-form.tsx
```diff
@@ -41,7 +42,10 @@ export const OperandForm: React.FC<OperandFormProps> = ({
 
   const handleSubmit = ({ formData: submitFormData }) => {
     k8sCreate(model, processFormData(submitFormData))
-      .then(() => next && history.push(next))
+      .then((res) => {
+        postFormCallback(res);
+        return next && history.push(next);
+      })
```
christianvogt, line 48:
```suggestion
      .then((res) => postFormCallback(res))
      .then(() => next && history.push(next))
```

invincibleJai:
Thanks @sahil143 , verified the changes works as expected

PR #7086, opened 2020-11-03
Bug 1891820: Add error message

rebeccaalpert:
The Edit Update Strategy modal allowed users to submit incorrect values (0). I added an error message so users aren't able to submit these values.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1891820.


spadgett:
These values *are* allowed to be 0. We just weren't showing the right value in the UI. See #7083

PR #7092, opened 2020-11-03
CONSOLE-2325: Add machine translations for Japanese

rebeccaalpert:
Added Japanese machine translations from Terry.

[CONSOLE-2325](https://issues.redhat.com/browse/CONSOLE-2325)


spadgett (approved):
/lgtm
/hold

No code changes here. This is just the initial drop of Japanese translations.

spadgett:
@yapei This can be tested by adding `?lng=ja` to the URL or by setting language to `ja` in the browser settings. (We're working on adding a modal to switch languages.)

yapei:
@spadgett  I have a general question about the translation user stories. How can we make sure all necessary contents are translated? For example, when only part of buttons/headings on one page is translated, how can we know if this is bug or working as intended?

spadgett:
> @spadgett I have a general question about the translation user stories. How can we make sure all necessary contents are translated? For example, when only part of buttons/headings on one page is translated, how can we know if this is bug or working as intended?

@yapei In general, I would reach out and ask if it's not clear. We still have some common components shared across pages that aren't finished, and parts of the UI we haven't gotten to yet. Some terms are treated as proper names and don't translated. For instance, "Namespace" was externalized, but appears to be intentionally left as-is in the Japanese translations. The Red Hat Globalization team will also be doing translation verification testing to make sure the content looks right for each language.

We'll do our best to let you know what is expected to be translated in the individual PRs.

PR #7095, opened 2020-11-04
Add API to create user settings

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4755

**Analysis / Root cause**:

Currently you need create this namespace yourself, **see Test Setup below**. This namespace will be automatically created by the console-operator, see PR https://github.com/openshift/console-operator/pull/479

**Solution Description**: 
Because the user could not create a ConfigMap here, we create a new `ConfigMap`, `Role` and `RoleBinding` with the service account role over the bridge. This code provides 3 new endpoints to create, get and delete the user settings. (The second and third API calls primary for debugging / development.) The console will fetch the ConfigMap over the known k8sWatch API.

**API**
`GET /api/console/user-settings`
Get the current user settings (returns a ConfigMap or 404)

`POST /api/console/user-settings`
Create a user setting ConfigMap for the current user. It returns the existing ConfigMap if it is already exist. This method does not accept an initial content for the ConfigMap.

`DELETE /api/console/user-settings`
Delete the user setting ConfigMap for the current user. Returns `204 No Content` if the ConfigMap was deleted, also if it was not created yet.

Irrelevant

```bash
➜  console git:([odc-4755](https://issues.redhat.com/browse/odc-4755)) go test -coverprofile=coverage.out github.com/openshift/console/pkg/usersettings  
ok  	github.com/openshift/console/pkg/usersettings	0.006s	coverage: 16.2% of statements

➜  console git:([odc-4755](https://issues.redhat.com/browse/odc-4755)) ✗ go tool cover -func=coverage.out
github.com/openshift/console/pkg/usersettings/handlers.go:37:	HandleUserSettings		0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:86:	getUserSettings			0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:92:	createUserSettings		0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:121:	deleteUserSettings		0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:140:	createServiceAccountClient	0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:149:	createUserProxyClient		0.0%
github.com/openshift/console/pkg/usersettings/handlers.go:158:	getUserSettingMeta		0.0%
github.com/openshift/console/pkg/usersettings/helpers.go:11:	getUserSettingMeta		100.0%
github.com/openshift/console/pkg/usersettings/helpers.go:29:	createRole			100.0%
github.com/openshift/console/pkg/usersettings/helpers.go:61:	createRoleBinding		100.0%
github.com/openshift/console/pkg/usersettings/helpers.go:85:	createConfigMap			100.0%
github.com/openshift/console/pkg/usersettings/types.go:10:	getConfigMapName		100.0%
github.com/openshift/console/pkg/usersettings/types.go:14:	getRoleName			100.0%
github.com/openshift/console/pkg/usersettings/types.go:18:	getRoleBindingName		100.0%
total:								(statements)			16.2%
```

**Test setup:**
This PR require some additional RBAC rules.

1. If PR https://github.com/openshift/console-operator/pull/479 is not part of your cluster you need to create a namespace:

1a Create namespace:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
    openshift.io/node-selector: ""
```

Until https://github.com/openshift/console-operator/pull/486/files is merged and available on your cluster you need to add this as well:

1b Create Role:
```yaml
kind: Role
metadata:
  name: console-user-settings-admin-fixed
  namespace: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
rules:
- apiGroups:
  - rbac.authorization.k8s.io
  resources:
  - roles
  - rolebindings
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
  - patch
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
  - patch
```

1c Create RoleBinding:
```yaml
kind: RoleBinding
metadata:
  name: console-user-settings-admin-fixed
  namespace: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
roleRef:
  kind: Role
  name: console-user-settings-admin-fixed
  apiGroup: rbac.authorization.k8s.io
subjects:
  - kind: ServiceAccount
    name: console
    namespace: openshift-console
```

2. Because the API calls needs a authentication and CSRF token, you can test the API calls from a logged in browser session with this JavaScript snippet:

In frontend code this is automatically done by `coFetch`, see [co-fetch.js](https://github.com/openshift/console/blob/master/frontend/public/co-fetch.js#L88-L97). But to test this without additional frontend code or postman I extracted the code for you:

2a Extract crcf token:
```js
let cookiePrefix = 'csrf-token=';
let getCSRFToken = () =>
  document &&
  document.cookie &&
  document.cookie
    .split(';')
    .map((c) => c.trim())
    .filter((c) => c.startsWith(cookiePrefix))
    .map((c) => c.slice(cookiePrefix.length))
    .pop();
getCSRFToken();
```

2b Get user-settings:
```js
fetch('/api/console/user-settings', { method: 'GET', headers: { 'X-CSRFToken': getCSRFToken() } }).then(
    (res) => {
        console.log('res', res);
        res.text().then(
            (text) => {
                console.log('text', text);
            },
            (parseErr) => console.warn('parseErr', parseErr)
        );
    },
    (requestErr) => {
        console.warn('requestErr', requestErr);
    }
);
```

2c Create user-settings:
```js
fetch('/api/console/user-settings', { method: 'POST', headers: { 'X-CSRFToken': getCSRFToken() } }).then(
    (res) => {
        console.log('res', res);
        res.text().then(
            (text) => {
                console.log('text', text);
            },
            (parseErr) => console.warn('parseErr', parseErr)
        );
    },
    (requestErr) => {
        console.warn('requestErr', requestErr);
    }
);
```

**Browser conformance**: 
Irrelevant

### DIFF
--- a/cmd/bridge/main.go
+++ b/cmd/bridge/main.go
@@ -543,29 +543,28 @@ func main() {
 		bridge.FlagFatalf("user-auth", "must be one of: oidc, openshift, disabled")
 	}
 
-	var resourceListerToken string
 	switch *fK8sAuth {
 	case "service-account":
 		bridge.ValidateFlagIs("k8s-mode", *fK8sMode, "in-cluster")
 		srv.StaticUser = &auth.User{
 			Token: k8sAuthServiceAccountBearerToken,
 		}
-		resourceListerToken = k8sAuthServiceAccountBearerToken
+		srv.ServiceAccountToken = k8sAuthServiceAccountBearerToken
 	case "bearer-token":
 		bridge.ValidateFlagNotEmpty("k8s-auth-bearer-token", *fK8sAuthBearerToken)
 		srv.StaticUser = &auth.User{
 			Token: *fK8sAuthBearerToken,
 		}
-		resourceListerToken = *fK8sAuthBearerToken
+		srv.ServiceAccountToken = *fK8sAuthBearerToken
 	case "oidc", "openshift":
 		bridge.ValidateFlagIs("user-auth", *fUserAuth, "oidc", "openshift")
-		resourceListerToken = k8sAuthServiceAccountBearerToken
+		srv.ServiceAccountToken = k8sAuthServiceAccountBearerToken
 	default:
 		bridge.FlagFatalf("k8s-mode", "must be one of: service-account, bearer-token, oidc, openshift")
 	}
 
 	srv.MonitoringDashboardConfigMapLister = server.NewResourceLister(
-		resourceListerToken,
+		srv.ServiceAccountToken,
 		&url.URL{
 			Scheme: k8sEndpoint.Scheme,
 			Host:   k8sEndpoint.Host,
@@ -583,7 +582,7 @@ func main() {
 	)
 
 	srv.KnativeEventSourceCRDLister = server.NewResourceLister(
-		resourceListerToken,
+		srv.ServiceAccountToken,
 		&url.URL{
 			Scheme: k8sEndpoint.Scheme,
 			Host:   k8sEndpoint.Host,
@@ -601,7 +600,7 @@ func main() {
 	)
 
 	srv.KnativeChannelCRDLister = server.NewResourceLister(
-		resourceListerToken,
+		srv.ServiceAccountToken,
 		&url.URL{
 			Scheme: k8sEndpoint.Scheme,
 			Host:   k8sEndpoint.Host,
--- a/pkg/usersettings/handlers.go
+++ b/pkg/usersettings/handlers.go
@@ -1,0 +1,178 @@
+package usersettings
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	core "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+
+	"github.com/openshift/console/pkg/auth"
+	"github.com/openshift/console/pkg/proxy"
+	"github.com/openshift/console/pkg/serverutils"
+)
+
+const namespace = "openshift-console-user-settings"
+
+var USER_RESOURCE = schema.GroupVersionResource{
+	Group:    "user.openshift.io",
+	Version:  "v1",
+	Resource: "users",
+}
+
+type UserSettingsHandler struct {
+	K8sProxyConfig      *proxy.Config
+	Client              *http.Client
+	Endpoint            string
+	ServiceAccountToken string
+}
+
+func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.ResponseWriter, r *http.Request) {
+	context := context.TODO()
+
+	serviceAccountClient, err := h.createServiceAccountClient()
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to create service account to handle user setting request: %v", err)
+		klog.Errorf(errMsg)
+		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+		return
+	}
+
+	userSettingMeta, err := h.getUserSettingMeta(context, user)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to get user data to handle user setting request: %v", err)
+		klog.Errorf(errMsg)
+		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+		return
+	}
+
+	switch r.Method {
+	case http.MethodGet:
+		configMap, err := h.getUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
+			errMsg := fmt.Sprintf("Failed to get user settings: %v", err)
+			klog.Errorf(errMsg)
+			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			return
+		}
+		serverutils.SendResponse(w, http.StatusOK, configMap)
+	case http.MethodPost:
+		configMap, err := h.createUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
+			errMsg := fmt.Sprintf("Failed to create user settings: %v", err)
+			klog.Errorf(errMsg)
+			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			return
+		}
+		serverutils.SendResponse(w, http.StatusOK, configMap)
+	case http.MethodDelete:
+		err := h.deleteUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
+			errMsg := fmt.Sprintf("Failed to delete user settings: %v", err)
+			klog.Errorf(errMsg)
+			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			return
+		}
+		w.WriteHeader(http.StatusNoContent)
+	default:
+		w.Header().Set("Allow", "GET, POST, DELETE")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Unsupported method, supported methods are GET POST DELETE"})
+	}
+}
+
+// Fetch the user-setting ConfigMap of the current user, by using his token.
+func (h *UserSettingsHandler) getUserSettings(ctx context.Context, client *kubernetes.Clientset, userSettingMeta *UserSettingMeta) (*core.ConfigMap, error) {
+	return client.CoreV1().ConfigMaps(namespace).Get(ctx, userSettingMeta.getConfigMapName(), meta.GetOptions{})
+}
+
+// Create a new user-setting ConfigMap, incl. Role and RoleBinding for the current user.
+// Returns the existing ConfigMap if it is already exist.
+func (h *UserSettingsHandler) createUserSettings(ctx context.Context, client *kubernetes.Clientset, userSettingMeta *UserSettingMeta) (*core.ConfigMap, error) {
+	role := createRole(userSettingMeta)
+	roleBinding := createRoleBinding(userSettingMeta)
+	configMap := createConfigMap(userSettingMeta)
+
+	_, err := client.RbacV1().Roles(namespace).Create(ctx, role, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		h.deleteUserSettings(ctx, client, userSettingMeta)
+		return nil, err
+	}
+
+	_, err = client.RbacV1().RoleBindings(namespace).Create(ctx, roleBinding, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		h.deleteUserSettings(ctx, client, userSettingMeta)
+		return nil, err
+	}
+
+	configMap, err = client.CoreV1().ConfigMaps(namespace).Create(ctx, configMap, meta.CreateOptions{})
+	if err != nil {
+		// Return actual ConfigMap if it is already created
+		if apierrors.IsAlreadyExists(err) {
+			klog.Infof("User settings ConfigMap \"%s\" already exist, will return existing data.", userSettingMeta.getConfigMapName())
+			return client.CoreV1().ConfigMaps(namespace).Get(ctx, userSettingMeta.getConfigMapName(), meta.GetOptions{})
+		}
+		h.deleteUserSettings(ctx, client, userSettingMeta)
+		return nil, err
+	}
+	return configMap, nil
+}
+
+// Deletes the user-setting ConfigMap, Role and RoleBinding of the current user.
+// It handles not found responses, so that it does not fail on multiple calls.
+func (h *UserSettingsHandler) deleteUserSettings(ctx context.Context, client *kubernetes.Clientset, userSettingMeta *UserSettingMeta) error {
+	err := client.RbacV1().RoleBindings(namespace).Delete(ctx, userSettingMeta.getRoleBindingName(), meta.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	err = client.RbacV1().Roles(namespace).Delete(ctx, userSettingMeta.getRoleName(), meta.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	err = client.CoreV1().ConfigMaps(namespace).Delete(ctx, userSettingMeta.getConfigMapName(), meta.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	return nil
+}
+
+func (h *UserSettingsHandler) createServiceAccountClient() (*kubernetes.Clientset, error) {
+	config := &rest.Config{
+		Host:        h.Endpoint,
+		BearerToken: h.ServiceAccountToken,
+		Transport:   h.Client.Transport,
+	}
+	return kubernetes.NewForConfig(config)
+}
+
+func (h *UserSettingsHandler) createUserProxyClient(user *auth.User) (dynamic.Interface, error) {
+	config := &rest.Config{
+		Host:        h.Endpoint,
+		BearerToken: user.Token,
+		Transport:   h.Client.Transport,
+	}
+	return dynamic.NewForConfig(config)
+}
+
+func (h *UserSettingsHandler) getUserSettingMeta(context context.Context, user *auth.User) (*UserSettingMeta, error) {
+	client, err := h.createUserProxyClient(user)
+	if err != nil {
+		return nil, err
+	}
+
+	userInfo, err := client.Resource(USER_RESOURCE).Get(context, "~", meta.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return newUserSettingMeta(userInfo.GetName(), string(userInfo.GetUID()))
+}
--- a/pkg/usersettings/handlers.go
+++ b/pkg/usersettings/handlers.go
@@ -39,17 +39,13 @@ func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.Respons
 
 	serviceAccountClient, err := h.createServiceAccountClient()
 	if err != nil {
-		errMsg := fmt.Sprintf("Failed to create service account to handle user setting request: %v", err)
-		klog.Errorf(errMsg)
-		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+		h.sendErrorResponse("Failed to create service account to handle user setting request: %v", err, w)
 		return
 	}
 
 	userSettingMeta, err := h.getUserSettingMeta(context, user)
 	if err != nil {
-		errMsg := fmt.Sprintf("Failed to get user data to handle user setting request: %v", err)
-		klog.Errorf(errMsg)
-		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+		h.sendErrorResponse("Failed to get user data to handle user setting request: %v", err, w)
 		return
 	}
 
@@ -57,27 +53,21 @@ func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.Respons
 	case http.MethodGet:
 		configMap, err := h.getUserSettings(context, serviceAccountClient, userSettingMeta)
 		if err != nil {
-			errMsg := fmt.Sprintf("Failed to get user settings: %v", err)
-			klog.Errorf(errMsg)
-			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			h.sendErrorResponse("Failed to get user settings: %v", err, w)
 			return
 		}
 		serverutils.SendResponse(w, http.StatusOK, configMap)
 	case http.MethodPost:
 		configMap, err := h.createUserSettings(context, serviceAccountClient, userSettingMeta)
 		if err != nil {
-			errMsg := fmt.Sprintf("Failed to create user settings: %v", err)
-			klog.Errorf(errMsg)
-			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			h.sendErrorResponse("Failed to create user settings: %v", err, w)
 			return
 		}
 		serverutils.SendResponse(w, http.StatusOK, configMap)
 	case http.MethodDelete:
 		err := h.deleteUserSettings(context, serviceAccountClient, userSettingMeta)
 		if err != nil {
-			errMsg := fmt.Sprintf("Failed to delete user settings: %v", err)
-			klog.Errorf(errMsg)
-			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
+			h.sendErrorResponse("Failed to delete user settings: %v", err, w)
 			return
 		}
 		w.WriteHeader(http.StatusNoContent)
@@ -85,6 +75,18 @@ func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.Respons
 		w.Header().Set("Allow", "GET, POST, DELETE")
 		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Unsupported method, supported methods are GET POST DELETE"})
 	}
+}
+
+func (h *UserSettingsHandler) sendErrorResponse(format string, err error, w http.ResponseWriter) {
+	errMsg := fmt.Sprintf(format, err)
+	klog.Errorf(errMsg)
+	code := http.StatusBadGateway
+	if apierrors.IsNotFound(err) {
+		code = http.StatusNotFound
+	} else if apierrors.IsForbidden(err) {
+		code = http.StatusForbidden
+	}
+	serverutils.SendResponse(w, code, serverutils.ApiError{Err: errMsg})
 }
 
 // Fetch the user-setting ConfigMap of the current user, by using his token.
--- a/pkg/usersettings/helpers.go
+++ b/pkg/usersettings/helpers.go
@@ -1,0 +1,95 @@
+package usersettings
+
+import (
+	"errors"
+
+	core "k8s.io/api/core/v1"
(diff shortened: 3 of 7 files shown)


jerolimov:
Note for myself, we can fetch the current user uid with this command!?

```bash
oc get users.user.openshift.io/~ -o template={{.metadata.uid}}
```

jerolimov:
@spadgett Can you take a look or assign this issue to someone who can review this change and https://github.com/openshift/console-operator/pull/479?

I need to remove the fmt.Print debug output and add some unit tests.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+
+	response, err := h.Client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("getCurrentUser %d %s\n", response.StatusCode, response.Status)
+
+	if response.StatusCode != 200 {
+		return nil, errors.New(fmt.Sprintf("Unexpected status code (%d instead of 200) when getting current user", response.StatusCode))
```
spadgett, line 240: Log this error.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+	request.Header.Set("Accept", "application/json")
+
+	response, err := h.Client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("getCurrentUser %d %s\n", response.StatusCode, response.Status)
+
+	if response.StatusCode != 200 {
```
spadgett, line 239: Use `http.StatusOK`

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+
+	fmt.Printf("sendRequest %d %s\n", response.StatusCode, response.Status)
+
+	return response, nil
+}
+
+func (h *UserSettingsHandler) copy(receivedResponse *http.Response, w http.ResponseWriter) {
+	w.Header().Set("Content-Type", receivedResponse.Header.Get("Content-Type"))
+	w.Header().Set("Content-Length", receivedResponse.Header.Get("Content-Length"))
+	w.WriteHeader(receivedResponse.StatusCode)
```
spadgett, line 331: At the very least, we need `Content-Encoding` as well if present. It's probably a good idea to add cache control headers. We should make sure the security-related headers are coming through as well.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+	path := fmt.Sprintf("/apis/rbac.authorization.k8s.io/v1/namespaces/%s/roles/%s", namespace, name)
+	return h.sendRequest(http.MethodDelete, path, nil)
+}
+
+func (h *UserSettingsHandler) deleteRoleBinding(namespace string, name string) (*http.Response, error) {
+	path := fmt.Sprintf("/apis/rbac.authorization.k8s.io/v1/namespaces/%s/rolebindings/%s", namespace, name)
+	return h.sendRequest(http.MethodDelete, path, nil)
+}
+
+func (h *UserSettingsHandler) sendRequest(method string, path string, content interface{}) (*http.Response, error) {
```
spadgett, line 296: Is there a reason we can't use client-go instead of rolling our own? We should have the library already as a dependency.

pkg/usersettings/types.go
```diff
@@ -0,0 +1,62 @@
+package usersettings
+
+// TODO: Convert this to official Namespace, ConfigMap types, etc.?
+
+type Metadata struct {
```
spadgett, line 5: We shouldn't need to do this ourselves since we have client-go.

spadgett (commented):
I've made a few comments, but it would be good to get @jhadvig to look since he is more familiar with the Go parts of console.

/assign @jhadvig

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// user id is missing, auth is used that does not support user info propagated, like OpenShift OAuth
+	currentUser, err := h.getCurrentUser(user)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Printf("Current user: %v\n", currentUser)
+	fmt.Printf("  Name: %s\n", currentUser.Metadata.Name)
+	fmt.Printf("  UID: %s\n", currentUser.Metadata.UID)
```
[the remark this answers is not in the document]
jerolimov: Yeah sure, will drop them all

pkg/usersettings/handlers.go
jerolimov: Should I pass all headers from the k8 response to the bridge response? I skipped this because that are more lines of code, but I think this make sense then.

spadgett: It's hard to make a recommendation since we're sort of reimplementing the proxy here. These are the headers we filter from response:

https://github.com/openshift/console/blob/master/pkg/proxy/proxy.go#L34-L40

Can we reuse the any of the proxy code here?

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+		Kind:       "ConfigMap",
+		Metadata: Metadata{
+			Namespace: namespaceName,
+			Name:      configMapName,
+		},
+	}
+
+	roleResponse, err := h.createRole(&role)
+	if err != nil {
+		fmt.Printf("createRole error: %v", err)
```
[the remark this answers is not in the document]
jerolimov: Replaced all fmt.Printfs with klog and removed all debug outputs.

pkg/usersettings/handlers.go
jerolimov: I replaced all http request calls with the `client-go` library, incl. fetching the ConfigMap. So we can not "proxy" the response here, but can just return the 'parsed' ConfigMap as json again.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+		TypeMeta: meta.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: meta.ObjectMeta{
+			Name: resourceData.getConfigMapName(),
+		},
+	}
+
+	client, err := h.createServiceAccountClient()
```
jhadvig, line 156: Also this call can be moved to the `HandleUserSettings()` and pas as param to the appropriate `UserSettingsHandler` method.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		klog.Warningf("Could not create user-settings Role \"%s\" for user \"%s\" (UID: %s)", resourceData.getRoleName(), resourceData.Username, resourceData.UID)
+	}
+
+	_, err = client.RbacV1().RoleBindings(USER_SETTINGS_NAMESPACE).Create(context, roleBinding, meta.CreateOptions{})
+	if err != nil {
+		klog.Warningf("Could not create user-settings Role \"%s\" for user \"%s\" (UID: %s)", resourceData.getRoleBindingName(), resourceData.Username, resourceData.UID)
+	}
+
+	configMap, err = client.CoreV1().ConfigMaps(USER_SETTINGS_NAMESPACE).Create(context, configMap, meta.CreateOptions{})
```
jhadvig, line 171: Lets make this shorter by:
```go
configMap, err = client.CoreV1().ConfigMaps(USER_SETTINGS_NAMESPACE).Create(context, configMap, meta.CreateOptions{})
if err != nil {
	klog.Warningf("Could not create user-settings ConfigMap \"%s\" for user \"%s\" (UID: %s)", resourceData.getConfigMapName(), resourceData.Username, resourceData.UID)
	// Return actual ConfigMap if it is already created
	if apierrors.IsAlreadyExists(err) {
		return client.CoreV1().ConfigMaps(USER_SETTINGS_NAMESPACE).Get(context, resourceData.getConfigMapName(), meta.GetOptions{}), err
	}
	return nil, err
}
return configMap, nil
```

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = client.RbacV1().Roles(USER_SETTINGS_NAMESPACE).Create(context, role, meta.CreateOptions{})
+	if err != nil {
+		klog.Warningf("Could not create user-settings Role \"%s\" for user \"%s\" (UID: %s)", resourceData.getRoleName(), resourceData.Username, resourceData.UID)
+	}
+
+	_, err = client.RbacV1().RoleBindings(USER_SETTINGS_NAMESPACE).Create(context, roleBinding, meta.CreateOptions{})
```
jhadvig, line 166: What happens when one of these resources(Role/Rolebinding/SA) fails to get created? Don't think we should proceed with creating the rest of them and leave it in an inconsistent state.

pkg/usersettings/types.go
```diff
@@ -0,0 +1,20 @@
+package usersettings
+
+type ResourceData struct {
```
jhadvig, line 3: ResourceData sounds a bit odd. How about `UserSettingData` or maybe `UserSettingMeta` ?

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, err
+	}
+
+	userInfo, err := client.Resource(USER_RESOURCE).Get(context, "~", meta.GetOptions{})
+	if err != nil {
+		klog.Infof("getCurrentUser error: %v", err)
+		return nil, err
+	}
+
+	name := string(userInfo.GetName())
```
jhadvig, line 248: `GetName()` is returning a string.
```suggestion
	name := string(userInfo.GetName())
```

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (h *UserSettingsHandler) getResourceData(context context.Context, user *auth.User) (*ResourceData, error) {
+	client, err := h.createUserProxyClient(user)
+	if err != nil {
+		return nil, err
+	}
+
+	userInfo, err := client.Resource(USER_RESOURCE).Get(context, "~", meta.GetOptions{})
+	if err != nil {
+		klog.Infof("getCurrentUser error: %v", err)
```
jhadvig, line 244:
```suggestion
		klog.Errorf("failed to GET the user resource: %v", err)
```

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+	return dynamic.NewForConfig(config)
+}
+
+func (h *UserSettingsHandler) getResourceData(context context.Context, user *auth.User) (*ResourceData, error) {
+	client, err := h.createUserProxyClient(user)
+	if err != nil {
+		return nil, err
+	}
+
+	userInfo, err := client.Resource(USER_RESOURCE).Get(context, "~", meta.GetOptions{})
```
jhadvig, line 242: OOC, does this match `oc get users.user.openshift.io/~` command ?

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, err
+	}
+
+	name := string(userInfo.GetName())
+	uid := string(userInfo.GetUID())
+	resourceSuffix := fmt.Sprintf("-%s", uid)
+
+	if name == "kube:admin" {
+		resourceSuffix = "-kubeadmin"
+	} else if uid == "" {
```
jhadvig, line 254: Lets move this right after you initialise the `uid` var, so we are retuning prematurely

pkg/usersettings/types.go
```diff
@@ -0,0 +1,20 @@
...
+	Username string
+	UID      string
+	// The resource suffix uses the User uid like "-12341234-1234-1234-1234-123412341234" or "" for kube:admin
+	ResourceSuffix string
```
jhadvig, line 7: Since this is not an suffix in case of the Role and RoleBinding I would name it `ResourceIdentifier`. Maybe?

pkg/usersettings/types.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+func (r *ResourceData) getConfigMapName() string {
```
jhadvig, line 10: Would be good to add unit tests for these methods.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+					"watch",
+				},
+				ResourceNames: []string{
+					resourceData.getConfigMapName(),
+				},
+			},
+		},
+	}
+
+	roleBinding := &rbac.RoleBinding{
```
jhadvig, line 124: Lets create a helper for this, so we can add unit-test.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+// Returns the existing ConfigMap if it is already exist.
+func (h *UserSettingsHandler) createUserSettings(user *auth.User) (*core.ConfigMap, error) {
+	context := context.TODO()
+
+	resourceData, err := h.getResourceData(context, user)
+	if err != nil {
+		return nil, err
+	}
+
+	role := &rbac.Role{
```
jhadvig, line 94: Lets create a helper for this, so we can add unit-test.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+			},
+		},
+		RoleRef: rbac.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     resourceData.getRoleName(),
+		},
+	}
+
+	configMap := &core.ConfigMap{
```
jhadvig, line 146: Lets create a helper for this, so we can add unit-test.

cmd/bridge/main.go
```diff
@@ -543,29 +543,28 @@ func main() {
 		bridge.FlagFatalf("user-auth", "must be one of: oidc, openshift, disabled")
 	}
 
-	var resourceListerToken string
```
jhadvig, line 546: 👍

jhadvig (changes requested):

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+	default:
+		w.Header().Set("Allow", "GET, POST, DELETE")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Unsupported method, supported methods are GET POST DELETE"})
+	}
+}
+
+// Fetch the current user with auth.User token and returning the user-setting ConfigMap.
+func (h *UserSettingsHandler) getUserSettings(user *auth.User) (*core.ConfigMap, error) {
+	context := context.TODO()
+	resourceData, err := h.getResourceData(context, user)
```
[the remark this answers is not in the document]
jerolimov: I thought this also, but are unsure how an OPTION request would handle this. Should I add extra if for an OPTION request before that or should we check if this works locally and on a cluster. I will change this and we see what happen :)

pkg/usersettings/handlers.go
jerolimov: Outdated since I switched to client-go, also added some klogs

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,340 @@
...
+	Endpoint            string
+	ServiceAccountToken string
+}
+
+func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case http.MethodGet:
+		response, err := h.getUserSettings(user)
+		if err != nil {
+			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: fmt.Sprintf("Failed to get user settings: %v", err)})
```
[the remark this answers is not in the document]
jerolimov: Added klog to all error cases in `HandleUserSettings`

pkg/usersettings/handlers.go
jerolimov: Outdated after the switch to `go-client`, we now proxy no data in user settings handler anymore. Data are returned by `serverutils.SendResponse`

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	default:
+		w.Header().Set("Allow", "GET, POST, DELETE")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Unsupported method, supported methods are GET POST DELETE"})
+	}
+}
+
+// Fetch the current user with auth.User token and returning the user-setting ConfigMap.
+func (h *UserSettingsHandler) getUserSettings(user *auth.User) (*core.ConfigMap, error) {
+	context := context.TODO()
```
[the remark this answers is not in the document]
jerolimov: I moved the context, createServiceAccountClient and getResourceData now at the top of `HandleUserSettings` so that there is just one place for it

pkg/usersettings/handlers.go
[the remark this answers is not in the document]
jerolimov: Moved context, createServiceAccountClient and getResourceData to `HandleUserSettings`

pkg/usersettings/handlers.go
jerolimov: Done

pkg/usersettings/handlers.go
jerolimov: :+1: Adopted

pkg/usersettings/handlers.go
jerolimov: Yep, was ignored before. I switched to this code to ignore just an `AlreadyExists` error. Everything else let the call fail and was logged then in `HandleUserSettings`.

```go
	if err != nil && !apierrors.IsAlreadyExists(err) {
		return nil, err
	}
```

pkg/usersettings/types.go
jerolimov: Had the same feeling but no better name. :smile: Change it to `UserSettingMeta`

pkg/usersettings/handlers.go
jerolimov: Dropped this log because the error was returned and logged now already in `HandleUserSettings`

pkg/usersettings/handlers.go
jerolimov: Yes :+1:

jerolimov: @jhadvig we have some similar code in [terminal/proxy.go line 119-136](https://github.com/openshift/console/blob/master/pkg/terminal/proxy.go#L119-L136), but I don't want touch other code in the first step. What do you think?

pkg/usersettings/handlers.go
jerolimov: We couldn't, the kube:admin has no uid. So I need keep that check first. See `helpers.go` `getUserSettingMeta` now

pkg/usersettings/types.go
jerolimov: Renamed.

Alternative, I could create one type like this:

```go
type resourceIdentifier struct {
	userName string
	configMapName string
	roleName string
	roleBindingName string
}
```

Wdyt? I like the idea.. :thinking:

pkg/usersettings/types.go
jerolimov: I would like to move this from "name getters / helpers" to the moment when creating the ResourceNames. See comment above.

But for the moment I added tests for the "name getters / helpers" here as well.

pkg/usersettings/handlers.go
jerolimov: Done, see helpers.go

pkg/usersettings/handlers.go
jerolimov: Done, see helpers.go and helpers_test.go

pkg/usersettings/handlers.go
jerolimov: Done, see helpers.go and helpers_test.go

jerolimov:
@jhadvig All done, I had one small idea for another cleanup. And I can squash all commits leter. Thanks for reviewing this.

pkg/usersettings/helpers.go
```diff
@@ -0,0 +1,95 @@
...
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func getUserSettingMeta(name string, uid string) (*UserSettingMeta, error) {
+	resourceIdentifier := ""
+
+	if name == "kube:admin" {
+		resourceIdentifier = "kubeadmin"
+	} else if uid != "" {
+		resourceIdentifier = uid
```
christianvogt, line 17: Make sure there is no UID for the kubeadmin case:
```suggestion
	if uid != "" {
		resourceIdentifier = uid
	} else if name == "kube:admin" {
		resourceIdentifier = "kubeadmin"
```

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+		configMap, err := h.getUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
+			klog.Errorf("Failed to get user settings: %v", err)
+			serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: fmt.Sprintf("Failed to get user settings: %v", err)})
+		} else {
+			serverutils.SendResponse(w, http.StatusOK, configMap)
+		}
+	case http.MethodPost:
+		configMap, err := h.createUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
```
jhadvig, line 65: The preferred idiomatic way for go, how to structure the response should be:
```go
if err != nil {
        errMsg := fmt.Sprintf("Failed to create user settings: %v", err)
	klog.Errorf(errMsg)
	serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: errMsg})
	return
}
serverutils.SendResponse(w, http.StatusOK, configMap)
```
so we return immediatelly

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+		} else {
+			w.WriteHeader(http.StatusNoContent)
+		}
+	default:
+		w.Header().Set("Allow", "GET, POST, DELETE")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Unsupported method, supported methods are GET POST DELETE"})
+	}
+}
+
+// Fetch the current user with auth.User token and returning the user-setting ConfigMap.
```
jhadvig, line 85:
```suggestion
// Fetch the user-setting ConfigMap of the current user, by using his token
```

pkg/usersettings/handlers.go
jhadvig: yeah, thats probably ok

pkg/usersettings/helpers.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"errors"
+
+	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func getUserSettingMeta(name string, uid string) (*UserSettingMeta, error) {
```
jhadvig, line 11:
```suggestion
func newUserSettingMeta(name string, uid string) (*UserSettingMeta, error) {
```

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+	configMap := createConfigMap(userSettingMeta)
+
+	_, err := client.RbacV1().Roles(namespace).Create(ctx, role, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return nil, err
+	}
+
+	_, err = client.RbacV1().RoleBindings(namespace).Create(ctx, roleBinding, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return nil, err
```
jhadvig, line 104: we should call `deleteUserSettings()` method upon error. So we dont end up in an inconsistent state, where Role and RoleBinding was created but the CM for some reason was not.
Either that or we rely that the request would be repeated until all the resources are created.

jhadvig (changes requested):

pkg/usersettings/helpers.go
jerolimov: Changed.

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+	Endpoint            string
+	ServiceAccountToken string
+}
+
+func (h *UserSettingsHandler) HandleUserSettings(user *auth.User, w http.ResponseWriter, r *http.Request) {
+	context := context.TODO()
+
+	serviceAccountClient, err := h.createServiceAccountClient()
+	if err != nil {
+		klog.Errorf("Failed to create service account to handle user setting request: %v", err)
```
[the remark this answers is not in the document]
jerolimov: done

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: fmt.Sprintf("Failed to create service account to handle user setting request: %v", err)})
+		return
+	}
+
+	userSettingMeta, err := h.getUserSettingMeta(context, user)
+	if err != nil {
+		klog.Errorf("Failed to get user data to handle user setting request: %v", err)
```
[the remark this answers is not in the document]
jerolimov: done

pkg/usersettings/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+		serverutils.SendResponse(w, http.StatusBadGateway, serverutils.ApiError{Err: fmt.Sprintf("Failed to get user data to handle user setting request: %v", err)})
+		return
+	}
+
+	switch r.Method {
+	case http.MethodGet:
+		configMap, err := h.getUserSettings(context, serviceAccountClient, userSettingMeta)
+		if err != nil {
+			klog.Errorf("Failed to get user settings: %v", err)
```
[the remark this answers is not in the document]
jerolimov: done

pkg/usersettings/handlers.go
jerolimov: done

pkg/usersettings/handlers.go
jerolimov: done

pkg/usersettings/helpers.go
jerolimov: Renamed

pkg/usersettings/handlers.go
jerolimov: As discussed in slack, I added `deletedUserSettings` before returning the create error.

jhadvig (approved):
/lgtm

PR #7096, opened 2020-11-04
Bug 1894330: update channels list

yapei:
Updated channels list for 4.7

### DIFF
--- a/frontend/public/module/k8s/cluster-settings.ts
+++ b/frontend/public/module/k8s/cluster-settings.ts
@@ -42,7 +42,7 @@ export const getSortedUpdates = (cv: ClusterVersionKind): ClusterUpdate[] => {
 export const getAvailableClusterChannels = (cv) => {
   // temporarily fall back to hard-coded values when `cv.status.desired.channels` are not present
   // TODO: remove fall back values once `cv.status.desired.channels` are widespread
-  return cv?.status?.desired?.channels || ['stable-4.6', 'fast-4.6', 'candidate-4.6'];
+  return cv?.status?.desired?.channels || ['stable-4.7', 'fast-4.7', 'candidate-4.7'];
 };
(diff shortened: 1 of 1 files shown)


spadgett:
Hi, @yapei. This should no longer be necessary as we now get the list of channels from the API since #6283, although I see you're updating the fallback. It might be better to just remove the fallback so we don't have to maintain this release-to-release. The ClusterVersion API should have the valid values.

cc @rhamilto

rhamilto:
> Hi, @yapei. This should no longer be necessary as we now get the list of channels from the API since #6283, although I see you're updating the fallback. It might be better to just remove the fallback so we don't have to maintain this release-to-release. The ClusterVersion API should have the valid values.
> 
> cc @rhamilto

Yeah.  We should remove the fallback **and**  there is also [a TODO to not link to the modal if `cv.spec.desired.channels` does not exist for a build](https://github.com/openshift/console/blob/master/frontend/public/components/cluster-settings/cluster-settings.tsx#L144) (e.g., a nightly).  If I remember correctly, @megan-hall had a design task for this.

megan-hall:
@rhamilto yes - I do. I will circle back with you offline and get that taken care of.

rhamilto:
IMO, we should probably close this PR since we don't want to remove the fallback without implementing the pending design changes at the same time as we'll end up with a modal containing no options in the dropdown if we do.

rhamilto:
@megan-hall's design story is https://issues.redhat.com/browse/PD-579

yapei:
Got it, let me close the PR

spadgett:
Let's go ahead and update the fallback for now so it's not wrong, and we can circle back when we have a design for when there are no channels.

/reopen

spadgett (approved):
/lgtm

PR #7099, opened 2020-11-04
Bug 1894374: prevent prevent loading by type

yaacov:
The file to upload using the upload screen does not have to have any extension, it is quite normal to have a raw file without any extension (e.g. my-fedora-disk and not my-fedora-disk.img). The problem is, that we block the user from uploading such a file with an error.

Screenshots:
After:
Peek 2020-11-04 10-28

Before:
Peek 2020-11-04 09-41

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
@@ -81,27 +81,12 @@ const templatesResource: WatchK8sResource = {
 enum uploadErrorType {
   MISSING = 'missing',
   ALLOCATE = 'allocate',
-  TYPE = 'type',
   CERT = 'cert',
 }
 
 const uploadErrorMessage = {
   [uploadErrorType.MISSING]: 'File input is missing',
   [uploadErrorType.ALLOCATE]: 'Could not create persistent volume claim',
-  [uploadErrorType.TYPE]: (
-    <>
-      <p>
-        The format of the file you are uploading is not supported. Please use one of the supported
-        formats
-      </p>
-      <p>
-        <ExternalLink
-          text="Learn more about supported formats"
-          href={CDI_UPLOAD_SUPPORTED_TYPES_URL}
-        />
-      </p>
-    </>
-  ),
   [uploadErrorType.CERT]: (uploadProxy) => (
     <>
       It seems that your browser does not trust the certificate of the upload proxy. Please{' '}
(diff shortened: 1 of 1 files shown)


yaacov:
@matthewcarleton @glekner @pcbailey please review

yaacov:
@matthewcarleton I added a hint about the file extensions as you suggested in Jira, please re review:
screenshot-localhost_9000-2020 11 05-12_30_17
screenshot-localhost_9000-2020 11 05-12_31_25

/retest

matthewcarleton:
@yaacov yes that's better :) 
Can you also use an[ inline alert](https://www.patternfly.org/v4/components/alert/html#inline-types) (.pf-m-inline)

yaacov:
> Can you also use an inline alert (.pf-m-inline)

Thanks, inline now, please re-review:
screenshot-localhost_9000-2020 11 06-11_48_05

matthewcarleton:
lgtm!

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
```diff
@@ -529,6 +513,9 @@ export const UploadPVCPage: React.FC<UploadPVCPageProps> = (props) => {
   const handleFileChange = (value, filename) => {
     setFileName(filename);
     setFileValue(value);
+
+    setFileNameExtension(/[.][^.]+$/.exec(filename)?.toString());
+    setIsFileRejected(false);
```
glekner, line 518: when is isFileRejected true? @yaacov

glekner:
@yaacov please squash the commits so I can lgtm

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
yaacov: it is set by the `UploadPVCForm` when file extension is unexpected, but is not reset back on field clear, this line reset it back when the offending file is cleared but no new file is selected.

yaacov:
> please squash the commits so I can lgtm

@glekner squashed :+1:

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
```diff
@@ -574,6 +561,26 @@ export const UploadPVCPage: React.FC<UploadPVCPageProps> = (props) => {
...
+                </p>
+                <p>
+                  <ExternalLink
+                    text="Learn more about supported formats"
+                    href={CDI_UPLOAD_SUPPORTED_TYPES_URL}
+                  />
+                </p>
+              </Alert>
+            ) : (
+              undefined
```
glekner, line 582: why ternary condition?
just do `isFileRejected &&`

glekner (changes requested):

frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
yaacov: `isFileRejected` is boolean, so if we use `&&` we get into a grey area of how to render `false` [1] , I preferred here to be explicit and make sure I understand `isFileRejected` is boolean, by explicitly render `{undefined}` instead of `false` which is redundant but explicit. On the other hand it does looks prettier with `&&` so changing to using `&&` :+1: 

[1] https://react-cn.github.io/react/tips/false-in-jsx.html

PR #7112, opened 2020-11-05
Bug 1891047: Access server API via kubernetes.default.svc from Helm endpoints

pedjak:
Previously, `Server.KubeAPIServerURL` was used and on some clusters this URL certificate might not verifiable using in-cluster CA.


jhadvig (approved):
/lgtm

spadgett (approved):
/lgtm

PR #7131, opened 2020-11-05
generic component to highlight any part of the ui

nemesis09:
**Fixes:**
https://issues.redhat.com/browse/ODC-4999

**Description:**
Create a generic component to highlight any part of UI with animation

**Screens:**
_highlight component behaviour ideal_
Note: three components are highlighted simultaneously for illustration only.
quickstarts_highlight

_highlight component behaviour when clicked anywhere on the ui after animation has started_
stophighlightanimation

_highlight component behaviour autoscroll to bring component to highlight in viewport if it is hidden_
highlightautoscroll

_guided tour behaviour unchanged with updated `Spotlight` component_
guidedtour_highlight

**Test Coverage:**
Screenshot from 2020-11-11 01-37-50

**Browser Conformance:**
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


frontend/packages/console-shared/src/components/highlight/Highlight.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const element = document.querySelector(selector);
+  const { height, width, top, left } = element.getBoundingClientRect();
+  const style: React.CSSProperties = {
+    top,
+    left,
+    height,
+    width,
+  };
+  return (
+    <Portal>
```
rohitkrai03, line 19: The implementation is exactly same as the Spotlight component from guided tour. Instead of creating a new one you should just move that one to console-shared and add some customization for this use case.

rohitkrai03 (changes requested):

frontend/packages/console-shared/src/components/highlight/Highlight.tsx
nemesis09: updated with customized spotlight component

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <Portal>
+      {hasBackdrop ? (
+        <div className="pf-c-backdrop ocs-spotlight__with-backdrop">
+          <div
+            className={cx({
+              'ocs-spotlight__element-highlight-animate': animate,
+              'ocs-spotlight__element-highlight-noanimate': !animate,
+            })}
+            style={style}
```
christianvogt, line 30: DRY. Store this component in a variable and render conditionally.

lwrigh:
@nemesis09 Here are the animation specifications from the final design doc: 

Blink animation is triggered by click/tap initiating step
Blink animation is delayed start by 0.5 seconds ⬅️
Blink animation consists of (all at the same time and over 1.2 sec duration)
Scale: shape increases size from 100% to 150% ⬅️
Border width: from 4px to 8px ⬅️
Opacity: from 100% to 0%

I took a closer look at the GIF and CSS and I think the elements that are missing are the specifications with the arrows next to them. Let me know if you need me to take a closer look at any animation specifications. 

Here's the link to the final design doc if you need to reference it: https://docs.google.com/document/d/1T41IzREwgfqR6NVpXI8e4psfkBR39gWt0eLfvA9pkGs/edit?usp=sharing

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: updated using variable.

nemesis09:
> @nemesis09 Here are the animation specifications from the final design doc:
> 
> Blink animation is triggered by click/tap initiating step
> Blink animation is delayed start by 0.5 seconds arrow_left
> Blink animation consists of (all at the same time and over 1.2 sec duration)
> Scale: shape increases size from 100% to 150% arrow_left
> Border width: from 4px to 8px arrow_left
> Opacity: from 100% to 0%
> 
> I took a closer look at the GIF and CSS and I think the elements that are missing are the specifications with the arrows next to them. Let me know if you need me to take a closer look at any animation specifications.
> 
> Here's the link to the final design doc if you need to reference it: https://docs.google.com/document/d/1T41IzREwgfqR6NVpXI8e4psfkBR39gWt0eLfvA9pkGs/edit?usp=sharing

updated animation addressing the suggested requirements. PTAL

lwrigh:
> > @nemesis09 Here are the animation specifications from the final design doc:
> > Blink animation is triggered by click/tap initiating step
> > Blink animation is delayed start by 0.5 seconds arrow_left
> > Blink animation consists of (all at the same time and over 1.2 sec duration)
> > Scale: shape increases size from 100% to 150% arrow_left
> > Border width: from 4px to 8px arrow_left
> > Opacity: from 100% to 0%
> > I took a closer look at the GIF and CSS and I think the elements that are missing are the specifications with the arrows next to them. Let me know if you need me to take a closer look at any animation specifications.
> > Here's the link to the final design doc if you need to reference it: https://docs.google.com/document/d/1T41IzREwgfqR6NVpXI8e4psfkBR39gWt0eLfvA9pkGs/edit?usp=sharing
> 
> updated animation addressing the suggested requirements. PTAL

The updates you made @nemesis09 look good to me! From the GIFs it looks like the behaviors and timing work out well.

lwrigh:
@mceledonia Here's the PR associated with the animated quick start hints. Feel free to take a look. The GIFs at the top of the PR are the most recent version of the implementation.

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
```diff
@@ -0,0 +1,91 @@
+@keyframes highlight {
+  0% {
+    opacity: 1;
+    border: var(--pf-global--BorderWidth--xl) solid var(--pf-global--palette--blue-200);
```
sahil143, line 4: Why didn't we use box shadow here as shown in https://codesandbox.io/s/highlight-9oizs

nemesis09: the UX for animation was updated
please find the updated UX for animation [here](https://xd.adobe.com/view/dd20d755-8def-4536-af2a-86ba83a8080c-8873/)
Slack Conversation links: 
https://coreos.slack.com/archives/C014NQG3H5L/p1604609753173100
https://coreos.slack.com/archives/C014NQG3H5L/p1604610208176400

christianvogt:
@nemesis09 a few observations from your images:
- While the highlight border animates, a border around the item should remain in place.
- The animation spec listed here mentions nothing about the outline shrinking.
- The border should remain for 10 seconds after the animation, without fading IMO, and then fade out quickly or simply disappear. Your animation looks like a 10second slow fade out.
- Your 3rd gif shows that when the menu is scrolled, the outline fails to follow the nav item.
- I believe the guided tour outline should be the same color as the quick start highlight now.

I'm waiting on @mceledonia to provide updates on the animation before reviewing the code.

mceledonia:
I'll run through each step of the animation and add the transitions and timings in code styling. I'll also add some visual aids for each step.

For the animation of all hint borders: 

- The user clicks the text link to activate the hint border.
- The border appears as two exact copies around the element, since they are two exact copies stacked on top of each other they just look like one. They should both `Fade in over 0.4 seconds` Right now they are static and do not move. The reason for 2 copies to appear is to achieve the desired animation effect described below.
image 
image
- After a brief `delay of 1.2 seconds` one copy of the border begins it's animation sequence. The animation sequence is as follows: `increase in scale from 100% to 150% over 1.2 seconds` `increase in border width from 4px to 8px over 1.2 seconds` `decrease in opacity/transparency from 100% to 0% over 1.2 seconds` Note - These all happen simultaneously at the same time, and finish at the same time. They do not happen one after another.
image
image
- The leftover static copy `waits/delays 10 seconds` and then triggers a `fade out over 5 seconds`
image

The spec for the starting border styling is the following:
Border width: 4px
Border color: `$pf-color-blue-200:` or `#73bcf7`

christianvogt:
Here's an updated implementation of the requested animation:
https://codesandbox.io/s/highlight-9oizs?file=/src/Highlight.css

Note that I'm currently working with @mceledonia to tweak the animation (eg border expansion from 4px -> 12px instead of 4px -> 8px)

christianvogt:
@nemesis09 please use the same outline for both the guided tour and quick starts highlight.

christianvogt:
(I modified the guided tour to test this)
You need to use popperjs to position the highlight so that it moves when the node underneath it moves. Otherwise this happens:
image

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
```diff
@@ -0,0 +1,64 @@
+@keyframes expand {
```
christianvogt, line 1:
```suggestion
@keyframes co-spotlight-expand {
```

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+    opacity: 1;
+  }
+  100% {
+    outline-offset: 21px;
+    outline-width: 12px;
+    opacity: 0;
+  }
+}
+
+@keyframes fade-in {
```
christianvogt, line 14:
```suggestion
@keyframes co-spotlight-fade-in {
```

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+  0% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
+
+@keyframes fade-out {
```
christianvogt, line 23:
```suggestion
@keyframes co-spotlight-fade-out {
```

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import cx from 'classnames';
+import Portal from '@console/shared/src/components/popper/Portal';
+import './Spotlight.scss';
+
+type SpotlightProps = {
+  selector: string;
+  animate?: boolean;
+  hasBackdrop?: boolean;
```
christianvogt, line 9: There's a bug where if `hasBackdrop === false && animate === false` then this happens:
image

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <div
+      className={cx('ocs-spotlight', {
+        'ocs-spotlight__element-highlight-animate': animate && !clicked,
+        'ocs-spotlight__element-highlight-noanimate': !animate,
+      })}
+      style={style}
+    />
+  );
+  return (
+    <Portal>
```
christianvogt, line 58: We will need to use popperjs instead so that the spotlight moves as the node underneath it moves.

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { height, width, top, left } = element.getBoundingClientRect();
+  const style: React.CSSProperties = {
+    top,
+    left,
+    height,
+    width,
+    ['--ocs-spotlight-top' as React.ReactText]: `${top}px`,
+    ['--ocs-spotlight-left' as React.ReactText]: `${left}px`,
+    ['--ocs-spotlight-height' as React.ReactText]: `${height}px`,
+    ['--ocs-spotlight-width' as React.ReactText]: `${width}px`,
```
christianvogt, line 33: Clean up dead code.

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
christianvogt, line 9: Combine these 2 props into one called `interactive` and add a description.
If `interactive` it will allow click through and animate. If `!interactive` it will display the background and prevent click through.

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [clicked, setClicked] = React.useState(false);
+  React.useEffect(() => {
+    if (!isInViewport(element)) {
+      element.scrollIntoView();
+    }
+    const handleClick = () => setClicked(true);
+    document.addEventListener('mousedown', handleClick);
+    return () => {
+      document.removeEventListener('mousedown', handleClick);
+    };
```
christianvogt, line 45: Only listen for clicks if in `interactive` mode.
You should also make sure the click occurs on the target element and not anywhere on screen.

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+  100% {
+    opacity: 0;
+  }
+}
+
+.ocs-spotlight {
+  pointer-events: none;
+  position: absolute;
+  &__with-backdrop {
+    mix-blend-mode: hard-light;
```
christianvogt, line 36: This causes the blue for the guided tour to be slightly a different color than the one we actually want. Perhaps raise a separate issue for this since it was an issue before as well.
image

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
andrewballantyne:
```suggestion
    const handleClick = () => setClicked(true);
    element.addEventListener('mousedown', handleClick);
    return () => {
      element.removeEventListener('mousedown', handleClick);
    };
```

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [clicked, setClicked] = React.useState(false);
+  React.useEffect(() => {
+    if (!isInViewport(element)) {
+      element.scrollIntoView();
+    }
+    if (!interactive) {
+      const handleClick = () => setClicked(true);
+      document.addEventListener('mousedown', handleClick);
+      return () => {
+        document.removeEventListener('mousedown', handleClick);
```
andrewballantyne, line 38:
```suggestion
      element.addEventListener('mousedown', handleClick);
      return () => {
        element.removeEventListener('mousedown', handleClick);
```
This was not updated...

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: raised issue https://issues.redhat.com/browse/ODC-5178

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: using popperjs instead of portal

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: the goal for this function is to detect any click outside the `link` and stop the animation.

> The solid rectangle hint persists until the user clicks somewhere else on the UI or waits the 10 seconds until it fades away. 
although reading the uxd again, it seems the `click-anywhere-to-disappear` should only apply to the solid rectangle persisting after the expand animation. I'll check this with ux

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: I'm not sure if we need to check for click event on the element, 
`element` is the component to highlight and the UXD says a click anywhere on the screen should stop the animation

christianvogt: I'll check with ux on the design again

christianvogt: I checked and you are right. They want it to disappear when clicking anywhere.

christianvogt:
@nemesis09 this is what the guided tour now looks like:
image

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <div
+      className={cx('ocs-spotlight', {
+        'ocs-spotlight__element-highlight-animate': interactive && !clicked,
+        'ocs-spotlight__element-highlight-noanimate': !interactive,
+      })}
+      style={style}
+    />
+  );
+  return (
+    <Popper reference={element} placement="top-start">
```
andrewballantyne, line 54: Your popper is trying to be too smart... it's pushing away from the edge of the screen.

```suggestion
    <Popper reference={element} placement="top-start" popperOptions={popperOptions}>
```

```
  import { PopperOptions } from 'popper.js';
  ...
  const popperOptions: PopperOptions = {
    modifiers: {
      preventOverflow: {
        boundariesElement: 'window',
        padding: 0,
      },
    },
  };
```

frontend/packages/console-shared/src/components/spotlight/__tests__/Spotlight.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should render StaticSpotlight if interactive is not set to true', () => {
+    wrapper = shallow(<Spotlight selector="selector" />);
+    expect(
+      wrapper
+        .find(StaticSpotlight)
+        .first()
```
andrewballantyne, line 24:
```suggestion
```

frontend/packages/console-shared/src/components/spotlight/__tests__/Spotlight.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+        .exists(),
+    ).toBe(true);
+  });
+
+  it('should render InteractiveSpotlight if interactive is set to true', () => {
+    wrapper = shallow(<Spotlight selector="selector" interactive />);
+    expect(
+      wrapper
+        .find(InteractiveSpotlight)
+        .first()
```
andrewballantyne, line 34:
```suggestion
```

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.ocs-spotlight {
+  pointer-events: none;
+  position: absolute;
+  &__with-backdrop {
+    mix-blend-mode: hard-light;
+  }
+  &__element-highlight-noanimate {
+    position: absolute;
```
andrewballantyne, line 39:
```suggestion
```

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+    border: var(--pf-global--BorderWidth--xl) solid var(--pf-global--palette--blue-200);
+    background-color: var(--pf-global--palette--black-500);
+    z-index: 9999;
+  }
+  &__element-highlight-animate {
+    pointer-events: none;
+    position: absolute;
```
andrewballantyne, line 46:
```suggestion
```

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  React.useEffect(() => {
+    if (!isInViewport(element)) {
+      element.scrollIntoView();
+    }
+    const handleClick = () => setClicked(true);
+    document.addEventListener('mousedown', handleClick);
+    return () => {
+      document.removeEventListener('mousedown', handleClick);
+    };
```
andrewballantyne, line 42:
```suggestion
    const handleClick = () => setClicked(true);
    document.addEventListener('click', handleClick);
    return () => {
      document.removeEventListener('click', handleClick);
    };
```
Changing these to `click` will help support mobile. `mousedown` is a non touch screen event.

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
```diff
@@ -0,0 +1,64 @@
...
+    opacity: 1;
+  }
+  100% {
+    opacity: 0;
+  }
+}
+
+.ocs-spotlight {
+  pointer-events: none;
+  position: absolute;
```
andrewballantyne, line 34: I don't think this `position` is needed either since you're not specifying the `top`/`left` esk attributes.

andrewballantyne (changes requested):
nits

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: updated.

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
nemesis09: this is being used by guided tours as well which uses top/left.

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/__tests__/Spotlight.spec.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/__tests__/Spotlight.spec.tsx
nemesis09: done

nemesis09:
> @nemesis09 this is what the guided tour now looks like:
> image

fixed

frontend/packages/console-shared/src/components/spotlight/spotlight.scss
andrewballantyne: 🤔  Pretty sure if you don't use `top` / `left` of 0 it does wonky things... but okay. Let's leave it and worry about it if it's an issue.

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
rohitkrai03: Shouldn't this be `ocs-spotlight-expand`?

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
rohitkrai03: @christianvogt @nemesis09 Shouldn't this be `ocs-spotlight-expand` since this is in `console-shared` package?

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
rohitkrai03: @christianvogt @nemesis09 Shouldn't this be `ocs-spotlight-expand` since this is in `console-shared` package?

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
andrewballantyne: Not sure it matters... we have a poor consistency across the console for keyframes and their names having the appropriate module. I think it's because they are only ever used in the one file. I'm indifferent either way though.

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
christianvogt: yes it should be. my mistake

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
christianvogt: yes it should be. my mistake

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
christianvogt: yes it should be. my mistake

andrewballantyne (approved):
/lgtm

/assign @christianvogt  @rohitkrai03 

Need an approval here when one of you get a chance.

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: renamed to `ocs-spotlight-expand`

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: enamed to ocs-spotlight-expand

frontend/packages/console-shared/src/components/spotlight/Spotlight.scss
nemesis09: renamed to `ocs-spotlight-expand`

andrewballantyne:
Odd, pod didn't start...

> * 1x kubelet: Successfully pulled image "quay.io/prometheus/busybox:latest" in 1m20.65114373s
2020/11/30 16:34:27 pod didn't start running within 15m0s: 
> * Container artifacts is not ready with reason ContainerCreating
> * Container release is not ready with reason ContainerCreating

/retest

rohitkrai03 (changes requested):
@nemesis09 I was testing this with guided tours by passing interactive prop to spotlight. I noticed some weird behaviors.

- The outside click doesn't work on any part of the document. It only works when clicked on the popper part of the guided tour.
- The borders on the help menu looks off. It looks okay without the `interactive` prop passed but the arrow is still off a bit from the center.
 
Screenshot from 2020-11-30 22-30-54

Screenshot from 2020-11-30 22-29-21

christianvogt:
Interesting that the popover for help is also offset.

andrewballantyne:
I don't understand how others are reproducing this situation lol... Christian on his Mac can reproduce... Aritra on Linux and Rohit on Linux -- but I can't on my Mac lol

Chrome
Screen Shot 2020-11-30 at 1 57 39 PM

FF
Screen Shot 2020-11-30 at 1 58 12 PM

andrewballantyne:
> The outside click doesn't work on any part of the document. It only works when clicked on the popper part of the guided tour.

@rohitkrai03 are you saying during Guided Tour you are expecting to click on the mask and exit? Because the interactive mode is false and thus the only way to close the modal is to close it via the `x` or finish the flow (or use one the exit at the start)

nemesis09:
> @nemesis09 I was testing this with guided tours by passing interactive prop to spotlight. I noticed some weird behaviors.
> 
>     * The outside click doesn't work on any part of the document. It only works when clicked on the popper part of the guided tour.
> 
>     * The borders on the help menu looks off. It looks okay without the `interactive` prop passed but the arrow is still off a bit from the center.
> 
> 
> Screenshot from 2020-11-30 22-30-54
> 
> Screenshot from 2020-11-30 22-29-21

This is odd. As far as I understand, there has been no substantial change in the spotlight as far as the implementation for Guided Tour is concerned.

andrewballantyne:
> This is odd. As far as I understand, there has been no substantial change in the spotlight as far as the implementation for Guided Tour is concerned.

@rohitkrai03 / @christianvogt / @nemesis09  Can you fire up any cluster (before this is merged) and check the Guided Tour? Is this a pre-existing issue? I obviously can't reproduce this 😞

nemesis09:
> I don't understand how others are reproducing this situation lol... Christian on his Mac can reproduce... Aritra on Linux and Rohit on Linux -- but I can't on my Mac lol
> 
> Chrome
> Screen Shot 2020-11-30 at 1 57 39 PM
> 
> FF
> Screen Shot 2020-11-30 at 1 58 12 PM

Update: I tried it out in the cluster, rather than localhost to check the differences. The offset in the arrow seems to be a pre-existing issue. I think this could be related to a similar behaviour in popover arrow that i noticed in the modals for cluster status card items, which later turned out to be a patternfly issue. here is the slack thread https://coreos.slack.com/archives/C6A3NV5J9/p1603962222473200

nemesis09:
> @nemesis09 I was testing this with guided tours by passing interactive prop to spotlight. I noticed some weird behaviors.
> 
>     * The outside click doesn't work on any part of the document. It only works when clicked on the popper part of the guided tour.
> 
>     * The borders on the help menu looks off. It looks okay without the `interactive` prop passed but the arrow is still off a bit from the center.
> 
> 
> Screenshot from 2020-11-30 22-30-54
> 
> Screenshot from 2020-11-30 22-29-21

Update: I tried it out in the cluster, rather than localhost to check the differences. The offset in the arrow seems to be a pre-existing issue. I think this could be related to a similar behaviour in popover arrow that i noticed in the modals for cluster status card items, which later turned out to be a patternfly issue. here is the slack thread https://coreos.slack.com/archives/C6A3NV5J9/p1603962222473200

rohitkrai03:
> > The outside click doesn't work on any part of the document. It only works when clicked on the popper part of the guided tour.
> 
> @rohitkrai03 are you saying during Guided Tour you are expecting to click on the mask and exit? Because the interactive mode is false and thus the only way to close the modal is to close it via the `x` or finish the flow (or use one the exit at the start)

@andrewballantyne I tested guided tour after passing `interactive` prop to the `Spotlight` component. So my expectation was that clicking anywhere should stop the animation and fade out the border but it only works when clicked on the popover.

The below screenshot shows the original use of `Spotlight` without interactive prop. The arrow seems to be offset which is a pre-existing issue even before this PR but the border box around help is correctly placed.
Screenshot from 2020-11-30 22-30-54

The below screenshot shows the `Spotlight` with `interactive` prop passed so that it uses `InteractiveSpotlight` with animation. It has the border box around help icon offset to the icon but the arrow is now centered with the border box.
Screenshot from 2020-11-30 22-29-21

andrewballantyne:
> @andrewballantyne I tested guided tour after passing `interactive` prop to the `Spotlight` component. So my expectation was that clicking anywhere should stop the animation and fade out the border but it only works when clicked on the popover

Well that's disconcerting... I'll have to test this out once I am finished with my current task.


> The below screenshot shows the original use of `Spotlight` without interactive prop. The arrow seems to be offset which is a pre-existing issue even before this PR but the border box around help is correctly placed.
> Screenshot from 2020-11-30 22-30-54
> 
> The below screenshot shows the `Spotlight` with `interactive` prop passed so that it uses `InteractiveSpotlight` with animation. It has the border box around help icon offset to the icon but the arrow is now centered with the border box.
> Screenshot from 2020-11-30 22-29-21

This is very confusing lol... I can't reproduce this in any fashion... I will have to try all the browsers under my disposal to see if I can reproduce this. Either way, I think we are addressing this post this PR merge.

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
```diff
@@ -0,0 +1,19 @@
...
+import InteractiveSpotlight from './InteractiveSpotlight';
+import StaticSpotlight from './StaticSpotlight';
+
+type SpotlightProps = {
+  selector: string;
+  interactive?: boolean;
+};
+
+const Spotlight: React.FC<SpotlightProps> = ({ selector, interactive }) => {
+  const element = document.querySelector(selector);
```
christianvogt, line 11:
```suggestion
  const element = React.useMemo(() => document.querySelector(selector), [selector]);
```

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
```diff
@@ -0,0 +1,54 @@
...
+  };
+  const isInViewport = (elementToCheck: Element) => {
+    const rect = elementToCheck.getBoundingClientRect();
+    return (
+      rect.top >= 0 &&
+      rect.left >= 0 &&
+      rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
+      rect.right <= (window.innerWidth || document.documentElement.clientWidth)
+    );
+  };
```
christianvogt, line 31: This is a utility function. Move it out of the render closure.

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
```diff
@@ -0,0 +1,54 @@
...
+    height,
+    width,
+  };
+  const popperOptions: PopperOptions = {
+    modifiers: {
+      preventOverflow: {
+        enabled: false,
+      },
+    },
+  };
```
christianvogt, line 22: Make this a const outside of the render function.

frontend/packages/console-shared/src/components/spotlight/Spotlight.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
nemesis09: done

frontend/packages/console-shared/src/components/spotlight/InteractiveSpotlight.tsx
nemesis09: done

andrewballantyne (approved):
/lgtm

PR #7135, opened 2020-11-06
Fix failing unit test which was being ignored by CI reporting.

akashshinde:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1894860


akashshinde:
Closing since no longer needed

PR #7140, opened 2020-11-06
Bug 1878701: Filter out events from old VM with the same nam

jelkosz:
Added 2 checks:
1: the VM object events filter now makes sure it checks the VM UID, not only the name

2: filtered out the events which happened before the VM was created.
Unfortunately, there is no way to filter them based on the UID of the VM since
the object which contains the owner reference with UID (like the migration pod)
is/might be deleted by now and all what is left is an event

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
+++ b/frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
@@ -1,6 +1,6 @@
-import { getName, getNamespace } from '@console/shared';
+import { getName, getNamespace, getUID, getCreationTimestamp } from '@console/shared';
 import { PodModel } from '@console/internal/models';
-import { EventInvolvedObject } from '@console/internal/module/k8s';
+import { EventInvolvedObject, EventKind } from '@console/internal/module/k8s';
 import {
   VirtualMachineImportModel,
   VirtualMachineInstanceMigrationModel,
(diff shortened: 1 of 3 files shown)


jelkosz:
@rawagner @suomiy PTAL

frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
```diff
@@ -10,18 +10,33 @@ import {
...
 const vmEventFilter = (vm: VMILikeEntityKind): EventFilterFunction => ({
   kind,
   namespace,
   name,
+  uid,
 }) => {
   return (
-    kind === VirtualMachineModel.kind && name === getName(vm) && namespace === getNamespace(vm)
+    kind === VirtualMachineModel.kind &&
+    name === getName(vm) &&
```
rawagner, line 25: i think we dont need to check name and namespace anymore, `uid` is enough

frontend/public/components/events.jsx
```diff
@@ -386,7 +386,10 @@ class EventStream extends React.Component {
       if (kind && !kindFilter(kind, obj)) {
         return false;
       }
-      if (filter && !filter.some((flt) => flt(obj.involvedObject))) {
+      if (
+        filter &&
+        !filter.some((flt) => flt({ ...obj.involvedObject, timestamp: getFirstTime(obj) }))
```
rawagner, line 391: I suggest to pass the whole event object and keep the backwards compatibility:

```suggestion
        !filter.some((flt) => flt(obj.involvedObject, obj))
```

passing just `getFirstTime` does not give us enough flexibility for the future

frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
```diff
@@ -10,18 +10,33 @@ import {
 import { VIRT_LAUNCHER_POD_PREFIX } from '../../constants/vm';
 import { VMILikeEntityKind } from '../../types/vmLike';
 
-type EventFilterFunction = (src: EventInvolvedObject) => boolean;
+type EventInvolvedObjectWithTimestamp = EventInvolvedObject & { timestamp?: string };
+
+type EventFilterFunction = (src: EventInvolvedObjectWithTimestamp) => boolean;
```
rawagner, line 15: given what I wrote about passing whole event to filter function this will become
```suggestion
type EventFilterFunction = (src: EventInvolvedObject, event: EventKind) => boolean;
```

frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
jelkosz: done

frontend/public/components/events.jsx
jelkosz: done

frontend/packages/kubevirt-plugin/src/selectors/event/filters.ts
jelkosz: done

PR #7143, opened 2020-11-06
i18n support in devconsole Pipeline components

vikram-raj:
Fixes: https://issues.redhat.com/browse/ODC-5096


invincibleJai:
/lgtm cancel

@vikram-raj this needs rebase

frontend/packages/dev-console/src/components/pipelineruns/PipelineRunsPage.tsx
```diff
@@ -22,7 +24,9 @@ const PipelineRunsPage: React.FC<PipelineRunsPageProps> = (props) => {
       title={PipelineRunModel.labelPlural}
```
karthikjeeyar, line 24: Needs to be translated

karthikjeeyar:
Components that need to be translation

1. RunDetailsErrorLog.tsx (Fields - Message, Log Snippet) 
2. StartPipelineForm.tsx (All the components inside it has title, which needs to be changed)
3. Pipeline Actions and Kebab Menus.
  
image
4. PipelineDetails Page ( Trigger Templates Section)

image
5. ResourceLinkList.tsx Component (title)

frontend/packages/dev-console/src/components/pipelineruns/PipelineRunsPage.tsx
vikram-raj: Adding translation to it doesn't affect as it a variable.

jerolimov:
@vikram-raj @andrewballantyne do we want translate this status messages as well? They are part of the runStatus enum in https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/pipelineruns/status/PipelineBars.tsx#L24-L28 so we need a mapping for all status keys (or values)?

image

vikram-raj:
@jerolimov Translated the status.

Screenshot 2020-11-12 at 3 14 43 PM

vikram-raj:
> @vikram-raj @andrewballantyne should we translate the status column and status overlay for PipelineRuns as well?
> 
> They are part of the runStatus enum in https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/pipelineruns/status/PipelineBars.tsx#L24-L28 so we need a mapping for all status keys (or values)?
> 
> image

@jerolimov We are using the status string in some places and enum for status at someplace and status is being used in condition at many places. So, adding the translation to it will be tricky and I would suggest handling the status translation into a separate PR. let me know WDYT?

karthikjeeyar:
Thanks @vikram-raj Since it includes refactoring  in many places, then its good idea to have it separate PR so that we can have test cases to ensure that it doesn't break existing behaviour.
 /lgtm

jerolimov:
+1 @vikram-raj, please create a new jira ticket for the status fields. I think that's possible to use translation right in the `PipelineBars` and `TaskStatusToolTip` component to replace status labels with translated status labels without refactoring the enums.

frontend/public/components/utils/kebab.tsx
```diff
@@ -543,6 +544,7 @@ export type KebabAction = (
   obj: K8sResourceKind,
   resources?: any,
   customData?: any,
+  t?: TFunction,
```
andrewballantyne, line 547: 🤔  Has the Admin team not done any kebab translations? Seems odd for us to be adding it in this PR.

vikram-raj: No, I did not see any action translations yet.

andrewballantyne: @vikram-raj Please check with the tmp-i18n channel. This feels weird, not sure what they are doing about this right now.

christianvogt: You can use the `labelKey` property for now.

andrewballantyne:
This unfortunately needs to be cherry-picked / rebased onto the new pipeline-plugin package. Should be a fairly simple 1:1 transition. Let me know if you have any concerns @vikram-raj

frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
```diff
@@ -197,7 +203,8 @@ export const stopPipelineRun: KebabAction = (kind: K8sKind, pipelineRun: Pipelin
 };
 
 const addTrigger: KebabAction = (kind: K8sKind, pipeline: Pipeline) => ({
-  label: 'Add Trigger',
+  // t('pipelines-plugin~Add Trigger')
```
karthikjeeyar, line 206: Please remove all the commented statements in this file and also rebase the PR.

frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
```diff
@@ -162,14 +166,16 @@ export const rerunPipelineRunAndRedirect: KebabAction = (
 ) => {
   return rerunPipeline(kind, pipelineRun, null, {
     onComplete: handlePipelineRunSubmit,
-    label: 'Rerun',
+    // t('pipelines-plugin~Rerun')
+    label: 'pipelines-plugin~Rerun',
```
karthikjeeyar, line 170:
```suggestion
    labelKey: 'pipelines-plugin~Rerun',
```

frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
christianvogt: I believe these are necessary for the locale parser to generate the strings

vikram-raj: These comments are needed.

PR #7144, opened 2020-11-06
CONSOLE-2464: Automatically collapse metadata.managedFields

JPinkney:
This PR makes it so that when a yaml editor opens, it checks to see if metadata.managedFields exists and if it does then collapse it. Unfortunately, the YAML language server doesn't provide folding ranges yet so the only alternative is to look through the AST manually.

It ends up looking like: https://www.youtube.com/watch?v=_FDwyjQN444

### DIFF
--- a/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
+++ b/frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
@@ -9,6 +9,7 @@ import { getLanguageService, TextDocument } from 'yaml-language-server';
 import { openAPItoJSONSchema } from '@console/internal/module/k8s/openapi-to-json-schema';
 import { getSwaggerDefinitions } from '@console/internal/module/k8s/swagger';
 import { global_BackgroundColor_dark_100 as editorBackground } from '@patternfly/react-tokens';
+import * as yaml from 'yaml-ast-parser';
 
 window.monaco.editor.defineTheme('console', {
   base: 'vs-dark',
@@ -28,7 +29,16 @@ window.monaco.editor.defineTheme('console', {
   },
 });
 
+const { Range } = window.monaco;
+
 export const defaultEditorOptions = { readOnly: false, scrollBeyondLastLine: false };
+
+const MODEL_URI = 'inmemory://model.yaml';
+const MONACO_URI = monaco.Uri.parse(MODEL_URI);
+
+const createDocument = (model) => {
+  return TextDocument.create(MODEL_URI, model.getModeId(), model.getVersionId(), model.getValue());
+};
 
 // Unfortunately, `editor.focus()` doesn't work when hiding the shortcuts
 // popover. We need to find the actual DOM element.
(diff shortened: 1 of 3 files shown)


spadgett:
@alimobrem @eparis FYI

frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
```diff
@@ -147,6 +148,42 @@ export const registerYAMLHover = (languageID, monaco, m2p, p2m, createDocument,
   });
 };
 
+export const registerYAMLFolding = (languageID, monaco, createDocument) => {
+  monaco.languages.registerFoldingRangeProvider(languageID, {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    provideFoldingRanges(model, context, token) {
```
spadgett, line 154: Will this disable all of the other folding ranges so I can't fold sections like `metadata`, `spec`, or `status`?

JPinkney: Yeah, I just realized this. I have another alternative way that would keep all the folding sections + be able to fold managedFields. I'm just making those changes now

JPinkney: I've made the changes and now the managedFields property is collapsed and all the other folding sections show up as well!

sleshchenko (approved):
:+1: This feature will make people prefer OpenShift Console rather than chectl which can't provide such UX =)

frontend/packages/console-shared/src/components/editor/yaml-editor-utils.ts
```diff
@@ -177,8 +221,16 @@ export const enableYAMLValidation = (monaco, p2m, monacoURI, createDocument, yam
...
   };
 
+  let initialFoldingTriggered = false;
+
   getModel().onDidChangeContent(() => {
     const document = createDocument(getModel());
+
+    if (!initialFoldingTriggered && document.getText() !== '') {
+      fold(editor, monaco);
+      initialFoldingTriggered = true;
```
spadgett, line 231: It looks like the managed fields will get expanded again when the user clicks "Reload"

spadgett (commented):
Thanks @JPinkney. Really appreciate your tracking down how to make this work.

There's a bit of flicker when the editor loads, but I'm not sure there's a way around that. Even so, I think it's still better to fold the managed fields by default.

schituku:
/qe-approved
  tested the fix and the field comes collapsed by default.

spadgett:
@JPinkney FYI, there's a pending review comment if you could look. The PR will need to be rebased as well.

JPinkney:
@spadgett I've rebased and everything should be working now

spadgett (approved):
/lgtm
/hold for approvals
/assign @yapei @ahardin-rh @reestr

spadgett:
/unassign @reestr 
/assign @sferich888
for px-approval

PR #7145, opened 2020-11-06
[ODC-4388] Allow cluster admins to create terminals

JPinkney:
This PR is part of https://issues.redhat.com/browse/ODC-4388 which is for giving privileged users access to create terminal.

Currently, the flow is:
If the user is a cluster-admin then deny them from opening a terminal.

Now, the flow is:
If the user is a cluster-admin then automatically create their terminal in a secure openshift-terminal namespace. If the openshift-terminal namespace does not exist then create it.

The backend code then checks if the current user is a cluster-admin and if they are then check that the workspace they are trying to access is in the openshift-terminal namespace.

Before: https://www.youtube.com/watch?v=TUri-TE52UA
After: https://www.youtube.com/watch?v=v6Xy7LKbYrU

You can test with: `quay.io/jpinkney/console:515`

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
@@ -8,11 +8,13 @@ import { withUserSettingsCompatibility, WithUserSettingsCompatibilityProps } fro
 import CloudshellExec from './CloudShellExec';
 import TerminalLoadingBox from './TerminalLoadingBox';
 import { TerminalInitData, initTerminal } from './cloud-shell-utils';
-import CloudShellSetup from './setup/CloudShellSetup';
 import useCloudShellWorkspace from './useCloudShellWorkspace';
 import { CLOUD_SHELL_NAMESPACE, CLOUD_SHELL_NAMESPACE_CONFIG_STORAGE_KEY } from './const';
 
 import './CloudShellTerminal.scss';
+import CloudShellAdminSetup from './setup/CloudShellAdminSetup';
+import CloudShellDeveloperSetup from './setup/CloudShellDeveloperSetup';
+import { useAccessReview2 } from '@console/internal/components/utils/rbac';
 
 type StateProps = {
   user: UserKind;
@@ -33,8 +35,12 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps &
 }) => {
   const [initData, setInitData] = React.useState<TerminalInitData>();
   const [initError, setInitError] = React.useState<string>();
-
-  const [workspace, loaded, loadError] = useCloudShellWorkspace(user, namespace);
+  const [isAdmin, isAdminCheckLoading] = useAccessReview2({
+    namespace: 'openshift-terminal',
+    verb: 'create',
+    resource: 'pods',
+  });
+  const [workspace, loaded, loadError] = useCloudShellWorkspace(user, isAdmin, namespace);
 
   const workspacePhase = workspace?.status?.phase;
   const workspaceName = workspace?.metadata?.name;
@@ -113,7 +119,7 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps &
   }
 
   // loading the workspace resource
-  if (!loaded) {
+  if (!loaded || isAdminCheckLoading) {
     return <TerminalLoadingBox message="" />;
   }
 
@@ -138,9 +144,19 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps &
     );
   }
 
+  if (isAdmin) {
+    return (
+      <CloudShellAdminSetup
+        onInitialize={(ns: string) => {
+          setNamespace(ns);
+        }}
+      />
+    );
+  }
+
   // show the form to let the user create a new workspace
   return (
-    <CloudShellSetup
+    <CloudShellDeveloperSetup
       onCancel={onCancel}
       onSubmit={(ns: string) => {
         setNamespace(ns);
--- a/frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
@@ -1,0 +1,74 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+
+import { k8sCreate, k8sGet } from '@console/internal/module/k8s';
+import { WorkspaceModel } from '../../../models';
+import {
+  newCloudShellWorkSpace,
+  createCloudShellResourceName,
+  CLOUD_SHELL_PROTECTED_NAMESPACE,
+} from '../cloud-shell-utils';
+import { NamespaceModel } from '@console/internal/models';
+import TerminalLoadingBox from '../TerminalLoadingBox';
+import { LoadError } from '@console/internal/components/utils/status-box';
+import { useTranslation } from 'react-i18next';
+
+type Props = {
+  onInitialize: (namespace: string) => void;
+};
+
+const CloudShellAdminSetup: React.FunctionComponent<Props> = ({ onInitialize }) => {
+  const { t } = useTranslation();
+
+  const [initError, setInitError] = React.useState<string>();
+  React.useEffect(() => {
+    (async () => {
+      async function namespaceExists(): Promise<boolean> {
+        try {
+          await k8sGet(NamespaceModel, CLOUD_SHELL_PROTECTED_NAMESPACE);
+          return true;
+        } catch (error) {
+          if (error.json.code !== 404) {
+            setInitError(error);
+          }
+          return false;
+        }
+      }
+
+      try {
+        const protectedNamespaceExists = await namespaceExists();
+        if (!protectedNamespaceExists) {
+          await k8sCreate(NamespaceModel, {
+            metadata: {
+              name: CLOUD_SHELL_PROTECTED_NAMESPACE,
+            },
+          });
+        }
+        await k8sCreate(
+          WorkspaceModel,
+          newCloudShellWorkSpace(createCloudShellResourceName(), CLOUD_SHELL_PROTECTED_NAMESPACE),
+        );
+        onInitialize(CLOUD_SHELL_PROTECTED_NAMESPACE);
+      } catch (error) {
+        setInitError(error);
+      }
+    })();
+    // Don't include dependencies because if the CLOUD_SHELL_PROTECTED_NAMESPACE
+    // is not found a refresh will be triggered, creating an extra terminal
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (initError) {
+    return (
+      <LoadError message={initError} label={t('cloudshell~OpenShift command line terminal')} />
+    );
+  }
+
+  return (
+    <div className="co-cloudshell-terminal__container">
+      <TerminalLoadingBox />
+    </div>
+  );
+};
+
+export default connect()(CloudShellAdminSetup);
--- a/frontend/packages/console-app/src/components/cloud-shell/useCloudShellWorkspace.ts
+++ b/frontend/packages/console-app/src/components/cloud-shell/useCloudShellWorkspace.ts
@@ -11,6 +11,7 @@ import {
   CloudShellResource,
   CLOUD_SHELL_RESTRICTED_ANNOTATION,
   startWorkspace,
+  CLOUD_SHELL_PROTECTED_NAMESPACE,
 } from './cloud-shell-utils';
 import { useAccessReview2 } from '@console/internal/components/utils';
 import { ProjectModel } from '@console/internal/models';
@@ -30,6 +31,7 @@ const findWorkspace = (data?: CloudShellResource[]): CloudShellResource | undefi
 
 const useCloudShellWorkspace = (
   user: UserKind,
+  isClusterAdmin: boolean,
   defaultNamespace: string = null,
 ): WatchK8sResult<CloudShellResource> => {
   const [namespace, setNamespace] = useSafetyFirst(defaultNamespace);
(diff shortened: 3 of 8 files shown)


pkg/terminal/proxy.go
```diff
@@ -89,16 +89,6 @@ func (p *Proxy) HandleProxy(user *auth.User, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	enabledForUser, err := p.checkUserPermissions(user.Token)
-	if err != nil {
-		http.Error(w, "Failed to check workspace operator state. Cause: "+err.Error(), http.StatusInternalServerError)
-		return
-	}
-	if !enabledForUser {
-		http.Error(w, "Terminal is disabled for cluster-admin users.", http.StatusForbidden)
```
sleshchenko, line 98: I wonder if it makes sense to keep logic on proxy side that will check that workspace is run in protected namespace if the current user is cluster admin?
It would help to provide a bit safer solution if cluster-admin created a terminal in non-protected namespace, and then opened terminal via direct URL, like `${OS_CONSOLE_URL}/console/terminal/non-protected-namespace/workpsace-name` (<- I'm not sure it's the right URL for openning terminal in a dedicated tab)

P.S. This may be already discussed on WTO Cabal or in tmp-terminal channel but I haven't reviewed them yet.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -138,9 +161,20 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps> = ({ user, onCancel
     );
   }
 
+  if (admin) {
+    return (
+      <CloudShellAdminSetup
```
sleshchenko, line 166: Seems in addition to admin-specific logic here, we need to modify place where we pick the existing dashboard. To reproduce it just apply with terminal the following and try console
```yaml
apiVersion: workspace.devfile.io/v1alpha1
kind: DevWorkspace
metadata:
  annotations:
    controller.devfile.io/restricted-access: "true"
  labels:
    console.openshift.io/terminal: "true"
    controller.devfile.io/creator: ""
  name: terminal-kasasa
  namespace: default
spec:
  routingClass: web-terminal
  started: true
  template:
    components:
    - plugin:
        id: redhat-developer/web-terminal/4.5.0
        name: web-terminal
```

Simpler admin specific logic for picking existing devworkspace - could be just ignore terminals in different namespace from openshift-terminal.

A bit better but requires some UI - if any found out from openshift-terminal namespace - propose admin to remove it and create in openshift-terminal.

I'm fully fine with simpler option.

sleshchenko (commented):
Everything apart from the shared corner-case works just fine

pkg/terminal/proxy.go
```diff
@@ -89,19 +89,19 @@ func (p *Proxy) HandleProxy(user *auth.User, w http.ResponseWriter, r *http.Requ
...
 	}
 
-	enabledForUser, err := p.checkUserPermissions(user.Token)
+	ok, namespace, workspaceName, path := stripTerminalAPIPrefix(r.URL.Path)
+	if !ok {
+		http.NotFound(w, r)
+		return
+	}
+
+	enabledForUser, err := p.checkUserPermissions(user.Token, namespace)
```
sleshchenko, line 98: Should we sync checkUserPermissions with the frontend part which has
```
if (group === 'system:cluster-admins') {
```

JPinkney: Technically yes, but this isn't as easy on the backend as the user object doesn't have groups AFAICT.

@spadgett or @christianvogt Do you know how to do this? Essentially the flow could be: When the user tries to open a terminal living in the openshift-terminal namespace check if their user is in the cluster-admin group. If they are not then deny.

On the frontend, we seem to get the groups that the user is associated with but on the backend it's not as simple

sleshchenko: The logic in https://github.com/openshift/console/pull/7145/commits/59b3c77556b89748e8eda0878c9233a905c9e956 does not seem to be right but I may be wrong.
I assume it should be rewritten in the following way: https://github.com/JPinkney/console/compare/secured-namespaces...sleshchenko:secured-namespace-patch?expand=1

serenamarie125:
This is aligned with one of our committed epics - confirmed that this is not dependent on updated to the WTO.

@christianvogt can you help get reviews on this?

serenamarie125 (approved):
Thanks for doing this!  LGTM

serenamarie125:
FYI @parvathyvr @alimobrem

sleshchenko (approved):
Tested on crc (1.18.0+bb304aa with OpenShift 4.6) with my built console and everything works as expected:
1. Cluster admin is not able to exec into developer workspace.
2. Cluster admin gets terminal in openshift-terminal namespace.
3. Console ignores all terminals which are not in openshift-terminal namespace for cluster admin.
4. Backend denies to /exec/init for cluster admin if terminal is not in openshift-terminal.

Good job!

frontend/packages/console-app/src/components/cloud-shell/useCloudShellWorkspace.ts
```diff
@@ -70,8 +72,12 @@ const useCloudShellWorkspace = (
     if (!canListWorkspaces) {
       result.namespace = namespace;
     }
+
+    if (isClusterAdmin) {
+      result.namespace = CLOUD_SHELL_PROTECTED_NAMESPACE;
+    }
```
christianvogt, line 78:
```suggestion
    if (isClusterAdmin) {
      result.namespace = CLOUD_SHELL_PROTECTED_NAMESPACE;
    } else if (!canListWorkspaces) {
      result.namespace = namespace;
    }
```

frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  React.useEffect(() => {
+    (async () => {
+      async function namespaceExists(): Promise<boolean> {
+        try {
+          await k8sGet(NamespaceModel, CLOUD_SHELL_PROTECTED_NAMESPACE);
+          return true;
+        } catch (error) {
+          const namespaceError = `namespaces "${CLOUD_SHELL_PROTECTED_NAMESPACE}" not found`;
+          if (error.message !== namespaceError) {
```
christianvogt, line 33: I don't like this error message comparison. Can you use the status code instead? eg 404

frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
```diff
@@ -0,0 +1,77 @@
...
+  createCloudShellResourceName,
+  CLOUD_SHELL_PROTECTED_NAMESPACE,
+} from '../cloud-shell-utils';
+import { NamespaceModel } from '@console/internal/models';
+import TerminalLoadingBox from '../TerminalLoadingBox';
+import { LoadError } from '@console/internal/components/utils/status-box';
+import { useTranslation } from 'react-i18next';
+
+type Props = {
+  onInitialize?: (namespace: string) => void;
```
christianvogt, line 17:
```suggestion
  onInitialize: (namespace: string) => void;
```

frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
```diff
@@ -0,0 +1,77 @@
...
+            metadata: {
+              name: CLOUD_SHELL_PROTECTED_NAMESPACE,
+            },
+          });
+        }
+        await k8sCreate(
+          WorkspaceModel,
+          newCloudShellWorkSpace(createCloudShellResourceName(), CLOUD_SHELL_PROTECTED_NAMESPACE),
+        );
+        onInitialize && onInitialize(CLOUD_SHELL_PROTECTED_NAMESPACE);
```
christianvogt, line 57: `onInitialize` should be a required prop
```suggestion
        onInitialize(CLOUD_SHELL_PROTECTED_NAMESPACE);
```

christianvogt:
Tried this and works as expected on the surface.
However, when I first installed the operator and opened the terminal as an admin, I noticed that 3 DevWorkspace resources were created inside the `openshift-terminal` namespace.
I cannot reproduce this again on the same cluster anymore though....

JPinkney:
@christianvogt Regarding:

> However, when I first installed the operator and opened the terminal as an admin, I noticed that 3 DevWorkspace resources were created inside the openshift-terminal namespace.
> I cannot reproduce this again on the same cluster anymore though....

Can you try with the latest changes? I just got a fresh cluster from cluster bot and deployed the changes and looks like I can't reproduce anymore

christianvogt:
@sleshchenko @JPinkney is everything on the backend good now? Sorry I haven't been following what's needed on the backend changes.

JPinkney:
Yeah, everything should be good on the backend and frontend now

christianvogt:
I tried again on cluster bot and everything worked for the admin and normal user.

christianvogt:
@spadgett @rhamilto this PR needs your approval

@JPinkney please rebase

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -27,12 +28,25 @@ type Props = {
...
+const isAdmin = (user: UserKind): boolean => {
+  if (!user.groups) {
+    return false;
+  }
+  for (const group of user.groups) {
+    if (group === 'system:cluster-admins') {
+      return true;
+    }
+  }
+  return false;
```
spadgett, line 40:
```suggestion
  return user.groups.some((group) => group === 'system:cluster-admins');
```

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -27,12 +28,25 @@ type Props = {
 
 type CloudShellTerminalProps = StateProps & Props;
 
+const isAdmin = (user: UserKind): boolean => {
+  if (!user.groups) {
+    return false;
+  }
+  for (const group of user.groups) {
+    if (group === 'system:cluster-admins') {
```
spadgett, line 36: I don't believe this is a reliable check for system admin. RBAC is more nuanced than this. You can have admin privileges without being in this group.

Is there a specific permission we are worried about? It would be better to do a self-subject access review.

frontend/packages/console-app/src/components/cloud-shell/setup/CloudShellAdminSetup.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const [initError, setInitError] = React.useState<string>();
+  React.useEffect(() => {
+    (async () => {
+      async function namespaceExists(): Promise<boolean> {
+        try {
+          await k8sGet(NamespaceModel, CLOUD_SHELL_PROTECTED_NAMESPACE);
+          return true;
+        } catch (error) {
+          return false;
```
spadgett, line 31: We should be checking the error code. If it's a 404 error, the namespace doesn't exist. If the error code is different, something else went wrong.

pkg/terminal/auth.go
```diff
@@ -1,33 +1,22 @@
...
-			},
-		},
+	if !isFound {
+		return false, nil
 	}
-	res, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), sar, metav1.CreateOptions{})
-	if err != nil || res == nil {
-		return false, err
+	for _, group := range groups {
+		if group == "system:cluster-admins" {
```
spadgett, line 17: Again, I don't think this is a reliable way to determine if the user is a cluster admin.

spadgett (changes requested):

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
sleshchenko: Previously, we tested `pods:write` in `openshift-operators` namespace with self-subject access review. @spadgett Do you have any concerns about returning back to that check? Any suggestion which permission is better to check?

spadgett: I think that would be a better check. We might want to use a different namespace like `openshift-terminal` or maybe `default`.

pkg/terminal/auth.go
openshifty: @spadgett is there a consistent way or better way to do this?

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
JPinkney: I just switched them to use `openshift-terminal`. I thought there might be issues if the openshift-terminal namespace doesn't exist but everything seems to be working fine

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
```diff
@@ -94,6 +96,26 @@ const CloudShellTerminal: React.FC<CloudShellTerminalProps &
...
+        if (resp.status.allowed) {
+          setAdmin(true);
+          return;
+        }
+        setAdmin(false);
+      })
+      .catch((e) => {
+        setInitError(e);
+      });
+  }, []);
```
christianvogt, line 117: You need to capture if the rbac check is still pending show the `TerminalLoadingBox`.

If catching the error is an edge case, we could simplify to use `useAccessReview2`. Otherwise add a separate loading state.
```
  const [isAdmin, isAdminCheckLoading] = useAccessReview2({
      namespace: 'openshift-terminal',
      verb: 'create',
      resource: 'pods',
    });
```

frontend/packages/console-app/src/components/cloud-shell/CloudShellTerminal.tsx
spadgett: > I just switched them to do a subject access review on the `openshift-terminal` namespace. I thought there might be issues if the openshift-terminal namespace doesn't exist but everything seems to be working fine

If you have permission to create pods in any namespace, then the RBAC check will pass even if the namespace doesn't exist yet.

christianvogt:
front end changes lgtm. Waiting for cluster bot to test again.

pkg/terminal/auth.go
```diff
@@ -7,10 +7,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// checkUserPermissions checks if the terminal proxy is supported for a given user.
-// Returns true if we're willing to proxy the user's token, false otherwise. We don't
-// want to proxy highly privileged tokens to avoid privilege escalation issues.
-func (p *Proxy) checkUserPermissions(token string) (bool, error) {
+// isClusterAdmin does a subject access review to see if the user can create pods in openshift-operators
```
spadgett, line 10: The namespace in the comment doesn't match the code.

spadgett (approved):
/approve

Backend changes LGTM. I will let @christianvogt give final lgtm. Thanks!

pkg/terminal/auth.go
JPinkney: Darn, that was the one spot that I forgot to change! Should I change it quickly or will it remove the LGTM status?

PR #7150, opened 2020-11-06
add quick starts card on cluster overview

debsmita1:
**Story**: 
https://issues.redhat.com/browse/ODC-4922

**Analysis / Root cause**: 
As an Admin, I want to easily be able to access and discover Quick Starts from the Cluster Overview.

**Solution Description**: 
- shows a list of max 3 quick starts, with View all option in the end
- the quick starts with status as `in progress` are listed first
- the completed ones are not listed
- if all the quick starts are completed then the catalog tile is removed from both the Cluster Overview and the Add page 
- the quick start badge is removed from the Cluster Overview page
- The visibility of the card is unique in each view. Removing the card will be saved in storage for that view, so it will not show up on that page again.
- added localization

**Gifs for design review**: 
quick-starts

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsTileLoader.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import QuickStartTile from '@console/dev-console/src/components/QuickStartTile';
+import QuickStartsLoader from './QuickStartsLoader';
+
+interface QuickStartsTileLoaderProps {
+  onQuickStartCatalogTileRemove?: () => void;
+}
+
+const QuickStartsTileLoader: React.FC<QuickStartsTileLoaderProps> = (props) => (
```
rohitkrai03, line 9: I am not sure if we even need this small utility component.

frontend/packages/dev-console/src/components/QuickStartTile.tsx
```diff
@@ -16,32 +17,62 @@ import {
...
+const QuickStartTile: React.FC<Props> = ({
+  setActiveQuickStart,
+  quickStarts,
+  allQuickStartStates,
+  perspective,
+  onRemoveTile,
+}) => {
+  const { t } = useTranslation();
+  const storageKey =
+    perspective === 'dev'
```
rohitkrai03, line 65: You don't need to check for perspective like this. Better to just send the storage_key prop from the parent component. That way this tile can be re-used in any number of places in any perspective.

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
```diff
@@ -16,11 +17,39 @@ const mainCards = [{ Card: StatusCard }, { Card: UtilizationCard }];
...
+    !isQuickStartTileHidden,
+  );
+
+  const onQuickStartCatalogTileRemove = () => {
+    setShowQuickStartCatalogTile(false);
+  };
+
+  const rc = React.useMemo(
+    () =>
+      showQuickStartCatalogTile
```
rohitkrai03, line 39: Why do you need this? Shouldn't the tile component not rendering anything rearrange the grid itself?

rohitkrai03 (changes requested):

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
debsmita1: @rohitkrai03  yes, it does but it adds a bit of space on top when I remove the quick-start tile

Screenshot from 2020-11-10 15-59-22

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsTileLoader.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import QuickStartTile from '@console/dev-console/src/components/QuickStartTile';
```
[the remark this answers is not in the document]
debsmita1: removed this component

frontend/packages/console-app/src/components/quick-starts/loader/QuickStartsTileLoader.tsx
debsmita1: removed this component

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+    onRemoveTile && onRemoveTile();
+  };
+
+  const onToggle = () => setOpen(!isOpen);
+
+  const actionDropdownItem = [
+    <DropdownItem
+      onClick={() => {
+        onRemove();
+      }}
```
rohitkrai03, line 80:
```suggestion
      onClick={onRemove}
```

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+        slicedQuickStarts.length < 3 &&
+        getQuickStartStatus(allQuickStartStates, quickStart.metadata.name) ===
+          QuickStartStatus.NOT_STARTED
+      ) {
+        slicedQuickStarts.push(quickStart);
+      }
+    });
+  }
+
+  if (quickStarts.length > 0 && slicedQuickStarts.length === 0) {
```
rohitkrai03, line 115: Why are you setting local storage to hide the card if there are no quick starts to show? That would mean the card would always be hidden even when a quick start is added later on.

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+    localStorage.setItem(HIDE_QUICK_START_DASHBOARD_TILE_STORAGE_KEY, 'true');
+    localStorage.setItem(HIDE_QUICK_START_ADD_TILE_STORAGE_KEY, 'true');
```
rohitkrai03, line 117: This component shouldn't have any idea about there storage keys.

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
rohitkrai03: Better just render null when there's nothing to show.

rohitkrai03 (changes requested):

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
rohitkrai03: @debsmita1  Can you take a look at why it's rendering empty grid item there?

rohitkrai03 (approved):
/lgtm
/approve

Need to remove the unused quick starts badge files since we don't use that anymore. Butthat can be done in a follow up PR. Rest works fine.

rhamilto:
When there are no QuickStarts, there is a bug where the wrapping `div.pf-l-grid__item.pf-m-12-col` for the QuickStarts card is still output, resulting in a gap above the `Activity` card.

Screen Shot 2020-11-24 at 2 51 32 PM

cc:  @spadgett

PR #7157, opened 2020-11-09
Fix to use graph view for default for dev perspective topology page

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5108

**Analysis / Root cause**: 
Default was mistakenly set to list view for dev perspective topology page

**Solution Description**: 
Default set to graph view with the option to default to list view. Update workloads tab to set the default to list view.
Update code to use the enum for view type rather than passing `graph` or `list` string values.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


jeff-phillips-18:
@christianvogt rebased

PR #7158, opened 2020-11-09
Bug 1896296: Ensure correct git URL for topology edit icon

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5040

**Analysis / Root cause**: 
Git URLs with `.git` suffix do not work when branch info is appended to the URL.

**Solution Description**: 
Check and remove `.git` in the URL before appending branch info.
Check if url is an SSH url (ex. `git@github.com:sclorg/nodejs-ex.git`) transform it into HTTP URL.

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/topology/topology-utils.ts
+++ b/frontend/packages/dev-console/src/components/topology/topology-utils.ts
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import GitUrlParse from 'git-url-parse';
 import {
   K8sResourceKind,
   K8sResourceKindReference,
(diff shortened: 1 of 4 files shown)


christianvogt:
@rottencandy I think you should use https://github.com/IonicaBizau/git-url-parse to parse the URL and construct the string. It has API to create URLs: `toString("https")`

frontend/packages/dev-console/src/components/topology/topology-utils.ts
```diff
@@ -32,9 +33,11 @@ export const getServiceBindingStatus = ({ FLAGS }: RootState): boolean =>
 export const getCheURL = (consoleLinks: K8sResourceKind[]) =>
   _.get(_.find(consoleLinks, ['metadata.name', 'che']), 'spec.href', '');
 
-export const getEditURL = (gitURL: string, gitBranch: string, cheURL: string) => {
+export const getEditURL = (vcsURI?: string, gitBranch?: string, cheURL?: string) => {
```
jeff-phillips-18, line 36: I don't see a reason to make the `vicURI` parameter optional.
```suggestion
export const getEditURL = (vcsURI: string, gitBranch?: string, cheURL?: string) => {
```

rottencandy: Updated.

rohitkrai03 (approved):
/approve

PR #7159, opened 2020-11-09
Topology, fetch service and route data when necessary

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5113

**Solution Description**: 
Remove routes and services fields from OverviewItem type. Add hooks to retrieve routes and services for a given resource. Update instances to use hooks rather than fields in the OverviewItem type.

Non-Functional Change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useRoutesWatcher.ts
+++ b/frontend/packages/console-shared/src/hooks/useRoutesWatcher.ts
@@ -1,0 +1,36 @@
+import * as React from 'react';
+import { K8sResourceKind, RouteKind } from '@console/internal/module/k8s';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import { getRoutesForServices } from '../utils';
+import { useServicesWatcher } from './useServicesWatcher';
+
+export const useRoutesWatcher = (
+  resource: K8sResourceKind,
+): { loaded: boolean; loadError: string; routes: RouteKind[] } => {
+  const { namespace } = resource.metadata;
+  const watchedServices = useServicesWatcher(resource);
+  const [allRoutes, loaded, loadError] = useK8sWatchResource<RouteKind[]>({
+    isList: true,
+    kind: 'Route',
+    namespace,
+  });
+
+  const servicesNames = React.useMemo(
+    () =>
+      !watchedServices.loadError && watchedServices.loaded
+        ? watchedServices.services.map((s) => s.metadata.name)
+        : [],
+    [watchedServices.loadError, watchedServices.loaded, watchedServices.services],
+  );
+
+  const routes = React.useMemo(() => getRoutesForServices(servicesNames, allRoutes), [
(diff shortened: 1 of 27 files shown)


frontend/packages/console-shared/src/hooks/useRoutesWatcher.ts
```diff
@@ -0,0 +1,52 @@
...
+        namespace,
+      },
+      routes: {
+        isList: true,
+        kind: 'Route',
+        namespace,
+      },
+    }),
+    [namespace],
+  );
```
jerolimov, line 27: Could not find a usage of `useRoutesWatcher` which also uses the fetched services. Can we remove the services resource lookup here?
And can we use then `useK8sWatchResource` instead of `useK8sWatchResources` then?

frontend/packages/console-shared/src/hooks/useServicesWatcher.ts
```diff
@@ -0,0 +1,44 @@
...
+      services: {
+        isList: true,
+        kind: 'Service',
+        namespace,
+      },
+    }),
+    [namespace],
+  );
+
+  const resources = useK8sWatchResources(watchedResources);
```
jerolimov, line 24: Can we simplify the error code below when using `useK8sWatchResource` instead of `useK8sWatchResources` here?

```suggestion
  const result = useK8sWatchResource(watchedResources);
  return {
    services: result[0],
    loaded: result[1],
    loadError, result[2],
  }
```

jerolimov (changes requested):
Added question to remove services from useRoutesWatcher and if we can use `useK8sWatchResource` instead of `useK8sWatchResources`. This could be improve all new watchers with only one resource.

frontend/packages/console-shared/src/hooks/useRoutesWatcher.ts
jeff-phillips-18: The call to `getServicesForResource` requires the services to be loaded, the returned services are used to determine the routes to be returned.

jerolimov: Didn't see that, thanks for the explanation :heavy_check_mark:

jerolimov (approved):
If you want change from `useK8sWatchResources` to `useK8sWatchResource` this is fine for me and I would retest it then again. Everything else looks already good.

/lgtm

jeff-phillips-18:
@jerolimov Updated and simplified. Thanks!

jerolimov (approved):
Tested it locally, shows services and routes :)

/lgtm

rohitkrai03 (approved):
/approve

andrewballantyne:
/lgtm

Rebase made, adding back the lgtm label.

PR #7162, opened 2020-11-09
Topology, fetch build config data when necessary

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5114

**Solution Description**: 
Remove build configs from OverviewItem type. Add hook to retrieve build configs for a given resource. Update instances to use hooks rather than fields in the OverviewItem type.

The alerts field for OverviewItem type is also removed as it is not being used anywhere (but required loading buildConfig resources).

Non-Functional Change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useBuildConfigsWatcher.ts
+++ b/frontend/packages/console-shared/src/hooks/useBuildConfigsWatcher.ts
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
+import { getBuildConfigsForResource } from '../utils';
+import { BuildConfigOverviewItem } from '../types';
+
+export const useBuildConfigsWatcher = (
+  resource: K8sResourceKind,
+): { loaded: boolean; loadError: string; buildConfigs: BuildConfigOverviewItem[] } => {
+  const [loaded, setLoaded] = React.useState<boolean>(false);
+  const [loadError, setLoadError] = React.useState<string>(null);
+  const [buildConfigs, setBuildConfigs] = React.useState<BuildConfigOverviewItem[]>();
+  const { namespace } = resource.metadata;
+  const watchedResources = React.useMemo(
+    () => ({
+      buildConfigs: {
+        isList: true,
+        kind: 'BuildConfig',
+        namespace,
+      },
+      builds: {
+        isList: true,
+        kind: 'Build',
+        namespace,
+      },
+    }),
+    [namespace],
+  );
+
+  const resources = useK8sWatchResources(watchedResources);
+
+  React.useEffect(() => {
+    const resourceWithLoadError = Object.values(resources).find((r) => r.loadError);
+    if (resourceWithLoadError) {
+      setLoadError(resourceWithLoadError.loadError);
+      return;
+    }
+    setLoadError(null);
+    if (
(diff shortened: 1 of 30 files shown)


frontend/packages/console-shared/src/hooks/useBuildConfigsWatcher.ts
```diff
@@ -0,0 +1,53 @@
...
+  );
+
+  const resources = useK8sWatchResources(watchedResources);
+
+  React.useEffect(() => {
+    const errorKey = Object.keys(resources).find((key) => resources[key].loadError);
+    if (errorKey) {
+      setLoadError(resources[errorKey].loadError);
+      return;
+    }
```
jerolimov, line 37: Small detail, but can simplified with Object.values:

```suggestion
    const resourceWithLoadError = Object.values(resources).find((r) => r.loadError);
    if (resourceWithLoadError) {
      setLoadError(resourceWithLoadError.loadError);
      return;
    }
```

For me both is fine, doesn't block this PR for me.

frontend/packages/console-shared/src/hooks/useBuildConfigsWatcher.ts
```diff
@@ -0,0 +1,53 @@
...
+import { CronJobKind, K8sResourceKind } from '@console/internal/module/k8s';
+import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
+import { getBuildConfigsForCronJob, getBuildConfigsForResource } from '../utils';
+import { BuildConfigOverviewItem } from '../types';
+
+export const useBuildConfigsWatcher = (
+  resource: K8sResourceKind,
+): { loaded: boolean; loadError: string; buildConfigs: BuildConfigOverviewItem[] } => {
+  const [loaded, setLoaded] = React.useState<boolean>(false);
+  const [loadError, setLoadError] = React.useState<string>('');
```
jerolimov, line 11: Please default with null. The useEffect also sets setLoadError to null if there is new resources available without any error. In this case we can save us from one re-rendering

frontend/packages/dev-console/src/utils/application-utils.ts
```diff
@@ -240,14 +244,33 @@ const deleteWebhooks = (
...
+    builds: {
+      data: builds,
+      loaded: true,
+      loadError: null,
+    },
+  };
+  const resourceBuildConfigs =
+    resource.kind === 'CronJob'
+      ? getBuildConfigsForCronJob(resource as CronJobKind, resources)
+      : getBuildConfigsForResource(resource, resources);
```
jerolimov, line 271: Just an idea: We can move this check also into `getBuildConfigsForResource` ? So that it starts with

```tsx
export const getBuildConfigsForResource = (
  resource: K8sResourceKind,
  resources: any,
): BuildConfigOverviewItem[] => {
  if (resource.kind === 'CronJob') {
    return getBuildConfigsForCronJob(resource as CronJobKind, resources);
  }
  // ....
```

Makes this sense? Not required to get a lgtm :)

jerolimov (changes requested):
Finished review, small change requested and added two ideas.

Tested the topology graph and sidebar locally with some git deployment which works fine.

jerolimov (approved):
Tested locally, Topology shows BuildConfigs

/lgtm

rohitkrai03 (approved):
/approve

PR #7164, opened 2020-11-09
Topology, fetch jobs for cronjobs when necessary

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5115

**Solution Description**: 
Remove jobs from OverviewItem type. Add hook to retrieve jobs for a given cronjob resource. Update instances to use hooks rather than fields in the OverviewItem type.

Non-Functional Change

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useJobsForCronJobWatcher.ts
+++ b/frontend/packages/console-shared/src/hooks/useJobsForCronJobWatcher.ts
@@ -1,0 +1,43 @@
+import * as React from 'react';
+import { K8sResourceKind, JobKind } from '@console/internal/module/k8s';
+import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
+import { getJobsForCronJob } from '../utils';
+
+export const useJobsForCronJobWatcher = (
+  cronJob: K8sResourceKind,
+): { loaded: boolean; loadError: string; jobs: JobKind[] } => {
+  const { namespace, uid } = cronJob.metadata;
+  const [loaded, setLoaded] = React.useState<boolean>(false);
+  const [loadError, setLoadError] = React.useState<string>('');
+  const [jobs, setJobs] = React.useState<JobKind[]>([]);
+  const watchedResources = React.useMemo(
+    () => ({
+      jobs: {
+        isList: true,
+        kind: 'Job',
+        namespace,
+      },
+    }),
+    [namespace],
+  );
+  const resources = useK8sWatchResources(watchedResources);
+
+  React.useEffect(() => {
+    const errorKey = Object.keys(resources).find((key) => resources[key].loadError);
+    if (errorKey) {
+      setLoadError(resources[errorKey].loadError);
+      return;
+    }
+    setLoadError('');
+    if (
+      Object.keys(resources).length > 0 &&
+      Object.keys(resources).every((key) => resources[key].loaded)
+    ) {
+      const resourceJobs = getJobsForCronJob(uid, resources);
+      setJobs(resourceJobs);
+      setLoaded(true);
+    }
+  }, [uid, resources]);
+
+  return { loaded, loadError, jobs };
+};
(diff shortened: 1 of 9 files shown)


frontend/packages/console-shared/src/hooks/useJobsForCronJobWatcher.ts
```diff
@@ -0,0 +1,43 @@
...
+import { K8sResourceKind, JobKind } from '@console/internal/module/k8s';
+import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
+import { getJobsForCronJob } from '../utils';
+
+export const useJobsForCronJobWatcher = (
+  cronJob: K8sResourceKind,
+): { loaded: boolean; loadError: string; jobs: JobKind[] } => {
+  const { namespace, uid } = cronJob.metadata;
+  const [loaded, setLoaded] = React.useState<boolean>(false);
+  const [loadError, setLoadError] = React.useState<string>('');
```
jerolimov, line 11:
```suggestion
  const [loadError, setLoadError] = React.useState<string>(null);
```

jerolimov:
@jeff-phillips-18 Tested this with the CronJob example below which doesn't link any resource and got an error in [pods-overview.tsx#L254-L257](https://github.com/openshift/console/blob/master/frontend/public/components/overview/pods-overview.tsx#L254-L257) because `PodsOverviewMultiple` prop `podResources` is undefined.

```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: example
  namespace: christoph-test
spec:
  schedule: '@daily'
  jobTemplate:
    spec:
      template:
        spec:
          containers:
            - name: hello
              image: busybox
              args:
                - /bin/sh
                - '-c'
                - date; echo Hello from the Kubernetes cluster
          restartPolicy: OnFailure
```

So you need to return jobs state in `useJobsForCronJobWatcher` with default [] or ensure that `PodsOverviewMultiple` handles this case.

jeff-phillips-18:
Thanks @jerolimov, updated.

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -838,7 +838,7 @@ export const createCronJobItem = (
   utils?: ResourceUtil[],
 ): OverviewItem => {
   const buildConfigs = getBuildConfigsForCronJob(cronJob, resources);
-  const jobs = getJobsForCronJob(cronJob, resources);
+  const jobs = getJobsForCronJob(cronJob.metadata.uid, resources);
```
dtaylor113, line 841: Should this be `cronJobUid`?

dtaylor113 (approved):
/lgtm

frontend/public/components/overview/pods-overview.tsx
```diff
@@ -251,10 +251,12 @@ export const PodsOverviewMultiple: React.SFC<PodsOverviewMultipleProps> = ({
...
-      let updatedPods = podResources.reduce((acc, resource) => {
-        acc.push(...getPodsForResource(resource, resources));
-        return acc;
-      }, []);
+      let updatedPods = podResources?.length
+        ? podResources.reduce((acc, resource) => {
+            acc.push(...getPodsForResource(resource, resources));
+            return acc;
+          }, [])
+        : [];
```
jerolimov, line 259: I personally would go with this, but its okay :smirk: 

```suggestion
      let updatedPods = podResources?.reduce((acc, resource) => {
        acc.push(...getPodsForResource(resource, resources));
        return acc;
      }, []) ?? [];
```

jerolimov (approved):
Verified locally that it did not crash anymore, changed to setLoadError(""), has already lgtm, and I would give it as well :+1: 

/lgtm

rohitkrai03 (approved):
/approve

jeff-phillips-18:
Rebased

jeff-phillips-18:
rebased again

PR #7167, opened 2020-11-10
Bug 1888870: Remove resolveCompletionItem from yaml editor since it's not needed

JPinkney:
This PR removes the resolveCompletionItem from the yaml editor since it's not actually needed and is causing errors.

Associated issue: https://bugzilla.redhat.com/show_bug.cgi?id=1888870


spadgett (approved):
/lgtm
/ok-to-test

PR #7168, opened 2020-11-10
add i18n for kubevirt-plugin [details page]

yaacov:
i18n of details page, part i

Screenshot:
screenshot-localhost_9000-2020 11 10-10_04_39


yaacov:
@pcbailey please review

frontend/packages/kubevirt-plugin/src/utils/guest-agent-strings.ts
```diff
@@ -1,31 +1,29 @@
...
-    return NOT_AVAILABLE_MESSAGE;
+    return t('kubevirt-plugin~Not available');
   }
 
   if (numLoggedInUsers === 0) {
-    return NO_LOGGED_IN_USERS_MSG;
+    return t('kubevirt-plugin~No users logged in');
   }
 
   return `${numLoggedInUsers} ${pluralize(numLoggedInUsers, 'user')}`;
```
pcbailey, line 14: Shouldn't this be translated using `count`?

frontend/packages/kubevirt-plugin/src/components/vms/vm-details.tsx
```diff
@@ -93,11 +96,18 @@ export const VMDetails: React.FC<VMDetailsProps> = (props) => {
...
-    <Alert className="co-alert" variant="warning" title="Operating system mismatch" isInline>
-      The operating system defined for this virtual machine does not match what is being reported by
+    <Alert
+      className="co-alert"
+      variant="warning"
+      title={t('kubevirt-plugin~Operating system mismatch')}
+      isInline
+    >
+      {t(
+        `kubevirt-plugin~The operating system defined for this virtual machine does not match what is being reported by
```
pcbailey, line 106: Don't use backticks inside the `t()` function. The i18n parser can't handle them for some reason and it won't pick up the key.

pcbailey (changes requested):
Yay for the first i18n patch for the kubevirt-plugin! 

I just had a couple of small comments.

frontend/packages/kubevirt-plugin/src/components/vms/vm-details.tsx
yaacov: yea, it's for multi line text, the i18n don't like backticks because it will not parse them, but here is just text,
@glekner do you think backticks are legit here (given they will not be parsed inside the `t` method) ? do you have a nicer solution for multi line translatable text ?

yaacov: changing to "" + "" ...
if anyone have a nicer way please comment.

frontend/packages/kubevirt-plugin/src/utils/guest-agent-strings.ts
yaacov: Yeee, i18n is great, using `count` now

yaacov:
@pcbailey please re review

frontend/packages/kubevirt-plugin/src/components/vms/vm-details.tsx
pcbailey: The `Trans` component looks nice there. Let's hope that there aren't any unforeseen side effects that make us remove it. =)

yaacov:
@spadgett PTAL

yaacov:
@pcbailey  @rawagner please review

pcbailey:
We need to wait for guidance on how to handle multi-line strings. We're being told that we shouldn't be using the `Trans` component unless there are React nodes or HTML tags in the string.

yaacov:
@pcbailey @rawagner hi, fixed the multi-line string as suggested by Rebbeca using this example [1] in our chat discussion, please review

[1] https://github.com/openshift/console/blob/master/frontend/public/components/monitoring/receiver-forms/alert-manager-receiver-forms.tsx#L195

frontend/packages/kubevirt-plugin/src/components/vm-snapshots/vm-snapshot-details.tsx
```diff
@@ -58,7 +59,7 @@ const SnapshotDetails: React.FC<DetailsProps> = ({ obj, customData: { restores }
                 }
               >
                 {obj?.metadata?.annotations?.description || (
-                  <div className="text-muted">{NOT_AVAILABLE_MESSAGE}</div>
+                  <div className="text-muted">{t('kubevirt-plugin~Not available')}</div>
```
rawagner, line 62: would it be possible to extract the keys into separate file ?
For example - we would create `translation-keys.ts`
```js
export const NOT_AVAILABLE = 'kubevirt-plugin~Not available'
```
and then use it `t(NOT_AVAILABLE)` ?

yaacov: > use it t(NOT_AVAILABLE) ?

No we can't, AFAIU the translation guidelines is to have clear texts in the `t` functions. so we must convert all our text constants to texts as part of the i18n effort.

To enforce this rule, the `yarn i18n` script will not translate this strings. 

@pcbailey am I currect ?

pcbailey: I asked about doing this. We could do it, but only with a hack. In general, it's being discouraged, but not forbidden. Below is Sam's response when I asked about our constants files. I'm open to the conversation if we want to do it a different way. I don't particularly like the inline approach. 

> You’ll have to refactor that. If you use the global i18next instance there, the strings won’t update on language change until after a page refresh. You could change the constants to refer to a translation key and use it like `t(CUSTOM_FLAVOR_LABEL_KEY)`. If you do that, you’ll need a comment in the code like `// t('kubevirt-plugin~Custom')` so the i18next parser knows about the string.
> 
> If you’re only using them one or two places, I’d just inline the keys in the component. Or you could refactor the code so that commonly used strings are in a reusable component.

yaacov: I like the inline approach, 
a - it's more clear when reading the file what is the tring shown.
b - the strings are actually keys, for the constant strings stored in the i18n `.json` file, so keeping another constant file for strings will be actually a duplication.

@pcbailey @rawagner can we keep it inline ?

pcbailey: The primary benefit to using constants file(s) like we currently have, is that you can change the string in one place and it updates for everyone. The i18n file doesn't work like that. Since the string is the key, it will still need to be changed everywhere it's used. I'm ok with using the inline approach if @rawagner is, though.

But, I did think of another option that may work. What if we convert the constants to functions that take the `t()` function, like this:

`const NOT_AVAILABLE = (t: TFunction) => t('kubevirt-plugin~Not available');`

yaacov: I prefer the inline approach,
a - the dictionary file is in our code base (under `kubevirt-plugin`) and we can see every new string added.
so if we forget to change a string somware, we will imidiatly see it as a new string that appear.

for example I had two strings:
`"{{label}} Description"` and `"{{name}} Descriotion"`

I Immediately saw that I have two similar strings in the `.json` file and converged them to be one string.

b - the hopes that we need to jump through in order to have a common place for strings ( e.g. moving the `t` function around) and having two files with strings are "not nice" IMHO.

rawagner: looking at guidelines I think having file with keys should work

translation-keys.ts
```js
export const NOT_AVAILABLE = 'kubevirt-plugin~Not available'
```

foo-component.tsx
```js
<div className="text-muted">{t(NOT_AVAILABLE)}</div>
```



what does not work is if we want to convert our existing constants to use `t` directly like

```
export const NOT_AVAILABLE = t('kubevirt-plugin~Not available');
```
foo-component.tsx (does not require any change) - this wont update if we change language
```js
<div className="text-muted">{NOT_AVAILABLE}</div>
```

But I didnt try it, so I may be wrong :) 

I just want to make sure we consider extracting keys and using constants. I think there's less chance to have typos but on the other hand looking at the core console we wont be aligned with them as they just use strings (and not being aligned with core is IMO is a big disadvantage and because of this I actually prefer strings).

pcbailey: It sounds like everyone is ok with strings, so strings it is. =)

PR #7172, opened 2020-11-10
Bug 1895372: Tolerate missing `status` on OperatorGroups

spadgett:
Use optional chaining to avoid runtime errors during OperatorHub subscribe when an OperatorGroup doesn't have a `status` stanza.

/assign @rhamilto


rhamilto (approved):
/lgtm

Thank you, @spadgett!

PR #7173, opened 2020-11-10
Bug 1888530 - [RHOCS Usability Study]No clear statement what rules users violated when failed the naming requirements

cyril-ui-developer:
/assign @spadgett 

This PR exclude the fix for OBC Name input field.


spadgett (approved):
/lgtm

Let's fix these two forms, and we can update the BZ component for the OCS team to fix the ObjectBucketClaim form.

PR #7174, opened 2020-11-10
Bug 1896229: Hide empty cards

rebeccaalpert:
We currently don't have code to handle gauge panels. (Both of the Current Rate of Bytes Received and Current Rate of Bytes Transmitted cards call for them.) Sam suggested hiding these cards for now and creating a story to support gauge panels. (See https://issues.redhat.com/browse/CONSOLE-2473 for the story.)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1896229.

### DIFF
--- a/frontend/public/components/monitoring/dashboards/index.tsx
+++ b/frontend/public/components/monitoring/dashboards/index.tsx
@@ -356,7 +356,12 @@ const Card: React.FC<CardProps> = ({ panel }) => {
     );
   }
 
+  if (!['grafana-piechart-panel', 'graph', 'singlestat', 'table'].includes(panel.type)) {
+    return null;
+  }
+
   const panelClassModifier = getPanelClassModifier(panel);
+
   return (
     <div
       className={`monitoring-dashboards__panel monitoring-dashboards__panel--${panelClassModifier}`}


frontend/public/components/monitoring/dashboards/index.tsx
```diff
@@ -358,21 +358,23 @@ const Card: React.FC<CardProps> = ({ panel }) => {
 
   const panelClassModifier = getPanelClassModifier(panel);
   return (
-    <div
-      className={`monitoring-dashboards__panel monitoring-dashboards__panel--${panelClassModifier}`}
-    >
-      <DashboardCard
-        className="monitoring-dashboards__card"
-        gradient={panel.type === 'grafana-piechart-panel'}
+    panel.type !== 'gauge' && (
```
spadgett, line 361: We should generalize this so that we hide any panel type we don't recognize.

rebeccaalpert: All set!

frontend/public/components/monitoring/dashboards/index.tsx
```diff
@@ -358,21 +358,26 @@ const Card: React.FC<CardProps> = ({ panel }) => {
...
-    <div
-      className={`monitoring-dashboards__panel monitoring-dashboards__panel--${panelClassModifier}`}
-    >
-      <DashboardCard
-        className="monitoring-dashboards__card"
-        gradient={panel.type === 'grafana-piechart-panel'}
+    (panel.type === 'grafana-piechart-panel' ||
+      panel.type === 'graph' ||
+      panel.type === 'singlestat' ||
+      panel.type === 'table') && (
```
kyoto, line 364: I think using an early return would be a little more readable. Also, since `panel.type` is repeated 4 times, maybe use `Array.includes`?

```
if (!['grafana-piechart-panel', 'graph', 'singlestat', 'table'].includes(panel.type)) {
  return null;
}
```

rebeccaalpert: All set!

frontend/public/components/monitoring/dashboards/index.tsx
```diff
@@ -357,6 +357,9 @@ const Card: React.FC<CardProps> = ({ panel }) => {
   }
 
   const panelClassModifier = getPanelClassModifier(panel);
```
kyoto, line 359: Nit: Since `panelClassModifier` is not used in the early return case, better to move it after the early return.

rebeccaalpert: Nit, uh, combed? All set.

kyoto:
/lgtm

Thanks @rebeccaalpert!

PR #7176, opened 2020-11-10
Fix i18n key in ResourceQuotas

rebeccaalpert:
The previous key was difficult for translators to work with. I updated the key to match what we've done elsewhere in the application.


frontend/public/components/resource-quota.jsx
```diff
@@ -298,9 +298,9 @@ const Details = ({ obj: rq }) => {
               )}
             </p>
             <p>
-              {`${t(
-                'resource-quota~A cluster administrator can establish limits on both the amount you can Request and your Limits with a',
-              )} ${ResourceQuotaModel.label}${t('resource-quota~.')}`}
+              {t(
+                'resource-quota~A cluster administrator can establish limits on both the amount you can request and your Limits with a ResourceQuota.',
```
spadgett, line 302: I’d make limits lowercase here too since it’s not a k8s resource

rebeccaalpert: All set!

spadgett (approved):
/lgtm

PR #7183, opened 2020-11-11
Bug 1896732: disable os upload if no os avaliable

yaacov:
OCP 4.6 with CNV 2.4 - golden images is not supported in CNV templates but th option to attach a PVC to a template is available in the UI.

This PR removes the checkbox if golden images is not supported in the templates.

Screenshot:
before:
screenshot-localhost_9000-2020 11 11-17_47_37

after:
screenshot-localhost_9000-2020 11 11-17_46_27


yaacov:
@glekner @matthewcarleton please review

PR #7184, opened 2020-11-11
Bug 1895263: Allow creating storage cluster irrespective of LSO namespace

afreen23:
- removes the hard-coded value of `openshift-local-storage` for fetching the LSO subscription
- if a single instance of LSO is not found, shows a dropdown to select one


a2batic:
lgtm

PR #7185, opened 2020-11-11
Update i18n json files

rebeccaalpert:
I pulled down the latest and it looks like a couple of json files weren't checked in after `yarn i18n` was run. `yarn i18n` will remove trailing spaces and sometimes update string order. If we don't fix these, developers will see these files update every time they run `yarn i18n`.


spadgett:
> I pulled down the latest and it looks like a couple of json files weren't checked in after `yarn i18n` was run. `yarn i18n` will remove trailing spaces and sometimes update string order.

We have a test that runs `yarn i18n` and fails if there is a diff... Do we have a bug in the test?

spadgett (approved):
/lgtm

rebeccaalpert:
Totally possible there's a bug in the diff test. I'll take a look and put up a second PR if needed.

PR #7186, opened 2020-11-11
Bug 1766287:  - "hello-openshift" template application wrong binary format

cyril-ui-developer:
/assign @spadgett 

Changed the image to `registry.redhat.io/openshift4/ose-hello-openshift-rhel8` but I'm getting `ImagePullBackOff` error for pods

Screen Shot 2020-11-11 at 11 56 21 AM

### DIFF
--- a/frontend/public/models/yaml-templates.ts
+++ b/frontend/public/models/yaml-templates.ts
@@ -7,6 +7,10 @@ import { YAMLTemplate } from '@console/plugin-sdk';
 /**
  * Sample YAML manifests for some of the statically-defined Kubernetes models.
  */
+
+const sampleContainerImage = 'registry.redhat.io/openshift4/ose-hello-openshift-rhel8';
+const sampleContainerCmd = '[ "/bin/bash", "-c", "sleep infinity" ]';
+
 export const baseTemplates = ImmutableMap<GroupVersionKind, ImmutableMap<string, string>>()
   .setIn(
     ['DEFAULT', 'default'],
@@ -229,6 +233,17 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: example
+  annotations:
+    image.openshift.io/triggers: |-
+      [
+        {
+          "from": {
+            "kind": "ImageStreamTag",
+            "name": "openshift/hello-openshift:latest"
+          },
+          "fieldPath": "spec.template.spec.containers[0].image"
+        }
+      ]
 spec:
   selector:
     matchLabels:
@@ -241,9 +256,10 @@ spec:
     spec:
       containers:
       - name: hello-openshift
-        image: openshift/hello-openshift
+        image: ${sampleContainerImage}
         ports:
         - containerPort: 8080
+        command: ${sampleContainerCmd}
 `,
   )
   .setIn(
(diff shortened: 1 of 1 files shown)


frontend/public/models/yaml-templates.ts
```diff
@@ -229,6 +233,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: example
+  annotations:
+    image.openshift.io/triggers: ${sampleContainerImage}
```
spadgett, line 237: If we're using `image.openshift.io/triggers`, we should refer to the image stream. This should look like

```suggestion
    image.openshift.io/triggers: |-
      [
        {
          "from": {
            "kind": "ImageStreamTag",
            "name": "openshift/hello-openshift:latest"
          },
          "fieldPath": "spec.template.spec.containers[0].image"
        }
      ]
```
```

Comment (cyril-ui-developer) [reply frontend/public/models/yaml-templates.ts]: @spadgett Updated, thanks.

File: frontend/public/models/yaml-templates.ts [hunk 2]
```diff
@@ -403,14 +431,26 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: example
+  annotations:
+    image.openshift.io/triggers: |-
```
spadgett, line 435: I don't think this will work for an individual pod since `image` is immutable after the pod is created.

frontend/public/models/yaml-templates.ts
```diff
@@ -338,6 +354,17 @@ apiVersion: apps.openshift.io/v1
 kind: DeploymentConfig
 metadata:
   name: example
+  annotations:
+    image.openshift.io/triggers: |-
```
spadgett, line 358: DeploymentConfigs have first class API support for triggers and don't need to use the annotation.

```
$ oc explain deploymentconfig.spec.triggers
KIND:     DeploymentConfig
VERSION:  apps.openshift.io/v1

RESOURCE: triggers <[]Object>

DESCRIPTION:
     Triggers determine how updates to a DeploymentConfig result in new
     deployments. If no triggers are defined, a new deployment can only occur as
     a result of an explicit client update to the DeploymentConfig with a new
     LatestVersion. If null, defaults to having a config change trigger.

     DeploymentTriggerPolicy describes a policy for a single trigger that
     results in a new deployment.

FIELDS:
   imageChangeParams    <Object>
     ImageChangeParams represents the parameters for the ImageChange trigger.

   type <string>
     Type of the trigger
```

cyril-ui-developer: @spadgett Removed the trigger for Pod and DeploymentConfigs.
Do we need to revert the image to `openshift/hello-openshift` for Pod and DeploymentConfigs?

spadgett: @cyril-ui-developer You shouldn't remove the trigger for DeploymentConfigs. Instead update it to use the `spec.triggers` instead of the annotation.

I'd just use `registry.redhat.io/openshift4/ose-hello-openshift-rhel8` directly for pods.

cyril-ui-developer: @spadgett Updated, thanks! However the status is still `ImagePullBackOff` instead of `Running`. 

Screen Shot 2020-11-16 at 2 14 34 PM
Screen Shot 2020-11-16 at 2 14 52 PM

spadgett (approved):
/lgtm

PR #7187, opened 2020-11-11
Fix i18n file diff test

rebeccaalpert:
The diff test for i18n files was still using the old file location (`./locales`). I updated it to handle public and packages locales folders.


test-frontend.sh
```diff
@@ -21,7 +21,7 @@ if git status --short | grep 'yarn.lock' > /dev/null; then
 fi
 
 yarn i18n
-GIT_STATUS="$( git status --short --untracked-files -- ./locales )"
+GIT_STATUS=[$(git status --short --untracked-files -- public/locales) || $(git status --short --untracked-files -- packages/**/locales)]
```
spadgett, line 24: I believe this can be one command.

```suggestion
GIT_STATUS=[$(git status --short --untracked-files -- public/locales packages/**/locales)]
```

rebeccaalpert: All set!

spadgett (approved):
/lgtm

Thanks!

rebeccaalpert:
Frontend test is going to fail until https://github.com/openshift/console/pull/7185 merges.

spadgett:
Looks like it's working :)

/hold until #7185 merges to fix the JSON files

rebeccaalpert:
/retest

#7185 merged.

rebeccaalpert:
Rebasing.

rebeccaalpert:
Another JSON file came in; updated to see if frontend tests will now pass.

rebeccaalpert:
Another JSON file wasn't checked in.

frontend/public/i18next-parser.config.js
```diff
@@ -1,14 +1,22 @@
+const PARSER_SUPPORTED_LOCALES = {
+  en: 'English',
+};
+
+/* eslint-disable no-unused-vars, @typescript-eslint/no-unused-vars */
```
spadgett, line 5: Oh, we shouldn't be referencing variables in the parser config in the app. I think we need to just make the locales map a separate variable in the language selector modal.

rebeccaalpert:
One of the i18n files that hadn't been checked in was conflicting with a namespace in another package. I updated the namespace that was smaller to fix the e2e tests. Long term, we should have a tool to check for these. Can you review this @spadgett assuming tests now pass? Thanks.

frontend/public/i18next-parser.config.js
```diff
@@ -1,14 +1,22 @@
+const PARSER_SUPPORTED_LOCALES = {
```
spadgett, line 1: This file shouldn't be in the `public` folder if it's just the parser config. `public` is for code that is part of the app itself.

spadgett (changes requested):

frontend/public/i18next-parser.config.js
rebeccaalpert: @spadgett - We're also using it in webpack.config.ts - Joachim had it set up to process Chinese a certain way when it's available. Are you ok with maintaining it in two places?

frontend/public/i18next-parser.config.js
rebeccaalpert: I'll try to move it. Fingers crossed it builds.

rebeccaalpert:
@spadgett - I updated file locations and imports; still have an open query on the support locales object.

rebeccaalpert:
```
Failed Specs:
1) Modal Annotations : Delete Annotation
   StaleElementReferenceError: stale element reference: element is not attached to the page document
```

/retest

frontend/i18next-parser.config.js
```diff
@@ -55,7 +54,7 @@ module.exports = {
   lineEnding: 'auto',
   // Control the line ending. See options at https://github.com/ryanve/eol
 
-  locales: Object.keys(SUPPORTED_LOCALES),
+  locales: Object.keys(PARSER_SUPPORTED_LOCALES),
```
spadgett, line 57: If we only need the keys, no reason for this to be a map with display names.

frontend/public/components/modals/language-preferences-modal.tsx
```diff
@@ -10,13 +10,16 @@ import {
...
-import { FALLBACK_LOCALE, SUPPORTED_LOCALES } from '../../i18next-parser.config';
 
 const LanguagePreferencesModal = (props: LanguagePreferencesModalProps) => {
   const { i18n, t } = useTranslation();
-  const langOptions = Object.keys(SUPPORTED_LOCALES).map((lang) => ({
+  const supportedLocales = {
+    en: 'English',
+    ja: '日本語',
+  };
+  const langOptions = Object.keys(supportedLocales).map((lang) => ({
```
spadgett, line 20: Instead of using a map and transforming it, just initial the variable in the format we need.

frontend/webpack.config.ts
```diff
@@ -19,7 +19,11 @@ interface Configuration extends webpack.Configuration {
 const MonacoWebpackPlugin = require('monaco-editor-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
-const { SUPPORTED_LOCALES } = require('./public/i18next-parser.config');
+
+const SUPPORTED_LOCALES = {
+  en: 'English',
+  ja: '日本語',
+};
```
spadgett, line 26: Don't initial this as a map and format later. Better to just specify it inline in the format we need.

frontend/i18next-parser.config.js
rebeccaalpert: Good point, thanks!

frontend/public/components/modals/language-preferences-modal.tsx
rebeccaalpert: We're also using supportedLocales as-is to fill the dropdown on line 45 though.

frontend/webpack.config.ts
```diff
@@ -207,7 +206,7 @@ const config: Configuration = {
     new CopyWebpackPlugin([{ from: './packages/knative-plugin/locales', to: 'locales' }]),
     new CopyWebpackPlugin([{ from: './packages/container-security/locales', to: 'locales' }]),
     new MomentLocalesPlugin({
-      localesToKeep: Object.keys(SUPPORTED_LOCALES).map((key) => (key === 'zh' ? 'zh-cn' : key)),
+      localesToKeep: ['en', 'ja'].map((key) => (key === 'zh' ? 'zh-cn' : key)),
```
spadgett, line 209:
```suggestion
      localesToKeep: ['en', 'ja'],
```

And when we get Chinese, we can just add `zh-cn` to the array without needing `map`

spadgett (approved):
/lgtm

PR #7189, opened 2020-11-11
Add Japanese to language switcher

spadgett:
We now have some translations for Japanese. Add the option to the language switcher.

/assign @rhamilto @rebeccaalpert

### DIFF
--- a/frontend/public/i18next-parser.config.js
+++ b/frontend/public/i18next-parser.config.js
@@ -1,5 +1,6 @@
 const SUPPORTED_LOCALES = {
   en: 'English',
+  ja: '日本語',
 };
 
 const FALLBACK_LOCALE = 'en';


frontend/public/i18next-parser.config.js
```diff
@@ -1,5 +1,7 @@
+// TODO: Each language option should be displayed in its own native language instead of English.
 const SUPPORTED_LOCALES = {
   en: 'English',
+  ja: 'Japanese',
```
rhamilto, line 4: This should be `ja: '日本語'`?  That's what the Customer Portal does.
Screen Shot 2020-11-11 at 2 08 26 PM

rebeccaalpert: I agree with Robb! That's how it was originally in Joachim's PR as well.

spadgett: Yeah, that was the TODO. I just didn't know the right value to use. Thanks! Updated.

rhamilto (commented):
/lgtm

rebeccaalpert (approved):
Lgtm!

frontend/public/i18next-parser.config.js
rebeccaalpert: I know enough Japanese to confirm that value is correct. :)

PR #7193, opened 2020-11-12
Cypress fixes:  'Create Project' timing issue and temporarily disable a11y check for 'aria-hidden-focus'

dtaylor113:
This PR addresses 2 Cypress issues:

1. Timing issue launching "Create Project" modal.   Cypress would immediatly click on the "Create Project" button to launch modal, then Project List page would finish loading and dismiss the modal!   Cypress now waits for Project List page to load before launching Create modal.

2. Continue to get `'aria-hidden-focus' accesibility violations` when launching modals.  #6910 was the first attempt to fix the violations; this PR is a continuation in the right direction, but not the complete solution, so this PR **temporarily disables the a11y check for aria-hidden-focus** until we correctly implement [aria-hidden elements do not contain focusable elements](https://dequeuniversity.com/rules/axe/3.4/aria-hidden-focus?application=axeAPI).  
Follow up [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1897008).

### DIFF
--- a/frontend/packages/integration-tests-cypress/views/list-page.ts
+++ b/frontend/packages/integration-tests-cypress/views/list-page.ts
@@ -18,7 +18,7 @@ export const listPage = {
       });
   },
   clickCreateYAMLbutton: () => {
-    return cy.byTestID('yaml-create').click();
+    cy.byTestID('yaml-create').click({ force: true });
   },
   filter: {
     byName: (name: string) => {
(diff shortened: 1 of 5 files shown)


rhamilto:
LGTM but will wait for other to weigh in.

spadgett:
> Cypress would immediatly click on the "Create Project" button to launch modal, then Project List page would finish loading and dismiss the modal! Cypress now waits for Project List page to load before launching Create modal.

I don't disagree with waiting for the page to load, but the fact that the modal is dismissed is probably a bug in the console code. I wonder if that's something a user could ever see.

dtaylor113:
> I don't disagree with waiting for the page to load, but the fact that the modal is dismissed is probably a bug in the console code. I wonder if that's something a user could ever see.

I don't think so, one of the commands was also `return`'ing a `cy...` which is also know to cause timing issues.

frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -18,7 +18,7 @@ export const listPage = {
       });
   },
   clickCreateYAMLbutton: () => {
-    return cy.byTestID('yaml-create').click();
+    cy.byTestID('yaml-create').click({ force: true });
```
spadgett, line 21: Why is force needed?

spadgett (approved):
/lgtm

PR #7196, opened 2020-11-12
Get customized catalog categories in dev catalog from server flags

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5019

**Analysis / Root cause**: 
Catalog categories are hard coded before. Based on the [Customize Developer Catalog Categories enhancement](https://github.com/openshift/enhancements/pull/508) we need to load them and fallback to the default categories.

**Solution Description**: 
Created a `useCatalogCategories` hook and refactored `CatalogFileViewPage` to a function component to use this hook. Also extracted the default categories and use a naming in sync with @rohitkrai03 upcoming catalog extensibility changes, see PR #7233.

Because of some security concerns / best practices the frontend could not read the ConfigMap directly. Instead of that the hook reads the customized developer catalog categories from `SERVER_FLAGS`. To add the categories to the server flags requires some changes in the `bridge` code. The bridge code now automatically 'add' a new `developerCatalogCategories` flag with an JSON encoded string for all categories.

All SERVER_FLAGS are rendered automatically to the `index.html` and are available as global `window.SERVERFLAGS....`.

When an admin changes the `Console` "cluster" resource, a `ConfigMap` "cluster" was automatically created. It contains a `ConsoleConfig.yaml` entry which was then provided as file to a new console Pod. Each change to the Console resource triggers a new console instance.

Catalog looks like before, but has now customized categories:
image

Updated existing tests, no update for the code coverage here.

**Test setup, with a cluster-bot cluster**
1. Launch a new cluster with cluster-bot `launch openshift/console#7196`
2. WIP... ;)

**Test setup, with a local running bridge and a remote cluster:**
1. Checkout this branch
2. `./build-backend.sh`
3. Login to your cluster as always (`oc login...`)
4. Setup the local bridge environment variables `source ./contrib/oc-environment.sh`
5. **New: Define the Developer Catalog Categories**

**a.) Configure your remote cluster and fetch the cluster customization config for local testing:**

Run the script `source ./contrib/oc-get-customization.sh` similar to oc-env. This new script needs to be called with kube:admin or service account permissions and it requires the yq command. It automatically fetches the customization config and sets the environment variable `BRIDGE_CONFIG`.

You can switch the user after this call or just run `bin/bridge`

**b.) Alternative you can create your own `cluster-consoleconfig-customization.yaml` and insert for example:**

```yaml
apiVersion: console.openshift.io/v1
kind: ConsoleConfig
customization:
  branding: ocp
  documentationBaseURL: https://docs.openshift.com/container-platform/4.6/
```

Export the config filename in your shell (see 5a) or just add the config file to your `bin/bridge` call:
```bash
bin/bridge -config cluster-consoleconfig-customization.yaml
```

**c.) Set the developer catalog categories directly as command line argument:**

```bash
bin/bridge -developer-catalog-categories '[ { "id": "java", "label": "Java", "tags": [ "java" ] } ]'
```

Please note that categories which doesn't match any item are not displayed by the catalog. ;)

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/cmd/bridge/main.go
+++ b/cmd/bridge/main.go
@@ -14,7 +14,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/coreos/pkg/flagutil"
 	"github.com/openshift/console/pkg/auth"
 	"github.com/openshift/console/pkg/bridge"
 	"github.com/openshift/console/pkg/knative"
@@ -61,11 +60,13 @@ func main() {
 	klog.InitFlags(fs)
 	defer klog.Flush()
 
+	// Define commandline / env / config options
+	fs.String("config", "", "The YAML config file.")
+
 	fListen := fs.String("listen", "http://0.0.0.0:9000", "")
 
 	fBaseAddress := fs.String("base-address", "", "Format: <http | https>://domainOrIPAddress[:port]. Example: https://openshift.example.com.")
 	fBasePath := fs.String("base-path", "/", "")
-	fConfig := fs.String("config", "", "The YAML config file.")
 
 	// See https://github.com/openshift/service-serving-cert-signer
 	fServiceCAFile := fs.String("service-ca-file", "", "CA bundle for OpenShift services signed with the service signing certificates.")
@@ -121,20 +122,15 @@ func main() {
 
 	fLoadTestFactor := fs.Int("load-test-factor", 0, "DEV ONLY. The factor used to multiply k8s API list responses for load testing purposes.")
 
-	if err := fs.Parse(os.Args[1:]); err != nil {
+	fDevCatalogCategories := fs.String("developer-catalog-categories", "", "Allow catalog categories customization. (JSON as string)")
+
+	if err := serverconfig.Parse(fs, os.Args[1:], "BRIDGE"); err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
 	}
-
-	if err := flagutil.SetFlagsFromEnv(fs, "BRIDGE"); err != nil {
+	if err := serverconfig.Validate(fs); err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
-	}
-
-	if *fConfig != "" {
-		if err := serverconfig.SetFlagsFromConfig(fs, *fConfig); err != nil {
-			klog.Fatalf("Failed to load config: %v", err)
-		}
 	}
 
 	baseURL := &url.URL{}
@@ -232,6 +228,7 @@ func main() {
 		ThanosPublicURL:       thanosPublicURL,
 		LoadTestFactor:        *fLoadTestFactor,
 		InactivityTimeout:     *fInactivityTimeout,
+		DevCatalogCategories:  *fDevCatalogCategories,
 	}
 
 	// if !in-cluster (dev) we should not pass these values to the frontend
--- a/frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogCategories.tsx
+++ b/frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogCategories.tsx
@@ -2,11 +2,11 @@ import * as React from 'react';
 import * as _ from 'lodash';
 import * as cx from 'classnames';
 import { VerticalTabs, VerticalTabsTab } from '@patternfly/react-catalog-view-extension';
-import { CatalogCategories } from '../utils/types';
+import { CatalogCategory } from '../utils/types';
 import { hasActiveDescendant, isActiveTab } from '../utils/category-utils';
 
 type CatalogCategoriesProp = {
-  categories: CatalogCategories;
+  categories: CatalogCategory[];
   categorizedIds: Record<string, string[]>;
   selectedCategory: string;
   onSelectCategory: (category: string) => void;
@@ -20,12 +20,15 @@ const CatalogCategories: React.FC<CatalogCategoriesProp> = ({
 }) => {
   const activeTab = _.has(categories, selectedCategory);
 
-  const renderTabs = (category, selectedCategoryId) => {
+  const renderTabs = (
+    category: CatalogCategory & { numItems?: number },
+    selectedCategoryID: string,
+    toplevelCategory: boolean,
+  ) => {
     if (!categorizedIds[category.id]) return null;
 
     const { id, label, subcategories, numItems } = category;
     const active = id === selectedCategory;
-    const shown = _.has(categories, id);
 
     const tabClasses = cx('text-capitalize', { 'co-catalog-tab__empty': !numItems });
 
@@ -37,11 +40,13 @@ const CatalogCategories: React.FC<CatalogCategoriesProp> = ({
         className={tabClasses}
         onActivate={() => onSelectCategory(id)}
         hasActiveDescendant={hasActiveDescendant(selectedCategory, category)}
-        shown={shown}
+        shown={toplevelCategory}
       >
         {subcategories && (
-          <VerticalTabs restrictTabs activeTab={isActiveTab(selectedCategoryId, category)}>
-            {_.map(subcategories, (subcategory) => renderTabs(subcategory, selectedCategoryId))}
+          <VerticalTabs restrictTabs activeTab={isActiveTab(selectedCategoryID, category)}>
+            {_.map(subcategories, (subcategory) =>
+              renderTabs(subcategory, selectedCategoryID, false),
+            )}
           </VerticalTabs>
         )}
       </VerticalTabsTab>
@@ -50,7 +55,7 @@ const CatalogCategories: React.FC<CatalogCategoriesProp> = ({
 
   return (
     <VerticalTabs restrictTabs activeTab={activeTab}>
-      {_.map(categories, (category) => renderTabs(category, selectedCategory))}
+      {_.map(categories, (category) => renderTabs(category, selectedCategory, true))}
     </VerticalTabs>
   );
 };
--- a/frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
+++ b/frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
@@ -1,6 +1,36 @@
-import { defaultCatalogCategories } from '../utils/default-categories';
-import { CatalogCategories } from '../utils/types';
+import * as React from 'react';
 
-const useCatalogCategories = (): CatalogCategories => defaultCatalogCategories;
+import { defaultCatalogCategories } from '../utils/default-categories';
+import { CatalogCategory } from '../utils/types';
+
+const useCatalogCategories = (): CatalogCategory[] => {
+  const categories = React.useMemo<CatalogCategory[]>(() => {
+    try {
+      const categoriesString = window.SERVER_FLAGS.developerCatalogCategories;
+      if (!categoriesString) {
+        return defaultCatalogCategories;
+      }
+
+      const categoriesArray: CatalogCategory[] = JSON.parse(categoriesString);
+
+      if (!Array.isArray(categoriesArray)) {
+        // eslint-disable-next-line no-console
+        console.error(
+          `Unexpected server flag "developerCatalogCategories" format. Expected array, got ${typeof categoriesArray}:`,
+          categoriesArray,
+        );
+        return defaultCatalogCategories;
+      }
+
+      return categoriesArray;
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Could not parse server flag "developerCatalogCategories":', error);
+      return defaultCatalogCategories;
+    }
+  }, []);
+
+  return categories;
+};
 
 export default useCatalogCategories;
--- a/frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
+++ b/frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
@@ -1,22 +1,23 @@
 import * as _ from 'lodash';
 import { CatalogItem } from '@console/plugin-sdk';
-import { CatalogCategories, CatalogCategory, CatalogSubcategory } from './types';
+import { CatalogCategory, CatalogSubcategory } from './types';
+
+export const NO_GROUPING = 'none';
+export const ALL_CATEGORY = 'all';
+export const OTHER_CATEGORY = 'other';
 
 export const matchSubcategories = (
   category: CatalogCategory,
   item: CatalogItem,
 ): (CatalogCategory | CatalogSubcategory)[] => {
   if (!category.subcategories) {
-    if (!category.values) {
+    if (!category.tags) {
       return [];
     }
 
-    let values = item[category.field];
-    if (!Array.isArray(values)) {
-      values = [values];
-    }
-
-    const intersection = [category.values, values].reduce((a, b) => a.filter((c) => b.includes(c)));
+    const intersection = [category.tags, item.tags || []].reduce((a, b) =>
+      a.filter((c) => b.includes(c)),
+    );
     if (!_.isEmpty(intersection)) {
       return [category];
     }
@@ -24,15 +25,9 @@ export const matchSubcategories = (
     return [];
   }
 
-  const matchedSubcategories = [];
-  _.forOwn(category.subcategories, (subCategory) => {
-    let values = item[category.field];
-
-    if (!Array.isArray(values)) {
-      values = [values];
-    }
-
-    const valuesIntersection = [subCategory.values, values].reduce((a, b) =>
+  const matchedSubcategories: (CatalogCategory | CatalogSubcategory)[] = [];
+  _.each(category.subcategories, (subCategory) => {
+    const valuesIntersection = [subCategory.tags, item.tags || []].reduce((a, b) =>
       a.filter((c) => b.includes(c)),
     );
     if (!_.isEmpty(valuesIntersection)) {
@@ -63,7 +58,7 @@ export const addItem = (
  */
 export const categorize = (
   items: CatalogItem[],
-  categories: CatalogCategories,
+  categories: CatalogCategory[],
 ): Record<string, string[]> => {
   const categorizedIds = {};
 
@@ -71,9 +66,9 @@ export const categorize = (
   _.each(items, (item) => {
     let itemCategorized = false;
 
-    addItem(categorizedIds, item.uid, categories.all.id); // add each item to all category
+    addItem(categorizedIds, item.uid, ALL_CATEGORY); // add each item to all category
 
-    _.each(categories, (category) => {
+    _.each(categories, (category: CatalogCategory) => {
       const matchedSubcategories = matchSubcategories(category, item);
       _.each(matchedSubcategories, (subcategory) => {
         addItem(categorizedIds, item.uid, category.id);
@@ -83,7 +78,7 @@ export const categorize = (
     });
 
     if (!itemCategorized) {
-      addItem(categorizedIds, item.uid, categories.other.id); // add to Other category
+      addItem(categorizedIds, item.uid, OTHER_CATEGORY); // add to Other category
     }
   });
 
@@ -92,17 +87,17 @@ export const categorize = (
 
 export const findActiveCategory = (
   activeId: string,
-  categories: CatalogCategories,
+  categories: CatalogCategory[],
 ): CatalogCategory => {
   let activeCategory = null;
-  _.forOwn(categories, (category) => {
+  _.each(categories, (category) => {
     if (activeCategory) {
       return;
     }
 
     if (category.id === activeId) {
       activeCategory = category;
-    } else {
+    } else if (category.subcategories) {
       activeCategory = findActiveCategory(activeId, category.subcategories);
     }
   });
@@ -110,15 +105,15 @@ export const findActiveCategory = (
 };
 
 export const hasActiveDescendant = (activeId: string, category: CatalogCategory): boolean => {
-  if (_.has(category?.subcategories, activeId)) {
+  if (_.has(category.subcategories, activeId)) {
     return true;
   }
 
-  return _.some(category?.subcategories, (subcategory) =>
+  return _.some(category.subcategories, (subcategory) =>
     hasActiveDescendant(activeId, subcategory),
   );
 };
 
 export const isActiveTab = (activeId: string, category: CatalogCategory): boolean => {
-  return _.has(category?.subcategories, activeId);
+  return _.has(category.subcategories, activeId);
 };
--- a/pkg/serverconfig/config.go
+++ b/pkg/serverconfig/config.go
@@ -1,14 +1,51 @@
 package serverconfig
 
 import (
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"strconv"
 
+	"github.com/coreos/pkg/flagutil"
 	"gopkg.in/yaml.v2"
+	"k8s.io/klog"
 )
+
+// Parse configuration from
+// 1. Config file
+// 2. Environment variables (overrides config file)
+// 3. Commandline arguments (overrides config file and environment varibles)
+//
+// Because the config filename could be defined as commandline argument or
+// environment variable, we need to parse these inputs before reading the
+// config file and need to override the config values after this again.
+func Parse(fs *flag.FlagSet, args []string, envPrefix string) error {
+	if err := flagutil.SetFlagsFromEnv(fs, envPrefix); err != nil {
+		return err
+	}
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	configFile := fs.Lookup("config").Value.String()
+
+	if configFile != "" {
+		if err := SetFlagsFromConfig(fs, configFile); err != nil {
+			klog.Fatalf("Failed to load config: %v", err)
+			return err
+		}
+		if err := flagutil.SetFlagsFromEnv(fs, envPrefix); err != nil {
+			return err
+		}
+		if err := fs.Parse(args); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
 
 // SetFlagsFromConfig sets flag values based on a YAML config file.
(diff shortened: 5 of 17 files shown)


frontend/public/components/catalog/hooks/useCatalogCategories.ts
```diff
@@ -0,0 +1,85 @@
...
+    };
+  };
+}
+
+export const useCatalogCategories = () => {
+  const [configMap, loaded, loadError] = useK8sWatchResource<ConfigMapKind>({
+    kind: 'ConfigMap',
+    namespace: 'openshift-console',
+    name: 'console-config',
+  });
```
spadgett, line 35: The console frontend shouldn't be fetching this config map directly. Normal users won't have access. The right way to do this is to update the console-operator to watch the custom categories config map and roll out a new console revision when it changes, wiring through updated config to the console backend.

cc @jhadvig

jerolimov: Oh, okay. Just to verify if I get the right way now, which I just discussed with @christianvogt:

Instead of reading the `ConfigMap` via k8 watch, extract the `customization.developerCatalog.categories` as SERVER_FLAG "developerCatalog.categories" in [pkg/serverconfig/config.go](https://github.com/openshift/console/blob/43e041787cf3899cf6ec4c6ea5e279e37b64c1c6/pkg/serverconfig/config.go#L165-L181). Then it will be automatically available as `windows.SERVER_FLAG.developerCatalog.categories`

And then I just need to return the server flag || defaultCatalogCategories in the hook?

spadgett: Right, the backend sets the server flag with the categories. We might not need a hook at all since the data won't change after console load.

pkg/serverconfig/config.go
```diff
@@ -117,8 +118,13 @@ func addClusterInfo(fs *flag.FlagSet, clusterInfo *ClusterInfo) {
...
 	// Assume "openshift" if config file is used.
-	fs.Set("k8s-auth", "openshift")
-	fs.Set("user-auth", "openshift")
+
+	// FIXME!!!
+	fmt.Printf("k8s-auth: %s\n", fs.Lookup("k8s-auth").Value.String())
+	fmt.Printf("user-auth: %s\n", fs.Lookup("user-auth").Value.String())
+
+	// fs.Set("k8s-auth", "openshift")
+	// fs.Set("user-auth", "openshift")
```
jerolimov, line 127: @spadgett @jhadvig @christianvogt At the moment I have the problem, that when we use a local `ConsoleConfig`, this code here previously overrides the auth=disabled option which is set by `contrib/oc-environment.sh`.

Without commenting out this two lines I can fetch the `ConsoleConfig` with this command:

```bash
oc get -o 'template={{index .data "console-config.yaml"}}' -n openshift-console configmaps console-config
```

But the fetched `ConsoleConfig` has more "problems", because it tries set some TLS files which are only available on a cluster. As example here is what I get from my local CRC:

```yaml
apiVersion: console.openshift.io/v1
auth:
  clientID: console
  clientSecretFile: /var/oauth-config/clientSecret
  oauthEndpointCAFile: /var/default-ingress-cert/ca-bundle.crt
clusterInfo:
  consoleBaseAddress: https://console-openshift-console.apps-crc.testing
  masterPublicURL: https://api.crc.testing:6443
customization:
  branding: ocp
  documentationBaseURL: https://docs.openshift.com/container-platform/4.6/
kind: ConsoleConfig
monitoringInfo:
  alertmanagerPublicURL: https://alertmanager-main-openshift-monitoring.apps-crc.testing
  grafanaPublicURL: https://grafana-openshift-monitoring.apps-crc.testing
  prometheusPublicURL: https://prometheus-k8s-openshift-monitoring.apps-crc.testing
  thanosPublicURL: https://thanos-querier-openshift-monitoring.apps-crc.testing
providers: {}
servingInfo:
  bindAddress: https://[::]:8443
  certFile: /var/serving-cert/tls.crt
  keyFile: /var/serving-cert/tls.key
```

And here one from a shared cluster:

```yaml
apiVersion: console.openshift.io/v1
auth:
  clientID: console
  clientSecretFile: /var/oauth-config/clientSecret
  oauthEndpointCAFile: /var/default-ingress-cert/ca-bundle.crt
clusterInfo:
  consoleBaseAddress: https://console-openshift-console.apps.REMOVED.devcluster.openshift.com
  masterPublicURL: https://api.REMOVED.devcluster.openshift.com:6443
customization:
  branding: ocp
  documentationBaseURL: https://docs.openshift.com/container-platform/4.7/
kind: ConsoleConfig
monitoringInfo:
  alertmanagerPublicURL: https://alertmanager-main-openshift-monitoring.apps.REMOVED.devcluster.openshift.com
  grafanaPublicURL: https://grafana-openshift-monitoring.apps.REMOVED.devcluster.openshift.com
  prometheusPublicURL: https://prometheus-k8s-openshift-monitoring.apps.REMOVED.devcluster.openshift.com
  thanosPublicURL: https://thanos-querier-openshift-monitoring.apps.REMOVED.devcluster.openshift.com
providers: {}
servingInfo:
  bindAddress: https://[::]:8443
  certFile: /var/serving-cert/tls.crt
  keyFile: /var/serving-cert/tls.key
```

When I reduce the `ConsoleConfig` to this example:

```yaml
apiVersion: console.openshift.io/v1
kind: ConsoleConfig
customization:
  developerCatalog:
    categories:
    - id: java
      label: Java
      tags: [java, jvm]
```

I can start the bridge with this file:

```bash
export BRIDGE_CONFIG=consoleconfig-with-dev-catalog-categories.yaml
bin/bridge
```

jerolimov: @spadgett @jhadvig @christianvogt Two ideas to solve the two problem above:

**Issue a) Could not run disabled authentification locally with a ConsoleConfig file**

Currently the [config.go addAuth method](https://github.com/openshift/console/blob/e92878c521f90cc09de714ef4b61840a45ec963e/pkg/serverconfig/config.go#L118-L121) overrides the flags `k8s-auth` and `user-auth`. I expect just removing this is currently a risk. The current  code assume "if there is any config, we use openshift for auth". There is currently no way to override this.

The reason for this is, in [main.go line 124-138](https://github.com/openshift/console/blob/e92878c521f90cc09de714ef4b61840a45ec963e/cmd/bridge/main.go#L124-L138) the order of reading configuration flags is:

1. Program arguments
2. Environment variables
3. Config file

As user of the bridge I had expected a revere order. So that config parameters are overridden by environment variables, and both can be overridden by program arguments.

If we would change this order, we could provide the customization (and all other ConsoleConfig arguments) as config file. This could then still enable "openshift" authentification by default.

But the exported environment variables from `oc-environment.sh` can override this, so that the authentication is disabled for local development.

What do you think?

**Issue b) Cluster ConsoleConfig YAML could not be used locally**

I talked with Christian about the problem, that the `ConsoleConfig` contains many keys which doesn't work locally. Esp. the TLS configuration. His recommendation is that we fetch the yaml file and accept only "well known" keys.

This must not be part of the `oc-environment.sh` script. For example, a script `contrib/oc-customization.sh` could download the yaml and accept only customization data and `export BRIDGE_CONFIG`.

What your opinion here? Have you another recommendation?

frontend/packages/dev-console/src/components/catalog/utils/custom-catalog-categories.ts
```diff
@@ -0,0 +1,47 @@
...
+    out[subcategory.id] = mapSubcategory(subcategory);
+  });
+  return out;
+};
+
+const mapCategory = (category: DeveloperConsoleCatalogCategory): CatalogCategory => ({
+  id: category.id,
+  label: category.label,
+  field: 'tags',
+  values: category.tags,
```
christianvogt, line 37: We should align the catalog apis and the console config schema for catalog categories.
We should be able to drop `field` and change `values` to `tags`.

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
```diff
@@ -1,6 +1,30 @@
...
+  const categories = React.useMemo<CatalogCategories>(() => {
+    try {
+      const categoriesString = window.SERVER_FLAGS.developerCatalogCategories;
+      if (!categoriesString) {
+        return defaultCatalogCategories;
+      }
+
+      const categoriesArray = JSON.parse(categoriesString);
+
+      if (!categoriesArray || !categoriesArray.length) {
```
christianvogt, line 17: Seems valid to me if someone supplies an empty array or a null value that we hide all categories.

christianvogt (commented):
I tried this:
`bin/bridge -developer-catalog-categories '[]'`
and got the default categories.
I expected to be able to hide all categories.

christianvogt:
I tried your script `oc-get-customization.sh`
Got the following output:
```
▶ ./contrib/oc-get-customization.sh
Use customization:
  branding: ocp
  documentationBaseURL: https://docs.openshift.com/container-platform/4.7/
  developerCatalog:
    categories:
      - id: java
        label: Testing Java
        tags:
          - java
```

But when I ran bridge I still saw the default categories.

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
jerolimov: Removed the second condition, catalog already skipped "All" and "Others" in this case. Result looks like this:

image

jerolimov:
> ./contrib/oc-get-customization.sh

@christianvogt Unfortunately you need to run the script in context of your current shell, because it sets an environment variable, similar to `source ./contrib/oc-environment.sh`

Please try `source ./contrib/oc-get-customization.sh`

Or you need to add a bridge argument: `bin/bridge -config cluster-consoleconfig-customization.yaml`

I can add a usage output to make this clear?

frontend/packages/dev-console/src/components/catalog/utils/custom-catalog-categories.ts
jerolimov: @christianvogt @rohitkrai03 I changed the default categories now to the CRD format. (Arrays instead of objects and replaced field and values with tags. Changed all necessary frontend code to keep the arrays up to rendering.

jerolimov:
Rebased and applied recommended changes from Christian. Go code does not changed.

pkg/serverconfig/config.go
```diff
@@ -178,4 +184,21 @@ func addCustomization(fs *flag.FlagSet, customization *Customization) {
 	if customization.CustomLogoFile != "" {
 		fs.Set("custom-logo-file", customization.CustomLogoFile)
 	}
+
+	categories, err := json.Marshal(&customization.DeveloperCatalog.Categories)
+	if err == nil {
+		fs.Set("developer-catalog-categories", string(categories))
+	} else {
+		fmt.Printf("Could not marshal ConsoleConfig customization.developerCatalog.categories: %v", err)
```
spadgett, line 192: This should be a fatal error.

cmd/bridge/main.go
```diff
@@ -121,20 +122,36 @@ func main() {
...
+	fs.String("developer-catalog-categories", "", "Allow catalog categories customization. (JSON as string)")
+
+	// Read configuration from
+	// 1. Config file
+	// 2. Environment variables (overrides config file)
+	// 3. Commandline arguments (overrides config file and environment varibles)
+	//
+	// Because the config filename could be defined as commandline argument or
+	// environment variable, we need to parse these inputs before reading the
+	// config file and need to override the config values after this again.
```
spadgett, line 134: We should add some tests to make sure we're really honoring this and to avoid regressions.

cmd/bridge/main.go
```diff
@@ -232,6 +249,7 @@ func main() {
 		ThanosPublicURL:       thanosPublicURL,
 		LoadTestFactor:        *fLoadTestFactor,
 		InactivityTimeout:     *fInactivityTimeout,
+		DevCatalogCategories:  fs.Lookup("developer-catalog-categories").Value.String(),
```
spadgett, line 252: We need to be doing some validation here. We should fail if the JSON is not well-formed or the content doesn't match what we expect.

contrib/oc-get-customization.sh
```diff
@@ -0,0 +1,44 @@
...
+
+LOCAL_FILE="cluster-consoleconfig-customization.yaml"
+
+if ! which yq > /dev/null; then
+    echo "yq cli not found, please checkout https://github.com/mikefarah/yq#install"
+    exit 1
+fi
+
+consoleConfig=$(
+    oc get -n openshift-console configmaps console-config -o 'template={{index .data "console-config.yaml"}}'
```
spadgett, line 31: I'd think we'd want to read the console config custom resource directly instead of using the config map. If you read the custom resource and you  it as JSON, you can use `jq` instead of `yq` which is more likely to be installed and already a prereq.

contrib/oc-get-customization.sh
```diff
@@ -0,0 +1,44 @@
+# shellcheck shell=bash
+#
+# This file is an example of how you might set up your environment to run the
+# console against an OpenShift cluster during development. To use it for
+# running bridge, do
+#
+# . contrib/oc-environment.sh
+# . contrib/oc-get-customization.sh
```
spadgett, line 8: It would be better to make this one script instead of two (even if this script calls the other).

contrib/oc-get-customization.sh
```diff
@@ -0,0 +1,44 @@
+# shellcheck shell=bash
```
spadgett, line 1: We should use `set -euo pipefail`

frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogCategories.tsx
```diff
@@ -20,12 +20,15 @@ const CatalogCategories: React.FC<CatalogCategoriesProp> = ({
 }) => {
   const activeTab = _.has(categories, selectedCategory);
 
-  const renderTabs = (category, selectedCategoryId) => {
+  const renderTabs = (
+    category: CatalogCategory & { numItems?: number },
+    selectedCategoryId: string,
```
spadgett, line 25: With console naming conventions, this would be

```suggestion
    selectedCategoryID: string,
```

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
```diff
@@ -1,6 +1,29 @@
...
+  const categories = React.useMemo<CatalogCategory[]>(() => {
+    try {
+      const categoriesString = window.SERVER_FLAGS.developerCatalogCategories;
+      if (!categoriesString) {
+        return defaultCatalogCategories;
+      }
+
+      const categoriesArray: CatalogCategory[] = JSON.parse(categoriesString);
+
+      if (!Array.isArray(categoriesArray)) {
```
spadgett, line 16: This needs to be logged and probably surfaced in the UI somewhere.

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
```diff
@@ -1,6 +1,29 @@
...
+        return defaultCatalogCategories;
+      }
+
+      return categoriesArray;
+    } catch (error) {
+      return defaultCatalogCategories;
```
spadgett, line 22: Again we shouldn't swallow the error. That makes it impossible to debug problems. At the very least, this needs to be logged in the console.

frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
```diff
@@ -1,38 +1,37 @@
...
-    if (!category.values) {
+    if (!category.tags) {
       return [];
     }
 
-    let values = item[category.field];
-    if (!Array.isArray(values)) {
-      values = [values];
+    let { tags } = item;
+    if (!Array.isArray(tags)) {
```
spadgett, line 15: When would this not be an array?

frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
```diff
@@ -1,38 +1,37 @@
...
-  const matchedSubcategories = [];
-  _.forOwn(category.subcategories, (subCategory) => {
-    let values = item[category.field];
-
-    if (!Array.isArray(values)) {
-      values = [values];
+  const matchedSubcategories: (CatalogCategory | CatalogSubcategory)[] = [];
+  _.each(category.subcategories, (subCategory) => {
+    let { tags } = item;
+    if (!Array.isArray(tags)) {
```
spadgett, line 30: When is this not an array?

pkg/serverconfig/types.go
```diff
@@ -65,6 +65,36 @@ type Customization struct {
...
+	Categories []DeveloperConsoleCatalogCategory `json:"categories"`
+}
+
+type DeveloperConsoleCatalogCategory struct {
+	// identifier used in the URL to enable deeplinking in console
+	ID string `json:"id"`
+	// category display label
+	Label string `json:"label"`
+	// the list of item tags this category will match
+	// +optional
```
spadgett, line 84: I don't think these comments have any meaning in the console repository, only in openshift/api.

frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogCategories.tsx
jerolimov: I just added the typing here, I can add a second commit to just rename this.

pkg/serverconfig/config.go
jerolimov: Changed to klog.Fatalf instead of fmt.Printf

contrib/oc-get-customization.sh
jerolimov: Yeah thats also cooler because you will get the data right after changing the console config and don't need to wait for a new deployment. Expected you want use the ConfigMap, will change this.

frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
jerolimov: Keeped the old code, cleaned it up and expect that it's now always an array (which we know for defaults and the CRD)

frontend/packages/dev-console/src/components/catalog/utils/category-utils.ts
jerolimov: Keeped the old code, cleaned it up and expect that it's now always an array (which we know for defaults and the CRD)

cmd/bridge/main.go
jerolimov: Extract flag set parsing into serverconfig package, please take a look at main.go, config.go and config_test.go :)

cmd/bridge/main.go
jerolimov: Added validation for the dev catalog categories, see new `pkg/serverconfig/validate.go` and `pkg/serverconfig/validate_test.go`. It fails when required fields "id" or "label" is missing. When using the cli argument is validates also that the input could parsed as json, is an array, doesn't contain invalid keys.

When using a yaml file, it was parsed by `config.yaml`. In this case it only validates that there is an id and label prop.

This works:

```yaml
apiVersion: console.openshift.io/v1
kind: ConsoleConfig
customization:
  developerCatalog:
    categories: []
```

This crashs:

```yaml
apiVersion: console.openshift.io/v1
kind: ConsoleConfig
servingInfo:
  bindAddress: http://localhost:9000
customization:
  developerCatalog:
    categories:
      -
```

Output:

```bash
go run github.com/openshift/console/cmd/bridge -config bind-address-config.yaml
Developer catalog category at index 0 must have at least id and label properties.
exit status 1
```

contrib/oc-get-customization.sh
jerolimov: @spadgett I fetch customization now via `oc get consoles.operator.openshift.io cluster -o yaml` 

I stay with yq for now, I didn't find a clean way to convert the extracted data with jq to yaml. And we need yaml because the bridge reads this ConsoleConfig as yaml. Wdyt?

contrib/oc-get-customization.sh
jerolimov: The problem with this is, oc-get-customization works only as kube:admin, while oc-environment works also with other users. And its only required to test customization locally, so I don't want to force anyone to use this but want provide a way to use it optionally.

We can improve this when we how we run the bridge with different users. Aka, after FF? Wdyt?

contrib/oc-get-customization.sh
jerolimov: As long as this script should be invoked via `source contrib/oc-get-customization.sh` to export `BRIDGE_CONFIG`, this doesn't work, because each error while close your current shell.

If we want that the user can just run `./contrib/oc-get-customization.sh` (or maybe `./hack/get-customization.sh`) and configure the config **manually** (`bin/bridge -config ...`), we can active this options. I would prefer to keep this for now and that we meet with Christian and Jakub to talk about some cleanups?

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
jerolimov: Add console.error logging

frontend/packages/dev-console/src/components/catalog/hooks/useCatalogCategories.ts
jerolimov: Add console.error logging

jerolimov:
@jhadvig @spadgett Updated this PR. See comments above. I added some go tests, some error logging in frontend and so on.

rohitkrai03 (changes requested):
@jerolimov UI crashes when I go to the Catalog page. 
Screenshot from 2020-12-03 19-09-30

Also getting some i18n keys missing error in console.

jerolimov:
> @jerolimov UI crashes when I go to the Catalog page.

Fixed. This was the reason there was a Array.isArray(item.tags) checks before. I need to add `|| []`-fallback here again.

> Also getting some i18n keys missing error in console.

Don't want fix this as part of this PR. We should go over all components at the end to find missing i18n keys here.

rohitkrai03 (approved):
/lgtm

The UI changes looks good to me. Verified locally.

cmd/bridge/main.go
```diff
@@ -121,22 +122,17 @@ func main() {
 
 	fLoadTestFactor := fs.Int("load-test-factor", 0, "DEV ONLY. The factor used to multiply k8s API list responses for load testing purposes.")
 
-	if err := fs.Parse(os.Args[1:]); err != nil {
+	fs.String("developer-catalog-categories", "", "Allow catalog categories customization. (JSON as string)")
```
jhadvig, line 125: Why not
```suggestion
	fDevCatalogCategories := fs.String("developer-catalog-categories", "", "Allow catalog categories customization. (JSON as string)")
```

and the below just:
```go
DevCatalogCategories:  fs.fDevCatalogCategories,
```

contrib/oc-get-customization.sh
jhadvig: Since this is only for local testing of the custom Dev catalog categories it could be a separate script. Would be good to update Readme with it.
@spadgett what do you think ?

jhadvig (commented):
The backend changes looks good.
Big 👍  for the tests.
Added couple of comments. Would deffer to @spadgett regarding the `oc-get-customization.sh` script.

pkg/serverconfig/types.go
```diff
@@ -65,6 +65,33 @@ type Customization struct {
...
+	DeveloperCatalog DeveloperConsoleCatalogCustomization `yaml:"developerCatalog,omitempty"`
+}
+
+// developerCatalog allow the cluster admin to configure developer catalog.
+type DeveloperConsoleCatalogCustomization struct {
+	// categories which are shown the in developer catalog.
+	Categories []DeveloperConsoleCatalogCategory `json:"categories,omitempty" yaml:"categories,omitempty"`
+}
+
+// category for the developer console catalog.
```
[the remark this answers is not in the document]
jerolimov: Updated to exact this commiets (without kubebuilder part)

cmd/bridge/main.go
jerolimov: Good catch, changed and tested locally with, and without config file.

contrib/oc-get-customization.sh
jhadvig: As @spadgett mentioned in the Slack discussion, lets leave this script out of the scope of this PR.

pkg/serverconfig/config.go
```diff
@@ -178,4 +220,21 @@ func addCustomization(fs *flag.FlagSet, customization *Customization) {
 	if customization.CustomLogoFile != "" {
 		fs.Set("custom-logo-file", customization.CustomLogoFile)
 	}
+
+	categories, err := json.Marshal(&customization.DeveloperCatalog.Categories)
```
jhadvig, line 224: Since this is a pointer we should add check
```suggestion
	if &customization.DeveloperCatalog != nil && &customization.DeveloperCatalog.Categories != nil {
		categories, err := json.Marshal(&customization.DeveloperCatalog.Categories)
```

pkg/serverconfig/config.go
```diff
@@ -178,4 +220,21 @@ func addCustomization(fs *flag.FlagSet, customization *Customization) {
...
+	if err == nil {
+		fs.Set("developer-catalog-categories", string(categories))
+	} else {
+		klog.Fatalf("Could not marshal ConsoleConfig customization.developerCatalog.categories: %v", err)
```
jhadvig, line 228:
```suggestion
		klog.Fatalf("Could not marshal ConsoleConfig customization.developerCatalog.categories field: %v", err)
```

pkg/serverconfig/config.go
jerolimov: Fixed, first check doesn't work. Second added as discussed in slack

jhadvig (approved):
The backend changes look good 👍 
/lgtm

andrewballantyne:
/retest 

Memory error again, should be intermittent.

PR #7200, opened 2020-11-13
Bug 1897423: Update dc strategy type and rolling params

yapei:
Deployment Config Details page is not showing `Max unavailable` and `Max surge`
Fixed it to use correct strategy type and parse from correct path
@jhadvig  Could you help review?


spadgett (approved):
/lgtm

PR #7210, opened 2020-11-13
Fix continual poll for pod graph data in pod details page

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5137

**Analysis / Root cause**: 
Area charts were updated to use generic prometheus poll props and set the endTime property to `Date.now()` by default. This causes the component to constantly change the URL since the endTime is changed every time.

**Solution Description**: 
Remove the default setting for endTime so that it is only set if the caller provides one.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


andrewballantyne (approved):
/lgtm

PR #7212, opened 2020-11-13
Fix loading errors for i18n keys

zherman0:
While the pods table i18n works, there are loading errors because the keys values are loaded before the initialization of the i18next code.  I have refactored to prevent this from being an issue.


spadgett (approved):
/lgtm

PR #7214, opened 2020-11-13
Add Cypress Operator single install mode test

rhamilto:
Replace disabled Protractor-based test with new Cypress one.

Resolves https://issues.redhat.com/browse/CONSOLE-2482


frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  afterEach(() => {
+    checkErrors();
+  });
+
+  after(() => {
+    cy.deleteProject(testName);
+    cy.logout();
+  });
+
+  const operatorPkgName = '3scale-community-operator';
```
spadgett, line 21: We might consider installing a custom catalog source like we were before to lock to a specific operator version, but I think this is fine for now. We'll see if it becomes an issue.

spadgett (commented):
We'll need to update the test-prow-e2e.sh script to run this test when `olmFull` is passed as an argument. That will enable the Cypress OLM tests in the OLM merge queue.

https://github.com/openshift/console/blob/master/test-prow-e2e.sh#L44-L48

spadgett:
We should remove the legacy Protractor test as well.

rhamilto:
/hold

there be bugs

spadgett (commented):
edit: Nevermind, the tests are in the right place.

/cc @dtaylor113

spadgett (approved):
/lgtm

PR #7217, opened 2020-11-16
Bug 1879140: Set proper loglevel for auth errors

jhadvig:
Print the authentication failures that indicate that the user is not logged in on a `Debug` loglevel
```
  2020-07-01T17:39:31.533556276Z 2020/07/1 17:39:31 server: authentication failed: http: named cookie not present
  2020-07-01T17:39:34.888946631Z 2020/07/1 17:39:34 server: authentication failed: unauthenticated, no value for cookie openshift-session-token
```

Also updated the `klog` methods upon calling `redirectAuthError()` 

/assign @spadgett

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -369,13 +369,13 @@ func (a *Authenticator) CallbackFunc(fn func(loginInfo LoginJSON, successURL str
 		}
 
 		if code == "" {
-			klog.Info("missing auth code in query param")
+			klog.Error("missing auth code in query param")
 			a.redirectAuthError(w, errorMissingCode)
 			return
 		}
 
 		if urlState != cookieState.Value {
-			klog.Error("State in url does not match State cookie")
+			klog.Error("state in url does not match State cookie")
 			a.redirectAuthError(w, errorInvalidState)
 			return
 		}
@@ -383,7 +383,7 @@ func (a *Authenticator) CallbackFunc(fn func(loginInfo LoginJSON, successURL str
 		oauthConfig, lm := a.authFunc()
 		token, err := oauthConfig.Exchange(ctx, code)
 		if err != nil {
-			klog.Infof("unable to verify auth code with issuer: %v", err)
+			klog.Errorf("unable to verify auth code with issuer: %v", err)
 			a.redirectAuthError(w, errorInvalidCode)
 			return
 		}
--- a/pkg/server/middleware.go
+++ b/pkg/server/middleware.go
@@ -25,7 +25,7 @@ func authMiddlewareWithUser(a *auth.Authenticator, handlerFunc func(user *auth.U
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		user, err := a.Authenticate(r)
 		if err != nil {
-			klog.Infof("authentication failed: %v", err)
+			klog.V(4).Infof("authentication failed: %v", err)
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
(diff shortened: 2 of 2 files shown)


pkg/auth/auth.go
```diff
@@ -369,21 +369,21 @@ func (a *Authenticator) CallbackFunc(fn func(loginInfo LoginJSON, successURL str
...
 		if code == "" {
-			klog.Info("missing auth code in query param")
+			klog.Error("missing auth code in query param")
 			a.redirectAuthError(w, errorMissingCode)
 			return
 		}
 
 		if urlState != cookieState.Value {
-			klog.Error("State in url does not match State cookie")
+			klog.Error("state in url does not match State cookie")
```
spadgett, line 378: This one we should log at a higher level. I don't think we ever expect to see this.

pkg/auth/auth.go
```diff
@@ -369,21 +369,21 @@ func (a *Authenticator) CallbackFunc(fn func(loginInfo LoginJSON, successURL str
...
 			a.redirectAuthError(w, errorInvalidState)
 			return
 		}
 		ctx := oidc.ClientContext(context.TODO(), a.clientFunc())
 		oauthConfig, lm := a.authFunc()
 		token, err := oauthConfig.Exchange(ctx, code)
 		if err != nil {
-			klog.Infof("unable to verify auth code with issuer: %v", err)
+			klog.Errorf("unable to verify auth code with issuer: %v", err)
```
spadgett, line 386: This one is not expected as well.

pkg/server/middleware.go
```diff
@@ -43,13 +43,13 @@ func authMiddlewareWithUser(a *auth.Authenticator, handlerFunc func(user *auth.U
 		}
 		if !safe {
 			if err := a.VerifySourceOrigin(r); err != nil {
-				klog.Infof("invalid source origin: %v", err)
+				klog.V(4).Infof("invalid source origin: %v", err)
```
spadgett, line 46: Also not expected. We should surface this one.

pkg/server/middleware.go
```diff
@@ -43,13 +43,13 @@ func authMiddlewareWithUser(a *auth.Authenticator, handlerFunc func(user *auth.U
...
 				w.WriteHeader(http.StatusForbidden)
 				return
 			}
 
 			if err := a.VerifyCSRFToken(r); err != nil {
-				klog.Infof("invalid CSRFToken: %v", err)
+				klog.V(4).Infof("invalid CSRFToken: %v", err)
```
spadgett, line 52: Also not expected.

spadgett (commented):
Some of these errors are unexpected and indicate a real problem. I think we should leave those logged by default.

jhadvig:
@spadgett comments addressed. PTAL

pkg/server/middleware.go
```diff
@@ -25,7 +25,7 @@ func authMiddlewareWithUser(a *auth.Authenticator, handlerFunc func(user *auth.U
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		user, err := a.Authenticate(r)
 		if err != nil {
-			klog.Infof("authentication failed: %v", err)
+			klog.V(4).Infof("authentication failed: %v", err)
```
spadgett, line 28: Will this get logged by default? We want to avoid the named cookie not present errors in the log, which is a red herring.

jhadvig: So the default log level `Normal` which is represented by `2`, so no it should not be logged by default. Only if the console logLevel is set to `Debug`, which is represented by `4`.

spadgett: You have everything at 4, but some of these we want to see by default. (See my initial review.)

spadgett: Ping @jhadvig, this comment is still pending :)

pkg/server/middleware.go
jhadvig: if this one is expected, then shouldn't we put it on higher log level ?

pkg/auth/auth.go
jhadvig: then it should be logged by default, right?

pkg/auth/auth.go
jhadvig: > I don't think we ever expect to see this.

in that case there should not be any log level set, since its not expected, right?

jhadvig:
@spadgett I've addressed your comments. Although not both of these two errors will be surfaced only on higher loglevel:
```
2020-07-01T17:39:31.533556276Z 2020/07/1 17:39:31 server: authentication failed: http: named cookie not present
2020-07-01T17:39:34.888946631Z 2020/07/1 17:39:34 server: authentication failed: unauthenticated, no value for cookie openshift-session-token
```

since they are both executed upon [authentication](https://github.com/openshift/console/blob/master/pkg/auth/auth_openshift.go#L175-L181). If we would want to differentiate between them we would need to to some additional rafactoring.

spadgett (approved):
/lgtm

PR #7220, opened 2020-11-16
Correct the order of dev perspective nav for pipelines

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5138

**Analysis / Root cause**: 
Pipelines was moved to a separate package and adds the `Pipelines` navigation item. I attempts to add it before an item that may or may not exist so simply adding `mergeBefore` is not acceptable.

**Solution Description**: 
Adds the ability for plugins to add items before or after other items by Id or an order of ids to look for.
Changed `mergeBefore` to `insertBefore`
Added `insertAfter`
Allow for arrays of items rather than a single item.
Use ids rather than names since names will not be consistent with i18n.
Add Ids to current nav items to allow plugins to use those Ids for insertion points.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug
/kind feature

### DIFF
--- a/frontend/packages/console-app/src/plugin.tsx
+++ b/frontend/packages/console-app/src/plugin.tsx
@@ -230,6 +230,7 @@ const plugin: Plugin<ConsumedExtensions> = [
   {
     type: 'NavItem/ResourceCluster',
     properties: {
+      id: 'storage',
       section: 'Storage',
       componentProps: {
         name: 'Volume Snapshot Contents',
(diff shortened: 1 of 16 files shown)


christianvogt:
@jeff-phillips-18 the side effect of your change is that the `id` attribute now appears on each nav item. I think we should remove this as we don't want to use html `id` attributes.

jeff-phillips-18:
Updated to put the `id` for nav items on the properties rather than using the componentProperties

PR #7221, opened 2020-11-16
add expect library for kubevirt tests

suomiy:
@spadgett can you please review and update the tectonic-console-builder image?

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -27,7 +27,7 @@ RUN chmod 777 -R ${HOME}
 
 RUN apt-get update \
     && apt-get install --no-install-recommends -y -q \
-    curl wget git unzip bzip2 jq \
+    curl wget git unzip bzip2 jq expect \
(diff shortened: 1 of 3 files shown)


suomiy:
bump: did a PR against https://github.com/openshift/release/pull/13715

Dockerfile.builder
```diff
@@ -27,7 +27,7 @@ RUN chmod 777 -R ${HOME}
 
 RUN apt-get update \
     && apt-get install --no-install-recommends -y -q \
-    curl wget git unzip bzip2 jq \
+    curl wget git unzip bzip2 jq expect \
```
spadgett, line 30: We probably should update node and probably kubectl at the same time so we don't have to replace the builder image twice. @zherman0 was looking at that.

suomiy: so that means, once this PR is merged? https://github.com/openshift/console/pull/7280

should I update kubectl in this PR to some version? or is/will be there a special PR for that?

spadgett: Sorry @suomiy, I missed this question. Basically I think #7306 and this should be combined. That way we only need to update the builder image once since there are several steps.

spadgett:
Closing in favor of #7306, which contains this commit.

/close

PR #7222, opened 2020-11-16
Convert Guided tour data from local storage to user settings

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5083

**Analysis / Root cause**: convert the local storage to user settings

**Solution Description**: use `useUserSettings` hook to store the local storage data into a config map.

**Unit test coverage report**: Added

**Test setup:**

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/tour/tour-context.ts
+++ b/frontend/packages/console-app/src/components/tour/tour-context.ts
@@ -6,6 +6,7 @@ import {
   Dispatch,
   ReducerAction,
   useEffect,
+  useCallback,
 } from 'react';
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
@@ -16,12 +17,9 @@ import { getFlagsObject } from '@console/internal/reducers/features';
 import { useExtensions } from '@console/plugin-sdk';
 import { RootState } from '@console/internal/redux';
 import { isGuidedTour, GuidedTour } from '@console/plugin-sdk/src/typings/guided-tour';
-import { TourActions } from './const';
-import {
-  getTourLocalStorageForPerspective,
-  filterTourBasedonPermissionAndFlag,
-  setTourCompletionLocalStorageDataForPerspective,
-} from './utils';
+import { useUserSettingsCompatibility } from '@console/shared/src/hooks/useUserSettingsCompatibility';
+import { TourActions, TOUR_LOCAL_STORAGE_KEY } from './const';
+import { filterTourBasedonPermissionAndFlag } from './utils';
 import { TourDataType, Step } from './type';
 
 type TourStateAction = { type: TourActions; payload?: { completed?: boolean } };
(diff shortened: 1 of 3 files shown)


frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -81,6 +82,43 @@ const getRequiredFlagsByTour = createSelector(
...
+
+const getTourLocalStorageData = (): TourLocalStorageData =>
+  JSON.parse(localStorage.getItem(TOUR_LOCAL_STORAGE_KEY));
+
+const useTourStateForPerspective = (
+  perspective: string,
+): [TourLocalStorageType, (completed: boolean) => void] => {
+  const [tourLocalState, setTourLocalState] = useState<TourLocalStorageData>(
+    getTourLocalStorageData() ?? {},
+  );
```
sahil143, line 101: Replace this `useState` hook with `useUserSettings` hook

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -16,11 +17,12 @@ import { getFlagsObject } from '@console/internal/reducers/features';
 import { useExtensions } from '@console/plugin-sdk';
 import { RootState } from '@console/internal/redux';
 import { isGuidedTour, GuidedTour } from '@console/plugin-sdk/src/typings/guided-tour';
-import { TourActions } from './const';
+import { useUserSettingsCompatibility } from '@console/shared/src/hooks/useUserSettingsCompatibility';
+import { TourActions, TOUR_LOCAL_STORAGE_KEY } from './const';
 import {
-  getTourLocalStorageForPerspective,
+  // getTourLocalStorageForPerspective,
```
invincibleJai, line 23: nit: can remove comment

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -81,6 +83,43 @@ const getRequiredFlagsByTour = createSelector(
...
 
+type TourLocalStorageType = {
+  completed: boolean;
+};
+
+type TourLocalStorageData = {
+  [key: string]: TourLocalStorageType;
+};
+
+const TOUR_CONFIGMAP_KEY = `console.guidedTour`;
```
invincibleJai, line 94: maybe can store `console` as const here console-shared/src/constants/common.ts , can be used for all scenarios. WDYT?

frontend/packages/console-app/src/components/tour/tour-context.ts
```diff
@@ -107,12 +149,13 @@ export const useTourValuesForContext = (): TourContextType => {
...
 
   useEffect(() => {
-    const { completed: initCompleted } = getTourLocalStorageForPerspective(activePerspective);
-    tourDispatch({ type: TourActions.initialize, payload: { completed: initCompleted } });
+    tourDispatch({ type: TourActions.initialize, payload: { completed } });
     setPerspective(activePerspective);
-  }, [activePerspective]);
+    // only run effect when the active perspective changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activePerspective, loaded]);
```
invincibleJai, line 156: is `loaded` intentional here as only used as dependency ?

frontend/packages/console-app/src/components/tour/tour-context.ts
sahil143: It's fine to keep these in their respective files as they are not being used anywhere else.

frontend/packages/console-app/src/components/tour/tour-context.ts
sahil143: Yes, loaded is intentional. If state doesn't exist for a perspective after the config map is loaded, this effect sets the default state for that perspective

invincibleJai:
Thanks @sahil143 looks good , we'll revisit this later once we have required changes in

invincibleJai:
/lgtm

Verified the changes , works as expected

jerolimov:
Tested this locally, works fine.
/lgtm

@sahil143 @invincibleJai Noticed one little strange thing here, because the component and hook was never unmounted, it does not removes the local storage key. But when removing the entry from the ConfigMap the old localStorage key is still available.
It's an edge case, but maybe we should remove the localStorage keys when the ConfigMap was loaded or updated successfully instead of unmount? Wdyt?

/assign @christianvogt
Has lgtm from Sahil, Jai and me. Can you check it and give an approval if everything is fine? Thanks

christianvogt:
@invincibleJai @jerolimov maybe we go back to also removing the key when we set a new value

PR #7225, opened 2020-11-16
Add Cypress Operator global install mode test

rhamilto:
Replace disabled Protractor-based test with new Cypress one.

Resolves https://issues.redhat.com/browse/CONSOLE-2483


rhamilto:
/hold

for https://github.com/openshift/console/pull/7214 to merge first

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      'contain',
+      'Operands are declarative components used to define the behavior of the application.',
+    );
+  });
+
+  it(`displays ${operatorName} ${operatorInstance} creation form`, () => {
+    cy.log('navigate to the form');
+    cy.byLegacyTestID('dropdown-button')
+      .click()
+      .byTestDropDownMenu('storageclusters.core.libopenstorage.org')
```
dtaylor113, line 78: I'm getting concerns about using custom commands in this mannor, noticing some flakes.  This issue may be that by executing `.byTestDropDownMenu` in the middle of the chain, executes `cy.get('[data-test-dropdown-menu="${selector}"]')` -which is not part of the chain, rather starts a new one!   I think we need to make some of our helper commands [Dual-Commands](https://docs.cypress.io/api/cypress-api/custom-commands.html#Dual-Commands). A dual command can either start a chain of commands or be chained off of an existing one.

For this particular case you may just want to use:
```
cy.log('navigate to the form');
    cy.byLegacyTestID('dropdown-button')
      .click()
      .get(`[data-test-dropdown-menu="${selector}"]`)`
```
...until we work out what we need to do.

dtaylor113 (changes requested):
HI @rhamilto, looks great!  I like how our helper methods/objs make the tests extremely readable and concise!  One comment about using custom commands in a chain of commands.  -thanks

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
dtaylor113: We may need to rewrite our `byXXX` custom selector methods if we want to also use them in middle of a chain (for example):
```
Cypress.Commands.add('byTestDropDownMenu', {
  prevSubject: 'optional'
}, (subject, selector) => {
  // subject may be defined or undefined
  // so you likely want to branch the logic
  // based off of that

  if (subject) {
    // wrap the existing subject
    // and do something with it
    cy.wrap(subject).get(`[data-test-operator-row="${selector}"]`),
  } else {
    cy.get(`[data-test-operator-row="${selector}"]`),
  }
});
```

rhamilto: Thanks, @dtaylor113.  Updated per your recommendation.  Will open a PR to make the same change in #7214 .

dtaylor113 (approved):
/lgtm

PR #7236, opened 2020-11-17
Query Browser: Remove Y-axis ticks

kyoto:
We have Y-axis strokes running the width of the graph, so the ticks seem
redundant.

Before | After
-|-
before | after

FYI @cshinn


spadgett (approved):
/lgtm

PR #7237, opened 2020-11-17
add default workspace when auto starting pipeline from Add flow

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5118

**Analysis / Root cause**: 
As a user, when I'm getting my Pipeline Run auto-started off the add flow, I need a PVC volumeClaimTemplate otherwise my Pipeline Run doesn't successfully run.

**Solution Description**: 
1. Attach a default volumeClaimTemplate to the pipelineRun when auto starting a pipeline through the +Add flow with pipeline metadata label attached to it.
2. Use the pipeline metadata label pre select the `PVC` in the Start pipeline modal and Add Trigger modal or fallback to select the `EmptyDirectory` option

1. PVC getting created through the add flow pipeline auto start feature (via `volumeclaimTemplate` in the pipelinerun).

pvc
2. In Start Modal, PVC getting auto selected using the pipeline label attached to it.
auto-selecting-pvc
3. In Add trigger, PVC getting auto selected
trigger

4. Pipelines are now getting executed successfully
image

image

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @andrewballantyne @jerolimov  @vdemeester @openshift/team-devconsole-ux

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
@@ -1,10 +1,13 @@
 import * as _ from 'lodash';
 import { k8sCreate } from '@console/internal/module/k8s';
 import { PipelineData } from '../import-types';
-import { Pipeline, PipelineRun } from '../../../utils/pipeline-augment';
+import { Pipeline, PipelineRun, PipelineWorkspace } from '../../../utils/pipeline-augment';
 import { PipelineModel } from '../../../models';
 import { createPipelineResource } from '../../pipelines/pipeline-resource/pipelineResource-utils';
-import { convertPipelineToModalData } from '../../pipelines/modals/common/utils';
+import {
+  convertPipelineToModalData,
+  getDefaultVolumeClaimTemplate,
+} from '../../pipelines/modals/common/utils';
 import { submitStartPipeline } from '../../pipelines/modals/start-pipeline/submit-utils';
 import { StartPipelineFormValues } from '../../pipelines/modals/start-pipeline/types';
 
@@ -67,6 +70,11 @@ export const createPipelineForImportFlow = async (
 export const createPipelineRunForImportFlow = async (pipeline: Pipeline): Promise<PipelineRun> => {
   const pipelineInitialValues: StartPipelineFormValues = {
     ...convertPipelineToModalData(pipeline),
+    workspaces: (pipeline.spec.workspaces || []).map((workspace: PipelineWorkspace) => ({
+      ...workspace,
+      type: 'volumeClaimTemplate',
+      data: getDefaultVolumeClaimTemplate(pipeline?.metadata?.name),
+    })),
     secretOpen: false,
   };
   return submitStartPipeline(pipelineInitialValues, pipeline);
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/hooks.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/hooks.ts
@@ -1,10 +1,15 @@
 import { match as RMatch } from 'react-router-dom';
 import { useTabbedTableBreadcrumbsFor } from '@console/shared';
-import { K8sKind, referenceForModel } from '@console/internal/module/k8s';
+import {
+  K8sKind,
+  PersistentVolumeClaimKind,
+  referenceForModel,
+} from '@console/internal/module/k8s';
 import {
   useK8sWatchResource,
   WatchK8sResource,
 } from '@console/internal/components/utils/k8s-watch-hook';
+import { PersistentVolumeClaimModel } from '@console/internal/models';
 import { pipelinesTab } from '../../utils/pipeline-utils';
 import { PipelineRun, getLatestRun } from '../../utils/pipeline-augment';
 import { TektonResourceLabel } from './const';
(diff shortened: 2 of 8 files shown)


vdemeester (commented):
This uses `volumeClaim` by default right ? If that's the case, it works, *but* this means there might be a few "gotchas":
- 2 pipelines running in parallel might run into problems as they will want to write to the same PVC
- the `git-clone` task needs to set a parameter (that "reset" the workspace before cloning) — that parameter should default to "true" but doesn't as of today I think.

karthikjeeyar:
@vdemeester Thank you for the reviewing :) 
> This uses `volumeClaim` by default right ? If that's the case, it works, _but_ this means there might be a few "gotchas"
>
> * 2 pipelines running in parallel might run into problems as they will want to write to the same PVC

Hmmm... 2 different pipelines will have 2 different PVC generated here, but for a single pipeline there can be parallel pipelineruns which might use same PVC to read/write. Any suggestions to overcome this issues?

> * the `git-clone` task needs to set a parameter (that "reset" the workspace before cloning) — that parameter should default to "true" but doesn't as of today I think.

In pipeline 1.2, `git-clone` task has `deleteExisting` parameter which has default value set to "true". Is this what you are refferring when you say `reset` ? If yes, then we are good there.

vdemeester:
> Hmmm... 2 different pipelines will have 2 different PVC generated here, but for a single pipeline there can be parallel pipelineruns which might use same PVC to read/write. Any suggestions to overcome this issues

Right, I wasn't clear indeed, I was talking about one pipeilne and two parallel pipelinerun. This could happen "relatively" quickly if this is hooked to a git repository and there is multiple developer on it. Using `volumeClaimTemplate` would help there (but this would mean a new PVC is created for each run.. which means at some point, the user might run out of space if pipelinerun are not cleaned up)

> In pipeline 1.2, `git-clone` task has `deleteExisting` parameter which has default value set to "true". Is this what you are refferring when you say `reset` ? If yes, then we are good there.

Yes :+1: very good then :stuck_out_tongue_closed_eyes:

karthikjeeyar:
Agreed, PVC isn't a best option in triggerTemplates @vdemeester We have a story to track this [ODC-4772]( https://issues.redhat.com/browse/ODC-4772), but it dint make it to 4.7 though.

frontend/packages/pipelines-plugin/src/components/pipelines/hooks.ts
```diff
@@ -9,6 +13,7 @@ import { pipelinesTab } from '../../utils/pipeline-utils';
 import { PipelineRun, getLatestRun } from '../../utils/pipeline-augment';
 import { TektonResourceLabel } from './const';
 import { PipelineRunModel } from '../../models';
+import { PersistentVolumeClaimModel } from '@console/internal/models';
```
abhinandan13jan, line 16: nit: move import

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -67,6 +72,11 @@ export const createPipelineForImportFlow = async (
 export const createPipelineRunForImportFlow = async (pipeline: Pipeline): Promise<PipelineRun> => {
   const pipelineInitialValues: StartPipelineFormValues = {
     ...convertPipelineToModalData(pipeline),
+    workspaces: (pipeline.spec.workspaces || []).map((workspace: PipelineWorkspace) => ({
```
abhinandan13jan, line 75: is it better to have a null check? pipeline.spec?.workspaces

karthikjeeyar: I have added a fallback  `|| []`  above, so it will not break.

abhinandan13jan: sure but if a pipeline doesn't have specs yet, could it through cannot read property `workspaces` of undefined

andrewballantyne: You can't create a Pipeline without a `spec`:
image

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -56,6 +59,8 @@ export const createPipelineForImportFlow = async (
         return { ...param, default: getImageUrl(name, namespace) };
       case 'DOCKERFILE':
         return { ...param, default: dockerfilePath };
+      case 'MAJOR_VERSION':
+        return { ...param, default: 'latest' }; // This needs to be removed when backend sends default value
```
andrewballantyne, line 63: Probably should remove this 🤔  It's in the `preview` channel, we can install 4.6 which is 1.1.z and that will work fine. They need to update 1.2.z to have this default.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/AddTriggerModal.tsx
```diff
@@ -48,6 +51,7 @@ const AddTriggerModal: React.FC<AddTriggerModalProps> = ({ pipeline, close }) =>
       initialValues={initialValues}
       onSubmit={handleSubmit}
       validationSchema={addTriggerSchema(t)}
+      enableReinitialize
```
andrewballantyne, line 54: Hmmm, not sure about this... kinda would want to wait for the data before rendering Formik. Reinitializing data could have side-effects mid-usage.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/start-pipeline/StartPipelineModal.tsx
```diff
@@ -54,6 +56,7 @@ const StartPipelineModal: React.FC<StartPipelineModalProps & ModalComponentProps
       initialValues={initialValues}
       onSubmit={handleSubmit}
       validationSchema={startPipelineSchema(t)}
+      enableReinitialize
```
andrewballantyne, line 59: Same issue here...

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/AddTriggerModal.tsx
karthikjeeyar: I have removed the `enableReInitiaze` prop and waiting for the data to render Formik.

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
karthikjeeyar: removed it.

andrewballantyne (approved):
/lgtm
/retest

PR #7241, opened 2020-11-17
add non blocking pipeline creation flow

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5048

**Analysis / Root cause**: 
As a user, I'd like to still be able to finish my add flow even if the Pipeline Run cannot be created for the added Pipeline.

**Solution Description**: 
Removed the pipelinerun creation request from the promise collection and added it in a try/catch block seperately.

**Before**:
Blocks the add flow redirection and shows the error.
image

**After**:
Does not block the add flow redirection and suppresses the pipelinerun creation error
non-blocking-flow
**Topology Sidebar**
image

import-submit-utils.spec.ts 
image
PipelineOverview.spec.tsx
image
pipeline-overview-utils.spec.ts
image

**Test setup:**
1. Install Openshift pipeilnes operator with `preview` channel to get 1.2 operator
2. Go to add from git flow
3. Use dotnet image `https://github.com/redhat-developer/s2i-dotnetcore-ex.git` and choose dotnet builder image
4. Add Pipeline option
5. click on create button

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @andrewballantyne @bgliwa01  @openshift/team-devconsole-ux

### DIFF
--- a/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
@@ -10,7 +10,7 @@ import {
   SecretModel,
 } from '@console/internal/models';
 import * as pipelineUtils from '@console/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils';
-import { PipelineModel, PipelineRunModel } from '@console/pipelines-plugin/src/models';
+import { PipelineModel } from '@console/pipelines-plugin/src/models';
 import { Resources } from '../import-types';
 import * as submitUtils from '../import-submit-utils';
 import { defaultData, nodeJsBuilderImage as buildImage } from './import-submit-utils-data';
@@ -183,13 +183,7 @@ describe('Import Submit Utils', () => {
         'createPipelineRunForImportFlow',
       );
 
-      const returnValue = await createOrUpdateResources(
-        mockData,
-        buildImage.obj,
-        false,
-        false,
-        'create',
-      );
+      await createOrUpdateResources(mockData, buildImage.obj, false, false, 'create');
       expect(createPipelineResourceSpy).toHaveBeenCalledWith(
         mockData.name,
         mockData.project.name,
@@ -200,8 +194,7 @@ describe('Import Submit Utils', () => {
         mockData.docker.dockerfilePath,
       );
       expect(createPipelineRunResourceSpy).toHaveBeenCalledTimes(1);
-      const models = returnValue.map((data) => _.get(data, 'model.kind'));
-      expect(models.includes(PipelineRunModel.kind)).toEqual(true);
+      expect(createPipelineRunResourceSpy).not.toThrowError();
       done();
     });
 
@@ -230,6 +223,44 @@ describe('Import Submit Utils', () => {
       );
       const pipelineRunResource = returnValue[1].data;
       expect(pipelineRunResource.metadata.name.includes(mockData.name)).toEqual(true);
+      done();
+    });
+    it('should suppress the error if the pipelinerun creation fails with the error', async (done) => {
+      const mockData = _.cloneDeep(defaultData);
+      mockData.resources = Resources.Kubernetes;
+      mockData.pipeline.enabled = true;
+      const createPipelineResourceSpy = jest.spyOn(pipelineUtils, 'createPipelineForImportFlow');
+
+      const createPipelineRunSpy = jest
+        .spyOn(pipelineUtils, 'createPipelineRunForImportFlow')
+        .mockImplementation(() => Promise.reject(new Error('PipelineRun error')));
+
+      const returnValue = await createOrUpdateResources(
+        mockData,
+        buildImage.obj,
+        false,
+        false,
+        'create',
+      );
+      const models = returnValue.map((r) => r['model.kind']);
+      expect(createPipelineResourceSpy).not.toThrow();
+      expect(createPipelineRunSpy).not.toThrow();
+      expect(returnValue).toBeDefined();
+      expect(models).toHaveLength(6);
+      done();
+    });
+
+    it('should throw error if the deployment creation fails with the error', async (done) => {
+      const mockData = _.cloneDeep(defaultData);
+      mockData.resources = Resources.Kubernetes;
+      mockData.pipeline.enabled = true;
+      jest
+        .spyOn(submitUtils, 'createOrUpdateDeployment')
+        .mockImplementation(() => Promise.reject(new Error('Deployment')));
+
+      await expect(
+        createOrUpdateResources(mockData, buildImage.obj, false, false, 'create'),
+      ).rejects.toEqual(new Error('Deployment'));
       done();
     });
 
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
@@ -14,6 +14,8 @@ import { PipelineRunModel, PipelineModel } from '../../../models';
 import TriggerLastRunButton from './TriggerLastRunButton';
 import PipelineRunItem from './PipelineRunItem';
 import PipelineStartButton from './PipelineStartButton';
+import { isPipelineNotStarted, removePipelineNotStarted } from './pipeline-overview-utils';
+import PipelineOverviewAlert from './PipelineOverviewAlert';
 
 const MAX_VISIBLE = 3;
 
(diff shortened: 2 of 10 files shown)


andrewballantyne:
Ah, I missed AC in the Story, apologizes for that @karthikjeeyar - If you look at the [design](https://docs.google.com/document/d/1dH8w11ZPKplScMwA93rQ0je8NPrTToKIikFtc7sZvBc/edit#heading=h.4me1hiepdkww), we have a need to show a message in the sidebar.

🤔 To do this, perhaps we find a way to add some form of `sessionStorage` (NOT `localStorage`) value when it fails?

> The read-only `sessionStorage` property accesses a session Storage object for the current origin. `sessionStorage` is similar to `localStorage`; the difference is that while data in `localStorage` doesn't expire, data in `sessionStorage` is cleared when the page session ends. 

Ref: [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

Our goal here is to just inform the user immediately, not to persist it indefinitely.

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
```diff
@@ -232,6 +225,39 @@ describe('Import Submit Utils', () => {
...
+
+      const returnValue = await createOrUpdateResources(
+        mockData,
+        buildImage.obj,
+        false,
+        false,
+        'create',
+      );
+      const models = returnValue.map((r) => r['model.kind']);
+      expect(models).toHaveLength(6);
```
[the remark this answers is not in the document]
karthikjeeyar: Pipeline resource creation call is not pushed into the requests array (uses await) , so i will add a spy on the creation call and expect it to not throw the error.

karthikjeeyar:
@andrewballantyne I have updated the code to show the alert in the sidebar, updated screenshots and added more tests.

frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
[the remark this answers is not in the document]
andrewballantyne: Ah so it does... apologizes for not looking at the code! I misremembered that the PipelineRun was added to the list and you removed it to use await with this PR. 👍

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
```diff
@@ -31,9 +33,24 @@ const PipelinesOverview: React.FC<PipelinesOverviewProps> = ({
   const {
     metadata: { name, namespace },
   } = pipeline;
+  const [showAlert, setShowAlert] = React.useState(isPipelineNotStarted(name));
+
+  React.useEffect(() => {
+    setShowAlert(isPipelineNotStarted(name));
+  }, [name]);
```
andrewballantyne, line 40: I think we want to just live read it on each render... without the hooks. This way wouldn't need to worry so much about unsetting it when the user clicks the 'x'.

```suggestion
  const showAlert = isPipelineNotStarted(name);
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
```diff
@@ -31,9 +33,24 @@ const PipelinesOverview: React.FC<PipelinesOverviewProps> = ({
...
+
   return (
     <>
       <SidebarSectionHeading text={PipelineRunModel.labelPlural}>
+        {showAlert && (
```
andrewballantyne, line 45:
```suggestion
        {(showAlert && pipelineRuns.length === 0) && (
```
If they start up a Pipeline Run we don't want to show it.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/pipeline-overview-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const setPipelineNotStarted = (pipelineName: string): void => {
+  if (!pipelineName) return;
+  const pipelineList = getNotStartedPipelines();
+  if (pipelineList.length === 0 || !pipelineList.includes(pipelineName)) {
+    pipelineList.push(pipelineName);
+  }
+  sessionStorage.setItem(PIPELINE_RUN_AUTO_START_FAILED, JSON.stringify(pipelineList));
+};
```
andrewballantyne, line 22: We want to hold onto the namespace as well... if you just do name, I could switch between projects and get unexpected results.

Plus you probably want to only write it back out when you modify it... no need to write the same value you got back to the storage if it didn't change.

```suggestion
export const setPipelineNotStarted = (namespace: string, pipelineName: string): void => {
  if (!pipelineName || !namespace) return;

  const pipelineData = getNotStartedPipelines();
  if (!pipelineData[namespace]) {
    pipelineData[namespace] = [];
  }
  if (!pipelineData[namespace].includes(pipelineName)) {
    pipelineData[namespace].push(pipelineName);
    sessionStorage.setItem(PIPELINE_RUN_AUTO_START_FAILED, JSON.stringify(pipelineData));
  }
};
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/pipeline-overview-utils.ts
```diff
@@ -0,0 +1,22 @@
+import { PIPELINE_RUN_AUTO_START_FAILED } from '@console/dev-console/src/components/topology/redux/const';
+
+export const getNotStartedPipelines = () => {
+  return JSON.parse(sessionStorage.getItem(PIPELINE_RUN_AUTO_START_FAILED) ?? '[]');
```
andrewballantyne, line 4: Based on my last comment... needs to be an object

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/pipeline-overview-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+export const isPipelineNotStarted = (pipelineName: string): boolean => {
+  return getNotStartedPipelines().includes(pipelineName);
+};
+
+export const removePipelineNotStarted = (pipelineName: string): void => {
+  const newList = getNotStartedPipelines().filter((pName) => pName !== pipelineName);
```
andrewballantyne, line 11: Based on my last(last) comment... need namespace here.

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
karthikjeeyar: removed UseEffect but still using local state as the re render is not quick enough to remove the Alert.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/pipeline-overview-utils.ts
karthikjeeyar: @andrewballantyne Updated the code to store/remove a namespaced session object.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
```diff
@@ -31,9 +33,20 @@ const PipelinesOverview: React.FC<PipelinesOverviewProps> = ({
   const {
     metadata: { name, namespace },
   } = pipeline;
+  const [showAlert, setShowAlert] = React.useState(isPipelineNotStarted(name, namespace));
```
andrewballantyne, line 36: This is not reactive 😞  

I'll need to do some investigation into why we can't just use it in the render.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
andrewballantyne: Ah I see... my bad. 🤔

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-overview/PipelineOverview.tsx
andrewballantyne: Ah okay, I see what is happening here... yeah your original solution made more sense. I'm not sure if we can entirely base our data on the sessionStorage 🤔  (which was my original reason for my suggestion)

andrewballantyne: Yeah, I think we can't get a "single state" of storage here... which is unfortunately. Okay, reset it back to the useEffect hook and we should be good to go.

```
  const [showAlert, setShowAlert] = React.useState(isPipelineNotStarted(name, namespace));

  React.useEffect(() => {
    setShowAlert(isPipelineNotStarted(name, namespace));
  }, [name, namespace]);
```

Sorry for this round-about solution... you had it right the first time.

andrewballantyne (approved):
Just need to rebase and reimplement the hook @karthikjeeyar - Get a LGTM from someone on the team or if I see the push I'll give it. Thanks!

karthikjeeyar:
Rebased and added the hook back.

andrewballantyne (approved):
/lgtm

PR #7242, opened 2020-11-17
adds useUserSettings hook and associated HOC

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4753

**Analysis / Root cause**: 
User preferences were saved in localStorage so not persisted across browsers 

**Solution Description**: 
Added a hook with support for localStorage as fallback

NA

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -1,0 +1,149 @@
+import * as React from 'react';
+// FIXME upgrading redux types is causing many errors at this time
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useSelector } from 'react-redux';
+import { RootState } from '@console/internal/redux';
+import { ConfigMapModel, ProjectRequestModel, ProjectModel } from '@console/internal/models';
+import {
+  K8sResourceKind,
+  k8sGet,
+  k8sCreate,
+  k8sPatch,
+  ConfigMapKind,
+} from '@console/internal/module/k8s';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+
+// can't create project with name prefix with 'openshift-*', once we have proxy need to update
+const USER_SETTING_CONFIGMAP_NAMESPACE = 'console-user-settings';
+
+// This won't be needed once we have proxy api
+const getProject = async () => {
+  try {
+    await k8sGet(ProjectModel, USER_SETTING_CONFIGMAP_NAMESPACE);
+  } catch {
+    await k8sCreate(ProjectRequestModel, {
+      metadata: {
+        name: USER_SETTING_CONFIGMAP_NAMESPACE,
+      },
+    });
+  }
+};
+
+const createConfigMap = async (configMapData: K8sResourceKind): Promise<boolean> => {
+  try {
+    await k8sCreate(ConfigMapModel, configMapData);
+    return true;
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+    return false;
+  }
+};
+
+const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: string) => {
+  if (value !== configMap.data?.[key]) {
+    const patch = [
+      {
+        op: 'replace',
+        path: `/data/${key}`,
+        value,
+      },
+    ];
+    try {
+      await k8sPatch(ConfigMapModel, configMap, patch);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(err);
+    }
+  }
+};
+
+const deseralizeData = <T>(data: T) => {
+  if (typeof data !== 'string') {
+    return data;
+  }
+  try {
+    return JSON.parse(data);
+  } catch {
+    return data;
+  }
+};
+
+const seralizeData = <T>(data: T) => {
+  if (typeof data === 'string') {
+    return data;
+  }
+  return JSON.stringify(data);
+};
+
+export const useUserSettings = <T>(
+  key: string,
+  defaultValue?: T,
+): [T, React.Dispatch<React.SetStateAction<T>>, boolean] => {
+  const defaultValueRef = React.useRef<T>(defaultValue);
+  const keyRef = React.useRef<string>(key);
+  const userUid = useSelector(
+    (state: RootState) => state.UI.get('user')?.metadata?.uid ?? 'kubeadmin',
+  );
+  const configMapResource = React.useMemo(
+    () => ({
+      kind: ConfigMapModel.kind,
+      namespace: USER_SETTING_CONFIGMAP_NAMESPACE,
+      isList: false,
+      name: `user-settings-${userUid}`,
+    }),
+    [userUid],
+  );
+  const [cfData, cfLoaded, cfLoadError] = useK8sWatchResource<K8sResourceKind>(configMapResource);
+  const [settings, setSettings] = React.useState<T>();
+  const [loaded, setLoaded] = React.useState(false);
+
+  React.useEffect(() => {
+    if (cfLoadError || (!cfData && cfLoaded)) {
+      (async () => {
+        await getProject();
+        const cmCreated = await createConfigMap({
+          apiVersion: ConfigMapModel.apiVersion,
+          kind: ConfigMapModel.kind,
+          metadata: {
+            name: `user-settings-${userUid}`,
+            namespace: USER_SETTING_CONFIGMAP_NAMESPACE,
+          },
+          data: {
+            ...(defaultValueRef.current !== undefined && {
+              [keyRef.current]: seralizeData(defaultValueRef.current),
+            }),
+          },
+        });
+        if (!cmCreated) {
+          setSettings(deseralizeData(defaultValueRef.current));
+          setLoaded(true);
+        }
+      })();
+    } else if (
+      cfData &&
+      cfLoaded &&
+      (!cfData.data?.hasOwnProperty(keyRef.current) ||
+        seralizeData(settings) !== cfData.data?.[keyRef.current])
+    ) {
+      setSettings(deseralizeData(cfData.data?.[keyRef.current]) ?? defaultValueRef.current);
+      setLoaded(true);
+    } else if (cfLoaded) {
+      setSettings(deseralizeData(defaultValueRef.current));
+      setLoaded(true);
+    }
+    // This effect should only be run on change of configmap data, status.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cfLoaded, cfLoadError]);
+
+  React.useEffect(() => {
+    if (cfData && cfLoaded && settings !== undefined) {
+      updateConfigMap(cfData, keyRef.current, seralizeData(settings));
+    }
+    // This effect should only be run on change of settings state.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [settings]);
+
+  return [settings, setSettings, loaded];
+};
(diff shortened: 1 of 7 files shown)


frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+    } else if (cfData && cfLoaded && (!settings || settings !== cfData.data?.[key])) {
+      setSettings(deseralizeData(cfData.data?.[key] ?? defaultValue));
+      setLoaded(true);
+    }
+    // This effect should only be run once per key.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [key, cfData, cfLoaded, cfLoadError]);
+
+  React.useEffect(() => {
+    if (cfData && cfLoaded && settings) {
```
christianvogt, line 130: If `settings` is purposely set to a falsey value, it'll never be updated.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+  React.useEffect(() => {
+    if (cfLoadError || (!cfData && cfLoaded)) {
+      (async () => {
+        await getProject();
+        const cmCreated = await createConfigMap(configMapData);
+        if (!cmCreated) {
+          setLoaded(true);
+        }
+      })();
+    } else if (cfData && cfLoaded && (!settings || settings !== cfData.data?.[key])) {
```
christianvogt, line 121: Settings should be allowed to be a falsey value.
You are comparing `settings` (a deserialized value) with `cfData.data[key]` which is a serialized value.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+      setSettings(deseralizeData(cfData.data?.[key] ?? defaultValue));
+      setLoaded(true);
+    }
+    // This effect should only be run once per key.
```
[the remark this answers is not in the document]
invincibleJai: This will be used to update the state whenever configData Changes and have updated check for the key

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+};
+
+const seralizeData = <T>(data: T) => {
+  if (typeof data === 'string') {
+    return data;
+  }
+  return JSON.stringify(data);
+};
+
+export const useUserSettings = <T>(key: string, defaultValue?: T) => {
```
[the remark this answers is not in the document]
invincibleJai: added

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+  const userUid = useSelector(
+    (state: RootState) => state.UI.get('user')?.metadata?.uid ?? 'kubeadmin',
+  );
+  const [configMapResource, configMapData] = getConfigResourceData(
+    userUid,
+    key,
+    seralizeData(defaultValue ?? ''),
+  );
```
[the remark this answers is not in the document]
invincibleJai: updated it

frontend/packages/console-shared/src/hooks/useUserSettings.ts
invincibleJai: yeah make sense , have removed check here

frontend/packages/console-shared/src/hooks/useUserSettings.ts
invincibleJai: missed here, doing the comparison with  serialized value now

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -0,0 +1,26 @@
...
+  const [settings, setSettings, loaded] = useUserSettings(
+    key,
+    localStorage.getItem(storageKey) || defaultValue,
+  );
+  const updaterFunction = React.useCallback(
+    (data) => {
+      setSettings(data);
+    },
+    [setSettings],
+  );
```
sahil143, line 18: no need for this, we can directly pass the setSettings

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  React.useEffect(() => {
+    return () => localStorage.removeItem(storageKey);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
sahil143, line 23:
```suggestion
  React.useEffect(() => () => localStorage.removeItem(storageKey)
    // eslint-disable-next-line react-hooks/exhaustive-deps
  , []);
```

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import { useUserSettings } from './useUserSettings';
+
+export const useUserSettingsCompatibility = <T>(
+  key: string,
+  storageKey: string,
+  defaultValue?: T,
+) => {
+  const [settings, setSettings, loaded] = useUserSettings(
```
sahil143, line 9:
```suggestion
  const [settings, setSettings, loaded] = useUserSettings<T>(
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,136 @@
...
+    () => ({
+      kind: ConfigMapModel.kind,
+      namespace: USER_SETTING_CONFIGMAP_NAMESPACE,
+      isList: false,
+      name: `user-settings-${userUid}`,
+    }),
+    [userUid],
+  );
+  const [cfData, cfLoaded, cfLoadError] = useK8sWatchResource<K8sResourceKind>(configMapResource);
+  const [settings, setSettings] = React.useState();
```
sahil143, line 92:
```suggestion
  const [settings, setSettings] = React.useState<T>();
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const seralizeData = <T>(data: T) => {
+  if (typeof data === 'string') {
+    return data;
+  }
+  return JSON.stringify(data);
+};
+
+export const useUserSettings = <T>(key: string, defaultValue?: T) => {
```
sahil143, line 77:
```suggestion
export const useUserSettings = <T>(key: string, defaultValue?: T): [T, React.Dispatch<React.SetStateAction<T>>, boolean] => {
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,136 @@
...
+  const defaultValueRef = React.useRef(defaultValue);
```
sahil143, line 78:
```suggestion
  const defaultValueRef = React.useRef<T>(defaultValue);
```

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
invincibleJai: yeah passed `setSettings`

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
invincibleJai: Have updated as below as could be string as well in case of localStorage value

```
const [settings, setSettings, loaded] = useUserSettings<string | T>(
    key,
    localStorage.getItem(storageKey) || defaultValue,
  );
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,138 @@
...
+  );
+  const [cfData, cfLoaded, cfLoadError] = useK8sWatchResource<K8sResourceKind>(configMapResource);
+  const [settings, setSettings] = React.useState();
+  const [loaded, setLoaded] = React.useState(false);
+
+  React.useEffect(() => {
+    if (cfLoadError || (!cfData && cfLoaded)) {
+      (async () => {
+        await getProject();
+        const cmCreated = await createConfigMap(configMapData);
```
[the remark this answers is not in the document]
christianvogt: I meant do the following because there's no other use for this data outside of this function.
Also, we don't need to define a default value if it is `undefined` or `null`.
```suggestion
        const cmCreated = await createConfigMap({
      apiVersion: ConfigMapModel.apiVersion,
      kind: ConfigMapModel.kind,
      metadata: {
        name: `user-settings-${userUid}`,
        namespace: USER_SETTING_CONFIGMAP_NAMESPACE,
      },
      data: {
        [keyRef.current]: seralizeData(defaultValueRef.current),
      },
    });
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (!cmCreated) {
+          setSettings(defaultValueRef.current);
+          setLoaded(true);
+        }
+      })();
+    } else if (
+      cfData &&
+      cfLoaded &&
+      (!cfData.data?.hasOwnProperty(keyRef.current) ||
+        seralizeData(settings) !== cfData.data?.[keyRef.current])
```
christianvogt, line 128: You still need to call `setLoaded(true)` even if these conditions fail and `cfLoaded === true`

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -0,0 +1,21 @@
...
+  const [settings, setSettings, loaded] = useUserSettings<T | string>(
+    key,
+    localStorage.getItem(storageKey) || defaultValue,
+  );
+
+  React.useEffect(
+    () => () => localStorage.removeItem(storageKey),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
+  );
```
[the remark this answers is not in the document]
invincibleJai: `setSettings` is being called by component using hook or first useEffect on configMap Data fetch (and this will update ConfigMap) so added on unmount else from `useUserSetting` need return patchStatus or something and update based on that

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -0,0 +1,146 @@
...
+    ) {
+      setSettings(deseralizeData(cfData.data?.[keyRef.current]) ?? defaultValueRef.current);
+      setLoaded(true);
+    }
+    // This effect should only be run on change of configmap data, status.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cfLoaded, cfLoadError]);
+
+  React.useEffect(() => {
+    if (cfData && cfLoaded && settings !== undefined) {
```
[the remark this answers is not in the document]
invincibleJai: yeah , needed this check as initially we are not setting any value to `useState` for settings so it'll be undefined and for that din't wanted to updated Configmap.

i was not sure if feature would set undefined as state explicitly. Any suggestion here

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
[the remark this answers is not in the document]
christianvogt: @invincibleJai right. Forgot about this call. Make sure to only clear it if we have loaded then:
```suggestion
  React.useEffect(
    () => () => {
      if (loaded) {
        localStorage.removeItem(storageKey);
      }
    },
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [loaded],
  );
```

sahil143 (approved):
/lgtm

PR #7243, opened 2020-11-17
Fix issue using 'byTestDropDownMenu'  in a chain

rhamilto:
Follow up fix to #7214 per https://github.com/openshift/console/pull/7225#pullrequestreview-532381003


dtaylor113 (approved):
/lgtm

PR #7244, opened 2020-11-17
Add Cypress OperatorHub test

rhamilto:
Resolves https://issues.redhat.com/browse/CONSOLE-2484


frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-hub.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  it('displays OperatorHub tiles filtered by "Provider Type"', () => {
+    cy.log('enable the Red Hat filter');
+    cy.byTestID('providerType-red-hat').click();
+    cy.log('more than one tile should be present');
+    cy.get('.catalog-tile-pf')
+      .its('length')
+      .should('be.gt', 0);
+    cy.log('track which tile is first');
+    const firstOperatorWithRHFilterApplied = cy
```
dtaylor113, line 37: I'm not sure this is working the way you expect.  Pls see: https://docs.cypress.io/guides/references/best-practices.html#Assigning-Return-Values.   You'll probably want to use closures as descripted [here](https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Return-Values).

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-hub.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    cy.byTestID('search-operatorhub').should('be.empty');
+    cy.get('.catalog-tile-pf')
+      .its('length')
+      .should('be.gt', 0);
+  });
+
+  it('filters Operators by category', () => {
+    cy.log('click the second category filter');
+    cy.get('.vertical-tabs-pf-tab > a')
+      .eq(2)
```
dtaylor113, line 87: This seems somewhat prone to changing/breaking.  Anyway to find the tab/a more deterministacally?

dtaylor113 (changes requested):
Looks pretty good to me, just a couple of concerns.  -thanks

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-hub.spec.ts
dtaylor113: I think something like:
```
cy.get('.catalog-tile-pf').first().find('.catalog-tile-pf-title').then(($catalogTitle => {
    cy.log('diable the Red Hat filter');
    cy.byTestID('providerType-red-hat').click();
    cy.log('enable the Certified filter');
    cy.byTestID('providerType-certified').click();
    cy.log('more than one tile should be present');
    cy.get('.catalog-tile-pf')
      .its('length')
      .should('be.gt', 0);
    cy.log('the first tile for Certified should not be the same as Red Hat');
    cy.get('.catalog-tile-pf')
      .first()
      .find('.catalog-tile-pf-title')
      .should('not.equal', catalogTitle);  // maybe catalogTitle.text or something
})
```

rhamilto:
Thanks for the helpful feedback, @dtaylor113.  Feedback incorporated.  PTAL.

dtaylor113 (approved):
/lgtm

PR #7248, opened 2020-11-18
Bug 1872166: Fix Silences link redirection from developer perspective

vikram-raj:
**Fixes**: 
 https://issues.redhat.com/browse/ODC-4901

**Analysis / Root cause**: 
In the Developer perspective, a Silences link redirects to an unexpected 'Alerts' page after creating a silence.

The page that is shown includes Alerts, Silences, and Alerting Rules tabs, which I believe should only be accessible from the Administrator's perspective.

**Solution Description**: 
Users are taken back to the alerts list from the silence details page where they can see that the alert has been silenced.

Kapture 2020-12-07 at 21 14 31

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -21,6 +21,7 @@ import {
   RedExclamationCircleIcon,
   YellowExclamationTriangleIcon,
 } from '@console/shared';
+import { useActivePerspective } from '@console/shared/src/hooks/useActivePerspective';
 import { withFallback } from '@console/shared/src/components/error/error-boundary';
 import * as UIActions from '../../actions/ui';
 import { coFetchJSON } from '../../co-fetch';
@@ -1011,6 +1012,7 @@ const SilencesDetailsPage = withFallback(
       startsAt = '',
       updatedAt = '',
     } = silence || {};
+    const [activePerspective] = useActivePerspective();
     const { t } = useTranslation();
 
     return (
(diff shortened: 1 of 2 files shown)


kyoto:
Before proceeding with this, I'd like to confirm with @cshinn that we really want to add the silence details page and edit form to the dev console?

If they're not needed, this bug fix would simply be a case of modifying the breadcrumbs link URL.

cshinn:
@kyoto thanks for pointing this out. I think I misunderstood the initial bug that @pneedle-rh filed. I think the best thing to do here is just to take users to the alerts list once the "silence" button is pressed. Sorry the response took a while, I've been out of the office for the past several days

vikram-raj:
@kyoto I updated the PR. Now I update the breadcrumbs link as per perspective. PTAL.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1029,9 +1031,10 @@ const SilencesDetailsPage = withFallback(
               breadcrumbs={[
                 {
                   name: t('monitoring~Silences'),
-                  path: namespace
-                    ? `/dev-monitoring/ns/${namespace}/silences`
-                    : '/monitoring/silences',
+                  path:
+                    activePerspective === 'dev'
```
kyoto, line 1035: I think we should also change the link text to `Alerts` for the dev perspective.

vikram-raj: Updated the text to `Alerts` for the dev perspective.

kyoto:
/lgtm

Thanks @vikram-raj

PR #7251, opened 2020-11-18
Fix redirect loop when running local bridge without valid auth token

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5167

**Analysis / Root cause**: 
When the `oc login` token expires the console triggers a login() call which redirects the user to `/auth/login`, which returns a 404 Page Not Found screen which redirects then again to `/auth/login`.

**Solution Description**: 
Do not redirect the user if the current URL is similar to the target URL.

Only affects local development

Only affects local development

**Test setup:**
```
oc login ....
source ./contrib/oc-environment.sh
export BRIDGE_K8S_AUTH_BEARER_TOKEN=sha256~DISABLEDTOKEN
bin/bridge
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/module/auth.js
+++ b/frontend/public/module/auth.js
@@ -118,6 +118,13 @@ export const authSvc = {
   },
 
   login: () => {
-    window.location = window.SERVER_FLAGS.loginURL;
+    // Ensure that we don't redirect to the current URL in a loop
+    // when using local bridge in development mode without authorization.
+    if (
+      window.location.href !== window.SERVER_FLAGS.loginURL &&
+      window.location.pathname !== window.SERVER_FLAGS.loginURL
+    ) {
+      window.location = window.SERVER_FLAGS.loginURL;
+    }
   },
 };


frontend/public/module/auth.js
```diff
@@ -118,6 +118,13 @@ export const authSvc = {
   },
 
   login: () => {
-    window.location = window.SERVER_FLAGS.loginURL;
+    // Ensure that we don't redirect to the current URL in a loop
+    // when using local bridge in development mode without authorization.
+    if (
+      window.location.href !== window.SERVER_FLAGS.loginURL &&
+      window.location.pathname !== window.SERVER_FLAGS.loginURL
```
spadgett, line 125: I believe `window.SERVER_FLAGS.loginURL` is always the fully qualified URL, so the second check shouldn't be necessary.

jerolimov: I thought this also, but at least on a local dev server it's just the pathname. 🤷‍♂️

jerolimov: @spadgett Want live with `window.location.pathname !== window.SERVER_FLAGS.loginURL` or should we keep both for the case where the loginURL is an absolute url instead of local path (what it is currently).

jerolimov:
Peek 2020-12-09 09-40

spadgett (approved):
/lgtm

PR #7252, opened 2020-11-18
Fix UI crash on opening topology sidebar when a pipeline is associated with an app

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5148

**Gif:**
Peek 2020-11-18 17-21
image

/kind bug

### DIFF
--- a/frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
+++ b/frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
@@ -149,7 +149,7 @@ const rerunPipeline: KebabAction = (
 };
 
 export const rerunPipelineAndStay: KebabAction = (kind: K8sKind, pipelineRun: PipelineRun) => {
-  return rerunPipeline(kind, pipelineRun, null, null);
+  return rerunPipeline(kind, pipelineRun);
 };
 
 export const rerunPipelineAndRedirect: KebabAction = (kind: K8sKind, pipelineRun: PipelineRun) => {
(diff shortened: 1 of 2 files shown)


frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
```diff
@@ -149,7 +149,7 @@ const rerunPipeline: KebabAction = (
 };
 
 export const rerunPipelineAndStay: KebabAction = (kind: K8sKind, pipelineRun: PipelineRun) => {
-  return rerunPipeline(kind, pipelineRun, null, null);
+  return rerunPipeline(kind, pipelineRun, null);
```
karthikjeeyar, line 152: no need to pass  `null`
```suggestion
  return rerunPipeline(kind, pipelineRun);
```

karthikjeeyar: I think we can safely remove the third argument `resources` from the `rerunPipeline` function as it not used anywhere.

divyanshiGupta: Yeah I also thought about that and it makes sense to remove this.

andrewballantyne: 100% agree, why was this changed in the first place? @vikram-raj Based on the git history, you just added two `null` arguments onto the end of this function call during I18n work.

divyanshiGupta: Done

frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
```diff
@@ -116,7 +116,6 @@ type RerunPipelineData = {
 const rerunPipeline: KebabAction = (
   kind: K8sKind,
   pipelineRun: PipelineRun,
-  resources: any,
```
andrewballantyne, line 119: Unfortunately this violates the `KebabAction` type. Please undo this change.

karthikjeeyar:
In topology sidebar, the button name is empty, but it should read 'Start Last  Run'
image

As part of i18n PR all the kebab action `label`  is changed to `labelKey`, Looks like it is missed in  `TriggerLastRunButton.tsx`, this component still uses `label` instead of `labelKey`. @divyanshiGupta This should fix the button text.

karthikjeeyar:
Verified locally, works fine.
/lgtm

andrewballantyne (approved):

PR #7256, opened 2020-11-18
Bug 1899056: Update dc strategy type and rolling params

yapei:
@spadgett  @jhadvig  Could you help review? Thank you


jhadvig (commented):
So I see that this PR does not contain any changes. Dont think thats right.

yapei:
> So I see that this PR does not contain any changes. Dont think thats right.

yeah,sorry I didn't notice that.  Let me close and create a new one tomorrow

PR #7261, opened 2020-11-18
CONSOLE-2485: Add Cypress descriptors test

rhamilto:
Resolves https://issues.redhat.com/browse/CONSOLE-2485


frontend/packages/integration-tests-cypress/support/selectors.ts
```diff
@@ -11,7 +11,6 @@ declare global {
       byTestDropDownMenu(selector: string): Chainable<Element>;
       byTestOperatorRow(selector: string): Chainable<Element>;
       byTestSectionHeading(selector: string): Chainable<Element>;
-      byTestDropdownMenu(selector: string): Chainable<Element>;
```
rhamilto, line 14: I mistakenly duplicated this in #7214

frontend/packages/integration-tests-cypress/support/index.ts
```diff
@@ -71,3 +71,13 @@ export const editKind = (kind: string, humanizeKind: boolean) =>
   actionOnKind(actions.edit, kind, humanizeKind);
 export const deleteKind = (kind: string, humanizeKind: boolean) =>
   actionOnKind(actions.delete, kind, humanizeKind);
+
+export const create = (obj) => {
+  const filename = [
+    Cypress.config('screenshotsFolder'),
+    `${obj.metadata.name}.${obj.kind.toLowerCase()}.json`,
```
dtaylor113, line 78: Hi @rhamilto, `Cypress.config('screenshotsFolder')` refers to `/frontend/gui_test_screenshots/cypress/screenshots/` the location where the cypress screenshots go.  You probably want to create these in `/frontend/gui_test_screenshots/` where the other result .json files get created.  Probably not a big issue as I see you are deleting these files after the test runs.

dtaylor113: This seems to work for me:
```
Cypress.config('screenshotsFolder')
    .toString()
    .replace('/cypress/screenshots', '')
```

rhamilto: Why not just hard code the path rather than manipulating the config value?  For example, 
```
const filename = [
    '/frontend/gui_test_screenshots',
    `${obj.metadata.name}.${obj.kind.toLowerCase()}.json`,
```

dtaylor113: Because we `cd` to the OLM package/dir and that will write the file in `packages/operator-lifecycle-manager/integration-tests-cypress/frontend/gui_test_screenshots`!   Each package (devconsole, olm, ceph) has their own `cypress.json` config file with a `Cypress.config('screenshotsFolder')` which points to the console's gui  `gui_test_screenshots/cypress/screenshots` dir.  Ie. OLM it is `"screenshotsFolder": "../../../gui_test_screenshots/cypress/screenshots"`, so trimming off the `/cypress/screenshots` writes files to the correct place.

Or, If you just do `const filename = './${obj.metadata.name}.${obj.kind.toLowerCase()}.json';`   That should write the files to `packages/operator-lifecycle-manager/integration-tests-cypress/` which should be ok as you are cleaning them up after tests.  I guess the only issue would be if tests fails and cleanup doesn't happen, then you'll have these files left in the location you decide.  (console's `gui_test_screenshots/cypress/screenshots` is already in .gitignore)

rhamilto: Gotcha.  Thanks for clarifying.  Updated with your initial suggestion.

dtaylor113 (approved):
/lgtm

PR #7262, opened 2020-11-18
Cypress: test for existance of 'cypress-a11y-report.json' before running 'yarn cypress-a11y-report'

dtaylor113:
This PR unblocks OLM queue!

It appears that our a11y reporting doesn't scale.  A `cypress-a11y-report.json` is created in each package separately and the a11y script is only looking in `packages/integration-tests-cypress`.  This PR checks for the existance of `packages/integration-tests-cypress/cypress-a11y-report.json` first before executing `yarn cypress-a11y-report`  -so a11y results will only output for `console` tests.

A longer term solution is needed to figure out how to merge all the `cypress-a11y-report.json` files which maybe created in each package dir into a single json file.  Should be straight forward as we already do  `mochawesome-merge` for the cypress-unit json files.


spadgett (approved):
/lgtm

PR #7265, opened 2020-11-19
Bug 1822744: Add Internal Mode Installation Test to Cypress

bipuladh:


frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  cy.clickNavLink(['Operators', 'OperatorHub']);
+  cy.byTestID('search-operatorhub').type('Openshift Container Storage');
+  cy.byTestID('ocs-operator-redhat-operators-openshift-marketplace').click();
+
+  cy.log('Subscribe to OCS Operator');
+  cy.byLegacyTestID('operator-install-btn').click({ force: true });
+  cy.byTestID('Operator recommended namespace:-radio-input').should('be.checked');
+  cy.byTestID('enable-monitoring').click();
+  cy.byTestID('install-operator').click();
+  cy.byTestID('success-icon', { timeout: 180000 }).should('be.visible');
```
dtaylor113, line 23: Are we really expecting to wait 3 minutes for this to succeed?  Is there a `cy.exec('oc ...')` cmd we can call to check status -it's sometimes faster than the UI.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  cy.byLegacyTestID('horizontal-link-Storage Cluster').click();
+  cy.byTestID('yaml-create').click();
+
+  cy.log(`Select ${mode}`);
+  cy.byTestID('Internal-radio-input').should('be.checked');
+
+  // Step 1
+  // Select all worker Nodes
+  cy.get('input[name=check-all]').check();
+  cy.get('input[name=check-all').should('be.checked');
```
dtaylor113, line 52: Duplicate code.  Consider adding a helper method, ex `wizard.checkAll()`.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    .click();
+
+  // Step 2
+  if (encrypted) {
+    cy.log('Enabling Encryption');
+    cy.byTestID('encryption-checkbox').click();
+  }
+  cy.get('.pf-c-button.pf-m-primary')
+    .contains('Next')
+    .click();
```
dtaylor113, line 70: Duplicate code.  Consider adding a helper method, ex `wizard.clickNext()`.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      );
+  });
+
+  // Wait for the storage cluster to reach Ready
+  // Storage Cluster CR flickers so wait for 10 seconds
+  // Disablng until ocs-operator fixes above issue
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(10000);
+
+  return cy.byTestID('resource-status').contains('Ready', { timeout: 900000 });
```
dtaylor113, line 94: We're expecting to wait 15 minutes for a Ready status?  Is there a cy.exec('oc ...') cmd we can call to check status?   Something like [this](https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/support/resources.ts#L32)?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        return cy.exec('oc get po -n openshift-storage -o json');
+      })
+      .then((res) => {
+        const pods = JSON.parse(res.stdout);
+        _.set(initialState, 'pods', pods);
+
+        cy.clickNavLink(['Operators', 'Installed Operators']);
+        cy.byLegacyTestID('item-filter').type('ocs-operator');
+        cy.byTestOperatorRow('OpenShift Container Storage').click();
```
dtaylor113, line 87: Duplicate code.  Consider helper method such as `navToInstalledOperator('ocs-operator')'

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+        modal.submit();
+        modal.shouldBeClosed();
+
+        // Wait for the storage cluster to reach Ready
+        // Storage Cluster CR flickers so wait for 10 seconds
+        // Disablng until ocs-operator fixes above issue
+        // eslint-disable-next-line cypress/no-unnecessary-waiting
+        cy.wait(10000);
+        cy.byTestOperandLink('ocs-storagecluster').click();
+        cy.byTestID('resource-status').contains('Ready', { timeout: 900000 });
```
dtaylor113, line 112: There are reasons why this against best practices.   
https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting
https://medium.com/better-programming/dont-use-wait-in-your-cypress-tests-cbe14746da71

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/ocp-dashboards.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { getPVCJSON } from '../helpers/pvc';
+import { checkErrors } from '../../../integration-tests-cypress/support';
+
+describe('Check OCS Dashboards', () => {
+  before(() => {
+    cy.login();
+    // cy.install();
```
dtaylor113, line 7: Uneccesary comment?

dtaylor113 (changes requested):
Looks good.  There is some duplicate code which could use some helper methods or perhaps new catalog/operator [view](https://github.com/openshift/console/tree/master/frontend/packages/integration-tests-cypress/views).
There is also a lot of waiting which goes against Cypress best practices and can lead to many test flakes:  
https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting
https://medium.com/better-programming/dont-use-wait-in-your-cypress-tests-cbe14746da71

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
bipuladh: `cy.exec` unlike other commands would not get retried in case of failure. We would have to come up with a method that polls the CLI. Even if we go that route we are shaving 15 to 30 seconds max.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
bipuladh: Again similar to checking if the Operator is installed, I think this would be better than to poll the CLI.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
bipuladh: I agree these are against the best practice. But issues in the operator ( as mentioned in the comment ) force us to use these hacks. :/

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
dtaylor113: Ok, I think we might have to come up with a better solution, we'll see how often this test 'flakes'.

dtaylor113 (approved):
/lgtm Thanks!

PR #7269, opened 2020-11-19
Check certificate before uploading

rawagner:


frontend/packages/kubevirt-plugin/src/components/modals/add-template-source/add-template-source.tsx
```diff
@@ -101,7 +118,16 @@ export const AddTemplateSourceModal: React.FC<ModalComponentProps &
       .setNamespace(baseImageNamespace)
       .asResource();
     try {
-      if (dataSource?.value === DataVolumeSourceType.PVC.getValue()) {
+      if (dataSource?.value === DataVolumeSourceType.UPLOAD.getValue()) {
+        try {
```
glekner, line 122: we should avoid nesting try/catch

frontend/packages/kubevirt-plugin/src/components/modals/add-template-source/add-template-source.tsx
```diff
@@ -101,7 +118,16 @@ export const AddTemplateSourceModal: React.FC<ModalComponentProps &
...
+          await coFetch(CDI_UPLOAD_URL_BUILDER(uploadProxyURL));
+        } catch (err) {
+          // the GET request will return an error everytime, but it will be undefined only if the certificate is invalid.
+          if (err?.response === undefined) {
+            setCertError(true);
+            return;
```
glekner, line 128: why not `Throw`?

glekner (changes requested):

rawagner:
@glekner removed nested try/catch

frontend/packages/kubevirt-plugin/src/components/modals/add-template-source/add-template-source.tsx
```diff
@@ -82,6 +96,7 @@ export const AddTemplateSourceModal: React.FC<ModalComponentProps &
   const { dataSource, file, isValid } = state;
 
   const onSubmit = async () => {
+    setError(undefined);
     setAllocating(true);
```
glekner, line 100: IMO we should not `setAllocating(true)` until the certificate check is successful, because it causes the modal to flicker between checks.
Lets create a new boolean `isCheckingCert` and pass it to the `ModalFooter`  `inProgress ` prop

glekner (changes requested):
one comment @rawagner

rawagner:
@glekner fixed all, ptal

PR #7270, opened 2020-11-19
Update CDI importing message in VM status

rawagner:


frontend/packages/kubevirt-plugin/src/constants/vm/vm-status.ts
```diff
@@ -54,25 +54,22 @@ export class VMStatus extends StatusEnum<VMStatusSimpleLabel | StatusSimpleLabel
...
   );
-  static readonly CDI_IMPORT_ERROR = new VMStatus('VMStatus_CDI_IMPORT_ERROR', 'Import error', {
+  static readonly CDI_IMPORT_ERROR = new VMStatus('VMStatus_CDI_IMPORT_ERROR', 'Disk image error', {
     isError: true,
-    group: StatusGroup.CDI,
   });
   static readonly CDI_IMPORTING = new VMStatus(
     'VMStatus_CDI_IMPORTING',
-    StatusSimpleLabel.Importing,
+    StatusSimpleLabel.Copying,
```
suomiy, line 62: This will cause the filtering to break as it will no longer be possible to filter it by `Importing` filter.

frontend/packages/kubevirt-plugin/src/constants/status-constants.ts
```diff
@@ -3,6 +3,7 @@ export enum StatusSimpleLabel {
   Completed = 'Completed',
   Pending = 'Pending',
   Importing = 'Importing',
+  Copying = 'Copying disk image',
```
suomiy, line 6: please don't add it to StatusSimpleLabel since these should be shared acrros multiple enums. And `Copying disk image` is not generic enough. It should be included in `VMStatusSimpleLabel`.

Also please look at getSimpleLabel function. If you add a simple label, you should make sure it is working in places where this is used sorting/filtering (dashboard inventory and table filters)

frontend/packages/kubevirt-plugin/src/constants/vm/vm-status.ts
```diff
@@ -54,25 +54,22 @@ export class VMStatus extends StatusEnum<VMStatusSimpleLabel | StatusSimpleLabel
...
-      group: StatusGroup.CDI,
     },
   );
   static readonly CDI_IMPORT_PENDING = new VMStatus(
     'VMStatus_CDI_IMPORT_PENDING',
-    'Import pending',
+    'Disk image pending',
     {
       isImporting: true,
       isPending: true,
```
suomiy, line 72: just fyi, the groups made the sorting and filtering easier accross similar statuses

suomiy (changes requested):

spadgett:

PR #7281, opened 2020-11-20
Enable quickstart hints in quickstart sidepanel

nemesis09:
**Fixes:**
https://issues.redhat.com/browse/ODC-4964

**Description:**
- define a markdown syntax for links that trigger hints in the UI.
- enable the Spotlight component in the QuickStart side panel using the syntax

**Screens:**
enable-tour-highlight

**Test Coverage:**
Screenshot from 2020-12-02 17-18-43

**Browser Conformance:**
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
+++ b/frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
@@ -1,0 +1,33 @@
+import * as React from 'react';
+import { Spotlight } from '../spotlight';
+
+type MarkdownHighlightExtensionProps = {
+  docContext: HTMLDocument;
+};
+const MarkdownHighlightExtension: React.FC<MarkdownHighlightExtensionProps> = ({ docContext }) => {
+  const [selector, setSelector] = React.useState<string>(null);
+  React.useEffect(() => {
+    const elements = docContext.querySelectorAll('[data-highlight]');
+    let timeoutId: NodeJS.Timeout;
+    function startHighlight(e) {
+      const highlightId = e.target.getAttribute('data-highlight');
+      if (!highlightId) {
+        return;
+      }
+      setSelector(null);
+      timeoutId = setTimeout(() => {
+        setSelector(`[data-tour-id="${highlightId}"]`);
+      }, 0);
+    }
+    elements && elements.forEach((elm) => elm.addEventListener('click', startHighlight));
+    return () => {
+      clearTimeout(timeoutId);
+      elements && elements.forEach((elm) => elm.removeEventListener('click', startHighlight));
+    };
+  }, [docContext]);
+  if (!selector) {
+    return null;
+  }
+  return <Spotlight selector={selector} interactive />;
+};
+export default MarkdownHighlightExtension;
--- a/frontend/packages/console-shared/src/components/markdown-highlight-extension/__tests__/highlight-consts.spec.ts
+++ b/frontend/packages/console-shared/src/components/markdown-highlight-extension/__tests__/highlight-consts.spec.ts
@@ -1,0 +1,50 @@
+import {
+  INVALID_HIGHLIGHT_LABELS,
+  INVALID_HIGHLIGHT_LINKS,
+  INVALID_IDS,
+  VALID_HIGHLIGHT_LABELS,
+  VALID_HIGHLIGHT_LINKS,
+  VALID_IDS,
+} from './test-data';
+import { HIGHLIGHT_REGEXP, LINK_LABEL, SELECTOR_ID } from '../highlight-consts';
+
+const createTestLoop = (
+  shouldPass: boolean,
+  label: string,
+  values: string[],
+  regExpString: string,
+) => {
+  const regExp = new RegExp(`^${regExpString}$`);
+  describe(`${label} - ${shouldPass ? '' : 'not '}${regExpString}`, () => {
+    values.forEach((value: string) => {
+      it(`expect to successfully ${shouldPass ? '' : 'not '}match against "${value}"`, () => {
+        expect(regExp.test(value)).toBe(shouldPass);
+      });
+    });
+  });
+};
+
+const successfullyTestValues = (label: string, values: string[], regExpString: string) =>
+  createTestLoop(true, label, values, regExpString);
+const failTestValuesSuccessfully = (label: string, values: string[], regExpString: string) =>
+  createTestLoop(false, label, values, regExpString);
+
+describe('Convert Markdown To Highlight RegExp Tests', () => {
+  successfullyTestValues('Valid Labels', VALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  successfullyTestValues('Valid Ids', VALID_IDS, SELECTOR_ID);
+  successfullyTestValues('Valid Links', VALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  failTestValuesSuccessfully('Invalid Labels', INVALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  failTestValuesSuccessfully('Invalid Ids', INVALID_IDS, SELECTOR_ID);
+  failTestValuesSuccessfully('Invalid Links', INVALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  it('should get back 3 matches from the HIGHLIGHT_REGEXP for label, highlight and Id', () => {
+    const matches = HIGHLIGHT_REGEXP.exec('[label]{{highlight tour-perspective-switcher}}');
+    expect(matches).not.toBe(null);
+    expect(matches.length).toBe(4);
+    // index 0 is the whole string
+    expect(matches[1]).toBe('label');
+    expect(matches[2]).toBe('highlight');
+    expect(matches[3]).toBe('tour-perspective-switcher');
+  });
+});
--- a/frontend/packages/console-shared/src/components/markdown-highlight-extension/highlight-consts.ts
+++ b/frontend/packages/console-shared/src/components/markdown-highlight-extension/highlight-consts.ts
@@ -1,0 +1,9 @@
+export const LINK_LABEL = '[\\d\\w\\s-()$!]+';
+export const HIGHLIGHT_ACTIONS = ['highlight'];
+export const SELECTOR_ID = `[\\w-]+`;
+
+// [linkLabel]{{action id}}
+export const HIGHLIGHT_REGEXP = new RegExp(
+  `\\[(${LINK_LABEL})]{{(${HIGHLIGHT_ACTIONS.join('|')}) (${SELECTOR_ID})}}`,
+  'g',
+);
--- a/frontend/public/components/markdown-view.tsx
+++ b/frontend/public/components/markdown-view.tsx
@@ -5,12 +5,13 @@ import * as sanitizeHtml from 'sanitize-html';
 
 const tableTags = ['table', 'thead', 'tbody', 'tr', 'th', 'td'];
 
-const markdownConvert = (markdown) => {
+const markdownConvert = (markdown, extensions?: string[]) => {
   const unsafeHtml = new Converter({
     tables: true,
     openLinksInNewWindow: true,
     strikethrough: true,
     emoji: true,
+    extensions,
   }).makeHtml(markdown);
 
   return sanitizeHtml(unsafeHtml, {
@@ -33,10 +34,12 @@ const markdownConvert = (markdown) => {
       'li',
       'code',
       'pre',
+      'button',
       ...tableTags,
     ],
     allowedAttributes: {
-      a: ['href', 'target', 'rel'],
+      a: ['href', 'target', 'rel', 'data-*'],
+      button: ['class', 'data-*'],
     },
     allowedSchemes: ['http', 'https', 'mailto'],
     transformTags: {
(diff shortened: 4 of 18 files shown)


frontend/public/components/markdown-view.tsx
```diff
@@ -1,16 +1,63 @@
 import * as React from 'react';
 import * as _ from 'lodash-es';
-import { Converter } from 'showdown';
+import { Converter, extension, subParser } from 'showdown';
 import * as sanitizeHtml from 'sanitize-html';
+import { Spotlight } from '@console/shared';
 
 const tableTags = ['table', 'thead', 'tbody', 'tr', 'th', 'td'];
 
+extension('quickstart', () => {
```
andrewballantyne, line 9: Okay, send this as a DM, but I figured I'd post the same here:

Make a new component for this extension and the needed execution code:
```
import { extension } from 'showdown';
import { Spotlight } from '../spotlight';
extension('quickstart', () => {
  return [
    {
      type: 'lang',
      regex: /\[(\w+)]{{(highlight)\s([\w-]+)}}/g,
      replace: (text: string, linkLabel: string, linkType: string, linkId: string): string => {
        return `<button class="pf-c-button pf-m-inline pf-m-link" data-highlight="${linkId}">${linkLabel}</button>`;
      },
    },
  ];
});
type MarkdownHighlightExtensionProps = {
  docContext: HTMLDocument;
};
const MarkdownHighlightExtension: React.FC<MarkdownHighlightExtensionProps> = ({ docContext }) => {
  const [selector, setSelector] = React.useState<string>(null);
  React.useEffect(() => {
    // console.debug('highlight nodes', this.frame.contentDocument.querySelectorAll('a[data-highlight]'));
    const elements = docContext.querySelectorAll('[data-highlight]');
    function startHighlight(e) {
      const highlightId = e.target.getAttribute('data-highlight');
      if (!highlightId) {
        return;
      }
      setSelector(null);
      setTimeout(() => {
        setSelector(`[data-tour-id="${highlightId}"]`);
      }, 0);
    }
    elements.forEach((elm) => elm.addEventListener('click', startHighlight));
    return () => {
      elements.forEach((elm) => elm.removeEventListener('click', startHighlight));
    };
  }, [docContext]);
  if (!selector) {
    return null;
  }
  return <Spotlight selector={selector} animate />;
};
export default MarkdownHighlightExtension;
```

andrewballantyne: You'll also need to make other changes in this file:

```
diff --git frontend/public/components/markdown-view.tsx frontend/public/components/markdown-view.tsx
index 33a2c341e..3032ca74d 100644
--- frontend/public/components/markdown-view.tsx
+++ frontend/public/components/markdown-view.tsx
@@ -80,10 +35,12 @@ const markdownConvert = (markdown) => {
       'li',
       'code',
       'pre',
+      'button',
       ...tableTags,
     ],
     allowedAttributes: {
-      a: ['href', 'target', 'rel'],
+      a: ['href', 'target', 'rel', 'data-*'],
+      button: ['class', 'data-*'],
     },
     allowedSchemes: ['http', 'https', 'mailto'],
     transformTags: {
@@ -177,14 +134,20 @@ export class SyncMarkdownView extends React.Component<
       <body class="pf-m-redhat-font"><div style="overflow-y: auto;">${markdownConvert(
         content || 'Not available',
       )}</div></body>`;
+
     return (
-      <iframe
-        sandbox="allow-popups allow-same-origin"
-        srcDoc={contents}
-        style={{ border: '0px', display: 'block', width: '100%', height: '0' }}
-        ref={(r) => (this.frame = r)}
-        onLoad={() => this.updateDimensions()}
-      />
+      <>
+        <iframe
+          sandbox="allow-popups allow-same-origin"
+          srcDoc={contents}
+          style={{ border: '0px', display: 'block', width: '100%', height: '0' }}
+          ref={(r) => (this.frame = r)}
+          onLoad={() => this.updateDimensions()}
+        />
+        {this.frame?.contentDocument && (
+          <MarkdownHighlightExtension docContext={this.frame.contentDocument} />
+        )}
+      </>
     );
   }
 }
```

frontend/packages/console-app/src/components/quick-starts/data/serverless-application-quickstart.ts
```diff
@@ -26,6 +26,7 @@ export const serverlessApplicationQuickStart: QuickStart = {
 4. In the **Git Repo URL** field, type \`https://github.com/sclorg/django-ex.git\`.
 5. Under Resources, select **Knative Service**.
 6. At the end of the page, click **Create**.
+8. test extension: [link]{{highlight tour-search-nav}}
```
andrewballantyne, line 29: Not sure what our mechanism will be for the link... we may need to work with UX next week on that... But I'm currently thinking this example was akin to `[data-tour-id="tour-search-nav"]`... that probably isn't extensible enough, but like I said, we can circle back to that.

andrewballantyne (changes requested):
Also the quick-start data files are now YAMLs... so that's going to require you to rebase...

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
```diff
@@ -0,0 +1,43 @@
...
+  React.useEffect(() => {
+    const elements = docContext.querySelectorAll('[data-highlight]');
+    function startHighlight(e) {
+      const highlightId = e.target.getAttribute('data-highlight');
+      if (!highlightId) {
+        return;
+      }
+      setSelector(null);
+      setTimeout(() => {
+        setSelector(`[data-tour-id="${highlightId}"]`);
```
christianvogt, line 30: We need to settle on the correct data marker for what we allow to be targeted by the highlight.
It shouldn't be arbitrary.

We also need to sanitize the `highlightId` otherwise someone could create a selector that closes the first rule and opens another.

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [selector, setSelector] = React.useState<string>(null);
+  React.useEffect(() => {
+    const elements = docContext.querySelectorAll('[data-highlight]');
+    function startHighlight(e) {
+      const highlightId = e.target.getAttribute('data-highlight');
+      if (!highlightId) {
+        return;
+      }
+      setSelector(null);
+      setTimeout(() => {
```
christianvogt, line 29: Capture the handle and `clearTimeout` on effect dispose.

frontend/public/components/markdown-view.tsx
```diff
@@ -139,13 +143,18 @@ export class SyncMarkdownView extends React.Component<
...
+      <>
+        <iframe
+          sandbox="allow-popups allow-same-origin"
+          srcDoc={contents}
+          style={{ border: '0px', display: 'block', width: '100%', height: '0' }}
+          ref={(r) => (this.frame = r)}
+          onLoad={() => this.updateDimensions()}
+        />
+        {this.frame?.contentDocument && (
+          <MarkdownHighlightExtension docContext={this.frame.contentDocument} />
```
christianvogt, line 155: setup a `key` incase the `content` changes so that we unmount and re-mount the extensions

andrewballantyne:
Recommend closing this one and putting your 2nd commit on top of #7131 that way we can do the review of both and close off. Update the description accordingly.

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
nemesis09: done

frontend/public/components/markdown-view.tsx
nemesis09: done

frontend/packages/console-shared/src/components/markdown-highlight-extension/__tests__/highlight-consts.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('Convert Markdown To Highlight RegExp Tests', () => {
+  successfullyTestValues('Valid Labels', VALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  successfullyTestValues('Valid Ids', VALID_IDS, SELECTOR_ID);
+  successfullyTestValues('Valid Links', VALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  failTestValuesSuccessfully('Invalid Labels', INVALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  failTestValuesSuccessfully('Invalid Ids', INVALID_IDS, SELECTOR_ID);
+  failTestValuesSuccessfully('Invalid Links', INVALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  it('expect to get back 3 matches from the HIGHLIGHT_REGEXP', () => {
```
abhinandan13jan, line 41: nit: should get back

nemesis09: done

abhinandan13jan:
ran locally and it runs fine. /lgtm

abhinandan13jan:
Ran locally!! Looks good
/LGTM

frontend/packages/console-shared/src/components/markdown-highlight-extension/__tests__/highlight-consts.spec.ts
andrewballantyne: Hah! My live coding template slaps `expect` in there :P I didn't even read it, good catch Abhi!

frontend/packages/console-shared/src/components/markdown-highlight-extension/__tests__/highlight-consts.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('Convert Markdown To Highlight RegExp Tests', () => {
+  successfullyTestValues('Valid Labels', VALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  successfullyTestValues('Valid Ids', VALID_IDS, SELECTOR_ID);
+  successfullyTestValues('Valid Links', VALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  failTestValuesSuccessfully('Invalid Labels', INVALID_HIGHLIGHT_LABELS, LINK_LABEL);
+  failTestValuesSuccessfully('Invalid Ids', INVALID_IDS, SELECTOR_ID);
+  failTestValuesSuccessfully('Invalid Links', INVALID_HIGHLIGHT_LINKS, HIGHLIGHT_REGEXP.source);
+
+  it('should get back 3 matches from the HIGHLIGHT_REGEXP', () => {
```
andrewballantyne, line 41:
```suggestion
  it('should get back matches in a specific order from the HIGHLIGHT_REGEXP', () => {
```
I think we need to updating the wording here to clarify why this text exists... it's no longer getting back 3 (+ the full string).

andrewballantyne (approved):
LGTM

andrewballantyne:
/assign @christianvogt @rohitkrai03 

Please apply an approver-level review.

andrewballantyne:
Logged https://issues.redhat.com/browse/ODC-5201 to handle the missed requirement of allowing all the needed highlights -- will be followed up as a bug.

frontend/packages/console-shared/src/components/markdown-highlight-extension/highlight-consts.ts
```diff
@@ -0,0 +1,10 @@
+export const LINK_LABEL = '[\\d\\w\\s-()$!]+';
+export const HIGHLIGHT_ACTIONS = ['highlight'];
+const ID_PREFIX = '(qs-|tour-)';
```
christianvogt, line 3: I don't think we need to limit IDs to a specific prefix considering we only look at a specific data attribute.

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
christianvogt: Based on the regex of the extension, it will already ensure that the highlight ID is `[\w-]+`. Therefore at this time I don't think any further sanitation is necessary.
When we start to support more actions and if we make the regex more generic, we may need to revisit the sanitizing here.

frontend/public/components/markdown-view.tsx
```diff
@@ -11,6 +12,7 @@ const markdownConvert = (markdown) => {
     openLinksInNewWindow: true,
     strikethrough: true,
     emoji: true,
+    extensions: ['quickstart'],
```
christianvogt, line 15: Is there a way we can add the highlight extension only in the quick starts so that it's not possible to be used by other parts of the application inadvertently.
Perhaps compose a `QuickStartsMarkdownView` which contributes the extra extension.

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
```diff
@@ -0,0 +1,47 @@
...
+    elements.forEach((elm) => elm.addEventListener('click', startHighlight));
+    return () => {
+      clearTimeout(timeoutId);
+      elements.forEach((elm) => elm.removeEventListener('click', startHighlight));
+    };
+  }, [docContext]);
+  if (!selector) {
+    return null;
+  }
+  return <Spotlight selector={selector} interactive />;
```
christianvogt, line 45: Error is thrown if the selector has no valid matches.

christianvogt (commented):
I've encountered an issue where the highlight doesn't work.

- go to import yaml
- refresh the browser on the import yaml page
- open quick start
- click highlight link => nothing happens

frontend/packages/console-shared/src/components/markdown-highlight-extension/highlight-consts.ts
nemesis09: done

frontend/public/components/markdown-view.tsx
nemesis09: done

frontend/packages/console-shared/src/components/markdown-highlight-extension/MarkdownHighlightExtension.tsx
nemesis09: done

nemesis09:
> I've encountered an issue where the highlight doesn't work.
> 
>     * go to import yaml
> 
>     * refresh the browser on the import yaml page
> 
>     * open quick start
> 
>     * click highlight link => nothing happens

Updated the PR to address this issue

andrewballantyne (approved):
/lgtm

PR #7285, opened 2020-11-20
Bug 1900010: Fix i18n sentence fragment

rebeccaalpert:
Translation came back to us with questions about the sentence fragment. I updated it to use the Trans component instead so we can export the entire sentence for translation.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1900010.


spadgett (approved):
/lgtm

PR #7291, opened 2020-11-22
Convert Pinned Resources local storage to user settings

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5109

**Analysis / Root cause**: Convert Pinned Resources local storage to user settings

**Solution Description**: Convert Pinned Resources local storage to user settings

**Screen shots / Gifs for design review**: NO UI changes

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/usePinnedResources.ts
+++ b/frontend/packages/console-shared/src/hooks/usePinnedResources.ts
@@ -1,0 +1,42 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
+import { useSelector } from 'react-redux';
+import { getActivePerspective } from '@console/internal/reducers/ui';
+import { useUserSettingsCompatibility } from './useUserSettingsCompatibility';
+import { PINNED_RESOURCES_LOCAL_STORAGE_KEY } from '../constants';
+import { useMemo } from 'react';
+import { isPerspective, Perspective, useExtensions } from '@console/plugin-sdk';
+
+type PinnedResourcesType = {
+  [perspective: string]: string[];
+};
+
+const PINNED_RESOURCES_CONFIG_MAP_KEY = 'console.pinnedResources';
+
+export const usePinnedResources = (): [string[], (pinnedResources: string[]) => void, boolean] => {
+  const activePerspective = useSelector(getActivePerspective);
+  const perspectiveExtensions = useExtensions<Perspective>(isPerspective);
+  const activePerspectiveExtension = perspectiveExtensions.find(
+    (p) => p.properties.id === activePerspective,
+  );
+  const [pinnedResources, setPinnedResources, loaded] = useUserSettingsCompatibility<
+    PinnedResourcesType
+  >(
+    PINNED_RESOURCES_CONFIG_MAP_KEY,
+    PINNED_RESOURCES_LOCAL_STORAGE_KEY,
+    { [activePerspective]: activePerspectiveExtension.properties.defaultPins },
+    true,
+  );
+  const pins = useMemo(() => (loaded ? pinnedResources[activePerspective] : []), [
+    loaded,
+    pinnedResources,
+    activePerspective,
+  ]);
+  return [
+    pins,
+    (pr: string[]) => {
+      setPinnedResources((prevPR) => ({ ...prevPR, [activePerspective]: pr }));
+    },
+    loaded,
+  ];
+};
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -16,6 +16,17 @@ import {
   USER_SETTING_CONFIGMAP_NAMESPACE,
 } from '../utils/user-settings';
 
+const useCounterRef = (initialValue: number = 0): [boolean, () => void, () => void] => {
+  const counterRef = React.useRef<number>(initialValue);
+  const increment = React.useCallback(() => {
+    counterRef.current += 1;
+  }, []);
+  const decrement = React.useCallback(() => {
+    counterRef.current -= 1;
+  }, []);
+  return [counterRef.current !== initialValue, increment, decrement];
+};
+
 export const useUserSettings = <T>(
   key: string,
   defaultValue?: T,
@@ -23,7 +34,7 @@ export const useUserSettings = <T>(
 ): [T, React.Dispatch<React.SetStateAction<T>>, boolean] => {
   const defaultValueRef = React.useRef<T>(defaultValue);
   const keyRef = React.useRef<string>(key);
-  const isRequestPending = React.useRef<boolean>(false);
+  const [isRequestPending, increaseRequest, decreaseRequest] = useCounterRef();
   const userUid = useSelector(
     (state: RootState) => state.UI.get('user')?.metadata?.uid ?? 'kubeadmin',
   );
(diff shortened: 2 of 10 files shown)


frontend/public/components/search.tsx
```diff
@@ -147,13 +142,14 @@ const SearchPage_: React.FC<SearchProps & StateProps & DispatchProps> = (props)
   const pinToggle = (e: React.MouseEvent<HTMLElement>, resource: string) => {
     e.preventDefault();
     e.stopPropagation();
-
-    const index = props.pinnedResources.indexOf(resource);
-    if (index >= 0) {
-      confirmNavUnpinModal(resource, pinnedResources, props.onPinnedResourcesChange);
-      return;
+    if (pinnedResourcesLoaded) {
```
christianvogt, line 145: Unnecessary because `pinToggle` is only available to be called when `pinnedResourcesLoaded === true`

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -138,12 +145,38 @@ export const useUserSettings = <T>(
...
+      }
     }
-    // This effect should only be run on change of settings state.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [settings]);
+  }, [cfData, cfLoaded, sync]);
+
+  const callback = React.useCallback<React.Dispatch<React.SetStateAction<T>>>(
+    (action: React.SetStateAction<T>) => {
+      if (isRequestPending.current) return;
```
christianvogt, line 161: Why does it matter if there's a request pending? what harm is there in saving the next value?

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -138,12 +145,38 @@ export const useUserSettings = <T>(
...
+      const previousSettings = settingsRef.current;
+      const newState =
+        typeof action === 'function' ? (action as (prevState: T) => T)(previousSettings) : action;
+      setSettings(newState);
+      if (cfLoaded) {
+        isRequestPending.current = true;
+        updateConfigMap(cfData, keyRef.current, seralizeData(newState))
+          .then(() => {
+            isRequestPending.current = false;
```
christianvogt, line 170: Move to finally block if required.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -138,12 +145,38 @@ export const useUserSettings = <T>(
...
+        cfLoaded &&
+        seralizeData(settingsRef.current) !== cfData?.data?.[keyRef.current]
+      ) {
+        setSettings(deseralizeData(cfData?.data?.[keyRef.current]));
+      }
     }
-    // This effect should only be run on change of settings state.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [settings]);
+  }, [cfData, cfLoaded, sync]);
```
christianvogt, line 157: Would be good if we can find a way to achieve this with refs and useMemo instead of an effect which forces another render.

christianvogt (commented):
Tested this and when loaded, my saved pinned resources never loaded as the defaults from my localstorage settings.

frontend/packages/console-shared/src/hooks/useUserSettingsCompatibility.ts
```diff
@@ -21,5 +23,5 @@ export const useUserSettingsCompatibility = <T>(
     [loaded],
   );
 
-  return [settings, setSettings, loaded];
+  return [settings as T, setSettings, loaded];
```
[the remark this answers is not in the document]
invincibleJai: i fixed that in https://github.com/openshift/console/pull/7301 , we don't need this now

christianvogt:
> Tested this and when loaded, my saved pinned resources never loaded as the defaults from my localstorage settings.

This was due to me forgetting to clear the existing user settings config ;/

frontend/packages/console-shared/src/hooks/usePinnedResources.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const PINNED_RESOURCES_CONFIG_MAP_KEY = 'console.pinnedResources';
+
+export const usePinnedResources = (): [string[], (pinnedResources: string[]) => void, boolean] => {
+  const activePerspective = useSelector(getActivePerspective);
+  const [pinnedResources, setPinnedResources, loaded] = useUserSettingsCompatibility<
+    PinnedResourcesType
+  >(PINNED_RESOURCES_CONFIG_MAP_KEY, PINNED_RESOURCES_LOCAL_STORAGE_KEY, {}, true);
+  return [
+    loaded ? pinnedResources[activePerspective] : [],
```
jerolimov, line 20: Get an error here "Cannot read property 'admin' of undefined".

loaded is true and pinnedResources is undefined, maybe because it waits on the sync? This works locally as workaround:

```suggestion
    loaded && pinnedResources ? pinnedResources[activePerspective] : [],
```

jerolimov:
Migration works fine, but I could a NPE, see above.

@christianvogt @invincibleJai @sahil143  should we create another ticket to sync this between browsers??

frontend/packages/console-shared/src/hooks/usePinnedResources.ts
christianvogt: image

christianvogt: App won't load at all for me. white screens

frontend/public/components/search.tsx
christianvogt: @sahil143  ^^

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -64,13 +75,28 @@ export const useUserSettings = <T>(
...
+      /**
+       * if key doesn't exist in config map send patch request to add the key with default value
+       */
+      !fallbackLocalStorage &&
+      cfData &&
+      cfLoaded &&
+      !cfData.data?.hasOwnProperty(keyRef.current)
+    ) {
+      updateConfigMap(cfData, keyRef.current, seralizeData(defaultValueRef.current));
+      setSettings(defaultValueRef.current);
```
christianvogt: only if `defaultValueRef.current !== undefined`

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -80,39 +106,53 @@ export const useUserSettings = <T>(
...
+     */
+    if (
+      cfLoaded &&
+      cfData &&
+      !cfData.data?.hasOwnProperty(keyRef.current) &&
+      settings &&
+      !isRequestPending
+    ) {
+      return defaultValueRef.current;
+    }
```
christianvogt: Shouldn't this only happen if `sync == true` ?
And why does it matter if `settings` is truthy?

sahil143: We are checking for settings here because if `settings`is undefined then it means we haven't initialized the key and that should happen with the patch request in the `useEffect`. I guess we should specifically check for `settings !== undefined`.
> Shouldn't this only happen if sync == true

I'm not sure here because if user deleted key from config map that would be to reset the state and it made sense to return default value here regardless of sync.
Another question that comes to mind is should also make patch request with the default value or call `setSettings` with defaultValue.

invincibleJai:
/lgtm

Verified the changes , was able to add / remove pin and even tried in different tabs

PR #7297, opened 2020-11-23
Provide code snippet for default console catalog categories

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5080

**Analysis / Root cause**: 
To enable the admin to customize the categories in Developer Catalog in the `Console` CRD (group operator.openshift.io), we need a way to provide the default catalog categories.

**Solution Description**:
PR is based on #7196, you can focus on the second commit in this commit.

This PR adds a new snippet to the `Console` resources which provides a list of the default catalog categories.

To save the customized Developer Catalog you need to extend the CRD as well. See Test setup below.

This PR converts the internally saved "default categories" into the "simplified" CRD format. Instead of a map where the id matches the map key, and an entry with `field: 'tags'` and `values: <array of tags>`:

```yaml
spec:
  customization:
    developerCatalog:
      categories:
      - java:
        - id: java
          label: 'Java',
          field: 'tags'
          values: ['java']
```

the CRD defines the categories as arrays with just an array of `tags`:

```yaml
spec:
  customization:
    developerCatalog:
      categories:
        - id: java
          label: 'Java'
          tags: ['java']
```

@openshift/team-devconsole-ux 
This adds a new snippet for `Console` CRD (group operator.openshift.io) in the sidebar of the `cluster` configuration.

It adds the snippet incl. the complete "yaml path", "spec.customization.developerCatalog.categories" which makes it hopefully easier for the admin to keep the correct indentation.

image

image

Unchanged

**Test setup:**
1. The PR https://github.com/openshift/api/pull/763/files extends the `Console` CRD (group operator.openshift.io) with a new `developerCatalog.categories` area. To save the snippet you need add these fields to the CRD before saving it:

CRD URL example: http://localhost:9000/k8s/cluster/customresourcedefinitions/consoles.operator.openshift.io

Search for `customProductName`, add this yaml between `customProductName` and `documentationBaseURL`.

```yaml
                    developerCatalog:
                      description: >-
                        console allow the cluster admin to configure the shown
                        categories.
                      type: object
                      properties:
                        categories:
                          description: categories which are shown the in the devconsole.
                          type: array
                          items:
                            type: object
                            properties:
                              id:
                                description: >-
                                  identifier used in the URL to enable
                                  deeplinking in console
                                type: string
                              label:
                                description: category display label
                                type: string
                              subcategories:
                                description: a list of child categories
                                type: array
                                items:
                                  type: object
                                  properties:
                                    id:
                                      description: >-
                                        identifier used in the URL to enable
                                        deeplinking in console
                                      type: string
                                    label:
                                      description: category display label
                                      type: string
                                    tags:
                                      description: >-
                                        the list of item tags this category will
                                        match
                                      type: array
                                      items:
                                        type: string
                              tags:
                                description: the list of item tags this category will match
                                type: array
                                items:
                                  type: string
```

2. Open the `cluster` resource of the YAML and open the sidebar.

Cluster config example: http://localhost:9000/k8s/cluster/operator.openshift.io~v1~Console/cluster/yaml

3. Insert the default snippet and save the resource.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/sidebars/resource-sidebar-samples.tsx
+++ b/frontend/public/components/sidebars/resource-sidebar-samples.tsx
@@ -70,24 +70,32 @@ const PreviewYAML = ({ maxPreviewLines = 20, yaml }) => {
   );
 };
 
-const ResourceSidebarSnippet: React.FC<any> = ({ snippet, insertSnippetYaml }) => {
+interface ResourceSidebarSnippetProps {
+  snippet: Sample;
+  insertSnippetYaml(id: string, yaml: string, reference: string);
+}
+
+const ResourceSidebarSnippet: React.FC<ResourceSidebarSnippetProps> = ({
+  snippet,
+  insertSnippetYaml,
+}) => {
   const [yamlPreviewOpen, setYamlPreviewOpen] = React.useState(false);
   const toggleYamlPreview = () => setYamlPreviewOpen(!yamlPreviewOpen);
 
-  const { highlightText, title, id, yaml, targetResource, description } = snippet;
-  const reference = referenceFor(targetResource);
+  const { highlightText, title, id, yaml, lazyYaml, targetResource, description } = snippet;
+
+  const insertSnippet = () => {
+    const reference = referenceFor(targetResource);
+    insertSnippetYaml(id, typeof lazyYaml === 'function' ? lazyYaml() : yaml, reference);
+  };
+
   return (
     <li className="co-resource-sidebar-item">
       <h3 className="h4">
         <span className="text-uppercase">{highlightText}</span> {title}
       </h3>
       <p>{description}</p>
-      <Button
-        type="button"
-        variant="link"
-        isInline
-        onClick={() => insertSnippetYaml(id, yaml, reference)}
-      >
+      <Button type="button" variant="link" isInline onClick={insertSnippet}>
         <PasteIcon className="co-icon-space-r" />
         Insert Snippet
       </Button>
@@ -110,12 +118,20 @@ const ResourceSidebarSnippet: React.FC<any> = ({ snippet, insertSnippetYaml }) =
           </>
         )}
       </Button>
-      {yamlPreviewOpen && <PreviewYAML yaml={yaml} />}
+      {yamlPreviewOpen && <PreviewYAML yaml={typeof lazyYaml === 'function' ? lazyYaml() : yaml} />}
     </li>
   );
 };
 
-export const ResourceSidebarSnippets = ({ snippets, insertSnippetYaml }) => {
+interface ResourceSidebarSnippetsProps {
+  snippets: Sample[];
+  insertSnippetYaml(id: string, yaml: string, reference: string);
+}
+
+export const ResourceSidebarSnippets: React.FC<ResourceSidebarSnippetsProps> = ({
+  snippets,
+  insertSnippetYaml,
+}) => {
   return (
     <ul className="co-resource-sidebar-list" style={{ listStyle: 'none', paddingLeft: 0 }}>
       {_.map(snippets, (snippet) => (
--- a/frontend/packages/console-shared/src/utils/sample-utils.ts
+++ b/frontend/packages/console-shared/src/utils/sample-utils.ts
@@ -1,7 +1,13 @@
 import { Map as ImmutableMap } from 'immutable';
+import * as _ from 'lodash';
+import YAML from 'js-yaml';
+
+import { defaultCatalogCategories } from '@console/dev-console/src/components/catalog/utils/default-categories';
+
 import {
   BuildConfigModel,
   ClusterRoleModel,
+  ConsoleModel,
   ConsoleLinkModel,
   NetworkPolicyModel,
   ResourceQuotaModel,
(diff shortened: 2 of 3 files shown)


frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -228,6 +231,48 @@ const defaultSamples = ImmutableMap<GroupVersionKind, Sample[]>()
...
+          const yaml = {
+            spec: {
+              customization: {
+                developerCatalog: {
+                  categories,
+                },
+              },
+            },
+          };
+          return YAML.dump(yaml);
```
invincibleJai, line 271: can make use if `safeJSToYAML`

jerolimov: Don't see a reason why our generated structure should fail. But that could be also "famous last words"..

But I don't see a good fallback here, so I would prefer to keep it simple and just dump the data.

I could extract this into a utils function and write a unit test for this to ensure that it doesn't fail in the future. Okay with that?

invincibleJai: yeah as it's own json so should be good :)

frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -26,6 +28,7 @@ import {
   referenceFor,
   referenceForModel,
 } from '../../module/k8s';
+import { catalogCategories } from '../catalog/catalog-items';
```
christianvogt, line 31: I don't like this dependency. Please put the default categories in a const file.

frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -228,6 +231,48 @@ const defaultSamples = ImmutableMap<GroupVersionKind, Sample[]>()
...
+          const mapCategory = (category) => ({
+            id: category.id,
+            label: category.label,
+            tags: !category.field || category.field === 'tag' ? category.values : [],
+            subcategories: category.subcategories
+              ? Object.values(category.subcategories).map(mapSubcategory)
+              : [],
+          });
+
+          const categories = Object.values(catalogCategories).map(mapCategory);
```
christianvogt, line 260: Why bother with this conversion. Shouldn't the mapping be 1:1 between the default data structure and the yaml?

frontend/public/components/sidebars/resource-sidebar-samples.tsx
jerolimov: Code reviews 1, Auto import 0

Create catch, used the wrong import here, changed it to 

```tsx
import { defaultCatalogCategories } from '@console/dev-console/src/components/catalog/utils/default-categories';
```

frontend/public/components/sidebars/resource-sidebar-samples.tsx
jerolimov: Rebased this PR now on #7196 where I changed all `defaultCatalogCategories` to the same structure which is used by the CRD. After this I now removed all the mapping from this PR and just return this YAML snippet:

```tsx
          const yaml = {
            spec: {
              customization: {
                developerCatalog: {
                  categories: defaultCatalogCategories,
                },
              },
            },
          };
          return YAML.dump(yaml);
```

jerolimov:
@christianvogt @rohitkrai03 This PR is now based on #7196, you can focus on the second commit in this PR for "adding code snippet support". Thanks.

frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -228,6 +232,31 @@ const defaultSamples = ImmutableMap<GroupVersionKind, Sample[]>()
...
+        snippet: true,
+        lazyYaml: () => {
+          const yaml = {
+            spec: {
+              customization: {
+                developerCatalog: {
+                  categories: defaultCatalogCategories,
+                },
+              },
+            },
```
christianvogt, line 253: I think if the snippet should be `defaultCatalogCategories` instead of a full spec.
This would allow someone to insert just the categories themselves.
Perhaps the description should be used to help guide the user where the snippet belongs

frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -228,6 +232,31 @@ const defaultSamples = ImmutableMap<GroupVersionKind, Sample[]>()
...
     ],
+  )
+  .setIn(
+    [referenceForModel(ConsoleModel)],
+    [
+      {
+        title: 'Add catalog categories',
+        description:
+          'Provides a list of default categories which are shown in the Developer Catalog.',
+        id: 'default',
```
christianvogt, line 243: Perhaps `devcatalog-categories` is a better id. Although I don't know how these ids are used.

frontend/public/components/sidebars/resource-sidebar-samples.tsx
jerolimov: Changed this to 

```tsx
return YAML.dump(defaultCatalogCategories);
```

Because the editor keeps indentation fine.

frontend/public/components/sidebars/resource-sidebar-samples.tsx
jerolimov: Changed this to 

```tsx
return YAML.dump(defaultCatalogCategories);
```

Because the editor keeps indentation fine.

christianvogt:
snippet addition looks good
/lgtm
/approve

waiting on dependent pr

rohitkrai03 (approved):
/lgtm

PR #7298, opened 2020-11-23
add sync capability to useUserSeetings hook

sahil143:
This PR adds the sync capability to `useUserSettings` hook, This can be tested through https://github.com/openshift/console/pull/7291

This PR is part of the https://issues.redhat.com/browse/ODC-4753

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -52,11 +52,14 @@ const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: str
     ];
     try {
       await k8sPatch(ConfigMapModel, configMap, patch);
+      return Promise.resolve();
     } catch (err) {
       // eslint-disable-next-line no-console
       console.error(err);
+      return Promise.reject();
     }
   }
+  return Promise.resolve();
 };
 
 const deseralizeData = <T>(data: T) => {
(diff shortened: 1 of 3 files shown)


frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -52,11 +52,14 @@ const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: str
...
     try {
       await k8sPatch(ConfigMapModel, configMap, patch);
+      return Promise.resolve();
     } catch (err) {
       // eslint-disable-next-line no-console
       console.error(err);
+      return Promise.reject();
     }
   }
+  return Promise.resolve();
```
invincibleJai, line 62: i don't think we need to return here as well as we are already returning in try catch block and this is an async function so we only need to return the value or throw an error. There's no need to return a Promise

sahil143: the try catch is inside If block. So we need to return outside of the if block as well

invincibleJai: ah i should have had this condition out of `updateConfigMap` and in caller, my bad and can still avoid return explicit Promise

sahil143:
Closing this as all the changes are in PR https://github.com/openshift/console/pull/7291 which is being reviewed actively

PR #7301, opened 2020-11-23
adds fallback to localstorage for usersettings if 403 for configmap

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5082

**Analysis / Root cause**: 
useUserSettings doesn't support fallback to localStorage

**Solution Description**: 
adds support fallback for support for `useUserSettings`

Implementation can be tested with https://github.com/openshift/console/pull/7051

NA

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -4,78 +4,19 @@ import * as React from 'react';
 // @ts-ignore
 import { useSelector } from 'react-redux';
 import { RootState } from '@console/internal/redux';
-import { ConfigMapModel, ProjectRequestModel, ProjectModel } from '@console/internal/models';
-import {
-  K8sResourceKind,
-  k8sGet,
-  k8sCreate,
-  k8sPatch,
-  ConfigMapKind,
-} from '@console/internal/module/k8s';
+import { ConfigMapModel } from '@console/internal/models';
+import { K8sResourceKind } from '@console/internal/module/k8s';
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
-
-// can't create project with name prefix with 'openshift-*', once we have proxy need to update
-const USER_SETTING_CONFIGMAP_NAMESPACE = 'console-user-settings';
-
-// This won't be needed once we have proxy api
-const getProject = async () => {
-  try {
-    await k8sGet(ProjectModel, USER_SETTING_CONFIGMAP_NAMESPACE);
-  } catch {
-    await k8sCreate(ProjectRequestModel, {
-      metadata: {
-        name: USER_SETTING_CONFIGMAP_NAMESPACE,
-      },
-    });
-  }
-};
-
-const createConfigMap = async (configMapData: K8sResourceKind): Promise<boolean> => {
-  try {
-    await k8sCreate(ConfigMapModel, configMapData);
-    return true;
-  } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(err);
-    return false;
-  }
-};
-
-const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: string) => {
-  if (value !== configMap.data?.[key]) {
-    const patch = [
-      {
-        op: 'replace',
-        path: `/data/${key}`,
-        value,
-      },
-    ];
-    try {
-      await k8sPatch(ConfigMapModel, configMap, patch);
-    } catch (err) {
-      // eslint-disable-next-line no-console
-      console.error(err);
-    }
-  }
-};
-
-const deseralizeData = <T>(data: T) => {
-  if (typeof data !== 'string') {
-    return data;
-  }
-  try {
-    return JSON.parse(data);
-  } catch {
-    return data;
-  }
-};
-
-const seralizeData = <T>(data: T) => {
-  if (typeof data === 'string') {
-    return data;
-  }
-  return JSON.stringify(data);
-};
+import { useUserSettingsLocalStorage } from './useUserSettingsLocalStorage';
+import {
+  createConfigMap,
+  deseralizeData,
+  getProject,
+  createProject,
+  seralizeData,
+  updateConfigMap,
+  USER_SETTING_CONFIGMAP_NAMESPACE,
+} from '../utils/user-settings';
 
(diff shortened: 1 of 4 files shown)


frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,81 @@
...
+  const localStorageUpdated = React.useCallback(
+    (ev: StorageEvent) => {
+      if (ev.key === storageConfigNameRef.current) {
+        const lsConfigMapData = deseralizeData(localStorage.getItem(storageConfigNameRef.current));
+        if (
+          lsData !== undefined &&
+          lsConfigMapData?.[keyRef.current] &&
+          seralizeData(lsConfigMapData[keyRef.current]) !== seralizeData(lsData)
+        ) {
+          setLsData(lsConfigMapData[keyRef.current]);
```
christianvogt, line 39: Shouldn't this be deserialized?

frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,81 @@
...
+  watch: boolean = true,
+): [T, React.Dispatch<React.SetStateAction<T>>] => {
+  const keyRef = React.useRef(key);
+  const defaultValueRef = React.useRef(defaultValue);
+  const userUid = useSelector(
+    (state: RootState) => state.UI.get('user')?.metadata?.uid ?? 'kubeadmin',
+  );
+  const storageConfigNameRef = React.useRef(`${CONFIGMAP_LS_KEY}-${userUid}`);
+  const [lsData, setLsData] = React.useState(() => {
+    const valueInLocalStorage = deseralizeData(localStorage.getItem(storageConfigNameRef.current));
```
christianvogt, line 23: You could use `localStorage.hasItem` to check whether we fallback to the default or not

frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,81 @@
...
+    return valueInLocalStorage?.[keyRef.current]
+      ? valueInLocalStorage[keyRef.current]
+      : defaultValueRef.current;
+  });
+  const lsDataRef = React.useRef<T>(lsData);
```
christianvogt, line 28: This never gets updated but is passed along as the previous value each time.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -115,27 +65,34 @@ export const useUserSettings = <T>(
               [keyRef.current]: seralizeData(defaultValueRef.current),
             }),
           },
-        });
-        if (!cmCreated) {
-          setSettings(deseralizeData(defaultValueRef.current));
-          setLoaded(true);
-        }
+        })
+          .then(() => setFallbackLocalStorage(false))
```
christianvogt, line 69: fallback already starts as false

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -115,27 +65,34 @@ export const useUserSettings = <T>(
...
+          .catch((err) => {
+            if (err.response.status === 403) {
+              setFallbackLocalStorage(true);
+            } else {
+              setFallbackLocalStorage(false);
+              setSettings(deseralizeData(defaultValueRef.current));
+              setLoaded(true);
+            }
```
christianvogt, line 77: How will this work at all if we have any error creating the configmap or if the new backend returns an error?
I suppose that it should never error out if the service is working correctly.
Perhaps it shouldn't go to the loaded state if we have an error.
Thoughts?

frontend/packages/console-shared/src/hooks/useUserSettings.ts
christianvogt, line 69: we should be using `try / catch` if it's an async function

frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
invincibleJai: in case of localStorage we are seralizing and deseralizing complete data in Configmap key so din't need to do it

```
const lsConfigMapData = deseralizeData(localStorage.getItem(storageConfigNameRef.current));
```

frontend/packages/console-shared/src/hooks/useUserSettings.ts
invincibleJai: yeah ideally it should never hit this block , have added it if in considering if it errors our and not with forbidden(403) then pass loaded with default value

frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
invincibleJai: `hasItem` i din't see in API for localStorage so made a check against null

```
localStorage.getItem(storageConfigNameRef.current) !== null
```

sahil143 (commented):
Tested this functionality with PR #7051 . Works as expected. Looks good to me.

jerolimov (approved):
Tested is with some local changes to 'fake the 403' case. Works fine. :+1: 

/lgtm

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -99,43 +40,58 @@ export const useUserSettings = <T>(
...
+              }),
+            },
+          });
+        } catch (err) {
+          if (err?.response?.status === 403) {
+            setFallbackLocalStorage(true);
+          } else {
+            setSettings(defaultValueRef.current);
+            setLoaded(true);
+          }
```
jerolimov, line 76: All other cases of an error are just ignored a.t.m? Can you add here a console.warn / error as well?

invincibleJai: we are logging for all in utils like updateConfigmap, createProject, get Project will add there for createConfigMap as well.

PR #7302, opened 2020-11-23
CONSOLE-2024: Add popover for Unschedulable Pending, CrashLoopBackoff, ErrImagePull, ImagePullBackOff statuses

rhamilto:
* If pod is `Unschedulable`, add a popover to `Pending` status that includes the reason the pod is unschedulable on the pod list and pod details pages.
  Screen Shot 2020-11-23 at 11 34 49 AM
  Screen Shot 2020-11-23 at 11 40 10 AM
* If pod status is `CrashLoopBackoff`, add a popover to the `CrashLoopBackoff` status that includes the reason for the error on the pod list and pod details pages.
  Screen Shot 2020-12-08 at 11 09 29 AM
  Screen Shot 2020-12-08 at 11 09 36 AM
* If pod status is `ErrImagePull`, add a popover to the `ErrImagePull` status that includes the reason for the error on the pod list and pod details pages.
  Screen Shot 2020-11-24 at 11 12 52 AM
  Screen Shot 2020-11-24 at 11 12 58 AM
* If pod status is `ImagePullBackOff`, add a popover to the `ImagePullBackOff` status that includes the reason for the error on the pod list and pod details pages.
  Screen Shot 2020-11-24 at 8 54 41 AM
  Screen Shot 2020-11-24 at 8 54 52 AM

### DIFF
--- a/frontend/public/components/pod.tsx
+++ b/frontend/public/components/pod.tsx
@@ -9,6 +9,7 @@ import { useTranslation } from 'react-i18next';
 import i18next from 'i18next';
 import * as classNames from 'classnames';
 import * as _ from 'lodash-es';
+import { Button, Popover } from '@patternfly/react-core';
 import { Status, TableColumnsType } from '@console/shared';
 import { ByteDataTypes } from '@console/shared/src/graph-helper/data-utils';
 import {
@@ -352,7 +353,7 @@ const PodTableRow = connect<PodTableRowPropsFromState, null, PodTableRowProps>(p
             columns={columns}
             columnID={podColumnInfo.status.id}
           >
-            <Status status={phase} />
+            <PodStatus pod={pod} />
           </TableData>
           <TableData
             className={podColumnInfo.ready.classes}
@@ -571,7 +572,53 @@ const PodGraphs = requirePrometheus(({ pod }) => {
   );
 });
 
-export const PodStatus: React.FC<PodStatusProps> = ({ pod }) => <Status status={podPhase(pod)} />;
+const PodStatusPopover: React.FC<PodStatusPopoverProps> = ({
+  bodyContent,
+  headerContent,
+  status,
+}) => {
+  return (
+    <Popover headerContent={headerContent} bodyContent={bodyContent}>
+      <Button variant="link" isInline>
+        <Status status={status} />
+      </Button>
+    </Popover>
+  );
+};
+
+export const PodStatus: React.FC<PodStatusProps> = ({ pod }) => {
+  const status = podPhase(pod);
+  const unschedulableCondition = pod.status?.conditions?.find(
+    (condition) => condition.reason === 'Unschedulable' && condition.status === 'False',
+  );
+  const containerStatusStateWaiting = pod.status?.containerStatuses?.find(
+    (cs) => cs.state?.waiting,
+  );
+  const { t } = useTranslation();
+
+  if (status === 'Pending' && unschedulableCondition) {
+    return (
+      <PodStatusPopover
+        bodyContent={unschedulableCondition.message}
+        headerContent={t('workload~Pod unschedulable')}
+        status={status}
+      />
+    );
+  }
+  if (
+    (status === 'CrashLoopBackOff' || status === 'ErrImagePull' || status === 'ImagePullBackOff') &&
+    containerStatusStateWaiting
+  ) {
+    return (
+      <PodStatusPopover
+        bodyContent={containerStatusStateWaiting.state.waiting.message}
+        status={status}
+      />
+    );
+  }
+
+  return <Status status={status} />;
+};
 
 export const PodDetailsList: React.FC<PodDetailsListProps> = ({ pod }) => {
   const { t } = useTranslation();
@@ -881,6 +928,12 @@ type PodContainerTableProps = {
   heading: string;
   containers: ContainerSpec[];
   pod: PodKind;
+};
+
+type PodStatusPopoverProps = {
+  bodyContent: string;
+  headerContent?: string;
+  status: string;
 };
 
 type PodStatusProps = {
--- a/frontend/packages/console-shared/src/components/status/Status.tsx
+++ b/frontend/packages/console-shared/src/components/status/Status.tsx
@@ -29,7 +29,7 @@ export const Status: React.FC<StatusProps> = ({
       return <StatusIconAndText {...statusProps} icon={<HourglassStartIcon />} />;
 
     case 'Pending':
-      return <StatusIconAndText {...statusProps} icon={<HourglassHalfIcon />} />;
+      return <StatusIconAndText {...statusProps} icon={<HourglassHalfIcon color="inherit" />} />;
 
     case 'Planning':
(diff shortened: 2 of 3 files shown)


frontend/public/components/pod.tsx
```diff
@@ -551,7 +552,22 @@ const PodGraphs = requirePrometheus(({ pod }) => {
   );
 });
 
-export const PodStatus: React.FC<PodStatusProps> = ({ pod }) => <Status status={podPhase(pod)} />;
+export const PodStatus: React.FC<PodStatusProps> = ({ pod }) => {
+  const status = podPhase(pod);
+  const unschedulableCondition = pod.status?.conditions?.find(
+    (condition) => condition.reason === 'Unschedulable',
```
spadgett, line 558:
```suggestion
    (condition) => condition.reason === 'Unschedulable' && condition.status === 'True',
```

spadgett:
We might check for `pod.containerStatuses[n].state.waiting.message` and show it in a popover if available. That should handle the `CrashLoopBackOff` case, which is common.

frontend/public/components/pod.tsx
rhamilto: This actually breaks the check as `condition.type === PodScheduled`.  Think we want `(condition) => condition.reason === 'Unschedulable' && condition.status === 'False',`
Screen Shot 2020-11-23 at 3 25 25 PM

rhamilto:
> We might check for `pod.containerStatuses[n].state.waiting.message` and show it in a popover if available. That should handle the `CrashLoopBackOff` case, which is common.

We could also cover the `pod.containerStatuses[n].state.reason === ErrImagePull` case as well.

rhamilto:
> > We might check for `pod.containerStatuses[n].state.waiting.message` and show it in a popover if available. That should handle the `CrashLoopBackOff` case, which is common.
> 
> We could also cover the `pod.containerStatuses[n].state.waiting.reason === ErrImagePull` case as well.

Although retry happens so quickly, maybe this isn't necessary?

spadgett:
It should back off at some point where the message would be useful. We can probably skip checking `reason` and just show the message in the popover (although we might have to be careful if the status flaps quickly back and forth).

rhamilto:
> It should back off at some point where the message would be useful. We can probably skip checking `reason` and just show the message in the popover (although we might have to be careful if the status flaps quickly back and forth).

Conveniently, the popover switches as the same time as the status and maintains its open state. 
https://www.youtube.com/watch?v=mB7lb8uiNMc

rhamilto:
Updated with `ErrImagePull`, `ImagePullBackOff` popover status.  Any others we should add?

frontend/public/components/pod.tsx
```diff
@@ -551,7 +552,60 @@ const PodGraphs = requirePrometheus(({ pod }) => {
...
+  const status = podPhase(pod);
+  const unschedulableCondition = pod.status?.conditions?.find(
+    (condition) => condition.reason === 'Unschedulable' && condition.status === 'False',
+  );
+  const containerStatusImagePullBackOff = pod.status?.containerStatuses?.find(
+    (cs) => cs.state?.waiting?.reason === 'ImagePullBackOff',
+  );
+  const containerStatusErrImagePull = pod.status?.containerStatuses?.find(
+    (cs) => cs.state?.waiting?.reason === 'ErrImagePull',
+  );
```
spadgett, line 579: Do you see any weird behavior if you remove the `reason` check and just always show `state.waiting.message` if present?

rhamilto: Good call.  I was able to do that and DRY up the code significantly.

frontend/public/components/pod.tsx
```diff
@@ -551,7 +552,49 @@ const PodGraphs = requirePrometheus(({ pod }) => {
...
+  const containerStatusStateWaiting = pod.status?.containerStatuses?.find(
+    (cs) => cs.state?.waiting,
+  );
+
+  if (status === 'Pending' && unschedulableCondition) {
+    return (
+      <PodStatusPopover
+        bodyContent={unschedulableCondition.message}
+        headerContent={`${PodModel.label} unschedulable`}
```
rhamilto, line 582: Switched this to use the model.label so at least one of the English words is from a single source.  I assume we want all this content in English since the content from the API is English.

rhamilto:
@spadgett, comments addressed.  PTAL.

frontend/public/components/pod.tsx
```diff
@@ -571,7 +572,49 @@ const PodGraphs = requirePrometheus(({ pod }) => {
...
+        status={status}
+      />
+    );
+  }
+  if ((status === 'ErrImagePull' || status === 'ImagePullBackOff') && containerStatusStateWaiting) {
```
spadgett, line 607: I'm still not sure we need to check `status` at all. Should we simply show any `containerStatusStateWaiting` message if present?

frontend/public/components/pod.tsx
spadgett: In general, I'd err on the side of externalizing the string. FWIW, I think in the other API doc popovers we show the translated property name with the English API doc string.

frontend/public/components/pod.tsx
rhamilto: In testing, it looks like we end up with a popover with an empty message for `ContainerCreating` if we do that.  Not sure what other statuses may be affected...

rhamilto: And we end up with a popover for `Terminating` as well.  Not sure that is desired.
Screen Shot 2020-12-08 at 11 00 08 AM

rhamilto: We *do* want `CrashLoopBackoff` though.  My initial PR did not cover that.

rhamilto: Updated to include `CrashLoopBackoff` and externalize `unschedulable`.

frontend/public/components/pod.tsx
```diff
@@ -571,7 +572,53 @@ const PodGraphs = requirePrometheus(({ pod }) => {
...
+  const containerStatusStateWaiting = pod.status?.containerStatuses?.find(
+    (cs) => cs.state?.waiting,
+  );
+  const { t } = useTranslation();
+
+  if (status === 'Pending' && unschedulableCondition) {
+    return (
+      <PodStatusPopover
+        bodyContent={unschedulableCondition.message}
+        headerContent={`${PodModel.label} ${t('workload~unschedulable')}`}
```
spadgett, line 603: The kind should be a parameter to the message if we do it this way, but I'd just make it part of the message so it can get translated as needed. (Some k8s kinds we are translating depending on the language.)

```suggestion
        headerContent={t('workload~Pod unschedulable'}`}
```

rhamilto: Updated

spadgett (approved):
/lgtm
/hold for approvals

yapei:
tested locally with a cluster created via cluster-bot, no issues found

rhamilto:
@sferich888, PTAL and tag.

@jhadvig, mind re-tagging?  I had to rebase to remove the kubevirt test.

jhadvig (approved):
/lgtm

PR #7305, opened 2020-11-23
Bug 1888036: Only show latest version for each CSV and PackageManifest provided API.

TheRealJon:
- Add useK8sModels hook so that we can easily filter out APIs without a model. We need this for the
  "All Instances" page so that we can filter before we run a API requests.
- Update provideAPIsFor into providedAPIsForCSV and break out a shared filter
  function to share betweeen providedAPIsForCSV and providedAPIsForChannel
- Update providedAPIsForChannel function to use new filter function.
- Replace usage of providedAPIsFor().find() chains with helper functions to find a specific API by
  model or reference.
- Add new util function `providedAPIForReference` for consistency with `providedAPIForModel`
- Define constant for internal object annotation key and replace string literals throughout the olm
  package. Also update existing constants to match our UPPERCASE_CONSTANT convention
- Remove redundant util functions in ./packages/operator-lifecycle-manager/src/utils.ts so that the
  logic for filtering internal objects could be centralized in the 'providedAPIsForCSV' function.
- Update `getInternalObjects` function args. Now accepts an `annotations` argument, and uses shared
  `parseJSONAnnotation` util function.
- Update ProvidedAPIs page to account for built in filtering in providedAPIsForCSV function. Also,
  remove usage of 'modelFor' function since it only checks statically defined models. We need to use
  the new useK8sModels hook to make sure we get any models that have been added by operators or
  crds.
- Fix unit tests

### DIFF
--- a/frontend/packages/console-shared/src/utils/annotations.ts
+++ b/frontend/packages/console-shared/src/utils/annotations.ts
@@ -1,0 +1,17 @@
+import { ObjectMetadata } from '@console/internal/module/k8s';
+
+export const parseJSONAnnotation = (
+  annotations: ObjectMetadata['annotations'],
+  annotationKey: string,
+  onError?: (err: Error) => void,
+  defaultReturn?: any,
+): any => {
+  try {
+    return annotations?.[annotationKey] ? JSON.parse(annotations?.[annotationKey]) : defaultReturn;
+  } catch (e) {
+    onError?.(e);
+    // eslint-disable-next-line no-console
+    console.warn(`Could not parse annotation ${annotationKey} as JSON: `, e);
+    return defaultReturn;
+  }
+};
--- a/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
@@ -37,8 +37,6 @@ import {
   referenceForModel,
   referenceFor,
   groupVersionFor,
-  GroupVersionKind,
-  K8sKind,
   k8sKill,
   k8sPatch,
   k8sGet,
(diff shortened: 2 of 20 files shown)


frontend/packages/console-shared/src/utils/annotations.ts
```diff
@@ -0,0 +1,17 @@
...
+export const parseJSONAnnotation = (
+  annotations: ObjectMetadata['annotations'],
+  annotationKey: string,
+  onError?: (err: Error) => void,
+  defaultReturn?: any,
+): any => {
+  try {
+    return annotations?.[annotationKey] ? JSON.parse(annotations?.[annotationKey]) : defaultReturn;
+  } catch (e) {
+    onError && onError(e);
```
spadgett, line 12:
```suggestion
    onError?.(e);
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -708,14 +703,19 @@ export const MarkdownView = (props: {
...
+  const createRoute = React.useMemo(
+    () =>
+      csv
+        ? `/k8s/ns/${csv.metadata.namespace}/${ClusterServiceVersionModel.plural}/${csv.metadata.name}/${reference}/~new`
+        : null,
+    [csv, reference],
+  );
 
-  return (
+  return model ? (
```
spadgett, line 718: We're now purposely showing cards for resources that don't yet have a model. See #7393

spadgett (commented):
We should make sure we continue to show tiles and tabs for APIs that don't have a model. We assume that the models simply haven't refreshed yet.

cc @rhamilto

TheRealJon:
@spadgett @rhamilto Tiles should now be shown even for API's without a model, they just won't be creatable until the model is loaded. I've addressed feedback so far. Sorry for the size of this one. Just had to make sure that the cards for package manifests (non-installed operators) match the cards for CSVs (installed operators) and wanted to make sure that they were sharing logic. Before they were not.

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

PR #7307, opened 2020-11-24
Localize VM NICs and disks modals

pcbailey:
This PR updates the modals for VM NICs and disks. It does not update the options in the drop down menus. That will be completed in a separate PR.

Disks modal:
Selection_686

Network interfaces modal:
Selection_687


pcbailey:
@yaacov @rawagner Please took a look.

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -354,12 +348,19 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
   return (
     <div className="modal-content">
       <ModalTitle>
-        {isEditing ? EDIT : ADD} {type.toString()}
+        {isEditing ? t('kubevirt-plugin~Edit') : t('kubevirt-plugin~Add')}{' '}
+        {t('kubevirt-plugin~{{type}}', { type: type.toString() })}
```
yaacov, line 352: Notes to next pass:
this will not get translated unless we have comments with `// t('kubevirt-plugin~<all posible types... >` ... do we need it translated ?

pcbailey: You're right. I added these earlier on before I understood about the comments hack. I've been meaning to take them out and forgot. I'll remove them in the next pass.

frontend/packages/kubevirt-plugin/src/components/form/size-unit-form-row.tsx
```diff
@@ -33,41 +34,47 @@ export const SizeUnitFormRow: React.FC<SizeUnitFormRowProps> = ({
...
-          ))}
-        </FormSelect>
-      </SplitItem>
-    </Split>
-  </FormRow>
-);
+}) => {
+  const { t } = useTranslation();
+
+  const titleText = title || t('kubevirt-plugin~Size');
```
suomiy, line 40: @pcbailey @rawagner this breaks the wizard

wizard

PR #7318, opened 2020-11-24
Bug 1900022: Search Page - Top Name/Label filter is not applied to selected Pipeline resources

dtaylor113:
This PR fixes two issue with how the Pipeline plugin interacts with the console's search page:

1. Search page top Name/Label filter not being applied to Pipelines resource section
2. Pipeline resource section shows 2nd Name/Label filter.

## Before

image

## After

image

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx
@@ -14,7 +14,7 @@ interface PipelinesResourceListProps extends React.ComponentProps<typeof FireMan
 
 const PipelinesResourceList: React.FC<PipelinesResourceListProps> = (props) => {
   const { t } = useTranslation();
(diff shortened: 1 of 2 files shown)


andrewballantyne:
Ah I see, this is a failure of passing forward params in our list pages. 🤔 

Guess this has always been the case since the filters changes got added. Thanks @dtaylor113 ! I'll review this in more detail soon.

dtaylor113:
> Ah I see, this is a failure of passing forward params in our list pages. 
> 
> Guess this has always been the case since the filters changes got added. Thanks @dtaylor113 ! I'll review this in more detail soon.

Hi @andrewballantyne, note there is still an issue in this area: [Bug 1901207: Search Page - Pipeline resources table not immediately updated after Name filter applied or removed](https://bugzilla.redhat.com/show_bug.cgi?id=1901207).  
I'm not familiar enough with 'PipelineAugmentRuns[Wrapper]' to attempt a fix.  -thanks

PR #7319, opened 2020-11-24
Bug 1899200: - In Installed Operators page I cannot search for an Operator by it's name

cyril-ui-developer:
/assign @spadgett 

Screen Shot 2020-11-24 at 11 48 38 AM

### DIFF
--- a/frontend/public/components/factory/table-filters.ts
+++ b/frontend/public/components/factory/table-filters.ts
@@ -29,6 +29,9 @@ import { Alert, Rule, Silence } from '../monitoring/types';
 export const fuzzyCaseInsensitive = (a: string, b: string): boolean =>
   fuzzy(_.toLower(a), _.toLower(b));
 
+const clusterServiceVersionDisplayName = (csv: K8sResourceKind): string =>
+  csv?.spec?.displayName || csv?.metadata?.name;
+
 // TODO: Table filters are undocumented, stringly-typed, and non-obvious. We can change that.
 export const tableFilters: TableFilterMap = {
   name: (filter, obj) => fuzzyCaseInsensitive(filter, obj.metadata.name),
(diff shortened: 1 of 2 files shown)


frontend/public/components/factory/table-filters.ts
```diff
@@ -277,6 +278,12 @@ export const tableFilters: TableFilterMap = {
   'image-name': (str, imageManifestVuln) => fuzzyCaseInsensitive(str, imageManifestVuln.spec.image),
   vulnerability: (str, imageVulnerability) =>
     fuzzyCaseInsensitive(str, imageVulnerability.vulnerability.name),
+
+  // Filter installed operators by text match - name or displayName
+  'installed-operator': (str, installedOperator) => {
```
spadgett, line 283: I'd call this `cluster-service-version` since that's the resource type.

frontend/public/components/factory/table-filters.ts
```diff
@@ -277,6 +278,12 @@ export const tableFilters: TableFilterMap = {
   'image-name': (str, imageManifestVuln) => fuzzyCaseInsensitive(str, imageManifestVuln.spec.image),
   vulnerability: (str, imageVulnerability) =>
     fuzzyCaseInsensitive(str, imageVulnerability.vulnerability.name),
+
+  // Filter installed operators by text match - name or displayName
```
spadgett, line 282: The comment doesn't match the code. It looks like we don't search by name, only display name. (I'd probably err on the side of matching both.)

frontend/public/module/k8s/installed-operators.ts
```diff
@@ -0,0 +1,6 @@
+import * as _ from 'lodash-es';
+
+import { K8sResourceKind } from '../../module/k8s';
+
+export const installedOperatorDisplayName = (installedOperator: K8sResourceKind): string =>
```
spadgett, line 5: These are ClusterServiceVersion resources, so we should use that terminology in all names/filenames.

frontend/public/module/k8s/installed-operators.ts
```diff
@@ -0,0 +1,6 @@
...
+  _.get(installedOperator, 'spec.displayName') || _.get(installedOperator, 'metadata.name');
```
spadgett, line 6: Prefer optional chaining

```suggestion
  installedOperator?.spec?.displayName || installedOperator?.metadata?.name;
```

frontend/public/module/k8s/installed-operators.ts
```diff
@@ -0,0 +1,6 @@
+import * as _ from 'lodash-es';
```
spadgett, line 1: It's too bad this can't go in the OLM package, but I think it needs to be here since it's reference from table-filters.ts.

I might just inline this in table-filters since we don't want to encourage a lot of OLM code outside of the package, and it's only used in one place.

cyril-ui-developer: @spadgett Updated! Thanks

spadgett (approved):
/lgtm

PR #7322, opened 2020-11-24
Removing Protractor annotations test from e2e, crud, and all test suites due to flakyness

dtaylor113:
We are currently porting over annotation tests to Cypress:  https://issues.redhat.com/browse/CONSOLE-2478

### DIFF
--- a/frontend/integration-tests/tests/modal-annotations.scenario.ts
+++ b/frontend/integration-tests/tests/modal-annotations.scenario.ts
@@ -15,7 +15,7 @@ const Actions = {
   delete: 'delete',
 };
 
-describe('Modal Annotations', () => {
+xdescribe('Modal Annotations', () => {
   beforeAll(async () => {
     await browser.get(`${appHost}/k8s/ns/${testName}/configmaps`);
     await crudView.isLoaded();


frontend/integration-tests/protractor.conf.ts
```diff
@@ -59,7 +59,6 @@ const testSuites = {
     'tests/crud.scenario.ts',
     'tests/secrets.scenario.ts',
     'tests/filter.scenario.ts',
-    'tests/modal-annotations.scenario.ts',
```
spadgett, line 62: We should leave the scenario and use `xdescribe` so we see they're disabled in the log and JUnit report.

spadgett (changes requested):

frontend/integration-tests/protractor.conf.ts
rhamilto: The downside to that is `beforeAll` and `afterAll` will still run...

rhamilto: > The downside to that is `beforeAll` and `afterAll` will still run...

i should say, `base.scenario.ts` including `Create a test namespace` will still run.

spadgett (approved):
/lgtm

frontend/integration-tests/protractor.conf.ts
spadgett: I think that only runs once for the entire e2e suite?

rhamilto: it runs for each suite.  but it's fine.  dave's gonna remove these soon anyhow.

dtaylor113: yes, and I just checked a few annotation failures, they seem to happen during the actual test, not in before hooks:
```

dtaylor113:
/hold
Until "quay.io has a service degratation incident." is resolved.

PR #7323, opened 2020-11-24
Add quick-search and add functionality in topology

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5021
https://issues.redhat.com/browse/ODC-5023

**UXD** - https://docs.google.com/document/d/1V1QibdzA9HCru3EGKvfWHR8T0RMG5c99_rVa6Z6wrEs/edit

**Gif:**
Peek 2020-11-25 02-36

**Browser Conformance:**
- [x] Chrome

- [x] Firefox


divyanshiGupta:
@bgliwa01 Please take a look at this PR. Also when the catalogItems are taking time in getting loaded what should be the UX in that case? Should we show a loading spinner or something in the search bar?

divyanshiGupta:
> @bgliwa01 Please take a look at this PR. Also when its taking time in loading all catalog items to search from then what should be the UX in that case? Should we show a loading spinner or something in the search bar?

@bgliwa01 I have added a spinner in search bar for the above scenario. So the items are loaded only once each time the user navigates to topology and hence the spinner will only be shown in this case. Let me know what you think about it.
Peek 2020-11-26 01-30

frontend/packages/topology/src/components/quick-search/QuickSearchBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+        autoFocus={autoFocus}
+        value={searchTerm}
+      />
+      {!itemsLoaded && (
+        <InputGroupText>
+          <Spinner size="lg" />
+        </InputGroupText>
+      )}
+      {itemsLoaded && showNoResults && (
+        <InputGroupText className="odc-quick-search-bar__no-results">-- No Results</InputGroupText>
```
andrewballantyne, line 47: I18n

frontend/packages/topology/src/components/quick-search/QuickSearchButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+const QuickSearchButton: React.FC<QuickSearchButtonProps> = ({ onClick }) => {
+  const { t } = useTranslation();
+
+  return (
+    <Tooltip position="right" content={t('topology~Add to project')}>
+      <Button
+        className="odc-quick-search-button"
+        variant="plain"
+        onClick={onClick}
+        aria-label="Quick Search"
```
andrewballantyne, line 20: I18n

frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
```diff
@@ -0,0 +1,13 @@
+.odc-quick-search-content {
+  min-height: 390px;
+  max-height: 390px;
```
andrewballantyne, line 3: 🤔  is this not just `height: 390px` -- am I missing other styles that make use of `min/max` heights?

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.scss
```diff
@@ -0,0 +1,35 @@
+.odc-quick-search-details {
+  display: flex;
+  flex-direction: column;
+  padding: 6px 12px 0px 15px;
```
andrewballantyne, line 4: Hmmm, can we use PF styles here?

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.scss
```diff
@@ -0,0 +1,35 @@
...
+  overflow-y: auto;
+  overflow-wrap: break-word;
+  width: 98%;
```
andrewballantyne, line 7: Hmmm, 98% is fishy... what was this in regards to?

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -0,0 +1,46 @@
...
+  selectedItem: CatalogItem;
+}
+
+const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem }) => {
+  return (
+    <div className="odc-quick-search-details">
+      <Title headingLevel="h4">{selectedItem?.name}</Title>
+      {selectedItem?.provider && (
+        <span className="odc-quick-search-details__provider">
+          Provided by {selectedItem.provider}
```
andrewballantyne, line 24: I18n

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
```diff
@@ -0,0 +1,29 @@
...
+  &__item {
+    &--highlight {
+      background: var(--pf-global--active-color--200);
+    }
+    &:hover {
+      background: #cbe9f9;
+    }
+  }
+  &__item-row {
+    padding: 0px 8px;
```
andrewballantyne, line 11: PF styles?

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
```diff
@@ -0,0 +1,29 @@
+.odc-quick-search-list {
+  &__item {
+    &--highlight {
+      background: var(--pf-global--active-color--200);
+    }
+    &:hover {
+      background: #cbe9f9;
```
andrewballantyne, line 7: Hmmm, no PF colour?

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
```diff
@@ -0,0 +1,29 @@
...
+  &__item-icon {
+    width: 30px;
+  }
+  &__item-name {
+    font-size: var(--pf-global--FontSize--md);
+    font-weight: 400;
+    font-family: var(--pf-global--FontFamily--heading--sans-serif);
+  }
+  &__all-items-link {
+    padding: 4px 0px;
```
andrewballantyne, line 25: PF styles

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+  searchTerm,
+  namespace,
+  onSelectListItem,
+}) => {
+  const { t } = useTranslation();
+  const ref = React.useRef(null);
+
+  const openForm = (e: React.SyntheticEvent) => {
+    e.persist();
+    e.stopPropagation();
```
andrewballantyne, line 44: Not sure you need `stopPropagation` 🤔 

Who else is catching the double-click?

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+    e.preventDefault();
+    const targetItemUID = ref?.current?.props?.id;
+    const item = listItems.find((listItem) => listItem.uid === targetItemUID);
+    removeQueryArgument('catalogSearch');
+    history.push(item?.cta?.href);
```
andrewballantyne, line 49: Hmmm... if we don't get the href we wanted here... you're going to redirect a bit wildly against an unknown/falsey value. Do we have an error state from UX instead?

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  const goToCatalogPage = () => {
+    removeQueryArgument('catalogSearch');
+    history.push(`/catalog/ns/${namespace}?keyword=${searchTerm}`);
```
andrewballantyne, line 54: I think you'll have a potential for stale value here in `searchTerm` ... perhaps use `React.useCallback`?

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+          </DataListItemRow>
+        </DataListItem>
+      ))}
+      <DataListItem id="viewAll" onClick={goToCatalogPage}>
+        <DataListItemRow className="odc-quick-search-list__all-items-link">
+          <DataListItemCells
+            dataListCells={[
+              <DataListCell>
+                <Button variant="link">
+                  {t('topology~View all results for')} &quot;{searchTerm}&quot; ({totalItems})
```
andrewballantyne, line 106: I think translations should have the swap terms in them 🤔 

It's been a long time since I dealt with I18n, but I think some languages are structured in such a fashion that the full sentence structure could flip around on you and you'd lose order of variables.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -0,0 +1,150 @@
...
+    (value: string) => {
+      setSearchTerm(value);
+      setCatalogItems(value ? searchCatalog(value) : null);
+      setSelectedItemId('');
+      setSelectedItem(null);
+    },
+    [searchCatalog],
+  );
+
+  const onEscapePress = React.useCallback(() => {
```
andrewballantyne, line 56: Small nit... rename this to `onCancel`? I think tying it to the button is a coupling aspect, whereas the functionality of "escape" is really just to cancel, exit or close... so I think if we call it that, it has more flexibility and is not tied to a single usage of the action. 

Like consider if a button got added to close the modal -- could use the same function/action but wouldn't be escape.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -0,0 +1,150 @@
...
+        if (searchInput?.value) {
+          setQueryArgument('catalogSearch', searchInput.value);
+        } else {
+          removeQueryArgument('catalogSearch');
+        }
+        closeModal();
+      }
+    };
+
+    document.addEventListener('click', onOutsideClick);
```
andrewballantyne, line 115: Interesting... Odd the Modal from PF doesn't _just_ support this 🤔 

I wonder if there should be some separation of concerns here... this component is for the content of the modal but is doing some "modal as a whole" actions (like closing via outside/mask click & keyboard controls). 

I'm thinking:

- `QuickStartModal` handles mask/outside clicks
- `QuickStartModal` handles escape keyboard press
- `QuickStartModalBody` handles up/down for selection of items

Thoughts?

andrewballantyne (changes requested):

andrewballantyne:
The Results also sorta scroll... from the design (and epic AC), it says:
> As a developer, when typing in a search string that has a lot of matches, show me the top 5 best matches, and allow me to navigate to the catalog to view the entire set of matches

So I think we can make sure it does not scroll any.

Eg:
results-sorta-scroll

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.scss
andrewballantyne: I removed this and it appears to work fine... what was the intent? (is it browser conformance? I only tested on Chrome)

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+  selectedItemId,
+  searchTerm,
+  namespace,
+  onSelectListItem,
+}) => {
+  const { t } = useTranslation();
+  const ref = React.useRef(null);
+
+  const openForm = (e: React.SyntheticEvent) => {
+    e.persist();
```
andrewballantyne, line 43: Oooh, just noticed this... what's the goal here persisting the event?

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+    e.stopPropagation();
+    e.preventDefault();
+    const targetItemUID = ref?.current?.props?.id;
```
andrewballantyne, line 46:
```suggestion
    const targetItemUID = ref.current?.props?.id;
```
Pretty sure ref ALWAYS has `current` as it's the structure of the `useRef` 🤔

andrewballantyne (changes requested):
Noticed a couple more things...

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: I had to add this to stop the event to propagate to its children since if it does the target item (which in this case I always want it to be the `li` element since from there I get the UID of the item) changes based on where you are clicking inside the list item block.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: We do not have an error state for this as of now but for the time being what I can do is that navigate the user to the form page only when we have the href available.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: I am not sure if I understand this completely but you mean swap terms for the variables or the entire sentence?

andrewballantyne: @divyanshiGupta A crude example:

In English: "I will be 25 on Thursday 1234"
In German: "Ich werde am Donnerstag 1234 25 sein"

It does not always pan out that your numbers and context of the phrase are always going to be at the end of a phrase in other languages. So we need to inject the variables into the phrase in order to allow translators to re-arrange to meet proper grammatical structure for the given language.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
andrewballantyne: Hmmm, I wonder if fallback should be to route them just to the catalog as if they selected "all"? I know it's not ideal and would appear to be a bug... but I think it's better than `history.push(undefined)`

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
divyanshiGupta: Yeah initially I was going to structure the event handlers this way but if you see the events on Modal and ModalBody are not really separated from each other. 

Like when the user clicks outside we are supposed to close the modal and if there is a value in the search bar we have to save it and present it to the user next time they open the modal. So one action happens on the Modal and the other on ModalBody. 

Similarly in case of escape key press we are supposed to clear the text if there is any text in the search bar and if then the user again presses escape we have to close the modal.

Handling these events in the body itself made more sense to me as the search bar state is being handled in the body and can be easily updated accordingly. But I am open to suggestions if you have any.

andrewballantyne: 🤔  Valid thoughts... I didn't follow it all the way through and you did haha... Perhaps we leave it as is for now. Sounds like there is an avenue to doing this, but it may be a bit messy. I'm more sad PF doesn't support clicking on a mask to close the modal...

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
divyanshiGupta: No on hover I wanted a color lighter than what appears when we select an item but didnt find it in pf

andrewballantyne: Hmmm... @bgliwa01 for Monday, can you provide some feedback on the chosen colour?

frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
divyanshiGupta: So when I used just the height somehow the container was shrinking when there were less than 6 items, not sure if it was a browser issue but now it seems to work fine.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.scss
divyanshiGupta: So I have kept the inner containers width to be 98% because I wanted a space between the parent container and where the scrollbar appears for the inner container. If I remove the width the space wont be there anymore as shown in the UX doc. Any suggestions?
image

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: Hmm yeah makes sense. @bgliwa01 It will be really helpful if you can give some input on how do we handle this on Monday.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
divyanshiGupta: Yeah definitely it can be done in the modal itself but that would make the code really complex as we will have to gain reference for the search bar in the modal and then use/edit its value from there.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    <SplitItem>
+                      <Label>{item.type}</Label>
+                    </SplitItem>
+                    <SplitItem>
+                      <TextContent>
+                        <Text component={TextVariants.small}>{item.provider}</Text>
+                      </TextContent>
+                    </SplitItem>
+                  </Split>
+                </DataListCell>,
```
andrewballantyne, line 93: PF looks to need us to provide keys on each one... perhaps combine UID + a static string?

From the console:
image

From PF docs:
image

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+          })}
+          onDoubleClick={openForm}
+          ref={ref}
+        >
+          <DataListItemRow className="odc-quick-search-list__item-row">
+            <DataListItemCells
+              className="odc-quick-search-list__item-content"
+              dataListCells={[
+                <DataListCell isIcon>
+                  <img className="odc-quick-search-list__item-icon" src={item.icon?.url} alt="" />
```
andrewballantyne, line 79: What happens if we don't have an icon 🤔 

Do we have a fallback? 

I think we need to also try to do an alt text here -- @christianvogt thoughts? I'm thinking maybe like `"${displayName} icon"`... not great, but better than a confusing no-named image for screenreaders.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,117 @@
...
+                </DataListCell>,
+                <DataListCell style={{ paddingTop: '6px' }} width={2} wrapModifier="truncate">
```
andrewballantyne, line 81: Probably should be a PF spacing... you can use the PF variable here instead of adding a className...

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
andrewballantyne: You're going to have to help me understand this... I am not following 🙂 

`stopPropagation` prevents event bubbling from continuing. Events bubble from the touched item (text, icon, etc) up indefinitely until they hit the top of the document. If we say had this appear as a node in Topology, one could want to stop propagation on double clicks to avoid the Topology Graph from zooming... but I don't know what is happening here that we want to stop 🤔

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: I added this because initially I was using the event target to get the UID and it was showing a warning saying that the target will always be null as the event is reused so use persist. But now since I am using ref instead of event target I dont need this anymore. Removed it.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: Got it. Thanks!

andrewballantyne (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.scss
andrewballantyne: Interesting... yeah it does appear like that in the design. Padding is a better way to go here... `width` of a % can be misleading and less extensible. Really what you're looking for is the gap, not a percentage of the width. If we made the modal wide or dynamically scaled it, % means something misleading for what we want... padding is preferred here.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+        }}
+      >
+        {selectedItem?.cta?.label}
+      </Button>
+      <TextContent>
+        <Text component={TextVariants.p} className="odc-quick-search-details__description">
+          {selectedItem?.description}
+        </Text>
+      </TextContent>
+    </div>
```
andrewballantyne, line 45: This component kinda falls over if it gets partial / invalid data... This is not good. I think we may need an error state, or we just put an alert saying we had troubles reading the type. 

Right now, it's going to be awkwardly empty... with a Button "shell" that has no label.

andrewballantyne (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: Yeah you are absolutely right. But initially if I clicked on for example the label inside the list item, current of the ref was not `li` but that `label` and thus navigation was failing but when I added `stopPropagation` current was always `li` and it somehow worked. But now again when I removed it, everything seems to work fine :see_no_evil: If you see any issues with double click now please let me know.

andrewballantyne:
You have a failure in I18n... I think you need to run the command in the `package.json` to fix it.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
divyanshiGupta: I didnt came across that scenario yet but it seems like a possibility. We can have have state that says `Data not available for this item` or something similar. cc: @bgliwa01

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: It makes sense to add an alt text. Done

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem }) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className="odc-quick-search-details">
+      <Title headingLevel="h4">{selectedItem?.name}</Title>
+      {selectedItem?.provider && (
+        <span className="odc-quick-search-details__provider">
+          {t('topology~Provided by')} {selectedItem.provider}
```
andrewballantyne, line 27: Think you're looking at doing this: https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/catalog/CatalogTile.tsx#L23

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -0,0 +1,127 @@
...
+          ref={ref}
+        >
+          <DataListItemRow className="odc-quick-search-list__item-row">
+            <DataListItemCells
+              className="odc-quick-search-list__item-content"
+              dataListCells={[
+                <DataListCell isIcon key={`${item.uid}-icon`}>
+                  <img
+                    className="odc-quick-search-list__item-icon"
+                    src={item.icon?.url}
```
andrewballantyne, line 80: So we have still a situation here where this can resolve to nothing and cause a blank. Perhaps we should check with @bgliwa01 if she prefers a blank/error image (depending on browser) or if she has a default to use.

andrewballantyne (changes requested):
Going to do one pass through the UI again, but looking good!

Minor change still needed and a note for Brigid.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
christianvogt: I suggest you do this instead so you can avoid the lookup again and the null case for item not found:
`onDoubleClick={(e) => openForm(e, item)}`

Furthermore the spec `CatalogItem` requires `cta.href` to always be defined. Make sure you aren't adding optional chaining when it isn't required.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
christianvogt: Icons aren't required but we have a fallback in catalog. Perhaps you need to use the same util and logic the catalog uses: `catalog-utils.tsx` `getIconProps`

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+        onClick={() => {
+          removeQueryArgument('catalogSearch');
+          history.push(selectedItem?.cta?.href);
+        }}
+      >
+        {selectedItem?.cta?.label}
+      </Button>
+      <TextContent>
+        <Text component={TextVariants.p} className="odc-quick-search-details__description">
+          {selectedItem?.description}
```
christianvogt, line 42: I see this a lot. The props say that `selectedItem` is required yet there's optional chaining here.
```suggestion
          {selectedItem.description}
```

andrewballantyne: This was not resolved @divyanshiGupta

andrewballantyne:
Looks good to me, but I am confused what happened here: https://github.com/openshift/console/pull/7323#discussion_r531681587

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
divyanshiGupta: @andrewballantyne this is resolved as I have made selectedItem optional in the props. But now I think a better solution would be to add a check for it in the parent itself so that it always has a selectedItem.

divyanshiGupta: Done.

christianvogt:
good job @divyanshiGupta 
A few minor nits.

Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
image

Pressing the `enter` key doesn't choose my item. I don't see it in the UX doc so this is more a question for @bgliwa01 if we should implement this. In testing it was natural for me to use the arrow keys and want to press the enter key instead of having to click the button.

The view all results list item is a tabbable item that does nothing. But the inner contents is also tabbable.
image
image

IMO there should only be 3 tabbable items in the dialog: text input, view all results link, item action button. I don't think there's value in having the individual list items tabbable since we can navigate these via arrow key up and down.

Focus isn't trapped inside this dialog. Focus should loop back to the text input but instead leaves the dialog.

Some funky borders going on the list items when the screen gets small:
image

Arrow key down navigates down the list and loops back to the top while arrow key up doesn't loop to the bottom if on the first item. What's the UX expectation here @bgliwa01

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
christianvogt: You can rely on the API of `CatalogItem` to be populated correctly.

divyanshiGupta:
> Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
> image

@bgliwa01 what should be the UX here? I think `No results` should be shown at the end of the search bar so that the input can be of full width at all times. Like this: 
image

andrewballantyne:
> > Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
> > image
> 
> @bgliwa01 what should be the UX here? I think `No results` should be shown at the end of the search bar so that the input can be of full width at all times. Like this:
> image

Recommend we go with this. UX can alter things if needed, but I don't want anything to stall given this is the last week before FF.

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
bgliwa01: @divyanshiGupta can we follow the same hover treatment that is used in the topology list view? you might be doing this - it looks very similar.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
bgliwa01: Based on Christian's comment, do we need to worry about this use case? I don't think we have this scenario in the developer catalog so I'm not sure how it would happen here.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
bgliwa01: +1 to Christian...staying consistent with what the catalog has would be the ideal solution!

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
andrewballantyne: Do we need to worry about it from a user point of view? No, probably not... but it is a concern right now from a gap in the code. This is how bugs happen. It's probably an edge case, but I'd prefer not to see an empty card with a broken looking button if something goes off the rail.

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
andrewballantyne: @divyanshiGupta see if the colours match, report that status. If they do, then we are _good_ here... otherwise we should align to Topology (and provide a screenshot for Brigid to verify).

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
bgliwa01: I don't see an error state in the developer catalog (I could be missing it) but I'm confused why we would need one here

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
divyanshiGupta: @andrewballantyne we dont need an empty state since this comp will always have a `selectedItem` and if some info is missing from the item itself then we will not show that particular info. Did you came across any scenario where you saw an empty details section?

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
divyanshiGupta: Yes, @bgliwa01 this was resolved. :)

bgliwa01:
> > Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
> > image
> 
> @bgliwa01 what should be the UX here? I think `No results` should be shown at the end of the search bar so that the input can be of full width at all times. Like this:
> image

If the box can grow and be responsive to the length of text so that no results would be at the end of the text, that would be ideal. If not, keeping it at the end of the box is also acceptable

christianvogt:
I agree that the text input should grow in size and continue to push the `-- no results` as additional text is entered.
That being said. I'm with @andrewballantyne that we shouldn't try to make changes here if it's going to hold up the PR. We can work on improving this later.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
andrewballantyne: Sorry, looking at a stale snippet above... you removed all optional conditions. My comments are moot now. Thanks!

bgliwa01:
> good job @divyanshiGupta
> A few minor nits.
> 
> Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
> image
> 
> Pressing the `enter` key doesn't choose my item. I don't see it in the UX doc so this is more a question for @bgliwa01 if we should implement this. In testing it was natural for me to use the arrow keys and want to press the enter key instead of having to click the button.
> 
> The view all results list item is a tabbable item that does nothing. But the inner contents is also tabbable.
> image
> image
> 
> IMO there should only be 3 tabbable items in the dialog: text input, view all results link, item action button. I don't think there's value in having the individual list items tabbable since we can navigate these via arrow key up and down.
> 
> Focus isn't trapped inside this dialog. Focus should loop back to the text input but instead leaves the dialog.
> 
> Some funky borders going on the list items when the screen gets small:
> image
> 
> Arrow key down navigates down the list and loops back to the top while arrow key up doesn't loop to the bottom if on the first item. What's the UX expectation here @bgliwa01

+1 to having enter choose the item and 3 tabbable items in the dialog: text input, view all results link, item action button.

I will get back to you on the arrow keys in about an hour, I want to confirm with the UX team.

bgliwa01:
Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.

frontend/packages/topology/src/components/quick-search/QuickSearchList.scss
divyanshiGupta: List view is using the same color on hover as I have used for highlighting the item when it is selected.
Screenshot from 2020-11-30 20-38-59

andrewballantyne:
> Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.

Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?

divyanshiGupta:
> > good job @divyanshiGupta
> > A few minor nits.
> > Whenever I enter a long string, as soon as it doesn't match any items the text input becomes very small and cuts off the input even though there's lots of room available:
> > image
> > Pressing the `enter` key doesn't choose my item. I don't see it in the UX doc so this is more a question for @bgliwa01 if we should implement this. In testing it was natural for me to use the arrow keys and want to press the enter key instead of having to click the button.
> > The view all results list item is a tabbable item that does nothing. But the inner contents is also tabbable.
> > image
> > image
> > IMO there should only be 3 tabbable items in the dialog: text input, view all results link, item action button. I don't think there's value in having the individual list items tabbable since we can navigate these via arrow key up and down.
> > Focus isn't trapped inside this dialog. Focus should loop back to the text input but instead leaves the dialog.
> > Some funky borders going on the list items when the screen gets small:
> > image
> > Arrow key down navigates down the list and loops back to the top while arrow key up doesn't loop to the bottom if on the first item. What's the UX expectation here @bgliwa01
> 
> +1 to having enter choose the item and 3 tabbable items in the dialog: text input, view all results link, item action button.
> 
> I will get back to you on the arrow keys in about an hour, I want to confirm with the UX team.

@bgliwa01 yes I already added the changes for having only 3 tabbable items - text input, view all results link, item action button.
Working on adding the functionality for opening the form page when enter is pressed on an already selected item.

bgliwa01:
> > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> 
> Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?

Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.

andrewballantyne:
> > > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> > 
> > 
> > Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?
> 
> Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.

Interesting, that probably should have been caught in the design review stage. Patternfly Modals don't support no mask (I think -- I can't find an example) 🤔 

So although I do agree we should follow the agreed upon designs, we probably should bow to the feature-set of PF when they overlap use-cases. Let's continue this discussion post this PR, unless you think this is critical enough to say this feature is incomplete without that change @bgliwa01

bgliwa01:
> > > > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> > > 
> > > 
> > > Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?
> > 
> > 
> > Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.
> 
> Interesting, that probably should have been caught in the design review stage. Patternfly Modals don't support no mask (I think -- I can't find an example) 🤔
> 
> So although I do agree we should follow the agreed upon designs, we probably should bow to the feature-set of PF when they overlap use-cases. Let's continue this discussion post this PR, unless you think this is critical enough to say this feature is incomplete without that change @bgliwa01

Leaving this for post PR works for me! I will check in with PF as well.

andrewballantyne:
> @bgliwa01 yes I already added the changes for having only 3 tabbable items - text input, view all results link, item action button.
> Working on adding the functionality for opening the form page when enter is pressed on an already selected item.

@divyanshiGupta Also the enter doesn't work on the "show all"... so this leaves an interesting UX. If they are on "show all" and hit enter, do you do the selected one or the "show all" link? @bgliwa01

andrewballantyne (approved):

andrewballantyne:
I'm okay with the current state of things -- the enter key is the only concern that I think I have left. I recommend we maybe merge this PR, log another ticket listing the remaining items and at least that way we have the bulk of the feature merged.

christianvogt:
> > @bgliwa01 yes I already added the changes for having only 3 tabbable items - text input, view all results link, item action button.
> > Working on adding the functionality for opening the form page when enter is pressed on an already selected item.
> 
> @divyanshiGupta Also the enter doesn't work on the "show all"... so this leaves an interesting UX. If they are on "show all" and hit enter, do you do the selected one or the "show all" link? @bgliwa01

If the user tabs to `show all`, enter key should action on the `show all` link.
If the user is focused on the text input, enter key should action on the selected item.

christianvogt:
If we move away from using a modal and instead move to a floating div that is centered over top topology, we can achieve the desired UX.

frontend/packages/topology/src/components/quick-search/QuickSearchModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  closeModal,
+  searchCatalog,
+  allCatalogItemsLoaded,
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <Modal
+      variant={ModalVariant.medium}
+      aria-label={t('topology~Quick Search Modal')}
```
christianvogt, line 27:
```suggestion
      aria-label={t('topology~Quick Search')}
```

divyanshiGupta:
> > > > > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> > > > 
> > > > 
> > > > Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?
> > > 
> > > 
> > > Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.

@bgliwa01 Sorry, I dont understand this usecase. Even if the modal is in the middle of topology view and not the entire window then also the user wont be able to see what is behind the modal even if we remove the mask. The modal in design is also not stated as being movable to improve the view and also if the user tries to move the topology nodes then the modal will be closed because of close on outside click functionality.

bgliwa01:
> > > > > > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> > > > > 
> > > > > 
> > > > > Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?
> > > > 
> > > > 
> > > > Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.
> 
> @bgliwa01 Sorry, I dont understand this usecase. Even if the modal is in the middle of topology view and not the entire window then also the user wont be able to see what is behind the modal even if we remove the mask, even with mask the topology view is visible just not interact able. The modal in design is also not stated as being movable to improve the view and also if the user tries to move the topology nodes then the modal will be closed because of close on outside click functionality.

Ideally, in a coming release this modal will be able to be resizable and moveable...I don't think this was possible for this release based on the epic ACs and sizing.

frontend/packages/topology/src/components/quick-search/QuickSearchModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <Modal
+      variant={ModalVariant.medium}
+      aria-label={t('topology~Quick Search')}
+      isOpen={isOpen}
+      showClose={false}
+      position="top"
+      positionOffset="30%"
```
andrewballantyne, line 31: 🤔  Is this new... because that's probably going to cause wonky stuff on Tablet-sized devices when the nav section collapses.

divyanshiGupta:
> > > @bgliwa01 yes I already added the changes for having only 3 tabbable items - text input, view all results link, item action button.
> > > Working on adding the functionality for opening the form page when enter is pressed on an already selected item.
> > 
> > 
> > @divyanshiGupta Also the enter doesn't work on the "show all"... so this leaves an interesting UX. If they are on "show all" and hit enter, do you do the selected one or the "show all" link? @bgliwa01
> 
> If the user tabs to `show all`, enter key should action on the `show all` link.
> If the user is focused on the text input, enter key should action on the selected item.

Added the enter functionality
Peek 2020-11-30 22-56
cc: @bgliwa01 @christianvogt @andrewballantyne

frontend/packages/topology/src/components/quick-search/QuickSearchModal.tsx
divyanshiGupta: Are you talking about the `positionOffset`?

divyanshiGupta: I added this because according to UXD the modal is neither at the top nor in the middle that is why had to add an offset.

divyanshiGupta: Peek 2020-11-30 23-02
 This is how it looks when the nav collapses.

andrewballantyne: Oh, it's a top offset, not a left offset. Apologizes!

divyanshiGupta:
> > > > > > > Great work @divyanshiGupta!! One nit - the modal should be centered in the topology view screen rather than the entire screen including the side nav.
> > > > > > 
> > > > > > 
> > > > > > Interesting... modal is a modal though. @christianvogt Thoughts here? Do we have a modal mount location feature in Console?
> > > > > 
> > > > > 
> > > > > Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.
> > 
> > 
> > @bgliwa01 Sorry, I dont understand this usecase. Even if the modal is in the middle of topology view and not the entire window then also the user wont be able to see what is behind the modal even if we remove the mask, even with mask the topology view is visible just not interact able. The modal in design is also not stated as being movable to improve the view and also if the user tries to move the topology nodes then the modal will be closed because of close on outside click functionality.
> 
> Ideally, in a coming release this modal will be able to be resizable and moveable...I don't think this was possible for this release based on the epic ACs and sizing.

Then we can look at changing it accordingly in the upcoming release. WDYT @andrewballantyne @christianvogt @rohitkrai03

andrewballantyne:
> > > > > > Well also, in the design I didn't include the modal background because we wanted the user to still be able to see the topology view...I agree if we can only do the modal with the greyed out background then it should be where it is today, but this strays from the design.
> > > 
> > > 
> > > @bgliwa01 Sorry, I dont understand this usecase. Even if the modal is in the middle of topology view and not the entire window then also the user wont be able to see what is behind the modal even if we remove the mask, even with mask the topology view is visible just not interact able. The modal in design is also not stated as being movable to improve the view and also if the user tries to move the topology nodes then the modal will be closed because of close on outside click functionality.
> > 
> > 
> > Ideally, in a coming release this modal will be able to be resizable and moveable...I don't think this was possible for this release based on the epic ACs and sizing.
> 
> Then we can look at changing it accordingly in the upcoming release. WDYT @andrewballantyne @christianvogt @rohitkrai03

As it stands, I agree with @divyanshiGupta that the no-mask version isn't going to really help. Imo, it's going to look goofy if we allow them to pan the Topology view behind the "modal". I am in favour of waiting for a future release where we want to allow the modal to move/resize/temporary hide for the user to get a glimpse of their topology rather than half-implementing it now.

@bgliwa01 It's more-or-less your call on how we approach this. We will likely do it in another PR though as I would like to see this merged to get a large chunk of the feature merged asap.

andrewballantyne (approved):
/lgtm

Let's get this merging now and we can work to support anything that is still under discussion.

bgliwa01 (approved):
lgtm

bgliwa01:
Just confirmed with PF accessibility - when on the top item and you press the up arrow, it should loop back to the bottom, as well as when on the bottom item and you press the down arrow, it should loop back to the top.

andrewballantyne:
> Just confirmed with PF accessibility - when on the top item and you press the up arrow, it should loop back to the bottom, as well as when on the bottom item and you press the down arrow, it should loop back to the top.

@divyanshiGupta will log a ticket for us to address the 3 questions:
1. Looping up/down arrows
2. Centering the "Modal" in Topology
3. Dropping the Mask and supporting a non-modal per the designs

We'll merge this PR without those and swing back to them. That way at least the bulk of the feature is implemented and we can have more time and effort into the best look and feel for the user for 4.7.

cc @bgliwa01 / @christianvogt

christianvogt:
@andrewballantyne i'm good with followup fixes.

As far as the model background goes.I thought that it would work like mac spotlight. There's no mask however clicking anywhere will immediately dismiss the modal. @bgliwa01 can correct me if my understanding was incorrect.

PR #7325, opened 2020-11-25
UI crashes on opening topology

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5173

**Analysis / Root cause**: 
Duplicate resource requests were sent to firehose. Firehose uses lodash `intersection` to check for changes to the resource list. Since there were duplicates, the intersection resulted in a different length of in compared to the original resource length. Even though the two resource lists (previous and current) were the same. 

**Solution Description**: 

Added `===` comparison to firehose to speed up the check and overcome the intersection comparison issue. 

De-duped the resource list in ApplicationDropdown since multiple providers were interested in the same resources with the same prop key.

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

_note: there seems to be an unrelated issue where on page refresh, the application selector isn't being populated right away_


sahil143 (approved):
/lgtm

invincibleJai (approved):

PR #7326, opened 2020-11-25
convert Topology filters localStorage to userSettings

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5038

Removed `filter` and `appliedFilters` from redux/local storage and added in userSettings.

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

### DIFF
--- a/frontend/packages/topology/src/components/page/TopologyView.tsx
+++ b/frontend/packages/topology/src/components/page/TopologyView.tsx
@@ -13,7 +13,6 @@ import { useAddToProjectAccess } from '@console/dev-console/src/utils/useAddToPr
 import { getEventSourceStatus } from '@console/knative-plugin/src/topology/knative-topology-utils';
 import {
   CreateConnectionGetter,
-  DisplayFilters,
   GraphData,
   TopologyApplyDisplayOptions,
   TopologyDisplayFilterType,
@@ -27,16 +26,12 @@ import {
 } from '../../extensions/topology';
 import { getTopologySearchQuery, useAppliedDisplayFilters, useDisplayFilters } from '../../filters';
 import { updateModelFromFilters } from '../../data-transforms/updateModelFromFilters';
-import {
-  setSupportedTopologyFilters,
-  setSupportedTopologyKinds,
-  setTopologyFilters,
-} from '../../redux/action';
+import { setSupportedTopologyFilters, setSupportedTopologyKinds } from '../../redux/action';
 import Topology from '../graph-view/Topology';
 import TopologyListView from '../list-view/TopologyListView';
 import TopologyFilterBar from '../../filters/TopologyFilterBar';
 import { getTopologySideBar } from '../side-bar/TopologySideBar';
-
+import { FilterContext } from '../../filters/FilterProvider';
 import './TopologyView.scss';
 
 const FILTER_ACTIVE_CLASS = 'odc-m-filter-active';
@@ -48,7 +43,6 @@ interface StateProps {
 
 interface DispatchProps {
   onSelectTab?: (name: string) => void;
-  onFiltersChange: (filters: DisplayFilters) => void;
   onSupportedFiltersChange: (supportedFilterIds: string[]) => void;
   onSupportedKindsChange: (supportedKinds: { [key: string]: number }) => void;
 }
@@ -67,10 +61,10 @@ export const ConnectedTopologyView: React.FC<ComponentProps> = ({
   viewType,
   eventSourceEnabled,
   application,
-  onFiltersChange,
   onSupportedFiltersChange,
   onSupportedKindsChange,
 }) => {
+  const { setTopologyFilters: onFiltersChange } = React.useContext(FilterContext);
   const [filteredModel, setFilteredModel] = React.useState<Model>();
   const [selectedEntity, setSelectedEntity] = React.useState<GraphElement>(null);
   const [visualization, setVisualization] = React.useState<Visualization>();
@@ -282,9 +276,6 @@ const TopologyStateToProps = (state: RootState): StateProps => {
 
 const TopologyDispatchToProps = (dispatch): DispatchProps => ({
   onSelectTab: (name) => dispatch(selectOverviewDetailsTab(name)),
-  onFiltersChange: (filters: DisplayFilters) => {
-    dispatch(setTopologyFilters(filters));
-  },
   onSupportedFiltersChange: (supportedFilterIds: string[]) => {
     dispatch(setSupportedTopologyFilters(supportedFilterIds));
   },
--- a/frontend/packages/topology/src/filters/FilterProvider.tsx
+++ b/frontend/packages/topology/src/filters/FilterProvider.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react';
+import { useUserSettingsCompatibility } from '@console/shared';
+import { DEFAULT_TOPOLOGY_FILTERS } from './const';
+import { DisplayFilters } from '../topology-types';
+import { TOPOLOGY_DISPLAY_FILTERS_LOCAL_STORAGE_KEY } from '../redux/const';
+
+const TOPOLOGY_DISPLAY_FILTERS_USER_SETTINGS_KEY = `devconsole.topology.filters`;
+
+const getTopologyFilters = (appliedFilters: AppliedFilters) => {
+  const filters = [...DEFAULT_TOPOLOGY_FILTERS];
+  filters.forEach((filter) => {
+    if (appliedFilters[filter.id] !== undefined) {
+      filter.value = appliedFilters[filter.id];
+    }
+  });
+
+  return filters;
+};
+
+const getAppliedFilters = (filters: DisplayFilters): { [id: string]: boolean } => {
+  if (!filters?.length) {
+    return {};
+  }
+
+  return filters.reduce((acc, filter) => {
+    acc[filter.id] = filter.value;
+    return acc;
+  }, {});
+};
+
+type AppliedFilters = { [filterKey: string]: boolean };
+type SetTopologyFilters = (filters: DisplayFilters) => void;
+
+const useFilterContextValues = (): [
+  DisplayFilters,
+  AppliedFilters,
+  boolean,
+  SetTopologyFilters,
+] => {
+  const [appliedFilters, setAppliedFilters, appliedFiltersLoaded] = useUserSettingsCompatibility(
+    TOPOLOGY_DISPLAY_FILTERS_USER_SETTINGS_KEY,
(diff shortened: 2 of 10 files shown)


frontend/packages/topology/src/components/page/TopologyView.tsx
```diff
@@ -48,7 +43,7 @@ interface StateProps {
 
 interface DispatchProps {
   onSelectTab?: (name: string) => void;
-  onFiltersChange: (filters: DisplayFilters) => void;
+  // onFiltersChange: (filters: DisplayFilters) => void;
```
invincibleJai, line 46: can remove comment

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
+import { useUserSettings } from '@console/shared';
+import { getDefaultAppliedFilters } from '../redux/reducer';
+import { DEFAULT_TOPOLOGY_FILTERS } from './const';
+import { getAppliedFilters } from '../redux/action';
+import { DisplayFilters } from '../topology-types';
+
+const TOPOLOGY_DISPLAY_FILTERS_CONFIG_MAP_KEY = `devcosole.topology.filters`;
+
+const getTopologyFilters = (appliedFilters) => {
```
invincibleJai, line 10: nit: can we have type for `appliedFilters`

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  DisplayFilters,
+  AppliedFilters,
+  boolean,
+  SetTopologyFilters,
+] => {
+  const [appliedFilters, setAppliedFilters, appliedFiltersLoaded] = useUserSettings(
+    TOPOLOGY_DISPLAY_FILTERS_CONFIG_MAP_KEY,
+    getDefaultAppliedFilters(),
+  );
+  const [filtersLoaded, setFiltersLoaded] = React.useState<boolean>(false);
```
invincibleJai, line 34: Do we need this useState , can't we rely on `appliedFiltersLoaded`?

sahil143: Yes, this is needed because filters depend on the data which is coming from user settings(appliedFilters). so once the user settings is loaded we need to update the filters as well and notify the component that all of the data is loaded.

frontend/packages/topology/src/filters/useAppliedDisplayFilters.ts
```diff
@@ -1,13 +1,8 @@
...
-import { useSelector } from 'react-redux';
-import { RootState } from '@console/internal/redux';
-import { DisplayFilters } from '../topology-types';
-import { getAppliedTopologyFilters } from './filter-utils';
+import { useContext } from 'react';
+import { FilterContext } from './FilterProvider';
 
-const useAppliedDisplayFilters = (): DisplayFilters => {
-  return useSelector((state: RootState) => getAppliedTopologyFilters(state));
+const useAppliedDisplayFilters = (): { [filiterKey: string]: boolean } => {
```
jerolimov, line 4: Typo

```suggestion
const useAppliedDisplayFilters = (): { [filterKey: string]: boolean } => {
```

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
+import { useUserSettings } from '@console/shared';
+import { getDefaultAppliedFilters } from '../redux/reducer';
+import { DEFAULT_TOPOLOGY_FILTERS } from './const';
+import { getAppliedFilters } from '../redux/action';
+import { DisplayFilters } from '../topology-types';
+
+const TOPOLOGY_DISPLAY_FILTERS_CONFIG_MAP_KEY = `devcosole.topology.filters`;
```
jerolimov, line 8:
```suggestion
const TOPOLOGY_DISPLAY_FILTERS_USERSETTINGS_KEY = `devcosole.topology.filters`;
```

Wdyt?

jerolimov:
@sahil143 When having a `Deployment` and `DeploymentConfig` in my topology and filter for only Deployments and reload the page, the filter was automatically overriden.

I noticed that by adding a console.warn to `useUserSetting` hook before calling `updateConfigMap`:

When selecting Deployment it saves:
```
update devcosole.topology.filters
{
apps~v1~Deployment: true <===
expand-app-groups: true
expand-groups: true
helmGrouping: true
knativeServices: true
operatorGrouping: true
show-groups: true
show-labels: true
show-pod-count: true
}
```

After reload the page it sends two updates (without any click by my):

```
update devcosole.topology.filters
{
expand-app-groups: true
expand-groups: true
show-groups: true
show-labels: true
show-pod-count: true
}
```

```
update devcosole.topology.filters 
{
expand-app-groups: true
expand-groups: true
helmGrouping: true
knativeServices: true
operatorGrouping: true
show-groups: true
show-labels: true
show-pod-count: true
}
```

Any ideas? Did you need more info? We can make a call if you cant reproduce this.

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+type AppliedFilters = { [filterKey: string]: boolean };
+type SetTopologyFilters = (filters: DisplayFilters) => void;
+
+const useFilterContextValues = (): [
+  DisplayFilters,
+  AppliedFilters,
+  boolean,
+  SetTopologyFilters,
+] => {
+  const [appliedFilters, setAppliedFilters, appliedFiltersLoaded] = useUserSettings(
```
andrewballantyne, line 30: @invincibleJai Should this not use the fallback wrapper?

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+    TOPOLOGY_DISPLAY_FILTERS_CONFIG_MAP_KEY,
+    getDefaultAppliedFilters(),
```
andrewballantyne, line 32: @sahil143 I thought our goal here was to get rid of the reducer and obfuscate local storage... why are we using this as the defaults?

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const setTopologyFilters = React.useCallback(
+    (displayFilters: DisplayFilters) => {
+      setFilters(displayFilters);
+      setAppliedFilters(getAppliedFilters(displayFilters));
+    },
+    [setAppliedFilters],
+  );
+
+  return [filters, appliedFilters, appliedFilters && filtersLoaded, setTopologyFilters];
```
andrewballantyne, line 52: If `appliedFilters` are what we have stored in user settings... and `filters` are just `appliedFilters` with added defaults... why are we providing both back? Wouldn't we want just `filters`?

frontend/packages/topology/src/filters/TopologyFilterBar.tsx
```diff
@@ -34,36 +31,29 @@ import QuickSearchButton from './quick-search/QuickSearchButton';
 import { getNamespaceDashboardKialiLink } from '../utils/topology-utils';
 
 import './TopologyFilterBar.scss';
+import { FilterContext } from './FilterProvider';
```
andrewballantyne, line 34: Nit: sort

frontend/packages/topology/src/filters/useDisplayFilters.ts
```diff
@@ -1,14 +1,10 @@
-// FIXME upgrading redux types is causing many errors at this time
-// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-// @ts-ignore
-import { useSelector } from 'react-redux';
-import { RootState } from '@console/internal/redux';
 import { DisplayFilters } from '../topology-types';
-import { getTopologyFilters } from './filter-utils';
 import { useDeepCompareMemoize } from '@console/shared/src';
+import { FilterContext } from './FilterProvider';
+import { useContext } from 'react';
```
andrewballantyne, line 4: Nit: Sort

sahil143:
> @sahil143 When having a Deployment and DeploymentConfig in my topology and filter for only Deployments and reload the page, the filter was automatically overridden / dropped.

I was able to reproduce this on master as well. I think we can log a bug to handle this.

frontend/packages/topology/src/filters/FilterProvider.tsx
sahil143: Thanks for noticing this, used the wrapper `useUserSettingsCompatibility`.

frontend/packages/topology/src/filters/FilterProvider.tsx
sahil143: fixed.

frontend/packages/topology/src/filters/FilterProvider.tsx
sahil143: Because `appliedFilters` is also being used in component.

sahil143:
@andrewballantyne @jerolimov Addressed all comments. Please review again.

invincibleJai:
> > @sahil143 When having a Deployment and DeploymentConfig in my topology and filter for only Deployments and reload the page, the filter was automatically overridden / dropped.
> 
> I was able to reproduce this on master as well. I think we can log a bug to handle this.

Verified the changes all works as expected apart from what mentioned above (this is observed in master as well so can be taken as separate issue)

jerolimov:
@sahil143 @invincibleJai Yeah works "similar" to master, it was saved successfully and then override when loading the page.

I will raise a bug for this.

/lgtm

frontend/packages/topology/src/filters/FilterProvider.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import { useUserSettingsCompatibility } from '@console/shared';
+import { DEFAULT_TOPOLOGY_FILTERS } from './const';
+import { getAppliedFilters } from '../redux/action';
```
jerolimov, line 4: `getAppliedFilters` was only used in this Provider. Can we remove it from `redux/action.ts` since it looks like that we still use redux here?!

frontend/packages/topology/src/filters/FilterProvider.tsx
andrewballantyne: Hmmm... we probably should look to refactor that -- let's go with this for now since it's the status quo with the implementation. But it seems like an odd situation.

andrewballantyne (approved):

PR #7327, opened 2020-11-25
Update user settings hook to use bridge call to create ConfigMap

jerolimov:
**Fixes**: 
~~https://issues.redhat.com/browse/ODC-4755~~ https://issues.redhat.com/browse/ODC-5189

This PR is based on #7095 which must be merged first (the go part).

It keep this to make it also easier to test the frontend change. A review here should focus on the TS change.

**Analysis / Root cause**:

Currently you need create this namespace yourself, **see Test Setup below**. This namespace will be automatically created by the console-operator, see PR https://github.com/openshift/console-operator/pull/479

**Solution Description**: 
Because the user could not create a ConfigMap here, we create a new `ConfigMap`, `Role` and `RoleBinding` with the service account role over the bridge. This code provides 3 new endpoints to create, get and delete the user settings. (The second and third API calls primary for debugging / development.) The console will fetch the ConfigMap over the known k8sWatch API.

Irrelevant

Updated `user-settings.spec.ts` and `useUserSettings.spec.ts`:

```
 PASS  packages/console-shared/src/utils/__tests__/user-settings.spec.ts
  createConfigMap
    ✓ calls user settings api  (4ms)
  updateConfigMap
    ✓ calls user settings api  (1ms)
  deseralizeData
    ✓ does not convert null or undefined
    ✓ converts valid json to an object (1ms)
    ✓ return invalid json strings as string
  seralizeData
    ✓ does not convert strings
    ✓ converts objects

 PASS  packages/console-shared/src/hooks/__tests__/useUserSettings.spec.ts
  useUserSettings
    ✓ should create and update user settings if watcher could not find a ConfigMap (297ms)
    ✓ should return default value for an empty configmap after switching from loading to loaded (4ms)
    ✓ should return saved value for an known key after switching from loading to loaded (3ms)
    ✓ should return default value for an unknown key if data is already loaded (hook is used twice) (3ms)
    ✓ should return saved value for an known key if data is already loaded (hook is used twice) (2ms)
    ✓ should return latest user settings value after switching from loading to loaded (3ms)
    ✓ should return latest user settings value in loaded state (hook is used twice) (2ms)
    ✓ should trigger update user settings when setter was called (3ms)
    ✓ should fallback to localStorage if creation fails with 403 (must not call updateConfigMap) (4ms)

Test Suites: 2 passed, 2 total
Tests:       16 passed, 16 total
```

**Test setup:**
This PR require some additional RBAC rules.

1. If PR https://github.com/openshift/console-operator/pull/479 is not part of your cluster you need to create a namespace:

1a Create namespace:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
    openshift.io/node-selector: ""
```

Until https://github.com/openshift/console-operator/pull/486/files is merged and available on your cluster you need to add this as well:

1b Create Role:
```yaml
kind: Role
metadata:
  name: console-user-settings-admin-fixed
  namespace: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
rules:
- apiGroups:
  - rbac.authorization.k8s.io
  resources:
  - roles
  - rolebindings
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
  - patch
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
  - patch
```

1c Create RoleBinding:
```yaml
kind: RoleBinding
metadata:
  name: console-user-settings-admin-fixed
  namespace: openshift-console-user-settings
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
roleRef:
  kind: Role
  name: console-user-settings-admin-fixed
  apiGroup: rbac.authorization.k8s.io
subjects:
  - kind: ServiceAccount
    name: console
    namespace: openshift-console
```

2. You need the backend functions from https://github.com/openshift/console/pull/7095.

You can checkout 7095, build the backend with `./build-backend.sh` and switch back to this PR.

3. You need a hook to test PR. I used this `AddPage.tsx`:

```jsx
import { Helmet } from 'react-helmet';
import { useTranslation } from 'react-i18next';
import { match as RMatch } from 'react-router';
import { useUserSettings, useUserSettingsCompatibility } from '@console/shared';

import { Firehose } from '@console/internal/components/utils';
import ODCEmptyState from './EmptyState';
import NamespacedPage from './NamespacedPage';
import ProjectsExistWrapper from './ProjectsExistWrapper';
import CreateProjectListPage from './projects/CreateProjectListPage';

export interface AddPageProps {
  match: RMatch<{
    ns?: string;
  }>;
}

interface AddUserSettings {
  value: string;
}

const AddPage: React.FC<AddPageProps> = ({ match }) => {
  const namespace = match.params.ns;

  const [userSettings, setUserSettings] = useUserSettings<AddUserSettings>('a_user_settings');
  const [userSettingsComp, setUserSettingsComp] = useUserSettingsCompatibility<string>(
    'a_user_settings_comp',
    'lskey',
  );

  return (
    <>
      <Helmet>
        <title>{`+${t('devconsole~Add')}`}</title>
      </Helmet>
      <NamespacedPage>
        <h1>
          user settings:
          <pre>{JSON.stringify(userSettings)}</pre>
          <button type="button" onClick={() => setUserSettings({ value: 'AAAAA' })}>
            AAAAA
          </button>
          <button type="button" onClick={() => setUserSettings({ value: 'BBBBB' })}>
            BBBBB
          </button>
        </h1>
        <h1>
          user settings compatibility:
          <pre>{JSON.stringify(userSettingsComp)}</pre>
          <button type="button" onClick={() => setUserSettingsComp('AAAAA')}>
            AAAAA
          </button>
          <button type="button" onClick={() => setUserSettingsComp('BBBBB')}>
            BBBBB
          </button>
        </h1>
        <Firehose resources={[{ kind: 'Project', prop: 'projects', isList: true }]}>
          <ProjectsExistWrapper title={t('devconsole~Add')}>
            {namespace ? (
              <ODCEmptyState title={t('devconsole~Add')} />
            ) : (
              <CreateProjectListPage title={t('devconsole~Add')}>
                {t('devconsole~Select a project to start adding to it')}
              </CreateProjectListPage>
            )}
          </ProjectsExistWrapper>
        </Firehose>
      </NamespacedPage>
    </>
  );
};

export default AddPage;
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/utils/user-settings.ts
+++ b/frontend/packages/console-shared/src/utils/user-settings.ts
@@ -1,66 +1,50 @@
-import { ConfigMapModel, ProjectRequestModel, ProjectModel } from '@console/internal/models';
-import {
-  K8sResourceKind,
-  k8sGet,
-  k8sCreate,
-  k8sPatch,
-  ConfigMapKind,
-} from '@console/internal/module/k8s';
+import { coFetch } from '@console/internal/co-fetch';
+import { ConfigMapModel } from '@console/internal/models';
+import { ConfigMapKind, resourceURL } from '@console/internal/module/k8s';
 
-// can't create project with name prefix with 'openshift-*', once we have proxy need to update
-export const USER_SETTING_CONFIGMAP_NAMESPACE = 'console-user-settings';
+export const USER_SETTING_CONFIGMAP_NAMESPACE = 'openshift-console-user-settings';
 
-export const getProject = async (): Promise<boolean> => {
+export const createConfigMap = async (): Promise<ConfigMapKind> => {
   try {
-    await k8sGet(ProjectModel, USER_SETTING_CONFIGMAP_NAMESPACE);
-    return true;
+    const response = await coFetch('/api/console/user-settings', { method: 'POST' });
+    return response.json();
   } catch (err) {
     // eslint-disable-next-line no-console
     console.error(err);
-    return false;
+    throw err;
   }
 };
 
-export const createProject = async () => {
-  try {
-    await k8sCreate(ProjectRequestModel, {
-      metadata: {
-        name: USER_SETTING_CONFIGMAP_NAMESPACE,
-      },
-    });
-  } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(err);
-  }
-};
-
-export const createConfigMap = async (configMapData: K8sResourceKind): Promise<K8sResourceKind> => {
-  try {
-    const configMapDataResp = await k8sCreate(ConfigMapModel, configMapData);
-    return configMapDataResp;
-  } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(err);
-    const error: Error & { response?: object } = new Error(err.message);
-    error.response = err.response;
-    throw error;
-  }
-};
-
-export const updateConfigMap = async (configMap: ConfigMapKind, key: string, value: string) => {
-  const operation = configMap.data?.hasOwnProperty(key) ? 'replace' : 'add';
-  const patch = [
-    {
-      op: operation,
-      path: `/data/${key}`,
-      value,
+export const updateConfigMap = async (
+  configMap: ConfigMapKind,
+  key: string,
+  value: string,
+): Promise<ConfigMapKind> => {
+  const url = resourceURL(ConfigMapModel, {
+    ns: configMap.metadata.namespace,
+    name: configMap.metadata.name,
+  });
+  const patch = {
+    data: {
+      [key]: value,
     },
-  ];
+  };
+  // Use JSON Merge Patch instead of normal JSON Patch because JSON Patch calls
+  // fail if there is no data defined.
   try {
-    await k8sPatch(ConfigMapModel, configMap, patch);
+    const response = await coFetch(url, {
+      method: 'PATCH',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/merge-patch+json;charset=UTF-8',
+      },
+      body: JSON.stringify(patch),
+    });
+    return response.json();
   } catch (err) {
     // eslint-disable-next-line no-console
     console.error(err);
+    throw err;
   }
 };
 
--- a/frontend/__tests__/utils/hooks-utils.tsx
+++ b/frontend/__tests__/utils/hooks-utils.tsx
@@ -1,11 +1,31 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
 
-const TestHook: React.FC<{ callback: () => void }> = ({ callback }) => {
-  callback();
+const useRerender = () => {
+  const [, setState] = React.useState(0);
+  return () => setState((value) => value + 1);
+};
+
+type ResultRef = { current: any };
+type RerenderRef = { current?: () => void };
+
+interface TestComponentProps {
+  hook: () => void;
+  result: ResultRef;
+  rerenderRef: RerenderRef;
+}
+
+const TestHook: React.FC<TestComponentProps> = ({ hook, result, rerenderRef }) => {
+  result.current = hook();
+  rerenderRef.current = useRerender();
   return null;
 };
 
-export const testHook = (callback: () => void) => {
-  mount(<TestHook callback={callback} />);
+export const testHook = <T extends any>(hook: () => T) => {
+  // Inspired by https://github.com/testing-library/react-hooks-testing-library
+  const result = { current: undefined as T };
+  const rerenderRef: RerenderRef = {};
+  const rerender = () => rerenderRef.current();
+  mount(<TestHook hook={hook} result={result} rerenderRef={rerenderRef} />);
+  return { result, rerender };
 };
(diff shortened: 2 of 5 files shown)


jerolimov:
/unassign @rhamilto @jhadvig 
/assign @christianvogt @invincibleJai 
/hold
Because it depends on #7095

contrib/oc-set-user-settings.sh
```diff
@@ -0,0 +1,42 @@
...
+if ! which yq > /dev/null; then
+    echo "yq cli not found, please checkout https://github.com/mikefarah/yq#install"
+    exit 1
+fi
+
+if [ "$USER_NAME" = "kube:admin" ]; then
+    RESOURCE="$RESOURCE-kubeadmin"
+elif [ -n "$USER_UID" ]; then
+    RESOURCE="$RESOURCE-$USER_UID"
+else
```
jerolimov, line 33: Todo: Check uid first, then username

frontend/packages/console-shared/src/utils/user-settings.ts
```diff
@@ -47,22 +17,37 @@ export const createConfigMap = async (configMapData: K8sResourceKind): Promise<K
...
+    name: configMap.metadata.name,
+  });
+  const patch = {
+    data: {
+      [key]: value,
+    },
+  };
+  // Use JSON Merge Patch instead of normal JSON Patch because JSON Patch calls
+  // fail if there is no data defined.
+  const response = await coFetchJSON(url, 'PATCH', {
```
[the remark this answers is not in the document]
jerolimov: Unfortunately `coFetchJSON` doesn't return a response, it returns already the parsed content. I can switch to coFetch and handle status code checking and response.json() myself. I will try that.

jerolimov:
I'm not happy with the unit tests yet. I currently rework them, I don't expect a change in the hook.

christianvogt:
ran as kube admin:
```
▶ ./contrib/oc-get-user-settings.sh
Get ConfigMap "user-settings-<no value>" from namespace "openshift-console-user-settings" and save it to "user-settings.yaml"

Error from server (NotFound): configmaps "user-settings-<no value>" not found
```

christianvogt:
```
▶ oc whoami
kube:admin

▶ ./contrib/oc-get-user-settings.sh
Get ConfigMap "user-settings-<no value>" from namespace "openshift-console-user-settings" and save it to "user-settings.yaml"

Error from server (NotFound): configmaps "user-settings-<no value>" not found
```

christianvogt:
Settings saved and loaded as expected for kubeadmin user in local dev.

frontend/packages/console-shared/src/utils/user-settings.ts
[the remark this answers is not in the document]
jerolimov: Changed to `coFetch`, please take a look

jerolimov:
@invincibleJai Can you please double check the unit tests, hope you are fine with them.

@christianvogt Fixed the get/set user settings script as well. Works now fine with other users. Also added the option to fetch other user settings from other usernames.

```bash
➜  contrib/oc-get-user-settings.sh 
Get ConfigMap "user-settings-kubeadmin" from namespace "openshift-console-user-settings" and save it to "user-settings.yaml"
  ...

➜  contrib/oc-get-user-settings.sh developer
Get ConfigMap "user-settings-12341234-1234-1234-1234-123412341234" from namespace "openshift-console-user-settings" and save it to "user-settings.yaml"
  ...
```

jerolimov:
@invincibleJai Updated all tests, looks fine for me, wdyt? Did you see a missing case? Need minimal tweaks also in the hook.

invincibleJai:
> @invincibleJai Updated all tests, I'm happy now, wdyt? Did you see a missing case? Tests also require some minimal tweaks in the hook.

Thanks @jerolimov this looks good to me , looks like we covered all.

invincibleJai:
i build the backend as described and cluster am in has necessary role and rolebindings but getting 404 on `/api/console/user-settings`

any help what i missed?

invincibleJai:
/lgtm

user-settings-example.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ConfigMap
+data:
+  console.communityProvidersWarning: "true"
+  console.guidedTour: '{"dev":{"completed":true}}'
```
christianvogt, line 5: Do we need this sample?
What's the intended use?
Might be better to simply have a comment in the `useUserSettings` hook.

jerolimov: Removed the example

jerolimov:
Just removed the example

jerolimov:
And another rebase was required after #7051 was merged. Will create a new PR for the scripts.

PR #7153, opened 2020-11-08
Kms cluster encryption

a2batic:
Merged after: #7062 
screencapture-localhost-9000-k8s-ns-openshift-storage-clusterserviceversions-ocs-operator-v4-6-0-ocs-openshift-io-v1-StorageCluster-new-2020-11-18-17_17_46
Screenshot from 2020-11-18 17-18-01

PR #7330, opened 2020-11-25
KMS flow for storage class creation

a2batic:
Merge after: #7153 
Screenshot from 2020-12-03 21-12-30
Screenshot from 2020-12-03 21-12-18

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/kms-config/utils.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/kms-config/utils.tsx
@@ -1,12 +1,19 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import { K8sResourceKind, ConfigMapKind, SecretKind } from '@console/internal/module/k8s/types';
-import { k8sCreate } from '@console/internal/module/k8s/resource';
+import { k8sCreate, k8sPatch } from '@console/internal/module/k8s/resource';
 import { ConfigMapModel, SecretModel } from '@console/internal/models';
-import { CEPH_STORAGE_NAMESPACE, MODES, KMSConfigMapName, KMSSecretName } from '../../constants';
+import {
+  CEPH_STORAGE_NAMESPACE,
+  MODES,
+  KMSConfigMapName,
+  KMSSecretName,
+  KMSConfigMapCSIName,
+} from '../../constants';
 import { Action } from '../ocs-install/attached-devices/create-sc/state';
 import { InternalClusterAction } from '../ocs-install/internal-mode/reducer';
 import { KMSConfig, KMSConfigMap } from '../ocs-install/types';
+import { StorageClassClusterAction } from '../../utils/storage-pool';
 
 export const parseURL = (url: string) => {
   try {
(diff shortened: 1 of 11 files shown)


frontend/packages/ceph-storage-plugin/src/components/kms-config/utils.tsx
```diff
@@ -66,15 +140,39 @@ export const createKmsResources = (kms: KMSConfig) => {
...
+    //   apiVersion: ConfigMapModel.apiVersion,
+    //   kind: ConfigMapModel.kind,
+    //   data: {
+    //     [kms.name.value]: JSON.stringify(csiConfigData),
+    //   },
+    //   metadata: {
+    //     name: KMSConfigMapCSIName,
+    //     namespace: CEPH_STORAGE_NAMESPACE,
+    //   },
+    // };
```
cloudbehl, line 154: can you remove this or put a comment why this is required?

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -151,3 +188,267 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+                  <div className="ocs-install-kms__save-button">
+                    <ButtonBar errorMessage={errorMessage} inProgress={progress}>
+                      <ActionGroup>
+                        <Button variant="secondary" onClick={updateKMS} isDisabled={!validSave}>
+                          Save
+                        </Button>
+                        <Button
+                          variant="plain"
+                          onClick={() => {
+                            if (editKMS) {
```
cloudbehl, line 420: move to a function

cloudbehl (commented):
nit:

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -161,3 +189,264 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+        setErrorMessage('');
+      } catch (error) {
+        setErrorMessage(error.message);
+      } finally {
+        setInProgress(false);
+      }
+    } else {
+      setErrorMessage(`KMS service ${state.kms.name.value} already exist`);
+    }
+  };
```
afreen23, line 311: I think you also need to `setInProgress` initially ?
And make it false later.

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -161,3 +189,264 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+  );
+
+  const EncryptionCardLoading: React.FC = () => (
+    <Grid className="skeleton-box">
+      <GridItem span={12} className="skeleton-activity" />
+      <GridItem span={12} className="skeleton-activity" />
+      <GridItem span={12} className="skeleton-activity" />
+      <GridItem span={12} className="skeleton-activity" />
+    </Grid>
+  );
```
afreen23, line 287: Components dont need to be inside existing component.
These should be moved out

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -161,3 +189,264 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+    if (csiConfigMapLoaded && csiConfigMap) {
+      const serviceNames: string[] = Object.keys(csiConfigMap?.data);
+      const targetServiceName: string = serviceNames[serviceNames.length - 1];
+      const delimiterIndex: number = targetServiceName.indexOf('-');
+      onParamChange(targetServiceName.slice(delimiterIndex + 1));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [csiConfigMap]);
+
+  return <></>;
```
afreen23, line 451: I cannot understand the use of this component, if this returns nothing.

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -161,3 +189,264 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+        caCertFile: state.kms.caCertFile,
+        tls: kmsData.VAULT_TLS_SERVER_NAME,
+        clientCert: state.kms.clientCert ?? null,
+        clientCertFile: state.kms.clientCertFile,
+        clientKey: state.kms.caCert ?? null,
+        clientKeyFile: state.kms.clientKeyFile,
+        providerNamespace: kmsData.VAULT_NAMESPACE,
+        hasHandled: true,
+      };
+      setEncryptionDispatch(SCActionType.SET_KMS_ENCRYPTION, '', dispatch, kmsObj);
```
afreen23, line 244: you should be dispatching on click of an action ideally .
This will set the state whenever map is not empty -> which does not seems like ideal condition

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -161,3 +189,264 @@ export const PoolResourceComponent: React.FC<ProvisionerProps> = ({ onParamChang
...
+    state.kms.port.value,
+    state.kms.name.valid,
+    state.kms.address.valid,
+    state.kms.port.valid,
+  ]);
+
+  React.useEffect(() => {
+    onParamChange(checked.toString());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [checked]);
```
afreen23, line 271: This can be just inside `if`

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
a2batic: This is because we need to send a value without rendering anything....since plugin accepts a component only, its like this.

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
a2batic: The reason for doing this is to set initial state of component with respect to value of configMap, so as config map loads, we have to set the state.

PR #7188, opened 2020-11-11
Add VM Create to Dev console

glekner:

PR #7334, opened 2020-11-25
allow providers to override catalog items

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5177

**Analysis / Root cause**: 
A VM template provider needs the ability to override the default template behavior and ensure that duplicate items are not present in the catalog. This is because the default template flow is not sufficient for vm templates even though they are OpenShift templates.

**Solution Description**: 
Catalog item providers that want to override existing items must supply a catalog item with the same UID. The priority of the provider will determine which catalog item is presented to the user.

Added `priority?: number` to catalog item provider extension.

Contributed VM template named `test`

Before: The VM template appeared in the new type as well as the old template type. 2 catalog tiles were displayed each with different contents.
image

After: Only a single catalog item is present. The VM template has priority and therefore the standard template catalog item is removed.
image

**Test setup:**

Tested with https://github.com/openshift/console/pull/7188
Need to contribute a catalog item provider which supplies catalog items with the same item UID.

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

cc @matthewcarleton  @serenamarie125 @glekner @rohitkrai03


frontend/packages/dev-console/src/components/catalog/service/CatalogServiceProvider.tsx
```diff
@@ -34,23 +34,31 @@ const CatalogServiceProvider: React.FC<CatalogServiceProviderProps> = ({
...
+  const catalogItems = React.useMemo(() => {
+    if (!loaded) {
+      return [];
+    }
+    const itemMap = _.flatten(catalogProviderExtensions.map((e) => extItemsMap[e.uid])).reduce(
+      (acc, item) => {
+        acc[item.uid] = item;
+        return acc;
+      },
+      {} as { string: CatalogItem },
```
andrewballantyne, line 54:
```suggestion
      {} as { [key: string]: CatalogItem },
```

What you have says you'll have a key `string` equal to `CatalogItem`... which I'm sure is not what you're looking for :)

christianvogt: Updated.
Surprisingly it was not a TS error...

matthewcarleton (approved):
LGTM :)

andrewballantyne (approved):
/lgtm

PR #7335, opened 2020-11-25
Localize Virtualization nav name

pcbailey:
This PR localizes the Virtualization side navigation tab name.

Screenshot:
Selection_688

### DIFF
--- a/frontend/packages/kubevirt-plugin/locales/en/kubevirt-virtualization.json
+++ b/frontend/packages/kubevirt-plugin/locales/en/kubevirt-virtualization.json
@@ -1,0 +1,3 @@
+{
+  "Virtualization": "Virtualization"
+}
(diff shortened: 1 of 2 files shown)


pcbailey:
@yaacov Please review.

christianvogt:
@pcbailey I suggest holding off on this for another day. I will be posting a pr tomorrow to address the dev console nav localization. It uses the `%...%` pattern for extension localization. These keys get localized with a call to `useExtension`. However I didn't implement support for localizing nested extension props which is required by nav which I have fixed on my branch.

christianvogt:
see https://github.com/openshift/console/pull/7341

pcbailey:
@christianvogt Thanks for the heads up. I have updated this PR to use the `%...%` pattern.

PR #7339, opened 2020-11-26
Convert custom table column configuration from localStorage to user settings

debsmita1:
**Jira Story:**
https://issues.redhat.com/browse/ODC-5110

**Analysis / Root cause**: 
table column configuration stored in local storage should be persisted

**Solution Description**: 
used HOC `withUserSettingsCompatibility` to save configurations in config map

**Screenshots / Gifs for design review**: 
no UI changes

updated existing test case

**Test setup:**
NA

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/nodes/NodesPage.tsx
+++ b/frontend/packages/console-app/src/components/nodes/NodesPage.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 // FIXME upgrading redux types is causing many errors at this time
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore
-import { connect, useSelector } from 'react-redux';
+import { connect } from 'react-redux';
 import * as _ from 'lodash';
 import { sortable } from '@patternfly/react-table';
 import { useTranslation } from 'react-i18next';
@@ -13,6 +13,11 @@ import {
   getNodeRole,
   getLabels,
   getNodeMachineNameAndNamespace,
+  WithUserSettingsCompatibilityProps,
+  TableColumnsType,
+  withUserSettingsCompatibility,
+  COLUMN_MANAGEMENT_LOCAL_STORAGE_KEY,
+  COLUMN_MANAGEMENT_CONFIGMAP_KEY,
 } from '@console/shared';
 import { NodeModel, MachineModel } from '@console/internal/models';
 import { NodeKind, referenceForModel } from '@console/internal/module/k8s';
@@ -40,7 +45,6 @@ import { nodeStatus } from '../../status/node';
 import NodeRoles from './NodeRoles';
 import { menuActions } from './menu-actions';
 import NodeStatus from './NodeStatus';
-import { RootState } from '@console/internal/redux';
 import MarkAsSchedulablePopover from './popovers/MarkAsSchedulablePopover';
 
 // t('nodes~Name')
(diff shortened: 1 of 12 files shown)


frontend/packages/console-shared/src/hooks/useTableColumns.ts
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
```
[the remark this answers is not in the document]
debsmita1: removed it

frontend/packages/console-app/src/components/nodes/NodesPage.tsx
```diff
@@ -402,9 +401,12 @@ const NodesTable: React.FC<NodesTableProps> = React.memo((props) => {
     [],
   );
   const { t } = useTranslation();
+  const [tableColumns, , loaded] = useTableColumns(columnManagementID);
```
abhinandan13jan, line 404: I'm not sure, but do we need a blank `updateTableColumn` or can we return/alter it in the `useTableColumns` definition return statement

frontend/packages/console-shared/src/hooks/useTableColumns.ts
```diff
@@ -0,0 +1,29 @@
...
+  );
+  const updateTableColumn = React.useCallback(
+    (selectedColumns: Set<string>) => {
+      setTableColumns((prevState) => {
+        return { ...prevState, [id]: [...selectedColumns] };
+      });
+    },
+    [id, setTableColumns],
+  );
+  return [loaded ? tableColumns[id] : [], updateTableColumn, loaded];
```
abhinandan13jan, line 28: do we need to return `updateTableColumn` here?

abhinandan13jan, line 28: nit: loaded is checked twice. Once here and once inside every call in individual components. Can we merge these 2 while we check so that we don't have to hold the loaded value in the component. Then we could also get rid of  statements like
```const [state, , loaded] to const [state];```

**Comment by invincibleJai** [comment]
i tested below scenarios

- namespace table
- project table
- pods table
- node table

Overall it works good , at times i observed flicker when i go to project for example shows shows all columns then update based on what's in Configmap

Nov-30-2020 19-04-57

File: frontend/packages/console-shared/src/hooks/useTableColumns.ts [hunk 4]
```diff
@@ -0,0 +1,27 @@
...
+  );
+  const updateTableColumn = React.useCallback(
+    (selectedColumns: Set<string>) => {
+      setTableColumns((prevState) => {
+        return { ...prevState, [id]: [...selectedColumns] };
+      });
+    },
+    [id, setTableColumns],
+  );
+  return [loaded && tableColumns?.[id] ? tableColumns[id] : [], updateTableColumn];
```
christianvogt, line 26: Clients of this hook need to know if the data is still loading:
```suggestion
  const columns = React.useMemo(() => loaded && tableColumns?.[id] || [], [loaded, tableColumns, id]);
  return [columns, updateTableColumn, loaded];
```

frontend/packages/console-shared/src/hooks/useTableColumns.ts
```diff
@@ -0,0 +1,27 @@
...
+  [id: string]: string[];
+};
+
+export const COLUMN_MANAGEMENT_CONFIGMAP_KEY = 'console.tableColumns';
+
+export const useTableColumns = (id: string): [string[], (selectedColumns: Set<string>) => void] => {
+  const [tableColumns, setTableColumns, loaded] = useUserSettingsCompatibility<TableColumnsType>(
+    COLUMN_MANAGEMENT_CONFIGMAP_KEY,
+    COLUMN_MANAGEMENT_LOCAL_STORAGE_KEY,
+    {},
```
christianvogt, line 15: Can the default value be `undefined` ?

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -94,12 +98,36 @@ export const useUserSettings = <T>(
   }, [cfLoadError, cfLoaded, fallbackLocalStorage]);
 
   React.useEffect(() => {
-    if (cfData && cfLoaded && settings !== undefined) {
-      updateConfigMap(cfData, keyRef.current, seralizeData(settings));
+    if (sync && !isRequestPending.current) {
```
christianvogt, line 101: The changes in `useUserSettings` will conflict with @sahil143's PR. I suggest you base this PR on his work.

frontend/packages/console-shared/src/utils/user-settings.ts
```diff
@@ -58,11 +58,14 @@ export const updateConfigMap = async (configMap: ConfigMapKind, key: string, val
     ];
     try {
       await k8sPatch(ConfigMapModel, configMap, patch);
+      return Promise.resolve();
     } catch (err) {
       // eslint-disable-next-line no-console
       console.error(err);
+      return Promise.reject();
```
christianvogt, line 65: This is an async function. You should be throwing errors and returning values instead of returning promises.

frontend/public/components/namespace.jsx
```diff
@@ -534,127 +534,122 @@ const projectHeaderWithoutActions = () =>
...
-            className={namespaceColumnInfo.displayName.classes}
-            columns={columns}
-            columnID={namespaceColumnInfo.displayName.id}
-          >
-            <span className="co-break-word co-line-clamp">
-              {getDisplayName(project) || (
-                <span className="text-muted">{t('namespace~No display name')}</span>
-              )}
+  withTranslation()(({ obj: project, index, rowKey, style, customData = {}, metrics, t }) => {
+    const [tableColumns] = useTableColumns(projectColumnManagementID);
```
christianvogt, line 541: You need to check if the data has `loaded` otherwise this will flash the default columns first.
Same comment wherever this hook is used.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
debsmita1: these changes have been copied from Sahil's https://github.com/openshift/console/pull/7298 PR, which he has closed yesterday. I have updated my PR description that it depends on https://github.com/openshift/console/pull/7291.

frontend/packages/console-shared/src/utils/user-settings.ts
debsmita1: same as above

sahil143: I will fix this in my PR. 😄

sahil143 (changes requested):

frontend/packages/console-shared/src/hooks/useTableColumns.ts
```diff
@@ -0,0 +1,34 @@
...
+        return { ...prevState, [id]: [...selectedColumns] };
+      });
+    },
+    [id, setTableColumns],
+  );
+  const columns = React.useMemo(() => (loaded && tableColumns?.[id]) || [], [
+    loaded,
+    tableColumns,
+    id,
+  ]);
```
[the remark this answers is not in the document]
christianvogt: You want to always maintain stable identity checks on data structures. If you don't you risk triggering dep changes of other hooks.
Of course this is short lived while `!loaded`.
I guess an alternative is to make `[]` a constant instead of using memo.

frontend/public/components/namespace.jsx
```diff
@@ -264,20 +265,19 @@ const getNamespacesSelectedColumns = () => {
...
     const bytes = metrics?.memory?.[name];
     const cores = metrics?.cpu?.[name];
     const description = getDescription(ns);
     const labels = ns.metadata.labels;
-    const columns = new Set(
-      selectedColumns?.get(NamespacesColumnManagementID) || getNamespacesSelectedColumns(),
-    );
+    const [tableColumns, , loaded] = useTableColumns(NamespacesColumnManagementID);
+    const columns =
+      loaded && tableColumns?.length > 0 ? new Set(tableColumns) : getNamespacesSelectedColumns();
```
christianvogt, line 280: You cannot render the table until the settings have loaded otherwise there's potential for the table to flash the default columns.
Same comment for all usages in this PR.

frontend/packages/console-shared/src/hooks/useTableColumns.ts
```diff
@@ -0,0 +1,34 @@
...
+    {},
+    true,
+  );
+  const updateTableColumn = React.useCallback(
+    (selectedColumns: Set<string>) => {
+      setTableColumns((prevState) => {
+        return { ...prevState, [id]: [...selectedColumns] };
+      });
+    },
+    [id, setTableColumns],
```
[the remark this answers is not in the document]
christianvogt: Technically this is correct to have `setTableColumns` as a dependency in case it happens to change.

frontend/public/components/namespace.jsx
debsmita1: Done

invincibleJai:
/lgtm

This PR Depends on https://github.com/openshift/console/pull/7291 

Have verified the changes with pod , namespace, node table. Looks good !!

christianvogt:
/hold

waiting on https://github.com/openshift/console/pull/7291

christianvogt:
/approve

will need to rebase after dependent pr merges

karthikjeeyar:
Verified the PR, works fine
/lgtm

PR #7343, opened 2020-11-27
Bug 1901594: fix item-create unstability

yapei:
@jhadvig  Can we fix bug 1901594 in this way? 
[Bug 1901594 - Kubernetes resource CRUD operations.Kubernetes resource CRUD operations Pod "before all" hook for "creates the resource instance"](https://bugzilla.redhat.com/show_bug.cgi?id=1901594)

### DIFF
--- a/frontend/packages/integration-tests-cypress/views/list-page.ts
+++ b/frontend/packages/integration-tests-cypress/views/list-page.ts
@@ -17,6 +17,7 @@ export const listPage = {
       });
   },
   clickCreateYAMLbutton: () => {
+    cy.get(`[data-test="item-create"]`).should('be.visible');
     cy.byTestID('item-create').click({ force: true });
   },
   filter: {


frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -17,6 +17,7 @@ export const listPage = {
       });
   },
   clickCreateYAMLbutton: () => {
+    cy.get(`[data-test="item-create"]`, { timeout: 180 }).should('be.visible');
```
spadgett, line 20: Is there a reason not to use the default timeout?

spadgett (commented):
Thanks for the PR 👍

frontend/packages/integration-tests-cypress/views/list-page.ts
yapei: @spadgett  Good question, didn't take default timeout into consideration. default timeout should also be ok

frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -17,6 +17,7 @@ export const listPage = {
       });
   },
   clickCreateYAMLbutton: () => {
+    cy.get(`[data-test="item-create"]`).should('be.visible');
```
dtaylor113, line 20: This should be `cy.byTestID('item-create').should('be.visible');`

dtaylor113:
I've been looking into this as well (didn't see the BZ for it till now :-) ) , I'm unsure why it cannot find the Create button at times.  I'm suspicious of this recent [change](https://github.com/openshift/console/commit/4642ae2d34f63794a2ca8012bcedc7690bfee6ac#diff-a0e1ce23609c622f716567e38406636f98f8da1d681b5d32697cc12b14c38979) to  `data-test="item-create"` for consistency.  Did I miss something in this change?  I've inspected the pages in question and the Create button does have `data-test="item-create"` so still not sure what is causing this.

spadgett:
Closing in favor of #7365, which should fix the openshift-ocp-installer-console-aws-4.7 job.

/close

PR #7345, opened 2020-11-27
add pipeline metrics dashboard

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5127

**Analysis / Root cause**: 

Pipeline metrics dashboard for each pipeline, gives an overall stats about the pipeline runs.

**Solution Description**: 

Added metrics per pipeline:

- PipelineRun duration
- TaskRuns duration
- Number of PipelineRun per day
- Pipeline success/all ratio per day

Empty State 1: 
image

Empty State 2: (with no data points)
image

Metrics Dashboard:
pipeline_metrics_chart

Interactive Legend:

interactive-legend

success Ratio test:
image
PipelineMetrics:
image
PipelineRunDuration:
image
PipelineRunTaskRunGraph:
image
pipeline-metrics-utils.ts:
image

**Test setup:**

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/assign @andrewballantyne 
cc: @bgliwa01 @siamaksade @vdemeester


andrewballantyne (commented):
Three quick comments -- I'm sure you're working on them, but just wanted to make myself feel like I did a review 😜 

Based on your gif... 
- Your legends look like they don't adjust well for 3days... 
- Your tooltip on the TaskRun Duration charts need to be improved to show more details
- Your x-axis legends for Pipeline names looks rough with the screen size -- I wonder if we can increase the angle or something to help with that (might need UX for other ideas)

Great looking work though! Thanks so much @karthikjeeyar

karthikjeeyar:
Made improvements around the legends, tooltips etc.
image

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetrics.scss
```diff
@@ -0,0 +1,23 @@
...
+  &__body {
+    margin-top: 0;
+    background: var(--pf-global--palette--black-150);
+    height: 100%;
+    &-content {
+      margin-top: var(--pf-global--spacer--lg);
+    }
+  }
+  &__big-card {
+    width: 100%;
```
andrewballantyne, line 12: 🤔  not sure this is needed, should be 100% by default.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetricsRefreshDropdown.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const PipelineMetricsRefreshDropdown: React.FC<PipelineMetricsRefreshDropdownProps> = ({
+  interval,
+  setInterval,
+}) => {
+  const { t } = useTranslation();
+  return (
+    <div className="form-group pipeline-metrics-dashboard__dropdown-wrap">
+      <label className="pipeline-metrics-dashboard__dropdown-title">
```
andrewballantyne, line 17: Don't use classes that are not directly imported, makes the component dependant on another arbitrary component.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetricsTimeRangeDropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const [isOpen, setValue] = React.useState(false);
+  const toggleIsOpen = React.useCallback(() => setValue((v) => !v), []);
+  const setClosed = React.useCallback(() => setValue(false), []);
+  const onChange = React.useCallback((v: string) => setTimespan(parsePrometheusDuration(v)), [
+    setTimespan,
+  ]);
+  const { t } = useTranslation();
+  return (
+    <div className="form-group pipeline-metrics-dashboard__dropdown-wrap">
+      <label className="pipeline-metrics-dashboard__dropdown-title">
```
andrewballantyne, line 29: Same concern here, import scss to use classes

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunDurationGraph.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const averageDuration = finalArray.totalDuration
+    ? formatDuration((finalArray.totalDuration * 1000) / finalArray.duration.length)
+    : 0;
+
+  return (
+    <div className="row pipeline-double-chart-container">
+      <div className="col-lg-3 pipeline-run-average-container">
+        <div className="pipeline-run-average-container__duration">{`${averageDuration}`}</div>
+        <div className="pipeline-run-average-container__duration-desc">
+          Average Pipeline Run Duration
```
andrewballantyne, line 58: I18n

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunDurationGraph.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </div>
+      </div>
+      <div className="col-lg-9">
+        <LineChart
+          ariaDesc="Pipeline Run Duration"
+          ariaTitle="Pipeline Run Duration"
```
andrewballantyne, line 64: I believe these need to be translated as well for consistency with screen readers + visible text.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunDurationGraph.tsx
```diff
@@ -0,0 +1,84 @@
...
+    },
+    { totalDuration: 0, duration: [] },
+  );
+  const averageDuration = finalArray.totalDuration
+    ? formatDuration((finalArray.totalDuration * 1000) / finalArray.duration.length)
+    : 0;
+
+  return (
+    <div className="row pipeline-double-chart-container">
+      <div className="col-lg-3 pipeline-run-average-container">
```
andrewballantyne, line 55: Avoid bootstrap grids

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
```diff
@@ -0,0 +1,81 @@
...
+      return r.metric.pipelinerun;
+    }) ?? [];
+  const finalObj: { [x: string]: { x: string; y: number }[] } = pRuns.reduce((acc, prun) => {
+    if (!prun) return acc;
+
+    const obj = prun[prun.length - 1];
+    if (obj?.metric?.task && !acc[obj.metric.task]) {
+      acc[obj.metric.task] = [];
+    }
+    acc[obj.metric.task].push(obj);
```
andrewballantyne, line 55: If you do have a task, create an array on it then push into it... if you don't have a task, NPE? Or am I misreading this?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  const pRuns =
+    getRangeVectorData(runData, (r) => {
+      return r.metric.pipelinerun;
+    }) ?? [];
+  const finalObj: { [x: string]: { x: string; y: number }[] } = pRuns.reduce((acc, prun) => {
+    if (!prun) return acc;
+
+    const obj = prun[prun.length - 1];
```
andrewballantyne, line 51: For every loop use the last one? This loop doesn't do anything different each loop, does it?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const pipelineRunTaskRunResultData = runData?.data?.result ?? [];
+  if (runDataError || pipelineRunTaskRunResultData.length === 0) {
+    return <GraphEmpty height={DEFAULT_CHART_HEIGHT} />;
+  }
+
+  const pipelineRuns = getRangeVectorData(runData, (r) => r.metric.status) ?? [];
+  const pipelineRunsCollection = pipelineRuns.reduce((acc, prun) => {
+    if (!prun) return acc;
+    const obj = prun[prun.length - 1];
+    acc.push(obj);
```
andrewballantyne, line 47: Confused about this... each loop would push the same last item into the accumulator

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,121 @@
...
+        x: new Date().setHours(0, 0, 0, 0),
+        y: 0,
+      },
+    };
+  }
+  const successPoint = _.find(finalArray, { x: 'success' })?.['count'] ?? 0;
+
+  return (
+    <div className="row pipeline-double-chart-container">
+      <div className="col-lg-3">
```
andrewballantyne, line 90: Avoid bootstrap classes

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,121 @@
...
+import {
+  PipelineMetricsGraphProps,
+  getRangeVectorData,
+  formatDate,
+} from './pipeline-metrics-utils';
+import SuccessRatioDonut from './charts/successRatioDonut';
+import { TimeSeriesChart } from './charts/TimeSeriesChart';
+
+import './PipelineChart.scss';
+import { DEFAULT_CHART_HEIGHT } from '../const';
```
andrewballantyne, line 16: Nit: sort

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,121 @@
...
+          data={_.sortBy(finalArray, 'sortOrder')}
+          totalValue={totalValue}
+          successValue={successPoint}
+          width={(width * 30) / 100}
+        />
+      </div>
+      <div className="col-lg-9">
+        <TimeSeriesChart
+          ariaDesc="Pipeline Success"
+          ariaTitle="Pipeline Success per day"
```
andrewballantyne, line 101: I18n

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/charts/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const domainValue: DomainPropType = {
+    x: domain?.['x'] || [startDate, endDate],
+    y: domain?.['y'] || undefined,
+  };
+  let yTickFormat = formatValue;
+  const tickValues = getXaxisValues(timespan);
+  const gData: { [x: string]: any }[] = [];
+  _.each(data, (values) => {
+    values && gData.push(values);
+  });
```
andrewballantyne, line 48:
```suggestion
  const gData: { [x: string]: any }[] = data.filter((values) => !!values);
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/charts/successRatioDonut.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const SuccessRatioDonut: React.FC<SuccessRatioDonutProps & ChartDonutProps> = ({
+  data,
+  width,
+  totalValue = 0,
+  successValue = 0,
+}) => {
+  return (
+    <>
```
andrewballantyne, line 16: No need for Fragment

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-test-data.ts
```diff
@@ -0,0 +1,60 @@
+import { PrometheusResponse } from '@console/internal/components/graphs';
```
andrewballantyne, line 1: This file should be co-located with the test files

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
karthikjeeyar: `metric.task` & `metric.pipelinerun` will be always available in this prometheus  query response, but i agree the above code snippet looks confusing, so i have changed it to conditionally push the obj to accumulator only if the `metric.task` is  present.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
karthikjeeyar: For every loop, finding the last element in the values array and forming  an obj based on it.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
karthikjeeyar: Basically it flattens the multi dimensional array by picking the last element in every iteration.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/__tests__/PipelineRunDuration.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  it('Should render an empty state if query resulted in error', () => {
+    usePipelineRunDurationPollSpy.mockReturnValue([{ data: { result: [] } }, true, false]);
+    const PipelineRunDurationGraphWrapper = shallow(
+      <PipelineRunDurationGraph {...PipelineRunDurationGraphProps} />,
+    );
+    expect(PipelineRunDurationGraphWrapper.find(GraphEmpty).exists()).toBe(true);
+  });
+
+  it('Should render an LineChart if data is available', () => {
```
abhinandan13jan, line 70: nit: a LineChart

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/__tests__/PipelineRunTaskRunGraph.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  let PipelineRunTaskRunGraphProps: PipelineRunTaskRunGraphProps;
+  beforeEach(() => {
+    PipelineRunTaskRunGraphProps = {
+      pipeline,
+      timespan: DEFAULT_PROMETHEUS_TIMESPAN,
+      interval: parsePrometheusDuration(DEFAULT_REFRESH_INTERVAL),
+    };
+  });
+
+  it('Should render an LoadingInline if query result is loading', () => {
```
abhinandan13jan, line 42: nit: a LoadingInline

frontend/packages/pipelines-plugin/src/components/pipelines/const.ts
```diff
@@ -46,3 +48,17 @@ export const SecretAnnotationType = {
 
 export const PIPELINE_SERVICE_ACCOUNT = 'pipeline';
 export const PIPELINE_RUN_AUTO_START_FAILED = `bridge/pipeline-run-auto-start-failed`;
+
+export const PipelineMetricsTimeRangeOptions = (t: TFunction) => ({
```
abhinandan13jan, line 52: This is a more of a util than a const, should we position it here?

abhinandan13jan: or could we simply move the usage of t: TFunction at a util level?

karthikjeeyar: Sure, will move it.

abhinandan13jan:
The code looks fine. Ran it locally, the metrics is not always populated, because of API maybe. cc: @karthikjeeyar @andrewballantyne

bgliwa01 (approved):
Great work! I see you have the to do of interactive legend, other than that lgtm!!

andrewballantyne:
> Great work! I see you have the to do of interactive legend, other than that lgtm!!

@karthikjeeyar Can you code this up with mock data for now? I'd like us not to wait a day for you to finish up coding things.

karthikjeeyar:
@andrewballantyne Thats the plan, using mock data to finish the interactive legend work.

karthikjeeyar:
@bgliwa01 @andrewballantyne Added Interactive Legends
interactive-legend

bgliwa01:
interactive legend looks great!!

andrewballantyne (changes requested):
I noticed some things while testing...

Unbalanced screen:
Screen Shot 2020-12-01 at 6 37 36 PM

Axis label cuts into legend 🤔  Might need to try and truncate the text -- see `frontend/public/components/utils/truncate-middle.ts` and `truncateMiddle`
Screen Shot 2020-12-01 at 6 40 15 PM

Success is 100% but no bar:
Screen Shot 2020-12-01 at 6 44 40 PM

I think we may need a custom `labelComponent` ([PF property](https://www.patternfly.org/v4/charts/line-chart#chartvoronoicontainer) -- maybe we just style VictoryTooltip 🤔 ) we should avoid that awkward wrap
Screen Shot 2020-12-01 at 7 36 07 PM

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
andrewballantyne: Doh - this is what I get for doing reviews late at night.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
andrewballantyne: Yup, not sure why I didn't see that last night

andrewballantyne:
Bottom left chart crashes, when it gets data. 

image

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
```diff
@@ -0,0 +1,136 @@
...
+      tickValues={tickValues}
+      width={width}
+      height={chartHeight}
+      legendComponent={
+        <ChartLegend
+          gutter={25}
+          x={90}
+          y={DEFAULT_CHART_HEIGHT + 75}
+          itemsPerRow={4}
+          name={'legend'}
```
andrewballantyne, line 115:
```suggestion
          name="legend"
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const sortOrder = obj.x === 'success' ? 1 : 2;
+    return {
+      ...obj,
+      count: obj.y,
+      sortOrder,
+      y: parseFloat(percentage),
+      name: `${obj.x}: ${percentage}%`,
+    };
+  });
+  let finalTimeArray = finalArray.reduce((acc, obj) => {
```
andrewballantyne, line 65: Not an Array... it's a Map/Object.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -0,0 +1,125 @@
...
+          width={(width * 30) / 100}
+          ariaDesc={t('pipelines-plugin~Pipeline success Ratio Chart')}
+          ariaTitle={t('pipelines-plugin~Pipeline success ratio')}
+        />
+      </GridItem>
+      <GridItem span={9}>
+        <TimeSeriesChart
+          ariaDesc={t('pipelines-plugin~Pipeline Success Chart')}
+          ariaTitle={t('pipelines-plugin~Pipeline Success Per Day')}
+          data={_.values(finalTimeArray) ?? []}
```
andrewballantyne, line 106: Nit: `Object.values` is a thing, no need for lodash

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/charts/successRatioDonut.tsx
```diff
@@ -0,0 +1,39 @@
...
+}) => {
+  return (
+    <ChartDonut
+      ariaDesc={ariaDesc}
+      ariaTitle={ariaTitle}
+      constrainToVisibleArea
+      data={data}
+      sortKey={successValue ? ['success', 'failed'] : ['failed']}
+      labels={({ datum }) => `${datum.x}: ${datum.y}%`}
+      colorScale={successValue ? [ChartThemeColor.green, '#c9190b'] : ['#c9190b']}
```
andrewballantyne, line 25: PF colours?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/charts/successRatioDonut.tsx
```diff
@@ -0,0 +1,39 @@
...
+      padding={{
+        bottom: 20,
+        left: 20,
+        right: 20,
+        top: 20,
+      }}
+      subTitle={data ? `${successValue} succeeded of ${totalValue}` : ''}
```
andrewballantyne, line 32: I18n, follow the example used in [PipelineBuildDecoratorTooltip.tsx](https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/topology/build-decorators/PipelineBuildDecoratorTooltip.tsx#L44) (for instance)

andrewballantyne (changes requested):
I noticed the page does not at all scale with the screen... it makes the charts hard to follow. Make use of the screen size attributes (sm, md, lg, etc) for Grid and it's items...

PF: https://www.patternfly.org/v4/layouts/grid#grid

karthikjeeyar:
1. Added Scatter points
2. Responsive grids
3. i18n and improvements around the legend label , tooltips in the chart area.
pipeline_metric_legend

abhinandan13jan:
Run this locally and it runs fine.
/lgtm

Just to keep note of the following if these are needed for future as these are not specifically mentioned in Acceptance Criteria
1. Currently the metrics doesn't account for running Pipelines
2. The Taskname in the legend is different from that in the viz.

karthikjeeyar:
@abhinandan13jan  Thanks for the review
> 1. Currently the metrics doesn't account for running PipelineRuns (not yet ended execution)
     
   Only the Task run duration (4th) graph metrics are getting the results during the execution, other queries are resulting after the completion of the pipeline execution.

> 2. The Taskname in the legend is different from that in the viz.

There is a way to map the original task name with pipeline object to get the custom task name, but the issue is Pipeline is editable, so user can add/remove the tasks, so the mapping will no longer work for the removed task. @andrewballantyne Any suggestion to handle this ?

> 3. Intermittent- Unsure but if two tabs are open with PL metrics in one and we run the Pipeline in another, the metrics doesn't refresh automatically after 30 secs

As per my understanding, Prometheus Metrics results are not available instantaneously like we get updates from tkn, it takes a little extra time to scrape the data. Like I mentioned in 1st point, we need to wait till the execution completes + a small delay, but dashboard will update eventually as we are polling every 30 seconds.

andrewballantyne:
> @abhinandan13jan Thanks for the review
> 
> > 1. Currently the metrics doesn't account for running PipelineRuns (not yet ended execution)
> 
> Only the Task run duration (4th) graph metrics are getting the results during the execution, other queries are resulting after the completion of the pipeline execution.

Not sure there is much we can do here... for one, success rate cannot be true or false while running. Durations are weird during runs... we could show them I suspect (if Prometheus has the data). Perhaps the number of PLRs a day can also be updated. 

It's something we should inquire from UX @karthikjeeyar please open that discussion up once you know more about what Prometheus can provide us -- if we can't do it, might be still good to mention it in case we can do messaging or something. It's a valid misunderstanding/concern that Abhi is having.

> 
> > 1. The Taskname in the legend is different from that in the viz.
> 
> There is a way to map the original task name with pipeline object to get the custom task name, but the issue is Pipeline is editable, so user can add/remove the tasks, so the mapping will no longer work for the removed task. @andrewballantyne Any suggestion to handle this ?

This is just how Tekton works. I say we use the display name in the Pipeline (`pipeline.spec.tasks[].name`) and if they change it, manage it as a new Task.

> 
> > 1. Intermittent- Unsure but if two tabs are open with PL metrics in one and we run the Pipeline in another, the metrics doesn't refresh automatically after 30 secs
> 
> As per my understanding, Prometheus Metrics results are not available instantaneously like we get updates from tkn, it takes a little extra time to scrape the data. Like I mentioned in 1st point, we need to wait till the execution completes + a small delay, but dashboard will update eventually as we are polling every 30 seconds.

If this is how Prometheus works, so be it... not much we can do about that. How does other pages like Topology sidebar and the Monitoring page manage this? Doubt this is going to be something we can change.

andrewballantyne (changes requested):
> * Your tooltip on the TaskRun Duration charts need to be improved to show more details

This was not done... Designs call for the Pipeline name -- especially now that we are truncating it. Anyway we can get that easily into the tooltip? Probably need to put a gap between overlapping values (two points overlapping)

Odd balance when we get those first set of data... perhaps we lock out the data until we get the finished Pipeline Run? That way we can skip this scenario and swing back and improve the sizing of the graphs later.

Screen Shot 2020-12-02 at 9 19 17 PM

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineRunTaskRunGraph.tsx
```diff
@@ -0,0 +1,135 @@
...
+  if (runDataLoading) {
+    return <LoadingInline />;
+  }
+
+  const pipelineRunResultData = runData?.data?.result ?? [];
+  if (runDataError || pipelineRunResultData.length === 0) {
+    return <GraphEmpty height={chartHeight - 30} />;
+  }
+
+  const pRuns = getRangeVectorData(runData, (r) => truncateMiddle(r.metric.pipelinerun)) ?? [];
```
andrewballantyne, line 52: I think we want to truncate more to get us down to a small bitesized item... we definitely need that tooltip improvement though so they can see the full PipelineRun name.
```suggestion
  const pRuns = getRangeVectorData(runData, (r) => truncateMiddle(r.metric.pipelinerun, { length: 10 })) ?? [];
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetricsRefreshDropdown.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import IntervalDropdown from '@console/internal/components/monitoring/poll-interval-dropdown';
+
+export interface PipelineMetricsRefreshDropdownProps {
```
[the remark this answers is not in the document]
karthikjeeyar: removed

abhinandan13jan:
verified locally, looks good 
/lgtm

vikram-raj (commented):
Verified it 👍 . and it works as expected for all the users (Admin and normal) 
Thanks for fixing the tooltip issue and console warning.

/lgtm

andrewballantyne:
/approve

We should definitely consult with UX on the state of that TaskRun tooltip when multiple lines overlap.

PR #7346, opened 2020-11-27
Update associated pipeline during the edit flow

rottencandy:
**Story**:
https://issues.redhat.com/browse/ODC-5018 

**Description**: 

When there are edits made to git information like url/revision/docker file path in edit import flow, those changes should be automatically reflected in the associated pipeline as well.

When no pipeline is associated:
tmp0
When a pipeline is associated(checkbox is disabled):
tmp1
When the builder image is changed:
tmp2
When an unsupported builder image is selected:
tmp3

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -79,3 +106,44 @@ export const createPipelineRunForImportFlow = async (pipeline: Pipeline): Promis
...
+    );
+  } else {
+    const runtimeLabel = 'pipeline.openshift.io/runtime';
+    if (template.metadata?.labels[runtimeLabel] !== pipeline.metadata?.labels[runtimeLabel]) {
+      updatedPipeline = _.cloneDeep(template) as Pipeline;
+      updatedPipeline.metadata = {
+        name: `${name}`,
+        namespace,
+        labels: { ...template.metadata?.labels, 'app.kubernetes.io/instance': name },
+      };
```
divyanshiGupta, line 135: @rottencandy when you are updating the `pipeline` with the new `template` you will need to add the `resourceVersion` of the `pipeline` in `updatedPipeline` otherwise the update request fails with the below error:
` Error "Invalid value: 0x0: must be specified for an update" for field "metadata.resourceVersion".`
```suggestion
      updatedPipeline.metadata = {
        resourceVersion: pipeline.metadata.resourceVersion,
        name: `${name}`,
        namespace,
        labels: { ...template.metadata?.labels, 'app.kubernetes.io/instance': name },
      };
```

divyanshiGupta (changes requested):

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
rottencandy: Updated.

bgliwa01:
Good work! One thing - the checkboxes don't seem to be disabled...

rottencandy:
> Good work! One thing - the checkboxes don't seem to be disabled...

@bgliwa01 They are disabled, seems like it's hard to notice in Firefox,
this is how it looks like in Chrome:
scrt-screenshot-1606759964

andrewballantyne (commented):
Looking good on the surface -- will test more tomorrow

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
```diff
@@ -468,6 +469,25 @@ export const createOrUpdateResources = async (
...
+    const updatedPipeline = await updatePipelineForImportFlow(
+      pipeline,
+      formData.name,
+      formData.project.name,
+      formData.git.url,
+      formData.git.ref,
+      formData.git.dir,
+      formData.docker.dockerfilePath,
+    );
+    if (_.has(updatedPipeline.metadata.labels, 'pipeline.openshift.io/runtime')) {
```
divyanshiGupta, line 482: @rottencandy what was the reason for adding this check for `pipeline.openshift.io/runtime` label? I am asking because this wont be present in the pipeline that is associated with an app created using import from dockerfile and thus if the this app is edited, pipeline-run will not be started automatically.

rottencandy: I added this check to make sure that the pipeline is still associated to the application, realized now that I should be checking for the `app.kubernetes.io/instance` label instead. Updated.

divyanshiGupta: Yes. Thanks!

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -294,6 +294,7 @@ export const getCommonInitialValues = (
     serverless: getServerlessData(editAppResource),
     pipeline: {
       enabled: !_.isEmpty(pipelineData),
+      data: pipelineData,
```
divyanshiGupta, line 297: Also you don't need to add a new prop for storing the pipeline in the form state itself as if you go through `EditApplication` you will see that we are sending all the resources including pipeline to `createOrUpdateGitResources` so you can directly get the `pipeline` there.

rottencandy: Thanks, updated.

bgliwa01 (approved):
lgtm...based on pipelines meeting, will approve and then potentially handle changes as a bug, if needed.

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -79,3 +106,45 @@ export const createPipelineRunForImportFlow = async (pipeline: Pipeline): Promis
...
+): Promise<Pipeline> => {
+  let updatedPipeline = _.cloneDeep(pipeline);
+
+  if (!template) {
+    updatedPipeline.metadata.labels = _.omit(
+      updatedPipeline.metadata.labels,
+      'app.kubernetes.io/instance',
+    );
+  } else {
+    const runtimeLabel = 'pipeline.openshift.io/runtime';
```
andrewballantyne, line 128: We are starting to duplicate this label... I found 4 instances of it. Perhaps add it to `frontend/packages/pipelines-plugin/src/const.ts` and remove the duplicates.

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -79,3 +106,45 @@ export const createPipelineRunForImportFlow = async (pipeline: Pipeline): Promis
...
+    if (template.metadata?.labels[runtimeLabel] !== pipeline.metadata?.labels[runtimeLabel]) {
+      updatedPipeline = _.cloneDeep(template) as Pipeline;
```
andrewballantyne, line 130:
```suggestion
      updatedPipeline = _.cloneDeep(template);
```
Shouldn't need to do this

frontend/packages/pipelines-plugin/src/plugin.tsx
```diff
@@ -471,7 +471,7 @@ const plugin: Plugin<ConsumedExtensions> = [
       label: 'Pipeline',
       // t('pipelines-plugin~Create a Tekton Pipeline to automate delivery of your application')
       description:
-        '%pipelines-plugin~Create a Tekton Pipeline to automate delivery of your application~',
+        '%pipelines-plugin~Create a Tekton Pipeline to automate delivery of your application%',
```
andrewballantyne, line 474: 👍

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
```diff
@@ -468,6 +470,26 @@ export const createOrUpdateResources = async (
         generatedImageStreamName,
       ),
     );
+  } else if (!_.isEmpty(appResources?.pipeline?.data) && !dryRun) {
```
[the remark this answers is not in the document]
rottencandy: Updated.

divyanshiGupta:
Code looks good to me. Creating a cluster to test this on and will add lgtm after that.

divyanshiGupta:
Verified it locally, works fine for me. 
/lgtm

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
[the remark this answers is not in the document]
andrewballantyne: You combined the two if statements and changed the logic. 

If we have a pipeline template, we don't want a build config. But if it's a dry run, we don't want to do Pipeline stuff because the K8s resources that support DryRun can fail. 

If it's a dry run and we have a template -- you'll request creating a BuildConfig... it's no real harm in the end, but it could theoretically fail the flow and it's just a waste of a call.

```
  if (pipeline.enabled) {
    if (!dryRun) {
      await managePipelineResources(formData, appResources);
    }
  } else {
```
and
```
  if (pipeline.enabled && !dryRun) {
    managePipelineResources(formData, appResources);
  }
  } else {
```
Are two different logic flows. 

Also, always `await` on async methods... rarely do you want to unbound a promise and let it run on it's own without context of this thread -- even if you don't want the response. If an issue happens, it can be confusing where the issue lies if you're already navigating to the Topology screen.

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -63,6 +61,36 @@ export const createPipelineForImportFlow = async (
...
+export const createPipelineForImportFlow = async (
+  name: string,
+  namespace: string,
+  gitUrl: string,
+  gitRef: string,
+  gitDir: string,
+  pipeline: PipelineData,
+  dockerfilePath: string,
+) => {
+  const template = _.cloneDeep(pipeline.template) as Pipeline;
```
andrewballantyne, line 75:
```suggestion
  const template = _.cloneDeep(pipeline.template);
```

andrewballantyne (changes requested):

andrewballantyne:
One major change needed - https://github.com/openshift/console/pull/7346#discussion_r534412546

frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
rottencandy: Updated.

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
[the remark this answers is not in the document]
rottencandy: Sorry, didn't think it through before combining the statements. Updated.

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
```diff
@@ -2,8 +2,10 @@
   "Pipelines": "Pipelines",
   "Select a Builder Image and resource to see if there is a pipeline template available for this runtime.": "Select a Builder Image and resource to see if there is a pipeline template available for this runtime.",
   "The pipeline template for Dockerfiles is not available at this time.": "The pipeline template for Dockerfiles is not available at this time.",
+  "There are no pipeline templates available for {{builderImage}}, current pipeline will be dissociated from application.": "There are no pipeline templates available for {{builderImage}}, current pipeline will be dissociated from application.",
   "There are no pipeline templates available for {{builderImage}} and {{resourceType}} combination.": "There are no pipeline templates available for {{builderImage}} and {{resourceType}} combination.",
   "this Builder Image": "this Builder Image",
+  "Pipeline will be updated to match the builder image.": "Pipeline will be updated to match the builder image.",
```
andrewballantyne, line 8:
```suggestion
  "Pipeline will be updated to match the Builder Image.": "Pipeline will be updated to match the Builder Image.",
```
Builder Image is a resource, so a proper noun, must be title case.

andrewballantyne (approved):
Need to fix the Translation before this can merge so we don't need to do another PR to fix it.

andrewballantyne (approved):

andrewballantyne (approved):
/lgtm

PR #7351, opened 2020-11-30
Added init state to insights plugin

bond95:
Init state appear when value of at least one insight data is `-1`. Unfortunately if reload insights-operator values of insights data are back to `-1`, thus init state appear again.

Screenshot from 2020-11-30 10-03-16

Please, take a look @doruskova @tisnik @spadgett @vojtechszocs

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
@@ -48,8 +48,13 @@ export const mapMetrics = (response: PrometheusResponse): Metrics => {
       return null;
     }
     const metricName = response.data?.result?.[i]?.metric?.metric;
-    values[metricName] = parseInt(value, 10);
+    if (values[metricName] === -1 || values[metricName] === undefined) {
+      values[metricName] = parseInt(value, 10);
+    }
   }
 
   return values;
 };
+
+export const isInitState = (values: Metrics) =>
+  Object.values(values).some((cur: number) => cur === -1);
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -1,6 +1,13 @@
 import * as React from 'react';
 import { ChartDonut, ChartLegend, ChartLabel } from '@patternfly/react-charts';
-import { riskIcons, colorScale, legendColorScale, riskSorting, mapMetrics } from './mappers';
+import {
+  riskIcons,
+  colorScale,
+  legendColorScale,
+  riskSorting,
+  mapMetrics,
+  isInitState,
+} from './mappers';
 import { PrometheusHealthPopupProps } from '@console/plugin-sdk';
 import { K8sResourceKind } from '@console/internal/module/k8s';
 import { ExternalLink } from '@console/internal/components/utils';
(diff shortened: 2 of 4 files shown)


frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
```diff
@@ -48,8 +48,13 @@ export const mapMetrics = (response: PrometheusResponse): Metrics => {
...
+    if (values[metricName] === -1 || values[metricName] === undefined) {
+      values[metricName] = parseInt(value, 10);
+    }
   }
 
   return values;
 };
+
+export const isInitState = (values) =>
+  !Object.values(values).reduce((cur, acc) => (cur === -1 ? acc && false : acc), true);
```
tisnik, line 60: `acc &&` is probably not needed there

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
```diff
@@ -48,8 +48,13 @@ export const mapMetrics = (response: PrometheusResponse): Metrics => {
       return null;
     }
     const metricName = response.data?.result?.[i]?.metric?.metric;
-    values[metricName] = parseInt(value, 10);
+    if (values[metricName] === -1 || values[metricName] === undefined) {
```
tisnik, line 51: prob. I'm confused, but has `values[metricName]` numeric or string type?

tisnik (commented):
just two nitpicks, looks very ok in overall @bohdan!

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
bond95: string type

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
bond95: `&&` should be there

tisnik (approved):

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
bond95: I'll just quote here my previous answer to you:

> Nope, I'm saying that it's more clear version

I don't know where you saw that this works differently, but ok

doruskova:
@bond95 looks good to me :)

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
spadgett: I'm confused here. Isn't `acc && false` always the same as `false`?

If I'm reading the logic right, can this be...
```suggestion
  Object.values(values).some((cur: number) => cur === -1);
```

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
```diff
@@ -48,8 +48,13 @@ export const mapMetrics = (response: PrometheusResponse): Metrics => {
...
+      values[metricName] = parseInt(value, 10);
+    }
   }
 
   return values;
 };
+
+export const isInitState = (values) =>
```
spadgett, line 59: Missing type on `values`.

bond95: Fixed

frontend/packages/insights-plugin/src/components/InsightsPopup/mappers.ts
bond95: Fixed

tisnik: that one is much nicer!

spadgett (approved):
/lgtm

PR #7362, opened 2020-11-30
Import console i18n instance

pcbailey:
This PR imports the console i18n instance into the kubevirt-plugin and adds a directive to the frontend webpack config to copy kubevirt-plugin locales files to frontend/public/locales to make the keys discoverable.


pcbailey:
@rawagner @yaacov Please review.

PR #7364, opened 2020-11-30
Avoid erroneous use of python3 with old node-gyp

yselkowitz:
node-sass, used within console, requires an old version of node-gyp 
(specifically 3.8.0) which does not have Python 3 support.  Since the 
switch to Node.js 14, something has caused node-gyp to look only at 
python3 as a python interpreter regardless of its absence, or the 
presence of python2.

This fixes the downstream builds of console by forcing the python 
interpreter version.

/cc @spadgett

### DIFF
--- a/Dockerfile.product
+++ b/Dockerfile.product
@@ -16,7 +16,7 @@ FROM registry.svc.ci.openshift.org/ocp/builder:ubi8.nodejs.14 AS nodebuilder
 ADD . .
 
 USER 0
-# required for node-gyp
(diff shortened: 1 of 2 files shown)


spadgett:
Thanks, it might be better to simply upgrade to a newer version of node-sass:

https://github.com/sass/node-sass/releases/tag/v5.0.0

Let me look at that quickly. Otherwise, we can go with this fix.

yselkowitz:
/hold
pending decision on #7365

PR #7365, opened 2020-11-30
Bug 1901594: bump node-sass to 5.0.0

spadgett:
This addresses a build failure with Node.js 14. The older node-sass version required Python 2.

### DIFF
--- a/Dockerfile.product
+++ b/Dockerfile.product
@@ -17,7 +17,7 @@ ADD . .
 
 USER 0
 # required for node-gyp
-RUN yum install -y python2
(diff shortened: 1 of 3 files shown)


spadgett:
cc @yselkowitz

I'm not sure offhand if this fixes the problem if python3 is not available. Do you know? We could still need #7364.

yselkowitz:
A local build of downstream worked for me, let's see how CI goes.

spadgett:
This will indirectly fix the test failures we're seeing in the openshift-ocp-installer-console-aws-4.7 job.

PR #7366, opened 2020-11-30
Bug 1898532: Remove _.startCase from FieldSet title

TheRealJon:
This was missed when switching to `useSchemaLabel` hook.


spadgett (approved):
/lgtm

PR #7367, opened 2020-11-30
Cypress: disable Monitoring test suite

dtaylor113:
Too many flakes in recent days.  Will re-evaluate after speaking with Monitoring team.

Thought this BZ was the root cause of the issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1891815

Reopened: https://bugzilla.redhat.com/show_bug.cgi?id=1870287


spadgett (approved):
/lgtm

PR #7369, opened 2020-12-01
Helm chart list should not return helm charts if the repo is disabled.

akashshinde:
This PR adds supports to return helm chart listing of only enabled repositories.

### DIFF
--- a/pkg/helm/chartproxy/proxy.go
+++ b/pkg/helm/chartproxy/proxy.go
@@ -77,24 +77,26 @@ func (p *proxy) IndexFile(onlyCompatible bool) (*repo.IndexFile, error) {
 
 	indexFile := repo.NewIndexFile()
 	for _, helmRepo := range helmRepos {
-		idxFile, err := helmRepo.IndexFile()
-		if err != nil {
-			klog.Errorf("Error retrieving index file for %v: %v", helmRepo, err)
-			continue
-		}
+		if !helmRepo.Disabled {
+			idxFile, err := helmRepo.IndexFile()
+			if err != nil {
+				klog.Errorf("Error retrieving index file for %v: %v", helmRepo, err)
+				continue
+			}
 
-		for key, entry := range idxFile.Entries {
-			if onlyCompatible {
-				for i := len(entry) - 1; i >= 0; i-- {
-					if entry[i].Metadata.KubeVersion != "" && p.kubeVersion != "" {
-						if !chartutil.IsCompatibleRange(entry[i].Metadata.KubeVersion, p.kubeVersion) {
-							entry = append(entry[:i], entry[i+1:]...)
+			for key, entry := range idxFile.Entries {
+				if onlyCompatible {
+					for i := len(entry) - 1; i >= 0; i-- {
+						if entry[i].Metadata.KubeVersion != "" && p.kubeVersion != "" {
+							if !chartutil.IsCompatibleRange(entry[i].Metadata.KubeVersion, p.kubeVersion) {
+								entry = append(entry[:i], entry[i+1:]...)
+							}
 						}
 					}
 				}
-			}
-			if len(entry) > 0 {
-				indexFile.Entries[key+"--"+helmRepo.Name] = entry
+				if len(entry) > 0 {
+					indexFile.Entries[key+"--"+helmRepo.Name] = entry
+				}
 			}
 		}
 	}
--- a/pkg/helm/chartproxy/repos.go
+++ b/pkg/helm/chartproxy/repos.go
@@ -38,6 +38,7 @@ const (
 type helmRepo struct {
 	Name            string
 	URL             *url.URL
+	Disabled        bool
 	TLSClientConfig *tls.Config
 }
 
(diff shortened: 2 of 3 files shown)


pkg/helm/chartproxy/repos.go
```diff
@@ -101,6 +101,15 @@ type helmRepoGetter struct {
 
 func (b helmRepoGetter) unmarshallConfig(repo unstructured.Unstructured) (*helmRepo, error) {
 	h := &helmRepo{}
+
+	disabled, _, err := unstructured.NestedBool(repo.Object, "spec", "disabled")
+	if err != nil {
```
pedjak, line 106: if `disabled` field does not exist, would this trigger error?

I would not skip unmarshaling if repo is disabled - instead I would add add `disabled` field to `helmRepo` struct, unmarshall to it as usual, and then in `List` method decide based on the value of `disabled` field.

pkg/helm/chartproxy/repos.go
```diff
@@ -192,7 +201,9 @@ func (b *helmRepoGetter) List() ([]*helmRepo, error) {
 			klog.Errorf("Error unmarshalling repo %v: %v", item, err)
 			continue
 		}
-		helmRepos = append(helmRepos, helmConfig)
+		if helmConfig != nil {
```
pedjak, line 204: see above, then this becomes
if !helmConfig.disabled {
.
.

}
```

```

pkg/helm/chartproxy/repos_test.go
```diff
@@ -341,3 +341,148 @@ func TestHelmRepo_IndexFile(t *testing.T) {
...
+func TestHelmRepoGetter_SkipDisabled(t *testing.T) {
+	tests := []struct {
+		name             string
+		helmCRS          []*unstructured.Unstructured
+		expectedRepoName []string
+		disabled	bool
+		apiErrors        []apiError
+	}{
+		{
+			name: "list helm repos with no disabled repo should list all repositories",
```
pedjak, line 355: This test is not needed because it is already covered `TestHelmRepoGetter_List` function.

pkg/helm/chartproxy/repos_test.go
```diff
@@ -341,3 +341,148 @@ func TestHelmRepo_IndexFile(t *testing.T) {
...
+								"url": "http://foo2.com/bar",
+							},
+						},
+					},
+				},
+			},
+			expectedRepoName: []string{"repo1","repo2"},
+		},
+		{
+			name: "list helm repos with disabled flag true should be ignored",
```
pedjak, line 392: better reading:

```
return only enabled repos
```

pkg/helm/chartproxy/repos_test.go
```diff
@@ -341,3 +341,148 @@ func TestHelmRepo_IndexFile(t *testing.T) {
...
+								"url": "http://foo2.com/bar",
+							},
+						},
+					},
+				},
+			},
+			expectedRepoName: []string{"repo2"},
+		},
+		{
+			name: "list helm repos with all disabled flags true should be empty",
```
pedjak, line 430: better reading:

```
return empty list if no repos are enabled
```

pedjak (changes requested):
remove WIP from PR title

pkg/helm/chartproxy/repos.go
akashshinde: Done

pkg/helm/chartproxy/repos_test.go
akashshinde: Makes sense, removed

pkg/helm/chartproxy/repos_test.go
akashshinde: done

pkg/helm/chartproxy/repos_test.go
akashshinde: done

pkg/helm/chartproxy/repos.go
akashshinde: @pedjak if `disabled` field isn't found, `unstructured.NestedBool` returns false and `err` would be nil.
ref: https://github.com/openshift/console/blob/117280d206855362c3044b7fca51313492eaff8c/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go#L89

pkg/helm/chartproxy/proxy.go
```diff
@@ -72,13 +72,15 @@ func (p *proxy) IndexFile() (*repo.IndexFile, error) {
...
 	indexFile := repo.NewIndexFile()
 	for _, helmRepo := range helmRepos {
-		idxFile, err := helmRepo.IndexFile()
-		if err != nil {
-			klog.Errorf("Error retrieving index file for %v: %v", helmRepo, err)
-			continue
-		}
-		for key, entry := range idxFile.Entries {
-			indexFile.Entries[key+"--"+helmRepo.Name] = entry
+		if !helmRepo.Disabled {
```
pedjak, line 75: `p.helmRepoGetter.List()` should return only non disabled repos IMHO.

pedjak (approved):
please squash

rohitkrai03 (approved):
/lgtm

Verified UI with the changes. Works as expected.

PR #7379, opened 2020-12-01
Fix sorting of nav section items

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5158

**Analysis / Root cause**: 
Items within a section were only sorted linearly so an item specifying before/after an item declared later did not get sorted based on that item.

**Solution Description**: 
Create utilities from existing sorting done for sections. Update the section nav items to use the same utilities and sort in the same manner as the sections.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


andrewballantyne:
@jeff-phillips-18 I'm not seeing new tests -- can we make sure we write a test to cover this scenario? Hopefully extensions don't in the way of the test being written 🙂

jeff-phillips-18:
The tests for the utils covers what is done here. Code reuse of code that had tests already :)

jeff-phillips-18:
Rebased w/ no other changes

jeff-phillips-18:
Rebased again

PR #7382, opened 2020-12-01
Bug 1897008: Cypress: reenable check for 'aria-hidden-focus' rule & checkA11y test for modals

dtaylor113:
This PR re-enables the a11y check for 'aria-hidden-focus' rule and allows A11y testing of Modals in isolation.  Updated modal div to `<div id="modal-container" role="dialog" aria-modal="true" />`.

### DIFF
--- a/frontend/packages/integration-tests-cypress/support/a11y.ts
+++ b/frontend/packages/integration-tests-cypress/support/a11y.ts
@@ -6,7 +6,7 @@ declare global {
   namespace Cypress {
     interface Chainable<Subject> {
       logA11yViolations(violations: Result[], target: string): Chainable<Element>;
-      testA11y(target: string): Chainable<Element>;
+      testA11y(target: string, selector?: string): Chainable<Element>;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


spadgett:
@dtaylor113 Can we simply add `aria-disabled` to the app element when the modal is open instead of adding it to every input? I don't it's scalable to have the individual components need to check if there is a modal open.

dtaylor113:
> @dtaylor113 Can we simply add `aria-disabled` to the app element when the modal is open instead of adding it to every input? I don't it's scalable to have the individual components need to check if there is a modal open.

Tried this and it didn't work :-(  Neither did any of the suggestions [here](https://stackoverflow.com/questions/62677291/aria-hidden-elements-do-not-contain-focusable-elements-issue-when-modal-is-sho)

dtaylor113:
Updated `react-modal` to NOT add `aria-hidden="true"` to the app root, thus bypassing the `aria-hidden-focus` violations.  Updated modal div to `<div id="modal-container" role="dialog" aria-modal="true" />`.

dtaylor113:
Updated PR to re-enables the a11y check for 'aria-hidden-focus' rule and allows A11y testing of Modals in isolation.  Updated modal div to `<div id="modal-container" role="dialog" aria-modal="true" />`.

frontend/packages/integration-tests-cypress/support/a11y.ts
```diff
@@ -34,19 +34,18 @@ Cypress.Commands.add('logA11yViolations', (violations: Result[], target: string)
...
       { id: 'color-contrast', enabled: false }, // seem to be somewhat inaccurate and has difficulty always picking up the correct colors, tons of open issues for it on axe-core
       { id: 'focusable-content', enabled: false }, // recently updated and need to give the PF team time to fix issues before enabling
       { id: 'scrollable-region-focusable', enabled: false }, // recently updated and need to give the PF team time to fix issues before enabling
-      { id: 'aria-hidden-focus', enabled: false }, // disabling until we implement correct handling of Modals, see https://dequeuniversity.com/rules/axe/3.4/aria-hidden-focus?application=axeAPI
     ],
   });
   a11yTestResults.numberChecks += 1;
   cy.checkA11y(
-    null,
+    selector || null,
```
spadgett, line 48: I don't think you need `|| null` here? Does it cause a problem leaving that off?

dtaylor113: fixed, removed `|| null`

spadgett (approved):
/lgtm

PR #7384, opened 2020-12-01
Add scripts to get and set user settings on a cluster

jerolimov:
**Fixes**: 
~~https://issues.redhat.com/browse/ODC-4755~~ https://issues.redhat.com/browse/ODC-5189

Extracted this from PR #7327

**Analysis / Root cause**:

Because we developers switches cluster more often then users, its could be helpful to extract the own user settings from a cluster and restore them on another cluster.

**Solution Description**: 
Created two small scripts to get and set user settings on a cluster.

Requirements:
* `oc` cli, https://github.com/openshift/oc
* `yq` cli, see https://github.com/mikefarah/yq#install

Usage:

```bash
./contrib/oc-get-user-settings.sh [username] [filename]

./contrib/oc-set-user-settings.sh [username] [filename]
```

Both commands can be called without any argument.

Default username is the current oc user (`oc whoami`)

Default filename is `user-settings.yaml`

**Test setup:**
* You need a 4.7 cluster with PRs #7095 and #7327
* `oc login`
* Use the browser to store some user settings (topology view graph switch for example)
* Call get and set scripts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@
 /frontend/@types
 /frontend/package-lock.json
 /frontend/po-files
+/user-settings.yaml
--- a/contrib/oc-get-user-settings.sh
+++ b/contrib/oc-get-user-settings.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+# shellcheck shell=bash
+#
+# This file is an example of how you might set up your environment to
+# run the console against an OpenShift cluster during development. To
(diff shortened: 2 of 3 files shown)


jerolimov:
We solved https://issues.redhat.com/browse/ODC-5189 in PR #7327

And with PR #7586 developers uses the localstorage automatically again, so we can close this PR.

PR #7387, opened 2020-12-02
Bug 1903402: Added isVMPage boolean to choose correct links for Nics and Disks

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1903402

**Analysis / Root cause**: 
There wasn't a way to tell if you are on VM page or VMI page, which in result VM page was always choose.

**Solution Description**: 
Added a boolean to check if you areo n VM or VMI page


metalice:
@rawagner , @glekner please review it :)  Thanks!

PR #7389, opened 2020-12-02
Validation on storage & nodes install step

cloudbehl:


afreen23:
@cloudbehl we need to do the change for internal mode as well.

cloudbehl:
@afreen23 updated for internal mode as well

PR #7390, opened 2020-12-02
Save dropdown bookmarks and favoriteKey in user settings

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5198

**Analysis / Root cause**: 
As a user, I want sync my bookmarks for namespaces, applications and other dropdowns between browsers.

**Solution Description**: 
Update Dropdown component to use `useUserSettings` hook and save `booksmarks` and `favoriteKey` in user settings.

From user perspective nothing has changed. This video shows the migration and update of the configmap:
Peek 2020-12-02 14-20

Untouched

**Test setup:**
Open developer perspective and bookmark / unbookmark / favorite namespaces and applications.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
+++ b/frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
@@ -331,7 +331,7 @@ export const DropdownField: React.FC<FieldProps> = ({
   uiSchema = {},
 }) => {
   const { t } = useTranslation();
-  const { items, title } = getUiOptions(uiSchema);
+  const { items, title } = getUiOptions(uiSchema) as { items?: object; title?: string };
   return (
     <Dropdown
       id={idSchema.$id}
--- a/frontend/packages/console-shared/src/constants/common.ts
+++ b/frontend/packages/console-shared/src/constants/common.ts
@@ -33,7 +33,9 @@ export const STORAGE_PREFIX = 'bridge';
 export const USERSETTINGS_PREFIX = 'console';
 
 // This localStorage key predates the storage prefix.
+export const NAMESPACE_USERSETTINGS_PREFIX = `${USERSETTINGS_PREFIX}.namespace`;
 export const NAMESPACE_LOCAL_STORAGE_KEY = 'dropdown-storage-namespaces';
+export const APPLICATION_USERSETTINGS_PREFIX = `${USERSETTINGS_PREFIX}.applications`;
 export const APPLICATION_LOCAL_STORAGE_KEY = 'dropdown-storage-applications';
 export const LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY = `${STORAGE_PREFIX}/last-namespace-name`;
 export const API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY = `${STORAGE_PREFIX}/api-discovery-resources`;
--- a/frontend/public/components/utils/dropdown.jsx
+++ b/frontend/public/components/utils/dropdown.jsx
@@ -562,13 +562,26 @@ export const Dropdown = (props) => {
     [setBookmarks],
   );
 
+  // FIXME: Remove this after latest namespace wasn't fetched from localStorage anymore.
+  const onFavorite = React.useCallback(
+    (key) => {
+      setFavoriteKey(key);
+      if (key) {
+        localStorage.setItem(favoriteStorageKey, key);
+      } else {
+        localStorage.removeItem(favoriteStorageKey);
+      }
+    },
+    [setFavoriteKey, favoriteStorageKey],
+  );
+
   return (
     <Dropdown_
       {...props}
       bookmarks={bookmarks}
       onBookmark={onBookmark}
       favoriteKey={favoriteKey}
-      onFavorite={setFavoriteKey}
+      onFavorite={onFavorite}
     />
   );
 };
--- a/frontend/public/components/utils/dropdown.jsx
+++ b/frontend/public/components/utils/dropdown.jsx
@@ -196,7 +196,6 @@ class DropDownRow extends React.PureComponent {
   }
 }
 
-/** @augments {React.Component<any>} */
 class Dropdown_ extends DropdownMixin {
   constructor(props) {
     super(props);
@@ -600,6 +599,15 @@ Dropdown.propTypes = {
   textFilter: PropTypes.string,
   title: PropTypes.node,
   disabled: PropTypes.bool,
+  id: PropTypes.string,
+  onChange: PropTypes.func,
+  selectedKey: PropTypes.string,
+  titlePrefix: PropTypes.string,
+  ariaLabel: PropTypes.string,
+  name: PropTypes.string,
+  autoSelect: PropTypes.bool,
+  describedBy: PropTypes.string,
+  required: PropTypes.bool,
 };
 
 class ActionsMenuDropdown_ extends DropdownMixin {
--- a/frontend/public/components/utils/dropdown.jsx
+++ b/frontend/public/components/utils/dropdown.jsx
@@ -3,16 +3,18 @@ import * as React from 'react';
 import * as classNames from 'classnames';
 import * as PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { useTranslation, withTranslation } from 'react-i18next';
 import { CaretDownIcon, MinusCircleIcon, PlusCircleIcon, StarIcon } from '@patternfly/react-icons';
+import { useUserSettingsCompatibility } from '@console/shared';
+
 import { impersonateStateToProps } from '../../reducers/ui';
 import { checkAccess } from './rbac';
 import { history } from './router';
 import { KebabItems } from './kebab';
 import { ResourceName } from './resource-icon';
 import { useSafetyFirst } from '../safety-first';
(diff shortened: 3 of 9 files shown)


frontend/public/components/utils/dropdown.jsx
```diff
@@ -598,6 +532,43 @@ export class Dropdown extends DropdownMixin {
...
+  const { storageKey } = props;
+
+  // Should be undefined so that we don't save undefined-bookmarks.
+  const bookmarkStorageKey = storageKey ? `${storageKey}-bookmarks` : undefined;
+
+  const [favoriteKey, setFavoriteKey] = useUserSettingsCompatibility(
+    storageKey,
+    storageKey,
+    null,
+    // TODO: enable sync when sync PR is merged
```
sahil143, line 545: @jerolimov how would the returned state work here. Since the key is undefined and we don't want to initialize it in user settings.

sahil143: I'm guessing similar to `useState`?

sahil143 (changes requested):

frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
```diff
@@ -337,7 +337,7 @@ export const DropdownField: React.FC<FieldProps> = ({
       key={idSchema.$id}
       title={t('console-shared~Select {{title}}', { title: title || schema?.title || name })}
       selectedKey={formData}
-      items={items ?? {}}
+      items={(items as object) ?? {}}
```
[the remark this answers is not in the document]
jerolimov: Because items comes from `const { items, title } = getUiOptions(uiSchema);` which returns items as string | string[] | array | object .... In this case this must be an object, this is was Dropdown requested before.

frontend/public/components/utils/dropdown.jsx
```diff
@@ -598,6 +532,43 @@ export class Dropdown extends DropdownMixin {
...
 
+export const Dropdown = (props) => {
+  const { storageKey } = props;
+
+  // Should be undefined so that we don't save undefined-bookmarks.
+  const bookmarkStorageKey = storageKey ? `${storageKey}-bookmarks` : undefined;
+
+  const [favoriteKey, setFavoriteKey] = useUserSettingsCompatibility(
+    storageKey,
+    storageKey,
```
christianvogt, line 543: The format of the user-settings keys should be `namespace.key`. eg `console.foobar`

I'd prefer to see a temporary prop in place for the old key and new one that adheres to the new format.
The current keys I see created are:
```
  dropdown-storage-namespaces
  dropdown-storage-namespaces-bookmarks
```

frontend/public/components/utils/dropdown.jsx
```diff
@@ -598,6 +532,43 @@ export class Dropdown extends DropdownMixin {
...
+    null,
```
christianvogt, line 544: Prefer `undefined` because this is what's being saved in the user settings:
```
  dropdown-storage-namespaces: 'null'
```

frontend/public/components/utils/dropdown.jsx
```diff
@@ -598,6 +532,43 @@ export class Dropdown extends DropdownMixin {
...
+    // TODO: enable sync when sync PR is merged
+  );
+  const [bookmarks, setBookmarks] = useUserSettingsCompatibility(
+    bookmarkStorageKey,
+    bookmarkStorageKey,
+    {},
```
christianvogt, line 550: Can this be `undefined` to start?

christianvogt (commented):
Unable to clear the favorite item. After page refresh, it comes back. The config map does not update.

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -77,9 +77,9 @@ export const useUserSettings = <T>(
...
 
   React.useEffect(() => {
-    if (cfData && cfLoaded && settings !== undefined) {
+    if (keyRef.current && cfLoaded && cfData && settings !== undefined) {
       const value = seralizeData(settings);
-      if (value !== cfData.data?.[key]) {
+      if (value !== cfData.data?.[keyRef.current]) {
         updateConfigMap(cfData, keyRef.current, value);
       }
     }
```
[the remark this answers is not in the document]
jerolimov: Reverted this to don't get into a merge conflict. We need to check after both are merged it there are some settings are stored without key.

frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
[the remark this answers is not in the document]
jerolimov: Moved this cast to the line where getUiOptions is assigned to items

frontend/public/components/utils/dropdown.jsx
jerolimov: Okay, this (old) Dropdown component has now a `userSettingsPrefix` and `storageKey` prop, so that the users can set both attributes for the `useUserSettingsCompatibility`. I used the prefix name because the component automatically appends `.favorite` and `.bookmarks`.

frontend/public/components/utils/dropdown.jsx
jerolimov: Removed the default so its now undefined.

frontend/public/components/utils/dropdown.jsx
jerolimov: Yeah works fine. Removed the default so its now undefined.

jerolimov:
> Unable to clear the favorite item. After page refresh, it comes back. The config map does not update.

Fixed this. This sets now null instead of undefined to clear the `.favorite` entry.

Undefined is currently not supported, let add support for undefined after FF?

@sahil143 @christianvogt Can you take a new look? Thanks

invincibleJai:
Verified the changes , works as expected. Need some work tests looks like

frontend/packages/console-shared/src/constants/common.ts
```diff
@@ -33,7 +33,9 @@ export const STORAGE_PREFIX = 'bridge';
 export const USERSETTINGS_PREFIX = 'console';
 
 // This localStorage key predates the storage prefix.
+export const NAMESPACE_USERSETTINGS_PREFIX = 'console.namespace';
```
debsmita1, line 36: How about:

```suggestion
export const NAMESPACE_USERSETTINGS_PREFIX =`${USERSETTINGS_PREFIX}.namespace`;
```

frontend/packages/console-shared/src/constants/common.ts
```diff
@@ -33,7 +33,9 @@ export const STORAGE_PREFIX = 'bridge';
...
 export const NAMESPACE_LOCAL_STORAGE_KEY = 'dropdown-storage-namespaces';
+export const APPLICATION_USERSETTINGS_PREFIX = 'console.applications';
```
debsmita1, line 38:
```suggestion
export const APPLICATION_USERSETTINGS_PREFIX = `${USERSETTINGS_PREFIX}.applications`;
```

frontend/packages/console-shared/src/constants/common.ts
jerolimov: Yeah good idea. Done.

frontend/packages/console-shared/src/constants/common.ts
jerolimov: Yeah good idea. Done.

debsmita1:
tested this locally, works fine
/lgtm

jerolimov:
Integration tests found an issue. Working on that.
/hold

jerolimov:
/unhold
The previous version of this PR has a bug that the simple Dropdown in Admin > Workloads > Pods or Developer > Search was empty. This is fixed now. I restored the line ... in dropdown.jsx Dropdown_ constructor.

```tsx
    this.state.items = props.items;
```

The dynamic cases of projects and apps works fine because the items was also set when the props are updated.

Here are some working examples:

Simple:
Peek 2020-12-04 18-44

Search:
Peek 2020-12-04 18-45

Project and apps:
Peek 2020-12-04 18-452

frontend/public/components/utils/dropdown.jsx
```diff
@@ -562,13 +562,26 @@ export const Dropdown = (props) => {
     [setBookmarks],
   );
 
+  // FIXME: Remove this after latest namespace wasn't fetched from localStorage anymore.
```
andrewballantyne, line 565: This is a temporary solution? It's a FIXME but the text does not specify when we would fix it. Could you explain more of what the scenario is?

sahil143: I guess once my PR goes in.

andrewballantyne:
/lgtm
/hold

Putting a hold on this since it has a [dependency](https://github.com/openshift/console/pull/7390#discussion_r536288377)

andrewballantyne:
/hold cancel

Removing hold based on discussion with @jerolimov -- we'll log a ticket to address removing the FIXME

andrewballantyne:
/retest

Operator degraded issue -- don't think it's related, this task passed already.

PR #7393, opened 2020-12-02
Bug 1897354:  make CRDCard display consistent with tabs

rhamilto:
We were filtering out CRDs from the Provided API cards list if the model for the CRD didn't exist; this created a discrepancy between what the cards listed and what the tabs listed.  This aligns the cards with the tabs, and both now reflect what's in the CSV YAML.

Reproducer:

add a fake `spec.customresourcedefinitions.owned` to the CSV.  

```
spec:
  customresourcedefinitions:
    owned:
      - description: Foo is the Schema for the jaegers API
        displayName: Foo
        kind: Foo
        name: foo.jaegertracing.io
        version: v1
```

Before:
Screen Shot 2020-12-03 at 5 18 25 PM

After:
Screen Shot 2020-12-03 at 5 17 54 PM

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
@@ -753,9 +753,6 @@ const crdCardRowStateToProps = ({ k8s }, { crdDescs }) => {
     crdDescs.map((desc) => k8s.getIn(['RESOURCES', 'models', referenceForProvidedAPI(desc)])),
   );
   return {
-    crdDescs: crdDescs.filter((desc) =>
-      models.find((m) => referenceForModel(m) === referenceForProvidedAPI(desc)),
-    ),
     createable: models
       .filter((m) => (m.verbs || []).includes('create'))
       .map((m) => referenceForModel(m)),


frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -765,10 +766,15 @@ const crdCardRowStateToProps = ({ k8s }, { crdDescs }) => {
 export const CRDCardRow = connect(crdCardRowStateToProps)((props: CRDCardRowProps) => {
   const internalObjects = getInternalObjects(props.csv);
   const crds = props.crdDescs?.filter(({ name }) => !isInternalObject(internalObjects, name));
+  // if the CSV was created within the last 5 minutes (300000 milliseconds)
+  const recentlyCreated = moment(Date.now()).diff(props.csv.metadata.creationTimestamp) < 300000;
```
spadgett, line 770: Isn't 30,000ms == 30s? I would use `5 * 60 * 1000`.

I wonder if we should check the date at all since we could have clock drift between the client and server. I think we should a similar message elsewhere in the UI without the date check.

spadgett: https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx#L420

spadgett: @rhamilto Disregard, I missed a `0` :)

rhamilto:
@itsptk and @tlwu2013, y'all good with this approach?

itsptk:
@rhamilto I like it 👍, just curious:

Would the 'Create instance' button appear when the CRD is actually created then? (The same time the Foo tab stops saying "The server doesn't have a resource type Foo.  Try refreshing the page if it was recently added." ?) Would the user need to refresh the Details for that 'create instance' button to appear?

spadgett:
> @rhamilto I like it 👍, just curious:
> 
> Would the 'Create instance' button appear when the CRD is actually created then? (The same time the Foo tab stops saying "The server doesn't have a resource type Foo. Try refreshing the page if it was recently added." ?) Would the user need to refresh the Details for that 'create instance' button to appear?

Hi @itsptk. They should, although there might be a little bit of a delay for normal users. If it doesn't, I'd consider that a bug.

spadgett (approved):
/lgtm

rhamilto:
> Would the 'Create instance' button appear when the CRD is actually created then? (The same time the Foo tab stops saying "The server doesn't have a resource type Foo. Try refreshing the page if it was recently added." ?) Would the user need to refresh the Details for that 'create instance' button to appear?

Yes, the create button should appear automatically.  I haven't tested it, but in looking at the code, it should "React".

PR #7394, opened 2020-12-02
kubevirt i18n Actions

yaacov:
kubevirt add actions i18n


frontend/packages/kubevirt-plugin/src/topology/components/kubevirtComponentFactory.ts
```diff
@@ -45,6 +45,7 @@ export const vmActions = (
       return action(model, contextMenuResource, {
         vmi,
         vmStatusBundle,
+        t: (t) => t, // FIXME(yaacov): this should be t: TFunction (e.g. useTranslation or i18n.t)
```
yaacov, line 48: @pcbailey help,  i tried using `i18n.t` here, but it was undefined :-( what am i missing ?

yaacov:
@pcbailey @rawagner  please review

PR #7396, opened 2020-12-02
Add support for samples & snippets in pipeline builder yaml editor

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5102

**Analysis / Root cause**: `YAMLEditorField` didn't support samples and snippets in the sidebar.

**Solution Description**: This PR - 
- Adds new `YAMLEditorSidebar` component that uses`ResourceSidebar` to show schema, samples and snippets. This component also manages loading samples, inserting snippets and downloading yamls,
- Refactors `EditYaml` component to use the above component.
- Adds support for using `YAMLEditorSidebar` in `YAMLEditorField` based on the availability of sidebar contents.
- Other cleanup and related refactoring.

@openshift/team-devconsole-ux 

YAML Editor

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/console-shared/src/utils/sample-utils.ts
```diff
@@ -0,0 +1,263 @@
...
+};
+
+const getTargetResource = (model: K8sKind) => ({
+  apiVersion: apiVersionForModel(model),
+  kind: model.kind,
+});
+
+const clusterRoleBindingSamples: Sample[] = [
+  {
+    title: 'Allow reading Nodes in the core API groups (for ClusterRoleBinding)',
```
andrewballantyne, line 49: All these should be translated... but we can probably run through this as a bug if needed

andrewballantyne:
Everything else LGTM. I say let's get this in now and then swing back with the translations. We are at too high of a count right now so let's see this merged and swing back with translations as another PR.

/lgtm

PR #7397, opened 2020-12-02
Bug 1901531: Fix runtime error in create CR form

TheRealJon:
In some cases, an uncaught exception could be thrown from dynamically generated array form fields.
This was happening when a nested array field had an explicitly set 'null' value. This is a
workaround that adds a fallback value to object and array fields to prevent this runtime error. An
issue and fix have also been opened upstream in the react-jsonschema-form repo.
https://github.com/rjsf-team/react-jsonschema-form/issues/2153
https://github.com/rjsf-team/react-jsonschema-form/pull/2154


spadgett (approved):
/lgtm

PR #7398, opened 2020-12-02
remove vm templates from template catalog

christianvogt:
**Fixes**: 
fixes https://issues.redhat.com/browse/ODC-5197

**Analysis / Root cause**: 
VM templates were showing up in the standalone Templates catalog but not in the general dev catalog.

**Solution Description**: 
Add catalog filter extension allowing kubevirt plugin to filter out all Templates which have a label identifying them as vm templates.

**Test setup:**

- install OpenShift Virtualization operator
- configure operator
- create VM template (creation via yaml is easiest)
- go to dev catalog and search for the vm template
- perform the same search in the Template and VM Template catalogs

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

cc @rohitkrai03

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/dev-catalog.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/dev-catalog.ts
@@ -31,6 +31,13 @@ namespace ExtensionProperties {
         items provided by other providers. */
     priority?: number;
   }
+
+  export interface CatalogItemFilter {
+    /** Type ID for the catalog item type. */
+    type: string;
+    /** Filters items of a specific type. Value is a function that takes CatalogItem[] and returns a subset based on the filter criteria. */
+    filter: CodeRef<<T extends CatalogItem[]>(items: T) => T>;
+  }
 }
 
 export interface DevCatalogModel extends Extension<ExtensionProperties.DevCatalogModel> {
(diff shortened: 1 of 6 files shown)


frontend/packages/console-plugin-sdk/src/typings/dev-catalog.ts
```diff
@@ -31,6 +31,13 @@ namespace ExtensionProperties {
         items provided by other providers. */
     priority?: number;
   }
+
+  export interface CatalogItemFilter {
+    /** Type ID for the catalog item type. */
+    type: string;
+    /** Fetch items and normalize it for the catalog. Value is a react effect hook. */
```
rohitkrai03, line 38:
```suggestion
    /** Filters items of a specific type. Value is a function that takes CatalogItem[] and returns a subset of it based on a filter criteria. */
```

rohitkrai03 (changes requested):

frontend/packages/console-plugin-sdk/src/typings/dev-catalog.ts
christianvogt: fixed

rohitkrai03 (approved):
/lgtm

PR #7399, opened 2020-12-03
Bug 1879638: Support binary file type secret entries

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1879638

1. read in secret file using FileReader
2. test if contents contain non printable characters
3. if so reread file as binary data
4. do not show text area displaying contents of binary files
5. use data field instead of stringData for all secrets and base64 all data
6. modify the details page to also not show binary data
cc: @jhadvig 

binary-encode


frontend/public/components/utils/file-input.tsx
```diff
@@ -28,13 +37,20 @@ export class FileInput extends React.Component<FileInputProps, FileInputState> {
...
-        fileData: input,
-        fileName: file.name,
-      });
+      const input = reader.result as string;
+      if (containsNonPrintableCharacters(input) && !fileIsBinary) {
+        fileIsBinary = true;
+        reader.readAsBinaryString(file);
+      } else {
+        this.props.onChange({
+          fileData: fileIsBinary ? Base64.encode(input) : input,
```
spadgett, line 49: I wouldn't Base64 encode here. For secrets, the value is always encoded whether it's binary or not. I'd let the caller encode the value.

frontend/public/components/utils/file-input.tsx
```diff
@@ -28,13 +37,20 @@ export class FileInput extends React.Component<FileInputProps, FileInputState> {
...
+    let fileIsBinary = false;
     const reader = new FileReader();
     reader.onload = () => {
-      const input = reader.result;
-      this.props.onChange({
-        fileData: input,
-        fileName: file.name,
-      });
+      const input = reader.result as string;
+      if (containsNonPrintableCharacters(input) && !fileIsBinary) {
```
spadgett, line 44: Why is the `!fileIsBinary` check needed?

frontend/public/components/utils/file-input.tsx
```diff
@@ -151,25 +167,33 @@ export const DroppableFileInput = withDragDropContext(
...
-          () => this.props.onChange(input),
-        );
+        if (containsNonPrintableCharacters(input) && !inputFileIsBinary) {
+          inputFileIsBinary = true;
+          reader.readAsBinaryString(file);
+        } else {
+          this.setState(
+            {
+              inputFileName: file.name,
+              inputFileData: inputFileIsBinary ? Base64.encode(input) : input,
```
spadgett, line 181: Again, I wouldn't Base64 encode here

frontend/public/components/utils/file-input.tsx
```diff
@@ -151,25 +167,33 @@ export const DroppableFileInput = withDragDropContext(
...
+              inputFileIsBinary,
```
spadgett, line 182: We need to handle this in the caller if we're not already. If this is the import YAML page, we should show an error saying the file contains unprintable characters. For secrets, we should allow the value but show a message saying the file contains unprintable characters and not preview the content.

frontend/public/components/utils/file-input.tsx
jcaianirh: @spadgett we start out assuming we can readAsText, but when we come here after readingAsBinary, we no longer need to drop down into this code or we'll loop.

frontend/public/components/utils/file-input.tsx
jcaianirh: @spadgett ok, removed base64 encoding here.

spadgett:
Hey, @jcaianirh this PR has a pending review comment (when you have time to look)

frontend/public/components/utils/file-input.tsx
jcaianirh: @spadgett Yes, the import yaml page indeed shows the message you mention above. I added the message about non printable characters right in the file-input.

frontend/public/components/utils/file-input.tsx
```diff
@@ -94,7 +111,7 @@ export class FileInput extends React.Component<FileInputProps, FileInputState> {
                 }
                 className="pf-c-form-control co-file-dropzone__textarea"
                 onChange={this.onDataChange}
-                value={this.props.inputFileData}
+                value={fileIsBinary ? fileIsBinaryMsg : this.props.inputFileData}
```
spadgett, line 114: I think it would be better to hide the textarea instead of putting the message inside the textarea. (For one, there's no way to tell that the message is not the contents of the file you picked.)

spadgett:

frontend/public/components/utils/file-input.tsx
```diff
@@ -9,6 +9,15 @@ import withDragDropContext from './drag-drop-context';
 // Maximal file size, in bytes, that user can upload
 const maxFileUploadSize = 4000000;
 const fileSizeErrorMsg = 'Maximum file size exceeded. File limit is 4MB.';
+const fileIsBinaryMsg = 'File contains non-printable characters. Preview is not available.';
```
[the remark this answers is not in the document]
jcaianirh: added translations here.

frontend/public/components/utils/file-input.tsx
jcaianirh: hide text area, and add an info alert that the text contains non-printable characters.

jcaianirh:
@spadgett responded to review comments

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -154,7 +155,8 @@ export const withSecretForm = (SubForm, modal?: boolean) =>
     }
     onDataChanged(secretsData) {
       this.setState({
-        stringData: { ...secretsData.stringData },
+        stringData: { ...secretsData?.stringData },
+        base64Data: { ...secretsData?.base64Data },
```
spadgett, line 159: I would just have one `data` state property and always base64 encode the data. It keeps the logic simpler and works for plain strings, too.

frontend/public/components/utils/file-input.tsx
```diff
@@ -3,14 +3,26 @@ import * as classNames from 'classnames';
...
+import i18next from 'i18next';
+/* eslint-disable import/named */
+import { WithTranslation, withTranslation } from 'react-i18next';
 
 import withDragDropContext from './drag-drop-context';
 
 // Maximal file size, in bytes, that user can upload
 const maxFileUploadSize = 4000000;
-const fileSizeErrorMsg = 'Maximum file size exceeded. File limit is 4MB.';
+const fileSizeErrorMsg = i18next.t('utils~Maximum file size exceeded. File limit is 4MB.');
```
spadgett, line 15: Don't use `i18next.t` outside of a component or function. This message should be moved into the component.

spadgett (commented):
@jcaianirh Would you mind including a screenshot?

frontend/public/components/secrets/create-secret.tsx
jcaianirh: ok, sounds good, updated to have only one data state @spadgett

frontend/public/components/utils/file-input.tsx
jcaianirh: done thanks

frontend/public/components/utils/file-input.tsx
```diff
@@ -139,46 +164,56 @@ export const DroppableFileInput = withDragDropContext(
...
         const input = reader.result as string; // Note(Yaacov): we use reader.readAsText
-        this.setState(
-          {
-            inputFileName: file.name,
-            inputFileData: input,
-            errorMessage: '',
-          },
-          () => this.props.onChange(input),
-        );
+        if (containsNonPrintableCharacters(input) && !inputFileIsBinary) {
```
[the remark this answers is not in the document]
jcaianirh: @jhadvig added comment.

frontend/public/components/utils/file-input.tsx
```diff
@@ -24,19 +32,27 @@ class FileInput_ extends React.Component<FileInputProps, FileInputState> {
...
+    let fileIsBinary = false;
     const reader = new FileReader();
     reader.onload = () => {
-      const input = reader.result;
-      this.props.onChange({
-        fileData: input,
-        fileName: file.name,
-      });
+      const input = reader.result as string;
+      if (containsNonPrintableCharacters(input) && !fileIsBinary) {
```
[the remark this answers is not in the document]
jcaianirh: @jhadvig added comment

jhadvig (approved):
@jcaianirh really nice work! 🥇 
/lgtm
/approve

spadgett:
@jcaianirh We need some automated tests around this since it's easy to get things wrong and not something often tested manually. I think it would be good to test:

1. Create a secret with a binary file in the UI. Decode the secret using the CLI. Ensure that the file is identical to before.
2. Create a text secret with Unicode content. Make sure the encoding is correct when viewed in the UI and decoded in the CLI.

Do you mind creating an issue to track this?

jcaianirh:
@spadgett here is the tracking issue: https://issues.redhat.com/browse/CONSOLE-2789

PR #7402, opened 2020-12-03
Bug 1909610: Fix available capacity when no storage class selected

cloudbehl:
1. when no storage class is selected it shows `Not Available`.
2. pass persistent volume data from the parent component and remove the get call. 

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/pvs-available-capacity.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/pvs-available-capacity.tsx
@@ -22,11 +22,7 @@ export const PVsAvailableCapacity: React.FC<PVAvaialbleCapacityProps> = ({
     <div className="skeleton-text ceph-pvs-available-capacity__current-capacity--loading" />
   );
 
-  if ((loadError || data.length === 0 || !storageClass) && loaded) {
-    availableStatusElement = (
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/pvs-available-capacity.tsx
```diff
@@ -15,18 +15,14 @@ export const PVsAvailableCapacity: React.FC<PVAvaialbleCapacityProps> = ({
...
   let availableStatusElement = (
     <div className="skeleton-text ceph-pvs-available-capacity__current-capacity--loading" />
   );
 
-  if ((loadError || data.length === 0 || !storageClass) && loaded) {
-    availableStatusElement = (
-      <div className="text-muted">{t('ceph-storage-plugin~Not Available')}</div>
-    );
-  } else if (loaded) {
+  if (loaded && !!getName(storageClass)) {
```
afreen23, line 25: We should also check for `loadError` since when `loadError` is true, loaded` is also `true`.
In that scenario, when you have a pre selected storage class, it will pass this check.

afreen23 (changes requested):

afreen23 (approved):
You probably need to update i18n files to fix up the failing tests

afreen23:
/lgtm
needs BZ as well

PR #7410, opened 2020-12-03
Use extensions to add decorators to topology nodes

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4536

**Solution Description**: 
Add the ability to add node decorators thru extensions.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind cleanup
/kind feature


frontend/packages/knative-plugin/src/topology/components/groups/KnativeServiceGroup.tsx
```diff
@@ -110,6 +108,17 @@ const KnativeServiceGroup: React.FC<KnativeServiceGroupProps> = ({
...
+    element,
+    element.getGraph().getData().decorators,
+    x + width / 2,
+    y + height / 2,
+    -1,
+    DECORATOR_RADIUS,
+    width,
+    height,
+  );
+
```
christianvogt, line 121: Convert to a component `<Decorators ...`

In there you can use `useResolvedExtensions` to fetch the decorators instead of setting them in the graph data.

frontend/packages/topology/src/components/graph-view/components/nodes/decorators/EditDecorator.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { editURL, vcsURI, vcsRef } = workloadData;
+  const cheEnabled = !!cheURL;
+  const editUrl = editURL || getEditURL(vcsURI, vcsRef, cheURL);
+  const repoIcon = routeDecoratorIcon(editUrl, radius, t, cheEnabled);
+
+  if (!repoIcon) {
+    return null;
+  }
+  return (
+    <Tooltip key="edit" content={t('topology~Edit source code')} position={TooltipPosition.right}>
```
christianvogt, line 37: Decorators shouldn't have to worry about a key.

frontend/packages/knative-plugin/src/topology/components/groups/KnativeServiceGroup.tsx
jeff-phillips-18: I believe it would be too expensive to resolve extensions (and sort, etc) for every node instance.

christianvogt: After reviewing `useResolvedExtensions`, there is no cache yet. Brought this up with Vojtech to improve.

Still value in the Decorators component but we can tackle it later.

andrewballantyne:
/test analyze

Memory issue, flake.

PR #7411, opened 2020-12-03
convert last used namespace local storage to userSettings

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5111

This PR move the `bridge/last-namespace-name` local storage to the user settings.
No UI changes.

### DIFF
--- a/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
+++ b/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
@@ -1,0 +1,51 @@
+import * as React from 'react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
+import { useDispatch } from 'react-redux';
+import { useLocation } from 'react-router-dom';
+import { getNamespace } from '@console/internal/components/utils/link';
+import { useUserSettingsCompatibility } from '@console/shared/src/hooks/useUserSettingsCompatibility';
+import { setActiveNamespace } from '@console/internal/actions/ui';
+import { ALL_NAMESPACES_KEY } from '@console/shared/src/constants';
+
+type NamespaceContextType = {
+  namespace?: string;
+  setNamespace?: (ns: string) => void;
+};
+
+const LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY = `bridge/last-namespace-name`;
+const LAST_NAMESPACE_NAME_USER_SETTINGS_KEY = 'console.lastNamespace';
+
+export const NamespaceContext = React.createContext<NamespaceContextType>({});
+
+export const useValuesForNamespaceContext = () => {
+  const { pathname } = useLocation();
+  const urlNamespace = getNamespace(pathname);
+  /**
+   * [TODO]: use favorite namespace here if there is any
+   *  const [favorite] = useUserSettings()
+   */
+  const [namespace, setNs, loaded] = useUserSettingsCompatibility(
+    LAST_NAMESPACE_NAME_USER_SETTINGS_KEY,
+    LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
+    urlNamespace || ALL_NAMESPACES_KEY,
+  );
+  const dispatchreduxAction = useDispatch();
+  const setNamespace = React.useCallback(
+    (ns: string) => {
+      dispatchreduxAction(setActiveNamespace(ns));
+      setNs(ns);
+    },
+    [dispatchreduxAction, setNs],
+  );
+
+  React.useEffect(() => {
+    if (loaded) {
+      dispatchreduxAction(setActiveNamespace(urlNamespace || namespace || ALL_NAMESPACES_KEY));
+    }
+    // only run this hook when namespace is loaded from user settings
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loaded]);
+
+  return { namespace: urlNamespace || namespace || ALL_NAMESPACES_KEY, setNamespace, loaded };
+};
--- a/frontend/packages/console-shared/src/hoc/withLastNamespace.tsx
+++ b/frontend/packages/console-shared/src/hoc/withLastNamespace.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { useActiveNamespace } from '../hooks';
+
+type WithLastNamespaceProps = {
+  activeNamespace?: string;
+  setActiveNamespace?: (v: string) => void;
+};
+
+export const withLastNamespace = <Props extends WithLastNamespaceProps>(
+  WrappedComponent: React.ComponentType<Props>,
+): React.FC<Omit<Props, keyof WithLastNamespaceProps>> => (props: Props) => {
+  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+  return (
+    <WrappedComponent
+      {...props}
+      activeNamespace={activeNamespace}
+      setActiveNamespace={setActiveNamespace}
+    />
+  );
+};
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
@@ -2,6 +2,10 @@
 import * as _ from 'lodash';
 import * as React from 'react';
 import { connect } from 'react-redux';
+import { Prompt } from 'react-router';
+import { useTranslation } from 'react-i18next';
+import { useActiveNamespace } from '@console/shared';
+import { getActiveNamespace } from '@console/internal/actions/ui';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Wizard/wizard';
 import {
@@ -11,7 +15,6 @@ import {
   WizardContextConsumer,
   WizardStep,
 } from '@patternfly/react-core';
-import { Prompt } from 'react-router';
 import { useShowErrorToggler } from '../../hooks/use-show-error-toggler';
 import { getDialogUIError, getSimpleDialogUIError } from '../../utils/strings';
 import {
@@ -27,14 +30,12 @@ import {
   isWizardEmpty as _isWizardEmpty,
 } from './selectors/immutable/wizard-selectors';
 import { iGetCommonData } from './selectors/immutable/selectors';
-import { setActiveNamespace, getActiveNamespace } from '@console/internal/actions/ui';
 import { vmWizardActions } from './redux/actions';
 import { ActionType } from './redux/types';
 import { getGoToStep } from './selectors/selectors';
 import { iGetLoadError, iGetIsLoaded } from '../../utils/immutable';
 
 import './create-vm-wizard-footer.scss';
-import { useTranslation } from 'react-i18next';
 
 type WizardContext = {
   onNext: () => void;
@@ -53,7 +54,6 @@ type CreateVMWizardFooterComponentProps = {
   isSimpleView: boolean;
   isInvalidUserTemplate: boolean;
   onEdit: (activeStepID: VMWizardTab) => void;
-  setActiveNS: (ns: string) => void;
 };
 
 const CreateVMWizardFooterComponent: React.FC<CreateVMWizardFooterComponentProps> = ({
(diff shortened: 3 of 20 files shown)


frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  const [namespace, setNs, loaded] = useUserSettingsCompatibility(
+    LAST_NAMESPACE_NAME_USER_SETTINGS_KEY,
+    LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
+    '',
+  );
+  const dispatchreduxAction = useDispatch();
+  const setNamespace = React.useCallback(
+    (ns: string) => {
+      dispatchreduxAction(setActiveNamespace(ns));
```
rohitkrai03, line 36: If I understand this correctly, we're putting `activeNamespace` in a top level context provider. And also in redux? if it's already in top level context do we need to keep it in redux store as well? I am confused because i see in some places we're using value from redux and in other places using value directly from context.

frontend/packages/console-shared/src/hoc/withLastNamespace.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export type WithLastNamespaceProps = {
+  namespace: string;
+  setActiveNamespace: (v: string) => void;
+};
+
+export const withLastNamespace = <Props extends WithLastNamespaceProps>(
+  WrappedComponent: React.ComponentType<Props>,
+): React.FC<Omit<Props, keyof WithLastNamespaceProps>> => (props: Props) => {
+  const [state, setState] = useLastNamespace();
```
rohitkrai03, line 12:
```suggestion
  const [activeNamespace, setActiveNamespace] = useLastNamespace();
```

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -65,13 +65,12 @@ const CreateVMWizardFooterComponent: React.FC<CreateVMWizardFooterComponentProps
   isSimpleView,
   goToStep,
   onEdit,
-  setActiveNS,
   isInvalidUserTemplate,
 }) => {
   const { t } = useTranslation();
   const [showError, setShowError, checkValidity] = useShowErrorToggler();
+  const [, setActiveNS] = useLastNamespace();
   const activeNS = getActiveNamespace();
```
rohitkrai03, line 73: Why not use the value from the context here?

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
sahil143: Initial idea was that only, but in some of the utils getActiveNamespace is being called which access the store directly `store.get(...)`. Didn't want to refactor everything on the last day 😞 but Yeah I agree with you using the context data everywhere and not use redux state.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
sahil143: missed this one!

sahil143: fixed!

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
rohitkrai03: I am okay with mixed usage for now since it's the last day of feature freeze. We should definitely look at removing all usage of redux state later.

frontend/public/components/app.jsx
```diff
@@ -45,6 +45,7 @@ const NOTIFICATION_DRAWER_BREAKPOINT = 1800;
 // Edge lacks URLSearchParams
 import 'url-search-params-polyfill';
 import { graphQLReady } from '../graphql/client';
+import DetectNamespace from '@console/app/src/components/detect-namespace/DetectNamespace';
```
rohitkrai03, line 48: Nit: sort imports.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -27,14 +29,13 @@ import {
   isWizardEmpty as _isWizardEmpty,
 } from './selectors/immutable/wizard-selectors';
 import { iGetCommonData } from './selectors/immutable/selectors';
-import { setActiveNamespace, getActiveNamespace } from '@console/internal/actions/ui';
+import { getActiveNamespace } from '@console/internal/actions/ui';
```
rohitkrai03, line 32: Nit: sort imports.

rohitkrai03 (changes requested):

rohitkrai03 (approved):
Couple of nits. But otherwise looks good. Verified locally.
/approve

frontend/packages/console-shared/src/hoc/withLastNamespace.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { useLastNamespace } from '../hooks';
+
+export type WithLastNamespaceProps = {
```
vikram-raj, line 4: Do we need to export this type?

sahil143: Nope! fixed!

frontend/packages/console-shared/src/hooks/useLastNamespace.ts
```diff
@@ -0,0 +1,7 @@
+import { NamespaceContext } from '@console/app/src/components/detect-namespace/namespace';
+import { useContext } from 'react';
```
rottencandy, line 2: Nit: missed this

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.tsx
```diff
@@ -12,6 +12,9 @@ import {
   WizardStep,
 } from '@patternfly/react-core';
 import { Prompt } from 'react-router';
+import { useTranslation } from 'react-i18next';
+import { useLastNamespace } from '@console/shared';
```
rottencandy, line 16: Same here

vikram-raj (commented):
I noticed an issue, when I select a project from the projects page and switch to another page. I noticed that URL ns is different than the projects dropdown.

Kapture 2020-12-04 at 15 57 11

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
andrewballantyne: @sahil143 please log a ticket for this so it does not get lost.

rottencandy:
Tested locally, looks good.
cc: @vikram-raj @rohitkrai03

frontend/public/actions/ui.ts
```diff
@@ -204,7 +201,6 @@ export const setActiveNamespace = (namespace: string = '') => {
     }
     // remember the most recently-viewed project, which is automatically
     // selected when returning to the console
```
andrewballantyne, line 203: Nit: Dead comments? Remove

debsmita1:
verified this locally, works fine
/lgtm

andrewballantyne (commented):
I was mid review :D haha... but this works. Please log the needed ticket to fix the remaining use of Redux (directly 😨 ) and we get this merged for FF.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -0,0 +1,59 @@
...
+    [dispatchreduxAction, setNs],
+  );
+
+  React.useEffect(() => {
+    if (loaded) {
+      setNamespace(urlNamespace || namespace || ALL_NAMESPACES_KEY);
+    }
+    // only run this hook when namespace is loaded from user settings
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loaded]);
```
christianvogt, line 48: By using `useEffect` here, the first time `loaded === true`, the app will render with the namespace from the settings which may differ from the `urlNamespace`. After the effect runs, the namespace will change causing a flash from old state to new.

If you set the default value in `useUserSettingsCompatibility`, the settings will always either have a namespace or a default.

Therefore at this point you need to decide to return the `urlNamespace` or the settings `namespace`.

The effect afterwards will update the settings based on the `urlNamespace`.

You should be able to remove this and update the return value. 

```suggestion
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  React.useEffect(() => {
+    // when namespace is changed through url
+    if (loaded && namespace && urlNamespace && urlNamespace !== namespace) {
+      setNamespace(urlNamespace);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [urlNamespace]);
+
+  return { namespace, setNamespace, loaded };
```
christianvogt, line 58:
```suggestion
  return { namespace: urlNamespace || namespace, setNamespace, loaded };
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -0,0 +1,59 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loaded]);
+
+  React.useEffect(() => {
+    // when namespace is changed through url
+    if (loaded && namespace && urlNamespace && urlNamespace !== namespace) {
+      setNamespace(urlNamespace);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [urlNamespace]);
```
christianvogt, line 56: You need to also check if loaded has changed.
```suggestion
  }, [urlNamespace, loaded]);
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (loaded) {
+      setNamespace(urlNamespace || namespace || ALL_NAMESPACES_KEY);
+    }
+    // only run this hook when namespace is loaded from user settings
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loaded]);
+
+  React.useEffect(() => {
+    // when namespace is changed through url
+    if (loaded && namespace && urlNamespace && urlNamespace !== namespace) {
```
christianvogt, line 52: Not sure why you'd need to ensure `namspace` has a value.
```suggestion
    if (loaded && urlNamespace && urlNamespace !== namespace) {
```

frontend/packages/console-shared/src/hooks/useLastNamespace.ts
```diff
@@ -0,0 +1,7 @@
...
+import { NamespaceContext } from '@console/app/src/components/detect-namespace/namespace';
+
+export const useLastNamespace = (): [string, (ns: string) => void] => {
```
christianvogt, line 4: Probably should call this `useActiveNamespace` since that's the current term devs will be familiar with.

frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -1,10 +1,10 @@
+import i18next from 'i18next';
 import { K8sKind } from '../../module/k8s';
-import { LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY } from '@console/shared/src/constants';
 import { ALL_NAMESPACES_KEY } from '@console/shared/src/constants/common';
-import i18next from 'i18next';
+import { getActiveNamespace } from '../../actions/ui';
 
 export const getBreadcrumbPath = (match: any, customPlural?: string) => {
-  const lastNamespace = localStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
+  const lastNamespace = getActiveNamespace();
```
christianvogt, line 7: Log a ticket to remove the usage of `actions/ui#getActiveNamespace`

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
christianvogt: OOps `setNamespace` dispatches and `setNs` does not. We will still need to dispatch to sync redux.

sahil143:
Jira ticket to track the redux state removal https://issues.redhat.com/browse/ODC-5210

andrewballantyne:
/retest

Supposedly the issue has been fix with the CI 🤞

andrewballantyne:
Unsure why this test failed `crud/k8-openshift-cruds.spec.ts`.

/retest

andrewballantyne:
🤔  This is a repeat of the same error... one last retest and then I'll investigate more.

/retest

andrewballantyne:
Had a secret failure... think it's an infra flake... going to retest.

/retest

PR #7413, opened 2020-12-04
Bug 1903172: Background shadow appears under datalist in column management modal

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1903172

Make datalist background color transparent so the background shadow of the modal content appears when dialog is vertically shrunk.

cc: @gdoyle1 

Before:
columnManagementBug
After:
Screen Shot 2020-12-04 at 12 10 57 AM

### DIFF
--- a/frontend/public/style/_overrides.scss
+++ b/frontend/public/style/_overrides.scss
@@ -207,6 +207,12 @@ h6 {
 .pf-c-chip-group.pf-m-toolbar {
   margin-bottom: var(--pf-global--spacer--xs);
 }
+
+//set pf datalist item background to transparent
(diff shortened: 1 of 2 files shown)


gdoyle1:
Looks great!! Thanks @jcaianirh

frontend/public/style/_overrides.scss
```diff
@@ -273,6 +273,11 @@ h6 {
   }
 }
 
+//set pf datalist item background to transparent
```
rhamilto, line 276: Nit.  We try to alphabetize rules by selector.  So this should move to line 210.

PR #7414, opened 2020-12-04
 Bug 1904305: add scroll for monitoring dashboard dropdown menu 

yapei:
Before fix:
Screen Shot 2020-12-04 at 2 18 02 PM

After fix: there is scroll bar for long dropdown list 
Screen Shot 2020-12-04 at 2 17 54 PM

### DIFF
--- a/frontend/public/components/monitoring/_monitoring.scss
+++ b/frontend/public/components/monitoring/_monitoring.scss
@@ -205,6 +205,11 @@ $screen-phone-landscape-min-width: 567px;
   overflow-x: auto;
 }
 
+.monitoring-dashboards__variable-dropdown .pf-c-dropdown__menu {
+  max-height: 60vh;
+  overflow-y: auto;
+}
+
 .monitoring-dashboards__variables {
   display: flex;
   flex-wrap: wrap;
--- a/frontend/public/components/monitoring/dashboards/index.tsx
+++ b/frontend/public/components/monitoring/dashboards/index.tsx
@@ -103,6 +103,7 @@ const VariableDropdown: React.FC<VariableDropdownProps> = ({
               {items[selectedKey]}
             </DropdownToggle>
           }
+          className="monitoring-dashboards__variable-dropdown"
         />
       )}
     </div>


yapei:
@jhadvig  Can you help review?

yapei:
@spadgett Could you help review? Thanks

frontend/public/components/_dropdown.scss
```diff
@@ -238,6 +238,8 @@ $dropdown-background--hover: $color-pf-black-200; // pf-c-dropdown__menu-item--h
   border-width: 0 !important; // TEMP fix until https://github.com/patternfly/patternfly-next/issues/2019 is fixed upstream
   list-style: none;
   -webkit-overflow-scrolling: touch;
+  max-height: 60vh;
+  overflow-y: auto;
```
spadgett, line 242: We try to keep the CSS rules alphabetized, so these would go after `list-style` above.

frontend/public/components/_dropdown.scss
```diff
@@ -238,6 +238,8 @@ $dropdown-background--hover: $color-pf-black-200; // pf-c-dropdown__menu-item--h
   border-width: 0 !important; // TEMP fix until https://github.com/patternfly/patternfly-next/issues/2019 is fixed upstream
```
spadgett, line 238: Unrelated to this fix, but can we remove this rule now @rhamilto @sg00dwin @jcaianirh ? (We should do that as a separate PR.)

frontend/public/components/_dropdown.scss
```diff
@@ -238,6 +238,8 @@ $dropdown-background--hover: $color-pf-black-200; // pf-c-dropdown__menu-item--h
...
   list-style: none;
   -webkit-overflow-scrolling: touch;
+  max-height: 60vh;
```
spadgett, line 241: I'd make this a variable since it's used for other dropdown heights. That way we always use the same max-height.

spadgett (commented):
Thanks @yapei 👍

frontend/public/components/_dropdown.scss
rhamilto: I think we probably want to scope this fix to the places we know it is needed.  We've had issues in the past sizing all dropdowns generically and ended up creating new bugs in untested contexts.

[vh units are problematic on mobile.](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/)  I suggest we avoid using them here.  We've had a number of bugs in the past trying to size dropdown menus in this way.  @sg00dwin , do you remember where we landed on an approach that worked in all contexts?

frontend/public/components/_dropdown.scss
rhamilto: Yes, this override can be removed now.  Anyone want to tackle that?

frontend/public/components/_dropdown.scss
yapei: ok, will update together with following comment

frontend/public/components/_dropdown.scss
yapei: @sg00dwin Could you help point me some history or PRs? Thank you

rhamilto: @yapei, we talked about this at scrum today.  `60vh` is probably ok to use in this instance, but please scope the fix to this one dropdown.  We don't want this fix applied to all dropdowns.

rhamilto: PR to @yapei branch with a way to scope this fix to just the monitoring dashboard dropdowns.
https://github.com/yapei/console/pull/1

yapei: Thank you so much @rhamilto

yapei:
@rhamilto  @spadgett  requested changes are updated, please help PTAL

rhamilto:
@sg00dwin, mind reviewing this since I assisted with the fix?

sg00dwin (approved):
Looks good

PR #7415, opened 2020-12-04
Fix error state for Raw Capacity Card

bipuladh:
- Align Error message for capacity breakdown card to the center
Screenshot from 2020-12-04 12-56-03

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/raw-capacity-card/raw-capacity-card.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/raw-capacity-card/raw-capacity-card.tsx
@@ -20,7 +20,8 @@ const colorScale = ['#0166cc', '#d6d6d6'];
 const parser = compose((val) => val?.[0]?.y, getInstantVectorStats);
 
 const RawCapacityCard: React.FC = React.memo(() => {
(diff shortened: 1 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/raw-capacity-card/raw-capacity-card.tsx
```diff
@@ -90,8 +91,8 @@ const RawCapacityCard: React.FC = React.memo(() => {
             </div>
           </>
         )}
-        {!loadError && loading && <LoadingCardBody />}
-        {loadError && !loading && <ErrorCardBody />}
+        {loading && !loadError && <LoadingCardBody />}
```
cloudbehl, line 94: In case of all undefined values can't we should not available?

bipuladh: In case of all `undefined` values `loadError` will be `true` and this block will not run. Line 95 with `ErrorCardBody` having `Not Available` will get activated.

cloudbehl: righto!

PR #7420, opened 2020-12-04
Move gitops code out into its own monorepo package

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-4916

**Solution Description**: 
Created a new `gitops-plugin` package, moved gitops code from `dev-console` package there.

Non-functional change.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/gitops-plugin/src/plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+  {
+    type: 'NavItem/Href',
+    properties: {
+      id: 'environments',
+      perspective: 'dev',
+      section: 'resources',
+      insertAfter: 'pipelines',
+      insertBefore: 'helm',
+      componentProps: {
+        // t('gitops-plugin~Environments')
```
divyanshiGupta, line 37: remove the commented code.

divyanshiGupta (changes requested):
Code looks good to me. Will add lgtm after testing it out locally.

divyanshiGupta:
Verified it out locally, works fine for me. 
/lgtm

PR #7424, opened 2020-12-04
Bug 1908381: Update devfile handler

maysunfaisal:
This PR is a continuation of https://github.com/openshift/console/pull/7299

This PR updates:
- openshift/api to release 4.6 branch & dependencies to newer versions in go.mod and the vendor dir
- devfile handler to read relative dockerfile location from devfile.yaml & filters devfile container component

To test this PR, try either of the example repo and its devfile - https://github.com/maysunfaisal/node-bulletin-board-2 or https://github.com/maysunfaisal/node-bulletin-board-3


go.mod
```diff
@@ -6,27 +6,27 @@ require (
...
 	github.com/coreos/go-oidc v2.1.0+incompatible
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
-	github.com/devfile/api v0.0.0-20201103130402-29b8738e196e
-	github.com/devfile/library v0.0.0-20201125155652-6eabeab3ee52
+	github.com/devfile/api v0.0.0-20201126204309-ec222215253e
+	github.com/devfile/library v0.0.0-20201203185121-a216646cf21d
 	github.com/gorilla/websocket v1.4.2
 	github.com/graph-gophers/graphql-go v0.0.0-20200309224638-dae41bde9ef9
-	github.com/openshift/api v3.9.0+incompatible
+	github.com/openshift/api v0.0.0-20200930075302-db52bc4ef99f
```
andrewballantyne, line 13: @jhadvig Here is the dependency up for my Devfile work 😄

andrewballantyne:
Thanks for logging this and working through the dependencies @maysunfaisal !

pkg/server/devfile-handler.go
```diff
@@ -50,33 +45,53 @@ func (s *Server) devfileHandler(w http.ResponseWriter, r *http.Request) {
...
-	containerComponents := devfileObj.Data.GetDevfileContainerComponents() //TODO: filter thru Console attribute, right now if there is more than one component container err out
+	filterOptions := common.DevfileOptions{
+		Filter: map[string]interface{}{
+			"tool": "console-import",
+		},
+	}
+
+	containerComponents, err := devfileObj.Data.GetDevfileContainerComponents(filterOptions) //For Dev Preview, if there is more than one component container err out
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to get the container component from devfile with attribute 'tool: console-import': %v", err)
```
andrewballantyne, line 56: @serenamarie125 This will require the owner of the Devfile is to add a couple lines to their Devfile in order for the UI to pick it up. This is so the Devfile can have multiple components but for Dev Preview we only deal with a singular targeted one. The error the user will see is in a red "danger" Alert of:

> Failed to get the container component from devfile with attribute 'tool: console-import'

Plus any error contents that may accompany the error from the library. I'll try to get a screenshot of this in a bit.

pkg/server/devfile-handler.go
```diff
@@ -50,33 +45,53 @@ func (s *Server) devfileHandler(w http.ResponseWriter, r *http.Request) {
...
-	service, err := getService(devfileObj)
+	service, err := getService(devfileObj, filterOptions)
 	if err != nil {
 		errMsg := fmt.Sprintf("Failed to get service for the devfile: %v", err)
 		klog.Error(errMsg)
 		serverutils.SendResponse(w, http.StatusInternalServerError, serverutils.ApiError{Err: errMsg})
 		return
 	}
 
+	dockerfilePath := devfileObj.Data.GetMetadata().Attributes.GetString("alpha.build-dockerfile", &err)
```
andrewballantyne, line 84: @maysunfaisal -- not a Go expert by any means... but is the `err` not supposed to be a 2nd argument here? Or is the `&err` passing in a pointer that is updated inside?

```suggestion
	dockerfilePath, err := devfileObj.Data.GetMetadata().Attributes.GetString("alpha.build-dockerfile", &err)
```

maysunfaisal: That is correct. The err pointer is passed and updated if there is any err. Thats how the go module func was written.

maysunfaisal: We should also document the fact that dockerfile location is to be provided by attribute `"alpha.build-dockerfile"` - https://github.com/maysunfaisal/node-bulletin-board-3/blob/main/devfile.yaml#L6 @andrewballantyne @serenamarie125

andrewballantyne: Good point @maysunfaisal -- The fallback is co-located with the `devfile.yaml`, correct?

maysunfaisal: Well atm, if its not available, we err out by the pointer you mentioned..

Screen Shot 2020-12-04 at 2 53 14 PM

andrewballantyne: Good to note @maysunfaisal... @serenamarie125 this will be another required thing for them

andrewballantyne (approved):
UI Changes LGTM!

jhadvig (approved):
Backend changes look good 👍 
/lgtm

PR #7427, opened 2020-12-04
Bug 1903192: - Role name missing on create role binding form

cyril-ui-developer:
/assign @spadgett 

Screen Shot 2020-12-04 at 3 03 24 PM
Screen Shot 2020-12-04 at 3 03 44 PM

### DIFF
--- a/frontend/public/components/RBAC/role.jsx
+++ b/frontend/public/components/RBAC/role.jsx
@@ -33,9 +33,12 @@ const menuActions = [
   (kind, role) => ({
     label: 'Add Role Binding',
     href: `/k8s/${
-      role.metadata.namespace ? `ns/${role.metadata.namespace}` : 'cluster'
-    }/rolebindings/~new?rolekind=${roleKind(role)}&rolename=${role.metadata.name}${role.metadata
-      .namespace && `&namespace=${role.metadata.namespace}`}`,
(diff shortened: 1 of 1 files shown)


frontend/public/components/RBAC/role.jsx
```diff
@@ -33,9 +33,12 @@ const menuActions = [
   (kind, role) => ({
     label: 'Add Role Binding',
     href: `/k8s/${
-      role.metadata.namespace ? `ns/${role.metadata.namespace}` : 'cluster'
-    }/rolebindings/~new?rolekind=${roleKind(role)}&rolename=${role.metadata.name}${role.metadata
-      .namespace && `&namespace=${role.metadata.namespace}`}`,
+      role.metadata.namespace
+        ? `ns/${role.metadata.namespace}/rolebindings/~new?rolekind=${roleKind(role)}&rolename=${
+            role.metadata.name
+          }${role.metadata.namespace && `&namespace=${role.metadata.namespace}`}`
```
spadgett, line 39: We already check for `role.metadata.namespace` just above, so we shouldn't need to check again.

cyril-ui-developer: @spadgett Updated, thanks.

spadgett (approved):
/lgtm

PR #7430, opened 2020-12-04
CONSOLE-2457: Modify CatalogSource Details view and include new

harishsurf:
This feature improves the Catalog Source details page by: 

1. Adding new fields to the Details view:
Details_view

2. Adding a new Operators tab:
Operators_tab

3. Modifying the list view for PackageManifest under Home-> Search page: 
searchPage
Note: Sorting on "Latest Version" column is disabled since the versioning format would be different for different package manifests. I discussed this with @TheRealJon and agreed that the logic for sorting these versions is non-trivial.
cc @tlwu2013 @itsptk 


frontend/packages/operator-lifecycle-manager/src/plugin.tsx
```diff
@@ -290,6 +299,13 @@ const plugin: Plugin<ConsumedExtensions> = [
         (await import('./components/operator-hub/operator-hub-details')).OperatorHubDetailsPage,
     },
   },
+  {
+    type: 'Page/Resource/Details',
+    properties: {
+      model: models.CatalogSourceModel,
+      loader: async () => (await import('./components/catalog-source')).CatalogSourceDetailsPage,
```
spadgett, line 306: This should have a `webpackChunkName`

frontend/public/components/utils/details-item.tsx
```diff
@@ -36,14 +36,28 @@ export const PropertyPath: React.FC<{ kind: string; path: string | string[] }> =
...
-);
+type EditButtonProps = {
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  testId?: string;
+};
+
+const EditButton: React.SFC<EditButtonProps> = (props) => {
+  return (
+    <Button
+      variant="link"
```
spadgett, line 47: Not new, but this should have a `type="button"` so it never acts as a form submit button.

frontend/public/components/utils/details-item.tsx
```diff
@@ -36,14 +36,28 @@ export const PropertyPath: React.FC<{ kind: string; path: string | string[] }> =
...
+      isInline
+      onClick={props.onClick}
+      data-test={
+        `${props.testId}`
```
spadgett, line 51:
```suggestion
        props.testId
```

frontend/public/components/utils/details-item.tsx
```diff
@@ -36,14 +36,28 @@ export const PropertyPath: React.FC<{ kind: string; path: string | string[] }> =
...
-  props,
-) => (
-  <Button variant="link" isInline onClick={props.onClick}>
-    {props.children}
-    <PencilAltIcon className="co-icon-space-l pf-c-button-icon--plain" />
-  </Button>
-);
+type EditButtonProps = {
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  testId?: string;
```
spadgett, line 41: Console naming conventions would make this ID

```suggestion
  testID?: string;
```

spadgett (commented):
Thanks @harishsurf

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -112,77 +111,158 @@ const DefaultSourceKebab: React.FC<DefaultSourceKebabProps> = ({
...
+        <div className="col-sm-6 col-xs-12">
+          <div className="co-m-pane__body">
+            <DetailsItem
+              editAsGroup
+              label="Status"
+              obj={catalogSource}
+              path="status.connectionState.lastObservedState"
+            />
+            <DetailsItem label="Display Name" obj={catalogSource} path="spec.displayName">
+              {catalogSource?.spec.displayName !== '' ? catalogSource?.spec.displayName : '-'}
```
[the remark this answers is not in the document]
harishsurf: gotcha

frontend/packages/operator-lifecycle-manager/src/components/modals/edit-registry-poll-interval-modal.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {
+  withHandlePromise,
+  HandlePromiseProps,
+  Dropdown,
+} from '@console/internal/components/utils';
+import { CatalogSourceKind } from '../../types';
+import { CatalogSourceModel } from '../../models';
+import { Form, FormGroup } from '@patternfly/react-core';
+
+const availablePollIntervals = {
```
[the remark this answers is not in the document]
harishsurf: I have modified the code to add a new item to the dropdown in such a case

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -112,77 +111,145 @@ const DefaultSourceKebab: React.FC<DefaultSourceKebabProps> = ({
...
-          </dl>
+        <div className="col-sm-6 col-xs-12">
+          <div className="co-m-pane__body">
+            <DetailsItem
+              editAsGroup
+              label="Status"
+              obj={catalogSource}
+              path="status.connectionState.lastObservedState"
+            />
+            <DetailsItem label="Display Name" obj={catalogSource} path="spec.displayName" />
```
spadgett, line 169: nit: These should be sentence case "Display name," although we can change that when we localize.

spadgett (approved):
/lgtm
/hold for approvals 

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
harishsurf: Since it would be done as part of i18n, I kept this as-is for now

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -112,77 +111,145 @@ const DefaultSourceKebab: React.FC<DefaultSourceKebabProps> = ({
...
+};
+
+export const CatalogSourceDetailsPage: React.FC<CatalogSourceDetailsPageProps> = (props) => {
+  const [operatorHub, operatorHubLoaded, operatorHubLoadError] = useK8sWatchResource<
+    OperatorHubKind
+  >({
+    kind: referenceForModel(OperatorHubModel),
+    name: 'cluster',
+    isList: false,
+  });
```
TheRealJon, line 207: Since the OperatorHub resource is a singular, unique cluster configuration resource, we could probably abstract this out into a custom reusable hook like `useOperatorHubConfig`. I'd put this in the `./packages/operator-lifecycle-manager/src/utils` folder with the only other custom olm hook.

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -112,77 +111,145 @@ const DefaultSourceKebab: React.FC<DefaultSourceKebabProps> = ({
...
+
+  const isDefaultSource = React.useMemo(() => {
+    const defaultSource = operatorHub?.status?.sources?.find(
+      (source) =>
+        source.name === props.match.params.name &&
+        DEFAULT_SOURCE_NAMESPACE === props.match.params.ns,
+    );
+    return !_.isEmpty(defaultSource);
+  }, [operatorHub, props.match.params.name, props.match.params.ns]);
```
TheRealJon, line 216:
```suggestion
  const isDefaultSource = React.useMemo(() => (
    DEFAULT_SOURCE_NAMESPACE === props.match.params.ns && operatorHub?.status?.sources?.some(
      (source) => source.name === props.match.params.name
    )
  ), [operatorHub, props.match.params.name, props.match.params.ns]);
```

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -112,77 +111,145 @@ const DefaultSourceKebab: React.FC<DefaultSourceKebabProps> = ({
...
-      {
-        kind: referenceForModel(OperatorGroupModel),
-        isList: true,
-        namespace: props.match.params.ns,
-        prop: 'operatorGroups',
-      },
-    ]}
-  />
-);
+export const Operators: React.FC<PackageManifestPageProps> = (props) => {
```
TheRealJon, line 196: This should be a more descriptive name like `CatalogSourceOperatorsPage`. Exporting a module with a generic name like `Operators` could create confusion when importing modules in other areas of the app.

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -597,6 +638,10 @@ export type CreateSubscriptionYAMLProps = {
   location: Location;
 };
 
+export type PackageManifestPageProps = {
+  obj: CatalogSourceKind;
+} & ListPageProps;
+
```
TheRealJon, line 644: This type def should be named `<<ComponentName>>Props` (e.g. `OperatorsProps`). I suggested renaming this component, so adjust accordingly. `PackageManifestPageProps` collides with an existing type definition in `./packages/operator-lifecycle-manager/src/components/package-manifest.tsx`.

frontend/packages/operator-lifecycle-manager/src/components/modals/edit-registry-poll-interval-modal.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const patch = [
+      { op: 'add', path: '/spec/updateStrategy/registryPoll/interval', value: pollInterval },
+    ];
+    return handlePromise(k8sPatch(CatalogSourceModel, catalogSource, patch), close);
+  };
+
+  return (
+    <Form onSubmit={submit} name="form">
+      <div className="modal-content modal-content--no-inner-scroll">
+        <ModalTitle>Edit Registry Poll Interval</ModalTitle>
```
TheRealJon, line 49: nit:
```suggestion
        <ModalTitle>Edit registry poll interval</ModalTitle>
```

frontend/packages/operator-lifecycle-manager/src/components/modals/edit-registry-poll-interval-modal.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <ModalBody>
+          <FormGroup label="Registry Poll Interval" fieldId="">
```
TheRealJon, line 51:
```suggestion
          <FormGroup label="Registry Poll Interval" fieldId="">
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
-  classNames('col-lg-3', 'col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-lg-5', 'col-md-5', 'col-sm-4', 'col-xs-6'),
+  '',
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg'),
+  '',
 ];
 
-export const PackageManifestTableHeader = () => [
+export const PackageManifestTableHeaderWithCatalogSource = () => [
```
TheRealJon, line 27: Instead of repeating the same array elements, we should do something like:

```ts
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
-  classNames('col-lg-4', 'col-md-4', 'col-sm-4', 'col-xs-6'),
-  classNames('col-lg-3', 'col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-lg-5', 'col-md-5', 'col-sm-4', 'col-xs-6'),
+  '',
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg'),
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg'),
+  '',
 ];
 
-export const PackageManifestTableHeader = () => [
```
TheRealJon, line 38: A more DRY approach:

```ts
const PackageManifestTableHeader = () => [
  // ...Table header definitions
]

const PackageManifestTableHeaderWithCatalogSource = () => [
   ...PackageManifestTableHeader(),
  // ... CatalogSource table header definition
]
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
+export const PackageManifestTableHeaderWithCatalogSource = () => [
   {
     title: 'Name',
+    sortFunc: 'sortPackageManifest',
```
TheRealJon, line 30: Use a more descriptive name. By what distinguishing property are we sorting?

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
+    transforms: [sortable],
     props: { className: tableColumnClasses[0] },
   },
   {
     title: 'Latest Version',
```
TheRealJon, line 35:
```suggestion
    title: 'Latest version',
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
   },
   {
-    title: 'Subscriptions',
+    title: 'Created',
+    sortField: 'metadata.creationTimestamp',
+    transforms: [sortable],
     props: { className: tableColumnClasses[2] },
   },
+  {
+    title: 'Catalog Source',
```
TheRealJon, line 45:
```suggestion
    title: 'CatalogSource',
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
+    title: 'Latest Version',
+    props: { className: tableColumnClasses[1] },
+  },
+  {
+    title: 'Created',
+    sortField: 'metadata.creationTimestamp',
+    transforms: [sortable],
+    props: { className: tableColumnClasses[2] },
+  },
+];
```
TheRealJon, line 69: See my suggestion about a less repetitive solution.
```suggestion
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
       </TableData>
+      {!customData.catalogSource ? (
+        <TableData className={tableColumnClasses[3]}>
+          <ResourceLink
+            kind={referenceForModel(CatalogSourceModel)}
+            name={packageManifest.status?.catalogSource}
+            namespace={packageManifest.status?.catalogSourceNamespace}
+          />
+        </TableData>
+      ) : null}
```
TheRealJon, line 106:
```suggestion
      {!customData.catalogSource && (
        <TableData className={tableColumnClasses[3]}>
          <ResourceLink
            kind={referenceForModel(CatalogSourceModel)}
            name={packageManifest.status?.catalogSource}
            namespace={packageManifest.status?.catalogSourceNamespace}
          />
        </TableData>
      )}
```

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
-        namespace: status.catalogSourceNamespace,
-      }),
-    new Map<string, CatalogSourceInfo>(),
-  );
-
+export const PackageManifestList = (props: PackageManifestListProps) => {
+  const { customData } = props;
+  const TableHeader = customData.catalogSource
+    ? PackageManifestTableHeader
+    : PackageManifestTableHeaderWithCatalogSource;
```
TheRealJon, line 115: This is expression is a bit confusing given the names of the props and components. I'd add a comment here to clarify.

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
```diff
@@ -2,216 +2,144 @@ import * as React from 'react';
...
+      data={props.data || []}
+      filters={props.filters}
+      Header={TableHeader}
+      Row={PackageManifestTableRow}
       EmptyMsg={() => (
         <MsgBox
-          title="No Package Manifests Found"
-          detail="Package Manifests are packaged Operators which can be subscribed to for automatic upgrades."
+          title="No PackageManifests Found"
+          detail="The catalog author has not added any packages."
```
TheRealJon, line 128:
```suggestion
          detail="The CatalogSource author has not added any packages."
```

frontend/packages/operator-lifecycle-manager/src/plugin.tsx
```diff
@@ -205,6 +205,15 @@ const plugin: Plugin<ConsumedExtensions> = [
         ).ClusterServiceVersionsPage,
     },
   },
+  {
+    type: 'Page/Resource/List',
+    properties: {
+      model: models.PackageManifestModel,
+      loader: async () =>
+        (await import('./components/package-manifest' /* webpackChunkName: "packagemanifest" */))
```
TheRealJon, line 213:
```suggestion
        (await import('./components/package-manifest' /* webpackChunkName: "package-manifest" */))
```

frontend/public/components/factory/table.tsx
```diff
@@ -172,6 +173,14 @@ const sorts = {
   volumeSnapshotSource: (snapshot: VolumeSnapshotKind): string => snapshotSource(snapshot),
   snapshotLastRestore: (snapshot: K8sResourceKind, { restores }) =>
     restores[getName(snapshot)]?.status?.restoreTime,
+  sortPackageManifest: (packageManifest: PackageManifestKind): string => {
```
TheRealJon, line 176: See earlier comment about a more descriptive name for this. Maybe something like `sortPackageManifestByDefaultChannelName`. Also this logic is repeated in `PackageManifestTableRow` component, which means it is a good candidate for  a utility function. There is already a `defaultChannelFor` function in `./packages/operator-lifecycle-manager/src/components/index.tsx`, but it looks like that is essentially just a basic accessor for a channel name. You could rename that one to `defaultChannelNameFor` and make `defaultChannelFor` return the whole channel object.

frontend/public/components/utils/details-item.tsx
```diff
@@ -36,14 +36,27 @@ export const PropertyPath: React.FC<{ kind: string; path: string | string[] }> =
...
-  props,
-) => (
-  <Button variant="link" isInline onClick={props.onClick}>
-    {props.children}
-    <PencilAltIcon className="co-icon-space-l pf-c-button-icon--plain" />
-  </Button>
-);
+type EditButtonProps = {
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  testID?: string;
```
TheRealJon, line 41: Move to bottom of the file.

TheRealJon (commented):
A few small adjustments, but overall this looks really good!

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
harishsurf: makes sense

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
harishsurf: Added comment:  `If the CatalogSource is not present, display PackageManifests along with their CatalogSources (used in PackageManifest Search page)`

frontend/packages/operator-lifecycle-manager/src/components/package-manifest.tsx
harishsurf: much cleaner!

frontend/packages/operator-lifecycle-manager/src/utils/useOperatorHubConfig.tsx
```diff
@@ -0,0 +1,17 @@
...
+const useOperatorHubConfig = () => {
+  const [operatorHub, operatorHubLoaded, operatorHubLoadError] = useK8sWatchResource<
+    OperatorHubKind
+  >({
+    kind: referenceForModel(OperatorHubModel),
+    name: 'cluster',
+    isList: false,
+  });
+  return [operatorHub, operatorHubLoaded, operatorHubLoadError];
+};
```
TheRealJon, line 15: nit: this could be simplified:
```suggestion
const useOperatorHubConfig = () =>
  useK8sWatchResource<OperatorHubKind>({
    kind: referenceForModel(OperatorHubModel),
    name: 'cluster',
    isList: false,
  });
```

TheRealJon (approved):
Looks good!

/approve
/lgtm

spadgett:
@TheRealJon Can you review again? Looks like some changes were pushed after your last lgtm.

spadgett:
> Note: Sorting on "Latest Version" column is disabled since the versioning format would be different for different package manifests. I discussed this with @TheRealJon and agreed that the logic for sorting these versions is non-trivial.

@TheRealJon FYI, we have a semver library that we use to sort versions in other contexts which we might be able to use here (in a follow on PR).

TheRealJon:
> > Note: Sorting on "Latest Version" column is disabled since the versioning format would be different for different package manifests. I discussed this with @TheRealJon and agreed that the logic for sorting these versions is non-trivial.
> 
> @TheRealJon FYI, we have a semver library that we use to sort versions in other contexts which we might be able to use here (in a follow on PR).

@spadgett My main issue was whether it even makes sense to sort unrelated packages based on their latest versions. The latest version available in each package is only relevant within the scope of that package, so sorting against other packages is kind of meaningless.

PR #7433, opened 2020-12-05
convert last used namespace local storage to userSettings

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5111

This PR move the bridge/last-namespace-name local storage to the user settings.
No UI changes.
Raising this PR since CI is stuck on the old PR https://github.com/openshift/console/pull/7411.

### DIFF
--- a/frontend/packages/console-app/src/components/detect-namespace/DetectNamespace.tsx
+++ b/frontend/packages/console-app/src/components/detect-namespace/DetectNamespace.tsx
@@ -1,0 +1,17 @@
+import * as React from 'react';
+import { NamespaceContext, useValuesForNamespaceContext } from './namespace';
+
+type DetectNamespaceProps = {
+  children: React.ReactNode;
+};
+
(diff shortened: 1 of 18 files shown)


rohitkrai03 (approved):
/lgtm
/approve

frontend/public/locales/en/public.json
```diff
@@ -7,6 +7,7 @@
   "Confirm": "Confirm",
   "{{description}} for": "{{description}} for",
   "Edit": "Edit",
+  "{{dateFromNow}}": "{{dateFromNow}}",
```
spadgett, line 10: We shouldn't ever have a translation key that is only a parameter. It doesn't make sense. Can you revert this change?

I think we need to disable the i18n test on timestamp. We don't expect the moment.js relative times to be psuedolocalized since they're not using i18next.

dtaylor113: PR to disable i18n test for timestamps: https://github.com/openshift/console/pull/7445

PR #7438, opened 2020-12-07
Fix pipelinerun visualization inprogress crash

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5160

**Analysis / Root cause**: 
When starting a Pipeline multiple times the console crashes sometimes with this JavaScript error

**Solution Description**: 
If pipeline run is in progress and doesn't have enough information(pipelineSpec) to form the pipeline, then do not render the visualization graph.

pipelinerun-loaded

image

**Test setup:**

1. Install Openshift pipelines operator
2. Create a pipeline using a add flow or yaml
3. Start the pipeline run multiple times using `start last run` option.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug
/assign @jerolimov

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunVisualization.scss
+++ b/frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunVisualization.scss
@@ -1,0 +1,3 @@
+.pipeline-plr-loader {
+  margin-bottom: var(--pf-global--spacer--md);
+}
--- a/frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunVisualization.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunVisualization.tsx
@@ -1,16 +1,24 @@
 import * as React from 'react';
+import { LoadingInline } from '@console/internal/components/utils';
 import PipelineVisualization from '../../pipelines/detail-page-tabs/pipeline-details/PipelineVisualization';
 import { PipelineRun, getPipelineFromPipelineRun } from '../../../utils/pipeline-augment';
+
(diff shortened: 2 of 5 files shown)


jerolimov (approved):
Crash is solved. :+1:

The update is visible to the user, so that visualization comes after the data is available.

/lgtm
/assign @andrewballantyne

andrewballantyne (changes requested):
I think we actually want the solution to be in `frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunVisualization.tsx`... this is a problem with our pulling the Pipeline logic from the Pipeline Run... which is a matter of time.

Making the change in `frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx` sets us up for an awkward situation of allowing the creation of a PipelineVisualization without a Pipeline -- and that's not ideal.

The solution should also be a Loading indicator not a null as when I reproduce this scenario and get a `null` back from the Pipeline it removes the graph and then pops back in without warning. Probably a better use experience to spin.

andrewballantyne:
> Crash is solved. 👍

Definitely solved the crash... so this is an _okay_ solution to the problem, however...

> The update is visible to the user, so that visualization comes after the data is available.

This is probably not the best user experiernce.

/lgtm cancel

karthikjeeyar:
Moved the check to `pipelineRunVisualization.tsx` and added loader for better UX.

pipelinerun-loaded

andrewballantyne (approved):
cc @bgliwa01

bgliwa01 (approved):
lgtm!

andrewballantyne:
/lgtm

Slight improvement on the solution @jerolimov lgtm'ed -- we talked about this solution and he was in agreement. Let's get this merged 😄

PR #7439, opened 2020-12-07
Bug 1902981: Namespace is passed differently when using create vm from template.

metalice:
Added support for this type of url.

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1902981

**Analysis / Root cause**: 
Url format is different, should be used initData for namespace for this link.

**Solution Description**: 
Using init namespace

After:
image
Before:
image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vm-templates/menu-actions.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vm-templates/menu-actions.tsx
@@ -42,11 +42,11 @@ const newTemplateFromCommon: MenuAction = (kind, vmTemplate, { namespace }) => (
 const vmTemplateCreateVMAction: MenuAction = (
   kind,
   obj,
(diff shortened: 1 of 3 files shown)


metalice:
@rawagner , @yaacov please check :)

rawagner:
/hold

@metalice please pass current namespace to [createVMAction](https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/src/components/vm-templates/utils.ts#L61).  createVMAction creates link for VM Wizard. Your approach is fine for user templates but it wont work with common templates.

metalice:
> /hold
> 
> @metalice please pass current namespace to [createVMAction](https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/src/components/vm-templates/utils.ts#L61). createVMAction creates link for VM Wizard. Your approach is fine for user templates but it wont work with common templates.

even shorter :) done

PR #7440, opened 2020-12-07
Bug 1915220: Drop localStorage workaround after namespaces are now stored in user settings, use favorite from user settings

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5212
https://bugzilla.redhat.com/show_bug.cgi?id=1915220

**Analysis / Root cause**: 
We added this workaround to update the dropdown.js to support user settings. But for favorite project we needed this workaround until namespaces are also saved in user settings. This was merged with #7433, so we don't need this workaround anymore.

**Solution Description**: 
* Drop the workaround for localStorage.
* Enabled user settings sync function for bookmarked projects and applications and the favorite project between browser tabs.
* Integrate favorite project into 'actual' project / namespace hook so that the favorite project was automatically open.

Peek 2020-12-10 20-25

Untouched :(

**Test setup:**
1. Favorite a namespace and open the console again with `/`, the selected project should be opened.
2. Open dev console in two browsers with the same users and add / remove namespace bookmarks

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
+++ b/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
@@ -48,9 +48,13 @@ export const useValuesForNamespaceContext = () => {
   );
 
   // Keep namespace in sync with redux.
-  // Automatically sets favorited or latest namespace as soon as
-  // both informations are loaded from user settings.
   React.useEffect(() => {
+    // Url namespace overrides everything. We don't need to wait for loaded
+    if (urlNamespace && !favoriteLoaded && !lastNamespaceLoaded) {
+      dispatch(setActiveNamespace(urlNamespace));
+    }
+    // Automatically sets favorited or latest namespace as soon as
+    // both informations are loaded from user settings.
     if (!urlNamespace && favoriteLoaded && lastNamespaceLoaded) {
       dispatch(setActiveNamespace(favoritedNamespace || lastNamespace || ALL_NAMESPACES_KEY));
     }
--- a/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
+++ b/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
 import { getNamespace } from '@console/internal/components/utils/link';
 import { useUserSettingsCompatibility } from '@console/shared/src/hooks/useUserSettingsCompatibility';
 import { setActiveNamespace } from '@console/internal/actions/ui';
-import { ALL_NAMESPACES_KEY } from '@console/shared/src/constants';
+import {
+  ALL_NAMESPACES_KEY,
+  USERSETTINGS_PREFIX,
+  NAMESPACE_USERSETTINGS_PREFIX,
+  NAMESPACE_LOCAL_STORAGE_KEY,
+} from '@console/shared/src/constants';
 
 type NamespaceContextType = {
   namespace?: string;
   setNamespace?: (ns: string) => void;
 };
 
+const FAVORITE_NAMESPACE_NAME_USERSETTINGS_KEY = `${NAMESPACE_USERSETTINGS_PREFIX}.favorite`;
+const FAVORITE_NAMESPACE_NAME_LOCAL_STORAGE_KEY = NAMESPACE_LOCAL_STORAGE_KEY;
+
+const LAST_NAMESPACE_NAME_USER_SETTINGS_KEY = `${USERSETTINGS_PREFIX}.lastNamespace`;
 const LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY = `bridge/last-namespace-name`;
-const LAST_NAMESPACE_NAME_USER_SETTINGS_KEY = 'console.lastNamespace';
 
 export const NamespaceContext = React.createContext<NamespaceContextType>({});
 
 export const useValuesForNamespaceContext = () => {
   const { pathname } = useLocation();
   const urlNamespace = getNamespace(pathname);
-  /**
-   * [TODO]: use favorite namespace here if there is any
-   *  const [favorite] = useUserSettings()
-   */
-  const [namespace, setNs, loaded] = useUserSettingsCompatibility(
-    LAST_NAMESPACE_NAME_USER_SETTINGS_KEY,
-    LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
-    urlNamespace || ALL_NAMESPACES_KEY,
+
+  const [favoritedNamespace, , favoriteLoaded] = useUserSettingsCompatibility<string>(
+    FAVORITE_NAMESPACE_NAME_USERSETTINGS_KEY,
+    FAVORITE_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
   );
-  const dispatchreduxAction = useDispatch();
+  const [lastNamespace, setLastnameNamespace, lastNamespaceLoaded] = useUserSettingsCompatibility<
+    string
+  >(LAST_NAMESPACE_NAME_USER_SETTINGS_KEY, LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
+
+  const dispatch = useDispatch();
   const setNamespace = React.useCallback(
-    (ns: string) => {
-      dispatchreduxAction(setActiveNamespace(ns));
-      setNs(ns);
+    (namespace: string) => {
+      dispatch(setActiveNamespace(namespace));
+      setLastnameNamespace(namespace);
     },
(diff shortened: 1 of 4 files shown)


jerolimov:
/unhold
/cc @sahil143 @invincibleJai
/assign @christianvogt 

Can you please double check the code and I expect that you know some areas in console which are also affected by the "active / last / favorite namespace". I tested this primary on the topology page with and without a favorite namespace. With and without a namespace in URL. Switching from 4.6 to this PR and dropping some ConfigMap or localStorage entries.

Thanks.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
...
     }
-    // only run this hook when namespace is loaded from user settings
+    // only run this hook when favorite and last namespace is loaded from user settings
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [loaded]);
+  }, [urlNamespace, favoriteLoaded, lastNamespaceLoaded]);
 
-  return { namespace: urlNamespace || namespace || ALL_NAMESPACES_KEY, setNamespace, loaded };
+  return {
+    namespace: urlNamespace || lastNamespace || favoritedNamespace || ALL_NAMESPACES_KEY,
```
sahil143, line 64:
```suggestion
    namespace: urlNamespace || favoritedNamespace || lastNamespace ||  ALL_NAMESPACES_KEY,
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
...
+  // Keep namespace in sync with redux
   React.useEffect(() => {
-    if (loaded) {
-      dispatchreduxAction(setActiveNamespace(urlNamespace || namespace || ALL_NAMESPACES_KEY));
+    // Url namespace overrides everything. We don't need to wait for loaded
+    if (urlNamespace) {
+      dispatch(setActiveNamespace(urlNamespace));
+    } else if (favoriteLoaded && lastNamespaceLoaded) {
+      // Calculate namespace after loading: favorite > last used namespace > all
+      dispatch(setActiveNamespace(favoritedNamespace || lastNamespace || ALL_NAMESPACES_KEY));
```
sahil143, line 57:
```suggestion
 if (favoriteLoaded && lastNamespaceLoaded) {
      // Calculate namespace after loading: favorite > last used namespace > all
      dispatch(setActiveNamespace(urlNamespace  || favoritedNamespace || lastNamespace || ALL_NAMESPACES_KEY));
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
...
     }
-    // only run this hook when namespace is loaded from user settings
+    // only run this hook when favorite and last namespace is loaded from user settings
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [loaded]);
+  }, [urlNamespace, favoriteLoaded, lastNamespaceLoaded]);
```
sahil143, line 61:
```suggestion
  }, [ favoriteLoaded, lastNamespaceLoaded]);
```

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
...
+    [dispatch, setLastnameNamespace],
   );
 
+  // Keep namespace in sync with redux
   React.useEffect(() => {
-    if (loaded) {
-      dispatchreduxAction(setActiveNamespace(urlNamespace || namespace || ALL_NAMESPACES_KEY));
+    // Url namespace overrides everything. We don't need to wait for loaded
+    if (urlNamespace) {
+      dispatch(setActiveNamespace(urlNamespace));
```
sahil143, line 54: we don't need this because it's already being taken care of in the main app reducer `SetCurrentLocation` action.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
```diff
@@ -6,46 +6,63 @@ import { useLocation } from 'react-router-dom';
...
-   *  const [favorite] = useUserSettings()
-   */
-  const [namespace, setNs, loaded] = useUserSettingsCompatibility(
-    LAST_NAMESPACE_NAME_USER_SETTINGS_KEY,
-    LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
-    urlNamespace || ALL_NAMESPACES_KEY,
+
+  const [favoritedNamespace, , favoriteLoaded] = useUserSettingsCompatibility<string>(
+    FAVORITE_NAMESPACE_NAME_USERSETTINGS_KEY,
+    FAVORITE_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
```
sahil143, line 35: we might need sync here.

sahil143 (changes requested):

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
christianvogt: Not sure we need sync here since detection is a one time operation.

christianvogt:
@jerolimov somethings wrong with the user settings.
If i have 2 browsers open and make changes to fav's in one browser, then immediately go to the other browser, while I do see the new changes, then I make changes to the list, it will override the changes made in the first browser. It seems the local state is taking precedence over merging with the sync state.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
jerolimov: @sahil143 Exact. I checked this now again and we need the favorite here only once for the initial transition from "we don't know the favorited namespace" to "favorited namespace is loaded". In this case we should set it.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
jerolimov: @sahil143 I debugged `SetCurrentLocation` a little bit, but it was only called when the URL changes, not on initial load.

So at the moment I expect we should "activate" the "url" namespace as soon as possible. Because it overrides the favorited and last namespace.

After finish loading we could then set the favorited namespace, or last used namespace, or all as fallback. In this order.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
jerolimov: @sahil143 Thanks, your right here. The urlNamespace is already known and I can drop the first if statement with `dispatch(setActiveNamespace(urlNamespace));`.

I keeped the second one and added the check !urlNamespace to ensure that it doesn't set the urlNamespace again after loading is finished.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
jerolimov: Accepted. Removed the urlNamespace key here.

frontend/packages/console-app/src/components/detect-namespace/namespace.ts
jerolimov: Accepted, change to your order here.

jerolimov:
@christianvogt @sahil143 @invincibleJai  Thanks for the feedback and reviews. I added some separate commits to make a new review hopefully easier.

* The first two commits are part of the original PR to use the favorited namespace if the user opens path `/`.
* Commit 3 https://github.com/openshift/console/pull/7440/commits/81f84319d4d3d50e9ebfde6cd9b9e63bed1de524 solved findings from Sahil
* Commit 4 https://github.com/openshift/console/pull/7440/commits/3bbcf5a81882ae7155f3e02536823784c145f9b5 just extract some if conditions into "outer" if statements!
* Commit 5 and 6 adds unit tests and a fix for Christians issue (https://github.com/openshift/console/pull/7440#issuecomment-758133653), the problem was that the `oldValue` in a setter was not the latest version received from the k8s watcher in sync mode. I added just the `settingsRef.current = resultedSettings` after the useMemo hook.

Wdyt? Any feedback is welcome. :)

jerolimov:
@christianvogt Added https://github.com/openshift/console/pull/7440/commits/86c471102412a737f8114f76e1e8c3f5ea31577a to restore the application dropdown when open the add page in a namespace and reload the page (or just load `/deploy-image/ns/xyz`).

PR #7443, opened 2020-12-07
Fix to not show misleading alert message in Project Access Page

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-3289

**Analysis/Root cause:**
An alert with message `Click Save to save changes or Reload to cancel changes` was shown to the user after adding a empty access row even though when `Save` button is disabled.

**Solution:**
Show alert only when the fields are changed and `Save` is enabled. This PR also adds `resetLabel` and `submitLabel` to show in the alert message which were missed during i18n integration.

**Screenshots:**
image
image

/kind bug


frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -18,29 +19,35 @@ const ProjectAccessForm: React.FC<FormikProps<FormikValues>> = ({
...
+}) => {
+  const { t } = useTranslation();
+  const disableSubmit = !dirty || !_.isEmpty(errors);
+  return (
+    <Form onSubmit={handleSubmit}>
+      <div className="co-m-pane__form">
+        <MultiColumnField
+          name="projectAccess"
+          addLabel="Add Access"
+          headers={['Name', 'Role']}
```
vikram-raj, line 31: Add translation for `headers` and `addLabel`

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -18,29 +19,35 @@ const ProjectAccessForm: React.FC<FormikProps<FormikValues>> = ({
...
+          emptyValues={{ user: '', role: '' }}
+        >
+          <InputField name="user" type={TextInputTypes.text} placeholder="Name" />
```
vikram-raj, line 34: Add translation for `placeholder`.

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -3,6 +3,7 @@ import * as _ from 'lodash';
 import { Form, TextInputTypes } from '@patternfly/react-core';
 import { FormikProps, FormikValues } from 'formik';
 import { MultiColumnField, InputField, DropdownField, FormFooter } from '@console/shared';
+import { useTranslation } from 'react-i18next';
 
 enum accessRoles {
   '' = 'Select a role',
```
vikram-raj, line 9: Can you please also fix this. I don't think we need this empty string value. we can pass the `title` prop to `DropdownField` instead of this.

divyanshiGupta: Hmm this is more of refactoring unrelated code for this fix but since it is a small change I have removed this and added a `title`.

vikram-raj: Thanks @divyanshiGupta

vikram-raj (commented):
Verified it. Works as expected. 👍 

/lgtm

rohitkrai03 (approved):
/approve

PR #7444, opened 2020-12-07
Hide scale buttons of pod donut for Pod resources

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5145

**Analysis / Root cause**: 
Scale up / down buttons are available for a single Pod resource and results in an error when pressed.

**Solution Description**: 
The buttons are hidden if the resource is of type Pod.

tmp

**Test setup:**
- Create a single pod
- Go to Details tab in side panel of the pod node

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


christianvogt:
Can you make the original hook more efficient if we know that scaling is not allowed, there's no need to check the access in the effect.

rottencandy:
> Can you make the original hook more efficient if we know that scaling is not allowed, there's no need to check the access in the effect.

Updated.

PR #7445, opened 2020-12-07
Cypress: do not test for i18n pseudo-translation of timestamps

dtaylor113:
We use `momentjs` for our timestamps which may show absolute or relative times (relative = 'a few moments ago').  Momentjs does provide [relative time translations](https://momentjs.com/docs/#/customization/relative-time/), but it doesn't work with `pseduo-translation` which we use in our cypress testing. 

This PR removes Cypress testing of pseudo-translation for timestamps.


spadgett (approved):
/lgtm

PR #7449, opened 2020-12-07
Bug 1904993: Remove &amp; in strings

rebeccaalpert:
I didn't see any others that had been internationalized.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1904993.


spadgett (approved):
These are the only ones I could find as well. Thanks!

/lgtm

PR #7450, opened 2020-12-07
Bug 1898679: Fix .co-required class styling

TheRealJon:
In some cases, the `position: absolute; top: 0;` styling on the `.co-required:after` pseudo-element was causing it to be rendered outside of the parent component, making it invisible.

### DIFF
--- a/frontend/public/style/_forms.scss
+++ b/frontend/public/style/_forms.scss
@@ -12,16 +12,16 @@
   flex-flow: row-reverse;
 }
 
-.co-required {
-  position: relative;
-  &:after {
-    color: $color-pf-red-200;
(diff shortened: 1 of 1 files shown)


sg00dwin:
@TheRealJon looking back I remembered why I used `position: absolute; top: 0`. We had a case where the asterisk needed explicit positioning. So I just tested removing it and swapping `vertical-align: top;` in place of `position: absolute; top: 0` and it works.  
Can you include `vertical-align: top;` on that `.co-required` rule

sg00dwin (commented):
add `vertical-align: top;`

sg00dwin (approved):
/lgtm

PR #7452, opened 2020-12-08
route eventsource from Admin to catalog for source and workload on create

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5220

**Analysis / Root cause**: 
route on eventsource create from serverless eventing tab in Admin perspective goes to old route

**Solution Description**: 
route eventsource create from serverless eventing tab in Admin perspective to catalog for source and workload on create

Dec-08-2020 13-44-33

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj (commented):
verified it.

/lgtm

PR #7456, opened 2020-12-08
WIP: fix gap above activity card

debsmita1:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5221

**Analysis / Root cause**: 
Before rendering the Quick Starts catalog card it is not being checked if there are Quick Starts available in the cluster

**Solution Description**: 
- checking if Quickstarts are available
- passing sync as true to `useUserSettings` hook such that the card on the right adjusts itself when the Quickstarts catalog card is removed

When no Quickstarts are available
no-quickstarts

qs

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
+++ b/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
@@ -17,7 +17,6 @@ import { ClusterDashboardContext } from './context';
 
 const mainCards = [{ Card: StatusCard }, { Card: UtilizationCard }];
 const leftCards = [{ Card: DetailsCard }, { Card: InventoryCard }];
-const rightCards = [{ Card: ActivityCard }];
 
 const HIDE_QUICK_START_DASHBOARD_TILE_USER_SETTINGS_KEY = 'console.dashboard.quickStartTile';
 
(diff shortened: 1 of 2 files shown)


frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
```diff
@@ -26,15 +29,24 @@ export const ClusterDashboard: React.FC<{}> = () => {
     InfrastructureModel,
     'cluster',
   );
-  // [TODO](sahil143): use sync capability here
+
+  const [consoleQuickStarts, consoleQuickStartsLoaded] = useK8sWatchResource<QuickStart[]>({
+    kind: referenceForModel(QuickStartModel),
+    isList: true,
+  });
```
sahil143, line 36: Since we are already loading quick starts here, I'm wondering if we can remove QuickStartsLoader from this file which does the same thing with additional permission checks. Can we create a hook that can fetch the quick starts and do permission checks as well?
Only checking for availability of quick starts seems flawed because there can be quickstarts that don't pass the permission check and if those are the only quickstarts available then the card might have zero quick starts and the gap will be still available.
cc @rohitkrai03

sahil143 (changes requested):

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
rohitkrai03: Quick starts shouldn't be loaded like this directly, it skips the permissions check which is eseential to determine what quikc starts are available for the current user based on his permissions. Also, the permission checker cannot be a simple hook since the quick starts are coming as `ConsoleQuickStart` CRs and we have to run permission check for each of them use `useAccessReview` hook. We cannot run the access review check safely over an array of items that are coming from backend and can change anytime.

rohitkrai03 (changes requested):

spadgett:

PR #7458, opened 2020-12-08
Bug 1881268: Image uploading failed but wizard claim the source

glekner:


glekner:
@gouyang we agreed to delete the dv when this error occurs.

PR #7461, opened 2020-12-08
Bug 1906872: Fix Pipeline Tech Preview badge alignment

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4915

**Analysis / Root cause**: 
The tech preview badge is not aligned with the Pipeline title

**Solution Description**: 
Basealign Pipeline form heading with Tech Preview badge

image

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.scss
+++ b/frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.scss
@@ -1,0 +1,3 @@
+.odc-form-section-pipeline {
+  align-items: baseline;
+}
--- a/frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.tsx
@@ -13,6 +13,8 @@ import { FLAG_OPENSHIFT_PIPELINE, CLUSTER_PIPELINE_NS } from '../../../const';
 import { Pipeline } from '../../../utils/pipeline-augment';
 import PipelineTemplate from './PipelineTemplate';
 
+import './PipelineSection.scss';
+
 type PipelineSectionProps = {
   flags: FlagsObject;
   builderImages: NormalizedBuilderImages;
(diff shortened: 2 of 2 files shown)


frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.tsx
```diff
@@ -12,6 +12,7 @@ import { PipelineModel, PipelineResourceModel } from '../../../models';
 import { FLAG_OPENSHIFT_PIPELINE, CLUSTER_PIPELINE_NS } from '../../../const';
 import { Pipeline } from '../../../utils/pipeline-augment';
 import PipelineTemplate from './PipelineTemplate';
+import './PipelineSection.scss';
```
karthikjeeyar, line 15: nit: new line

frontend/packages/pipelines-plugin/src/components/import/pipeline/PipelineSection.scss
```diff
@@ -0,0 +1,3 @@
+.odc-form-section-pipeline {
```
karthikjeeyar, line 1: now that we have pipelines as a separate package, I think it will be better if we should use `pipeline-` prefix instead of `odc`, thoughts?

christianvogt: Agreed, but let's create a separate task for this as there's many instances in the package that need updating.
Maybe something short like `op` (OpenShift Pipelines) or `tp` (Tekton Pipelines).

vikram-raj: Created an issue for this https://issues.redhat.com/browse/ODC-5253

vikram-raj:
/retitle Bug 1906872: Fix Pipeline Tech Preview badge alignment

jerolimov (approved):

PR #7462, opened 2020-12-08
Bug 1900377: Add selector for active users on Overview tab

pcbailey:
This PR adds a selector to target the active users message in the Details card of the Overview tab when the guest agent is installed and at least one user is active.


pcbailey:
@omril-redhat @yaacov Please review.

PR #7463, opened 2020-12-08
Bug 1905542: Replace 'Baremetal' with 'BareMetal' in the supported platforms for I…

ashishkumsingh:
Replace 'Baremetal' with 'BareMetal' in the supported platforms for Independent mode of OCS deployment.

Signed-off-by: Ashish Singh <assingh@redhat.com>


spadgett (approved):
/lgtm

PR #7465, opened 2020-12-08
Bug 1906760: Fixes to prevent topology from unnecessary updates/renders

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5169

**Analysis / Root cause**: 
Updates were occurring too frequently and when data had not changed.

**Solution Description**: 
Check for data changes before updating topology data and pod data. Throttle incoming updates to topology resource data.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/topology/src/data-transforms/TopologyDataRetriever.tsx
+++ b/frontend/packages/topology/src/data-transforms/TopologyDataRetriever.tsx
@@ -2,12 +2,13 @@ import * as React from 'react';
 import {
   useK8sWatchResources,
   WatchK8sResources,
+  WatchK8sResults,
 } from '@console/internal/components/utils/k8s-watch-hook';
-import { usePrometheusRulesPoll } from '@console/internal/components/graphs/prometheus-rules-hook';
-import { getAlertsAndRules } from '@console/internal/components/monitoring/utils';
+import { useDebounceCallback } from '@console/shared';
 import { TopologyResourcesObject, TrafficData } from '../topology-types';
 import { ModelContext, ExtensibleModel } from './ModelContext';
 import { getFilterById, SHOW_GROUPS_FILTER_ID, useDisplayFilters } from '../filters';
+import { useMonitoringAlerts } from './useMonitoringAlerts';
 import { updateTopologyDataModel } from './updateTopologyDataModel';
 
 type TopologyDataRetrieverProps = {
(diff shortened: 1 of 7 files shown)


frontend/packages/topology/src/components/graph-view/components/nodes/WorkloadNode.tsx
```diff
@@ -53,6 +53,14 @@ const WorkloadPodsNode: React.FC<WorkloadPodsNodeProps> = observer(
     const showPodCount = showPodCountFilter?.value ?? false;
     const { decorators } = element.getGraph().getData();
     const nodeDecorators = getNodeDecorators(element, decorators, cx, cy, radius, decoratorRadius);
+
+    const podSet = React.useMemo(() => {
+      if (donutStatus) {
+        return <PodSet size={size} x={cx} y={cy} data={donutStatus} showPodCount={showPodCount} />;
+      }
+      return null;
+    }, [cx, cy, donutStatus, showPodCount, size]);
```
christianvogt, line 62: I don't see the value in doing this.
I think what you want to do is `React.memo` on `PodSet`.
Just be careful to ensure that we don't mutate data and expect it to re-render.

frontend/packages/topology/src/data-transforms/TopologyDataRetriever.tsx
```diff
@@ -18,34 +26,21 @@ const TopologyDataRetriever: React.FC<TopologyDataRetrieverProps> = ({ trafficDa
...
-        if (alert.rule) {
-          delete alert.rule.evaluationTime;
-          delete alert.rule.lastEvaluation;
-          alert.rule.alerts &&
-            alert.rule.alerts.forEach((ruleAlert) => {
-              delete ruleAlert.activeAt;
-            });
-        }
-      });
+  const updateResources = _.debounce((newResources) => {
```
christianvogt, line 36: Need to use the `debounce` hook otherwise this creates a new debounce function on ever update.

frontend/packages/topology/src/components/page/TopologyView.tsx
```diff
@@ -185,6 +186,17 @@ export const ConnectedTopologyView: React.FC<ComponentProps> = ({
...
+        const oldModel = visualization.toModel();
+        if (newModel.nodes && oldModel.nodes) {
+          newModel.nodes.forEach((node) => {
+            const oldNode = oldModel.nodes.find((n) => n.id === node.id);
+            if (oldNode && _.isEqual(oldNode.data, node.data)) {
+              node.data = oldNode.data;
+            }
+          });
+        }
+      }
```
christianvogt, line 199: Why is this necessary?

PR #7467, opened 2020-12-08
Bug 1905253:  update events timeline empty state and end messages

rhamilto:
Before:
Screen Shot 2020-12-08 at 12 06 11 PM

After:
Screen Shot 2020-12-11 at 4 08 25 PM

### DIFF
--- a/frontend/public/components/events.jsx
+++ b/frontend/public/components/events.jsx
@@ -265,9 +265,7 @@ export const NoEvents = () => {
   const { t } = useTranslation();
   return (
     <Box className="co-sysevent-stream__status-box-empty">
-      <div className="text-center cos-status-box__detail">
-        {t('events~No events in the past hour')}
-      </div>
+      <div className="text-center cos-status-box__detail">{t('events~No events')}</div>
     </Box>
   );
 };
@@ -328,7 +326,6 @@ class EventStream extends React.Component {
       filteredEvents: [],
       error: null,
       loading: true,
-      oldestTimestamp: new Date(),
     };
     this.toggleStream = this.toggleStream_.bind(this);
   }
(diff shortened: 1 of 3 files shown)


andybraren (approved):
Thanks for the quick fix, lgtm!

frontend/public/components/events.jsx
```diff
@@ -516,7 +516,7 @@ class EventStream extends React.Component {
               className="co-sysevent-stream__timeline__btn"
             />
             <div className="co-sysevent-stream__timeline__end-message">
-              There are no events before <Timestamp timestamp={this.state.oldestTimestamp} />
+              Events older than 2 hours are not stored.
```
spadgett, line 519: The default TTL is one hour, but I believe it can be changed. We might not be able to hard-code the message.

rhamilto: I just noticed [the empty state message is `No events in the past hour`](https://github.com/openshift/console/blob/385f62dac05fad1757825dbc197ff86b96ea5c8c/frontend/public/components/events.jsx#L251).  If the TTL can be changed, the empty state message also needs updating.  Perhaps we just do:

`Older events are not stored.` and `No events` so as to keep the messages precise?  @andybraren, your thoughts?

andybraren: I like those tweaks @rhamilto, both of which are better than what they are today. I posted in [#forum-monitoring](https://coreos.slack.com/archives/C0VMT03S5/p1607723565406400) to see if the 1 hour is configurable.

andybraren: Oh, I just found this which suggests the default is 3 hours and isn't configurable? 🤷‍♂️
https://access.redhat.com/solutions/5170631

sttts: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v4.1.0/config/defaultconfig.yaml#L96

spadgett (approved):
/lgtm

spadgett:
@andybraren @itsptk Does the wording look good?

andybraren (approved):
`Older events are not stored` and `No events` seem like the safest ways to phrase this, so lgtm!

PR #7469, opened 2020-12-08
Bug 1906768: Fix to correctly sort nav items dependent on non-plugin items

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5226

**Analysis / Root cause**: 
Nav items within sections were sorted amongst themselves then added in that order

**Solution Description**: 
Sort the nav items by dependencies, then insert each in order so that they can then be placed before/after existing nav items.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


yaacov:
Works for me on Firefox v83.0 - fedora

PR #7471, opened 2020-12-08
CONSOLE-2381: Support dynamic demo plugin deployment on cluster

vojtechszocs:
- [x] Add Dockerfile to build the demo plugin image.
- [x] Add Kubernetes manifest to deploy demo plugin on the cluster.
- [x] Add basic README for the demo plugin.

The image is based on official [Node.js image](https://hub.docker.com/_/node), using `http-server.sh` script as the entry point and running as `node` user.

@spadgett Built image size is `985MB`, is this OK or should we consider using `node:alpine` as the base image to reduce its size?

```
$ docker image ls -f "reference=node:10" -f "reference=test"
REPOSITORY   TAG       IMAGE ID       CREATED        SIZE
test         latest    855fa54f99c3   24 hours ago   985MB
node         10        78e6fbf9befd   6 days ago     911MB
```

When building the Docker image, we need the whole Console repo as the build context. This is because `frontend/codegen.yml` is referencing GraphQL schemas at `pkg/graphql`.

The plugin HTTP server runs at port 9001 with caching disabled and CORS enabled.

### DIFF
--- a/frontend/dynamic-demo-plugin/README.md
+++ b/frontend/dynamic-demo-plugin/README.md
@@ -1,0 +1,62 @@
+# OpenShift Console Demo Plugin
+
+This project emulates a standalone repository hosting a sample
+[dynamic plugin](/frontend/packages/console-dynamic-plugin-sdk/README.md) for OpenShift Console.
+
+It is meant to serve as a reference for Console plugin developers and for testing dynamic plugin
+capabilities via end-to-end tests.
+
+## Local development
+
+1. `yarn build` to build the plugin, generating output to `dist` directory
+2. `yarn http-server` to start an HTTP server hosting the generated assets
+
+```
+Starting up http-server, serving ./dist
+Available on:
+  http://127.0.0.1:9001
+  http://192.168.1.190:9001
+  http://10.40.192.80:9001
+Hit CTRL-C to stop the server
+```
+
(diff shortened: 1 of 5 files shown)


frontend/dynamic-demo-plugin/README.md
```diff
@@ -0,0 +1,62 @@
...
+the script, for example:
+
+```sh
+yarn http-server -a 127.0.0.1 --ssl --cert /path/to/tls.crt --key /path/to/tls.key
+```
+
+## Deployment on cluster
+
+Console dynamic plugins are supposed to be deployed via [OLM operators](https://github.com/operator-framework).
+In case of demo plugin, we just apply a minimal Kubernetes manifest which adds the necessary resources.
```
jhadvig, line 34: Since we the manifest is containing also a console CRD I would rather use OpenShift.
```suggestion
In case of demo plugin, we just apply a minimal OpenShift manifest which adds the necessary resources.
```

jhadvig (changes requested):

vojtechszocs:
@jhadvig Thanks for the review! PR updated, please have a second look.

jhadvig (approved):
@vojtechszocs lets suqsh this into a single commit and we can tag it 👍 
/approve

rhamilto (approved):
/lgtm

PR #7475, opened 2020-12-08
Bug 1906899: fix offset in quick start highlight bounding box

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5234

**Analysis / Root cause**: 
The bounding box of the target node to highlight gets calculated immediately after the highlight is rendered to the page. However, this occurs before popper has positioned the highlight. Since the Popper component rendered the content without any positional information, it ended up rendering the content at the bottom of the page below the fold. This results in a vertical scrollbar appearing on the body and shifting the content over to the left. This only occurs the browser / os has scrollbars always visible.

**Solution Description**: 

The solution here is to default the popper wrapper with `position: absolute, top: 0, left: 0` so that it begins rendering in the top left corner. Also changed the scroll behavior to be smooth as it helps guide the user to understand where the element came from it if happens to be offscreen.

To test all spotlight usage, I had to fix an issue where the guided tour menu item wasn't rendering because of a change to active perspective fetching. This fix required also addressing a circular dependency.

image

**Test Setup**
Install the following Quick Start:

```
apiVersion: console.openshift.io/v1
kind: ConsoleQuickStart
metadata:
  name: test-highlight
spec:
  description: test desc
  displayName: test highlight
  durationMinutes: 1
  introduction: test intro
  tasks:
    - description: |
        1. [perspective]{{highlight tour-perspective-dropdown}}
        2. [masthead]{{highlight tour-help-button}}
        3. [search]{{highlight tour-search-nav}}
      title: test highlight
```

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/detect-perspective/useValuesForPerspectiveContext.ts
+++ b/frontend/packages/console-app/src/components/detect-perspective/useValuesForPerspectiveContext.ts
@@ -1,0 +1,24 @@
+import * as React from 'react';
+import { isPerspective, Perspective, useExtensions } from '@console/plugin-sdk';
+import { useUserSettingsCompatibility } from '@console/shared';
+import { PerspectiveType } from './perspective-context';
+
+const LAST_PERSPECTIVE_LOCAL_STORAGE_KEY = `bridge/last-perspective`;
+
+const LAST_PERSPECTIVE_USER_SETTINGS_KEY = 'console.lastPerspective';
+
+export const useValuesForPerspectiveContext = (): [
+  PerspectiveType,
+  React.Dispatch<React.SetStateAction<PerspectiveType>>,
+  boolean,
+] => {
+  const perspectiveExtensions = useExtensions<Perspective>(isPerspective);
+  const [perspective, setPerspective, loaded] = useUserSettingsCompatibility<PerspectiveType>(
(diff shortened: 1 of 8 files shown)


frontend/packages/console-app/src/components/detect-perspective/useValuesForPerspectiveContext.ts
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import { isPerspective, Perspective, useExtensions } from '@console/plugin-sdk';
+import { useUserSettingsCompatibility } from '@console/shared';
+import { PerspectiveType } from './perspective-context';
+
+const LAST_PERSPECTIVE_LOCAL_STORAGE_KEY = `bridge/last-perspective`;
+
+const LAST_PERSPECTIVE_USER_SETTINGS_KEY = 'console.lastPerspective';
+
+export const useValuesForPerspectiveContext = (): [
```
andrewballantyne, line 10:
```suggestion
default export (): [
```
Named files should default export. Generic files `utils...` can export multiple/named.

andrewballantyne (approved):
/lgtm

One nit, doesn't make a difference if it's left.

PR #7480, opened 2020-12-09
Bug 1906871: Fix white screen issue of devconsole monitoring alerts

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5219

**Analysis / Root cause**: 
In the dev console monitoring alerts page, when clicking on the Name column white screen the UI.

**Solution Description**: 
Fix the infinite loop of rendering.

Kapture 2020-12-09 at 17 10 20

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj:
/retitle Bug 1906871: Fix white screen issue of devconsole monitoring alerts

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlertColumn.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type MonitoringAlertColumn = {
+  title: string;
+  cellFormatters?: IFormatter[];
+  transforms?: ITransform[];
+  fieldName?: string;
+  sortFunc?: string;
+};
+
+export const MonitoringAlertColumn = (t: TFunction): MonitoringAlertColumn[] => [
```
jerolimov, line 12: We use this in different components, I'm unsure if this is the easiest solution, but I can live with this until we get some performance issues or want to change this on all components at the same time.

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
```diff
@@ -138,10 +98,10 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
     dispatch(
       sortList(
         reduxID,
-        monitoringAlertColumn[index - 1].fieldName,
-        monitoringAlertColumn[index - 1].sortFunc,
+        MonitoringAlertColumn(t)[index - 1].fieldName,
+        MonitoringAlertColumn(t)[index - 1].sortFunc,
         direction,
-        monitoringAlertColumn[index - 1].title,
+        MonitoringAlertColumn(t)[index - 1].title,
```
jerolimov, line 104: @vikram-raj But can you please call this method once or memorize this output of `MonitoringAlertColumn(t)`. Also if this method just return a static array we should not create this multiple times per rendering.

```tsx
const monitoringAlertColumn = MonitoringAlertColumn(t);
```

or

```tsx
const monitoringAlertColumn = React.useMemo(() => MonitoringAlertColumn(t), [t]);
```

Memo is option from my perspective here.

jerolimov (changes requested):
Fixes the rerendering issue but have a comment about reusing the function result.

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
vikram-raj: Updated it. Thanks

jerolimov (approved):
Works fine :+1:

/lgtm

invincibleJai:
/approve

PR #7481, opened 2020-12-09
Bug 1906684: handles application group issue with EventSource creation

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5239

**Analysis / Root cause**: 
Event Source creation fails if user selects no app group and switch to yaml and then to form as application name had text "no application group" as part of annotations

**Solution Description**: 
handles application group issue with EventSource creation

image

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
+++ b/frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
@@ -10,6 +10,8 @@ import {
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
 import { checkAccess, history } from '@console/internal/components/utils';
 import { safeYAMLToJS } from '@console/shared/src/utils/yaml';
+import { UNASSIGNED_APPLICATIONS_KEY } from '@console/shared/src/constants';
+import { CREATE_APPLICATION_KEY } from '@console/topology/src/const';
 import { parseALMExamples, ClusterServiceVersionKind } from '@console/operator-lifecycle-manager';
 import {
   getAppLabels,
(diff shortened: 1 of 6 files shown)


frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
```diff
@@ -466,3 +467,56 @@ export const handleRedirect = (
...
+    application: {
+      ...formDataValues.application,
+      ...(appGroupName &&
+        appGroupName !== formDataValues.application.name && {
+          name: appGroupName,
+          selectedKey: formDataValues.application.selectedKey ? CREATE_APPLICATION_KEY : '',
+        }),
+      ...(!appGroupName && {
+        name: '',
+        selectedKey: '#UNASSIGNED_APP#',
```
vikram-raj, line 489: Can we use const `UNASSIGNED_APPLICATIONS_KEY` here?

vikram-raj:
We need to handle the use case for non-admin users. We should display the alert message for the user has only view access for the namespace.

image

invincibleJai:
> We need to handle the use case for non-admin users. We might need to show an alert message instead of loading the form for the user has only view access for the namespace.
> 

@vikram-raj we can get this fixed as separate PR as this is not related to current issue being fixed , WDYT?

vikram-raj:
> > We need to handle the use case for non-admin users. We might need to show an alert message instead of loading the form for the user has only view access for the namespace.
> 
> @vikram-raj we can get this fixed as separate PR/ticket as this is not related to current issue being fixed (can add accesReview even before i.e don't show Kamelet source in catalog if user don't have create access for KB, WDYT?

Agree, we can handle it in a separate issue. And can we hide the Event Source card from the add page if the user has only view access to the namespace?

vikram-raj (commented):
Verified it. Works as expected. 

/lgtm

PR #7483, opened 2020-12-09
Bug 1879013: Improve CD-ROM interface selection

glekner:
Changes:

1. Type row is now above Interface
2. Upon changing Type to CD-ROM, Interface will change to SATA, Virtio option is removed.

image


matthewcarleton:
LGTM!

PR #7485, opened 2020-12-09
Add missing OS name to details popover.

rawagner:
Show 'Not available' if there's no workload profile

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/selectors/vm-template/advanced.ts
+++ b/frontend/packages/kubevirt-plugin/src/selectors/vm-template/advanced.ts
@@ -26,18 +26,13 @@ import {
   getCloudInitVolume,
   getCPU,
   getDataVolumeTemplates,
-  getDisks,
   getFlavor,
   getMemory,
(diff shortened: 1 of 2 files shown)


frontend/packages/kubevirt-plugin/src/selectors/vm-template/advanced.ts
```diff
@@ -162,13 +157,9 @@ export const getTemplateSizeRequirement = (
     sourceSize = convertToBaseValue(getDataVolumeStorageSize(templateSource.dvTemplate));
   }
 
-  return `${pluralize(
-    getDisks(vm).length + (isCDRom ? 1 : 0) + (isWindowsTemplate(template) ? 1 : 0),
```
yaacov, line 166: is it ok to drop the "N disks: ..." here ?

rawagner: Yes, thats on purpose. We want to show only storage requirement, not number of disks anymore

PR #7486, opened 2020-12-09
Bug 1907890: Fix to show correct alert when required in pipeline parameters and resources forms

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5229

**Analysis/Root cause:**
Alert with misleading and incomplete message `Click to save changes or to cancel changes` was shown to the user after adding a empty row even though when Save button is disabled.
During FormFooter comp localization the comp was also refactored but other comps such as PipelineParametersForm and PipelineResourcesForm which are using FormFooter were not fixed accordingly by passing required labels to FormFooter comp to show correct alert message.

**Solution:**
Show alert only when the fields are changed and `Save` is enabled. This PR also adds `resetLabel` and `submitLabel` to show in the alert message which were missed during localization.

**Screenshots:**
image
image
image
image

/kind bug


frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/PipelineResourcesForm.tsx
```diff
@@ -5,6 +5,7 @@ import { FormikProps, FormikValues, getIn } from 'formik';
 import { useAccessReview } from '@console/internal/components/utils';
 import { FormFooter } from '@console/shared';
 import PipelineResources from './PipelineResources';
+import { useTranslation } from 'react-i18next';
```
rottencandy, line 8: Nit: sort imports

rottencandy (commented):
Not sure if this is a bug, but if I delete an existing value and click add, the button gets disabled.
tmp

divyanshiGupta:
> Not sure if this is a bug, but if I delete an existing value and click add, the button gets disabled.
> tmp

@rottencandy it is supposed to work like that since any ways we don't allow saving an empty row and if the user has deleted a row and want to add another in its place either they have to add the required data or they can remove the empty row and then save and the new saved record will not have the deleted row.

rottencandy (approved):
/lgtm

rohitkrai03 (approved):
/approve

PR #7487, opened 2020-12-09
Bug 1866298: Label namespace during storage cluster creation

bipuladh:
- Removed option to label namespace during Operator Installation in OLM.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
@@ -57,7 +57,7 @@ const makeOCSRequest = (
   isEncrypted: boolean,
   isMinimal: boolean,
 ): Promise<any> => {
-  const promises = labelNodes(selectedNodes);
+  const promises: Promise<K8sResourceKind>[] = [...labelNodes(selectedNodes)];
   const ocsObj = getOCSRequestData(
     storageClass,
     defaultRequestSize.BAREMETAL,
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
@@ -1,6 +1,11 @@
 import * as _ from 'lodash';
-import { k8sPatch, NodeKind, StorageClassResourceKind } from '@console/internal/module/k8s';
-import { NodeModel } from '@console/internal/models';
+import {
+  k8sPatch,
+  NodeKind,
+  StorageClassResourceKind,
+  K8sKind,
+} from '@console/internal/module/k8s';
+import { NodeModel, NamespaceModel } from '@console/internal/models';
 import {
   NO_PROVISIONER,
   OCS_INTERNAL_CR_NAME,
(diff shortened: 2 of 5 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -71,6 +76,21 @@ export const labelNodes = (selectedNodes: NodeKind[]): Promise<NodeKind>[] => {
   );
 };
 
+export const labelNamespace = (name: string): Promise<K8sKind> =>
+  k8sPatch(
+    NamespaceModel,
+    {
+      metadata: {
+        name,
```
afreen23, line 84: I shall avoid passing dynamic namespace when we are labeling just a single namespace for ocs - `CEPH_STORAGE_NAMESPACE`

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
```diff
@@ -57,7 +57,7 @@ const makeOCSRequest = (
   isEncrypted: boolean,
   isMinimal: boolean,
 ): Promise<any> => {
-  const promises = labelNodes(selectedNodes);
+  const promises: Promise<K8sResourceKind>[] = [...labelNodes(selectedNodes), labelOCSNamespace()];
```
afreen23, line 60: This is a stale file. We are waiting for 4.6 testing and backports to finish to get rid of this.
```suggestion
  const promises: Promise<K8sResourceKind>[] = [...labelNodes(selectedNodes)];
```
Additionally having a comment that its "Not to be used for 4.7+" would be good.
Its not in scope of the PR, hence wont enforce!

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install-lso-sc.tsx
bipuladh: Thanks for pointing it out. Since it's stale I will keep this file untouched.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
```diff
@@ -71,6 +76,21 @@ export const labelNodes = (selectedNodes: NodeKind[]): Promise<NodeKind>[] => {
...
+    NamespaceModel,
+    {
+      metadata: {
+        name: CEPH_STORAGE_NAMESPACE,
+      },
+    },
+    {
+      op: 'add',
+      path: '/metadata/labels',
+      value: { 'openshift.op/cluster-monitoring': 'true' },
```
afreen23, line 90:
```suggestion
      value: { 'openshift.io/cluster-monitoring': 'true' },
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/ocs-request-data.ts
bipuladh: good catch!

afreen23 (approved):
/lgtm

PR #7488, opened 2020-12-09
Bug 1906898: Missing User RoleBindings in the Project Access Web UI

### LINKED ISSUE #6836: Missing User RoleBindings in the Project Access Web UI
**Describe the bug**

https://github.com/openshift/okd/issues/346

We've one RoleBinding for a project admin which includes two users.
Via the CLI we can see the two users if you describe the rolebinding.

_"[root@amu1123 ~]# oc describe rolebinding gitops-user-binding
Name:         gitops-user-binding
Labels:       argocd.argoproj.io/instance=ns-dev-c2-0eu1-
Annotations:  Role:
  Kind:       ClusterRole
  Name:       admin
Subjects:
  Kind  Name      Namespace
  User  GRESS
  User  HERRNB_F"_

But in the Web UI only the first user is displayed. (Developer -> Project -> Project Access)
image

If you search for RoleBindings two RoleBindings will be displayed:
image

Also If you delete the RoleBinding from the Project Access Web UI the complete RoleBinding will be deleted not only the first User will be removed

**Version**

4.5.0-0.okd-2020-10-03-012432 - UPI

**How reproducible**

Just create a RoleBinding which contains two users.

```
apiVersion: authorization.openshift.io/v1
kind: RoleBinding
metadata:
  name: gitops-user-binding
roleRef:
  name: admin
subjects:
- kind: User
  name: GRESS
- kind: User
  name: HERRNB_F
userNames:
- GRESS
- HERRNB_F
```

**Log bundle**

debsmita1:
**Fixes**:
https://issues.redhat.com/browse/ODC-4989
Closes #6836 

**Analysis / Root cause**: 
The User info is added in the `subjects` property of the RoleBinding CR. This `subjects` is of type array and can have multiple users assigned the same role. So, previously while fetching the users from all the Role Bindings only the first element of `subjects` was being read

**Solution Description**: 
- go over all the role binding subjects to fetch users
- if the role of one of the grouped subjects is changed, then it is deleted from the grouped subjects & a new role binding is created for that user

pa

Added new tests

**Test setup:**
NA

**Browser conformance:** 
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
```diff
@@ -1,24 +1,27 @@
...
-export const getUserRoleBindings = (roleBindings) => {
+export const getUsersFromSubject = (user: RoleBinding) => {
+  const users: UserRoleBinding[] = user.subjects?.reduce((acc, obj) => {
+    acc.push({
+      roleBindingName: user.metadata.name,
+      user: obj.name,
+      role: user.roleRef.name,
+    });
+    return acc;
+  }, []);
```
jerolimov, line 16: Isn't this a map function? :)

frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
```diff
@@ -1,24 +1,27 @@
...
-        ...userRoleBindings,
-        ...[
-          {
-            roleBindingName: user.metadata.name,
-            user: user.subjects[0].name,
-            role: user.roleRef.name,
-          },
-        ],
-      ]),
+    (user: RoleBinding) => (userRoleBindings = [...userRoleBindings, ...getUsersFromSubject(user)]),
```
jerolimov, line 24: Its similar to before, but the result of a map function should be used, otherwise can we switch to a .forEach here? Isn't this similar to this code?

```tsx
export const getUserRoleBindings = (roleBindings: RoleBinding[]) => {
  const userRoleBindings: UserRoleBinding[] = [];
  roleBindings.forEach(
    (user: RoleBinding) => userRoleBindings.push(...getUsersFromSubject(user)),
  );
  return userRoleBindings;
};
```

(untested code). Wdyt? :)

frontend/packages/dev-console/src/components/project-access/project-access-form-submit-utils.ts
```diff
@@ -73,3 +76,14 @@ export const sendRoleBindingRequest = (verb: string, roles: UserRoleBinding[], r
   });
   return finalArray;
 };
+
+export const getGroupedRole = (role: UserRoleBinding, roleBindings: RoleBinding[]) => {
+  let groupedRole: RoleBinding;
+  _.forEach(roleBindings, (roleBinding: RoleBinding) => {
+    if (roleBinding.metadata.name === role.roleBindingName && roleBinding.subjects.length > 1) {
+      _.remove(roleBinding.subjects, (subject) => subject.name === role.user);
```
jerolimov, line 84: Hmm, not a big fan of changing input parameters if a 'get' helper function. Esp in this case where getGroupedRole was called with some props we should avoid this.

@debsmita1 Did you see an option that this method return new values instead of modifying the input variables?

frontend/packages/dev-console/src/components/project-access/__tests__/project-access-form-utils.spec.tsx
```diff
@@ -1,20 +0,0 @@
...
-  it('should fetch the only the required rolebindings', async () => {
-    const filteredRoleBindings = filterRoleBindings(roleBindingsResourceData, Roles);
-    expect(filteredRoleBindings).toEqual(roleBindingsWithRequiredRolesResult);
-  });
-
-  it('should consider only the required attributes', async () => {
-    const userRoleBindings = getUserRoleBindings(roleBindingsWithRequiredRoles);
-    expect(userRoleBindings).toEqual(roleBindingsWithRequiredAttributes);
-  });
-});
```
jerolimov, line 20: Why remove this unit tests? The util functions still exist and are not changed, or? So I expect this tests would still run fine?

jerolimov (commented):
@debsmita1 Happy new year and sorry for the late review here. I tested the PR locally and it looks like that everything works fine for the user and I like all these new types :+1:

I added some small opinionated ideas and questions. Let me know what your opinion on that points.

jerolimov (changes requested):
-

frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
andrewballantyne, line 24:
```suggestion
  return roleBindings.map((user) => getUsersFromSubject(user)).flat();
```
Two major things here:
1. Don't assign in a return statement (we should probably get a linter to stop this)
2. Avoid mutations in the middle of `.map` function calls (the purer they are the better they are)

frontend/packages/dev-console/src/components/project-access/project-access-form-submit-utils.ts
andrewballantyne, line 84: Is there a reason you need to remove the matching subject?

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/project-access/project-access-form-submit-utils.ts
debsmita1: @andrewballantyne What I am trying to do here is, if I update the role of a user(suppose A) that belongs to grouped subjects(subjects: A, B) then  I am removing that user(A) from the group and sending a patch request to update the rolebinding with the user B.

frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
debsmita1: yes, I could have used `flat` here to concatenate the sub-array elements. Thanks!

debsmita1: used `.flat` here 
added the test for this util in `project-access-form-utils.spec.ts`

frontend/packages/dev-console/src/components/project-access/project-access-form-submit-utils.ts
debsmita1: fixed it

frontend/packages/dev-console/src/components/project-access/__tests__/project-access-form-utils.spec.tsx
debsmita1: fixed. Changed the file name to `project-access-form-utils.spec.ts`

frontend/packages/dev-console/src/components/project-access/project-access-form-submit-utils.ts
```diff
@@ -73,3 +76,14 @@ export const sendRoleBindingRequest = (verb: string, roles: UserRoleBinding[], r
   });
   return finalArray;
 };
+
+export const getGroupedRole = (role: UserRoleBinding, roleBindings: RoleBinding[]) => {
```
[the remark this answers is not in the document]
debsmita1: used find & filter

andrewballantyne (approved):
/lgtm

PR #7493, opened 2020-12-09
Remove border rule since upstream fix has merged

sg00dwin:
https://github.com/patternfly/patternfly/pull/2963


spadgett (approved):
/lgtm

PR #7494, opened 2020-12-09
Bug 1905941: Add optional chaining to catalog source Status field

harishsurf:
Add optional chaining to catalog source Status field

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
@@ -483,7 +483,7 @@ const flatten = ({
           operatorCount: getOperatorCount(catalogSource, packageManifests.data),
           publisher: catalogSource.spec.publisher,
           registryPollInterval: getRegistryPollInterval(catalogSource),
-          status: catalogSource.status.connectionState.lastObservedState,
+          status: catalogSource.status?.connectionState?.lastObservedState,
         }),
       };
     },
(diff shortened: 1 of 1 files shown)


frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -483,7 +483,7 @@ const flatten = ({
           operatorCount: getOperatorCount(catalogSource, packageManifests.data),
           publisher: catalogSource.spec.publisher,
           registryPollInterval: getRegistryPollInterval(catalogSource),
-          status: catalogSource.status.connectionState.lastObservedState,
+          status: catalogSource?.status?.connectionState?.lastObservedState,
```
spadgett, line 486: We know `catalogSource` isn't `undefined` since it would blow up on the lines above

```suggestion
          status: catalogSource.status?.connectionState?.lastObservedState,
```

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
```diff
@@ -501,7 +501,7 @@ const flatten = ({
     operatorHub,
     publisher: source.spec.publisher,
     registryPollInterval: getRegistryPollInterval(source),
-    status: source.status.connectionState.lastObservedState,
+    status: source?.status?.connectionState?.lastObservedState,
```
spadgett, line 504:
```suggestion
    status: source.status?.connectionState?.lastObservedState,
```

spadgett (commented):
Thanks @harishsurf

frontend/packages/operator-lifecycle-manager/src/components/catalog-source.tsx
harishsurf: True. I updated it

spadgett (approved):
/lgtm

PR #7502, opened 2020-12-09
Bug 1905621: Protractor login test fails against a 4.7 (nightly) Power cluster

dtaylor113:
Increased Protractor login redirect which seems to take a little longer on Power clusters.


spadgett (approved):
/lgtm

PR #7503, opened 2020-12-10
Bug 1906318: kubevirt: use proper term for Authorized SSH Keys

dankenigsberg:
cloud-init's ssh_authorized_keys should be called that way in code and
GUI because their important function is to AUTHORIZE a user to log into
a VM, not mere authentication. I suspect that the existing term was
introduced by mistake.

Signed-off-by: Dan Kenigsberg <danken@redhat.com>


dankenigsberg:
@jelkosz @yaacov would you consider this? The current text is an eye sore (once I spotted it)

yaacov:
/retitle Bug 1906318: kubevirt: use proper term for Authorized SSH Keys

PR #7505, opened 2020-12-10
Bug 1897003: VNC console main window can auto reconnect when full screen vnc console tab is closed

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1897003

**Analysis / Root cause**: 
Main window and new open tab/window, didn't had a way to communicate

**Solution Description**: 
Added a var to localStorage in order to communicate between main window and new tab/window

vnc-console

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/connected-vm-console/connected-vm-console.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/connected-vm-console/connected-vm-console.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { K8sResourceKind, PersistentVolumeClaimKind, PodKind } from '@console/internal/module/k8s';
 import { PersistentVolumeClaimModel, PodModel, ServiceModel } from '@console/internal/models';
 import { getVMStatus } from '../../statuses/vm/vm-status';
+import { getName } from '@console/shared';
 import { VMImportKind } from '../../types/vm-import/ovirt/vm-import';
 import { V1alpha1DataVolume } from '../../types/vm/disk/V1alpha1DataVolume';
 import {
@@ -17,6 +18,8 @@ import { VMConsolesWrapper } from '../vms/vm-console';
 import { getLoadedData } from '../../utils';
 import { ConsoleEmptyState } from './vm-console-empty-state';
 import { ConsoleType } from '../../constants/vm/console-type';
+import { useEventListener } from '../../hooks/use-event-listener';
+import { useRenderVNCConsole } from '../../hooks/use-render-vnc-console';
 
 const ConnectedVMConsole: React.FC<ConnectedVMConsoleProps> = ({
   type,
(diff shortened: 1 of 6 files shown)


metalice:
@yaacov , @rawagner please review :)

frontend/packages/kubevirt-plugin/src/hooks/use-render-vnc-console.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const useRenderVNCConsole = ({
+  shouldBeFullScreen,
+  initValue,
+}: RenderVNCConsoleArguments): boolean => {
+  const value = !!localStorage.getItem('isFullScreenVNC') || initValue;
+  const [renderVNCConsole, setRenderVNCConsole] = React.useState<boolean>(value);
+
+  useEventListener(window, 'storage', () => {
+    setRenderVNCConsole(!!localStorage.getItem('isFullScreenVNC'));
```
yaacov, line 17: will it work with multiple consoles open from different VMs ?

metalice: yes, in updated version, it is.

frontend/packages/kubevirt-plugin/src/hooks/use-render-vnc-console.ts
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { useEventListener } from './use-event-listener';
+
+interface RenderVNCConsoleArguments {
+  vmName: string;
+  shouldBeFullScreen: boolean;
+  initValue?: boolean;
+}
+
+export const useRenderVNCConsole = ({
```
rawagner, line 10: I think the implementation can be improved:

- We dont need `initValue` (as its the same as `shouldBeFullScreen`)
- you should reuse `useLocalStorage` here (it will need to be enhanced with `useEventListener(window, 'storage',..`)
- hook should return setter method for setting localstorage key value so you dont have to do things like

```js
useEventListener(window, 'beforeunload', () =>
    localStorage.removeItem(`isFullScreenVNC-${vmName}`),
  );
```
instead you would do
```
useEventListener(window, 'beforeunload', () =>
    setRenderVNCConsole(false);
  );
```

frontend/packages/kubevirt-plugin/src/components/vms/vm-console.tsx
```diff
@@ -129,20 +131,20 @@ const VMConsoles: React.FC<VMConsolesProps> = ({
...
-
   return (
     <Stack hasGutter>
       {showOpenInNewWindow && consoleType && (
         <StackItem>
           <Button
             variant="secondary"
-            onClick={() =>
+            onClick={() => {
+              localStorage.setItem(`isFullScreenVNC-${vmName}`, 'true');
```
rawagner, line 141: this should be done in `connected-vm-console` to make sure everytime we open it (even from other places in the future), the localstorage value will be set properly

frontend/packages/kubevirt-plugin/src/components/connected-vm-console/connected-vm-console.tsx
```diff
@@ -36,6 +39,15 @@ const ConnectedVMConsole: React.FC<ConnectedVMConsoleProps> = ({
   const loadedDataVolumes = getLoadedData(dataVolumes);
   const loadedImports = getLoadedData(vmImports);
   const vmi = loadedVMIs?.[0];
+  const vmName = getName(loadedVM);
+  const renderVNCConsole = useRenderVNCConsole({
+    vmName,
```
rawagner, line 44: we also need to make sure it will work only with VMI without VM object

rawagner (changes requested):

yaacov:
@metalice need to run `yarn i18n` again, strings changed and the tests are failing

irosenzw:
@metalice can you please squash all these commits together ?

PR #7507, opened 2020-12-10
Bug 1906689: Fix user can pin to nav configmaps and secrets multiple times

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5176

**Analysis / Root cause**: 
By default the developer console comes with ConfigMaps and Secrets pinned to the main navigation. and user can pin to nav configmaps and secrets multiple times.

**Solution Description**: 
Fix now can not able to pin default pin resource ConfigMaps and Secrets

pinned

### DIFF
--- a/frontend/packages/dev-console/src/plugin.tsx
+++ b/frontend/packages/dev-console/src/plugin.tsx
@@ -38,6 +38,7 @@ import {
   ConfigMapModel,
   DeploymentModel,
 } from '@console/internal/models';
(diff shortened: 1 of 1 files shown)


vikram-raj:
/retitle Bug 1906689: Fix user can pin to nav configmaps and secrets multiple times

sahil143 (approved):
/lgtm

verified changes locally. works as expected.

PR #4517, opened 2020-02-26
Bug 1807665: Edit Environment variables in edit flows

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-3134
Add support for editing environment variables under Advanced Section (Buildconfig, Deployment)

Problem:
Editing the evironment variables feature was missing
Solution:
Added the namevalue editor and prepopulating the environment variables in the editor so that user can add/remove/update the values.

Video-Thu-Feb-27-2020-02-17-05

PR #7515, opened 2020-12-10
Bug 1867965: Don't override volumes and volumeMounts in D/DC

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-4953

**Solution:**
Preserve the `originalResource` `volumes` and `volumeMounts` while updating D/DC using edit flows.

Peek 2020-12-14 18-00

**Unit tests:**
image

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/utils/__tests__/resource-label-utils.spec.tsx
+++ b/frontend/packages/dev-console/src/utils/__tests__/resource-label-utils.spec.tsx
@@ -1,0 +1,84 @@
+import { mergeData } from '../resource-label-utils';
+import {
+  devfileDeployment,
+  newBuildConfig,
+  newDeployment,
+  newDeploymentConfig,
+  originalBuildConfig,
+  originalDeployment,
+  originalDeploymentConfig,
+} from './resource-label-utils-data';
+
+describe('resource-label-utils', () => {
+  describe('mergeData', () => {
+    it('should return newResource when originalResource is empty', () => {
+      const mergedResource = mergeData({}, newDeployment);
+      expect(mergedResource).toEqual(newDeployment);
+    });
+
+    it('should return mergedData with newResource labels', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
+      expect(mergedResource.metadata.labels).toEqual(newDeployment.metadata.labels);
+    });
+
+    it('should return mergedData with newResource and originalResource labels if originalResource is a devfile resource', () => {
+      const mergedResource = mergeData(devfileDeployment, newDeployment);
+      expect(mergedResource.metadata.labels).toEqual({
+        ...newDeployment.metadata.labels,
+        ...devfileDeployment.metadata.labels,
+      });
+    });
+
+    it('should return mergedData with newResource annotations ', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
+      expect(mergedResource.metadata.annotations).toEqual(newDeployment.metadata.annotations);
+    });
+
+    it('should return mergedData with newResource and originalResource annotations if originalResource is a devfile resource', () => {
+      const mergedResource = mergeData(devfileDeployment, newDeployment);
+      expect(mergedResource.metadata.annotations).toEqual({
+        ...newDeployment.metadata.annotations,
+        ...devfileDeployment.metadata.annotations,
+      });
+    });
+
+    it('should return mergedData with newResource template labels', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
+      expect(mergedResource.spec.template.metadata.labels).toEqual(
+        newDeployment.spec.template.metadata.labels,
+      );
+    });
+
+    it('should return mergedData with newResource template containers', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
+      expect(mergedResource.spec.template.spec.containers).toEqual(
+        newDeployment.spec.template.spec.containers,
+      );
+    });
+
+    it('should return mergedData with newResource strategy', () => {
+      const mergedResource = mergeData(originalBuildConfig, newBuildConfig);
+      expect(mergedResource.spec.strategy).toEqual(newBuildConfig.spec.strategy);
+    });
+
+    it('should return mergedData with originalResource strategy if newResource strategy is undefined', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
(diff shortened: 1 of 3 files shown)


frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -80,6 +80,8 @@ export const getPodLabels = (name: string) => {
 };
 
 export const mergeData = (originalResource: K8sResourceKind, newResource: K8sResourceKind) => {
```
andrewballantyne, line 82: Can we write tests for this function?

divyanshiGupta: Done.

frontend/packages/dev-console/src/utils/__tests__/resource-label-utils.spec.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+describe('resource-label-utils', () => {
+  describe('mergeData', () => {
+    it('should return newResource when originalResource is empty', () => {
+      const mergedResource = mergeData({}, newDeployment);
+      expect(mergedResource).toEqual(newDeployment);
+    });
+
+    it('should return the mergedResource with correct data', () => {
+      const mergedResource = mergeData(originalDeployment, newDeployment);
```
andrewballantyne, line 12: This is good initial tests, but we should probably have a test (`it`) for each if block.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -92,11 +94,16 @@ export const mergeData = (originalResource: K8sResourceKind, newResource: K8sRes
     mergedData.spec.template.spec.containers = newResource.spec.template.spec.containers;
   }
   if (mergedData?.spec?.hasOwnProperty('strategy')) {
-    mergedData.spec.strategy = newResource.spec?.strategy ?? originalResource?.spec?.strategy;
+    mergedData.spec.strategy = newResource.spec?.strategy ?? originalResource.spec?.strategy;
   }
   if (mergedData.spec?.triggers) {
     mergedData.spec.triggers = newResource.spec.triggers;
   }
+  if (mergedData.spec?.template?.spec?.hasOwnProperty('volumes')) {
```
andrewballantyne, line 102: Still a little perplexed by why we need this... This feels like we won't really catch everything that one could add to Deployments.

Essentially this solution feels like it will close the specific ask we had in the bug but not address the underlying issue of "parameters being lost if we are not explicitly calling them out". 🤔

andrewballantyne (changes requested):
Need more tests I think (see comment).

This is a bizarre method of "some fields override specially"... not sure if this is the best solution but for the use-case we have, this might be the best for the bug-fixing sprints.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
divyanshiGupta: @andrewballantyne the issue here is that we dont have a way for adding/editing volumes/volumeMounts from the create/edit app forms but volumes/volumeMounts can be added/updated directly in the yaml itself so currently the solution that makes sense to me is to preserve whatever volumes/volumeMounts we have in the originalResource.

andrewballantyne: I understand the reason for the change -- what I have confusion around is the way the fix is implemented. I know this method has other statements like this, but it seems we are putting a bandaid on a stab wound. We are not allowing properties to flow through (like a spread or merge of objects), so we are stuck to manually green lighting each property as someone raises it to us. 

Not sure we need to address this issue here, it's a general issue with the edit flow in general today for all the resources we maintain. But it's definitely concerning.

frontend/packages/dev-console/src/utils/__tests__/resource-label-utils.spec.tsx
andrewballantyne: I believe we should still get more tests here. If nothing more than a specific test for volumes and a test for no volumes (ie the fix you did).

divyanshiGupta: Sure, will add more tests.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
divyanshiGupta: Do you have a suggestion on how do we handle this here?

divyanshiGupta: Do you have a suggestion on how do we handle this here?

andrewballantyne: Not without more tests 🙂 

Essentially you are already merging the two objects, I think if we deep merge it ... it MIGHT solve our issue. But I am a bit concerned about breaking everything else haha. I don't think this is the best time to do it. We can fix the direct issue now and then look to tighten up things in the future.

andrewballantyne (approved):
/lgtm

frontend/packages/dev-console/src/utils/resource-label-utils.ts
andrewballantyne: @divyanshiGupta It occurred again, I think we need to go through with this now.
RE: https://bugzilla.redhat.com/show_bug.cgi?id=1929577

divyanshiGupta: Yes, it's because how these statements got added without considering what happens to the old values. This particular issue is occurring because the old containers are directly replaced with what we generate not taking in account specific container attribute values like envs and that these values can be added from elsewhere so we cannot override everything.
https://github.com/openshift/console/pull/4517/files#diff-7fd530a4a1b4113f74f1e7e64beadbdb448845c1fad06d6a4d4824425583c2d6R53-R55

I think we should create a tech debt for this and then I can work on refactoring this function and add some unit tests.

divyanshiGupta: https://issues.redhat.com/browse/ODC-5539

andrewballantyne: We could also fix it part of the BZ I logged... this is the 2nd customer bug we have on this function; it needs to be fixed sooner rather than later.

PR #7516, opened 2020-12-10
Bug 1908035: Fix issue with dynamic-demo-plugin webpack build

vojtechszocs:
Console dynamic plugin SDK (and consequently `dynamic-demo-plugin`) currently uses webpack `5.0.0-beta.16` which doesn't support empty `entry` object yet. Quote from [webpack docs](https://webpack.js.org/concepts/entry-points/#object-syntax):

> You can pass empty object `{}` to `entry` when you have only entrypoints generated by plugins.

This will be fixed once we bump dynamic plugin SDK dependencies.

In the meantime, this PR fixes an issue where `dynamic-demo-plugin/dist` wasn't generated at all due to changes made in #7163 (`ConsoleAssetPlugin` not using `afterCompile` hook).

---

_Tech details:_ Dynamic plugin projects are not typical web applications with webpack-processed entry points. All dynamic plugin assets are generated via `ConsoleRemotePlugin` (which uses `ConsoleAssetPlugin` under the hood). However, webpack 4 requires us to specify at least one entry point (from which a bundle gets created). We're currently working around this issue, with webpack 5 latest this should be a non-issue.

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleAssetPlugin.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/webpack/ConsoleAssetPlugin.ts
@@ -24,7 +24,6 @@ const emitJSON = (compilation: webpack.Compilation, filename: string, data: any)
   // we just provide the content (source) and its length (size).
 
   // TODO(vojtech): revisit after bumping webpack 5 to latest stable version
-
   // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
   // @ts-ignore
   compilation.emitAsset(filename, {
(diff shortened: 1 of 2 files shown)


spadgett:
If we move to the latest beta, v5.0.0-beta.31, would that address the issue?

vojtechszocs:
> If we move to the latest beta, v5.0.0-beta.31, would that address the issue?

@spadgett I'd suggest moving to latest v5 stable (currently `5.10.0`) since there's been a lot of changes around module federation code between `5.0.0-beta.16` and the latest beta anyway.

jhadvig (approved):
/lgtm

PR #7518, opened 2020-12-10
Bug 1906770: Hide topology shortcuts in mobile view

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-2382

**Analysis / Root cause**: 
Short cuts in topology view take up a log of space at mobile view

**Solution Description**: 
Add a hook to watch window size and to detect when resolution is considered mobile.
Check for mobile and when at mobile resolution do not show the shortcuts

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

/kind bug


christianvogt:
@jeff-phillips-18 I think you can use `window.matchMedia` along with a `change` listener here.

PR #7519, opened 2020-12-11
Bug 1906683: fixes issue with kn resources if trigger goes to ksvc and IMC from same broker

invincibleJai:
**Fixes:**
https://issues.redhat.com/browse/ODC-5254

**Analysis / Root cause:**
Kn resources are not showing in Topology if triggers has KSVC and IMC as subscriber due to a NPE

**Solution Description:**
Fixed the NPE , so that kn data model returns proper data

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified and works as expected
/lgtm

PR #7520, opened 2020-12-11
Bug 1906685: show sinkbinding as resources in sidebar if owned by source

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5228

**Analysis / Root cause**: 
SinkBinding is shown in topology view along with actual source created in case of Container source and KameletBinding Creation

**Solution Description**: 
Show show only actual source created in topology view and sinkbinding in sidebar resources here (as is underlying resource here)

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/components/overview/EventSourceResources.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/EventSourceResources.tsx
@@ -14,12 +14,14 @@ import {
 } from '@console/internal/components/utils';
 import { PodModel } from '@console/internal/models';
 import { PodsOverview } from '@console/internal/components/overview/pods-overview';
+import EventSourceOwnedList from './EventSourceOwnedList';
 
-export type EventSinkServicesOverviewListProps = {
+type EventSourceResourcesProps = {
   obj: K8sResourceKind;
+  ownedSources?: K8sResourceKind[];
 };
 
-const EventSinkServicesOverviewList: React.FC<EventSinkServicesOverviewListProps> = ({ obj }) => {
+const EventSourceResources: React.FC<EventSourceResourcesProps> = ({ obj, ownedSources }) => {
   const { t } = useTranslation();
   const {
     kind,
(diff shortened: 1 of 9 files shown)


frontend/packages/knative-plugin/src/components/overview/EventSourceResources.tsx
```diff
@@ -15,11 +17,30 @@ import {
...
 import { PodsOverview } from '@console/internal/components/overview/pods-overview';
 
-export type EventSinkServicesOverviewListProps = {
+export type EventSourceResourcesProps = {
   obj: K8sResourceKind;
+  ownedSources?: K8sResourceKind[];
 };
 
-const EventSinkServicesOverviewList: React.FC<EventSinkServicesOverviewListProps> = ({ obj }) => {
+export const OwnedSourcesList: React.FC<{ source: K8sResourceKind }> = ({ source }) => {
```
karthikjeeyar, line 25: It would be better to have one component per file. Moving this out to new a file and we can then add its own test cases.

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -918,6 +919,27 @@ export const getRouteData = (resource: K8sResourceKind, ksroutes: K8sResourceKin
...
 };
 
+const getOwnedEventSourceData = (
+  resource: K8sResourceKind,
+  data: TopologyDataObject,
+  resources,
+) => {
+  const eventSourceProps = [...getDynamicEventSourcesModelRefs(), CamelKameletBindingModel.plural];
+  const eventSourcesData = getKnativeDynamicResources(resources, eventSourceProps);
+  let ownnedSourceData = getOwnedResources(resource, eventSourcesData);
```
karthikjeeyar, line 929: nit: typo
```suggestion
  let ownedSourceData = getOwnedResources(resource, eventSourcesData);
```

PR #7523, opened 2020-12-11
Bug 1906896: show empty message for no alerts

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5116

**Analysis / Root cause:**

**Solution Description:**
- added a loader
- it shows `No Alerts found` if there are no alerts

Added new test

Screenshot from 2020-12-11 15-28-07
alerts

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
@@ -13,7 +13,7 @@ import { getAlertsAndRules } from '@console/internal/components/monitoring/utils
 import { monitoringSetRules, monitoringLoaded, sortList } from '@console/internal/actions/ui';
 import { Rule } from '@console/internal/components/monitoring/types';
 import { RootState } from '@console/internal/redux';
-import { getURLSearchParams } from '@console/internal/components/utils';
+import { getURLSearchParams, EmptyBox, LoadingBox } from '@console/internal/components/utils';
 import { getFilteredRows } from '@console/internal/components/factory';
 import { alertingRuleStateOrder } from '@console/internal/reducers/monitoring';
 import { usePrometheusRulesPoll } from '@console/internal/components/graphs/prometheus-rules-hook';
@@ -107,6 +107,13 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
     );
     setSortBy({ index, direction });
   };
+
+  if (loading && !loadError) {
+    return <LoadingBox />;
+  }
+  if (_.isEmpty(response?.data?.groups)) {
+    return <EmptyBox label={t('devconsole~Alerts')} />;
+  }
 
   return (
     <>
--- a/frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/MonitoringAlerts.spec.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/MonitoringAlerts.spec.tsx
@@ -3,8 +3,10 @@ import { Map } from 'immutable';
 import * as redux from 'react-redux';
 import { shallow } from 'enzyme';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import * as prometheusHook from '@console/internal/components/graphs/prometheus-rules-hook';
 import { FilterToolbar } from '@console/internal/components/filter-toolbar';
 import { MonitoringAlerts } from '../MonitoringAlerts';
+import { EmptyBox } from '@console/internal/components/utils';
 
 jest.mock('react-i18next', () => {
   const reactI18next = require.requireActual('react-i18next');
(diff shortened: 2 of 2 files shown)


sahil143 (changes requested):
@debsmita1 When visiting the Alerts tab page starts to hang and becomes unresponsive. according to the warning in the console, setState is being called repeatedly.
Screenshot 2020-12-11 at 7 21 33 PM

sahil143:
@debsmita1 Unit tests are failing

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
```diff
@@ -147,6 +147,13 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
     setSortBy({ index, direction });
   };
 
+  if (loading && !loadError) {
+    return <LoadingBox />;
+  }
+  if (thanosAlertsAndRules?.alerts?.length === 0) {
+    return <EmptyBox label="Alerts" />;
```
karthikjeeyar, line 154: localize the string.

debsmita1:
> @debsmita1 When visiting the Alerts tab page starts to hang and becomes unresponsive. according to the warning in the console, useEffect hook is being called repeatedly.
> Screenshot 2020-12-11 at 7 21 33 PM

@sahil143  This issue is being fixed in PR https://github.com/openshift/console/pull/7480

debsmita1:
> @debsmita1 Unit tests are failing

fixed

frontend/packages/dev-console/src/components/monitoring/alerts/__tests__/MonitoringAlerts.spec.tsx
```diff
@@ -38,11 +40,70 @@ describe('MonitoringAlerts', () => {
...
     expect(wrapper.find(FilterToolbar).exists()).toBe(true);
     expect(wrapper.find(Table).exists()).toBe(true);
     expect(wrapper.find(TableHeader).exists()).toBe(true);
     expect(wrapper.find(TableBody).exists()).toBe(true);
   });
+  spyPrometheusRulesPoll.mockReturnValueOnce([{}, null, false]);
+  it('should show empty state message', () => {
+    const wrapper = shallow(<MonitoringAlerts {...monitoringAlertsProps} />);
+    expect(wrapper.find(EmptyBox).exists()).toBe(true);
+  });
```
jerolimov, line 108: @debsmita1 Really great to see more tests coming. In jest all testcases and code which is required to run a test case should be part of a `it` or `beforeEach`, `afterEach`, etc. function. Only test setup calls which are similar for all tests could be done before all tests.

In your case you modify the `spyPrometheusRulesPoll` mock between two test functions. This works only fine when you run your test from top to bottom 'once'. The first mock call saved one result, the second one the second (empty) version.

The `it` calls just register "test" calls (your callback function) and it will be run later. When you add console.log statements you can see that the tests are runned in this order:

```tsx
console.log('1')
it('test 1' => {
  console.log('3')
})
console.log('2')
it('test 2' => {
  console.log('4')
})
```

And this is an issue if you don't run all tests at the same time or in a different order.  You can test this with this command:

`yarn test packages/dev-console/src/components/monitoring/alerts/__tests__/MonitoringAlerts.spec.tsx -t empty`

This runs both mock "Once" calls and then runs the second it callback. But the mock returns the first value here.

So the general idea should be that each test works completely independent. You can reach this by just moving the setup of `spyPrometheusRulesPoll` also into the it function.

jerolimov (changes requested):
Empty message was shown and I really like the new tests here, but found a small nit in the tests, also if they are green.
After this small change I would like to add a lgtm here. :)

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
```diff
@@ -108,6 +108,13 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
     setSortBy({ index, direction });
   };
 
+  if (loading && !loadError) {
+    return <LoadingBox />;
+  }
+  if (_.isEmpty(response?.data)) {
+    return <EmptyBox label={t('devconsle~Alerts')} />;
```
vikram-raj, line 115:
```suggestion
    return <EmptyBox label={t('devconsole~Alerts')} />;
```

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
```diff
@@ -108,6 +108,13 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
     setSortBy({ index, direction });
   };
 
+  if (loading && !loadError) {
+    return <LoadingBox />;
+  }
+  if (_.isEmpty(response?.data)) {
```
vikram-raj, line 114:
```suggestion
  if (_.isEmpty(response?.data?.groups)) {
```

vikram-raj (changes requested):

vikram-raj (commented):
Thanks @debsmita1 

/lgtm

PR #7528, opened 2020-12-11
Bug 1909067: Keep pod terminal output after connection closes

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4935

**Analysis / Root cause**: 
When open a terminal on a Pod, the terminal output disappears as soon as the user exit it or the connection was terminated. But the terminal could contain essential informations about the issue or at least infos which the user is still interessted.

**Solution Description**: 
Preserve the terminal output after the connection is closed.

tmp

@openshift/team-devconsole-ux 
/cc @beaumorley 

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/public/components/pod-exec.jsx
+++ b/frontend/public/components/pod-exec.jsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import * as _ from 'lodash-es';
 import { Base64 } from 'js-base64';
+import { withTranslation } from 'react-i18next';
 import { ExpandIcon } from '@patternfly/react-icons';
-import { Button } from '@patternfly/react-core';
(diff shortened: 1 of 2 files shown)


beaumorley:
lgtm thank you

jerolimov:
As discussed in slack with Saul we should add the reconnect option only for Running Pods.

andrewballantyne (commented):
/assign @christianvogt

christianvogt:
Make the window vertically very small so that you need to scroll down to see the terminal. Then type `exit` and press the reconnect button. An exception is thrown:

image

christianvogt:
Cannot reproduce the error anymore.
/lgtm
/approve

PR #7531, opened 2020-12-11
Bug 1906840:  Change ClusterOperator status 'Updating' to 'Progressing'

rhamilto:
to reflect the name of the condition type since updating is a subset of progressing, and we now provide a progress bar showing operator updates on the Cluster Settings page.

### DIFF
--- a/frontend/packages/console-app/src/components/dashboards-page/status.ts
+++ b/frontend/packages/console-app/src/components/dashboards-page/status.ts
@@ -118,7 +118,7 @@ export const getClusterOperatorStatusPriority: GetOperatorStatusPriority<Cluster
   if (status === OperatorStatus.Unknown) {
     return { ...healthStateMapping[HealthState.UNKNOWN], title: status };
   }
-  if (status === OperatorStatus.Updating) {
+  if (status === OperatorStatus.Progressing) {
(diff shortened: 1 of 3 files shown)


spadgett:
cc @smarterclayton

frontend/public/module/k8s/cluster-settings.ts
```diff
@@ -87,13 +87,15 @@ export const getClusterVersionCondition = (
...
-    ),
+export const updateInProgress = (cv: ClusterVersionKind): boolean => {
+  return (
+    !_.isEmpty(
+      getClusterVersionCondition(
+        cv,
+        ClusterVersionConditionType.Progressing,
+        K8sResourceConditionStatus.True,
+      ),
+    ) && getDesiredClusterVersion(cv) !== getLastCompletedUpdate(cv)
```
spadgett, line 98: Sorry, I probably wasn't clear in the Bugzilla. The bug was specifically for ClusterOperator status rather than ClusterVersion. (The logic here: https://github.com/openshift/console/blob/master/frontend/public/module/k8s/cluster-operator.ts#L18-L21)

spadgett (approved):
/lgtm

PR #7534, opened 2020-12-14
Bug 1907282: fix topology page white screen, use ternary operator instead of logical operator

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5258

**Analysis / Root cause**: 
We are using logical operator in the PodSet component which causes it to pass a boolean value to PodStatus component title/subtitle and since we started to truncate the title/subtitle using `_.truncate` in PodStatus component with PR #7447, boolean values are converted to string by lodash truncate `false => 'false'`.

**Solution Description**: use ternary operator instead of logical operator and pass `undefined` if showPodCount is false

NO UI changes

**Test setup:**

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/pod/PodStatus.tsx
+++ b/frontend/packages/console-shared/src/components/pod/PodStatus.tsx
@@ -51,8 +51,8 @@ const PodStatus: React.FC<PodStatusProps> = ({
   size = 130,
   standalone = false,
   showTooltip = true,
-  title = '',
-  subTitle = '',
+  title,
+  subTitle,
   titleComponent,
   subTitleComponent,
   data,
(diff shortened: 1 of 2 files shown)


rohitkrai03 (approved):
/approve

The changes in the PR works fine to fix the issue and does make sense but I would like to point out that the analysis/root cause in the description mentions that the issue is caused because boolean value is being passed to `_.truncate` and it's changing `false` => `'false'`. But that wouldn't break the UI since the prop accepts a string as valid type, so the main issue was being caused because `false` was being passed to `titleComponent` prop. Its fixed after these changes because now `titleComponent` receives `undefined` instead of `false`. cc: @sahil143

sahil143:
> so the main issue was being caused because false was being passed to titleComponent prop. Its fixed after these changes because now titleComponent receives undefined instead of false

Agreed, titleComponent wasn't being invoked since title itself was a boolean `false` and since translate converted it to a string it became `true` and titleComponent was invoked which was a boolean and not a component.

PR #7537, opened 2020-12-14
Bug 1892896: Disables the block radio input if the provisioner is cephfs

vbnrh:
Disables the block radio input if the selected storage class provisioner is cephfs

Disabling radio input
Screenshot from 2020-12-14 15-19-26

Having readonly text
Screenshot from 2021-01-21 20-29-07

### DIFF
--- a/frontend/public/components/storage/create-pvc.tsx
+++ b/frontend/public/components/storage/create-pvc.tsx
@@ -2,8 +2,9 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
 import { Link } from 'react-router-dom';
-import { ActionGroup, Button } from '@patternfly/react-core';
-import { isCephProvisioner, isObjectSC } from '@console/shared/src/utils';
+import { ActionGroup, Button, Tooltip } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import { isCephProvisioner, isObjectSC, cephStorageProvisioners } from '@console/shared/src/utils';
 import { k8sCreate, K8sResourceKind, referenceFor } from '../../module/k8s';
 import { AsyncComponent, ButtonBar, RequestSizeInput, history, resourceObjPath } from '../utils';
 import { StorageClassDropdown } from '../utils/storage-class-dropdown';
@@ -20,6 +21,15 @@ import {
   dropdownUnits,
   getAccessModeForProvisioner,
 } from './shared';
+
+const InfoToolTip = () => (
+  <Tooltip
+    position="bottom"
+    content={<div>Only filesystem volume mode is available for cephfs</div>}
+  >
+    <OutlinedQuestionCircleIcon />
+  </Tooltip>
+);
 
 const NameValueEditorComponent = (props) => (
   <AsyncComponent
@@ -135,6 +145,9 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
     setAllowedAccessModes(modes);
     setStorageClass(updatedStorageClass?.metadata?.name);
     setStorageProvisioner(provisioner);
+    if (storageProvisioner.includes(cephStorageProvisioners[1])) {
+      setVolumeMode('Filesystem');
+    }
   };
 
   const handleRequestSizeInputChange = (obj) => {
@@ -266,16 +279,22 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
         Volume Mode
       </label>
       <div className="form-group">
-        {volumeModeRadios.map((radio) => (
-          <RadioInput
-            {...radio}
-            key={radio.value}
-            onChange={handleVolumeMode}
-            inline
-            checked={radio.value === volumeMode}
-            name="volumeMode"
-          />
-        ))}
+        {storageProvisioner.includes(cephStorageProvisioners[1]) ? (
+          <>
+            Filesystem <InfoToolTip />
+          </>
+        ) : (
+          volumeModeRadios.map((radio) => (
+            <RadioInput
+              {...radio}
+              key={radio.value}
+              onChange={handleVolumeMode}
+              inline
+              checked={radio.value === volumeMode}
+              name="volumeMode"
+            />
+          ))
+        )}
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


frontend/public/components/storage/create-pvc.tsx
```diff
@@ -269,6 +269,7 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
             key={radio.value}
             onChange={handleVolumeMode}
             inline
+            disabled={storageProvisioner.includes('cephfs') && radio.value === 'Block'}
```
a2batic, line 272: Declare const in `frontend/public/components/storage/shared.ts` file for `CephFS` like we have it for RBD and then use it.

bipuladh, line 272: If it's RBD then we have to disable the 'FileSystem' too if I am not mistaken.

cloudbehl: I think rbd supports both types, @vbnrh can you check with QE.

spadgett:
I wouldn't just disable a single radio button if there's only one choice. Instead, I'd suggest a plain text, readonly value of "Filesystem," possibly with a popover explaining why you can't change it.

vbnrh:
@cloudbehl @spadgett @yuvalgalanti I have screenshot for both the approaches..... Please suggest one of the them.. I find having a readonly text with a tooltip to be inconsistent with the current UX. 

I'll update the PR if read-only text is the preferred approach

cloudbehl:
> We've used a plain text label for readonly values in several other forms (create role binding comes to mind), so I don't think it's inconsistent. I prefer this approach.
> 
> If we're going to disable the radio buttons, we should disable both "Volume Mode" radio buttons, not just "Block."

I agree we can just shift to read-only mode as we have only two options & one is disabled.

vbnrh:
@cloudbehl @spadgett thanks for your confirmations.. I've updated the PR. Please review

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -266,16 +283,22 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-          <RadioInput
-            {...radio}
-            key={radio.value}
-            onChange={handleVolumeMode}
-            inline
-            checked={radio.value === volumeMode}
-            name="volumeMode"
-          />
-        ))}
+        {storageProvisioner.includes(cephFSProvisionerSuffix) ? (
```
cloudbehl, line 286: you can use 1st index of this list.

https://github.com/openshift/console/blob/eec5efd40c79b95ebf3ba4315d5719b197840aa9/frontend/packages/console-shared/src/utils/storage-utils.ts#L6

```suggestion
        {storageProvisioner.includes(cephFSProvisionerSuffix) ? (
```

yuvalgalanti:
in access mode, you see that one of the options is disabled, how is that different from what we are doing? It's actually more clear to users seeing something is disabled due to another selection they did rather than disappearing the option.
I think we can keep it as is and add a description text under the radio buttons row @spadgett @cloudbehl @vbnrh, how does that sounds to you?

frontend/public/components/storage/create-pvc.tsx
vbnrh: done

vbnrh:
> in access mode, you see that one of the options is disabled, how is that different from what we are doing? It's actually more clear to users seeing something is disabled due to another selection they did rather than disappearing the option.
> I think we can keep it as is and add a description text under the radio buttons row @spadgett @cloudbehl @vbnrh, how does that sounds to you?

@yuvalgalanti Can you create a design for this and post it here ? It will make it clear to everyone to look for...

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -266,16 +282,22 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-            onChange={handleVolumeMode}
-            inline
-            checked={radio.value === volumeMode}
-            name="volumeMode"
-          />
-        ))}
+        {storageProvisioner.includes(cephStorageProvisioners[0]) ? (
+          <div>
+            Filesystem <InfoToolTip />
+          </div>
```
rawagner, line 288: you dont have to wrap it in div
```suggestion
          <>
            Filesystem <InfoToolTip />
          </>
```

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -21,6 +22,18 @@ import {
...
+    <span>
+      <Tooltip
+        position="bottom"
+        content={<div>Only filesystem volume mode is available for cephfs</div>}
+      >
+        <OutlinedQuestionCircleIcon />
+      </Tooltip>
+    </span>
+  );
+};
```
rawagner, line 36: No need for `span`. I also miss type and you can do without explicit return 
```suggestion
const InfoToolTip: React.FC = () => (
      <Tooltip
        position="bottom"
        content={<div>Only filesystem volume mode is available for cephfs</div>}
      >
        <OutlinedQuestionCircleIcon />
      </Tooltip>
  );
```

rawagner: I tried with `ocs-storagecluster-cephfs` (provisioner `openshift-storage.cephfs.csi.ceph.com`), but I still could choose Filesystem or Block. The code disables the Block option only for provisioner `ceph.rook.io/block`. So I wonder if the message is correct ?

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -2,8 +2,9 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
 import { Link } from 'react-router-dom';
-import { ActionGroup, Button } from '@patternfly/react-core';
-import { isCephProvisioner, isObjectSC } from '@console/shared/src/utils';
+import { ActionGroup, Button, Tooltip } from '@patternfly/react-core';
+import OutlinedQuestionCircleIcon from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
```
rawagner, line 6:
```suggestion
import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
```

rawagner (changes requested):

frontend/public/components/storage/create-pvc.tsx
vbnrh: fixed

frontend/public/components/storage/create-pvc.tsx
vbnrh: oops index should be 1 :P changed it ! could you give it another shot

frontend/public/components/storage/create-pvc.tsx
vbnrh: fixed

PR #7539, opened 2020-12-14
Bug 1903400: Migration cannot be when machine is off or in error

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1903400

**Analysis / Root cause**: 
Missing checks for isError and isInProgress

**Solution Description**: 
Added checks before enabling migration


metalice:
@rawagner , @yaacov , please review :)

PR #7541, opened 2020-12-14
Bug 1906356: Unify Clone PVC boot source flow

glekner:
When clicking customize on the VM Wizard and choosing the Clone provision source, the user can now select the PVC from the General Tab.

Screen Shot 2020-12-14 at 15 07 23


glekner:
@matthewcarleton @rawagner

matthewcarleton:
awesome! My only question is "project" vs "namespace" I think I've seen PVC namespace in other places. @rawagner WDYT?

glekner:
@matthewcarleton in the simple wizard (before clicking customize) its `Persistent Volume Claim project`

PR #7542, opened 2020-12-14
Bug 1907892: Hide edit app action for apps created using devfile flow

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5225

**Analysis/Root cause:**
Import from devfile is currently in `DevPreview` and edit app feature was not planned for `DevPreview` and hence is not available currently but `edit <app-name>` action is still shown clicking on which the edit flow page errors out.

**Solution:**
Add a temporary annotation `isFromDevfile` on devfile resources and hide `edit <app-name>` action on deployment based on if above annotation exists on it. We can remove this once we plan to add edit feature for apps created using devfile flow.

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
+++ b/frontend/packages/dev-console/src/components/import/import-submit-utils.ts
@@ -471,16 +471,8 @@ export const createDevfileResources = async (
   const {
     name,
     project: { name: namespace },
-    application: { name: applicationName },
-    git: { url: repository, ref = 'master' },
     devfile: { devfileSuggestedResources },
   } = formData;
-
-  const defaultLabels = getAppLabels({ name, applicationName });
-  const defaultAnnotations = {
-    ...getGitAnnotations(repository, ref),
-    isFromDevfile: 'true',
-  };
 
   const devfileResourceObjects: DevfileSuggestedResources = Object.keys(
     devfileSuggestedResources,
@@ -493,13 +485,12 @@ export const createDevfileResources = async (
         metadata: {
           ...resource.metadata,
           annotations: {
-            ...defaultAnnotations,
             ...resource.metadata?.annotations,
+            isFromDevfile: 'true',
           },
           name,
           namespace,
           labels: {
-            ...defaultLabels,
             ...resource.metadata?.labels,
           },
         },
--- a/frontend/packages/dev-console/src/utils/resource-label-utils.ts
+++ b/frontend/packages/dev-console/src/utils/resource-label-utils.ts
@@ -81,9 +81,16 @@ export const getPodLabels = (name: string) => {
 
 export const mergeData = (originalResource: K8sResourceKind, newResource: K8sResourceKind) => {
   const mergedData = _.merge({}, originalResource || {}, newResource);
-  mergedData.metadata.labels = newResource.metadata.labels;
+  const isDevfileResource = originalResource?.metadata?.annotations?.isFromDevfile;
+  mergedData.metadata.labels = {
+    ...newResource.metadata.labels,
+    ...(isDevfileResource ? originalResource?.metadata?.labels : {}),
+  };
   if (mergedData.metadata.annotations) {
-    mergedData.metadata.annotations = newResource.metadata.annotations;
+    mergedData.metadata.annotations = {
+      ...newResource.metadata.annotations,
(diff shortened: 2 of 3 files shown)


divyanshiGupta:
@andrewballantyne @rohitkrai03 any suggestions on the annotation name?

andrewballantyne:
@divyanshiGupta The Devfile submit utils reuses resource creation (with the backend resources filled in as "original" resources) so I'm wondering if we can just remove the annotation that causes the edit flow based on a param the Devfile passes through. Thoughts?

divyanshiGupta:
> @divyanshiGupta The Devfile submit utils reuses resource creation (with the backend resources filled in as "original" resources) so I'm wondering if we can just remove the annotation that causes the edit flow based on a param the Devfile passes through. Thoughts?

@andrewballantyne Yeah initially I thought about doing that but that annotation is used at other places also for example - to determine if the deployment was created using the web-console and based on that showing build alert when the build hasn't completed, so I think we cannot remove that. That is why thought of adding a temp annotation which can be removed once we add edit feature for apps created using devfile flow.

andrewballantyne:
> @andrewballantyne Yeah initially I thought about doing that but that annotation is used at other places also for example - to determine if the deployment was created using the web-console and based on that showing build alert when the build hasn't completed, so I think we cannot remove that. That is why thought of adding a temp annotation which can be removed once we add edit feature for apps created using devfile flow.

There is an annotation on all resources called `isFromDevfile`... perhaps that's the useful one here? It's temporary and we can just make the change in the kebab render logic I would think 🤔

divyanshiGupta:
> > @andrewballantyne Yeah initially I thought about doing that but that annotation is used at other places also for example - to determine if the deployment was created using the web-console and based on that showing build alert when the build hasn't completed, so I think we cannot remove that. That is why thought of adding a temp annotation which can be removed once we add edit feature for apps created using devfile flow.
> 
> There is an annotation on all resources called `isFromDevfile`... perhaps that's the useful one here? It's temporary and we can just make the change in the kebab render logic I would think thinking

Okay so I dont see `isFromDevfle` annotation on the deployment created using devfile flow. This is because its being added on the backend resource and not on the resource created by us hence its getting overridden during merge. Will use this annotation then. Thanks!

frontend/packages/dev-console/src/components/import/import-submit-utils.ts
```diff
@@ -490,13 +490,11 @@ export const createDevfileResources = async (
         metadata: {
           ...resource.metadata,
           annotations: {
-            ...defaultAnnotations,
```
andrewballantyne, line 493: I think we need to keep this 🤔 

If annotations are provided by the Devfile Library, we don't want to lose those. I just noticed now that I failed to follow through on Regina's work and make sure these values were in the resources created 🤦 

Can we look at merging old and new annotations and labels? Is there any concern you can think of around this?

frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -51,9 +52,10 @@ export const getGitAnnotations = (gitURL: string, gitRef?: string) => {
   };
 };
 
-export const getCommonAnnotations = () => {
+export const getCommonAnnotations = (buildStrategy?: string) => {
   return {
     'openshift.io/generated-by': 'OpenShiftWebConsole',
+    ...(buildStrategy === BuildStrategyType.Devfile && { isFromDevfile: 'true' }),
```
andrewballantyne, line 58: I don't mind this solution that much -- I just would prefer to keep the Devfile specific stuff isolated if at all possible. We shouldn't look to pollute everything with Devfile if we can avoid it. Ideally if we went with a label "disable-edit" that would work too haha... it's a bit on the nose though.

It may also just be better to avoid this edit and go with my other comment about merging annotations 🤔 

I'm open to suggestions.

andrewballantyne (changes requested):
I think you removed needed merging of annotations and labels... it just failed to be properly echo'ed in the other methods. That's my bad, I didn't follow it through when I cloned the code that Regina did. 

Unless you can think of a good reason why we don't merge annotations and labels from the previous "original" objects, I suggest we just fix that here and go with your kebab hidden fix.

andrewballantyne (approved):
/lgtm

PR #7543, opened 2020-12-14
Bug 1907518: adds rbac for eventSource, KameletBinding and removes camelSource unwanted changes

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5255

**Analysis / Root cause**: 
RBAC checks were not there for eventSources and KameletBinding so self provisioner with view only access can see sources in catalog but can't create one

**Solution Description**: 
- Handled RBAC for EventSources
- Handled RBAC for KameletBinding
- Removes unneeded code for camel Source as we moved to Kamelets

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
```diff
@@ -264,59 +254,26 @@ export const sanitizeKafkaSourceResource = (formData: EventSourceFormData): Even
...
-        provider: isKnownEventSource(kind) ? 'Red Hat' : '',
+      uid,
+      name,
+      description,
+      icon: { url: getEventSourceIcon(referenceForModel(model)), class: null },
+      type: 'EventSource',
+      provider,
+      cta: { label: t('knative-plugin~Create Event Source'), href },
+      data: {
+        sourceType: kind,
```
sahil143, line 276: It would be better to keep this logic separate from this util. if we keep it here this would affect the catalog and quick search as well. I think we should just keep the normalization logic separate in the provider itself and return models from here like we did before

frontend/packages/knative-plugin/src/components/add/EventSourcePage.tsx
```diff
@@ -1,44 +1,45 @@
...
-import { CamelKameletBindingModel, CamelKameletModel } from '../../models';
+import { CamelKameletBindingModel } from '../../models';
 
 type EventSourcePageProps = RouteComponentProps<{ ns?: string }>;
 
 const EventSourcePage: React.FC<EventSourcePageProps> = ({ match, location }) => {
   const { t } = useTranslation();
   const namespace = match.params.ns;
-  const eventSourceList = useEventSourceList(namespace);
+  const eventSourceListData = useEventSourceList(namespace);
```
sahil143, line 23: we don't need to fetch all the event sources and perform an access check for all of them. Just get the data for the needed event source using the kind from the URL.
Icon, Description can be taken from their separate utilities

frontend/packages/knative-plugin/src/hooks/useEventSourceStatus.ts
```diff
@@ -0,0 +1,92 @@
...
+  const { eventSourceModels, loaded: eventSourceModelsLoaded } = useEventSourceModels();
+  const [kamelet, kameletLoaded] = useK8sGet<K8sResourceKind>(
+    CamelKameletModel,
+    kameletName,
+    namespace,
+  );
+  const [createSourceAccess, setCreateSourceAccess] = React.useState<boolean>(true);
+  const [sourceFound, setSourceFound] = React.useState<boolean>(true);
+  const [loaded, setLoaded] = React.useState<boolean>(false);
+  const [normalizedSource, setNormalizedSource] = React.useState<EventSourceMetaData>();
```
sahil143, line 35: Not sure if we are going to need all these states. I think We can get away with using `useMemo` for these.

frontend/packages/knative-plugin/src/hooks/useEventSourceStatus.ts
```diff
@@ -0,0 +1,92 @@
...
+  const eventSourceModel =
+    isEventSource && eventSourceModels?.find((model: K8sKind) => model.kind === sourceKindProp);
+  const sourceModel =
+    isKameletSource && kamelet && kameletLoaded ? CamelKameletBindingModel : eventSourceModel;
+
+  React.useEffect(() => {
+    let sourceData = {};
+    if (sourceModel) {
+      setSourceFound(true);
+      checkAccess({
```
sahil143, line 50: we can use `useAccessReview` instead of this.

invincibleJai: yeah made use of `useAccessReview2` as needed loading state

vikram-raj (commented):
Verified all the use cases. it works as expected. 👍

frontend/packages/knative-plugin/src/utils/create-eventsources-utils.ts
```diff
@@ -264,190 +248,59 @@ export const sanitizeKafkaSourceResource = (formData: EventSourceFormData): Even
...
+      name,
+      description,
+      provider,
+      iconUrl: getEventSourceIcon(referenceForModel(eventSourceModel)),
+    };
+  }
+  return normalizedSource as EventSourceMetaData;
+};
+
+export const getEventSourceModelsWithAccess = (namespace: string, eventSourceModels: K8sKind[]) => {
```
sahil143, line 288: Why not return the map itself here, instead of pushing to accessList?
```
return eventSourceModels.map((model) => {
    const { apiGroup, plural } = model;
   return checkAccess({
                group: apiGroup,
        	resource: plural,
        	namespace,
   	        verb: 'create',
	      }).then((result) => (result.status.allowed ? model : null)),

sahil143 (approved):
/lgtm

PR #7544, opened 2020-12-14
Add Chinese to webpack config

rebeccaalpert:
I was looking at our webpack config and noticed that we don't have Chinese listed, even though we should. I updated it.


rebeccaalpert:
Closing because it looks like this has already been fixed on rebase.

PR #7545, opened 2020-12-14
Bug 1907543: Set "en" as default for Moment.js

rebeccaalpert:
We only import the Moment.js locales we currently support in OpenShift. Moment.js didn't know how to handle locales we didn't import and seemed like it was defaulting to the last imported locale (Korean) when the language was set to German, Spanish, etc. Changing which locale was the last imported locale affected what language Moment.js set itself to.

We only support English, Japanese, Chinese, and Korean timestamps right now. I set Moment to default to English if it gets a language we don't handle.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1907543.

### DIFF
--- a/frontend/public/i18n.js
+++ b/frontend/public/i18n.js
@@ -5,10 +5,10 @@ import httpBackend from 'i18next-http-backend';
 import Pseudo from 'i18next-pseudo';
 
 // Load moment.js locales (en is default and always loaded)
-import 'moment/locale/zh-cn';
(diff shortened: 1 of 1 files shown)


jerolimov:
Could verify this PR on my machine, checked absolute and relative (x hours ago) timestamps.

My chrome settings:
image

jcaianirh (approved):
lgtm

rhamilto (approved):
/lgtm

PR #7547, opened 2020-12-15
Bug 1907888: Fix pipeline list page loader

karthikjeeyar:
Fixes: https://issues.redhat.com/browse/ODC-5264

**Analysis / Root cause**: 

"Pipelines" list page shows a message "No pipelines found" before it loads the data

**Solution Description**: 

Do not render the `No pipelines found`  text before the firehose call loads the data.

Screenshots:
pipeline-loader-text

image

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
@@ -1,9 +1,9 @@
 import * as _ from 'lodash';
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { ListPageWrapper_ as ListPageWrapper } from '@console/internal/components/factory';
-import { Firehose } from '@console/internal/components/utils';
+import { EmptyBox, Firehose, LoadingBox } from '@console/internal/components/utils';
 import { Resource, getResources } from '../../../utils/pipeline-augment';
-import { PipelineModel } from '../../../models';
 import PipelineAugmentRuns, { filters } from './PipelineAugmentRuns';
 import PipelineList from './PipelineList';
 
@@ -14,13 +14,15 @@ interface PipelineAugmentRunsWrapperProps {
 }
 
 const PipelineAugmentRunsWrapper: React.FC<PipelineAugmentRunsWrapperProps> = (props) => {
+  const { t } = useTranslation();
   const pipelineData = _.get(props.pipeline, 'data', []);
-  if (pipelineData.length < 1) {
-    return (
-      <div className="cos-status-box">
-        <div className="text-center">No {PipelineModel.labelPlural} Found</div>
-      </div>
-    );
+
+  if (!props.pipeline.loaded) {
+    return <LoadingBox />;
+  }
+
+  if (pipelineData.length === 0) {
+    return <EmptyBox label={t('pipelines-plugin~Pipelines')} />;
   }
   const firehoseResources: Resource = getResources(props.pipeline.data);
   return (
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import { EmptyBox, LoadingBox } from '@console/internal/components/utils';
+import { ListPageWrapper_ as ListPageWrapper } from '@console/internal/components/factory';
+import { PipelineExampleNames, pipelineTestData } from '../../../../test-data/pipeline-data';
+import PipelineAugmentRunsWrapper from '../PipelineAugmentRunsWrapper';
+
+const mockData = pipelineTestData[PipelineExampleNames.WORKSPACE_PIPELINE];
+const { pipeline } = mockData;
+
+type PipelineAugmentRunsWrapperProps = React.ComponentProps<typeof PipelineAugmentRunsWrapper>;
+
+jest.mock('react-i18next', () => {
+  const reactI18next = require.requireActual('react-i18next');
+  return {
+    ...reactI18next,
+    useTranslation: () => ({ t: (key) => key }),
+  };
+});
+
+describe('Pipeline Augment Run Wrapper', () => {
+  let pipelineAugmentRunsWrapperProps: PipelineAugmentRunsWrapperProps;
+  let wrapper: ShallowWrapper;
+  beforeEach(() => {
+    pipelineAugmentRunsWrapperProps = {
+      pipeline: {
+        data: [pipeline],
+        loaded: false,
+      },
+    };
+    wrapper = shallow(<PipelineAugmentRunsWrapper {...pipelineAugmentRunsWrapperProps} />);
+  });
+
+  it('Should render loader if data not yet loaded', () => {
+    expect(wrapper.find(LoadingBox).exists()).toBeTruthy();
+  });
+
+  it('Should render the EmptyBox if the data is empty', () => {
+    wrapper.setProps({ pipeline: { data: [], loaded: true } });
+    expect(wrapper.find(EmptyBox).exists()).toBeTruthy();
+  });
+
+  it('Should render ListpageWrapper if the pipeline data is loaded and available', () => {
+    wrapper.setProps({ pipeline: { data: [pipeline], loaded: true } });
+    expect(wrapper.find(ListPageWrapper).exists()).toBeTruthy();
+  });
+});


andrewballantyne:
/retest
/assign

Looks good on a high level, makes sense that's why it's happening. Want to investigate the code a bit more before I approve.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
```diff
@@ -15,7 +15,10 @@ interface PipelineAugmentRunsWrapperProps {
 
 const PipelineAugmentRunsWrapper: React.FC<PipelineAugmentRunsWrapperProps> = (props) => {
   const pipelineData = _.get(props.pipeline, 'data', []);
-  if (pipelineData.length < 1) {
+  if (!props.pipeline.loaded) {
+    return null;
+  }
+  if (props.pipeline.loaded && pipelineData.length < 1) {
```
andrewballantyne, line 21:
```suggestion
  if (pipelineData.length === 0) {
```
Best be explicit and not mathematically correct 🙂 

Also, don't need to include the inverse of previous boolean statement.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+        data: [pipeline],
+        loaded: false,
+      },
+    };
+    wrapper = shallow(<PipelineAugmentRunsWrapper {...pipelineAugmentRunsWrapperProps} />);
+  });
+
+  it('Should not render if the firehose call is not yet loaded', () => {
+    expect(wrapper.html()).toBeNull();
+  });
```
andrewballantyne, line 29: We should always favour loading spinners... There is almost never a good reason to do null if the content is still being processed by a hook or firehose.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
```diff
@@ -15,7 +15,10 @@ interface PipelineAugmentRunsWrapperProps {
...
     return (
       <div className="cos-status-box">
         <div className="text-center">No {PipelineModel.labelPlural} Found</div>
```
andrewballantyne, line 24: This needs to be translated... mind adding this to your PR?

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  it('Should render No Pipelines found text if the data is empty', () => {
+    wrapper.setProps({ pipeline: { data: [], loaded: true } });
+    expect(wrapper.find('div.text-center').text()).toBe('No Pipelines Found');
```
andrewballantyne, line 33: 🤔  I am not happy with these kind of tests... they are brittle and tie themselves to the exact implementation. If we swapped the divs out for PF components, this test would fail, but effectively not useful to us in anyway. If we changed the text, that would also not be ideal.

Perhaps the test here should be you don't render the ListPage 🤔

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it('Should render firehose if the pipeline data is loaded and available', () => {
+    wrapper.setProps({ pipeline: { data: [pipeline], loaded: true } });
+    const firehoseComponent = wrapper.find(Firehose);
+
+    expect(firehoseComponent.exists()).toBeTruthy();
```
andrewballantyne, line 40: Let's not write tests against how we load data... this test serves just to bind our implementation to Firehose. Really what we want to do is not worry about how it gets the data and that it just renders the needed wrapper.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
```diff
@@ -15,7 +15,10 @@ interface PipelineAugmentRunsWrapperProps {
 
 const PipelineAugmentRunsWrapper: React.FC<PipelineAugmentRunsWrapperProps> = (props) => {
   const pipelineData = _.get(props.pipeline, 'data', []);
-  if (pipelineData.length < 1) {
+  if (!props.pipeline.loaded) {
+    return null;
```
andrewballantyne, line 19: As noted on the tests, this needs to be a loading spinner not `null`.

andrewballantyne (changes requested):
Sorry, got to this now, not sure what I missed from my first look but there are some comments on your added tests and I think I need to clarify to the team when `null` vs `loading` is needed.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
karthikjeeyar: Moved it to spinner.

frontend/public/locales/en/public.json
```diff
@@ -13,5 +13,6 @@
   "Pause": "Pause",
   "Warning": "Warning",
   "Not available": "Not available",
-  "Resource": "Resource"
+  "Resource": "Resource",
+  "No Pipelines Found": "No Pipelines Found"
```
andrewballantyne, line 17: This should be in the Pipeline package not public

karthikjeeyar: Done.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/PipelineAugmentRunsWrapper.tsx
```diff
@@ -14,11 +14,17 @@ interface PipelineAugmentRunsWrapperProps {
...
+
+  if (!props.pipeline.loaded) {
+    return <LoadingBox />;
+  }
+
+  if (pipelineData.length === 0) {
     return (
       <div className="cos-status-box">
-        <div className="text-center">No {PipelineModel.labelPlural} Found</div>
+        <div className="text-center">{t('pipelines-plugin~No Pipelines Found')}</div>
```
debsmita1, line 27: can we use the `EmptyBox` component here ?

karthikjeeyar: Problem with EmptyBox is, it uses dynamic values (label) in the translation and generates keys in `utils.json`. More over the label in our case is `Pipelines` will not be translated.

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+    wrapper = shallow(<PipelineAugmentRunsWrapper {...pipelineAugmentRunsWrapperProps} />);
+  });
+
+  it('Should render loader if data not yet loaded', () => {
+    expect(wrapper.find(LoadingBox).exists()).toBeTruthy();
+  });
+
+  it('Should not render the listPageWrapper if the data is empty', () => {
+    wrapper.setProps({ pipeline: { data: [], loaded: true } });
+    expect(wrapper.find(ListPageWrapper).exists()).toBeFalsy();
```
debsmita1, line 40: can you also check if EmptyBox is rendered

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+      pipeline: {
+        data: [pipeline],
+        loaded: false,
+      },
+    };
+    wrapper = shallow(<PipelineAugmentRunsWrapper {...pipelineAugmentRunsWrapperProps} />);
+  });
+
+  it('Should render loader if data not yet loaded', () => {
+    expect(wrapper.find(LoadingBox).exists()).toBeTruthy();
```
andrewballantyne, line 35: Nit, we should always equate booleans to their boolean value. 

```suggestion
    expect(wrapper.find(LoadingBox).exists()).toBe(true);
```

It's not _wrong_ to stay truthy (true is truthy lol), it's just a bit inaccurate on a principle basis.

andrewballantyne (approved):

frontend/packages/pipelines-plugin/src/components/pipelines/list-page/__tests__/PipelineAugmentRunsWrapper.spec.tsx
andrewballantyne: I'm indifferent if this changes before merging. We need a linter to check for `.exists()` and `.tobeTruthy()` on the same expect.

PR #7522, opened 2020-12-11
Bug 1907307: Propagate provider and support annotations

rawagner:
Screenshot from 2020-12-11 10-36-50
Screenshot from 2020-12-11 10-22-26
Screenshot from 2020-12-11 10-22-07

@matthewcarleton can you take a look ?

PR #7549, opened 2020-12-15
Bug 1907793:  Show support info in VM template details

rawagner:
depends on https://github.com/openshift/console/pull/7522


rawagner:
Follows what we have in dev perspective @matthewcarleton 
Screenshot from 2020-12-15 10-15-07


Screenshot from 2020-12-15 10-15-44

matthewcarleton:
Should it say just "full"? Where did the "full" come from in general? I don't think we had that in any designs did we?

rawagner:
We didnt, support level (Full/Community) is coming from annotation on the template - see https://issues.redhat.com/browse/CNV-7742

matthewcarleton:
ah ok - it would always say "Red Hat - full" though right? Not just "full"?

rawagner:
Changed to always show `<provider> - <support_level>` (before I had only `<support_level>` if there was just one)

frontend/packages/kubevirt-plugin/src/components/vm-templates/label.scss
```diff
@@ -0,0 +1,3 @@
+.kv-template-support__label {
```
matthewcarleton, line 1: you could write this so there's only a margin if there's more than one

yaacov:
/close

CI failing, will continue in #7644

PR #7552, opened 2020-12-15
Bug 1897830: Fix cluster creation when using localvolume

afreen23:
- use the nodeSelector field to get the associated nodes for a PV
- drops the usage of labels for pv association

Before:
Screenshot from 2020-12-16 13-18-11

After:
Screenshot from 2020-12-16 13-18-20

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/utils/install.ts
+++ b/frontend/packages/ceph-storage-plugin/src/utils/install.ts
@@ -4,13 +4,17 @@ import {
   Taint,
   StorageClassResourceKind,
   K8sResourceKind,
+  MatchExpression,
 } from '@console/internal/module/k8s';
 import {
   humanizeBinaryBytes,
   convertToBaseValue,
   humanizeCpuCores,
 } from '@console/internal/components/utils';
-import { HOSTNAME_LABEL_KEY } from '@console/local-storage-operator-plugin/src/constants';
+import {
+  HOSTNAME_LABEL_KEY,
+  LABEL_OPERATOR,
+} from '@console/local-storage-operator-plugin/src/constants';
 import { getNodeCPUCapacity, getNodeAllocatableMemory, getName } from '@console/shared';
 import { ocsTaint, NO_PROVISIONER, AVAILABLE, MINIMUM_NODES, ZONE_LABELS } from '../constants';
 import { Discoveries } from '../components/ocs-install/attached-devices/create-sc/state';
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -48,10 +52,13 @@ export const getTotalDeviceCapacity = (list: Discoveries[]): number =>
 
 export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
   const nodes = pvs.reduce((res, pv) => {
-    const nodeName = pv?.metadata?.labels?.[HOSTNAME_LABEL_KEY];
-    if (nodeName) {
-      res.add(nodeName);
-    }
+    const matchExpressions: MatchExpression[] =
+      pv?.spec?.nodeAffinity?.required?.nodeSelectorTerms?.[0]?.matchExpressions;
```
cloudbehl, line 56: what if there are multiple node selectors? I think that was the reason we relied on PV label.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -48,10 +52,13 @@ export const getTotalDeviceCapacity = (list: Discoveries[]): number =>
 
 export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
   const nodes = pvs.reduce((res, pv) => {
-    const nodeName = pv?.metadata?.labels?.[HOSTNAME_LABEL_KEY];
```
cloudbehl, line 51: can you tell in which case the HOSTNAME_LABEL_KEY will not be present if =done from localVolumeSet?

cloudbehl (commented):
Just wanted to understand why we are doing this at first?

frontend/packages/ceph-storage-plugin/src/utils/install.ts
afreen23: For LSO PVs, it will always have hostname based selector.
I confirmed with LSO team on that.

afreen23: Relying on node affinity is more reliable over label selector. The node affinity field is a mandatory one for PV and will always be present on every PV unlike labels - added as per convenience.
Thats why we had issues in supporting 4.5 OCS on OCP 4.6 where users were trying to create storage cluster via local volume CR and they are blocked, since the pVs created via local volume are not labeled by hostname

frontend/packages/ceph-storage-plugin/src/utils/install.ts
afreen23: There is not any case for local volume set.
The bug is for local volume fix  and an improvement over the current implementation where we identify nodes via labels.
See https://github.com/openshift/console/pull/7552/files#r543672246

PR #7553, opened 2020-12-15
Bug 1907863: Update quick start actions to handle concurrent state updates

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5233
**Fixes**: https://issues.redhat.com/browse/ODC-5236
**Fixes**: https://issues.redhat.com/browse/ODC-5268

**Analysis / Root cause**: While moving from local storage to use settings the quick start state was moved from redux to top level context using context provider extension. The quick start state is very complex and require concurrent update to the state for each user action, that is why there were multiple calls to `setState((state) => newState)` which resulted in each call working with the older and stale state. this resulted in none of the quick start actions working correctly.

**Solution Description**: There were two possible solutions for the above problem - 

1. Use `redux` or `useReducer` for managing state of quick start. The problem with this approach is that the `initialState` of the reducer needs to be async coming from user settings. One way to solve this was to have a separate action just to initialize the state asynchronously but not sure if this is a widely accepted pattern for doing such things.
2. Convert existing state update callbacks to mimic user actions and handle all concurrent updates together. 

Went ahead with the second approach for now as that required less changes to the current code.

Quick Start Fix

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (commented):
@rohitkrai03 can we add the sync true to AllQuickStartStates only. This will be helpful if the Quickstarts are open in two different browser tabs.

rohitkrai03:
@sahil143 Added sync true for all quick start states.

sahil143 (commented):
The back button on the quick start is not working
quickstart-back

rohitkrai03:
> The back button on the quick start is not working
> quickstart-back

@sahil143 Fixed.

sahil143:
@rohitkrai03 I'm seeing a weird mismatch in the state. Not sure if it's to do with quick start state updates or the userSettings hook itself
The state is correct on the card(completed) in both tabs but once it is opened the state changes to InProgress
quick-start-tour

rohitkrai03:
> @rohitkrai03 I'm seeing a weird mismatch in the state. Not sure if it's to do with quick start state updates or the userSettings hook itself
> The state is correct on the card(completed) in both tabs but once it is opened the state changes to InProgress
> quick-start-tour

That's actually because `Start the tour` button sets the status to `In progress` as well. I think if user starts a completed tour again, the status should be updated to `In progress`. We can confirm this behaviour with UX if needed.

sahil143:
> That's actually because Start the tour button sets the status to In progress as well. I think if user starts a completed tour again, the status should be updated to In progress. We can confirm this behaviour with UX if needed.

Ah Okay! That makes sense

sahil143 (commented):
/lgtm

verified changes locally. works as expected

PR #7556, opened 2020-12-15
Bug 1907893: Fix nav items sort algorithm to work with latest node/browser versions

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5267

**Analysis / Root cause**: 
The Nav items sort algorithm used worked fine for Node 10 but Node 12 updated sort to use timsort which breaks the nav item sort algorithm.

**Solution Description**: 
Update the sort algorithm for nav items to not use `sort` but rather a simple insertion sort based on dependencies.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/public/components/nav/navSortUtils.ts
+++ b/frontend/public/components/nav/navSortUtils.ts
@@ -1,34 +1,30 @@
-import { NavSection as PluginNavSection, NavItem, SeparatorNavItem } from '@console/plugin-sdk';
+import {
+  NavSection as PluginNavSection,
+  NavItem as PluginNavItem,
+  SeparatorNavItem,
+} from '@console/plugin-sdk';
 
-const itemDependsOnItem = (
-  s1: PluginNavSection | NavItem | SeparatorNavItem,
-  s2: PluginNavSection | NavItem | SeparatorNavItem,
-) => {
+const toArray = (val) => (val ? (Array.isArray(val) ? val : [val]) : []);
+
+type NavItem = PluginNavSection | PluginNavItem | SeparatorNavItem;
+
+const itemDependsOnItem = (s1: NavItem, s2: NavItem): boolean => {
   if (!s1.properties.insertBefore && !s1.properties.insertAfter) {
     return false;
   }
-  const before = Array.isArray(s1.properties.insertBefore)
-    ? s1.properties.insertBefore
-    : [s1.properties.insertBefore];
-  const after = Array.isArray(s1.properties.insertAfter)
-    ? s1.properties.insertAfter
-    : [s1.properties.insertAfter];
+  const before = toArray(s1.properties.insertBefore);
+  const after = toArray(s1.properties.insertAfter);
   return before.includes(s2.properties.id) || after.includes(s2.properties.id);
 };
 
-const isPositioned = (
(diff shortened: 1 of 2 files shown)


frontend/public/components/nav/navSortUtils.ts
```diff
@@ -87,18 +85,18 @@ export const getSortedNavItems = (navItems: (PluginNavSection | NavItem | Separa
 
 export const sortExtensionItems = (
   extensionItems: (PluginNavSection | NavItem | SeparatorNavItem)[],
-) => {
+): (PluginNavSection | NavItem | SeparatorNavItem)[] => {
```
jerolimov, line 88: Yeah :love_you_gesture:

frontend/public/components/nav/navSortUtils.ts
```diff
@@ -109,5 +107,39 @@ export const sortExtensionItems = (
...
+        remainingItems.push(item);
+      }
+    });
+
+    if (remainingItems.length) {
+      // If nothing changed, just add the remaining items
+      if (!sortedItem) {
+        preSorted.push(...remainingItems);
+        return;
+      }
```
jerolimov, line 135: Is this an optimization? Maybe its me, but I don't get for which case this is needed. At least the tests works fine if I remove this 5 lines. :)

jerolimov (commented):
Test works fine again on node 14. In general looks like all is fine, just added one question. Otherwise this lgtm.

frontend/public/components/nav/navSortUtils.ts
jerolimov: And if you want make me super happy I think this file could use a type def at the top:

```tsx
import { NavSection as PluginNavSection, NavItem as PluginNavItem, SeparatorNavItem } from '@console/plugin-sdk';

type NavItem = PluginNavSection | PluginNavItem | SeparatorNavItem;
```

And then add types anywhere like:

```tsx
const sortItems = (preSorted: NavItem[], itemsToSort: NavItem[]) => {
```

:smirk_cat:

frontend/public/components/nav/navSortUtils.ts
jeff-phillips-18: If there are no items left that have dependencies on other items then we add them all. This is for error cases where two items might depend on each other.

frontend/public/components/nav/navSortUtils.ts
jeff-phillips-18: I've trie to make @jerolimov super happy 😉

frontend/public/components/nav/navSortUtils.ts
jeff-phillips-18: I've added a test for such thing

jerolimov: Yeah great. Without this code we get an endloss loop. Thanks for that additional test. Works fine! :+1:

frontend/public/components/nav/navSortUtils.ts
jerolimov: Thanks :)

jerolimov (approved):
Works fine!

/lgtm

PR #7557, opened 2020-12-15
Bug 1907896:  Update PF React Topology for new item placement issue

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5256
https://issues.redhat.com/browse/ODC-4988

**Analysis / Root cause**: 
Patternfly React Topology was not running any simulation on an add node so the node was simply centered on the graph without any adjustments for existing nodes.
Patternfly React's Topology ContextMenu was allowing submenus to overflow the viewport and cause a scroll bar to appear.

**Solution Description**: 
Update the version of PF React Topology to get a fix for this issue which runs one round of simulation after adding nodes to make slight adjustments to placement without totally overwriting user node placement. Also pick up a fix to the context menu that uses the viewport rather than the window to determine placement for context menus.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -112,7 +112,7 @@
     "@patternfly/react-core": "4.75.2",
     "@patternfly/react-table": "4.19.5",
     "@patternfly/react-tokens": "4.9.16",
-    "@patternfly/react-topology": "4.6.54",
+    "@patternfly/react-topology": "4.6.93",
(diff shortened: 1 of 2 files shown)


jeff-phillips-18:
PF React Topology Fix: https://github.com/patternfly/patternfly-react/pull/5249

PR #7558, opened 2020-12-15
[WIP] fix quickstart footer

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-4619
https://issues.redhat.com/browse/ODC-4780

# Screenshot 
quick-start-panel


christianvogt:
Subtle but I dislike how the scrollbars aren't aligned to the top and right edge of the bounding container. This also cuts off the content off by the padding offset below the header and above the footer.
image

lwrigh:
Was the shadow added to the content that is partially hidden? I would also add more consistent spacing around the buttons in the footer. I would suggest 16px spacing.

PR #7562, opened 2020-12-16
Bug 1873480: Fix accessibility issue on Topology page

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5156
https://issues.redhat.com/browse/ODC-4949

**Analysis / Root cause**: 
Accessibility issue on Topology page. Missing `aria-label` attribute for `View shortcuts`, `Find by Name`, and `View Change` icon button.

**Solution Description**: 
Add the `aria-label` attribute to `View shortcuts`, `Find by Name`, and `View Change` icon button.


frontend/packages/topology/src/components/page/TopologyPageToolbar.tsx
```diff
@@ -37,6 +40,7 @@ const TopologyPageToolbar: React.FC<TopologyPageToolbarProps> = observer(
             <Button
               type="button"
               variant="link"
+              aria-label={t('topology~View shortcuts')}
```
christianvogt, line 43: Fairly certain that `aria-label` for a button containing text is unnecessary unless the `aria-label` has more descriptive text which differs from the element text.

vikram-raj: Here we may need the UX input for text.

divyanshiGupta: @vikram-raj we can change it to `View topology shortcuts` or something similar to make it more descriptive.
cc: @beaumorley

christianvogt: In most cases the button text is sufficient and since we already decided that this text 'view shortcuts' is sufficient for the normal user.

christianvogt: I still do not feel it necessary to add an aria-label here unless you can say that UX says otherwise.

vikram-raj: We have an accessibility issue with this shortcut button that is the reason I am adding `aria-label` to this.

christianvogt: @vikram-raj are you sure we have an issue? The two linked jira issues do not mention the shortcuts button.

Please share the issue.

vikram-raj: It is the issue https://issues.redhat.com/browse/ODC-4949

christianvogt: Fairly certain they actually meant the `!` icon button because it has no aria-label or inner text.
There are no accessibility violations according to a tool i use (axe) to test the `? view shortcuts` button where as the `!` icon button is missing text.

Did you test with an accessibility tool to see the violation before and after?

vikram-raj: Removed this `aria-label` .

andrewballantyne (approved):
/lgtm

PR #7565, opened 2020-12-16
Bug 1908296: Fix pipeline builder form yaml switcher validation issue

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5249

**Analysis / Root cause**: 

In pipeline builder page,  pasting the yaml file on the yaml section and switching back to form view fails to enable the create button. Formik validation is not occurring against the immediate change in form values while switch back to form view.

**Solution Description**: 

Add a force validation to get correct error state while switching to form view.

submit-button-enabled

No UI changes

**Test Setup**:

1. Goto pipeline builder page, switch to yaml section
2. Paste a valid pipeline yaml  from the clipboard
3. Switch back to form view (create button should be enabled)

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
@@ -52,9 +52,18 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
     setFieldValue,
     setStatus,
     values,
+    validateForm,
   } = props;
   const statusRef = React.useRef(status);
   statusRef.current = status;
+
+  React.useEffect(() => {
+    if (values.editorType === EditorType.Form) {
+      // Force validation against the new data that was adjusted in the YAML
+      // Formik isn't properly handling the immediate state of the form values during the cycle of the editorType
+      setTimeout(() => validateForm(), 0);
+    }
+  }, [values.editorType, validateForm]);
 
   const updateErrors: UpdateErrors = React.useCallback(
     (taskErrors) => {


divyanshiGupta:
@karthikjeeyar did you find out why the form is not getting properly validated? Is it a formik issue?
Your solution is working but it is a hack and we use `SyncedEditorField` at other places also but we dont see this issue there.

@rohitkrai03 do you an idea about it?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
```diff
@@ -52,10 +52,19 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
...
   const statusRef = React.useRef(status);
   statusRef.current = status;
 
+  React.useEffect(() => {
+    if (values.editorType === EditorType.Form) {
+      // Force validation against the new data that was adjusted in the YAML
+      // Formik isn't properly handling the immediate state of the form values during the cycle of the editorType
+      setTimeout(() => validateForm(), 0);
+    }
+  }, [values.editorType, validateForm]);
```
rohitkrai03, line 66: Why not use `useFormikValidationFix` hook here? It does a similar hack.

rohitkrai03 (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
karthikjeeyar: I tried using this hook, but unfortunately it doesn't work.  with pipeline builder form component, validation happens before the task is preselected, so the above code is to validate the form right after all the new changes are in place, thus setTimeout is used.

rohitkrai03 (approved):
/lgtm
/approve

PR #7567, opened 2020-12-16
Bug 1889213: The error message of uploading failure is not clear enough

glekner:


glekner:
Screen Shot 2020-12-15 at 23 28 17

Screen Shot 2020-12-15 at 23 28 20

@matthewcarleton @rawagner @jelkosz

matthewcarleton:
nice work! 
Should the action say Back to form (deletes DV)? We don't mention the PVC at all prior to the action so it might be a bit confusing.

glekner:
changed button text to - `(Deletes DV)` @matthewcarleton

frontend/packages/kubevirt-plugin/src/k8s/requests/cdi-upload/cdi-upload-requests.ts
```diff
@@ -65,7 +64,7 @@ const createUploadToken = async (pvcName: string, namespace: string): Promise<st
   }
 };
 
-export const createUploadPVC = async (dataVolume: V1alpha1DataVolume, t: TFunction) => {
```
yaacov, line 68: nice !

PR #7576, opened 2020-12-16
Bug 1908715: Add fix to loop back to last quick-search list item on pressing arrow up key when on top

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5281

**Analysis/Root cause:**
When on the bottom item and you press the down arrow loop backs to the top but when on the top item and you press the up arrow, it doesn't loop back to the bottom. 

**Solution:**
When topmost item is selected and arrow up key is pressed, make last item as the selectedItem.

**Gif:**
Peek 2020-12-16 21-28

/kind bug


rohitkrai03 (approved):
/lgtm
/approve

bgliwa01 (approved):
lgtm! Thanks Divyanshi!

PR #7581, opened 2020-12-16
Bug 1908425: - Create Role Binding form subject type and name are undefined when All Project is selected 

cyril-ui-developer:
/assign @rhamilto

This PR is a follow on to https://github.com/openshift/console/pull/7546. This fix addressed the issue of subject type and name are set to undefined when All Projects is selected.

Screen Shot 2020-12-16 at 1 17 31 PM
Screen Shot 2020-12-16 at 12 52 17 PM


dtaylor113 (approved):
/lgtm

rhamilto (approved):
/lgtm

PR #7583, opened 2020-12-16
Bug 1908434: Remove &apos from metal3-plugin internationalized strings

rebeccaalpert:
Removed &apos from metal3-plugin internationalized strings.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1908434.


sg00dwin (approved):
/lgtm

frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
```diff
@@ -57,7 +57,7 @@ const StartNodeMaintenanceModal = withHandlePromise<StartNodeMaintenanceModalPro
           <StackItem>
             <Trans ns="metal3-plugin">
               If the node does not exit maintenance within <strong>30 minutes</strong>, the cluster
-              will automatically rebuild the node&apos;s data using replicated copies
+              will automatically rebuild the node{"'"}s data using replicated copies
```
jtomasek, line 60: Is this necessary?
```suggestion
              will automatically rebuild the node's data using replicated copies
```
should work just fine shouldn't it?

jtomasek (changes requested):

frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
rebeccaalpert: It doesn't. You have to escape it in the Trans component or you get "react/no-unescaped-entities" errors. We have "react/no-unescaped-entities" turned off in OpenShift and it didn't respond to comments disabling the rule, so I think it's coming from the `react-i18next` library.

jtomasek: Hmm, I see, so in that case It seems to be better to keep the html escape code then (`&apos;`). It is also recommended in the plugin docs [1]. But I am fine with both solutions.

[1] https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md#prevent-invalid-characters-from-appearing-in-markup-reactno-unescaped-entities)

jtomasek (approved):
/lgtm

PR #7586, opened 2020-12-17
Bug 1908687: Use localstorage fallback when running a local bridge

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/CONSOLE-2510
https://bugzilla.redhat.com/show_bug.cgi?id=1908687

**Analysis / Root cause**: 
With the new user settings feature its hard for console developers who shares and switch between many clusters.

The settings could confuse other developers on the same cluster / user. And they are often missed when switching to another cluster.

**Solution Description**: 
* This PR introduce a new bridge cli option `--user-settings-location=configmap|localstorage`.
* It was automatically set to `localstorage` in the `oc-enviconment.sh` script. So its automatically active for all console developers which runs a local bridge.
* The flag is available in frontend as `SERVER_FLAGS.userSettingsLocation`
* The user settings hook(s) automatically check this flags and use the localStorage if the flag is set to `localstorage`.

It's possible to override this settings with cli option like `bin/bridge --user-settings-location=configmap`

**Test setup:**

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


christianvogt:
/lgtm
/approve

@jerolimov it would be good to allow the dev to maybe specify a fixed user name. While logging in as kubeadmin, the configmap name is consistent, when using consoledeveloper, the configmap name gets concatenated with the user UID which differs on each dev cluster.

christianvogt:
fyi @spadgett

cmd/bridge/main.go
```diff
@@ -122,7 +122,8 @@ func main() {
 
 	fLoadTestFactor := fs.Int("load-test-factor", 0, "DEV ONLY. The factor used to multiply k8s API list responses for load testing purposes.")
 
-	fDevCatalogCategories := fs.String("developer-catalog-categories", "", "Allow catalog categories customization. (JSON as string)")
+	fDevCatalogCategories := fs.String("developer-catalog-categories", "", "DEV ONLY. Allow catalog categories customization. (JSON as string)")
```
spadgett, line 125: Is this flag really dev-only? I thought this was used in production (set by the operator).

cmd/bridge/main.go
```diff
@@ -122,7 +122,8 @@ func main() {
...
+	fUserSettings := fs.String("user-settings", "configmap", "DEV ONLY. Define if the user settings should be stored in a shared ConfigMap or in users browser (configmap | localstorage).")
```
spadgett, line 126: We should add validation when reading the flag.

Say what the default is if not set in the description.

spadgett: nit: maybe `user-settings-location`?

cmd/bridge/main.go
jerolimov: You're right. I removed this again.

cmd/bridge/main.go
jerolimov: I'm fine with that. Renamed everything to `userSettingsLocation` or `user-settings-location`

jerolimov:
/retest

@spadgett Thanks for the review. I applied all your requests. Can you please check again.

@christianvogt @invincibleJai I also added this commit https://github.com/openshift/console/pull/7586/commits/f206ace1e0f2428776d53a7c40f9303d4325f025 to do NOT append the user uid (or kubeadmin) if this `-user-settings-location configmap` is used.

It's not beautiful because I added the `localStorage getItem/setItem` key to the method signature. But I also don't want check the SERVER_FLAGS twice. This also enables me to remove the redux dependency from `useUserSettingsLocalStorage`. Wdyt?

invincibleJai:
/lgtm

Thanks @jerolimov , verified both in cluster and off cluster seems to work as expected. can update description as well to use `bin/bridge --user-settings-location=configmap`

christianvogt:
nice
I like how my pinned nav items and other setting stick with me in local dev

/approve

/assign @spadgett

spadgett (approved):
/lgtm

PR #7589, opened 2020-12-17
Bug 1908716: passes loadError message and handled same in pods-overview

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5287

**Analysis / Root cause**: 
LoadError was not string which was getting to render 

**Solution Description**: 
Pass LoadError message which is string and not render loadError

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


rohitkrai03 (approved):
/approve

PR #7593, opened 2020-12-17
Bug 1906118: Fix polling issues in OCS feature detector

bipuladh:
- Don't perform polling for all resources.
- Lower the number of CRs we are polling.


jhadvig (approved):
/lgtm

PR #7597, opened 2020-12-18
Bug 1908997: Add more provisioners with their accessmode mapping

yapei:
Fix [Bug 1908997](https://bugzilla.redhat.com/show_bug.cgi?id=1908997) - Unsupported access mode should not be available when creating pvc by cinder-csi-driver/gcp-pd-csi-driver from web-console

@jhadvig Can you help review?


jhadvig (approved):
/lgtm

PR #7600, opened 2020-12-18
Bug 1909092: adds feature flag with channel CRD for add action and route

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4940

**Analysis / Root cause**: 
User was seeing invalid message if clicks on Channel Card and Eventing CR is not created

**Solution Description**: 
Feature flag on Add Action/Route for channel was only for Knative Eventing CRD added Channel CRD as well

**Test setup:**
1. Install Serverless operator
2. Don't create CR for Knative Eventing and Knative Serving

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/locales/en/knative-plugin.json
+++ b/frontend/packages/knative-plugin/locales/en/knative-plugin.json
@@ -117,6 +117,7 @@
   "No sink found for this resource.": "No sink found for this resource.",
   "Filter": "Filter",
   "Autoscaled to 0": "Autoscaled to 0",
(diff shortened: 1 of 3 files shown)


karthikjeeyar:
Verified on clusterbot with the changes, it works as expected 
/lgtm

vikram-raj (commented):
Adding `lgtm` back as it was dropped after rebase.

/lgtm

PR #7606, opened 2020-12-18
Bug 1909236: Fix for overlapping remove pin icon on long nav items

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5205

**Analysis / Root cause**: 
The text for the nav item is truncated when it hits the end of the nav bar area.

**Solution Description**: 
On hover, truncate the text for the nav item when it hits the area where the remove icon is shown.

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/bug

### DIFF
--- a/frontend/public/components/nav/_perspective-nav.scss
+++ b/frontend/public/components/nav/_perspective-nav.scss
@@ -27,6 +27,7 @@
     white-space: nowrap;
 
     &:hover, &:focus {
+      padding-right: 30px;
       .oc-nav-pinned-item__unpin-button {
         .oc-nav-pinned-item__icon {
           opacity: 1;


frontend/public/components/utils/dropdown.jsx
```diff
@@ -150,6 +151,13 @@ class DropDownRow extends React.PureComponent {
             e.stopPropagation();
             onBookmark(itemKey, !isBookmarked);
           }}
+          aria-label={
+            _.isString(content)
+              ? isBookmarked
+                ? t('dropdown~Remove bookmark {{content}}', { content })
+                : t('dropdown~Add bookmark {{content}}', { content })
+              : null
+          }
```
jerolimov, line 160: Like the `isString` check. I directly thought "but what happen if the content is a react node."

I think you could still use the aria-label `"Remove bookmark "` if the content is not a string. This would help more than no aria-label in this case. So maybe:

```suggestion
          aria-label={
            isBookmarked
              ? t('dropdown~Remove bookmark {{content}}', { _.isString(content) ? content : '' })
              : t('dropdown~Add bookmark {{content}}', { _.isString(content) ? content : '' })
          }
```

jerolimov:
@jeff-phillips-18 Overlapping works fine and would receive a lgtm from me. But the second commit is already part of #7608. I liked this change as well but have a small request in the other PR. If you can drop this commit we can add some badges to this PR here. :smile:

frontend/public/components/utils/dropdown.jsx
jeff-phillips-18: not sure how I added this commit... sorry

jeff-phillips-18:
> @jeff-phillips-18 Overlapping works fine and would receive a lgtm from me. But the other commit is already part of #7608. I liked this change as well but have a small request in the other PR. If you can drop this commit we can add some badges to this PR here. 😄

Sorry, not sure how I added that commit but it is gone now.

jerolimov (approved):
Works fine. Thanks for cleaning up the history. Both PRs lgtm.

/lgtm

rohitkrai03 (approved):
/approve

PR #7608, opened 2020-12-18
Bug 1909276: Fix for a11y errors in project dropdown

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5129

**Analysis / Root cause**: 
There are no aria attributes on the bookmark or favorite link in the dropdown.

**Solution Description**: 
Add aria-labels to the favorite and bookmark links

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/bug

### DIFF
--- a/frontend/public/components/utils/dropdown.jsx
+++ b/frontend/public/components/utils/dropdown.jsx
@@ -105,7 +105,7 @@ class DropdownMixin extends React.PureComponent {
   }
 }
 
-class DropDownRow extends React.PureComponent {
+class DropDownRowWithTranslation extends React.PureComponent {
   render() {
     const {
       itemKey,
@@ -120,9 +120,11 @@ class DropDownRow extends React.PureComponent {
       favoriteKey,
       canFavorite,
       onFavorite,
+      t,
     } = this.props;
 
     let prefix;
+    const contentString = _.isString(content) ? content : '';
 
     if (!autocompleteFilter && !onBookmark) {
       //use pf4 markup if not using the autocomplete dropdown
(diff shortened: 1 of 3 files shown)


dtaylor113 (approved):
/lgtm

frontend/public/components/utils/dropdown.jsx
```diff
@@ -150,6 +151,13 @@ class DropDownRow extends React.PureComponent {
             e.stopPropagation();
             onBookmark(itemKey, !isBookmarked);
           }}
+          aria-label={
+            _.isString(content)
+              ? isBookmarked
+                ? t('dropdown~Remove bookmark {{content}}', { content })
+                : t('dropdown~Add bookmark {{content}}', { content })
+              : null
+          }
```
jerolimov, line 160: Like the `isString` check. I directly thought "but what happen if the content is a react node."

I think you could still use the aria-label `"Remove bookmark "` if the content is not a string. This would help more than no aria-label in this case. So maybe:

```suggestion
          aria-label={
            isBookmarked
              ? t('dropdown~Remove bookmark {{content}}', { _.isString(content) ? content : '' })
              : t('dropdown~Add bookmark {{content}}', { _.isString(content) ? content : '' })
          }
```

And maybe its also possible to extract this to the top of the render function so that we have this code only once and use `{{ content: contentAsString }}` or so. Wdyt?

jerolimov (changes requested):
Added a small tiny improvement idea. Let me know what you think. :smile:

jeff-phillips-18:
> Added a small tiny improvement idea. Let me know what you think. 😄

Like it! Done.

jerolimov (approved):
Tested it locally, works fine :+1:

/lgtm

rohitkrai03 (approved):
/approve

PR #7612, opened 2020-12-21
Bug 1908545: Clone modal fails to open

yaacov:
`nameError` can be `null` in clone modal


yaacov:
forced push to skip the buggy CI :bug:

PR #7618, opened 2020-12-21
Bug 1909790: fix pipeline yaml edit flow

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5270

**Analysis / Root cause:** 
The new pipeline obj has no resourceVersion property

**Solution Description:**
Merged the new pipeline and existing pipeline objects

**GIF:**
pipeline-yaml-edit

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
@@ -126,9 +126,13 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
   };
 
   const sanitizeToYaml = () =>
-    safeJSToYAML(convertBuilderFormToPipeline(values.formData, namespace), 'yamlData', {
-      skipInvalid: true,
-    });
+    safeJSToYAML(
+      convertBuilderFormToPipeline(values.formData, namespace, existingPipeline),
+      'yamlData',
+      {
+        skipInvalid: true,
+      },
+    );
 
   return (
     <>


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderPage.tsx
```diff
@@ -49,9 +50,7 @@ const PipelineBuilderPage: React.FC<PipelineBuilderPageProps> = (props) => {
     if (values.editorType === EditorType.YAML) {
       try {
         pipeline = safeLoad(values.yamlData);
-        if (!pipeline.metadata?.namespace) {
-          pipeline.metadata.namespace = ns;
```
rottencandy, line 53: I think we should keep the `namespace` check, or it would cause [ODC-5251](https://issues.redhat.com/browse/ODC-5251) to appear again.

rottencandy:
Not able to delete an individual task or a param from the pipeline:
tmp_pl

rottencandy (approved):
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderPage.tsx
```diff
@@ -49,6 +50,13 @@ const PipelineBuilderPage: React.FC<PipelineBuilderPageProps> = (props) => {
     if (values.editorType === EditorType.YAML) {
       try {
         pipeline = safeLoad(values.yamlData);
+        pipeline =
```
rohitkrai03, line 53: Why can't we do something similar to how edit-yaml component handles updates?

rohitkrai03 (changes requested):

karthikjeeyar:
@debsmita1
While editing the pipeline in the yaml view extra properties like `resourceVersion`, `managedFields` etc are not shown in the pipeline yaml view.  IMO we should show all the fields in the yaml view without hiding any fields.

The problem is here is in `PipelineBuilderForm.tsx`,  `sanitizeToYaml` function converts the form to yaml using `convertBuilderFormToPipeline` (missing the third argument `existingPipeline`). This should fix the issue and also show all the missing  fields in the yaml editor. Please revert the current fix and add the third argument to make it work.

karthikjeeyar:
Thanks @debsmita1 
/lgtm

andrewballantyne (approved):

PR #7619, opened 2020-12-21
Bug 1909681: Remove extension guards for feature detectors in Ceph Plugin

bipuladh:


cloudbehl:
can we remove the use of `CEPH_FLAG` as it's can be replaced with `OCS_FLAG` ?

cloudbehl:
will deal with ceph_flag in future PR while refatoring. Thanks

/lgtm

PR #7621, opened 2020-12-21
Bug 1909792: Internationalize details page empty states

rebeccaalpert:
No owner and no labels weren't internationalized.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1909792.

### DIFF
--- a/frontend/public/components/utils/label-list.tsx
+++ b/frontend/public/components/utils/label-list.tsx
@@ -2,6 +2,9 @@ import * as React from 'react';
 import { Link } from 'react-router-dom';
 import * as classNames from 'classnames';
 import * as _ from 'lodash-es';
+/* eslint-disable import/named */
+import { withTranslation, WithTranslation } from 'react-i18next';
+/* eslint-enable import/named */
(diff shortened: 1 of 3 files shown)


jcaianirh (approved):
lgtm

frontend/public/components/utils/label-list.tsx
```diff
@@ -19,21 +22,21 @@ export const Label: React.SFC<LabelProps> = ({ kind, name, value, expand }) => {
   );
 };
 
-export class LabelList extends React.Component<LabelListProps> {
+class TranslatedLabelList extends React.Component<LabelListProps> {
```
jhadvig, line 25: nit: noticed that for the components we are using `*WithTranslation` suffix.
```suggestion
class LabelListWithTranslation extends React.Component<LabelListProps> {
```

But wondering if we should rather just use the original component name as we do when use `connect`, since `withTranslation` is also just a HOC.

Would be good to align on the naming, to avoid any confusions.

jhadvig (approved):
One nit that can be solved as a follow up.

/lgtm
/approve

rebeccaalpert:
Rebased.

PR #7622, opened 2020-12-21
Bug 1909815: Internationalize Admin perspective in switcher

rebeccaalpert:
The perspective switcher only had the Developer perspective label internationalized. I added the Admin perspective label.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1909815.

### DIFF
--- a/frontend/packages/console-app/src/plugin.tsx
+++ b/frontend/packages/console-app/src/plugin.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import { CogsIcon } from '@patternfly/react-icons';
+import i18next from 'i18next';
 import { FLAGS } from '@console/shared/src/constants';
 import {
   Plugin,
@@ -81,7 +82,7 @@ const plugin: Plugin<ConsumedExtensions> = [
     type: 'Perspective',
     properties: {
       id: 'admin',
-      name: 'Administrator',
+      name: i18next.t('console-app~Administrator'),
       icon: <CogsIcon />,
       default: true,
       getLandingPageURL: (flags) =>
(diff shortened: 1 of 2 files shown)


frontend/packages/console-app/src/plugin.tsx
```diff
@@ -81,7 +82,8 @@ const plugin: Plugin<ConsumedExtensions> = [
     type: 'Perspective',
     properties: {
       id: 'admin',
-      name: 'Administrator',
+      // t('console-app~Administrator')
```
jcaianirh, line 85: hi @rebeccaalpert do we need this comment?

jcaianirh (changes requested):
looks good minor comment

frontend/packages/console-app/src/plugin.tsx
rebeccaalpert: All set @jcaianirh.

jcaianirh (approved):
lgtm thanks @rebeccaalpert

jhadvig (approved):
/lgtm
/approve

rebeccaalpert:
Rebased.

frontend/packages/console-app/src/plugin.tsx
```diff
@@ -81,7 +82,7 @@ const plugin: Plugin<ConsumedExtensions> = [
     type: 'Perspective',
     properties: {
       id: 'admin',
-      name: 'Administrator',
+      name: i18next.t('console-app~Administrator'),
```
spadgett, line 85: @rebeccaalpert Sorry, I know this is merged, but we shouldn't be using `i18next.t` here. It should be...

```
      // t('console-app~Administrator')
      name: '%console-app~Administrator%',
```

spadgett: Disregard, I see this has since been fixed.

PR #7623, opened 2020-12-21
Disable operator-install-* Cypress tests

rhamilto:
operator-install-* Cypress test are failing in OLM as it depends on catalog source and operators that aren't installed in CI.  The tests need to be updated to use a mock catalog source and an operator that exists in said catalog source so as not to depend on a catalog source and operator that may not be available.  https://github.com/openshift/console/blob/f96e815b633cea26b520536b636cf031d131a3e6/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/global-installmode.scenario.ts#L42-L53 is what we used in the proceeding Protractor test it replaced.

Bug to fix test:  https://bugzilla.redhat.com/show_bug.cgi?id=1909836

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
@@ -4,7 +4,7 @@ import { detailsPage } from '../../../integration-tests-cypress/views/details-pa
 
 const operatorName = 'Portworx Essentials';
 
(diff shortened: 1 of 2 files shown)


rhamilto:
cc:  @awgreene

dtaylor113 (approved):
/lgtm

rhamilto:
/hold

need to comment out operator-install-single-namespace for the same reason.

rhamilto:
/hold cancel

Updated to disable both operator-install tests.

dtaylor113 (commented):
/lgtm

dtaylor113 (commented):
/lgtm

PR #7627, opened 2020-12-22
Bug 1910263: Adjust wizard to make existing tests to be able to run

gouyang:
This is only focus on make the existing tests to be able to run, the simple flow is not covered.


yaacov:
@gouyang  hi, kubevirt-plugin CI is currently broken (https://github.com/openshift/kubernetes/pull/496), to pass it you need to push into the PR so the robot will catch a new change, and will remove the broken CI from the list of tests.

gouyang:
@yaacov Thanks, pushed again.

gouyang:
@yaacov Looks kubevirt-plugin tests is not here anymore. Could you or ask someone review the PR, I don't know who are available during the Holiday.

PR #7304, opened 2020-11-23
[CONSOLE-2494] Update Dockerfile.product for nodejs v14.15.0

yselkowitz:
**Fixes**: 
https://issues.redhat.com/browse/CONSOLE-2494

PR #7631, opened 2020-12-22
[WIP] Increase max-old-space-size value size

jhadvig:
Test is increasing `max-old-space-size` will solve the issue with `ci/prow/analyze` failing job 🤞 

Also this might be an issue associated with [updating to nodejs14](https://github.com/openshift/console/pull/7304) or memory leaks in some of our static plugins.

CC`ing @dtaylor113 @rhamilto @christianvogt


jhadvig:
increase to `5120` failed. Going for `6144`

jhadvig:
increase to `6144` failed. Going for `8192`

jhadvig:
Closing since even doubling the size did not fix the issue.

PR #7635, opened 2020-12-22
[release-4.6] Bug 1910066:  fix yaml editor from remounting

rhamilto:


jhadvig (approved):
/lgtm

rhamilto:
Patch manager, this bug has a customer case associated with it.  Can it be prioritized for cherry-pick?

PR #7639, opened 2020-12-23
Fix Installation Tests for Ceph

bipuladh:
/assign @cloudbehl 
- Enable labeling of the namespace in OLM Page
 - Click Next to adjust for Configure Page
- Support situations where Operators Panel is already open

### DIFF
--- a/frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
+++ b/frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
@@ -26,6 +26,7 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
       cy.log('Subscribe to OCS Operator');
       cy.byLegacyTestID('operator-install-btn').click({ force: true });
       cy.byTestID('Operator recommended namespace:-radio-input').should('be.checked');
+      cy.byTestID('enable-monitoring').click();
       cy.byTestID('install-operator').click();
       cy.byTestID('success-icon', { timeout: 180000 }).should('be.visible');
       cy.exec('oc get project openshift-storage -o json').then((res) => {
@@ -49,9 +50,9 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
 
       // Make changes to this once we add annotation
       cy.log(`Install OCS in ${mode} Mode`);
-      commonFlows.navigateToOCS();
+      commonFlows.navigateToOCS(true);
       cy.byLegacyTestID('horizontal-link-Storage Cluster').click();
-      cy.byTestID('yaml-create').click();
+      cy.byTestID('item-create').click();
 
       cy.log(`Select ${mode}`);
       cy.byTestID('Internal-radio-input').should('be.checked');
@@ -73,6 +74,7 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
         cy.log('Enabling Encryption');
         cy.byTestID('encryption-checkbox').click();
       }
+      wizard.next();
 
       // Final Step
       wizard.create();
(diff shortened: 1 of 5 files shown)


afreen23:
We should specify the issue we are fixing in the commit message.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -26,6 +26,7 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
       cy.log('Subscribe to OCS Operator');
       cy.byLegacyTestID('operator-install-btn').click({ force: true });
       cy.byTestID('Operator recommended namespace:-radio-input').should('be.checked');
+      cy.byTestID('enable-monitoring').click();
```
cloudbehl, line 29: is it to support 4.6?

cloudbehl: can you put a comment, if yes.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
```diff
@@ -27,10 +26,6 @@ describe('OCS Operator Expansion of Storage Class Test', () => {
     cy.visit('/');
   });
 
-  afterEach(() => {
-    checkErrors();
-  });
-
```
cloudbehl, line 33: put a comment.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
bipuladh: yes. We need to remove it for 4.7

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
bipuladh: To be added once we fix i18n issues.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
bipuladh: it's required all the time, but we can remove it once 4.7 comes.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/add-capacity.spec.ts
cloudbehl: can you just comment on the code instead of removing it so we don't miss out?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -26,6 +26,7 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
...
       cy.byTestID('install-operator').click();
       cy.byTestID('success-icon', { timeout: 180000 }).should('be.visible');
       cy.exec('oc get project openshift-storage -o json').then((res) => {
```
GowthamShanmugam, line 32: Up to this part, we can move to some global integration-tests-cypress folder, So anyone can reuse this installation logic by passing the operator name and namespace as param.

What I am suggesting here is, The operator install flow is a pre-requisite for OCS cluster creation flow. Instead of naming it as "install", we can name it something like prepate_storage_cluster or create_ocs_cluster. And prepare_ocs_cluster flow can the call install flow externally. For easier understanding separate these two install and create logic in different functions. (Or) Add "install" as  cypress global namespace from global cypress IT and Add "prepare_ocs_cluster" as cypress global namespace in ceph-storage-plugin cypress IT

PR #7643, opened 2020-12-23
Bug 1909821: Fix flexible scaling enablement

afreen23:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1909821

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/storage-and-nodes-step.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/storage-and-nodes-step.tsx
@@ -28,6 +28,7 @@ import {
 import {
   getNodeInfo,
   shouldDeployAsMinimal,
+  isFlexibleScaling,
   filterSCWithNoProv,
   getAssociatedNodes,
   nodesWithoutTaints,
@@ -53,8 +54,7 @@ const validate = (
   enableFlexibleScaling: boolean,
 ): ValidationType[] => {
   const validations = [];
-  if (enableFlexibleScaling) {
-    //  TODO: add check for arbiter
+  if (!stretchClusterChecked && enableFlexibleScaling) {
     validations.push(ValidationType.BAREMETAL_FLEXIBLE_SCALING);
   }
   if (enableMinimal) {
@@ -84,7 +84,6 @@ export const StorageAndNodes: React.FC<StorageAndNodesProps> = ({ state, dispatc
   } = state;
 
   let scNodeNames: string[] = []; // names of the nodes, backing the storage of selected storage class
-  const scNodeNamesLength = scNodeNames.length;
   const { cpu, memory, zones } = getNodeInfo(nodes);
   const scName: string = state.storageClassName;
   const validations: ValidationType[] = validate(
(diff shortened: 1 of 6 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard-steps/select-capacity-and-nodes-step.tsx
```diff
@@ -66,7 +66,7 @@ export const SelectCapacityAndNodes: React.FC<SelectCapacityAndNodesProps> = ({
   }, [cpu, dispatch, memory, nodesCount]);
 
   React.useEffect(() => {
-    const isFlexibleScaling = nodesCount && zonesCount < 3;
+    const isFlexibleScaling = !!(nodesCount && zonesCount < 3);
```
cloudbehl, line 69: shouldn't this be `!!(nodesCount>=3 && zonesCount < 3)` otherwise even one node is selected flexible scaling will get enabled.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/storage-and-nodes-step.tsx
```diff
@@ -109,9 +107,9 @@ export const StorageAndNodes: React.FC<StorageAndNodesProps> = ({ state, dispatc
   }, [cpu, dispatch, memory, nodesCount]);
 
   React.useEffect(() => {
-    const isFlexibleScaling: boolean = scNodeNamesLength && zonesCount < 3;
+    const isFlexibleScaling: boolean = !!(nodesCount && zonesCount < 3);
```
cloudbehl, line 110: shouldn't this be !!(nodesCount>=3 && zonesCount < 3) otherwise even one node is selected flexible scaling will get enabled.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/wizard-pages/storage-and-nodes-step.tsx
```diff
@@ -109,9 +107,9 @@ export const StorageAndNodes: React.FC<StorageAndNodesProps> = ({ state, dispatc
   }, [cpu, dispatch, memory, nodesCount]);
 
   React.useEffect(() => {
-    const isFlexibleScaling: boolean = scNodeNamesLength && zonesCount < 3;
+    const isFlexibleScaling: boolean = !!(nodesCount >= MINIMUM_NODES && zonesCount < 3);
```
cloudbehl, line 110: Sorry I missed one more thing.

if arbiter is enabled the nodesCount is 3 and zones is 2, flexible sizing will get in place which should not happen. we need to check for if arbiter is enabled then flexible should not be enabled.

Also if you can move this logic to a single place so we don't need to change both always.

afreen23: Okay arbiter is not for internal mode, so we can instead guard on arbiter itself.
We should not mix two diff logics. flexible scaling should stay intact.
I will make that change.

cloudbehl: for flexible scaling, we have single logic for internal & internal-attached.
attachedDevices
`!!(nodesCount >= MINIMUM_NODES && zonesCount < 3)` 
internal
`!!(nodesCount >= MINIMUM_NODES && zonesCount < 3)`

we can place them somewhere in utils so we don't miss out other if there is a change in one.

afreen23: Okay I guess the ask was for arbiter.

cloudbehl: Thanks, @afreen23 looks much better now. :)

PR #7646, opened 2020-12-23
Bug 1766287: Fix use of hello-openshift imagestream

yselkowitz:
The new imagestream was missing a lookupPolicy setting which prevented
it from being a drop-in replacement of the image on Docker Hub.  With
that fixed in the Samples operator, only the annotations are needed from
commit eee9a0cee2f766935be7fa9ed487ce015fbaeb64.

This is dependent on https://github.com/openshift/cluster-samples-operator/pull/350

/hold
/assign @cyril-ui-developer

### DIFF
--- a/frontend/public/models/yaml-templates.ts
+++ b/frontend/public/models/yaml-templates.ts
@@ -8,9 +8,6 @@ import { YAMLTemplate } from '@console/plugin-sdk';
  * Sample YAML manifests for some of the statically-defined Kubernetes models.
  */
 
-const sampleContainerImage = 'registry.redhat.io/openshift4/ose-hello-openshift-rhel8';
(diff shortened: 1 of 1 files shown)


yselkowitz:
When testing, the key is to check that hello-openshift is pulled from the payload instead of from docker.io.

yselkowitz:
> /igtm

typo?

yselkowitz:
Is there a reason the analyze test keeps failing?

yselkowitz:
/hold
Could someone please explain the consistent failures in the analyze test?

pweil-:
> Could someone please explain the consistent failures in the analyze test?

We began seeing this memory issue just before the break and bumped the memory limit but that doesn't appear to have worked.  @spadgett opened https://github.com/openshift/console/pull/7702 yesterday as another solution.  Hopefully that does the trick!

PR #7649, opened 2020-12-23
Bug 1910409: Add externalized strings for i18n for notification drawer

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1910409

Externalize notification drawer strings.


rebeccaalpert (approved):
Looks good to me. I'm not an actual approver though.

frontend/public/components/notification-drawer.tsx
```diff
@@ -51,36 +53,43 @@ import { PrometheusEndpoint } from './graphs/helpers';
...
+        <Link to="/monitoring/alerts" onClick={drawerToggle}>
+          {t('notification-drawer~View all alerts')}
+        </Link>
+      </EmptyStateSecondaryActions>
+    </EmptyState>
+  );
+};
 
-export const getAlertActions = (actionsExtensions: AlertAction[]) => {
+export const getAlertActions = (actionsExtensions: AlertAction[], t) => {
```
spadgett, line 90: Don't pass `t` as a parameter. Use the global i18next instance instead.

frontend/packages/patternfly/src/components/notification-drawer/notification-entry.tsx
```diff
@@ -19,6 +20,21 @@ export enum NotificationTypes {
   update = 'update',
 }
 
+export const notificationTypeString = (type: NotificationTypes, t: TFunction): string => {
```
spadgett, line 23: Don't pass `t` as a parameter. Use the global i18next instance instead.

spadgett: No need to export if this is only used in the same file.

frontend/packages/patternfly/src/components/notification-drawer/notification-entry.tsx
```diff
@@ -59,43 +75,50 @@ const NotificationEntry: React.FC<NotificationEntryProps> = ({
...
+    >
+      <div className="pf-c-notification-drawer__list-item-header">
+        <span className="pf-c-notification-drawer__list-item-header-icon">
+          <NotificationIcon type={type} />
+        </span>
+        <h4 className="pf-c-notification-drawer__list-item-header-title">
+          <span className="pf-screen-reader">
+            {t('notification-drawer~{{type}} notification:', {
+              type: notificationTypeString(type, t),
+            })}
```
spadgett, line 103: It would be better to make the entire string one message like `Warning notification:`. This is easier for the translators to localize.

frontend/public/components/notification-drawer.tsx
```diff
@@ -98,12 +107,13 @@ const getAlertNotificationEntries = (
   toggleNotificationDrawer: () => void,
   alertActionExtensions: AlertAction[],
   isCritical: boolean,
+  t: TFunction,
```
spadgett, line 110: Don't pass `t` as an argument

frontend/public/components/notification-drawer.tsx
```diff
@@ -229,7 +241,10 @@ export const ConnectedNotificationDrawer_: React.FC<ConnectedNotificationDrawerP
       poll(`${prometheusBaseURL}/${PrometheusEndpoint.RULES}`, 'notificationAlerts', getAlerts);
     } else {
       store.dispatch(
-        UIActions.monitoringErrored('notificationAlerts', new Error('prometheusBaseURL not set')),
+        UIActions.monitoringErrored(
+          'notificationAlerts',
+          new Error(t('notification-drawer~prometheusBaseURL not set')),
```
spadgett, line 246: Is this message actually shown in the UI? If not, we don't need to localize.

spadgett (changes requested):

frontend/public/components/notification-drawer.tsx
jcaianirh: @spadgett this is returned as loadError from the UI store which is displayed in the component AlertErrorState when an error occurs and no alerts are loaded.

jcaianirh:
@spadgett made requested changes thanks for the review

frontend/packages/patternfly/src/components/notification-drawer/notification-entry.tsx
```diff
@@ -19,6 +19,21 @@ export enum NotificationTypes {
   update = 'update',
 }
 
+const notificationTypeString = (type: NotificationTypes): string => {
```
spadgett, line 22: You should declare this method inside the component since it's only used in one place. That way you can use the `t` from `useTranslation` instead of the global `i18next` instance.

jcaianirh: sounds good @spadgett

spadgett (approved):
lgtm, please squash

jcaianirh:
@spadget squashed

/hold for approvals

/assign @ahardin-rh @yapei @sferich888

yapei:
spin up a cluster and tested notification drawer functions, such as
Screen Shot 2021-01-11 at 2 32 45 PM

no issues found

spadgett:
/unassign @sferich888 
/assign @reestr 

Strictly speaking, this shouldn't need approvals as it's a bug fix. I don't mind getting sign off, though, and it's good that QE has tested.

jcaianirh:
@spadgett I think this needs approvals only because this is part of the i18n story, but never got addressed, so its external to the jira story as a bug fix, so I think it should go through the same approval process.

spadgett:
Let's go ahead and remove the hold as this is a bug fix.

/hold cancel

jcaianirh:
@spadgett could you add an lgtm on this one

spadgett (approved):
/lgtm

spadget:
Please can you remove me from this thread?

I was only included in the first place because someone wrote the wrong
username.

spadgett:
> Please can you remove me from this thread? I was only included in the first place because someone wrote the wrong username.

I have no way to remove you (sorry), but GitHub has an Unsubscribe button you can use.

spadgett:
@jcaianirh Needs rebase :/

jcaianirh:
@spadgett rebased, thanks

spadgett (approved):
/lgtm

PR #7654, opened 2020-12-24
Bug 1784298: Query Browser: Fix "reduced resolution" showing unnecessarily

kyoto:
The way that `samples` was being updated sometimes caused a non-optimal
value to be used and could also cause the "Displaying with reduced
resolution due to large dataset" message to be displayed. This fixes the
issue by always setting `samples` to the preferred value whenever the
graph's timespan changes.


spadgett (approved):
/lgtm

spadgett:
The analyze failure should be fixed by #7702

PR #7470, opened 2020-12-08
Bug 1911280: Use context directory when selecting Builder images

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4776

**Analysis / Root cause**: 
Builder images aren't detected if context directory is not at the root.

**Solution Description**: 
Use `CONTEXT_DIR` to fetch repository files when checking for supported builder images.

For GitHub: Replace the [getTree](https://octokit.github.io/rest.js/v18#git-get-tree) with [getContents](https://octokit.github.io/rest.js/v18#repos-get-content) which supports a `path` param
For Bitbucket: Pass `path` param in [bitbucket API](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/src/%7Bcommit%7D/%7Bpath%7D)
For GitLab: Pass `path` param in [GitLab's repositories API](https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/repositories.md).

**Screen shots / Gifs for design review**: 
tmp

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

PR #7655, opened 2020-12-24
Bug 1910753: Use context directory to find devfile

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5297

**Analysis / Root cause**: 
When using the UI to select a directory path for the git, it fails to take this into account when we go searching for the Devfile.

**Solution Description**: 
Use context directory and determine the correct path of the devfile.

**NOTE:** Depends on #7470, only consider ef9ca7220293c71ca6b22ccfa46c6746ac1fea27

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/git-service/src/services/github-service.ts
```diff
@@ -105,7 +112,7 @@ export class GithubService extends BaseService {
       const resp = await this.client.repos.getContents({
         owner: this.metadata.owner,
         repo: this.metadata.repoName,
-        path: 'Dockerfile',
+        path: `${this.metadata.contextDir}/Dockerfile`,
```
jerolimov, line 115: I'm sure this works in almost all cases you tested. But because contextDir is defined as optional, I would like to request to check if contextDir is defined. So that we don't get an unexpected `undefined/Dockerfile` in some edge-cases. Similar you have done in `gitlab-service.ts`.

To archive this its maybe a good idea to create two helper functions `isFilePresent(filename: string)` and `getFileContent(filename: string)` which can do all these repeated worked in this file where `isDevfilePresent` ends as a redirect to just 

```tsx
  isDevfilePresent = () => isFilePresent('devfile.yaml')
  getDevfileContent = () => getFileContent('devfile.yaml')
```

rottencandy: The `getRepoMetadata` function will make sure that `contextDir` will always be available, at least as an empty string, so IMO checking for that isn't needed. What do you think?
There are even test cases where `contextDir` isn't provided.

Updated with helper functions `isFilePresent` & `getFileContent`.

rottencandy:
cc: @rohitkrai03

rohitkrai03 (approved):
/lgtm
/approve

PR #7660, opened 2020-12-28
Bug 1907989: dataVolumeTemplates can be empty

yaacov:
dataVolumeTemplates can be empty


yaacov:

PR #7661, opened 2020-12-28
Bug 1911213: VM with default flavor should not show a pending changes warning

yaacov:
When creating a VM (using Kubevirt API or just with a simple yaml file), I see a "pending change" warning


yaacov:

PR #7662, opened 2020-12-28
Bug 1910870: Translate create vm action label

yaacov:
translate action label in template details view

Screenshot:
screenshot-localhost_9000-2020 12 28-12_56_26


yaacov:

PR #7671, opened 2020-12-29
Bug 1908180: Add source for template is stucking in preparing pvc

glekner:
- Fixed suggested Volume, Access modes not updating upon SC change
- Showing alerts with documentation
- If modes are known, non-default modes will have the label `Not recommended`

Screen Shot 2020-12-29 at 12 56 08

--------------------

Screen Shot 2020-12-29 at 12 56 00


glekner:
- changed not recommended text to include storage class name 
`Not recommended for ${storageClassName} storage class`

yaacov:
/hold
for translations

PR #7677, opened 2020-12-29
Bug 1911474: Delete Devfile Application Group without any errors

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5298

**Analysis / Root cause:** 
before deleting the webhook secret, it is not being checked if the secret exists.

**Solution Description:**
listing the available secrets in the namespace, and added a check to verify if the associated secret exists

**GIF:**
delete-app

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/topology/src/utils/application-utils.ts
+++ b/frontend/packages/topology/src/utils/application-utils.ts
@@ -136,31 +136,31 @@ const safeKill = async (model: K8sKind, obj: K8sResourceKind) => {
   return null;
 };
 
-const deleteWebhooks = (
+const deleteWebhooks = async (
   resource: K8sResourceKind,
   buildConfigs: K8sResourceKind[],
   isKnativeResource: boolean,
 ) => {
   const deploymentsAnnotations = resource.metadata?.annotations ?? {};
   const gitType = detectGitType(deploymentsAnnotations['app.openshift.io/vcs-uri']);
+  const secretList = await k8sList(SecretModel, {
+    ns: resource.metadata.namespace,
+  });
   return buildConfigs?.reduce((requests, bc) => {
     const triggers = bc.spec?.triggers ?? [];
     const reqs = triggers.reduce((a, t) => {
-      let obj: K8sResourceKind;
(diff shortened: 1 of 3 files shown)


frontend/packages/topology/src/utils/application-utils.ts
```diff
@@ -136,29 +136,36 @@ const safeKill = async (model: K8sKind, obj: K8sResourceKind) => {
...
-              webhookTypeObj.secretReference?.name ??
-              `${resource.metadata.name}-${webhookType}-webhook-secret`,
-            namespace: resource.metadata.namespace,
-          },
-        };
+        const secretName =
+          webhookTypeObj.secretReference?.name ??
+          `${resource.metadata.name}-${webhookType}-webhook-secret`;
+        const secretResource = secretList.find((secret) => secret.metadata.name === secretName);
+        if (!_.isEmpty(secretResource)) {
```
divyanshiGupta, line 160:
```suggestion
        if (secretResource) {
```

frontend/packages/topology/src/utils/application-utils.ts
```diff
@@ -136,29 +136,36 @@ const safeKill = async (model: K8sKind, obj: K8sResourceKind) => {
...
+          obj = {
+            ...resource,
+            metadata: {
+              name: secretName,
+              namespace: resource.metadata.namespace,
+            },
+          };
+        }
       }
       return obj ? [...a, safeKill(SecretModel, obj)] : a;
```
divyanshiGupta, line 170: Also since here we are now getting the list of the secrets in the ns and finding if the secret we want to delete exists and then deleting it, why not use `secretResource` here rather than creating another `obj`?

debsmita1:
fixing the failed tc

debsmita1:
> fixing the failed tc

fixed

divyanshiGupta:
Verified locally, works fine for me.
/lgtm

rohitkrai03 (approved):
/approve

PR #7679, opened 2020-12-30
Bug 1911418: default storage class is not used in rootdisk

glekner:


frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/initial-state/vm-settings-tab-initial-state.ts
```diff
@@ -77,6 +77,7 @@ export const getInitialVmSettings = (data: CommonData): VMSettings => {
       isHidden: hiddenByProviderOrCloneCommonBaseDiskImage,
     },
     [VMSettingsField.CLONE_PVC_NAME]: {},
+    [VMSettingsField.DEFAULT_STORAGE_CLASS]: {},
```
suomiy, line 80: this would be better tied with the whole wizard, not just vm settings. Either directly in CommonData or as a CommonData reference

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/storage-tab-state-update.ts
```diff
@@ -200,9 +201,46 @@ export const internalStorageDiskBusUpdater = ({
...
+  const storageClassName = iGetVmSettingValue(state, id, VMSettingsField.DEFAULT_STORAGE_CLASS);
+
+  if (storageClassName) {
+    const iProvisionSourceStorage = iGetProvisionSourceStorage(state, id);
+    const provisionSourceStorage: VMWizardStorage =
+      iProvisionSourceStorage && iProvisionSourceStorage.toJSON();
+
+    if (
+      provisionSourceStorage &&
+      !getDataVolumeStorageClassName(provisionSourceStorage?.dataVolume)
```
suomiy, line 222: this might be too forceful. What about the storages with no SC?

suomiy (commented):
@glekner can you take a look?

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/storage-tab-state-update.ts
glekner: @suomiy 
this basically means update the sc to the default one if the storage doesn't have any sc, whats wrong?
if we'll remove this line the sc will always reset to the default one even if the user chooses a different sc.

suomiy: There might be also an option of not having the storage class for creating the templates too. Since the default SC might change when the template gets used.

suomiy: edit: ah sorry for the first point - in  that case there will not be any SC to select.

So what is left is to allow selecting empty SC in disk modal

suomiy: https://github.com/openshift/console/pull/7715

PR #7681, opened 2020-12-30
Bug 1908180: Add source for template is stucking in preparing pvc

glekner:


yaacov:
/hold

need to run `yarn i18n` again, other PRs changed the `.json` file and now this PR can't pass tests.

PR #7684, opened 2020-12-31
Bug 1911443: Remove optional email from ssh key validation

yaacov:
`ssh-keygen` tool generate a public key that include the users email at the end.
but it looks like the rfc4253 ssh-rsa format does not require the email.

This PR remove the validation check that the email exist.


yaacov:

PR #7685, opened 2020-12-31
Bug 1878953: Fix RBAC with upload form and golden images

glekner:


frontend/public/components/utils/rbac.tsx
```diff
@@ -124,6 +124,55 @@ export const useAccessReview2 = (
   return [isAllowed, loading];
 };
 
+export const useMultipleAccessReviews = (
+  multipleResourceAttributes: AccessReviewResourceAttributes[],
+  impersonate?,
```
rawagner, line 129:
```suggestion
  impersonate?: boolean,
```

rawagner (changes requested):

PR #7694, opened 2021-01-04
Bug 1910501: Use absolute paths when resource creation is cancelled

bipuladh:


frontend/packages/ceph-storage-plugin/src/components/independent-mode/install.tsx
```diff
@@ -142,9 +142,7 @@ const CreateExternalCluster = withHandlePromise((props: CreateExternalClusterPro
     );
   };
 
-  const onCancel = () => {
-    history.goBack();
-  };
+  const onCancel = () => history.push(`/k8s/ns/${ns}/clusterserviceversions/${appName}/`);
```
afreen23, line 145: The correct path should be one with the`groupVersionKind` instead of `clusterserviceversions`.
Any reason for `clusterserviceversions` ?
Additionally `resourcePathFromModel` would work here.

afreen23 (changes requested):
Same for all

frontend/packages/ceph-storage-plugin/src/components/independent-mode/install.tsx
bipuladh: @afreen23 why do you prefer GVK over plural, we use plural all over our plugin declaration file.

afreen23: I have observed in console that operator pages are following the:
`k8s/ns/openshift-storage/operators.coreos.com~v1alpha1~ClusterServiceVersion/ocs-operator.v4.6.0`
And resourcePathFromModel will redirect to the operator page since you have the operator name -> appName and ns.
Something like this:
`resourcePathFromModel(ClusterServiceVersionModel, appName, ns)`

bipuladh: cool

PR #7699, opened 2021-01-04
Bug 1909836: Re-enable and update operator-install-* tests to add catalog-source

rhamilto:
…alog-source

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
@@ -1,12 +1,22 @@
 import { checkErrors } from '../../../integration-tests-cypress/support';
 import { modal } from '../../../integration-tests-cypress/views/modal';
 import { detailsPage } from '../../../integration-tests-cypress/views/details-page';
+import { createCatalogSource, deleteCatalogSource } from '../views/catalog-source.view';
 
 const operatorName = 'Portworx Essentials';
+const catalogSourceName = 'console-e2e';
+const operatorID = 'portworx-essentials-console-e2e-openshift-marketplace';
+const operatorRow = 'Portworx Essentials';
+const operatorPkgName = 'portworx-essentials';
+const operatorInstallFormURL = `/operatorhub/subscribe?pkg=${operatorPkgName}&catalog=${catalogSourceName}&catalogNamespace=openshift-marketplace&targetNamespace=`;
+const operatorInstance = 'StorageCluster';
+const openshiftOperatorsNS = 'openshift-operators';
+const operandLink = 'portworx';
 
-xdescribe(`Interacting with a global install mode Operator (${operatorName})`, () => {
+describe(`Interacting with a global install mode Operator (${operatorName})`, () => {
   before(() => {
     cy.login();
+    createCatalogSource(operatorName, catalogSourceName);
   });
 
   afterEach(() => {
@@ -14,26 +24,16 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
   });
 
   after(() => {
+    cy.log('navigate to OperatorHub > Sources');
+    deleteCatalogSource(catalogSourceName);
     cy.logout();
   });
 
-  const operatorPkgName = 'portworx-essentials';
-  const operatorID = 'portworx-essentials-community-operators-openshift-marketplace';
-  const operatorRow = 'Portworx Essentials';
-  const operatorInstallFormURL = `/operatorhub/subscribe?pkg=${operatorPkgName}&catalog=community-operators&catalogNamespace=openshift-marketplace&targetNamespace=`;
-  const operatorInstance = 'StorageCluster';
-  const openshiftOperatorsNS = 'openshift-operators';
-  const operandLink = 'portworx';
-
   it(`displays subscription creation form for ${operatorName}`, () => {
-    cy.log('navigate to the operator install form from OperatorHub');
+    cy.log('navigate to the Operator install form from OperatorHub');
     cy.visit('/operatorhub/all-namespaces');
     cy.byTestID('search-operatorhub').type(operatorName);
     cy.byTestID(operatorID).click();
-    cy.log('dismiss Show Community Operator modal');
-    modal.shouldBeOpened();
-    modal.submit();
-    modal.shouldBeClosed();
     cy.log('go to the install form');
     cy.byLegacyTestID('operator-install-btn').click({ force: true });
     cy.url().should('include', operatorInstallFormURL);
@@ -42,7 +42,7 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
   it(`creates the global install mode ClusterServiceVersion for ${operatorName}`, () => {
     cy.visit(operatorInstallFormURL);
     cy.byTestID('install-operator').click();
-    cy.log('verify operator began installation');
+    cy.log('verify Operator began installation');
     cy.byTestID('view-installed-operators-btn').should(
       'contain',
       `View Installed Operators in namespace ${openshiftOperatorsNS}`,
@@ -50,7 +50,7 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
     cy.log('view the ClusterServiceVersion list page');
     cy.byTestID('view-installed-operators-btn').click();
     cy.log(`verify the ClusterServiceVersion row for ${operatorRow} exists`);
-    cy.byTestOperatorRow(operatorRow).should('exist');
+    cy.byTestOperatorRow(operatorRow, { timeout: 60000 }).should('exist');
   });
 
   it(`displays details about ${operatorName} ClusterServiceVersion on the "Details" tab`, () => {
@@ -104,15 +104,11 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
     );
   });
 
-  it(`uninstalls the operator`, () => {
-    cy.log('navigate to the operator uninstall modal in OperatorHub');
+  it(`uninstalls the Operator`, () => {
+    cy.log('navigate to the Operator uninstall modal in OperatorHub');
     cy.visit('/operatorhub/all-namespaces');
     cy.byTestID('search-operatorhub').type(operatorName);
     cy.byTestID(operatorID).click();
-    cy.log('dismiss Show Community Operator modal');
-    modal.shouldBeOpened();
-    modal.submit();
-    modal.shouldBeClosed();
     cy.log('uninstall the operator');
     cy.byLegacyTestID('operator-uninstall-btn').click({ force: true });
     cy.url().should(
@@ -123,7 +119,7 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
     modal.modalTitleShouldContain('Uninstall Operator?');
     modal.submit(true);
     modal.shouldBeClosed();
-    cy.log('verify the operator is not installed');
+    cy.log('verify the Operator is not installed');
     cy.get('.loading-skeleton--table').should('not.be.visible');
     cy.byTestOperatorRow(operatorRow).should('not.exist');
   });
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
@@ -1,12 +1,21 @@
 import { checkErrors, testName } from '../../../integration-tests-cypress/support';
 import { modal } from '../../../integration-tests-cypress/views/modal';
+import { createCatalogSource, deleteCatalogSource } from '../views/catalog-source.view';
 
-const operatorName = '3scale Operator';
+const operatorName = 'Couchbase Operator';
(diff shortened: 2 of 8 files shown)


frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -1,12 +1,47 @@
...
+      publisher: 'Red Hat, Inc',
+    },
+  };
+  const catalogNamespace = 'openshift-marketplace';
 
-xdescribe(`Interacting with a global install mode Operator (${operatorName})`, () => {
   before(() => {
     cy.login();
+    cy.exec(`echo '${JSON.stringify(catalogSource)}' | kubectl create -n ${catalogNamespace} -f -`);
+    // await new Promise((resolve) =>
```
rhamilto, line 27: Not sure how to best check to see if the `packagemanifest` has been created or not using Cypress.  This is how we did it in Protractor.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -17,23 +18,42 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
...
+    cy.byTestID('catalog-source-name').type(catalogSourceName);
+    cy.byTestID('catalog-source-image').type(
+      'quay.io/operator-framework/upstream-community-operators@sha256:9243fc9cfdbe3df502772b92abdb0b4794adb4898833fb0a8a5c2f34f6e7ff4e',
+    );
+    cy.byTestID('save-changes').click();
+    cy.log('verify the catalogsource is READY');
+    cy.visit('/k8s/cluster/config.openshift.io~v1~OperatorHub/cluster/sources');
+    cy.byLegacyTestID(catalogSourceName).should('exist');
+    cy.byTestID(`${catalogSourceName}-status`).should('contain', 'READY');
+    // cy.wait(5000);
```
rhamilto, line 45: I thought I could avoid using the wait by navigating to the packagemanifest list and confirming it exists (lines 46-49), but even that happens too quickly.  @dtaylor113, any better ideas on how to wait for the packagemanifests to be ready?

dtaylor113: Hi @rhamilto, can you do some `oc cli` [checking](https://github.com/openshift/console/blob/66a08e0c120137138888ca879cd7cd7bd70b1b75/frontend/packages/integration-tests-cypress/support/resources.ts#L33) to see if packagemanifests exists?   Can you [wait for a promise](https://docs.cypress.io/api/utilities/promise.html#Waiting-for-Promises)?  If not, you might have to do the `cy.wait`.  After the update to Cypress 6, we can [wait on a request](https://docs.cypress.io/api/commands/intercept.html#Waiting-on-a-request), but not sure that would help in this case?

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -127,4 +143,16 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
     cy.get('.loading-skeleton--table').should('not.be.visible');
     cy.byTestOperatorRow(operatorRow).should('not.exist');
   });
+
+  it(`deletes the catalogsource`, () => {
```
dtaylor113, line 147: Hi @rhamilto, I know this is a pain but Cypress best practices calls for each `it()` to be independant of each other.  In this case the `create catalogsource` should probably go in a `before()`, and the `delete catalogsource` should go in an `after()`.  -thanks

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -17,23 +18,42 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
...
+    cy.byTestID('save-changes').click();
+    cy.log('verify the catalogsource is READY');
+    cy.visit('/k8s/cluster/config.openshift.io~v1~OperatorHub/cluster/sources');
+    cy.byLegacyTestID(catalogSourceName).should('exist');
+    cy.byTestID(`${catalogSourceName}-status`).should('contain', 'READY');
+    // cy.wait(5000);
+    cy.log('verify the packagemanifest has been created');
+    cy.visit('/k8s/ns/openshift-marketplace/packages.operators.coreos.com~v1~PackageManifest');
+    listPage.filter.byName(operatorPkgName);
+    listPage.rows.shouldExist(operatorName);
```
dtaylor113, line 49: So is this where it fails?  Times out after 30 seconds (the default timeout)?  If so, you might want to add a longer timeout to `listPage.rows.shouldExist` (90 seconds) like we did for [listPage.rows.shouldNotExist](https://github.com/openshift/console/blob/6fc9b6a34e3d69de07f8288ca8ea5e6d3cf253e3/frontend/packages/integration-tests-cypress/views/list-page.ts#L65).

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
rhamilto: Thanks, @dtaylor113 .  I was seeing some weirdness where both the `before` *and* `after` were running before any of the `it`s.  It appears I have to stop and restart Cypress to resolve.  

Since the `CatalogSource` creation and deletion is duplicated between these two tests, I want to move it to a shared file, but I'm not sure where that should be.  Any advice?

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
rhamilto: If I load the list of `PackageManifests` before all the `PackageManifests` are created, they never will all load (I guess we're not watching them here?).  I put the wait of 60 seconds in after creation of the `CatalogSource` and everything works as expected.  I realize this isn't ideal, but I'm trying to just get something that works at the moment.

If we had a `CatalogSource` with just the two `Operators` we needed for the tests, that would certainly help, but I'm not sure we want to do that since that would require the creation of another image that someone would have to maintain.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
dtaylor113: Probably need to create new folder/file under `frontend/packages/operator-lifecycle-manager/integration-tests-cypress/views/catalog.ts`

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
dtaylor113: Hmm...ok, it might be flaky under slower CI, something to watch for.
If your interested :-)  https://clearmeasure.com/polling-a-page-in-cypress/

rhamilto: > If I load the list of `PackageManifests` before all the `PackageManifests` are created, they never will all load (I guess we're not watching them here?).

@spadgett, any idea why `/k8s/ns/openshift-marketplace/packages.operators.coreos.com~v1~PackageManifest` is not dynamically updating?  Seems like a bug?

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -104,15 +130,11 @@ xdescribe(`Interacting with a global install mode Operator (${operatorName})`, (
     );
   });
 
-  it(`uninstalls the operator`, () => {
-    cy.log('navigate to the operator uninstall modal in OperatorHub');
+  it(`uninstalls the Operator`, () => {
```
dtaylor113, line 133: Hi @rhamilto, we are running into the Cypress anti-pattern of [Having-tests-rely-on-the-state-of-previous-tests](https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests).  Here you have an `it` to create the operator, another `it` to test it's details, and a final `it` to uninstall the operator.  While this seems very logical, it is frowned upon by the Cypress gods :-)  All `its` should be able to run independently; you should be able to specify a `its.only` for any test and it should be able to run.  
To resolve I think these three tests should be in a single `it` test with alot of `cy.log` statements.  
From Cypress best practices about this:

"This ... is ideal because now we are resetting the state between each test and ensuring nothing in previous tests leaks into subsequent ones.
We’re also paving the way to make it less complicated to write multiple tests against the “default” state of the form. That way each test stays lean but each can be run independently and pass."

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
```diff
@@ -93,19 +119,15 @@ xdescribe(`Interacting with a single namespace install mode Operator (${operator
   it(`displays details about ${operatorName} ${operatorInstance} instance on the "Details" tab`, () => {
     cy.log(`navigate to the "Details" tab`);
     cy.byTestOperandLink(operandLink).click();
-    cy.byTestSectionHeading('APIManagerBackup Overview').should('exist');
+    cy.byTestSectionHeading('Couchbase Cluster Overview').should('exist');
   });
 
-  it(`uninstalls the operator from ${testName} namespace`, () => {
-    cy.log('navigate to the operator uninstall modal in OperatorHub');
+  it(`uninstalls the Operator from ${testName} namespace`, () => {
```
dtaylor113, line 125: Same comment about running into the Cypress anti-pattern of [Having-tests-rely-on-the-state-of-previous-tests](https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests).

dtaylor113 (changes requested):
Tests look good to me.  Only issue is the Cypress anti-pattern of [Having-tests-rely-on-the-state-of-previous-tests](https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests).

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
```diff
@@ -13,6 +13,28 @@ const operatorInstance = 'StorageCluster';
 const openshiftOperatorsNS = 'openshift-operators';
 const operandLink = 'portworx';
 
+const verifyPackageManifest = (opName: string, csName: string, depth: number = 0) => {
```
rhamilto, line 16: @dtaylor113, I made an attempt at adding polling per https://github.com/openshift/console/pull/7699#discussion_r552131837.  What do you think?

If this is good, I will move this and the rest of `CatalogSource` install and uninstall out to a separate file so it can be shared between the two tests.

dtaylor113: :+1: I like it!

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
rhamilto: Implemented the polling technique @dtaylor113 mentioned.

rhamilto:
@dtaylor113, @spadgett PTAL.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts
rhamilto: I'm not opposed to this, but I was trying to maintain focus on getting the tests re-enabled.  This would need to be one `it` in order to comply with the best practice since the first `it` installs the operator.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/views/catalog-source.view.ts
```diff
@@ -0,0 +1,57 @@
...
+  cy.log('verify the PackageManifest has been created');
+
+  return cy
+    .visit(
+      `/k8s/ns/openshift-marketplace/packages.operators.coreos.com~v1~PackageManifest?packagemanifest-name=${opName}`,
+    )
+    .get(`[data-test-rows="resource-row"]`)
+    .should('exist')
+    .then(() => {
+      // the opName can exist in multiple CatalogSources, so check for the csName
```
spadgett, line 16: We could use the new "Operators" tab on the CatalogSource added in #7430. I'd think it would simplify the logic here and add more test coverage for that feature.

rhamilto: Good suggestion. @spadgett.  I've implemented it.  Please take another look.

spadgett (approved):
👍 
/lgtm

PR #7701, opened 2021-01-04
Bug 1912558: Show pending status for taskRuns that haven't started

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5222

**Analysis / Root cause**: 
TaskRun list and detail page shows that a TaskRun is "Running" if it has not yet started with the reason "ExceededNodeResources".

**Solution Description**: 
Show "Pending" status for TaskRuns that have the reason "ExceededNodeResources" in the Succeeded condition.

tmp
tmp0

**Test setup:**
- Create a new namespace
- Add LimitRange and Pipeline
- Start as many PipelineRuns you need to exceed your memory limits

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
@@ -5,7 +5,7 @@ import { referenceForModel } from '@console/internal/module/k8s';
 import { TaskRunKind, getModelReferenceFromTaskKind } from '../../../utils/pipeline-augment';
 import { TaskRunModel, PipelineModel } from '../../../models';
 import { tableColumnClasses } from './taskruns-table';
-import { pipelineRunFilterReducer as taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
+import { taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
 import { TektonResourceLabel } from '../../pipelines/const';
 import TaskRunStatus from '../status/TaskRunStatus';
 
--- a/frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
+++ b/frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
@@ -1,23 +1,40 @@
 import * as _ from 'lodash';
 
+// Converts the PipelineRun (and TaskRun) condition status into a human readable string.
+// See also tkn cli implementation at https://github.com/tektoncd/cli/blob/release-v0.15.0/pkg/formatted/k8s.go#L54-L83
 export const pipelineRunStatus = (pipelineRun): string => {
   const conditions = _.get(pipelineRun, ['status', 'conditions'], []);
-  const isCancelled = conditions.find((c) =>
-    ['PipelineRunCancelled', 'TaskRunCancelled'].some((cancel) => cancel === c.reason),
-  );
-  if (isCancelled) {
-    return 'Cancelled';
-  }
   if (conditions.length === 0) return null;
 
-  const condition = conditions.find((c) => c.type === 'Succeeded');
-  return !condition || !condition.status
-    ? null
-    : condition.status === 'True'
-    ? 'Succeeded'
-    : condition.status === 'False'
-    ? 'Failed'
-    : 'Running';
+  const succeedCondition = conditions.find((c) => c.type === 'Succeeded');
+  if (!succeedCondition || !succeedCondition.status) {
+    return null;
+  }
+  const status =
+    succeedCondition.status === 'True'
+      ? 'Succeeded'
+      : succeedCondition.status === 'False'
+      ? 'Failed'
+      : 'Running';
+
+  if (succeedCondition.reason && succeedCondition.reason !== status) {
+    switch (succeedCondition.reason) {
+      case 'PipelineRunCancelled':
+      case 'TaskRunCancelled':
+      case 'Cancelled':
+        return 'Cancelled';
+      case 'PipelineRunStopping':
+      case 'TaskRunStopping':
+        return 'Failed';
+      case 'CreateContainerConfigError':
+      case 'ExceededNodeResources':
+      case 'ExceededResourceQuota':
+        return 'Pending';
+      default:
+        return status;
+    }
+  }
+  return status;
 };
 
 export const pipelineFilterReducer = (pipeline): string => {
--- a/frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-filter-reducer.spec.ts
+++ b/frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-filter-reducer.spec.ts
@@ -32,8 +32,38 @@ const mockPipelineRuns = [
     },
   },
   { status: { conditions: [{ status: 'Unknown', type: 'Succeeded' }] } },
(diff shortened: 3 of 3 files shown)


frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -61,5 +61,11 @@ export const pipelineResourceTypeFilter = (filters, pipelineResource): boolean =
 
 export const taskRunFilterReducer = (taskRun): string => {
   const status = pipelineRunStatus(taskRun);
+  if (status === 'Running') {
+    const condition = taskRun.status?.conditions?.find((c) => c.type === 'Succeeded');
+    if (condition.reason === 'ExceededNodeResources') {
+      return 'Pending';
+    }
+  }
   return status || '-';
```
jerolimov, line 70: That's not part of your change, but thats exactly the reason I would prefer to keep TaskRun and PipelineRuns separated. But for now this change is "technical fine".

But there are more reasons why a TaskRun is Pending... (see https://github.com/tektoncd/pipeline/blob/66096db2c12125693c39b32bb1cbfeff6cfdb644/pkg/pod/status.go#L284-L303).

So we need to check the other both states here as well (`CreateContainerConfigError` and `Pending`) or check for status == `Unknown` and reason != `Pending`. Maybe its safer to define an array of "taskRunPendingReasons"? Wdyt?

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -61,5 +61,11 @@ export const pipelineResourceTypeFilter = (filters, pipelineResource): boolean =
 
 export const taskRunFilterReducer = (taskRun): string => {
   const status = pipelineRunStatus(taskRun);
+  if (status === 'Running') {
+    const condition = taskRun.status?.conditions?.find((c) => c.type === 'Succeeded');
+    if (condition.reason === 'ExceededNodeResources') {
```
jerolimov, line 66: If taskRun.status could be empty (see optional check one line above), and the conditions could not include a type === 'Successeded', you must check here again if the condition is defined.

jerolimov: There are also more reasons why a TaskRun is pending. Maybe you can update the `pipelineRunStatus` instead. I now found the source code of the `tkn` cli which converts the "Succeeded" condition into a "human readable text".

Since the tkn cli do this for PipelineRun and TaskRun in a similar way, I think we could do this also for PipelineRuns. Otherwise I liked that you tried to ensure with this local change to not break any PipelineRun related check.

See https://github.com/tektoncd/cli/blob/master/pkg/formatted/k8s.go#L55-L82

Wdyt? :)

jerolimov (changes requested):
I think we should align this with the tkn cli and add more cases then just the one reported case (ExceededResourceQuota).

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
rottencandy: Thanks, I Initially avoided modifying the PipelineRun check because I wasn't sure if PipelineRun and TaskRun could have the same reasons. Updated now to handle all the cases.

frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
```diff
@@ -5,7 +5,7 @@ import { referenceForModel } from '@console/internal/module/k8s';
 import { TaskRunKind, getModelReferenceFromTaskKind } from '../../../utils/pipeline-augment';
 import { TaskRunModel, PipelineModel } from '../../../models';
 import { tableColumnClasses } from './taskruns-table';
-import { pipelineRunFilterReducer as taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
+import { taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
```
jerolimov, line 8: 😳 👍 Good catch

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -2,22 +2,39 @@ import * as _ from 'lodash';
...
+      : succeedCondition.status === 'False'
+      ? 'Failed'
+      : 'Running';
+
+  if (succeedCondition.reason && succeedCondition.reason !== status) {
+    switch (succeedCondition.reason) {
+      case 'PipelineRunStopping':
+      case 'TaskRunStopping':
+        return 'Failed';
+      case 'ContainerConfigError':
```
jerolimov, line 29: In go code this error is names

```suggestion
      case 'CreateContainerConfigError':
```

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -2,22 +2,39 @@ import * as _ from 'lodash';
 
 export const pipelineRunStatus = (pipelineRun): string => {
   const conditions = _.get(pipelineRun, ['status', 'conditions'], []);
+  if (conditions.length === 0) return null;
```
jerolimov, line 5: :+1:

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -2,22 +2,39 @@ import * as _ from 'lodash';
...
   const isCancelled = conditions.find((c) =>
-    ['PipelineRunCancelled', 'TaskRunCancelled'].some((cancel) => cancel === c.reason),
+    ['PipelineRunCancelled', 'TaskRunCancelled', 'Cancelled'].some((cancel) => cancel === c.reason),
   );
   if (isCancelled) {
     return 'Cancelled';
   }
```
jerolimov, line 11: Hmmm. I know its saver to keep the existing code. But what do you think if we move this (similar to the go code) also in the switch below?

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
```diff
@@ -2,22 +2,39 @@ import * as _ from 'lodash';
 
 export const pipelineRunStatus = (pipelineRun): string => {
```
jerolimov, line 3: Can you add an documentation link here? Maybe like this:

```suggestion
// Converts the PipelineRun (and TaskRun) condition status into a human readable string.
// See also tkn cli implementation at https://github.com/tektoncd/cli/blob/release-v0.15.0/pkg/formatted/k8s.go#L54-L83
export const pipelineRunStatus = (pipelineRun): string => {
```

jerolimov (approved):
Works fine, thanks for the cleanup :+1: 

Tested it with canceled, running, pending and successfully PipelineRuns and TaskRuns.

/lgtm

frontend/packages/pipelines-plugin/src/utils/pipeline-filter-reducer.ts
andrewballantyne: Probably could have used `/** ... */` for an official JSDoc of the method... but that's fine.

andrewballantyne (approved):
/lgtm

PR #7702, opened 2021-01-04
Bug 1912115: Fix analyze command

spadgett:
Make sure `NODE_OPTIONS` is used when calling `webpack-bundle-analyzer`.

See https://stackoverflow.com/questions/10856129/setting-an-environment-variable-before-a-command-in-bash-is-not-working-for-the


spadgett:
I opened #7709 to address the flake in the previous run.

PR #7707, opened 2021-01-05
Bug 1913306: quick-start-footer

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-4619,
https://issues.redhat.com/browse/ODC-4780/'
Fixes shadow retainment on QuickStart change
Fixes shadow update on Quickstart Load
Fixes shadow update on browser resize
Markdown area scrollbar bug -> https://issues.redhat.com/browse/ODC-5338

# Fix
**Issue**
1. Unnecessary scrollbar appears in Quick start panel - This is because some headings are two lines long and some are one line long and the height computation is made by assuming one line header.
2. No shadow for overlapping content and footer.

**Solution**
1. Converted panel into flex items
2. Added shadow to footer top onScroll

# ScreenShot
Shadow update on Quickstart change and browser resize
footer1
footer2

# Tests
none changed

# Browser conformance
Chrome

### DIFF
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
@@ -6,9 +6,15 @@ import { QuickStartContext, QuickStartContextValues } from './utils/quick-start-
 
 type QuickStartControllerProps = {
   quickStart: QuickStart;
+  footerClass: string;
+  contentRef: React.Ref<HTMLDivElement>;
 };
 
-const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart }) => {
+const QuickStartController: React.FC<QuickStartControllerProps> = ({
+  quickStart,
+  contentRef,
+  footerClass,
+}) => {
   const {
     metadata: { name },
     spec: { tasks = [] },
@@ -65,6 +71,7 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart 
         onTaskSelect={handleTaskSelect}
         onTaskReview={handleTaskStatusChange}
         onQuickStartChange={handleQuickStartChange}
+        ref={contentRef}
       />
       <QuickStartFooter
         status={status}
@@ -72,6 +79,7 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart 
         totalTasks={totalTasks}
         onNext={handleNext}
         onBack={handleBack}
+        footerClass={footerClass}
       />
     </>
   );
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.scss
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.scss
@@ -10,6 +10,10 @@
     background: inherit;
     z-index: var(--pf-global--ZIndex--xs);
   }
+  &__body {
+    display: flex;
+    flex-direction: column;
+  }
   &__title {
     display: flex;
     align-items: center;
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
@@ -10,7 +10,7 @@ import {
   Title,
 } from '@patternfly/react-core';
 import { AsyncComponent } from '@console/internal/components/utils';
-import { useScrollDirection, ScrollDirection } from '@console/shared/';
+import { useScrollShadows, Shadows } from '@console/shared';
 import { QuickStart } from './utils/quick-start-types';
 import './QuickStartPanelContent.scss';
 
@@ -27,18 +27,23 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   handleClose,
   activeQuickStartID,
 }) => {
-  const [scrollDirection, handleScrollCallback] = useScrollDirection();
   const { t } = useTranslation();
+  const [contentRef, setContentRef] = React.useState<HTMLDivElement>();
+  const shadows = useScrollShadows(contentRef);
+
   const quickStart = quickStarts.find((qs) => qs.metadata.name === activeQuickStartID);
 
-  const headerClasses = classNames('co-quick-start-panel-content-head', {
-    'pf-u-box-shadow-sm-bottom':
-      scrollDirection && scrollDirection !== ScrollDirection.scrolledToTop,
+  const headerClasses = classNames({
+    'pf-u-box-shadow-sm-bottom': shadows === Shadows.top || shadows === Shadows.both,
+  });
+
+  const footerClass = classNames({
+    'pf-u-box-shadow-sm-top': shadows === Shadows.bottom || shadows === Shadows.both,
   });
 
   return quickStart ? (
-    <DrawerPanelContent className="co-quick-start-panel-content" onScroll={handleScrollCallback}>
-      <div className={headerClasses}>
+    <DrawerPanelContent className="co-quick-start-panel-content">
+      <div className={`co-quick-start-panel-content-head ${headerClasses}`}>
         <DrawerHead>
           <div className="co-quick-start-panel-content__title">
             <Title
@@ -59,10 +64,12 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
           </DrawerActions>
         </DrawerHead>
       </div>
-      <DrawerPanelBody>
+      <DrawerPanelBody hasNoPadding className="co-quick-start-panel-content__body">
         <AsyncComponent
           loader={() => import('./QuickStartController').then((c) => c.default)}
           quickStart={quickStart}
+          footerClass={footerClass}
+          contentRef={setContentRef}
         />
       </DrawerPanelBody>
     </DrawerPanelContent>
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.scss
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.scss
@@ -1,7 +1,7 @@
 @import '~bootstrap-sass/assets/stylesheets/bootstrap/variables';
 
 .co-quick-start-content {
-  @media (min-width: $screen-lg-min) {
-    min-height: calc(100vh - (var(--pf-global--spacer--4xl) + var(--pf-global--spacer--3xl)));
-  }
+  flex: 1 1 0;
+  overflow: auto;
+  padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--lg);
 }
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
@@ -15,50 +15,49 @@ type QuickStartContentProps = {
   onQuickStartChange?: (quickStartId: string) => void;
 };
 
-const QuickStartContent: React.FC<QuickStartContentProps> = ({
-  quickStart,
-  taskNumber,
-  allTaskStatuses,
-  onTaskSelect,
-  onTaskReview,
-  onQuickStartChange,
-}) => {
-  const {
-    spec: { introduction, tasks, conclusion, nextQuickStart = [] },
-  } = quickStart;
-  const totalTasks = tasks.length;
-  const nextQS = nextQuickStart.length > 0 && nextQuickStart[0];
-  return (
-    <div className="co-quick-start-content">
-      {taskNumber === -1 && (
-        <QuickStartIntroduction
-          tasks={tasks}
-          allTaskStatuses={allTaskStatuses}
-          introduction={introduction}
-          onTaskSelect={onTaskSelect}
-        />
-      )}
-      {taskNumber > -1 && taskNumber < totalTasks && (
-        <QuickStartTasks
-          tasks={tasks}
-          taskNumber={taskNumber}
-          allTaskStatuses={allTaskStatuses}
-          onTaskReview={onTaskReview}
-          onTaskSelect={onTaskSelect}
-        />
-      )}
-      {taskNumber === totalTasks && (
-        <QuickStartConclusion
-          tasks={tasks}
-          conclusion={conclusion}
-          allTaskStatuses={allTaskStatuses}
-          nextQuickStart={nextQS}
-          onQuickStartChange={onQuickStartChange}
-          onTaskSelect={onTaskSelect}
-        />
-      )}
-    </div>
-  );
-};
+const QuickStartContent = React.forwardRef<HTMLDivElement, QuickStartContentProps>(
+  (
+    { quickStart, taskNumber, allTaskStatuses, onTaskSelect, onTaskReview, onQuickStartChange },
+    ref,
+  ) => {
+    const {
+      spec: { introduction, tasks, conclusion, nextQuickStart = [] },
+    } = quickStart;
+    const totalTasks = tasks.length;
+    const nextQS = nextQuickStart.length > 0 && nextQuickStart[0];
+
+    return (
+      <div className="co-quick-start-content" ref={ref}>
+        {taskNumber === -1 && (
+          <QuickStartIntroduction
+            tasks={tasks}
+            allTaskStatuses={allTaskStatuses}
+            introduction={introduction}
+            onTaskSelect={onTaskSelect}
+          />
+        )}
+        {taskNumber > -1 && taskNumber < totalTasks && (
+          <QuickStartTasks
+            tasks={tasks}
+            taskNumber={taskNumber}
+            allTaskStatuses={allTaskStatuses}
+            onTaskReview={onTaskReview}
+            onTaskSelect={onTaskSelect}
+          />
+        )}
+        {taskNumber === totalTasks && (
+          <QuickStartConclusion
+            tasks={tasks}
+            conclusion={conclusion}
+            allTaskStatuses={allTaskStatuses}
+            nextQuickStart={nextQS}
+            onQuickStartChange={onQuickStartChange}
+            onTaskSelect={onTaskSelect}
+          />
+        )}
+      </div>
+    );
+  },
+);
 
 export default QuickStartContent;
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
@@ -8,6 +8,7 @@ import './QuickStartFooter.scss';
 
 type QuickStartFooterProps = {
   status: QuickStartStatus;
+  footerClass: string;
   taskNumber: number;
   totalTasks: number;
   onNext: () => void;
@@ -20,6 +21,7 @@ const QuickStartFooter: React.FC<QuickStartFooterProps> = ({
   totalTasks,
   onNext,
   onBack,
+  footerClass,
 }) => {
   const location = useLocation();
   const { pathname: currentPath } = location;
@@ -47,7 +49,7 @@ const QuickStartFooter: React.FC<QuickStartFooterProps> = ({
   ]);
 
   return (
-    <div className="co-quick-start-footer">
+    <div className={`co-quick-start-footer ${footerClass}`}>
       <Button
         style={{
           marginRight: 'var(--pf-global--spacer--md)',
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.scss
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.scss
@@ -1,6 +1,5 @@
 .co-quick-start-footer {
   background-color: var(--pf-global--Color--light-100);
-  bottom: 0;
-  position: sticky;
-  padding: var(--pf-global--spacer--sm) 0;
+  flex: 0 0 auto;
+  padding: var(--pf-global--spacer--md) var(--pf-global--spacer--lg);
 }
--- a/frontend/packages/console-shared/src/hooks/index.ts
+++ b/frontend/packages/console-shared/src/hooks/index.ts
@@ -3,6 +3,7 @@ export * from './deep-compare-memoize';
 export * from './document-listener';
 export * from './fullscreen';
 export * from './scroll';
+export * from './useScrollShadows';
 export * from './plugins-overview-tab-section';
 export * from './debounce';
 export * from './select-list';
--- a/frontend/packages/console-shared/src/hooks/useScrollShadows.ts
+++ b/frontend/packages/console-shared/src/hooks/useScrollShadows.ts
@@ -1,0 +1,45 @@
+import * as React from 'react';
+import { useResizeObserver } from './useResizeObserver';
+
+export enum Shadows {
+  none = 'none',
+  both = 'both',
+  top = 'top',
+  bottom = 'bottom',
+}
+
+export const useScrollShadows = (node: HTMLElement): Shadows => {
+  const [shadows, setShadows] = React.useState(Shadows.none);
+  const computeShadows = React.useCallback(() => {
+    if (node) {
+      const { scrollTop, clientHeight, scrollHeight } = node;
+      const top = scrollTop !== 0;
+      const bottom = scrollTop + clientHeight < scrollHeight;
+      if (top && bottom) {
+        setShadows(Shadows.both);
+      } else if (top) {
+        setShadows(Shadows.top);
+      } else if (bottom) {
+        setShadows(Shadows.bottom);
+      } else {
+        setShadows(Shadows.none);
+      }
+    }
+  }, [node]);
+  // recompute when the scroll container changes in size
+  useResizeObserver(computeShadows, node);
+  React.useEffect(() => {
+    if (node) {
+      // compute initial shadows
+      computeShadows();
+      // listen for scroll events
+      node.addEventListener('scroll', computeShadows);
+    }
+    return () => {
+      if (node) {
+        node.removeEventListener('scroll', computeShadows);
+      }
+    };
+  }, [node, computeShadows]);
+  return shadows;
+};


abhinandan13jan:
cc: @christianvogt @lwrigh

lwrigh:
LGTM!

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -39,41 +52,73 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
...
-  }, [previousStep]);
+    if (taskNumber > -1) return setQuickStartTaskNumber(name, taskNumber - 1);
+
+    return null;
+  }, [name, setQuickStartTaskNumber, taskNumber]);
 
   const handleTaskSelect = React.useCallback(
     (selectedTaskNumber: number) => {
       setQuickStartTaskNumber(name, selectedTaskNumber);
+      startQuickStart();
```
christianvogt, line 94: Looks like this is unrelated to the footer issue. Please raise a separate bug / pr for these other changes.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -39,41 +52,73 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
...
     },
-    [name, setQuickStartTaskNumber],
+    [name, setQuickStartTaskNumber, startQuickStart],
   );
 
+  const headerHeight = panelHeadReference.current?.offsetHeight;
+  const computedMaxHeight = window.innerHeight - (headerHeight + 6 * 24); // CSS equivalent of calc(viewportheight - (headerheight + 6 * spacer-md))
```
christianvogt, line 100: Do not compute the height based on window height. Always create components to fit their available height without assuming you know the parent container and its size.

A flow layout that uses sticky or flex or grid would work best.

christianvogt (commented):
The scroll container should not be padded. The padding should be inside the scroll container.
The current implementation causes the shadow to appear above the scroll container and also the scroll bar is offset from the right edge:
image

christianvogt:
@abhinandan13jan I don't see a bottom shadow.
Also if the footer and header are converted from sticky to have the scrollbar appear within the content area instead of spanning the entire height of the quick start panel, I don't know if shadows are still necessary. @lwrigh what's the UX stance on when to show these shadows?

lwrigh:
> @abhinandan13jan I don't see a bottom shadow.
> Also if the footer and header are converted from sticky to have the scrollbar appear within the content area instead of spanning the entire height of the quick start panel, I don't know if shadows are still necessary. @lwrigh what's the UX stance on when to show these shadows?

If the scrollbar is always present inside of the quick start content area there shouldn't be a need to also have the shadows in my opinion. @bmignano What are your thoughts?

bmignano:
> If the scrollbar is always present inside of the quick start content area there shouldn't be a need to also have the shadows in my opinion. @bmignano What are your thoughts?
I think that logic makes sense, I don't really see a need for both. But can we be sure that the scrollbar is always present / visible, even if a user's browser settings are set otherwise? @christianvogt

christianvogt:
@lwrigh  @bmignano here are the two options and whether we still want the shadow in either or both cases. It makes some sense to have the shadow for (A) and less necessary for (B)

(A) With the entire panel scrollable and sticky header and footer:
image

(B) With a flex layout and the content container scrollable only:
image

christianvogt:
There's confusion here. The scrollbar being present always is an OS option. We shouldn't be trying to detect when it's present and do something different.

My inquiry was related to the location of the scrollbar as indicated in my previous comment.

lwrigh:
> @lwrigh @bmignano here are the two options and whether we still want the shadow in either or both cases. It makes some sense to have the shadow for (A) and less necessary for (B)
> 
> (A) With the entire panel scrollable and sticky header and footer:
> image
> 
> (B) With a flex layout and the content container scrollable only:
> image

Option B makes more sense to me. I definitely prefer having the scrollbar located inside of the content area and having the title header sticky.

abhinandan13jan:
@christianvogt @lwrigh Converted to flex, moved padding inside and added shadow to footer top on scroll. Please take another look.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
abhinandan13jan: removed

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
abhinandan13jan: converted to flex

lwrigh:
@abhinandan13jan Can you post an updated screenshot for reference?

abhinandan13jan:
@lwrigh 
> @abhinandan13jan Can you post an updated screenshot for reference?

quickstartfooterscroll

lwrigh:
LGTM!

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartFooter.tsx
```diff
@@ -47,7 +49,7 @@ const QuickStartFooter: React.FC<QuickStartFooterProps> = ({
   ]);
 
   return (
-    <div className="co-quick-start-footer">
+    <div className={footerClass}>
```
christianvogt, line 52: `co-quick-start-footer` should always be present. Do not rely on the parent to pass in this class.
Concatenate the incoming class to this.

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.scss
```diff
@@ -10,6 +10,10 @@
     background: inherit;
     z-index: var(--pf-global--ZIndex--xs);
   }
+  &-body {
```
christianvogt, line 13:
```suggestion
  &__body {
```

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.scss
```diff
@@ -1,7 +1,7 @@
 @import '~bootstrap-sass/assets/stylesheets/bootstrap/variables';
 
 .co-quick-start-content {
-  @media (min-width: $screen-lg-min) {
-    min-height: calc(100vh - (var(--pf-global--spacer--4xl) + var(--pf-global--spacer--3xl)));
-  }
+  flex: 1 1 0;
+  overflow: auto;
+  padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--lg);;
```
christianvogt, line 6:
```suggestion
  padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--lg);
```

christianvogt (commented):
If you start with no scroll bar and then reduce the vertical space of the browser, you end up with scrollbars appearing inside the markdown content:
image

If you scroll such that a shadow appears, then change quick starts such that there is no scrollbar present, the previous shadow remains. This is also visible in the GIF you posted where the bottom shadow does not disappear.
image

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.scss
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.scss
abhinandan13jan: updated

abhinandan13jan:
> If you start with no scroll bar and then reduce the vertical space of the browser, you end up with scrollbars appearing inside the markdown content:
> image
> 
> If you scroll such that a shadow appears, then change quick starts such that there is no scrollbar present, the previous shadow remains. This is also visible in the GIF you posted where the bottom shadow does not disappear.
> image

@christianvogt These are existing issues. The one with markup scroll is because of hardcoded styles in the markUp component.
The scroll issue is also there on the cluster. I tried to work around with that, but need to update the hook a little which may become a breaking change.

christianvogt:
@abhinandan13jan please log a bug for the markdown scrollbar appearing and please fix the shadows so they correctly disappear otherwise it's misleading.

abhinandan13jan:
> @abhinandan13jan please log a bug for the markdown scrollbar appearing and please fix the shadows so they correctly disappear otherwise it's misleading.

Fixed the shadow issue and for the markup logged https://issues.redhat.com/browse/ODC-5338

shadow-issue-fix

sahil143:
@abhinandan13jan I'm not seeing the shadow on the footer unless I scroll. The shadow should be present if the content is scrollable to indicate that there is content present off the screen.
Screenshot 2021-01-13 at 7 07 27 PM

sahil143:
The header shadow doesn't seem to be fixed.
quick-start-scroll

abhinandan13jan:
Maybe we need to redo the QuickStart SidePanel to achieve what we are looking for. But I believe, both the issues mentioned in the above JIRA tickets have been addressed. 
cc: @andrewballantyne @nimishamukherjee @christianvogt

abhinandan13jan:
@sahil143 I updated the PR. could you please re-review?

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
```diff
@@ -13,52 +14,74 @@ type QuickStartContentProps = {
...
+
+    const debouncedResize = useDebounceCallback(resizeCallback, [resizeCallback], 50);
+
+    React.useEffect(() => {
+      const resizeObserver: ResizeObserver = new ResizeObserver(debouncedResize);
+      resizeObserver.observe(ref.current);
+      return () => {
+        resizeObserver.disconnect();
+      };
+    }, [ref, onContentUpdate, debouncedResize]);
```
sahil143, line 51: use the hook here  `useResizeObserver`

sahil143, line 51: Why is `onContentUpdate` part of dependency array?

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
```diff
@@ -13,52 +14,74 @@ type QuickStartContentProps = {
...
+    } = props;
+    const {
+      spec: { introduction, tasks, conclusion, nextQuickStart = [] },
+    } = quickStart;
+    const totalTasks = tasks.length;
+    const nextQS = nextQuickStart.length > 0 && nextQuickStart[0];
+
+    const resizeCallback: ResizeCallback = () => {
+      ref?.current && onContentUpdate();
+    };
```
sahil143, line 41: seems like you are calling this on resizeObserver only. Why do you need to forwardRef here? Why not just use the hook `useRef` and pass the current value to `onContentUpdate(ref.current)`?

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -27,18 +37,40 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   handleClose,
   activeQuickStartID,
 }) => {
-  const [scrollDirection, handleScrollCallback] = useScrollDirection();
+  const [scrollDirection, handleScrollCallback, resetScrollDirection] = useScrollDirection();
   const { t } = useTranslation();
   const quickStart = quickStarts.find((qs) => qs.metadata.name === activeQuickStartID);
 
-  const headerClasses = classNames('co-quick-start-panel-content-head', {
+  const updateShadow: UpdateShadow = (contentRef) => {
```
sahil143, line 44: use `useCallback` here

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
```diff
@@ -13,52 +14,74 @@ type QuickStartContentProps = {
...
+      onContentUpdate,
+    } = props;
+    const {
+      spec: { introduction, tasks, conclusion, nextQuickStart = [] },
+    } = quickStart;
+    const totalTasks = tasks.length;
+    const nextQS = nextQuickStart.length > 0 && nextQuickStart[0];
+
+    const resizeCallback: ResizeCallback = () => {
+      ref?.current && onContentUpdate();
```
sahil143, line 40:
```suggestion
      ref.current && onContentUpdate();
```

sahil143 (changes requested):

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
abhinandan13jan: it's a lint error otherwise

abhinandan13jan: I don't need the width or the height. It is all about checking the scrollHeight vs clientHeight. Can you please tell me how the `onResizeObserver` would serve any better, given we have a more custom implementation of the same mechanism above??

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
abhinandan13jan: Updated. thanks for pointing it out.

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
abhinandan13jan: updated. Using a memoised callback now.

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
sahil143: replace this useEffect with
```
useResizeObserver(debouncedResize, ref.current)
```

sahil143: This is custom hook. you might need to rebase with master.

abhinandan13jan: updated this as well

christianvogt:
The logic for computing the shadows is incorrect.

You need an `onScroll` event listener on the content scroll container. There is no need for a resize observer. Ideally the parent component which is managing the shadows should simply pass along an `onScroll` event handler which gets attached to the content scroll container div.

This use to work in this way when the drawer itself was the scroll container. You still have the event handler attached to the `DrawerPanelContent` component as `onScroll={handleScrollCallback}` even though this component no longer scrolls with the current implementation.

abhinandan13jan:
@sahil143 @christianvogt Updated with the resizescroller hook.Please check

frontend/packages/console-shared/src/hooks/scroll.ts
```diff
@@ -45,6 +53,41 @@ export const useScrollDirection = (): [ScrollDirection, (event) => void] => {
     },
     [scrollDirection],
   );
-
   return [scrollDirection, handleScroll];
 };
+
+export const useScrollShadows = (node: HTMLElement, debounce = 200): Shadows => {
```
sahil143, line 59: move this to a separate file

sahil143, line 59:
```suggestion
export const useScrollShadows = (node: HTMLElement, debounce: number = 200): Shadows => {
```

frontend/packages/console-shared/src/hooks/scroll.ts
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useResizeObserver, useDebounceCallback } from '@console/shared';
```
sahil143, line 2: use relative path to import these
```suggestion
import { useResizeObserver } from './useResizeObserver';
import { useDebounceCallback } from './debounce';
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -27,18 +29,19 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   handleClose,
   activeQuickStartID,
 }) => {
-  const [scrollDirection, handleScrollCallback] = useScrollDirection();
   const { t } = useTranslation();
+  const [shadows, SetShadows] = React.useState<Shadows>(Shadows.none);
```
sahil143, line 33: use camelCasing for names
```suggestion
  const [shadows, setShadows] = React.useState<Shadows>(Shadows.none);
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -27,18 +29,19 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
...
   const quickStart = quickStarts.find((qs) => qs.metadata.name === activeQuickStartID);
 
-  const headerClasses = classNames('co-quick-start-panel-content-head', {
-    'pf-u-box-shadow-sm-bottom':
-      scrollDirection && scrollDirection !== ScrollDirection.scrolledToTop,
+  const updateShadow: UpdateShadow = (shadow) => SetShadows(shadow);
```
sahil143, line 36: You can directly pass `setShadow` to the component. Why do you need to wrap it?

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -28,6 +37,22 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
     (task, index) => activeQuickStartState[`taskStatus${index}`],
   ) as QuickStartTaskStatus[];
 
+  const [contentRef, SetContentRef] = React.useState<HTMLDivElement>(null);
```
sahil143, line 40: use camelCasing 
```suggestion
  const [contentRef, setContentRef] = React.useState<HTMLDivElement>(null);
```

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -28,6 +37,22 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
...
+  const ref = React.createRef<HTMLDivElement>();
+  const shadows = useScrollShadows(contentRef, 100);
```
sahil143, line 42: Why not pass ref.current directly here?
```suggestion
  const shadows = useScrollShadows(ref.current, 100);
```

abhinandan13jan: The hook will not update unless ref.current is set as a state. This is where `contentRef` comes in and stores ref.current as state

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
abhinandan13jan: updated

frontend/packages/console-shared/src/hooks/scroll.ts
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
christianvogt: Can you move this hook to `QuickStartPanelContent.tsx` and add a prop `contentRef` that allows you to pass a ref function down to the get to the content. That way `QuickStartPanelContent.tsx` can manage the the shadows.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -28,6 +34,22 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
...
+
+  React.useEffect(() => {
+    ref.current !== contentRef && setContentRef(ref.current);
+  }, [ref, contentRef]);
+
+  React.useEffect(() => {
+    setShadows(shadows);
+  }, [shadows, setShadows]);
```
christianvogt, line 47: You won't need these hooks if you move the `useScrollShadows` to `QuickStartPanelContent`. Furthermore the first `useEffect` is invalid because `ref` is a constant and would never update even if the `ref.current` value changed.

frontend/packages/console-shared/src/hooks/scrollShadows.ts
```diff
@@ -0,0 +1,46 @@
...
+export enum Shadows {
+  none = 'none',
+  both = 'both',
+  top = 'top',
+  bottom = 'bottom',
+}
+
+export const useScrollShadows = (node: HTMLElement, debounce: number = 200): Shadows => {
+  const [shadows, setShadows] = React.useState(Shadows.none);
+  const computeShadows = useDebounceCallback(() => {
```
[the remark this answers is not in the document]
abhinandan13jan: removed Debouncing

frontend/packages/console-shared/src/hooks/scrollShadows.ts
```diff
@@ -0,0 +1,46 @@
...
+import { useDebounceCallback } from './debounce';
+
+export enum Shadows {
+  none = 'none',
+  both = 'both',
+  top = 'top',
+  bottom = 'bottom',
+}
+
+export const useScrollShadows = (node: HTMLElement, debounce: number = 200): Shadows => {
```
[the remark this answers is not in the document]
abhinandan13jan: file renamed

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -6,9 +6,15 @@ import { QuickStartContext, QuickStartContextValues } from './utils/quick-start-
 
 type QuickStartControllerProps = {
   quickStart: QuickStart;
+  footerClass: string;
+  updateContentRef: (ref: HTMLDivElement) => void;
```
christianvogt, line 10: Rename and update type:
```suggestion
  contentRef: React.Ref<HTMLDivElement>;
```

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -28,6 +34,9 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
     (task, index) => activeQuickStartState[`taskStatus${index}`],
   ) as QuickStartTaskStatus[];
 
+  const ref = React.createRef<HTMLDivElement>();
+  React.useEffect(() => updateContentRef(ref?.current));
```
christianvogt, line 38:
```suggestion
```

This approach is incorrect because an update to the `ref` doesn't cause a re-render and therefore the effect won't get run unless something else triggers the re-render. You need to use a function as the ref.

But with my other suggestions, you won't need this.

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -65,13 +74,15 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({ quickStart
         onTaskSelect={handleTaskSelect}
         onTaskReview={handleTaskStatusChange}
         onQuickStartChange={handleQuickStartChange}
+        ref={ref}
```
christianvogt, line 77: Pass the incoming prop straight down to the content child:
```suggestion
        ref={contentRef}
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -10,12 +10,14 @@ import {
   Title,
 } from '@patternfly/react-core';
 import { AsyncComponent } from '@console/internal/components/utils';
-import { useScrollDirection, ScrollDirection } from '@console/shared/';
+import { useScrollShadows, Shadows } from '@console/shared/';
```
christianvogt, line 13:
```suggestion
import { useScrollShadows, Shadows } from '@console/shared';
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -10,12 +10,14 @@ import {
...
 import { QuickStart } from './utils/quick-start-types';
 import './QuickStartPanelContent.scss';
 
 type HandleClose = () => void;
 
+type UpdateContentRef = (newRef: HTMLDivElement) => void;
```
christianvogt, line 19:
```suggestion
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -27,18 +29,24 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   handleClose,
   activeQuickStartID,
 }) => {
-  const [scrollDirection, handleScrollCallback] = useScrollDirection();
   const { t } = useTranslation();
+  const [contentRef, setContentRef] = React.useState<HTMLDivElement>();
+  const shadows = useScrollShadows(contentRef);
+
+  const updateContentRef: UpdateContentRef = (newRef) => setContentRef(newRef);
```
christianvogt, line 36:
```suggestion
```

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -59,10 +67,12 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
           </DrawerActions>
         </DrawerHead>
       </div>
-      <DrawerPanelBody>
+      <DrawerPanelBody hasNoPadding className="co-quick-start-panel-content__body">
         <AsyncComponent
           loader={() => import('./QuickStartController').then((c) => c.default)}
           quickStart={quickStart}
+          footerClass={footerClass}
+          updateContentRef={updateContentRef}
```
christianvogt, line 75:
```suggestion
          contentRef={setContentRef}
```

PR #7709, opened 2021-01-05
tests: fix timeout loading logs in CRD extensions test

spadgett:
We don't need to wait for the log content itself to check if the log links are present. The log is sometimes slow to load, causing the tests to timeout.

/cc @dtaylor113 @rhamilto


dtaylor113 (approved):
/lgtm

PR #7715, opened 2021-01-05
allow empty storage class when creating templates

suomiy:
cc @yaacov @glekner @pcbailey @rawagner @metalice


spadgett:

PR #7717, opened 2021-01-06
Bug 1912703: Query Browser: Hide graph empty state when hideGraphs is true

kyoto:
This also fixes a bug where the graph was not re-rendered after clicking
"Show graph".


spadgett (approved):
/lgtm

PR #7718, opened 2021-01-06
Bug 1913196: handle resizing of browser for guided tour

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5235

**Analysis / Root cause**: When resizing the window in the middle of the guided tour, the spotlight area fails to move with the popover.

**Solution Description**: added a resize observer on the app which indicates any change in size of view and update the spotlight position.

guidedTour-resize

**Test setup:**

**Browser conformance**: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/spotlight/StaticSpotlight.tsx
+++ b/frontend/packages/console-shared/src/components/spotlight/StaticSpotlight.tsx
@@ -1,26 +1,29 @@
 import * as React from 'react';
+import { useBoundingClientRect } from '../../hooks';
 import Portal from '../popper/Portal';
 import './spotlight.scss';
 
 type StaticSpotlightProps = {
-  element: Element;
+  element: Element | HTMLElement;
 };
 
 const StaticSpotlight: React.FC<StaticSpotlightProps> = ({ element }) => {
-  const { top, left, height, width } = element.getBoundingClientRect();
-  const style: React.CSSProperties = {
-    top,
-    left,
-    height,
-    width,
-  };
-  return (
+  const clientRect = useBoundingClientRect(element as HTMLElement);
+  const style: React.CSSProperties = clientRect
+    ? {
+        top: clientRect.top,
+        left: clientRect.left,
+        height: clientRect.height,
+        width: clientRect.width,
+      }
+    : {};
+  return clientRect ? (
     <Portal>
       <div className="pf-c-backdrop ocs-spotlight__with-backdrop">
         <div className="ocs-spotlight ocs-spotlight__element-highlight-noanimate" style={style} />
       </div>
     </Portal>
-  );
+  ) : null;
 };
 
 export default StaticSpotlight;
--- a/frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
+++ b/frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import { useResizeObserver } from './useResizeObserver';
+
+type BoundingClientRect = ClientRect | null;
+
+export const useBoundingClientRect = (targetElement: HTMLElement | null): BoundingClientRect => {
+  const [clientRect, setClientRect] = React.useState<BoundingClientRect>(() =>
+    targetElement ? targetElement.getBoundingClientRect() : null,
+  );
+
+  const observerCallback = React.useCallback(() => {
+    setClientRect(targetElement ? targetElement.getBoundingClientRect() : null);
+  }, [targetElement]);
+
+  useResizeObserver(observerCallback);
+
+  return clientRect;
+};
--- a/frontend/packages/console-shared/src/hooks/useResizeObserver.ts
+++ b/frontend/packages/console-shared/src/hooks/useResizeObserver.ts
@@ -1,0 +1,18 @@
+import * as React from 'react';
+
+export const useResizeObserver = (
+  callback: ResizeObserverCallback,
+  targetElement?: HTMLElement | null,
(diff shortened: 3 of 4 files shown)


frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const observerCallback = React.useCallback(() => {
+    setClientRect(targetElement ? targetElement.getBoundingClientRect() : null);
+  }, [targetElement]);
+
+  React.useEffect(() => {
+    if (targetElement) {
+      setClientRect(targetElement.getBoundingClientRect());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rohitkrai03, line 17: Why do you need to disable the es lint rule here? Target element shouldn't get changed anyway.

rohitkrai03 (changes requested):

frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
sahil143: Only wanted this to run when the component is mounted.

frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { useResizeObserver } from './useResizeObserver';
+
+type BoundingClientRect = ClientRect | null;
+
+export const useBoundingClientRect = (targetElement: HTMLElement | null): BoundingClientRect => {
+  const [clientRect, setClientRect] = React.useState<BoundingClientRect>(null);
```
christianvogt, line 7: Remove the effect and use a function in `useState` as a lazy initializer.
```suggestion
  const [clientRect, setClientRect] = React.useState<BoundingClientRect>(() => targetElement ? targetElement.getBoundingClientRect() : null);
```

frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
christianvogt: Use `useState(function)` for lazy initialization

frontend/packages/console-shared/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+export const useResizeObserver = (
+  callback: ResizeObserverCallback,
+  targetElement: HTMLElement | null = document.querySelector('#app'),
```
christianvogt, line 5: Do not execute a query selector in the default value otherwise it's executed each time this function is called and requires a default value.
```
const element = React.useMemo(() => targetElement ?? document.querySelector('#app'));
```

Also, why `#app` instead of the body?

frontend/packages/console-shared/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,17 @@
...
+  observerOptions: ResizeObserverObserveOptions = undefined,
+): void => {
+  const targetElementRef = React.useRef(targetElement);
+  targetElementRef.current = targetElement;
```
christianvogt, line 9: No need for a `ref` as the target element should be a dependency on the effect hook.

frontend/packages/console-shared/src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,17 @@
...
+  React.useEffect(() => {
+    const observer = new ResizeObserver(callback);
+    observer.observe(targetElementRef.current, observerOptions);
+    return () => {
+      observer.disconnect();
+    };
+  }, [callback, observerOptions]);
```
christianvogt, line 16: You need a new resize observer when the target element changes.

frontend/packages/console-shared/src/components/spotlight/StaticSpotlight.tsx
```diff
@@ -1,19 +1,22 @@
...
 const StaticSpotlight: React.FC<StaticSpotlightProps> = ({ element }) => {
-  const { top, left, height, width } = element.getBoundingClientRect();
-  const style: React.CSSProperties = {
-    top,
-    left,
-    height,
-    width,
-  };
+  const clientRect = useBoundingClientRect(element as HTMLElement);
+  const style: React.CSSProperties = clientRect
```
christianvogt, line 12: If there's no `clientRect` you probably shouldn't render the portal.

frontend/packages/console-shared/src/hooks/useBoundingClientRect.ts
```diff
@@ -0,0 +1,23 @@
...
+  }, [targetElement]);
+
+  React.useEffect(() => {
+    if (targetElement) {
+      setClientRect(targetElement.getBoundingClientRect());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useResizeObserver(observerCallback);
```
christianvogt, line 20: I think we need a new issue to move to using popperjs for the static spotlight as well. There's also another issue with the way the see through spotlight is rendered in that it affects the colors due to the css effect.

For now the resize observer is ok.

frontend/packages/console-shared/src/hooks/useResizeObserver.ts
sahil143: used body instead of `#app`

sahil143:
@christianvogt Addressed review comments, ready for review again.

PR #7719, opened 2021-01-06
Bug 1913209: Adjust messaging in support modal for community supported templates

rawagner:
Screenshot from 2021-01-06 10-34-36


yaacov:
@rawagner need to run `yarn i18n` again ... automation doesn't catch string changes as git conflict, but you need to rebase and run it.

PR #7721, opened 2021-01-06
Bug 1912381: Add id to nad form inputs

yaacov:
Adding IDs to NAD form.

Note:
the bug is actually in patternfly implementation of `FormGroup` that ignores the `fieldId` prop, this adds IDs that can help QE automation work around that issue.

Screenshot:
screenshot-0 0 0 0_8000-2021 01 06-12_59_18


yaacov:
@gouyang @rawagner please review

PR #7723, opened 2021-01-06
Bug 1914405: Fix to keep quick-search modal opened with previous input on coming back from a selection

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5279

**Analysis/Root cause:**
This is a design change wherein the quick-search modal should be opened with the previous search when coming back to topology from a selection.

**Solution:**
Save the search input while navigating to a form or catalog-page from quick-search modal similar to what to is done when a user clicks outside the modal. When coming back to topology, keep the modal opened if the previous search was saved .  

**Gif:**
Peek 2021-01-06 17-25

/kind bug

### DIFF
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
@@ -1,5 +1,6 @@
-import { CatalogService } from '@console/dev-console/src/components/catalog/service/CatalogServiceProvider';
 import * as React from 'react';
+import { CatalogService } from '@console/dev-console/src/components/catalog/service/CatalogServiceProvider';
+import { getQueryArgument } from '@console/internal/components/utils';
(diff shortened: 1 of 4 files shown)


sahil143 (commented):
@divyanshiGupta Why not set the query arguments when the user types in the quick search (`onSearch()` method in `QuickSearchModalBody`) then you wouldn't have to set it onEnter or `QuickSearchList`.

rohitkrai03:
> @divyanshiGupta Why not set the query arguments when the user types in the quick search (`onSearch()` method in `QuickSearchModalBody`) then you wouldn't have to set it onEnter or `QuickSearchList`.

I agree. We can just update the query param as and when user types in the search box and then use that query when user comes back to the previous URL. That way there won't be any need for explicitly setting the query param anywhere because of some actions.

rohitkrai03 (approved):
/approve

PR #7725, opened 2021-01-06
Bug 1886977: show correct VM Import Provider in status

suomiy:
please review @yaacov @glekner @pcbailey @rawagner @metalice

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
@@ -30,8 +30,27 @@ import { saveAndRestartModal } from '../modals/save-and-restart-modal/save-and-r
 import { history } from '@console/internal/components/utils/router';
 import { getVMTabURL } from '../../utils/url';
 import { VMTabURLEnum } from '../vms/types';
+import { StatusGroup } from '../../constants/status-group';
+import { VMImportWrappper } from '../../k8s/wrapper/vm-import/vm-import-wrapper';
+import { VMImportType } from '../../constants/v2v-import/ovirt/vm-import-type';
 
 import './vm-status.scss';
+
+const getStatusSuffixLabelKey = (vmStatusBundle: VMStatusBundle) => {
+  if (vmStatusBundle.status.getGroup() === StatusGroup.VMIMPORT) {
+    switch (new VMImportWrappper(vmStatusBundle.vmImport).getType()) {
+      case VMImportType.OVIRT:
+        // t('kubevirt-plugin~RHV')
+        return 'kubevirt-plugin~RHV';
+      case VMImportType.VMWARE:
+        // t('kubevirt-plugin~VMware')
+        return 'kubevirt-plugin~VMware';
+      default:
+        break;
+    }
+  }
+  return undefined;
+};
 
 type LinkType = {
   to: string;
@@ -164,7 +183,11 @@ export const VMStatus: React.FC<VMStatusProps> = ({
   const vmiLike = vm || vmi;
 
   const { status, pod, progress, importerPodsStatuses } = vmStatusBundle;
-  const title = t(status.getLabelKey()) || status.toString() || t('kubevirt-plugin~Unknown'); // TODO status.toVerboseString() should be called to pass to popup header
+
+  const title =
+    t(status.getLabelKey()) ||
+    status.toString(t(getStatusSuffixLabelKey(vmStatusBundle))) ||
+    t('kubevirt-plugin~Unknown');
   const popoverTitle = arePendingChanges ? 'Pending Changes' : null;
   const message = vmStatusBundle.message || vmStatusBundle.detailedMessage;
   const detailedMessage = vmStatusBundle.message ? vmStatusBundle.detailedMessage : null;
--- a/frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
+++ b/frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
@@ -95,18 +95,21 @@ export abstract class StatusEnum<SIMPLE_LABEL = StatusSimpleLabel> extends Objec
 
   getSimpleLabel = () => this.simpleLabel;
 
-  toString() {
+  toString(group?: string) {
     const result = this.label || super.toString();
-    return this.group && !this._isUnknown ? `${result} (${this.group.toString()})` : result;
+    if (!this._isUnknown) {
+      if (group) {
+        return `${result} (${group})`;
+      }
+      if (this.group) {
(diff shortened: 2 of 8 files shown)


frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
```diff
@@ -164,7 +181,11 @@ export const VMStatus: React.FC<VMStatusProps> = ({
   const vmiLike = vm || vmi;
 
   const { status, pod, progress, importerPodsStatuses } = vmStatusBundle;
-  const title = t(status.getLabelKey()) || status.toString() || t('kubevirt-plugin~Unknown'); // TODO status.toVerboseString() should be called to pass to popup header
+
+  const title =
+    t(status.getLabelKey()) ||
+    status.toString(getStatusSuffix(vmStatusBundle, t)) ||
```
suomiy, line 187: I noticed this is not properly localized so I did not touch this issue in this bug. Are there any plans on fixing this? @yaacov @pcbailey

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
```diff
@@ -30,9 +30,26 @@ import { saveAndRestartModal } from '../modals/save-and-restart-modal/save-and-r
 import { history } from '@console/internal/components/utils/router';
 import { getVMTabURL } from '../../utils/url';
 import { VMTabURLEnum } from '../vms/types';
+import { StatusGroup } from '../../constants/status-group';
+import { VMImportWrappper } from '../../k8s/wrapper/vm-import/vm-import-wrapper';
+import { VMImportType } from '../../constants/v2v-import/ovirt/vm-import-type';
 
 import './vm-status.scss';
 
+function getStatusSuffix(vmStatusBundle: VMStatusBundle, t) {
```
glekner, line 39: can you change to arrow func

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
glekner, line 187: I believe you should call `t` on `status.toString` not inside it.  @yaacov wdyt

frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
```diff
@@ -95,20 +95,23 @@ export abstract class StatusEnum<SIMPLE_LABEL = StatusSimpleLabel> extends Objec
 
   getSimpleLabel = () => this.simpleLabel;
 
-  toString() {
+  toString(group?: string) {
```
glekner, line 98: This is too specific for `toString`

why not change `toVerboseString` accordingly?

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
suomiy: done

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
suomiy: I know it is not perfect but it is adhering to the recent convention. As I mentioned in https://github.com/openshift/console/pull/7725#discussion_r552608938 we can do a full refactoring but it is outside of the scope of this bug.

frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
suomiy: because other enums are using the toString for the status without the group..

suomiy: also toVerboseString was removed since it was not used

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
```diff
@@ -30,9 +30,26 @@ import { saveAndRestartModal } from '../modals/save-and-restart-modal/save-and-r
 import { history } from '@console/internal/components/utils/router';
 import { getVMTabURL } from '../../utils/url';
 import { VMTabURLEnum } from '../vms/types';
+import { StatusGroup } from '../../constants/status-group';
+import { VMImportWrappper } from '../../k8s/wrapper/vm-import/vm-import-wrapper';
+import { VMImportType } from '../../constants/v2v-import/ovirt/vm-import-type';
 
 import './vm-status.scss';
 
+const getStatusSuffix = (vmStatusBundle: VMStatusBundle, t) => {
```
yaacov, line 39:
```suggestion
const getStatusSuffixLabelKey = (vmStatusBundle: VMStatusBundle) => {
```

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
```diff
@@ -30,9 +30,26 @@ import { saveAndRestartModal } from '../modals/save-and-restart-modal/save-and-r
...
+  if (vmStatusBundle.status.getGroup() === StatusGroup.VMIMPORT) {
+    switch (new VMImportWrappper(vmStatusBundle.vmImport).getType()) {
+      case VMImportType.OVIRT:
+        return t('kubevirt-plugin~RHV');
```
yaacov, line 43:
```suggestion
        // t('kubevirt-plugin~RHV')
        return 'kubevirt-plugin~RHV';
```

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
yaacov, line 187:
```suggestion
    status.toString(getStatusSuffix(vmStatusBundle, t)) ||
    status.toString(t(getStatusSuffixLabelKey(vmStatusBundle))) ||
```

frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
yaacov: if we only using the `group` in one place ?
we can do `{{status}} ({{group}})` there ?

frontend/packages/kubevirt-plugin/src/components/vm-status/vm-status.tsx
suomiy: thanks, fixed

frontend/packages/kubevirt-plugin/src/constants/status-enum.ts
suomiy: it is not the only place, so unless we do larger refactoring we need the group inside the status

PR #7727, opened 2021-01-06
Bug 1894020: fix Internal registry deploy flow

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5259

**Analysis / Root cause**: 

Users with edit right are unable to deploy images from their own namespace in the UI but they can do it using the CLI.
**Solution Description**: 

Reverting back to 3.11 like flow for avoiding the permission issue caused while granting access via UI

1. User chooses the same project
image

2. User chooses a different project
image

image

**Test setup:**

1. Create a user with edit right to a namespace
2. go to the namespace and use internal image registry option
3. now choose an image stream from the user's own namespace for creation.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @parvathyvr @invincibleJai


parvathyvr (changes requested):
Hi @karthikjeeyar ! looking at the screenshot of the inline alert-I feel the readability of the message is not great. So can we try an alert with title,description and action?
As discussed over slack, here are the changes:
-the Title can be the first line of the current message. 
-Description can be the "You can grant authority with the command <oc command>" and can we differentiate the oc command with a different style?
-The action will be 'Click to copy the command'

karthikjeeyar:
@parvathyvr  Updated the code with requested changes

parvathyvr (changes requested):
Hi @karthikjeeyar After the update the message looks better. But now the blue background around the oc command part on the yellow alert background looks weird.
Can we use this PF 'Clipboard copy' component for the oc command part?
https://www.patternfly.org/v4/components/clipboard-copy/#read-only-expanded
This will have the built in 'copy to clipboard' action and user can expand and see the command if they want.

parvathyvr (approved):
LGTM!

rottencandy (approved):
Tested locally.
/lgtm

invincibleJai:
Thanks @karthikjeeyar  looks good and have verified below scenarios

Assumption:

-  user with admin privileges with username kubeadmin , creates a ns ns-admin1  and ns-admin2
-  kubeadmin creates a self-provisioner user Developer  and provides view access to ns-admin1 and edit access ns-admin2
-  Developer user create namespace ns-dev1 and ns-dev2

 
 
 Scenario1: Kubeadmin logs in and able to import images from across all namespace 🎉 
 
 Scenario2: Developer logs in
-  Developer is in namespace ns-admin1 and as he has view access so can't see option to deploy image
-  Developer is in namespace ns-admin1 and can successfully import image from ns-admin2 and ns-dev1
-  Developer is in namespace ns-dev1 and can successfully import image from ns-admin1
-  Developer is in namespace ns-dev2 and can successfully import image from ns-admin2

All works as expected and was able to import however i din't had to use `oc policy  ...` command.

cc @christianvogt

karthikjeeyar:
@invincibleJai As we have noticed that all the newly created projects has a default role-bindings, which includes `system:image-pullers` role attached to it. This role in every namespace allows the user to import imagestream across namespaces.

image

By deleting this `sytem:image-pullers` rolebinding in a namespace causes the import to fail with `errImagePull`. However there is no harm in showing the oc command in the alert, which will help the user to create if the necessary rolebinding is not available.
cc: @christianvogt

PR #7728, opened 2021-01-06
Bug 1908650:  The title of customize wizard is inconsistent

glekner:
image

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import * as _ from 'lodash';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
@@ -67,9 +68,9 @@ import { ValidTabGuard } from './tabs/valid-tab-guard';
 import { FirehoseResourceEnhanced } from '../../types/custom';
 import { parseVMWizardInitialData } from '../../utils/url';
 import { VMWizardInitialData } from '../../types/url';
+import { getTemplateName } from '../../selectors/vm-template/basic';
 
 import './create-vm-wizard.scss';
-import { useTranslation } from 'react-i18next';
 
 type CreateVMWizardComponentProps = {
   isSimpleView: boolean;
@@ -154,18 +155,32 @@ const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props) 
   });
 
   const getWizardTitle = () => {
-    const { isCreateTemplate, isProviderImport, iUserTemplate } = props;
+    const {
+      isCreateTemplate,
+      isProviderImport,
+      iUserTemplate,
+      commonTemplates,
+      initialData: { commonTemplateName },
+    } = props;
     if (isCreateTemplate) {
       return t('kubevirt-plugin~Create Virtual Machine template');
     }
     if (isProviderImport) {
       return t('kubevirt-plugin~Import Virtual Machine');
     }
-    return iUserTemplate
-      ? t('kubevirt-plugin~Create Virtual Machine from {{template}}', {
-          template: iGetName(iUserTemplate),
-        })
-      : t('kubevirt-plugin~Create Virtual Machine');
+    if (iUserTemplate) {
+      return t('kubevirt-plugin~Create Virtual Machine from {{template}}', {
+        template: iGetName(iUserTemplate),
+      });
+    }
+
+    const templates = immutableListToShallowJS(iGetLoadedData(commonTemplates));
+    const template = templates?.find((tmp) => tmp.metadata.name === commonTemplateName);
+    const templateName = getTemplateName(template);
+
+    return t('kubevirt-plugin~Create Virtual Machine from {{template}}', {
+      template: templateName,
+    });
   };
 
   const goBackToEditingSteps = () =>


glekner:
please review @suomiy @yaacov @metalice

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
```diff
@@ -154,18 +155,32 @@ const CreateVMWizardComponent: React.FC<CreateVMWizardComponentProps> = (props)
...
+      return t('kubevirt-plugin~Create Virtual Machine from {{template}}', {
+        template: iGetName(iUserTemplate),
+      });
+    }
+
+    const templates = immutableListToShallowJS(iGetLoadedData(commonTemplates));
+    const template = templates?.find((tmp) => tmp.metadata.name === commonTemplateName);
+    const templateName = getTemplateName(template);
+
+    return t('kubevirt-plugin~Create Virtual Machine from {{template}}', {
```
suomiy, line 181: Are you really sure this is what we want in this case? The template and os/flavor can change in the wizard.

Either we should change the title each time it gets switched (which is a bit weird IMO). Or keep the old title?

suomiy:
/hold

please take https://github.com/openshift/console/pull/7728#pullrequestreview-562764694 into consideration

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard.tsx
yaacov: > Are you really sure this is what we want in this case? The template and os/flavor can change in the wizard.

We get to this page after they choose a template, so users will see the os they selected in the first tab of the wizard,
The Template they choose may have different flavour then they want (we don't show flavor in the first step), this is why we use `getTemplateName` method that will remove the flavour from the name.

Note: the problem we are trying to fix is seen in this two similar flows, that will show different titles,
on the first one the user gets the title "Create Virtual Machine" but on the other "Create VIrtual Machine from <template name>"

Screenshot:
User choose a template in first step, and get the title "Create Virtual Machine":
Peek 2021-01-06 16-55

User choose a template in first stepp, and gets the title "create VIrtual Machine from my-template"
Peek 2021-01-06 16-58

> Either we should change the title each time it gets switched (which is a bit weird IMO). Or keep the old title?

The bug was about the different titles between the similar flows, so we need to adjuat the title to be the same in the too flows and also be helpful, reminding the users that the choose <template> in the first step and are now creating a VM from that template while customizing it is a good title IMHO

suomiy: I understand the original problem, but we will run into this inconsistency when running with this approach. Meaning the VM will not be created from the advertised RHEL in the title but from fedora template.

aa


So maybe we could switch the title upon the os change?

yaacov: > So maybe we could switch the title upon the os change?

The user already choose OS (and workload) in the first step when they choose the common template to start the customization from, IMHO removing the `Operating System` and `Workload` drop-downs is the correct way, we use this solution when a user choose a user defined template. WDYT ?

suomiy: I think the point of the advanced wizard is to fine tune the selection so I would keep the dropdowns. Because you either can't find the right combination in the template cards or you change your mind later in the flow. And it would be quite painful to restart everything.

yaacov: @rawagner @glekner @metalice @matthewcarleton  hi,

We have a question we need help on, 

a. On the one side, letting user choose OS again after they choose a template is redundant and creates a big logic mess in the wizard that need to re-calculate the common template each time use choose os/flavor/workflow, removing the options to choose OS and Workflow at this stage makes a lot of sense to me, both in the UX side, don't repeat yourself, and on the logic side it will remove a lot of code that check for OS/flavor/workload change.

b. On the other hand, @suomiy explains that it will give users more options, @suomiy can you add below more info about the pros and cons of letting users change OS here ?

suomiy: > a. On the one side, letting user choose OS again after they choose a template is redundant and creates a big logic mess in the wizard that need to re-calculate the common template each time use choose os/flavor/workflow, removing the options to choose OS and Workflow at this stage makes a lot of sense to me, both in the UX side, don't repeat yourself, and on the logic side it will remove a lot of code that check for OS/flavor/workload change.

There is no option to select the right flavor and workload in the simple template select. It just selects the os. So removing the logic from the advanced wizard will not be possible 

> b. On the other hand, @suomiy explains that it will give users more options, @suomiy can you add below more info about the pros and cons of letting users change OS here ?

I think the explanation of a. is explaing b. as well. The user wants to customize the selection since this is the job of the advanced wizard. And while the user is customizing, the user might as well change the os if s/he cannot find the right workload. If we remove these options it will become very painful to choose anything for the users. For every change they would have to restart both wizards (simple and advanced) and go through the whole process again.

yaacov: > There is no option to select the right flavor 

Yes, we should leave a Flavour drop down, but it should not change the template chosen in the first step, only the cpu/memory.

> and workload in the simple template select. It just selects the os. So removing the logic from the advanced wizard will not be possible

Removing the workflow option in the first step of choosing the template was a deliberate decision, the reason was because it does not change the resulting VM in a way that will affect the user. IMHO we don't need to re-introduce an option that has no real value to the user.

yaacov: > The user wants to customize the selection since this is the job of the advanced wizard. And while the user is customizing, 

maybe the flow of:
`choose a template -> customise a VM (including choosing a new template)`
is wrong ?
maybe we need to separate flows:
a. create a VM from template, (and if users need, they will customize it later in the details pages based on that base template)
b. create a new VM without a template where user can customize anything similar to new from YAML where user is not restricted by a base template.

WDYT ?

matthewcarleton: I don't think we should allow them to create a VM without a template. It's not very useful to throw them into a flow and ask them to do a bunch of stuff that we can do for them. 
I agree that we should change the OS dropdown to read only OR remove it entirely. They've chosen their OS from the list so it's reasonable to assume they won't need to commonly change this (maybe edge cases). If we take this approach then we don't need to dynamically change the title of the wizard but we still allow them to make any other customization that they want.
WDYT @suomiy @yaacov ?

yaacov: > I don't think we should allow them to create a VM without a template. It's not very useful to throw them into a flow and ask them to do a bunch of stuff that we can do for them.

I agree, also we have "create from YAML" to cover this adage cases.

>  agree that we should change the OS dropdown to read only OR remove it entirely. They've chosen their OS from the list so it's reasonable to assume they won't need to commonly change this (maybe edge cases). If we take this approach then we don't need to dynamically change the title of the wizard but we still allow them to make any other customization that they want.

I agree, we can make an inform decision about how exactly to implement that (disable/remove) in an off-line more relaxed discussion.

In the mean time I will remove the hold from this PR.

rawagner: I think disabling OS field and keeping flavor/workload changeable makes sense

suomiy: +1

suomiy: posted a fix https://github.com/openshift/console/pull/7744

PR #7729, opened 2021-01-06
Bug 1913332: Fix pipeline visualization issue when taskspecs is used

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5306

**Analysis / Root cause**: 
Pipeline details page is broken if the newly created pipeline contains multiple taskSpecs in the yaml.

**Solution Description**: 
Add check for `taskspec` before load the pipeline visulation 

image

**Test setup:**
Install Openshift Pipleine operator
and create this pipeline https://github.com/karthikjeeyar/pipelines/blob/main/taskspec.yaml

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
@@ -19,7 +19,6 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
   pipelineRun,
 }) => {
   const { t } = useTranslation();
-  const { nodes, edges } = getTopologyNodesEdges(pipeline, pipelineRun);
   let content: React.ReactElement;
   if (hasInlineTaskSpec(pipeline.spec.tasks)) {
     // TODO: Inline taskSpec is not yet supported feature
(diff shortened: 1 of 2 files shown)


jerolimov (approved):
Tested and verified. Works fine.

/lgtm

karthikjeeyar:
1.Copy a taskspec yaml  https://github.com/karthikjeeyar/pipelines/blob/main/taskspec.yaml
2. go to the pipeline builder page -> yaml view -> paste the yaml
3. switch back to pipeline builder breaks the page.

As we are not supporting the taskspec in builder component yet, we should gracefully discard the taskspec pipeline yaml or reset to initialvalues when switching to builder view. cc: @andrewballantyne

karthikjeeyar:
@vikram-raj Pipeline builder with taskSpec is not broken while switching from yaml view to builder view. It would be better if we add some test cases.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
```diff
@@ -32,7 +31,12 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
         )}
       />
     );
-  } else if (nodes.length === 0 && edges.length === 0) {
+    return <div className="odc-pipeline-visualization">{content}</div>;
```
christianvogt, line 34: Remove this duplicate `return` statement. Since you already assign the `content`, add an `} else {` statement after this to block the rest of the conditions and continue to use the single `return` statement.

karthikjeeyar:
works fine
/lgtm

PR #7733, opened 2021-01-06
Bug 1913660: update location and text of Pipeline edit flow alert

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5277

**Solution Description**: 
- changed the position of the  info alert when editing a pipeline from the pipeline section to the build config section and change the text
- changed the pipeline info alert message

**Screen shots:** 
p2
p1

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


bgliwa01 (approved):
lgtm!

frontend/packages/dev-console/src/components/import/builder/BuilderImageSelector.tsx
```diff
@@ -29,6 +41,35 @@ const BuilderImageSelector: React.FC<BuilderImageSelectorProps> = ({
...
+        isInline
+        variant="info"
+        title={t(
+          'devconsole~There are no supported pipelines available for {{builderImage}}. Changing to this builder image will disconnect your associated Pipeline.',
+          { builderImage: imageName },
+        )}
+      />
+      <br />
+    </>
+  );
```
rohitkrai03, line 72: Can you cleanup the logic a bit so that it's easier to follow? Also you don't need to repeat yourself with writing two alerts exactly same with just different titles. Just add a check to define this title in a variable and then use that in the alert jsx.

rohitkrai03 (changes requested):

andrewballantyne (approved):
/lgtm

PR #7736, opened 2021-01-06
CONSOLE-2400: Bump react to v17

zherman0:
Bump React from 16.x to 17.0.1 and other dependencies as needed.
[CONSOLE-2400](https://issues.redhat.com/browse/CONSOLE-2400)

Did not bump types/react to 17 as there were serious conflicts. Still investigating options for that.


spadgett:
@zherman0 Looks like enzyme doesn't support React 17, which is causing the test errors, so we might be blocked. See https://github.com/enzymejs/enzyme/issues/2429

/cc @christianvogt

zherman0:
/hold
We have several dependency issues that have not been resolved by third party modules.  We are holding this until 4.8 to give those external modules time to update to React 17.

spadgett:

zherman0:
@spadgett - Enzyme still does not support React 17, however yarn add --dev @wojtekmaj/enzyme-adapter-react-17 is suppose to be a viable workaround.  Shall I replace the official enzyme with this personal repo and test everything?

spadgett:
Yeah, I'm good with adding the adapter to unblock us until enzyme adds support.

frontend/before-tests.js
```diff
@@ -1,7 +1,8 @@
 /* eslint-env node */
 
 import { configure } from 'enzyme';
-import * as Adapter from 'enzyme-adapter-react-16';
+// import * as Adapter from 'enzyme-adapter-react-16';
```
spadgett, line 4: Remove commented-out code.

frontend/@types/gql/queries.d.ts
```diff
@@ -0,0 +1,18 @@
+
```
spadgett, line 1: I thought we had GraphQL types in `@types/console/generated` so this shouldn't be necessary?

cc @rawagner

rawagner: Thats right. The location changed recently in https://github.com/openshift/console/pull/8343. `@types/gql` is the old location which is not ignored by git anymore so got commited here probably.

zherman0: Removed the extra files

spadgett (approved):
/lgtm
/hold for approvals

@yapei This simply needs a basic regression test of console. There is no visible change to the UI.

PR #7201, opened 2020-11-13
Bug 1893601: fix filesystem queries

sjenning:
In RHCOS, the only part of the filesystem that is usable by OpenShift is `/var`.
- `/var/lib/kubelet` holds pod volumes
- `/var/lib/containers` holds the container writeable layers and images.
- `/var/log` contains system logs.

```
# lsblk
NAME                         MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda                            8:0    0   120G  0 disk 
|-sda1                         8:1    0   384M  0 part /boot
|-sda2                         8:2    0   127M  0 part /boot/efi
|-sda3                         8:3    0     1M  0 part 
`-sda4                         8:4    0 119.5G  0 part 
  `-coreos-luks-root-nocrypt 253:0    0 119.5G  0 dm   /sysroot

# df -h /sysroot
Filesystem                            Size  Used Avail Use% Mounted on
/dev/mapper/coreos-luks-root-nocrypt  120G   17G  104G  14% /sysroot

# df -h /var
Filesystem                            Size  Used Avail Use% Mounted on
/dev/mapper/coreos-luks-root-nocrypt  120G   17G  104G  14% /var
```
Thus it is the only mountpoint whose capacity and utilization we care about.

/cc @rawagner 

Also see previous attempt to fix this:
https://bugzilla.redhat.com/show_bug.cgi?id=1874028
https://github.com/openshift/console/pull/6571

PR #7739, opened 2021-01-07
Bug 1909004: make filesystem queries compatible with both RHCOS and RHEL nodes

sjenning:
https://github.com/openshift/console/pull/7201 fixed filesystem queries for RHCOS nodes but does not work for RHEL nodes.

For RHCOS, both `/` and `/var` are bind mounts to the same underlying `/sysroot`, which is just a dm-crypt on the 4th GPT disk partition.

```
$ lsblk
NAME                         MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
vda                          252:0    0   120G  0 disk 
├─vda1                       252:1    0   384M  0 part /boot
├─vda2                       252:2    0   127M  0 part /boot/efi
├─vda3                       252:3    0     1M  0 part 
└─vda4                       252:4    0 119.5G  0 part 
  └─coreos-luks-root-nocrypt 253:0    0 119.5G  0 dm   /sysroot
```

Thus, using `/` instead of `/var` is less precise for RHCOS nodes, but should still provide correct behavior for RHCOS and mostly correct behavior for RHEL nodes.  If the RHEL node is using a separate `/var` partition, it will not be correct.  Right now, however, RHEL nodes show no data at all, which is the bug for which this PR is opened.

@spadgett @jhadvig

### DIFF
--- a/frontend/packages/console-app/src/components/nodes/NodesPage.tsx
+++ b/frontend/packages/console-app/src/components/nodes/NodesPage.tsx
@@ -468,11 +468,11 @@ const fetchNodeMetrics = (): Promise<NodeMetrics> => {
     {
       key: 'usedStorage',
       query:
-        'sum by (instance) (node_filesystem_size_bytes{mountpoint="/var"} - node_filesystem_free_bytes{mountpoint="/var"})',
+        'sum by (instance) (node_filesystem_size_bytes{mountpoint="/"} - node_filesystem_free_bytes{mountpoint="/"})',
     },
     {
       key: 'totalStorage',
-      query: 'sum by (instance) (node_filesystem_size_bytes{mountpoint="/var"})',
+      query: 'sum by (instance) (node_filesystem_size_bytes{mountpoint="/"})',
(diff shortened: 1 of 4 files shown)


sjenning:
Verified this still works for a full RHCOS cluster in AWS.  Verifying on a RHEL worker next.

spadgett (approved):
/lgtm
/hold since @sjenning is still testing

sjenning:
Yep, it works on the RHEL worker
/hold cancel

kyoto:

aravindhp:
This ended up breaking the filesystem graph for Windows nodes 😞

aravindhp:
@openshift/openshift-team-windows-containers is looking into why and how to prevent such breakages in the future.

aravindhp:
Seems like https://github.com/openshift/console/pull/7201/ is the PR that broke Windows and not this PR. We are looking into adding an e2e test to the console repo or WMCO repo to catch such breakages in the future.

PR #7744, opened 2021-01-07
Bug 1913685: disable OS field when creating a VM from a template

suomiy:


suomiy:
please review @yaacov @glekner @rawagner @metalice

PR #7745, opened 2021-01-07
Bug 1895169:  VM Template does not properly manage Mount Windows guest tools check box

glekner:


glekner:
please review @yaacov @suomiy

frontend/packages/kubevirt-plugin/src/selectors/immutable/template/combined.ts
```diff
@@ -138,3 +139,8 @@ export const getITemplateDefaultWorkload = (template: ITemplate): WorkloadProfil
   WorkloadProfile.getAll().find(
     (w) => iGetLabels(template).get(`${TEMPLATE_WORKLOAD_LABEL}/${w.getValue()}`) === 'true',
   );
+
+export const iGetTemplateGuestToolsDisk = (tmp: ITemplate, diskName: string = WINTOOLS_DISK_NAME) =>
```
suomiy, line 143: I think we should rather use isWinToolsImage function to have the detection consistent across the code base

glekner:
please re-review @suomiy

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/redux/state-update/vm-settings-tab-state-update.ts
```diff
@@ -93,11 +94,14 @@ const osUpdater = ({ id, prevState, dispatch, getState }: UpdateOptions) => {
   const os = iGetVmSettingValue(state, id, VMSettingsField.OPERATING_SYSTEM);
   const isWindows = os?.startsWith('win');
 
+  const iUserTemplate = iGetLoadedCommonData(state, id, VMWizardProps.userTemplate);
+  const mountTools = !iUserTemplate || !!iGetTemplateGuestToolsDisk(iUserTemplate);
```
suomiy, line 98: the value still should depend on the OS no? Eg. when you are creating a template.

glekner: fixed

PR #7746, opened 2021-01-07
Bug 1907908: VM Wizard always uses VirtIO for the VM rootdisk

glekner:


glekner:
@suomiy @yaacov @metalice please review

PR #7747, opened 2021-01-07
Bug 1913725: Refactor Insights Operator Plugin states

gkarat:
Changes behavior of the Insights plugin for a case when the operator has been initialized or disabled. It showed a green OK icon with 0 issues found before, which was not correct.

https://issues.redhat.com/browse/CCXDEV-3321

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -1,12 +1,13 @@
 import * as React from 'react';
 import { ChartDonut, ChartLegend, ChartLabel } from '@patternfly/react-charts';
+
 import {
   riskIcons,
   colorScale,
   legendColorScale,
   riskSorting,
   mapMetrics,
-  isInitState,
+  isInitialized,
 } from './mappers';
 import { PrometheusHealthPopupProps } from '@console/plugin-sdk';
 import { K8sResourceKind } from '@console/internal/module/k8s';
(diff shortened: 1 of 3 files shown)


frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -19,14 +21,14 @@ const DataComponent: React.FC<DataComponentProps> = ({ x, y, datum }) => {
 };
 
 export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses, k8sResult }) => {
-  const resource = mapMetrics(responses[0].response);
-  const clusterID = (k8sResult as K8sResourceKind)?.data?.spec?.clusterID || '';
-  const riskEntries = Object.entries(resource).sort(
+  const metrics = mapMetrics(responses[0].response);
+  const clusterID = get(k8sResult as K8sResourceKind, 'data.spec.clusterID', '');
```
spadgett, line 25: Is there a reason to use lodash get over optional chaining?

tisnik (approved):
Looks good, but TBH I have not tried it on my side yet

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
gkarat: @spadgett  no special reason. Is it a worse solution than chaining? Why?

bond95: @gkarat because using of language construction is always more preferable then using function that is doing same.

gkarat: @bond95 got it, changed :)

spadgett (approved):
/lgtm

PR #7749, opened 2021-01-07
Bug 1911382: If source PVC is selected, volumeMode changes to source volumeMode.

metalice:
If source PVC is selected, volumeMode changes to source volumeMode. volumeMode field will be disabled for changing and info message will be presented

*Also fixed a small typo in 2 functions names.

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1911382

**Analysis / Root cause**: 
If source PVC was presented, volumeMode didn't change according to it.

**Solution Description**: 
Added support to read and apply the value of volumeMode from source PVC.

Bug 1911382

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/storage-tab/vm-wizard-storage-modal-enhanced.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/storage-tab/vm-wizard-storage-modal-enhanced.tsx
@@ -59,7 +59,7 @@ const VMWizardStorageModal: React.FC<VMWizardStorageModalProps> = (props) => {
   );
 
   const [namespace, setNamespace] = React.useState<string>(
-    new DataVolumeWrapper(dataVolume).getPesistentVolumeClaimNamespace() || vmNamespace,
+    new DataVolumeWrapper(dataVolume).getPersistentVolumeClaimNamespace() || vmNamespace,
   );
 
   const resources = [
--- a/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
@@ -177,6 +177,18 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
     (isEditing && combinedDisk.getVolumeMode()) || null,
   );
 
+  React.useEffect(() => {
+    if (source.requiresPVC()) {
+      const pvcNameDataVolume = dataVolume.getPersistentVolumeClaimName();
+      const pvcNamespaceDataVolume = dataVolume.getPersistentVolumeClaimNamespace();
+      const pvc = persistentVolumeClaims?.data.find(
+        ({ metadata }) =>
+          metadata?.name === pvcNameDataVolume && metadata?.namespace === pvcNamespaceDataVolume,
+      );
+      setVolumeMode((value) => VolumeMode.fromString(pvc?.spec?.volumeMode) || value);
+    }
+  }, [dataVolume, persistentVolumeClaims, source]);
+
   const [defaultAccessMode, defaultVolumeMode, isScModesKnown] = React.useMemo(() => {
     return [
       getDefaultSCAccessModes(storageClassConfigMap, storageClassName)?.[0],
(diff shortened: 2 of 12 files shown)


metalice:
please review @suomiy @rawagner @yaacov  :)

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/storage-tab/vm-wizard-storage-modal-enhanced.tsx
```diff
@@ -59,7 +59,7 @@ const VMWizardStorageModal: React.FC<VMWizardStorageModalProps> = (props) => {
   );
 
   const [namespace, setNamespace] = React.useState<string>(
-    new DataVolumeWrapper(dataVolume).getPesistentVolumeClaimNamespace() || vmNamespace,
+    new DataVolumeWrapper(dataVolume).getPersistentVolumeClaimNamespace() || vmNamespace,
```
yaacov, line 62: :-)

yaacov:
p.s.
we have two alerts, one on top and one on the bottom, can you consolidate them into one info block, ( bottom or top , just let them be in one place ) ?

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -177,6 +177,18 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
...
+    if (source.requiresPVC()) {
+      const pvcNameDataVolume = dataVolume.getPersistentVolumeClaimName();
+      const pvcNamespaceDataVolume = dataVolume.getPersistentVolumeClaimNamespace();
+      const pvc = persistentVolumeClaims?.data.find(
+        ({ metadata }) =>
+          metadata?.name === pvcNameDataVolume && metadata?.namespace === pvcNamespaceDataVolume,
+      );
+      setVolumeMode((value) => VolumeMode.fromString(pvc?.spec?.volumeMode) || value);
+    }
+  }, [dataVolume, persistentVolumeClaims, source]);
```
suomiy, line 190: please don't use dataVolume and persistentVolumeClaims as dependencies for useEffect as they can have a different reference each rerender and will fire every time. Please guard it with loaded status of persistentVolumeClaims, source, and volumeMode

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
```diff
@@ -627,33 +639,50 @@ export const DiskModal = withHandlePromise((props: DiskModalProps) => {
...
-                                    {
-                                      storageClassName,
-                                    },
-                                  )
-                                : '',
-                            )}
-                          />
-                        ))}
-                      </FormSelect>
+                      <Stack hasGutter>
```
suomiy, line 642: +1 this is a nice solution for disk modals, but it is not enough for the wizard. Template disks are not required to go through disk modals and such volumeMode mistakes will not be catched. Are you planning to fix it here or in the followup?

suomiy (changes requested):

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
metalice: it's a solid point. added a custom hook for this situation.

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
metalice: It's been added automatically by the linter, basically, there aren't any side effects that can cause performance issues. I agree that if there is a performance issue we should avoid it, but I don't think this is the case.

metalice:
@rawagner , @yaacov , @suomiy  - edit a bit the PR, please check :)

yaacov:
all comments addressed, canceling hold

/hold cancel
/lgtm

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
suomiy: the linter does't check for these things, only in very limited scope. It is not like you can measure the performance issues, but the code isn't crisp and fires setVolumeMode needlesly.

metalice: volumeMode is a string, a primitive type, useState can compare it easily,  it won't get updated if the value is the same, so in this case, I think no need for concern :)

frontend/packages/kubevirt-plugin/src/components/modals/disk-modal/disk-modal.tsx
suomiy: Can we please move this to an updater? The hook works nicely in this scenario, but it poses following issues.

1. Consistency: we try to have all logic in the updaters so we can have one place to look at for changes in the state. If we have the logic scattered everywhere it would become very hard to naviagate. Simple hooks work great for simple things, but if we try to use this approach for the whole wizard it would become a mess.

2. Optimization: this hook is called needlessly on any redux change and when it fires the update it fires the updaters  twice.

Also this logic should target all DV and not just a rootstorage (+ it isn't a good practice to target it by name)

PR #7750, opened 2021-01-07
Bug 1913568: Normal user cannot create URL template

glekner:


glekner:
@rawagner @suomiy can you approve this? 
Normal user cannot create a URL template with the following error:
```
"datavolumes.cdi.kubevirt.io "url-rootdisk-bzvw8" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"
```

since we have the following checkbox in the delete modal, I believe `blockOwnerDeletion` is redundant here since the user is choosing if to delete the DV or not.

image

PR #7754, opened 2021-01-07
Bug 1913851: sort cluster task in pipeline builder

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5304

**Analysis / Root cause**: 
Task selection dropdown in the pipeline builder contains an unsorted list of items.

**Solution Description**: 
Sort the cluster tasks in the pipeline builder

image

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/TaskListNode.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/TaskListNode.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as _ from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { Button, Flex, FlexItem, FocusTrap } from '@patternfly/react-core';
 import { CaretDownIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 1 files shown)


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/TaskListNode.tsx
```diff
@@ -43,10 +44,13 @@ const TaskListNode: React.FC<TaskListNodeProps> = ({ element, unselectedText })
...
-  const options = [
-    ...namespaceTaskList.map((task) => taskToOption(task, onNewTask)),
-    ...clusterTaskList.map((task) => taskToOption(task, onNewTask)),
-  ];
+  const options = _.sortBy(
+    [
+      ...namespaceTaskList.map((task) => taskToOption(task, onNewTask)),
+      ...clusterTaskList.map((task) => taskToOption(task, onNewTask)),
+    ],
+    (o) => o.label,
```
karthikjeeyar, line 52: You could simply use sort function instead of lodash :) 
```suggestion
  const options =  [
      ...namespaceTaskList.map((task) => taskToOption(task, onNewTask)),
      ...clusterTaskList.map((task) => taskToOption(task, onNewTask)),
    ].sort((a, b) => a.label.localeCompare(b.label))
```

christianvogt: Either or. Neither is wrong.
If we want to impose such restrictions, then we should write lint rules for them. Add it to the nit pile.

andrewballantyne: Definitely my preferred way, but it's effectively the same. If we were not so swamped, I'd have gone for this change.

PR #7755, opened 2021-01-07
Bug 1886154: - System roles are not present while trying to create new role binding through web console

cyril-ui-developer:
/assign @spadgett


spadgett (approved):
/lgtm

PR #7756, opened 2021-01-07
CONSOLE-2521: Migrate Role Binding CRUD tests to Cypress

rebeccaalpert:
Migrated Role Binding tests and removed the old ones from Protractor.

Fixes https://issues.redhat.com/projects/CONSOLE/issues/CONSOLE-2521.


spadgett:
@rebeccaalpert I think it's good to break this up into smaller PRs anyway.

rebeccaalpert:
I'm updating this to reflect that it's a sub-task PR now.

frontend/packages/integration-tests-cypress/tests/crud/rolebindings.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  afterEach(() => {
+    checkErrors();
+  });
+
+  after(() => {
+    cy.deleteProject(testName);
+    cy.logout();
+  });
+
+  it(`displays "Create RoleBinding" page`, () => {
```
spadgett, line 29: It looks like this entire test is covered below in the next test. Is it needed?

frontend/public/components/utils/dropdown.jsx
```diff
@@ -474,6 +474,7 @@ class Dropdown_ extends DropdownMixin {
                       className="pf-c-form-control"
                       onClick={(e) => e.stopPropagation()}
                       data-test-id="dropdown-text-filter"
+                      data-test="dropdown-text-filter"
```
spadgett, line 477: I'd just use the existing legacy test ID for now instead of having two test IDs.

frontend/packages/integration-tests-cypress/tests/crud/rolebindings.spec.ts
rebeccaalpert: I wasn't sure so I left that one alone. Do you think this is ok to drop @dtaylor113?

dtaylor113: yes, it's just a repeat of the test below.l

dtaylor113: yes, i'd drop it, it's a duplicate/subset of the test below.

dtaylor113 (approved):
/lgtm assuming Sam's comments are addressed -thanks

frontend/packages/integration-tests-cypress/tests/crud/rolebindings.spec.ts
rebeccaalpert: Thanks; will drop!

rebeccaalpert:
Updated id and removed duplicate test.

spadgett (approved):
/lgtm

PR #7762, opened 2021-01-08
Bug 1909043: Show an info instead of warning for flexible scaling on internal mode

afreen23:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1909043

Before:
Screenshot from 2021-01-13 16-48-24

After:
Screenshot from 2021-01-13 16-47-24


bipuladh (approved):
/lgtm

PR #7763, opened 2021-01-08
Bug 1913567: Monitoring: Change stacked graph tooltips to not use cumulative values

kyoto:
Change the tooltip to show the series value for both stacked and
non-stacked graphs.


frontend/public/components/monitoring/query-browser.tsx
```diff
@@ -879,7 +880,7 @@ type TooltipProps = {
   activePoints?: { x: number; y: number; _y1?: number }[];
   center?: { x: number; y: number };
   height?: number;
-  style?: { fill: string };
+  style?: { fill: string; name: string };
```
kyoto, line 883: Also noticed that this type definition was missing

spadgett (approved):
/lgtm

PR #7765, opened 2021-01-08
Bug 1883614: Warn user when restarting BMH

rawagner:
Screenshot from 2021-01-08 09-34-06

@andybraren are you ok with the message ? (its the same as non-graceful poweroff)


frontend/packages/metal3-plugin/src/components/modals/RestartHostModal.tsx
```diff
@@ -41,19 +39,12 @@ const RestartHostModal = ({
     [host, close, handlePromise],
   );
 
-  const text = nodeName
-    ? t(
-        'metal3-plugin~The bare metal host {{name}} will be restarted gracefully after all managed workloads are moved.',
```
andybraren, line 46: WDYT of adding back a tiny bit of `<p>` text above the warning that reads:

> The host will be powered off and on again.

andybraren (approved):
One small thing, but otherwise lgtm!

frontend/packages/metal3-plugin/src/components/modals/RestartHostModal.tsx
andybraren: I know it's a bit obvious 🙂 but I think we typically have _something_ in the body of modals explaining what's going to happen before warnings/alerts.

rawagner: sure, added!

jtomasek (approved):

jtomasek (approved):
/lgtm

PR #7766, opened 2021-01-08
Bug 1914183: Link to NAD Quickstarts

rawagner:
Screenshot from 2021-01-08 10-56-27

@matthewcarleton do we want to add some message between Empty state title and actions ?


frontend/packages/kubevirt-plugin/src/components/create-vm/tabs/select-template.tsx
```diff
@@ -243,12 +243,14 @@ export const SelectTemplate: React.FC<SelectTemplateProps> = ({
...
               <Trans t={t} ns="kubevirt-plugin">
-                The virtual machine can be customized in the next step.
+                The virtual machine can be <b>customized</b> in the next step.
               </Trans>
             </div>
             <VMTemplateSupport />
             <div>
-              <b>{t('kubevirt-plugin~Only templates with a valid boot source will be shown.')}</b>
+              <Trans t={t} ns="kubevirt-plugin">
+                Templates that are in <b>error</b> or in <b>progress</b> state will not be shown.
```
matthewcarleton, line 252:
```suggestion
                Templates that are in an <b>error</b> or <b>in-progress</b> state will not be shown.
```

rawagner: fixed

PR #7767, opened 2021-01-08
Bug 1914196: Use route port field for docker container port in Dockerfile import

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4866

**Analysis / Root cause**: 

When using the `from dockerfile` add flow, there are 2 similar fields that are used to specify which port the application is listening on: dockerfile conainer port & route target port
No matter what value is entered, `target port` never persisted anywhere.

**Solution Description**: 
Remove the docker container port, and use only the route target port to create the service and route.

https://user-images.githubusercontent.com/20013884/104005655-7c446480-51cb-11eb-9905-8d47049886b7.mp4

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/utils/shared-submit-utils.ts
+++ b/frontend/packages/dev-console/src/utils/shared-submit-utils.ts
@@ -43,8 +43,8 @@ export const createService = (
 
   let ports = imagePorts;
   const buildStrategy = formData.build?.strategy;
-  if (buildStrategy === 'Docker') {
-    const port = { containerPort: _.get(formData, 'docker.containerPort'), protocol: 'TCP' };
+  if (buildStrategy === 'Docker' && unknownTargetPort) {
+    const port = { containerPort: _.toInteger(unknownTargetPort), protocol: 'TCP' };
     ports = [port];
   } else if (buildStrategy === 'Devfile' && !_.isEmpty(originalService?.spec?.ports)) {
     ports = [
@@ -129,10 +129,9 @@ export const createRoute = (
 
   let targetPort: string;
   const buildStrategy = formData.build?.strategy;
-  if (buildStrategy === 'Docker') {
-    const port = _.get(formData, 'docker.containerPort');
+  if (buildStrategy === 'Docker' && unknownTargetPort) {
     targetPort = makePortName({
-      containerPort: _.toInteger(port),
+      containerPort: _.toInteger(unknownTargetPort),
       protocol: 'TCP',
     });
   } else if (buildStrategy === 'Devfile') {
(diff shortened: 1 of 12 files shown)


frontend/packages/dev-console/src/utils/shared-submit-utils.ts
```diff
@@ -43,8 +43,8 @@ export const createService = (
 
   let ports = imagePorts;
   const buildStrategy = formData.build?.strategy;
-  if (buildStrategy === 'Docker') {
-    const port = { containerPort: _.get(formData, 'docker.containerPort'), protocol: 'TCP' };
+  if (buildStrategy === 'Docker' && unknownTargetPort) {
+    const port = { containerPort: _.toInteger(unknownTargetPort), protocol: 'TCP' };
```
divyanshiGupta, line 47: Instead not adding a port at all maybe we can add the `defaultUnknownPort` here in case there is no `unknownTargetPort`

rottencandy: Updated.

frontend/packages/dev-console/src/utils/shared-submit-utils.ts
```diff
@@ -43,8 +43,9 @@ export const createService = (
 
   let ports = imagePorts;
   const buildStrategy = formData.build?.strategy;
-  if (buildStrategy === 'Docker') {
-    const port = { containerPort: _.get(formData, 'docker.containerPort'), protocol: 'TCP' };
+  if (buildStrategy === 'Docker' && unknownTargetPort) {
```
divyanshiGupta, line 46: Since you haven't removed the check for `unknownTargetPort` so even if there is no `unknownTargetPort`, `defaultUnkownPort` will not be used instead and it beats the purpose of my previous comment.

frontend/packages/dev-console/src/utils/shared-submit-utils.ts
```diff
@@ -43,8 +43,9 @@ export const createService = (
...
+    const containerPort = _.toInteger(unknownTargetPort) || defaultUnknownPort;
```
divyanshiGupta, line 47: I seems that you dont need to add `defaultUnknownPort` here since there is already a condition below which checks if the ports are empty then use `defaultUnknownPort`.

rottencandy: Removed `defaultUnknownPort`. Tested without providing `containerPort` and `defaultUnknownPort` is used if no ports are stated.

rohitkrai03 (approved):
/approve

PR #7774, opened 2021-01-08
Bug 1914309: Redirect user to root when terminal operator is not installed

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5078

**Analysis / Root cause**: Web terminal page was trying to load dev workspace resources even when the terminal operator was not installed. This resulted in showing empty state.

**Solution Description**: Check for dev workspace flag and redirect to root if not available.

https://user-images.githubusercontent.com/6041994/104031309-92b1e680-51f2-11eb-95f0-9d01696693cb.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
@@ -1,11 +1,17 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { InlineTechPreviewBadge } from '@console/shared';
+import { Redirect } from 'react-router';
+import { InlineTechPreviewBadge, useFlag } from '@console/shared';
+import { FLAG_DEVWORKSPACE } from '../../consts';
 import CloudShellTerminal from './CloudShellTerminal';
 import './CloudShellTab.scss';
 
 const CloudShellTab: React.FC = () => {
   const { t } = useTranslation();
+  const devWorkspaceFlag = useFlag(FLAG_DEVWORKSPACE);
+
+  if (!devWorkspaceFlag) return <Redirect to="/" />;
+
   return (
     <>
       <div className="co-cloud-shell-tab__header">
--- a/frontend/packages/console-app/src/components/cloud-shell/__tests__/CloudShellTab.spec.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/__tests__/CloudShellTab.spec.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
+import { Redirect } from 'react-router';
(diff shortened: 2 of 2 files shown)


frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
```diff
@@ -1,11 +1,17 @@
...
+import { history } from '@console/internal/components/utils';
+import { FLAG_DEVWORKSPACE } from '../../consts';
 import CloudShellTerminal from './CloudShellTerminal';
 import './CloudShellTab.scss';
 
 const CloudShellTab: React.FC = () => {
   const { t } = useTranslation();
+  const devWorkspaceFlag = useFlag(FLAG_DEVWORKSPACE);
+
+  if (!devWorkspaceFlag) history.push('/');
```
christianvogt, line 13: I think you should render a `<Redirect` component in this case instead of pushing to history. This will also replace the history instead of adding to the stack.

abhinandan13jan: IMO, this should be handled at the route level and not component level.

rohitkrai03: Done.

frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
```diff
@@ -1,11 +1,17 @@
...
+import { InlineTechPreviewBadge, useFlag } from '@console/shared';
+import { FLAG_DEVWORKSPACE } from '../../consts';
 import CloudShellTerminal from './CloudShellTerminal';
 import './CloudShellTab.scss';
 
 const CloudShellTab: React.FC = () => {
   const { t } = useTranslation();
+  const devWorkspaceFlag = useFlag(FLAG_DEVWORKSPACE);
+
+  if (!devWorkspaceFlag) return <Redirect to="/" />;
```
abhinandan13jan, line 13: is it better to handle this at the Route level instead of the component 
Changing it in the plugin

rohitkrai03: How do you suggest we do this at the route level? This route is in the root level `<App>` component which also renders the main redux store context which contains all the flag information. Do you have a solution for this in mind?

abhinandan13jan: I missed that it was placed beside the '/' at the route level. It is still doable with a manual API fetch for checkng the presence of the CRD using the model, but that would again slow it down I guess
Well, then the `<Redirect/>` is the best approach in this case.

PR #7779, opened 2021-01-08
Bug 1914427: Use kube_running_pod_ready to exclude terminal pods

smarterclayton:
Terminal pods have no cost to the system in most scenarios, so
our dashboards and queries should exclude metrics that include
terminal pods (which `kube_pod_info` includes). The new 4.7 metric
`kube_running_pod_ready` only includes non-terminal pods and is
the appropriate metric to display for counts and joins to used
capacity. On the dashboard this means the pod count reflects the
running state of the system, not completed pods which could be
in the thousands and have minimal overhead.


spadgett (approved):
/lgtm

smarterclayton:
one more change (forgot to save the file)

spadgett (approved):
/lgtm

kyoto:

PR #7783, opened 2021-01-10
Bug 1891189: Stop non-positive values from being entered by users for max disk size and limit

bipuladh:
Use Tooltip to provide basic guidance.
Before: 
Screenshot from 2021-01-11 01-59-03
After:
Screenshot from 2021-01-11 01-58-55
Screenshot from 2021-01-11 01-58-50
Screenshot from 2021-01-11 01-58-43

### DIFF
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
@@ -8,6 +8,7 @@ import {
   TextInputTypes,
   Text,
   TextVariants,
+  Tooltip,
 } from '@patternfly/react-core';
 import { Dropdown } from '@console/internal/components/utils';
 import { ListPage } from '@console/internal/components/factory';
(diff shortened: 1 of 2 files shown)


bipuladh:
@yuvalgalanti PTAL. Are the error info tips good enough?

yuvalgalanti:
looks good @bipuladh, thanks

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
```diff
@@ -212,14 +208,20 @@ export const LocalVolumeSetInner: React.FC<LocalVolumeSetInnerProps> = ({
...
             >
-              <TextInput
-                type={TextInputTypes.text}
-                id="create-lvs-max-disk-size"
-                value={state.maxDiskSize}
-                validated={state.isValidMaxSize ? 'default' : 'error'}
-                className="lso-create-lvs__disk-size-form-group-max-input"
-                onChange={onMaxSizeChange}
-              />
+              <Tooltip
```
cloudbehl, line 211: can we make these validations generic so it can be used with ant text input?

bipuladh: we can take it up as a refactor in the future where we write a generic component for this. For now, let's just fix the bug.

PR #7784, opened 2021-01-10
Bug 1911465: Fix humanize IOPS function

bipuladh:
- Add the missing unit which was dropped initially

Screenshot:
Screenshot from 2021-01-11 02-32-22

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/utilization-card/utils.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/utilization-card/utils.tsx
@@ -9,11 +9,10 @@ import {
 
 export const humanizeIOPS: Humanize = (value) => {
   const humanizedNumber = humanizeNumber(value);
-  const unit = 'IOPS';
   return {
     ...humanizedNumber,
-    string: `${humanizedNumber.value} ${unit}`,
-    unit,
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/components/dashboard-page/storage-dashboard/utilization-card/utils.tsx
```diff
@@ -9,11 +9,10 @@ import {
 
 export const humanizeIOPS: Humanize = (value) => {
   const humanizedNumber = humanizeNumber(value);
-  const unit = 'IOPS';
   return {
     ...humanizedNumber,
-    string: `${humanizedNumber.value} ${unit}`,
-    unit,
+    string: `${humanizedNumber.string} IPOS`,
```
cloudbehl, line 14:
```suggestion
    string: `${humanizedNumber.string} IOPS`,
```

cloudbehl, line 14: Do string and unit both need to have IOPS unit?

bipuladh: yes. Or else the two would not be consistent.
`value + unit === string` ( Format for humanize functions)

PR #7785, opened 2021-01-11
Allow customization in review step

yaacov:
We should not allow VM customization before VM name is defined.

Screenshot:
screenshot-localhost_9000-2021 01 11-09_15_19
screenshot-localhost_9000-2021 01 11-09_14_41
screenshot-localhost_9000-2021 01 11-09_14_59


yaacov:
closing, after off line discussion with @rawagner

PR #7788, opened 2021-01-11
Bug 1914793: HW devices dont need to be translated

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1914793

**Analysis / Root cause**: 
Devices used labelKey without t function

**Solution Description**: 
Using now the value untranslated as devices don't need to be translated.

After:
image
image

Before:
image
image


metalice:
@yaacov please check

PR #7797, opened 2021-01-11
Bug 1905416: Fix sandboxed links in MarkdownView iframe

TheRealJon:
Add `allow-popups-to-escape-sandbox` to MarkdownView iframe `sandbox` attribute so that links with
`target="_blank"` don't inherit the sandbox restrictions.


spadgett (approved):
/lgtm

PR #7798, opened 2021-01-11
Bug 1925562: Add new ArgoCD link from GitOps application environments page

keithchong:
Signed-off-by: Keith Chong <kykchong@redhat.com>

See GitOps 544.  The new link within each card will link to the ArgoCD app list page with a filter-by-label added to show only the apps associated with the environments.

Screenshot:

Screen Shot 2021-01-07 at 2 10 10 PM

For video see,

https://coreos.slack.com/archives/CMP95ST2N/p1610126728487400?thread_ts=1610118978.486800&cid=CMP95ST2N

### DIFF
--- a/frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
+++ b/frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
@@ -2,16 +2,20 @@ import * as React from 'react';
 import * as _ from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { Grid, GridItem, Card, CardBody } from '@patternfly/react-core';
-import { history, ResourceLink } from '@console/internal/components/utils';
+import { history, ResourceLink, ExternalLink } from '@console/internal/components/utils';
+import { K8sResourceKind } from '@console/internal/module/k8s';
 import { GitOpsAppGroupData } from '../utils/gitops-types';
+import { getArgoCDFilteredAppsURI } from '../utils/gitops-utils';
 import './GitOpsListItem.scss';
 
 interface GitOpsListItemProps {
   appGroup: GitOpsAppGroupData;
+  argocdLink?: K8sResourceKind;
 }
 
-const GitOpsListItem: React.FC<GitOpsListItemProps> = ({ appGroup }) => {
+const GitOpsListItem: React.FC<GitOpsListItemProps> = ({ appGroup, argocdLink }) => {
   const { t } = useTranslation();
+  const argocdLinkUrl = getArgoCDFilteredAppsURI(argocdLink.spec.href, appGroup.name);
   const handleCardClick = () => {
     history.push(`/environments/${appGroup.name}?url=${appGroup.repo_url}`);
   };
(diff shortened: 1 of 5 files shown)


keithchong:
Hi @invincibleJai , could you please review this? Let me know if there is something I need to do on my end.

spadgett:

keithchong:
Fixes https://github.com/openshift/console/issues/8068

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
```diff
@@ -2,32 +2,42 @@ import * as React from 'react';
...
-          <GridItem lg={6} md={6} sm={6}>
+          <GridItem lg={5} md={5} sm={5} onClick={handleCardClick}>
             {t('gitops-plugin~{{count, number}} Environment', {
               count: _.size(appGroup.environments),
             })}
           </GridItem>
+          <GridItem lg={2} md={2} sm={2}>
+            {argocdLink && (
+              <ExternalLink
+                href={`${argocdLink.spec.href}/applications?labels=app.kubernetes.io%252Fname%253D${appGroup.name}`}
```
divyanshiGupta, line 36: @keithchong please create a utility function in `gitops-utils.ts` for generating this url as desired. You can see `getPipelinesBaseURI` for reference.

divyanshiGupta (changes requested):

rohitkrai03:
/retitle Bug 1925562: Add new ArgoCD link from GitOps application environments page

keithchong:
Hi @divyanshiGupta , I updated my branch.  I also tested it again and the link still works properly.

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
keithchong: Updated.

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
```diff
@@ -2,32 +2,41 @@ import * as React from 'react';
...
+const GitOpsListItem: React.FC<GitOpsListItemProps> = ({ appGroup, argocdLink }) => {
   const { t } = useTranslation();
+  const argocdLinkUrl = getArgoCDFilteredAppsURI(argocdLink.spec.href, appGroup.name);
   const handleCardClick = () => {
     history.push(`/environments/${appGroup.name}?url=${appGroup.repo_url}`);
   };
 
   return (
-    <Card className="odc-gitops-list-item" onClick={handleCardClick} isHoverable>
+    <Card className="odc-gitops-list-item" isHoverable>
```
rohitkrai03, line 24: Whole card is not clickable anymore because of moving the onClick handler to two grid items only. I know you did the change because the click on argocd link would also result in clicking the card but I think a better way to solve this would be to stop propagation of event in `ExternalLink` component.

rohitkrai03 (changes requested):

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
```diff
@@ -2,32 +2,41 @@ import * as React from 'react';
...
           </GridItem>
-          <GridItem lg={6} md={6} sm={6}>
+          <GridItem lg={5} md={5} sm={5} onClick={handleCardClick}>
             {t('gitops-plugin~{{count, number}} Environment', {
               count: _.size(appGroup.environments),
             })}
           </GridItem>
+          <GridItem lg={2} md={2} sm={2}>
+            {argocdLinkUrl && (
+              <ExternalLink href={`${argocdLinkUrl}`} text={t('gitops-plugin~Argo CD')} />
```
rohitkrai03, line 37: I think you can just pass a prop to this component `stopPropagation` and update the component to add an onClick handler that stopPropagation if this prop is passed.
```suggestion
              <ExternalLink href={`${argocdLinkUrl}`} text={t('gitops-plugin~Argo CD')} stopPropagation />
```

rohitkrai03: In the ExternalLink component you can add onClick handler by passing the below prop to the `<a>` element - `{...(stopPropagation ? { onClick: (e) => e.stopPropagation() } : {})}`

rohitkrai03 (changes requested):

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
keithchong: Thanks Rohit.

frontend/packages/gitops-plugin/src/components/list/GitOpsListItem.tsx
keithchong: Thanks for the suggestion.  This is a useful tip.

rohitkrai03 (approved):
/lgtm
/approve

Tested it locally. Works as expected.

PR #7799, opened 2021-01-12
Bug 1915133: Fix missing default pinned nav items in dev perspective

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5320
https://bugzilla.redhat.com/show_bug.cgi?id=1915133

**Analysis / Root cause**: 
After changing `usePinnedResources` to user settings the hook sets the "default pins" only to the first perspective. This happen because all "pinned resources" are stored in one single object.

```tsx
type PinnedResourcesType = {
  [perspective: string]: string[];
};
```

**Solution Description**: 
Return the default pins from the `Perspective` extension until there is a list of pins defined in user settings for the active perspective.

No UX change.

Added new unit tests for usePinnedResources

```
 PASS  packages/console-shared/src/hooks/__tests__/usePinnedResources.spec.ts
  usePinnedResources
    ✓ returns an empty array if user settings are not loaded yet (31ms)
    ✓ returns no default pins if there are no other pins defined and no extension could be found (2ms)
    ✓ returns some default pins if there are no other pins defined and the extension has default pins (1ms)
    ✓ returns an array of pins saved in user settings for the current perspective (1ms)
```

**Test setup:**

1. Remove the pinned resources from user settings.
2. Reload the console or dev console

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/usePinnedResources.ts
+++ b/frontend/packages/console-shared/src/hooks/usePinnedResources.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, useCallback } from 'react';
 import { isPerspective, Perspective, useExtensions } from '@console/plugin-sdk';
 import { useUserSettingsCompatibility } from './useUserSettingsCompatibility';
 import { PINNED_RESOURCES_LOCAL_STORAGE_KEY } from '../constants';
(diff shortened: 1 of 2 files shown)


jeff-phillips-18 (approved):

frontend/packages/console-shared/src/hooks/usePinnedResources.ts
```diff
@@ -13,27 +13,29 @@ const PINNED_RESOURCES_CONFIG_MAP_KEY = 'console.pinnedResources';
...
-    (pr: string[]) => {
-      setPinnedResources((prevPR) => ({ ...prevPR, [activePerspective]: pr }));
+  >(PINNED_RESOURCES_CONFIG_MAP_KEY, PINNED_RESOURCES_LOCAL_STORAGE_KEY, {}, true);
+
+  const pins = useMemo(() => {
+    if (!loaded) {
+      return [];
+    }
+    if (pinnedResources?.[activePerspective]) {
+      return pinnedResources?.[activePerspective];
```
invincibleJai, line 25:
```suggestion
      return pinnedResources[activePerspective];
```

invincibleJai:
Thanks @jerolimov , verified the changes. Works as expected

frontend/packages/console-shared/src/hooks/usePinnedResources.ts
jerolimov: Removed both chars :D

jeff-phillips-18 (approved):

rohitkrai03 (approved):
/approve

PR #7801, opened 2021-01-12
Bug 1915176: Update snapshot API CRDs to v1 in web-console when creating volumesnapshot related resource

jhadvig:
Those resoruces are: 
- VolumeSnapshots
- VolumeSnapshotClasses
- VolumeSnapshotContents

/assign @spadgett

### DIFF
--- a/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
+++ b/frontend/packages/integration-tests-cypress/mocks/snapshot.ts
@@ -74,7 +74,7 @@ export const PVC = {
 };
 
 export const SnapshotClass = {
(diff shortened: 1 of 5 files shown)


spadgett (commented):
Are there any API changes from v1beta1 to v1 that we need to account for?

TheRealJon (approved):

spadgett:

spadgett:
> Are there any API changes from v1beta1 to v1 that we need to account for?

I looked at the API, and I don't see any changes that would break our templates. @jhadvig Can you rebase?

spadgett (approved):
/lgtm

PR #7802, opened 2021-01-12
Bug 1915277: [e2e][automation]fix cdi upload form test

glekner:


glekner:
@gouyang I believe tests should work now, please test in your env to confirm.

PR #7807, opened 2021-01-12
Bug 1911307:  scale pods using the 'scale' subresource

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1911307

### DIFF
--- a/frontend/packages/console-shared/src/components/pod/PodRing.tsx
+++ b/frontend/packages/console-shared/src/components/pod/PodRing.tsx
@@ -54,7 +54,8 @@ const PodRing: React.FC<PodRingProps> = ({
   const handleScaling = _.debounce(
     (operation: number) => {
       const patch = [{ op: 'replace', path, value: operation }];
-      const promise: Promise<K8sResourceKind> = k8sPatch(resourceKind, obj, patch);
+      const opts = { path: 'scale' };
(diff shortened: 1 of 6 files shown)


frontend/packages/console-shared/src/components/pod/PodRing.tsx
```diff
@@ -54,7 +54,8 @@ const PodRing: React.FC<PodRingProps> = ({
   const handleScaling = _.debounce(
     (operation: number) => {
       const patch = [{ op: 'replace', path, value: operation }];
-      const promise: Promise<K8sResourceKind> = k8sPatch(resourceKind, obj, patch);
+      const opts = { path: 'scale' };
```
spadgett, line 57: You'll need to update the pod ring access check as well to use the `scale` subresource.

spadgett (commented):
Thanks @rhamilto 👍

frontend/packages/console-shared/src/components/pod/PodRing.tsx
rhamilto: This right?  https://github.com/openshift/console/pull/7807/commits/335306b2a232e861d5b88c83d9c6c9e033e88d59

spadgett: @rhamilto Looks right

spadgett (approved):
/lgtm

PR #7810, opened 2021-01-12
Bug 1888554: Improve help text for label selectors on PVC creation form

TheRealJon:
Improve help text for label selectors on PVC creation form


spadgett (approved):
/lgtm

PR #7813, opened 2021-01-13
Bug 1913608: Query Browser: Fix bug where `None` table values were not cleared

kyoto:
We were modifying `result` instead of creating a copy which caused the
added `None` table entries to persist when they shouldn't.


spadgett (approved):
/lgtm

PR #7816, opened 2021-01-13
Bug 1915647: adds check for drophint for eventSource connector

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5328

**Analysis / Root cause**: 
intermittent NPE observed on drag and drop of eventSource connector

**Solution Description**: 
added check for data in drophint and for annotations

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
After the fix, I am not able to reproduce the issue by moving the pub-sub connector over the KSVC.
/lgtm

PR #7820, opened 2021-01-13
Bug 1915685: Tweak message for not supported template

rawagner:
Screenshot from 2021-01-13 09-46-59


rawagner:
/hold

lets wait for @matthewcarleton approval

frontend/packages/kubevirt-plugin/src/components/modals/support-modal/support-modal.tsx
```diff
@@ -46,14 +49,17 @@ const SupportModal: React.FC<SupportModalProps> = ({ onConfirm, close, community
           ) : (
             <>
               <StackItem>
-                {t('kubevirt-plugin~This template does not have a support statement defined.')}
+                {t(
+                  'kubevirt-plugin~The support level is not defined in the template yaml. To mark this template as supported, add these two annotations to the template yaml:',
```
yaacov, line 53:
```suggestion
                  'kubevirt-plugin~The support level is not defined in the template yaml. To mark this template as supported, add these two annotations to the template:',
```
the reason for removing the explicit YAML suggestion is because we can edit annotations also from details view

matthewcarleton: could we remove YAML and replace it with "add these two annotations in the template details"?

rawagner: msg updated

rawagner:
/hold cancel
@yaacov ptal

PR #7823, opened 2021-01-13
Bug 1911212: Monitoring dashboards: Support $__auto_interval_* variables

kyoto:
Treat all `$__auto_interval_*` variables as identical to `$__interval` and `$__rate_interval`.

With the current settings, this means that `$__interval`, `$__rate_interval` and all `$__auto_interval_*` variables will be set to either ("Time Range" / 30) or to 5 minutes, whichever is larger.


spadgett (approved):
/lgtm

PR #7829, opened 2021-01-13
Bug 1867400: Not allow creation of second storagecluster

afreen23:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867400
Screenshot from 2021-01-13 22-56-41


frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+export const ExistingClusterModal: React.FC<ExistingClusterModalProps> = ({
+  match,
+  clusterName,
+  hasStorageCluster,
+}) => {
+  const {
+    params: { ns, appName },
+  } = match;
+  const { t } = useTranslation();
+  const [isOpen, setIsOpen] = React.useState(false);
```
cloudbehl, line 21:
```suggestion
  const [isOpen, setIsOpen] = React.useState(hasStorageCluster);
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -136,6 +144,11 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
       {mode === MODES.ATTACHED_DEVICES && (
         <CreateAttachedDevicesCluster match={match} mode={mode} />
       )}
+      <ExistingClusterModal
+        match={match}
+        clusterName={sc?.items?.[0]?.metadata?.name}
```
cloudbehl, line 149: clusterName should be filtered out based on the item which has status != ignored.
Check this:
https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/features.ts#L116..L119

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: I have changed a bit of code now.
This suggestion does not holds now.PTAL again.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -35,21 +44,13 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
-  const csvResource = {
-    kind: referenceForModel(ClusterServiceVersionModel),
-    name: appName,
-    namespace: ns,
-    isList: false,
-  };
-
   const [csv, csvLoaded, csvError] = useK8sWatchResource<ClusterServiceVersionKind>(csvResource);
   const [infra, infraLoaded, infraError] = useK8sGet<any>(InfrastructureModel, 'cluster');
+  const [sc, scLoaded, scLoadError] = useK8sGet<ListKind<StorageClusterKind>>(
```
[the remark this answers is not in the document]
afreen23: sure, sounds good!

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
afreen23: We want to block cluster creation , even if a single item is present.
And this condition is only valid to check  when someone fires from CLI a cluster when another cluster is already present. It seems an edge case and i dropped it but no harm in putting a check there !!

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  storageCluster,
+}) => {
+  const {
+    params: { ns, appName },
+  } = match;
+  const { t } = useTranslation();
+  const [isOpen, setIsOpen] = React.useState(true);
+
+  const clusterName = storageCluster.items.find((sc) => sc.status.phase !== 'Ignored').metadata
+    .name;
```
bipuladh, line 24:
```suggestion
  const clusterName = storageCluster.items.find((sc) => sc.status.phase !== 'Ignored')?.metadata
   ?.name;
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const ExistingClusterModal: React.FC<ExistingClusterModalProps> = ({
+  match,
+  storageCluster,
+}) => {
+  const {
+    params: { ns, appName },
+  } = match;
+  const { t } = useTranslation();
+  const [isOpen, setIsOpen] = React.useState(true);
```
bipuladh, line 21:
```suggestion
  const [isOpen, setOpen] = React.useState(true);
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+    history.push(storageClusterPath);
+  };
+
+  return (
+    <Modal
+      title={t('ceph-storage-plugin~Storage Cluster exists')}
+      titleIconVariant="warning"
+      isOpen={isOpen}
+      variant="small"
+      onClose={() => history.goBack()}
```
bipuladh, line 46:
```suggestion
      onClose={() => resourcePathFromModel(ClusterServiceVersionModel, appName, ns)}
```
Let's return them back to the OLM page no matter how they reach here.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: This component will be called only when a cluster is present

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: What do you mean by OLM page ?
Its returning to Installed Operator's page.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: I prefer full names :)

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
bipuladh: what if I come here through a URL link? wouldn't we go back to the last url?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { match as RouteMatch } from 'react-router';
+import { Button } from '@patternfly/react-core';
+import { Modal } from '@console/shared/src/components/modal';
+import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager/src';
+import { history, resourcePathFromModel } from '@console/internal/components/utils';
+import { ListKind, referenceForModel } from '@console/internal/module/k8s';
+import { StorageClusterKind } from '../../types';
+import { OCSServiceModel } from '../../models';
+
+export const ExistingClusterModal: React.FC<ExistingClusterModalProps> = ({
```
bipuladh, line 13: Can we make this a default export of this file?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <Modal
+      title={t('ceph-storage-plugin~Storage Cluster exists')}
+      titleIconVariant="warning"
+      isOpen={isOpen}
+      variant="small"
+      onClose={() => history.goBack()}
+      isFullScreen={false}
+      actions={[
+        <Button key="confirm" variant="primary" onClick={onConfirm}>
```
bipuladh, line 49: aria-label for the button

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,69 @@
...
+          {t('ceph-storage-plugin~Back to operator page')}
+        </Button>,
+        <Button key="cancel" variant="link" onClick={onCancel}>
```
bipuladh, line 52: aria-label for the button

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: yes, on closing the icon it will go and that's what we need to do. Its on closing the modal.
If you want to go to operator's page then you can click on "Go to operator's page"
Back to OLM page is explicit here. Refer the screenshot

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: There is text in the button so it will be used.
Adding aria-label will be redundant here, since both the text will be remain same. aria-label is good to add where we dont have clue of what button is doing like <button class="burger"></button> or <button>+</button>

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: Same as above https://github.com/openshift/console/pull/7829/#discussion_r569971170

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: I used named because I want the name to be consistent all over.
Do you think we should make it default considering its  a single component ?

bipuladh: yes

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
bipuladh: `history.goBack()` will refer to the last url, if I am not mistaken, so if I reach here from say the dashboard pages by changing the url. Wouldn't closing it lead me back to dashboard?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
bipuladh: makes sense. thanks

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: I have updated but as a general rule I prefer to do named exports irrespective of a single file.
It messes with typos and intellisense, automcomplete. If this component want to be reused then anyone can export like `ExistingClusterModal  from './file'` or `existingClusterModal from './file'`

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const onConfirm = () => {
+    setIsOpen(false);
+    history.push(resourcePathFromModel(ClusterServiceVersionModel, appName, ns));
+  };
+
+  const onCancel = () => {
+    setIsOpen(false);
+    history.push(storageClusterPath);
+  };
```
afreen23, line 35: @bipuladh here we are setting url to OLM page

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
afreen23: @bipuladh here we are setting url to go back to the last page.
And here we are setting to operator and cluster page. see https://github.com/openshift/console/pull/7829/files#r569983381
A total of 3 urls.

bipuladh: Exactly my concern. If the user closes it, do we really want to send them to the last URL or to the OLM details page for OCS Operator?

afreen23: Yes we want to send them to last url.
If he wants to go to operator page, he will choose operator page

afreen23:
I have removed close button due to the concerns on redirection @bipuladh @cloudbehl @yuvalgalanti .
This wont be an issue since we have the respective urls for operator page and cluster page redirection present.
Hope this solves for all :)

PR #7832, opened 2021-01-13
Bug 1915939: Resizing the browser window removes Web Terminal Icon

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5313

**Analysis / Root cause**: 
Web terminal button missing from mobile menu.

**Solution Description**: 
Adds the web terminal action to the mobile menu.

Refactored redux action component, `CloudShellMastheadAction`, and refactored `CloudShellMastheadButton` to reuse the same hook and redux state.

Added a new redux state `isActive` which captures when the terminal window is actually present. Works in conjunction with `isExpanded` which denotes the drawer is opened. This was necessary to know when the close confirmation model was required to prompt the user. Previously even if the setup form was opened we prompted the user to confirm ending their session even though no session was present. With this new state, I was able to reconcile all close actions into one.

@openshift/team-devconsole-ux 

When the terminal is closed:
image

When the terminal is open:
image

Added unit tests for cloud-shell redux reducers, actions, and selectors.
Added unit tests for `useCloudShellAvailable` hook.
image

**Test setup:**
Install web terminal operator from operator hub.

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/CloudShell.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/CloudShell.tsx
@@ -1,11 +1,9 @@
 import * as React from 'react';
-import { Dispatch } from 'redux';
 import { connect } from 'react-redux';
 import { RootState } from '@console/internal/redux';
 import { WithFlagsProps, connectToFlags } from '@console/internal/reducers/features';
-import { isCloudShellExpanded } from '../../redux/reducers/cloud-shell-reducer';
+import { isCloudShellExpanded } from '../../redux/reducers/cloud-shell-selectors';
 import { toggleCloudShellExpanded } from '../../redux/actions/cloud-shell-actions';
-import cloudShellConfirmationModal from './cloudShellConfirmationModal';
 import CloudShellDrawer from './CloudShellDrawer';
 import CloudShellTerminal from './CloudShellTerminal';
 import { FLAG_DEVWORKSPACE } from '../../consts';
@@ -24,9 +22,8 @@ const CloudShell: React.FC<CloudShellProps> = ({ flags, open, onClose }) => {
   if (!flags[FLAG_DEVWORKSPACE]) {
     return null;
   }
-  const toggleWithModal = () => cloudShellConfirmationModal(onClose);
   return open ? (
-    <CloudShellDrawer onClose={toggleWithModal}>
+    <CloudShellDrawer onClose={onClose}>
       <CloudShellTerminal onCancel={onClose} />
     </CloudShellDrawer>
   ) : null;
@@ -36,7 +33,7 @@ const stateToProps = (state: RootState): StateProps => ({
   open: isCloudShellExpanded(state),
 });
 
-const dispatchToProps = (dispatch: Dispatch): DispatchProps => ({
+const dispatchToProps = (dispatch): DispatchProps => ({
   onClose: () => dispatch(toggleCloudShellExpanded()),
 });
 
--- a/frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { connect } from 'react-redux';
+import { connect, Dispatch } from 'react-redux';
 import { Base64 } from 'js-base64';
 import { useTranslation } from 'react-i18next';
 import { connectToFlags, WithFlagsProps } from '@console/internal/reducers/features';
@@ -8,6 +8,8 @@ import { FLAGS } from '@console/shared';
 import { WSFactory } from '@console/internal/module/ws-factory';
 import { resourceURL } from '@console/internal/module/k8s';
 import { PodModel } from '@console/internal/models';
+import { Button, EmptyState, EmptyStateBody } from '@patternfly/react-core';
+import { setCloudShellActive } from '../../redux/actions/cloud-shell-actions';
 import Terminal, { ImperativeTerminalType } from './Terminal';
 import TerminalLoadingBox from './TerminalLoadingBox';
 import useActivityTick from './useActivityTick';
@@ -17,7 +19,7 @@ import {
   startWorkspace,
   CloudShellResource,
 } from './cloud-shell-utils';
-import { Button, EmptyState, EmptyStateBody } from '@patternfly/react-core';
+
 import './CloudShellExec.scss';
 
 // pod exec WS protocol is FD prefixed, base64 encoded data (sometimes json stringified)
@@ -40,7 +42,11 @@ type StateProps = {
   };
 };
 
-type CloudShellExecProps = Props & StateProps & WithFlagsProps;
+type DispatchProps = {
+  onActivate: (active: boolean) => void;
+};
+
+type CloudShellExecProps = Props & DispatchProps & StateProps & WithFlagsProps;
 
 const NO_SH =
   'starting container process caused "exec: \\"sh\\": executable file not found in $PATH"';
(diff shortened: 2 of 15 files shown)


bgliwa01 (approved):
LGTM! thanks Christian!

frontend/packages/console-app/src/components/cloud-shell/__tests__/useCloudShellAvailable.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    checkTerminalAvailable: jest.fn(() => Promise.resolve()),
+  };
+});
+
+jest.mock('@console/shared', () => {
+  const originalModule = (jest as any).requireActual('@console/shared');
+  return {
+    ...originalModule,
+    useFlag: jest.fn(() => true),
```
andrewballantyne, line 20: I'm flip-flopping on making this comment... but I think it's better to not have a default for your methods. 

You're mocking them when you need to specify a sane test (all parts included). So if you have no default returns for these methods, a lack of a mock should fail due to no explicit definition of the return type (it'll get back `undefined` or never resolve)... which I think helps make the tests whole, if more are written in the future.

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
+import { Dispatch } from 'react-redux';
+import {
+  setCloudShellExpanded,
+  setCloudShellActive,
+  Actions,
+  toggleCloudShellExpanded,
+} from '../cloud-shell-actions';
+import cloudShellConfirmationModal from '../../../components/cloud-shell/cloudShellConfirmationModal';
```
andrewballantyne, line 8: Nit: Sort.

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        isActive: false,
+      },
+    });
+    expect(setCloudShellActive(true)).toEqual({
+      type: Actions.SetCloudShellActive,
+      payload: {
+        isActive: true,
+      },
+    });
+  });
```
andrewballantyne, line 49: I don't think these are great tests 🤔 

For the most part you're testing `action` and static values. The _real_ test here is that you set `isActive` or `isExpanded` to the provided value. If we change the action in the future to have another property, this would fail and just need to be updated. If we changed the type, it's a constant to a constant change... so it would just need to be updated. If we changed the logic around setting `isExpanded` or `isActive` based on the provided value, that **does** have value as it changed the expected output against the signature of the method. ie, "I once provided `true`, but now you're consuming it as a number? A string? Inverting it? I didn't expect to that!". Whereas everything else mentioned is not impacting the signature => expected output of the action, more so just adding to it (around the initial expected output).

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    const dispatch = jest.fn<Dispatch>();
+    // initial isExpanded state is false
+    const state = { plugins: { console: { cloudShell: { isExpanded: false } } } } as any;
+    toggleCloudShellExpanded()(dispatch, () => state);
+    expect(dispatch).toHaveBeenCalledWith({
+      type: Actions.SetCloudShellExpanded,
+      payload: {
+        isExpanded: true,
+      },
+    });
```
andrewballantyne, line 61: To my previous comment, I think we can improve this so it doesn't test `action` and avoids the "extra key" manual test update.

I created this helper method to simplify the usage because Jest doesn't have "partial nested" matches in function returns.
```
const payloadContaining = (v: Object) => expect.objectContaining({ payload: expect.objectContaining(v) });
```
```suggestion
    expect(dispatch).toHaveBeenCalledWith(payloadContaining({ isExpanded: true }));
```

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  it('should thunk dispatch toggle expand false action without confirmation', () => {
+    const dispatch = jest.fn<Dispatch>();
+    // initial isExpanded state is true but isActive is false
+    const state = {
+      plugins: { console: { cloudShell: { isExpanded: true, isActive: false } } },
+    } as any;
+    toggleCloudShellExpanded()(dispatch, () => state);
+    expect(dispatch).toHaveBeenCalledWith({
```
andrewballantyne, line 71: Same change here

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  it('should thunk dispatch toggle expand false action with confirmation', async () => {
+    const dispatch = jest.fn<Dispatch>();
+    // initial isExpanded state is true and isActive is true
+    const state = {
+      plugins: { console: { cloudShell: { isExpanded: true, isActive: true } } },
+    } as any;
+    await toggleCloudShellExpanded()(dispatch, () => state);
+    expect(dispatch).toHaveBeenCalledWith({
```
andrewballantyne, line 88: Same here

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(cloudShellConfirmationModal).toHaveBeenCalled();
+  });
+
+  it('should thunk dispatch toggle expand true action', () => {
+    const dispatch = jest.fn<Dispatch>();
+    // initial isExpanded state is false
+    const state = { plugins: { console: { cloudShell: { isExpanded: false } } } } as any;
+    toggleCloudShellExpanded()(dispatch, () => state);
+    expect(dispatch).toHaveBeenCalledWith({
```
andrewballantyne, line 103: Same here

andrewballantyne (changes requested):
Nits and some test improvements.

andrewballantyne:
🤔  You also failed a test for not running `yarn i18n` -- which is odd, because you didn't update the text? 🤔

frontend/packages/console-app/src/redux/actions/__tests__/cloud-shell-actions.spec.ts
christianvogt: Discussed with Andrew and we decided that using `expect.objectContaining` will suffice.

frontend/packages/console-app/src/components/cloud-shell/CloudShell.tsx
```diff
@@ -36,7 +33,7 @@ const stateToProps = (state: RootState): StateProps => ({
   open: isCloudShellExpanded(state),
 });
 
-const dispatchToProps = (dispatch: Dispatch): DispatchProps => ({
```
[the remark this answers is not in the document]
christianvogt: Redux dispatch typescript type doesn't support thunks. Need to investigate. Also we don't include redux-thunk. Instead console has a _copy_ of redux-thunk.

frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
```diff
@@ -17,7 +18,7 @@ import {
   startWorkspace,
   CloudShellResource,
 } from './cloud-shell-utils';
-import { Button, EmptyState, EmptyStateBody } from '@patternfly/react-core';
+import { setCloudShellActive } from '../../redux/actions/cloud-shell-actions';
```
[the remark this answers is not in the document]
christianvogt: How about that eslint rule we disabled :P

andrewballantyne (approved):
/lgtm

PR #7835, opened 2021-01-13
Fix broken olm cypress tests and add them to CI script.

TheRealJon:
Fix broken olm cypress tests and add them to CI script.


dtaylor113 (approved):
/lgtm

spadgett (approved):
/lgtm

spadgett:
@TheRealJon It looks like some of the OLM tests are running in the wrong perspective likely causing them to fail. It could be related to the user settings change to remember the last perspective.

spadgett:
@christianvogt It seems create operand in the admin perspective is switching the active perspective to dev. These tests are running in CI, so no one else should be using the cluster. Is this a bug?

spadgett (approved):
/lgtm

PR #7842, opened 2021-01-14
Add tests for VM template source and VM empty state

rawagner:


rawagner:
@gouyang please take a look

frontend/packages/kubevirt-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  cy.exec(
+    `kubectl delete --ignore-not-found=${ignoreNotFound} -n ${resource.metadata.namespace} --cascade ${kind} ${resource.metadata.name}`,
+  );
+});
+
+Cypress.Commands.add('createResource', (resource) => {
+  cy.exec(`echo '${JSON.stringify(resource)}' | kubectl create -f -`);
+});
+
+Cypress.Commands.add('createDataVolume', (name: string, namespace: string) => {
```
gouyang, line 29: Can we put the dv json in fixture and just call `createResource` to do it?

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
```diff
@@ -0,0 +1,35 @@
+import { ProvisionSource } from '../../integration-tests/tests/utils/constants/enums/provisionSource';
```
gouyang, line 1: I prefer copying the file into `integration-tests-cypress` directory, let's don't have dependencies on the old tests, wdyt?

frontend/packages/kubevirt-plugin/src/components/vm-templates/vm-template-source.tsx
```diff
@@ -89,6 +89,7 @@ const DeleteSourceButton: React.FC<DeleteSourceButtonProps> = ({ template, sourc
           isInline
           variant={ButtonVariant.link}
           onClick={() => createDeleteSourceModal({ sourceStatus })}
+          data-test="delete-template-source"
```
gouyang, line 92: Like this.

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
```diff
@@ -0,0 +1,35 @@
...
+  [ProvisionSource.URL.getValue()]: 'input[id="provision-source-url"]',
+  [ProvisionSource.CONTAINER.getValue()]: 'input[id="provision-source-container"]',
+};
+
+type DiskSourceOpts = {
+  pvcName: string;
+  pvcNamespace: string;
+};
+
+export const addSource = {
```
gouyang, line 13: It needs to take storageClass/accessMode/volueMode into considerations here, in the old tests we extract the values from configMap `KUBEVIRT_STORAGE_CLASS_DEFAULTS`, for simplicity I prefer to just use consts to define them for each storageClass. Either way works for me.

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
rawagner: ok, I agree to not have deps on old tests. But I also dont like to just copy the file. I created a new directory `integration-tests-utils` where we can put pieces which are meant to be shared between new and old tests - and moved `ProvisionSource` there.

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
rawagner: UI already sets the correct values. We dont need to change them in tests.

frontend/packages/kubevirt-plugin/integration-tests-cypress/support/index.ts
rawagner: moved data volume to fixture. But I still need this command because I need to change the default name/namespace/access and volume modes. The command however uses `createResource`

frontend/packages/kubevirt-plugin/integration-tests-utils/enums/provisionSource.ts
```diff
@@ -10,13 +10,15 @@ export class ProvisionSource extends ObjectEnum<string> {
   static readonly CONTAINER = new ProvisionSource(
     'Container',
     'Import via Registry (creates PVC)',
-    'quay.io/kubevirt/fedora-cloud-container-disk-demo:latest',
+    'quay.io/kubevirt/cirros-container-disk-demo:latest',
```
rawagner, line 13: I changed the default container to cirros to make it smaller/faster

rawagner:
@gouyang please take a look again

frontend/packages/kubevirt-plugin/integration-tests-utils/enums/provisionSource.ts
gouyang: We're using cirros long time ago and change it to fedora because some tests configured with cloud-init need to use fedora image, like guest-agent tests. 
How about let's have both of them and only use Fedora when necessary?

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
gouyang: Then let's add a step to change the storageClass from the environment variable.

frontend/packages/kubevirt-plugin/integration-tests-cypress/tests/vm-template/source-image.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    cy.login();
+    cy.visit('');
+    cy.createProject(testName);
+  });
+
+  beforeEach(() => {
+    virtualization.templates.visit();
+    cy.deleteResource({
+      kind: 'DataVolume',
+      metadata: { name: 'rhel6.9', namespace: 'kubevirt-os-images' },
```
gouyang, line 19: `kubevirt-os-images` is `openshift-virtualization-os-images` in downstream, not sure what's the best way to address it, it at least can be a const on top of the tests, wdyt?

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/add-source.ts
rawagner: Ok, you can specify `STORAGE_CLASS` env variable now.

frontend/packages/kubevirt-plugin/integration-tests-cypress/tests/vm-template/source-image.spec.ts
rawagner: added env variable `KUBEVIRT_OS_IMAGES_NS` which is set to `openshift-virtualization-os-images` by default

frontend/packages/kubevirt-plugin/integration-tests-cypress/fixtures/data-volume.json
```diff
@@ -0,0 +1,27 @@
...
+  "kind": "DataVolume",
+  "metadata": {
+    "name": "foo",
+    "namespace": "bar",
+    "annotations": {
+      "cdi.kubevirt.io/storage.bind.immediate.requested": "true"
+    }
+  },
+  "spec": {
+    "pvc": {
```
gouyang, line 12: It needs to add storageClassName for the PVC to accept STORAG_CLASS from environment variable.

rawagner: fixed

.gitignore
```diff
@@ -19,6 +19,7 @@
 /frontend/packages/integration-tests-cypress/cypress-a11y-report.json
 /frontend/packages/dev-console/integration-tests/cypress-a11y-report.json
 /frontend/packages/ceph-storage-plugin/integration-tests-cypress/cypress-a11y-report.json
+/frontend/packages/kubevirt-plugin/integration-tests-cypress/cypress-a11y-report.json
```
spadgett, line 22: I don't think this is needed after #7916.

frontend/package.json
```diff
@@ -37,6 +37,8 @@
     "test-cypress-devconsole-headless": "cd packages/dev-console/integration-tests && node --max-old-space-size=4096 ../../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless --spec \"features/project-creation.feature\";",
     "test-cypress-ceph": "cd packages/ceph-storage-plugin/integration-tests-cypress && ../../../node_modules/.bin/cypress open --config-file cypress-ceph.json --env openshift=true",
     "test-cypress-ceph-headless": "cd packages/ceph-storage-plugin/integration-tests-cypress && node --max-old-space-size=4096 ../../../node_modules/.bin/cypress run --config-file cypress-ceph.json --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
+    "test-cypress-kubevirt": "cd packages/kubevirt-plugin/integration-tests-cypress && ../../../node_modules/.bin/cypress open --env openshift=true",
+    "test-cypress-kubevirt-headless": "cd packages/kubevirt-plugin/integration-tests-cypress && node --max-old-space-size=4096 ../../../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
```
spadgett, line 41: @dtaylor113 We should investigate a way to add these tests that avoids two need commands in the package.json file every time. Maybe a script that takes a parameter or env var.

Nothing to change in this PR, just an observation.

test-prow-e2e.sh
```diff
@@ -49,4 +49,6 @@ elif [ "$SCENARIO" == "olmFull" ]; then
   ./test-cypress.sh -p olm -h true
 elif [ "$SCENARIO" == "ceph" ]; then
   ./test-cypress.sh -p ceph -h true
+elif [ "$SCENARIO" == "kubevirt-gating" ]; then
+  ./test-cypress.sh -p kubevirt -h true
```
spadgett, line 53: I assume you want to skip the Protractor tests as well (line 37 above).

frontend/package.json
dtaylor113: Hi @spadgett, created https://issues.redhat.com/browse/CONSOLE-2538 to track

.gitignore
rawagner: thats right, removed

test-prow-e2e.sh
rawagner: No, we still want to execute proctator tests too. We will remove them once we rewrite protractor gating suite to cypress.

spadgett:

frontend/packages/kubevirt-plugin/integration-tests-cypress/plugins/index.js
```diff
@@ -0,0 +1,63 @@
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+const wp = require('@cypress/webpack-preprocessor');
+const fs = require('fs');
+
+module.exports = (on, config) => {
```
spadgett, line 6: @dtaylor113 I assume all plugins are doing this. Is there a way we can keep the config in place instead of copying it across plugins?

spadgett (approved):
Hey @rawagner. Changes look good to me, but I will defer to @gouyang for final review.

PR #7844, opened 2021-01-14
add Andrew Ballantyne as an approver

christianvogt:
Andrew is the the team lead for the dev-console.

cc @spadgett


spadgett (approved):
/lgtm

PR #7846, opened 2021-01-15
Bug 1915995: - "Edit RoleBinding Subject" action under RoleBinding list page kebab actions causes unhandled exception

cyril-ui-developer:
/assign @spadgett

### DIFF
--- a/frontend/public/components/RBAC/bindings.jsx
+++ b/frontend/public/components/RBAC/bindings.jsx
@@ -397,7 +397,7 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
       'subjects',
     ]);
     existingData.kind = props.kind;
-    const { subjectKind, subjectName } = this.props.fixed.subjectRef;
+    const { subjectKind, subjectName } = this.props.fixed.subjectRef || {};
     const data = _.defaultsDeep({}, props.fixed, existingData, {
       apiVersion: 'rbac.authorization.k8s.io/v1',
       kind: 'RoleBinding',
@@ -500,6 +500,7 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
     const { fixed, saveButtonText, t } = this.props;
     const RoleDropdown = kind === 'RoleBinding' ? NsRoleDropdown : ClusterRoleDropdown;
     const title = `${this.props.titleVerb} ${kindObj(kind).label}`;
+    const isSubjectDisabled = fixed?.subjectRef?.subjectName ? true : false;
     const bindingKinds = [
       {
         value: 'RoleBinding',
@@ -517,9 +518,9 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
       },
     ];
     const subjectKinds = [
-      { value: 'User', title: t('bindings~User') },
-      { value: 'Group', title: t('bindings~Group') },
-      { value: 'ServiceAccount', title: t('bindings~ServiceAccount') },
+      { value: 'User', title: t('bindings~User'), disabled: false },
+      { value: 'Group', title: t('bindings~Group'), disabled: false },
+      { value: 'ServiceAccount', title: t('bindings~ServiceAccount'), disabled: false },
     ];
 
     return (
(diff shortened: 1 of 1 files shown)


frontend/public/components/RBAC/bindings.jsx
```diff
@@ -419,7 +419,7 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
       'subjects',
     ]);
     existingData.kind = props.kind;
-    const { subjectKind, subjectName } = this.props.fixed.subjectRef;
+    const { subjectKind, subjectName } = props.fixed.subjectRef || {};
```
spadgett, line 422: Should we expect `fixed` to always have a value?

cyril-ui-developer: @spadgett Yes

spadgett: @cyril-ui-developer So I just tested this, and I get an error even when I edit a role binding that has a subject. I don't think the problem is the null guard... It seems that `subjectRef` is not being set when it should have a value.

cyril-ui-developer: @spadgett The `subjectRef` was introduced for User and Group context. There is a  `subjects` object that is set for Role Binding and project pages. I have removed the `subjectRef` and bound the values with the `subjects` object. PTAL

frontend/public/components/RBAC/bindings.jsx
```diff
@@ -432,13 +431,12 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
       subjects: [
         {
           apiGroup: 'rbac.authorization.k8s.io',
-          kind: subjectKind || 'User',
-          name: subjectName || '',
+          kind: 'User',
+          name: '',
```
spadgett, line 435: I guess I'm unclear on the fix. Why can't be initialize the value to the subject that's currently set in the RoleBinding?

cyril-ui-developer: @spadgett The values are not available if the context is from Group and User pages. 
This is the PR that set Group and User values - https://github.com/openshift/console/pull/7546 using query parameters.

cyril-ui-developer: > I guess I'm unclear on the fix. Why can't be initialize the value to the subject that's currently set in the RoleBinding?

The Edit and Duplicate binding makes use of `BindingLoadingWrapper` component and it is connected to `Firehose` and Redux store.  So the subjects is from the store. The Create RoleBinding subjects is using the initialized values  `   
 {
          apiGroup: 'rbac.authorization.k8s.io',
          kind: 'User',
          name: '',
        }`
        
        The exception occurred because Edit and Duplicate doesn't have defined `subjectRef` from  query parameters. The subjects comes from the store and get set in this logic:
        [](https://github.com/openshift/console/blob/master/frontend/public/components/RBAC/bindings.jsx#L423)
        
     To get the Group and User context for Create Role Binding, I introduced `subjectRef` and used it in the subjects initialized in RoleBinding page and handled the null check to prevent Edit and Duplicate from exception. 
 
 I'm thinking that the previous submitted change described above looks clearer. 
 
 Another option might be to rewrite Create Binding component to make use of Redux store and Firehose.
 
 Any thoughts?

spadgett:

cyril-ui-developer:
@spadgett Reverted to the previous PR and added logic to disable the subject type if the context is User and Group. 
PTAL. Thanks

frontend/public/components/RBAC/bindings.jsx
```diff
@@ -500,6 +500,7 @@ class BaseEditRoleBindingWithTranslation extends React.Component {
     const { fixed, saveButtonText, t } = this.props;
     const RoleDropdown = kind === 'RoleBinding' ? NsRoleDropdown : ClusterRoleDropdown;
     const title = `${this.props.titleVerb} ${kindObj(kind).label}`;
+    const isSubjectDisabled = fixed?.subjectRef?.subjectName ? true : false;
```
spadgett, line 503: nit:

```suggestion
    const isSubjectDisabled = !!fixed.subjectRef?.subjectName;
```

spadgett (approved):
One small comment, but not enough to hold the PR on.

/lgtm

PR #7848, opened 2021-01-15
Bug 1916642: Remove period in Chinese key/value translation

rebeccaalpert:
Removing unnecessary period.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1916642.


spadgett (approved):
/lgtm

PR #7849, opened 2021-01-15
Bug 1899627: Update project status icon and text in the dashboard Status panel to match health-item display

sg00dwin:
Changes to project status Icon and text sizes and spacing with Image Vulnerabilities.

- Wrap with `<Gallery>` component to match Admin overview status panel.
- Remove `frontend/public/components/dashboard/project-dashboard/status-card.scss` styles and use `.co-status-card__health-item` for positioning, same as `< ResourceHealthItem>`

Nested `.co-icon-and-text--lg  .co-icon-and-text__icon` rule needed to set size and spacing.

fix https://bugzilla.redhat.com/show_bug.cgi?id=1899627

cc @itsptk 

**Before**
Screen Shot 2021-01-15 at 10 27 51 AM

----
**After**
Screen Shot 2021-01-13 at 4 50 00 PM copy

----
**Overview Dashboard**
Screen Shot 2021-01-15 at 10 37 08 AM

### DIFF
--- a/frontend/packages/console-shared/src/components/dashboard/dashboard-card/card.scss
+++ b/frontend/packages/console-shared/src/components/dashboard/dashboard-card/card.scss
@@ -1,6 +1,8 @@
 @import '~@patternfly/patternfly/sass-utilities/colors';
+@import '../../../../../../public/components/_icon-and-text.scss';
 
 $co-dashboard-card--line-height: 19px;
+$co-dashboard-text-font-size: 0.875rem;
 
 .co-dashboard-card {
   margin: 0;
@@ -57,9 +59,9 @@ $co-dashboard-card--line-height: 19px;
 }
 
 .co-dashboard-icon {
-  font-size: 1.2rem;
+  font-size: $co-icon-and-text-icon-lg;
 }
 
 .co-dashboard-text--small {
-  font-size: 0.875rem;
+  font-size: $co-dashboard-text-font-size;
 }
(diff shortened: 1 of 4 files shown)


frontend/packages/console-shared/src/components/dashboard/dashboard-card/card.scss
```diff
@@ -1,6 +1,8 @@
 @import '~@patternfly/patternfly/sass-utilities/colors';
 
 $co-dashboard-card--line-height: 19px;
+$co-dashboard-icon-font-size: 1.2rem;
```
rhamilto, line 4: As we discussed, I think it's better for this to be renamed to move to `_icon-and-text.scss` and be imported here rather than the other way around.

sg00dwin: updated pr

itsptk:
Thanks @sg00dwin , LGTM!

frontend/public/components/_icon-and-text.scss
```diff
@@ -1,3 +1,7 @@
+// @import '../../packages/console-shared/src/components/dashboard/dashboard-card/card.scss';
```
rhamilto, line 1: Delete instead of comment out?

rhamilto (commented):
One nit.  Otherwise LGTM.

frontend/public/components/_icon-and-text.scss
sg00dwin: removed

rhamilto (approved):
/lgtm

PR #7850, opened 2021-01-15
Bug 1916834: Pipeline metrics text updates

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5310

**Analysis / Root cause**: 
- Pipeline Run should be capitalized in "Average pipeline run 
- It should be 2 of 2 succeeded not 2 succeeded of 
- In the tooltip of the donut chart, Success should be capitalized to follow sentence case

**Solution Description**: 
Update the text

scrt0
scrt1

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -68,12 +68,13 @@ const PipelineSuccessRatioDonut: React.FC<PipelineMetricsGraphProps> = ({
...
       count: obj.y,
       sortOrder,
+      x: _.capitalize(String(obj.x)),
       y: parseFloat(percentage),
       name: `${obj.x}: ${percentage}%`,
     };
   });
   let successTimeSeriesObj = finalArray.reduce((acc, obj) => {
-    if (obj.x !== 'success') return acc;
+    if (obj.x !== 'Success') return acc;
```
abhinandan13jan, line 77: Shouldn't we use an enum for these checks?? something like `Status.Success`

abhinandan13jan (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
rottencandy: I was not able to find an existing enum for pipeline graph axes. Should I create one?
Not sure if it's required since we only check for the `Success` axis here.

abhinandan13jan: not very necessary, but it would be really good, if you can store this string `Success` in a variable as it is being used in multiple places, IMO

rottencandy: Updated.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -36,6 +36,7 @@ const PipelineSuccessRatioDonut: React.FC<PipelineMetricsGraphProps> = ({
     delay: interval,
   });
   const pipelineSuccessData = runData?.data?.result ?? [];
+  const successString = 'Success';
```
rohitkrai03, line 39: Any reason to capitalize this string? I don't see it being rendered anywhere. You wouldn't have to explicitly capitalize strings in the obj to check for it's equality if you just kept it `success`.

rohitkrai03 (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
rottencandy: It is used by the donut chart component on hover. I realized It could be capitalized directly inside the component itself. Updated.

rohitkrai03 (approved):
/lgtm
/approve

bgliwa01:
In the first image, what is the black box that says pipeline success ratio?

PR #7863, opened 2021-01-19
[WIP]allow users to set service port

nemesis09:
Fixes:
https://issues.redhat.com/browse/ODC-4867

Analysis / Root cause:
During import from git, container image, and from dockerfile flows, the target port is linked to the creation of a route. However the target port is used as part of the Service resource creation and not the Route specifically.

Solution Description:
create a section for user to specify custom routes for their application

Screens:
servicePort

Test coverage:
Screenshot from 2021-01-19 06-06-17
Screenshot from 2021-01-19 06-07-06

Browser Conformance:
[x] Firefox
[] Chrome
[] Safari
[] Edge


andrewballantyne:
This is being redirected as not-a-bug and we will look at the feature support for this in a future release.

PR #7867, opened 2021-01-19
Bug 1917660: Dont use the /scale endpoint for Job resource

jhadvig:
/assign @rhamilto

### DIFF
--- a/frontend/public/components/modals/configure-count-modal.tsx
+++ b/frontend/public/components/modals/configure-count-modal.tsx
@@ -15,6 +15,7 @@ export const ConfigureCountModal = withHandlePromise((props: ConfigureCountModal
     path,
     resource,
     resourceKind,
+    opts,
     handlePromise,
     title,
     titleKey,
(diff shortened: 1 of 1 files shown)


dtaylor113 (approved):
/lgtm

frontend/public/components/modals/configure-count-modal.tsx
```diff
@@ -32,7 +32,9 @@ export const ConfigureCountModal = withHandlePromise((props: ConfigureCountModal
     e.preventDefault();
 
     const patch = [{ op: 'replace', path, value: _.toInteger(value) }];
-    const opts = { path: 'scale' };
+    // Jobs are not using the `/scale` endpoint.
```
spadgett, line 35: It would be better to pass this as a prop to the dialog if possible instead of hard-coding kinds here.

jhadvig:
@spadgett comment addressed. PTAL

/hold cancel

spadgett (approved):
/lgtm

spadgett:
Thanks, this looks cleaner 👍

PR #7871, opened 2021-01-19
Bug 1915972: Global configuration breadcrumbs do not work as expected

dtaylor113:
**Before**
Previously, when navigating to Admin->Cluster Settings->Global configuration, click on any cluster config resource to drilldown to details page.  Detail page has breadcrumbs such as: "Builds > Build details", where "Builds" navs to list view of cluster build configs.

**After**
Now if you click into a Global configuration resource details page, the breadcrumbs are "Global Configuration >
[resource] details" where "Global Configruation" navigates back to Admin->Cluster Settings->Global configuration

### DIFF
--- a/frontend/public/components/cluster-settings/global-config.tsx
+++ b/frontend/public/components/cluster-settings/global-config.tsx
@@ -12,10 +12,22 @@ import { addIDPItems } from './oauth';
 import { TextFilter } from '../factory';
 import { fuzzyCaseInsensitive } from '../factory/table-filters';
 import { withExtensions, isGlobalConfig, GlobalConfig } from '@console/plugin-sdk';
+import i18next from 'i18next';
 
 const stateToProps = (state: RootState) => ({
   configResources: state.k8s.getIn(['RESOURCES', 'configResources']),
 });
+
+export const breadcrumbsForGlobalConfig = (detailsPageKind: string, detailsPagePath: string) => [
+  {
+    name: 'Global Configuration',
+    path: '/settings/cluster/globalconfig',
+  },
+  {
+    name: i18next.t('details-page~{{kind}} details', { kind: detailsPageKind }),
+    path: detailsPagePath,
+  },
+];
 
 const ItemRow = ({ item }) => {
   return (
--- a/frontend/public/components/utils/breadcrumbs.ts
+++ b/frontend/public/components/utils/breadcrumbs.ts
@@ -2,6 +2,7 @@ import i18next from 'i18next';
 import { K8sKind } from '../../module/k8s';
 import { LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY } from '@console/shared/src/constants';
 import { ALL_NAMESPACES_KEY } from '@console/shared/src/constants/common';
+import { breadcrumbsForGlobalConfig } from '../cluster-settings/global-config';
 
 export const getBreadcrumbPath = (match: any, customPlural?: string) => {
   const lastNamespace = sessionStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
@@ -13,13 +14,16 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
   return `/k8s/cluster/${customPlural || match.params.plural}`;
 };
 
-export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () => [
-  {
-    name: `${kindObj.labelPlural}`,
-    path: getBreadcrumbPath(match),
-  },
-  {
-    name: i18next.t('details-page~{{kind}} details', { kind: kindObj.label }),
-    path: `${match.url}`,
-  },
-];
+export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () =>
+  kindObj.apiGroup === 'config.openshift.io' && match.params.name === 'cluster'
+    ? breadcrumbsForGlobalConfig(kindObj.label, match.url)
+    : [
+        {
+          name: `${kindObj.labelPlural}`,
+          path: getBreadcrumbPath(match),
+        },
+        {
+          name: i18next.t('details-page~{{kind}} details', { kind: kindObj.label }),
+          path: `${match.url}`,
+        },
+      ];
--- a/frontend/public/components/cluster-settings/cluster-version.tsx
+++ b/frontend/public/components/cluster-settings/cluster-version.tsx
@@ -7,6 +7,7 @@ import { DetailsPage } from '../factory';
 import { Conditions } from '../conditions';
 import { ClusterVersionKind, K8sResourceKindReference, referenceForModel } from '../../module/k8s';
 import { navFactory, ResourceSummary, SectionHeading } from '../utils';
+import { breadcrumbsForGlobalConfig } from './global-config';
 
 const clusterVersionReference: K8sResourceKindReference = referenceForModel(ClusterVersionModel);
 
@@ -34,6 +35,7 @@ export const ClusterVersionDetailsPage: React.FC<ClusterVersionDetailsPageProps>
     {...props}
     kind={clusterVersionReference}
     pages={[navFactory.details(ClusterVersionDetails), navFactory.editYaml()]}
+    breadcrumbsFor={() => breadcrumbsForGlobalConfig(ClusterVersionModel.label, props.match.url)}
   />
 );
 
--- a/frontend/public/components/monitoring/alerting.tsx
+++ b/frontend/public/components/monitoring/alerting.tsx
@@ -1,7 +1,7 @@
 import * as classNames from 'classnames';
 import i18next from 'i18next';
 import * as _ from 'lodash-es';
-import { Button, Popover } from '@patternfly/react-core';
+import { Button, Popover, Split, SplitItem } from '@patternfly/react-core';
 import { sortable } from '@patternfly/react-table';
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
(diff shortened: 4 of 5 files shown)


frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -13,13 +13,26 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
...
-    name: `${kindObj.labelPlural}`,
-    path: getBreadcrumbPath(match),
-  },
-  {
-    name: i18next.t('details-page~{{kind}} details', { kind: kindObj.label }),
-    path: `${match.url}`,
-  },
-];
+export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () => {
+  return match.url.startsWith('/k8s/cluster/config')
```
dtaylor113, line 17: @spadgett Is this too broad a change?  Can we say that any urls which start with '/k8s/cluster/config' should have a 'Global Config' breadcrumb?

dtaylor113: Arg..already found an exception:
`http://localhost:9000/k8s/ns/knative-serving/operator.knative.dev~v1alpha1~KnativeServing/knative-serving`
image

spadgett: Instead of this, I would check if the resource is in the `config.openshift.io` API group and maybe that the name is `cluster`.

frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -13,13 +13,26 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
...
+      ]
+    : [
+        {
+          name: `${kindObj.labelPlural}`,
+          path: getBreadcrumbPath(match),
+        },
+        {
+          name: i18next.t('details-page~{{kind}} details', { kind: kindObj.label }),
+          path: `${match.url}`,
+        },
```
spadgett, line 36: I'd avoid repeating this twice since it's identical to above.

frontend/public/components/utils/breadcrumbs.ts
dtaylor113: Hmm..have to handle `withExtensions` & `GlobalConfigPageExtension`'s

dtaylor113: Updated to use `kindObj.apiGroup === 'config.openshift.io' && match.params.name === 'cluster'`, however still not working for `knative-serving` which is:
```
kindObj.apiGroup = "operator.knative.dev"
match.params.name = "knative-serving"
```
Another option would be creating a Details Page component for GlobalConfig resources, instead of using `DefaultDetailsPage`, would be same as `DefaultDetailsPage` but different breadcrumbs. We would have to update `app-contents.tsx`, not sure how we do that with dynamic `GlobalConfigPageExtension`'s?

Another option I was considering was adding a search param to the links to the Details pages from the Global Config page, something like "?backTo='globalConfig'" or "detailPageType='globalConfigDetail'" and we could dynamically set the `breadcrumbsFor` prop from [here](https://github.com/openshift/console/blob/f8ffd52597c86737547b2935063f71acd0ca59c7/frontend/public/components/default-resource.jsx#L147)

frontend/public/components/utils/breadcrumbs.ts
dtaylor113: fixed

frontend/public/components/utils/breadcrumbs.ts
dtaylor113: Removed `&& match.params.name === 'cluster'` as Cluster Version's `match.params.name` equals `version`.

spadgett: > Removed `&& match.params.name === 'cluster'` as Cluster Version's `match.params.name` equals `version`.

I would leave the `cluster` check and manually pass the breadcrumbs as props in the other special cases (knative-serving and cluster version).

dtaylor113: @spadgett, 

- Most of the global config resources use DefaultDetailsPage which calls `breadcrumbsForDetailsPage()` which will set global config breadcrumbs via condition test `kindObj.apiGroup === 'config.openshift.io' && match.params.name === 'cluster'`.  
- Plugin resources will have url to details page: `...?backTo=globalConfig`, which will manually pass in breadcrumbs in `DefaultDetailsPage`.  
- ClusterVersion has a dedicated Details page which I pass in breadcrumbs.

spadgett: @dtaylor113 I'm unclear on the need for the query param. What are "plugin resources"? Do the special cases not have a dedicated details page?

dtaylor113:
**Added global config breadcrumbs for Alertmananger:**
image

**Same page from Monitoring -> Alerting (no breadcrumbs):**
image

frontend/public/components/utils/breadcrumbs.ts
dtaylor113: `knative-serving` is a `GlobalConfigPageExtension` what I'm calling a 'plugin resource'.  It doesn't have a dedicated detail page AFAIK.

spadgett: It looks like `KnativeServing` doesn't have a dedicated page. The right fix might be to add one, but that's really owned by the dev console team. I might fix the core breadcrumbs now and open a follow up low severity BZ for the `KnativeServing` breadcrumb.

An alternative is to have the breadcrumbs component check `GlobalConfigPageExtension`. It's starting to feel like too much of the config page logic is creeping into the breadcrumbs component, though. Even checking the API group there isn't ideal.

dtaylor113: Ok, I'll remove the code to handle the plugins, and I'll create a BZ for 'dev-console` to create a dedicated details page for `KnativeServing` and include the breadcrumbing code they should use.

spadgett:
@dtaylor113 I believe you need to run `yarn i18n`

dtaylor113:
@spadgett ran `yarn i18n`, removed the query param for plugins, and filed BZ for dev-console:
_Need dedicated details page w/ global config breadcrumbs for 'KnativeServing' plugin_
https://bugzilla.redhat.com/show_bug.cgi?id=1918803
Removing WIP as I think this is ready for final review  -thanks

frontend/public/components/cluster-settings/global-config.tsx
```diff
@@ -12,11 +12,26 @@ import { addIDPItems } from './oauth';
...
+import i18next from 'i18next';
 
 const stateToProps = (state: RootState) => ({
   configResources: state.k8s.getIn(['RESOURCES', 'configResources']),
 });
 
+export const breadcrumbsForGlobalConfig = (
+  detailsPageKind: string,
+  detailsPageUrl: string,
+) => () => [
```
spadgett, line 24: Why does this return a function instead of just returning the breadcrumbs array?

dtaylor113: Currently DetailsPage only has a [breadcrumbsFor](https://github.com/openshift/console/blob/1c9b925b8d722e9781a60c11fe39258d90bf916a/frontend/public/components/factory/details.tsx#L102) prop., which is a function.  Breadcrumbs component does take a `breadcrumb` prop which is a straight array of breadcrumb objs.  I'll add it to DetailsPageProps.

dtaylor113: Hi @spadgett, updated

frontend/public/components/factory/details.tsx
```diff
@@ -113,6 +114,7 @@ export type DetailsPageProps = {
   name?: string;
   namespace?: string;
   resources?: FirehoseResource[];
+  breadcrumbs?: { name: string; path: string }[];
   breadcrumbsFor?: (obj: K8sResourceKind) => { name: string; path: string }[];
```
spadgett, line 118: We shouldn't have two props that are essentially the same. I would just leave `breadcrumbsFor` and callers can always use `breadCrumbsFor={() => breadcrumbs}` if needed.

dtaylor113: ok, reverting it back...

dtaylor113: Reverted to single `breadcrumbsFor` prop., made `breadcrumbsForGlobalConfig` return array of breadcrumbs, code clean up for `breadcrumbsForDetailsPage`

frontend/public/components/cluster-settings/global-config.tsx
```diff
@@ -12,11 +12,23 @@ import { addIDPItems } from './oauth';
 import { TextFilter } from '../factory';
 import { fuzzyCaseInsensitive } from '../factory/table-filters';
 import { withExtensions, isGlobalConfig, GlobalConfig } from '@console/plugin-sdk';
+import i18next from 'i18next';
 
 const stateToProps = (state: RootState) => ({
   configResources: state.k8s.getIn(['RESOURCES', 'configResources']),
 });
 
+export const breadcrumbsForGlobalConfig = (detailsPageKind: string, detailsPageUrl: string) => [
```
spadgett, line 21: This should say "path" instead of "url"

```suggestion
export const breadcrumbsForGlobalConfig = (detailsPageKind: string, detailsPagePath: string) => [
```

frontend/public/components/cluster-settings/global-config.tsx
```diff
@@ -12,11 +12,23 @@ import { addIDPItems } from './oauth';
...
 });
 
+export const breadcrumbsForGlobalConfig = (detailsPageKind: string, detailsPagePath: string) => [
+  {
+    name: 'Global Configuration',
+    path: '/settings/cluster/globalconfig',
+  },
+  {
+    name: i18next.t('details-page~{{kind}} details', { kind: detailsPageKind }),
+    path: `${detailsPagePath}`,
```
spadgett, line 28:
```suggestion
    path: detailsPagePath,
```

spadgett (commented):
Sorry, I missed this last time, but the template literal isn't necessary.

frontend/public/components/cluster-settings/global-config.tsx
dtaylor113: fixed

frontend/public/components/cluster-settings/global-config.tsx
dtaylor113: fixed

spadgett (approved):
/lgtm

PR #7872, opened 2021-01-19
Bug 1918019: support rendering markdown inline instead of iframe

christianvogt:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5338

**Analysis / Root cause**: 
Markdown renders inside an iframe and when the view resizes, the markdown view doesn't get resized with it.
Another side effect of the iframe is the slow loading of the text areas in the quick start views.

**Solution Description**: 
The approach I took is to render the markdown inline without the iframe. In doing so we no longer get scroll bars around the markdown.

The UI looks the same. No more scroll bars around the individual markdown text components.

quick-start-md-iframe

@openshift/team-devconsole-ux

Added new unit tests for the markdown view to test rendering inside and outside of iframes, as well as rendering markdown extensions.

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/__tests__/components/markdown-view.spec.tsx
+++ b/frontend/__tests__/components/markdown-view.spec.tsx
@@ -1,0 +1,69 @@
+import * as React from 'react';
+import { act } from 'react-dom/test-utils';
+import { mount } from 'enzyme';
+import { SyncMarkdownView } from '../../public/components/markdown-view';
+
+jest.mock('showdown', () => ({
+  Converter: class {
+    makeHtml = (markdown) => markdown;
+  },
+}));
+
+jest.mock('react-i18next', () => {
+  const reactI18next = require.requireActual('react-i18next');
+  return {
+    ...reactI18next,
+    useTranslation: () => ({ t: (key) => key }),
+  };
+});
+
+describe('markdown-view', () => {
+  it('should render markdown view inline and iframe', () => {
+    expect(
+      mount(<SyncMarkdownView />)
+        .find('iframe')
+        .exists(),
+    ).toBe(true);
+    expect(
+      mount(<SyncMarkdownView inline />)
+        .find('iframe')
+        .exists(),
+    ).toBe(false);
+  });
+
+  it('should call renderExtension', () => {
+    const renderExtension = jest.fn();
+    mount(<SyncMarkdownView renderExtension={renderExtension} />);
+    expect(renderExtension).not.toHaveBeenCalled();
+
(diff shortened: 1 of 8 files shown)


spadgett:
I'm a little hesitant about this since the markdown is no longer sandboxed, which is the reason we used an `iframe`. I see a call to `sanitize-html` in `markdownConvert`, but are we absolutely sure there is no way to run JavaScript? (For example, through an event listener on one of the elements in the allowlist?)

christianvogt:
@spadgett it's a valid concern.
I've looked over the docs and sanitize-html does address this but like anything, if there's a bug or exploit found it does open up for attacks. I did some testing and between the markdown -> html conversion plus the sanitizing, it didn't allow me to execute javascript. However my tests were simple. I'm going to additionally update the sanitize-html to the latest version.

christianvogt:
@spadgett updated the sanitize-html library

Tested several areas that have markdown and everything seems to continue to render fine.

Since the quick starts render several markdown areas at once, this definitely is a better experience.

rohitkrai03 (approved):
/lgtm

Code looks good to me. Tested it locally in various scenarios. Works fine.

christianvogt:
@rohitkrai03 rebased and fixed merge conflict

rohitkrai03:
Seems like some tests are failing.

christianvogt:
I'll regenerate the `yarn.lock` file...

andrewballantyne:
@christianvogt Looks as if you still have a yarn issue. What's your version of yarn? Looks as if the CI is using `yarn v1.22.10`

andrewballantyne (changes requested):
LGTM, I was not able to inject anything dangerous. Script tags are removed as non approved tags, anchor hrefs that have `javascript:` are removed as they don't meet `http:` / `https:` / `mailto:`. OnClicks are not approved attributes on anything so they are removed. 

Looks pretty sane on the surface. Not sure what one could do to inject runnable JavaScript into it. 

Once the yarn stuff is fixed, this is a 👍  from me. 

@spadgett Do you have any objections to this merging?
/hold

Waiting for Sam to finish his review.

spadgett:
No objections, thanks guys.

/hold cancel

andrewballantyne:
pre-emptive LGTM... trust that Christian won't make me come back and see another `yarn.lock` issue 😜 

/lgtm

christianvogt:
yay. yarn file fixed. Now to get through the e2e tests.

PR #7877, opened 2021-01-20
Bug 1918185: capitalize pipelinerun details

abhinandan13jan:
# Addresses 
https://issues.redhat.com/browse/ODC-5337

# Issue
Capitalize pipeline run details

# Screenshot
Screenshot from 2021-01-25 14-43-02

# Browser conformance
Chrome


frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
```diff
@@ -12,7 +12,7 @@
   "Namespace": "Namespace",
   "Type": "Type",
   "Created": "Created",
-  "Pipeline run details": "Pipeline run details",
+  "Pipeline Run details": "Pipeline Run details",
```
andrewballantyne, line 15: "Pipeline Run Details" (ticket description)

andrewballantyne: Actually, it might _not_ be true? Perhaps a mistype by @bgliwa01

christianvogt: The admin pages use a lower case `details`.

christianvogt (commented):
What about the breadcrumbs?
image

All the admin pages use lower case `details`.

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
bgliwa01: It should be "Pipeline Run details"

jerolimov:
@abhinandan13jan The breadcrumb was called "Pipeline Run Details", can we change this as well? See screenshot:

image

And I can find 28 results for `Pipeline Run Details` in our sourcecode, and only 6 results for `Pipeline Run details`. Most of them are test related and maybe they work case-insensitive. But I would personally prefer if we adjust the integration tests as well. Wdyt?

jerolimov:
"Detail" or "Details" is btw also used on other pipeline detail pages. I would recommend that we change them all togeother.

Here are the **current** (with this PR) breadcrumb / h2 titles:

* Pipeline Details** -- Pipeline details
* Pipeline Run Details -- Pipeline Run details
* Task Run Details -- Task Run Details
* Pipeline Resource Details -- PipelineResource Details (without space)
* Trigger Template Details -- Trigger Template details
* Event Listener Details -- Event listener details (small L in title)

christianvogt:
@bgliwa01 ^^

jerolimov:
> @jerolimov your suggestions are inconsistent...

That are the current (status quo) implemention. The suggestion is to use always "details" an unique Pascal Case (with space) for the resource names.

andrewballantyne:
> > @jerolimov your suggestions are inconsistent...
> 
> That are the current (status quo) implemention. The suggestion is to use always "details" and Pascal Case (with space) for the resource names.

Didn't see the last one...

Event Listener Details -- Event listener details (small L in h1 title)

EventListener is the resource, it should be capital `Event` and capital `Listener`.

bgliwa01:
Agree with changing them all to be "details" on both the header and breadcrumbs. Also +1 to adding the space! Thank you!

abhinandan13jan:
> "Details" or "details" is btw also used on other pipeline detail pages. I would recommend that we change them all together also if only Pipeline Run details are requested.
> 
> Here are the **current** (with this PR) breadcrumb (left one below) and ---- h1 titles (right one below):
> 
> * Pipeline Details -- Pipeline details
> * Pipeline Run Details -- Pipeline Run details
> * Task Run Details -- Task Run Details
> * Pipeline Resource Details -- PipelineResource Details (=> without space in the h1 title, should add one here)
> * Trigger Template Details -- Trigger Template details
> * Event Listener Details -- Event listener details (lowercase L in h1 title => should use a uppercase L here)
> 
> And obviously, use "details" anywhere. Wdyt?

All the 5 breadcrumbs are updated and now use a lower case `details`
These 4 details page looks okay now `Pipeline details`, `Pipeline Run details`, `Task Run details`, `Event Listener details`

Please note that the PipelineResource page uses `navFactory.details(detailsFor())` Either we need to use a custom details page or update the factory function which will reflect in all such details pages.
cc: @andrewballantyne @christianvogt @jerolimov

andrewballantyne:
I logged https://bugzilla.redhat.com/show_bug.cgi?id=1919936 to address the DetailsForKind page. Should solve the PipelineResource and ClusterTask pages having capital `Details`.

/approve

andrewballantyne:
/retest

Rate limit error 😞

jerolimov (approved):
Tested this locally with pseudolocalization and it works fine. Thanks @abhinandan13jan for the extra effort here :+1:

/lgtm

bgliwa01 (approved):
lgtm

PR #7885, opened 2021-01-20
Bug 1921762: Fix knative serving and eventing breadcrumbs

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5341

**Solution description:**
- using the hook `useTabbedTableBreadcrumbsFor` to fix the breadcrumb path
- fixed the issue with the Triggers list showing an error when `subscriber.ref` is empty
- using extension (created in PR https://github.com/debsmita1/console/pull/6) for Event Sources, Channels & Brokers

**GIF & Screenshot:**
breadcrumbs

when no subscriber is associated with Trigger
nosubscriberassociatedwithtrigger

**Browser conformance:** 
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (commented):
@debsmita1 This implementation will only work for the static model. In case of dynamic models eventsources/channels this will fail. I created plugin extension to handle the breadcrumbs for detail page which covers the dynamic event sources as well. I raised a PR against your PR https://github.com/debsmita1/console/pull/6.

Plugin Extension:
```
namespace ExtensionProperties {
  export interface DetailPageBreadCrumbs {
    /**
     * array of models(kindObj) against which bread crumb is formed
     */
    getModels: () => K8sKind[] | K8sKind;
    /**
     * returns breadcrumb for the given kindref
     */
    useBreadcrumbs: (kind: string, urlMatch: match<any>) => any;
  }
}

export interface DetailPageBreadCrumbs
  extends Extension<ExtensionProperties.DetailPageBreadCrumbs> {
  type: 'DetailPageBreadCrumbs';
}

export const isDetailPageBreadCrumbs = (e: Extension): e is DetailPageBreadCrumbs => {
  return e.type === 'DetailPageBreadCrumbs';
};
```

cc @christianvogt @invincibleJai Please take a look.

invincibleJai:
I verified the PR , it works as expected however Ci jobs are failing

frontend/packages/knative-plugin/src/hooks/useEventSourceDetailPageBreadcrumbs.ts
```diff
@@ -0,0 +1,18 @@
+import { useTranslation } from 'react-i18next';
+import { match } from 'react-router';
+import { K8sKind } from '@console/internal/module/k8s';
+import { useActivePerspective } from '@console/shared/src/hooks/useActivePerspective';
+import { serverlessTab } from '../utils/serverless-tab-utils';
+import { useTabbedTableBreadcrumbsFor } from '@console/shared';
```
invincibleJai, line 6: nit: sort imports

frontend/public/components/factory/details.tsx
```diff
@@ -49,53 +81,70 @@ export const DetailsPage = withFallback<DetailsPageProps>(({ pages = [], ...prop
         })),
     [resourcePageExtensions, props],
   );
-
+  const kindObj = props.kindObj ? props.kindObj : modelFor(props.kind);
```
invincibleJai, line 84: can we have

```suggestion
  const kindObj = props.kindObj ?? modelFor(props.kind);
```

invincibleJai:
/lgtm

Thanks , verified works as expected!!

frontend/public/components/factory/details-breadcrumb-resolver.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const DetailsBreadcrumbResolver: React.FC<DetailsBreadcrumbResolverType> = ({
+  useBreadcrumbs,
+  onBreadcrumbsResolved,
+  kind,
+  urlMatch,
+}) => {
+  const breadcrumbs = useBreadcrumbs(kind, urlMatch);
+  React.useEffect(() => {
+    onBreadcrumbsResolved(breadcrumbs);
```
rohitkrai03, line 26: There should be a check for value returned by `useBreadcrumbs` hook. An async hook can lead to callback being called with `undefined`.

rohitkrai03 (changes requested):

frontend/packages/knative-plugin/src/plugin.tsx
```diff
@@ -37,6 +38,16 @@ import { TopologyConsumedExtensions, topologyPlugin } from './topology/topology-
...
+import {
+  eventSourceBreadcrumbsProvider,
+  channelBreadcrumbsProvider,
+  brokerBreadcrumbsProvider,
+} from './hooks';
+import {
+  eventSourceModelsProviderForBreadcrumbs,
+  channelModelsProviderForBreadcrumbs,
+  brokerModelProviderForBreadcrumbs,
+} from './utils';
```
sahil143, line 50: move these utilities to there respective hooks file as this is specific for breadcrumbs

karthikjeeyar:
Works as expected.
/lgtm

spadgett:

frontend/public/components/factory/details.tsx
```diff
@@ -49,53 +81,70 @@ export const DetailsPage = withFallback<DetailsPageProps>(({ pages = [], ...prop
         })),
     [resourcePageExtensions, props],
   );
-
+  const kindObj = props.kindObj ?? modelFor(props.kind);
+  const resolvedBreadcrumbExtension = useBreadCrumbsForDetailPage(kindObj);
+  const onBreadcrumbsResolved = React.useCallback((breadcrumbs) => {
+    setPluginBreadcrumbs(breadcrumbs ? () => breadcrumbs : undefined);
```
rohitkrai03, line 87: Why are you setting `pluginBreadcrumbs` with `() => breadcrumbs` when you're sending it to `breadcrumbs` prop of `PageHeading`? The type of `breadcrumbs` prop is `{ name: string; path: string }[];`

rohitkrai03 (changes requested):

frontend/packages/knative-plugin/src/hooks/index.ts
```diff
@@ -1,2 +1,37 @@
-export * from './useEventSourceModelsWithAccess';
-export * from './useEventSourceStatus';
+import { getExecutableCodeRef } from '@console/dynamic-plugin-sdk/src/coderefs/coderef-utils';
+
+export const eventSourceBreadcrumbsProvider = getExecutableCodeRef(() =>
```
rohitkrai03, line 3: i don't think it's a good idea to replace index.ts with coderef hook exports here. This can be confusing as general imports would expect to work from `/hooks` directly which will not happen. Also this leads to changing older imports from the hooks folder which is not right. Better to create a new file or something that has all coderefs like these and then import from that.

rohitkrai03 (changes requested):

frontend/packages/knative-plugin/src/catalog/useEventSourceProvider.ts
```diff
@@ -5,7 +5,7 @@ import { CatalogItem, CatalogExtensionHook } from '@console/plugin-sdk';
 import { K8sKind, referenceForModel } from '@console/internal/module/k8s';
 import { getEventSourceIcon } from '../utils/get-knative-icon';
 import { getEventSourceCatalogProviderData } from './event-source-data';
-import { useEventSourceModelsWithAccess } from '../hooks';
+import { useEventSourceModelsWithAccess } from '../hooks/useEventSourceModelsWithAccess';
```
rohitkrai03, line 8: Do not change older imports. See my comment above.

frontend/packages/knative-plugin/src/components/add/EventSourcePage.tsx
```diff
@@ -11,7 +11,7 @@ import { QUERY_PROPERTIES } from '@console/dev-console/src/const';
 import ConnectedEventSource from './EventSource';
 import EventSourceAlert from './EventSourceAlert';
 import { CamelKameletBindingModel } from '../../models';
-import { useEventSourceStatus } from '../../hooks';
+import { useEventSourceStatus } from '../../hooks/useEventSourceStatus';
```
rohitkrai03, line 14: Same as above.

frontend/packages/knative-plugin/src/components/pub-sub/details/SubscriptionDetailsPage.tsx
```diff
@@ -1,18 +1,34 @@
...
 
 const SubscriptionDetailsPage: React.FC<React.ComponentProps<typeof DetailsPage>> = (props) => {
+  const { kindObj, match } = props;
   const pages = [navFactory.details(SubscriptionDetails), navFactory.editYaml()];
   const commonActions = Kebab.factory.common.map((action) => action);
   const menuActionsCreator = [
     ...Kebab.getExtensionsActionsForKind(EventingSubscriptionModel),
     ...commonActions,
   ];
+  const breadcrumbsFor = useTabbedTableBreadcrumbsFor(
```
rohitkrai03, line 17:
```suggestion
  const breadcrumbs = useTabbedTableBreadcrumbsFor(
```

frontend/packages/knative-plugin/src/components/pub-sub/details/SubscriptionDetailsPage.tsx
```diff
@@ -1,18 +1,34 @@
...
+    match,
+    'eventing',
+    serverlessTab(kindObj.kind),
+  );
 
-  return <DetailsPage {...props} pages={pages} menuActions={menuActionsCreator} />;
+  return (
+    <DetailsPage
+      {...props}
+      breadcrumbsFor={() => breadcrumbsFor}
```
rohitkrai03, line 27:
```suggestion
      breadcrumbsFor={() => breadcrumbs}
```

frontend/packages/knative-plugin/src/components/pub-sub/details/TriggerDetailsPage.tsx
```diff
@@ -1,18 +1,34 @@
...
 import { DetailsPage } from '@console/internal/components/factory';
 import { Kebab, navFactory } from '@console/internal/components/utils';
+import { useTabbedTableBreadcrumbsFor } from '@console/shared';
 import TriggerDetails from './TriggerDetails';
 import { EventingTriggerModel } from '../../../models';
+import { serverlessTab } from '../../../utils/serverless-tab-utils';
 
 const TriggerDetailsPage: React.FC<React.ComponentProps<typeof DetailsPage>> = (props) => {
+  const { kindObj, match } = props;
+  const breadcrumbsFor = useTabbedTableBreadcrumbsFor(
```
rohitkrai03, line 11: Same as above. Similar suggestion for all such usages.

rohitkrai03 (changes requested):

frontend/packages/knative-plugin/src/hooks/index.ts
debsmita1: moved these to `src/providers`

frontend/public/components/factory/details.tsx
debsmita1: fixed it

sahil143 (commented):
@debsmita1 everything looks good. just one nit. 
verified locally as well. works as expected.

sahil143 (approved):
/lgtm

rohitkrai03 (approved):
/approve

PR #7888, opened 2021-01-21
Bug 1916882: OCS 4.7 wizard does not show zone when zone's are added manually

SanjalKatiyar:
Before:
Screenshot from 2021-01-21 10-02-10

After:
Screenshot from 2021-01-21 10-36-29

### DIFF
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/nodes-selection-list.tsx
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/nodes-selection-list.tsx
@@ -19,9 +19,9 @@ import {
   getNodeAllocatableMemory,
 } from '@console/shared';
 import { useSelectList } from '@console/shared/src/hooks/select-list';
-import { hasNoTaints } from '../../utils';
+import { hasNoTaints, getZone } from '../../utils';
 import { GetRows } from './types';
-import { getZone } from '@console/ceph-storage-plugin/src/utils/install';
+
 import './node-selection-list.scss';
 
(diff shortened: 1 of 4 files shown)


frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/nodes-selection-list.tsx
```diff
@@ -21,6 +21,7 @@ import {
 import { useSelectList } from '@console/shared/src/hooks/select-list';
 import { hasNoTaints } from '../../utils';
 import { GetRows } from './types';
+import { getZone } from '@console/ceph-storage-plugin/src/utils/install';
```
cloudbehl, line 24: We don't import from the Ceph plugin to LSO plugin that creates a circular dependency. I think you can define a `getZone` function in LSO plugin itself

SanjalKatiyar: Is this the right path where I can can make this function, instead of importing from ceph plugin??
`console/frontend/packages/local-storage-operator-plugin/src/utils/index.ts`

cloudbehl: That sounds good to me!

SanjalKatiyar: okay, thanks.

cloudbehl:
Thanks for the PR!

PR #7889, opened 2021-01-21
Bug 1918639: add missing trigger types and update the optional fields

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5350

**Analysis / Root cause**: 
Event listeners has an optional `triggerRef` field which is not compatible in UI code causes the UI to crash

**Solution Description**: 
Add the missing trigger types such as `triggerRef` and update the checks for optional fields.

Eventlistener page:
image

**Test setup:**
1. Create an eventlistener with the yaml

```
kind: EventListener
metadata:
  name: vote-app
spec:
  serviceAccountName: pipeline
  triggers:
    - triggerRef: vote-app
```

**Unit test cases:**

  EventListener Details
    ✓ should render eventlistener details page (25ms)
    ✓ should not render EventListenerTriggers section if the trigger contains binding ref (4ms)
    ✓ should not render EventListenerTriggers section if triggers contains triggerRef (3ms)
    
**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerDetails.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerDetails.tsx
@@ -11,6 +11,7 @@ export interface EventListenerDetailsProps {
 
 const EventListenerDetails: React.FC<EventListenerDetailsProps> = ({ obj: eventListener }) => {
   const { t } = useTranslation();
+  const triggers = eventListener.spec.triggers?.filter((trigger) => trigger.template?.name) || [];
   return (
     <div className="co-m-pane__body">
       <SectionHeading text={t('pipelines-plugin~Event listener details')} />
@@ -23,10 +24,12 @@ const EventListenerDetails: React.FC<EventListenerDetailsProps> = ({ obj: eventL
             eventListener={eventListener}
             namespace={eventListener.metadata.namespace}
           />
-          <EventListenerTriggers
-            namespace={eventListener.metadata.namespace}
-            triggers={eventListener.spec.triggers}
-          />
+          {triggers.length > 0 && (
+            <EventListenerTriggers
+              namespace={eventListener.metadata.namespace}
+              triggers={triggers}
+            />
+          )}
         </div>
       </div>
     </div>
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerTriggers.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerTriggers.tsx
@@ -18,11 +18,15 @@ interface EventListenerTriggersProps {
 
 const EventListenerTriggers: React.FC<EventListenerTriggersProps> = ({ namespace, triggers }) => {
   const { t } = useTranslation();
+  const triggerTemplates = triggers.filter((tr) => tr.template?.name);
+  if (triggerTemplates.length === 0) {
+    return null;
+  }
   return (
     <dl>
       <dt>{t('pipelines-plugin~Triggers')}</dt>
       <dd>
-        {triggers.map((trigger) => {
+        {triggerTemplates.map((trigger) => {
           const triggerTemplateKind = referenceForModel(TriggerTemplateModel);
           const triggerTemplateName = trigger.template.name;
           const bindings: ResourceModelLink[] = getEventListenerTriggerBindingNames(
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/EventListenerDetails.spec.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/EventListenerDetails.spec.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import EventListenerDetails from '../EventListenerDetails';
+import {
+  EventlistenerTestData,
+  EventlistenerTypes,
+} from '../../../../test-data/event-listener-data';
+import EventListenerTriggers from '../EventListenerTriggers';
+
+jest.mock('react-i18next', () => {
+  const reactI18next = require.requireActual('react-i18next');
+  return {
+    ...reactI18next,
+    useTranslation: () => ({ t: (key) => key }),
+  };
+});
+
+type EventListenerDetailsProps = React.ComponentProps<typeof EventListenerDetails>;
+
+describe('EventListener Details', () => {
+  let wrapper: ShallowWrapper<EventListenerDetailsProps>;
+
+  beforeEach(() => {
+    wrapper = shallow(
+      <EventListenerDetails
+        obj={EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_NAME]}
+      />,
+    );
+  });
+
+  it('should not render EventListenerTriggers section if the trigger contains binding ref', () => {
+    wrapper.setProps({ obj: EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_REF] });
+    expect(wrapper.find(EventListenerTriggers).exists()).toBe(false);
+  });
+
+  it('should not render EventListenerTriggers section if triggers contains triggerRef', () => {
+    wrapper.setProps({ obj: EventlistenerTestData[EventlistenerTypes.TRIGGER_REF] });
+    expect(wrapper.find(EventListenerTriggers).exists()).toBe(false);
+  });
+});
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
@@ -1,4 +1,4 @@
-import { K8sResourceCommon } from '@console/internal/module/k8s';
+import { K8sResourceCommon, K8sResourceKind, Toleration } from '@console/internal/module/k8s';
 import { PipelineRun } from '../../../utils/pipeline-augment';
 
 export type TriggerBindingParam = {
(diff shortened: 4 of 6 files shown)


andrewballantyne:
This is a feature implementation. I discussed with @karthikjeeyar for us to focus on the fix here as a means not to crash the UI if they used the feature we don't support. Updating types are likely to stay as they are not user-facing and will help us not whitescreen.

karthikjeeyar:
@andrewballantyne Updated the code to handle the UI crashes, conditional rendering of the supported resource-links in the evenlistener details page.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerDetails.tsx
```diff
@@ -11,6 +11,7 @@ export interface EventListenerDetailsProps {
 
 const EventListenerDetails: React.FC<EventListenerDetailsProps> = ({ obj: eventListener }) => {
   const { t } = useTranslation();
+  const triggers = eventListener.spec?.triggers?.filter((trigger) => trigger?.template?.name) || [];
```
andrewballantyne, line 14: Nit: `spec` isn't optional.
```suggestion
  const triggers = eventListener.spec.triggers?.filter((trigger) => trigger?.template?.name) || [];
```

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerTriggers.tsx
```diff
@@ -24,7 +24,7 @@ const EventListenerTriggers: React.FC<EventListenerTriggersProps> = ({ namespace
       <dd>
         {triggers.map((trigger) => {
           const triggerTemplateKind = referenceForModel(TriggerTemplateModel);
-          const triggerTemplateName = trigger.template.name;
+          const triggerTemplateName = trigger.template?.name;
```
andrewballantyne, line 27: If you don't have a template name, I think we need to return `null` to just kill this node. No sense in trying to craft a broken block.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/EventListenerDetails.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <EventListenerDetails
+        obj={EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_NAME]}
+      />,
+    );
+  });
+  it('should render eventlistener details page', () => {
+    expect(wrapper.find(ResourceSummary).exists()).toBe(true);
+    expect(wrapper.find(EventListenerURL).exists()).toBe(true);
+    expect(wrapper.find(EventListenerTriggers).exists()).toBe(true);
+  });
```
andrewballantyne, line 36: Careful with these kind of tests. Most of this is testing static content. It'll always be there in the current iteration of the code. So this test serves (mostly) to just test we don't change the content... which is not what tests are for.

Test just the conditionals and things that can pivot on you during prop changes. In this case, just `triggers.length` not rendering the `EventListenerTriggers` component when there are no triggers in the `eventListener`.

But this looks like you cover in your next two tests... so this test has very little value and will just serve to break when we want to change the static content.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
```diff
@@ -28,29 +28,79 @@ export type TriggerTemplateKind = K8sResourceCommon & {
 
 export type EventListenerKindBindingReference = {
   // TriggerBinding / ClusterTriggerBinding reference
-  kind: string;
+  // Kind can only be provided if Ref is also provided. Defaults to TriggerBinding
+  kind?: string;
   // Ref is used since Tekton Triggers 0.5 (part of OpenShift Pipeline Operator 1.1)
-  ref: string;
+  // Mutually exclusive with Name
```
andrewballantyne, line 34: This is a key opportunity for us to use TypeScript unions. 

Craft two types, one for old version and one for next version and then do a union of the two types:

```
/** @deprecated use EventListenerKindBindingReference instead */
type EventListenerKindBindingReferenceLegacy = {
  // Support for older operators (<0.5 Tekton Triggers)
  // https://github.com/tektoncd/triggers/pull/603/files
  // https://github.com/tektoncd/triggers/releases/tag/v0.5.0 and
  // https://github.com/tektoncd/triggers/releases/tag/v0.6.0
  name: string;
  value: string;
};

export type EventListenerKindBindingReference =
  | {
      // Kind/Ref is used since Tekton Triggers 0.5 (part of OpenShift Pipeline Operator 1.1)
      // Defaults to TriggerBinding
      kind?: string;
      ref: string;
    }
  | EventListenerKindBindingReferenceLegacy;
```

Honestly we probably won't keep this for much longer. Perhaps part of the new parity support in 4.8 we remove the legacy type.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
```diff
@@ -28,29 +28,79 @@ export type TriggerTemplateKind = K8sResourceCommon & {
...
+  interceptors?: TriggerInterceptor;
+  template?: {
+    // Ref is used since Tekton Triggers 0.10.x (part of OpenShift Pipeline Operator 1.3)
+    ref?: string;
+    // We also support older operators, so need to show & save the old field as well.
     // TriggerTemplateKind name reference
-    name: string;
+    // https://github.com/tektoncd/triggers/pull/898/files
+    // name will be deprecated in TP1.4
+    name?: string;
```
andrewballantyne, line 85: Union type again?

frontend/packages/pipelines-plugin/src/components/pipelines/utils/triggers.ts
```diff
@@ -28,7 +28,7 @@ type TriggerTemplateMapping = { [key: string]: TriggerTemplateKind };
 
 const getResourceName = (resource: K8sResourceCommon): string => resource.metadata.name;
 const getEventListenerTemplateNames = (el: EventListenerKind): string[] =>
-  el.spec.triggers.map((elTrigger: EventListenerKindTrigger) => elTrigger.template?.name);
+  el.spec?.triggers?.map((elTrigger: EventListenerKindTrigger) => elTrigger.template?.name) || [];
```
andrewballantyne, line 31: Nit: `spec` is not optional.
```suggestion
  el.spec.triggers?.map((elTrigger: EventListenerKindTrigger) => elTrigger.template?.name) || [];
```

frontend/packages/pipelines-plugin/src/components/pipelines/utils/triggers.ts
```diff
@@ -185,8 +185,8 @@ export const useTriggerTemplateEventListenerNames = (triggerTemplate: TriggerTem
 
   return eventListenerResources
     .filter((eventListener: EventListenerKind) =>
-      eventListener.spec.triggers.find(
-        ({ template: { name } }) => name === getResourceName(triggerTemplate),
+      eventListener.spec?.triggers?.find(
```
andrewballantyne, line 188: Nit: `spec` is not optional

frontend/packages/pipelines-plugin/src/components/pipelines/utils/triggers.ts
```diff
@@ -196,8 +196,8 @@ export const useTriggerBindingEventListenerNames = (triggerBinding: TriggerBindi
   const eventListenerResources = useAllEventListeners(triggerBinding.metadata.namespace) || [];
   return eventListenerResources
     .filter((eventListener: EventListenerKind) =>
-      eventListener.spec.triggers.find(({ bindings }) =>
-        bindings.find(
+      eventListener.spec?.triggers?.find(({ bindings }) =>
```
andrewballantyne, line 199: here too

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/EventListenerDetails.spec.tsx
karthikjeeyar: removed it.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
karthikjeeyar: yeah, we can clean up all this after a couple releases.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerTriggers.tsx
karthikjeeyar: filtering triggers based on template?.name and returning null early on in the component, so that <dt>Tiggers <dt> is also not rendered.

andrewballantyne: I think there was a misunderstanding here.

You made this `trigger.template?.name` optional, but it will fail to render properly if the `triggerTemplateName` is `undefined`. So we have to do something about the iterations of this map that do NOT have a `triggerTemplateName`. 

My recommendation is you return `null` and drop this iteration from rendering a broken link. I'm open to other suggestions, but we can't just leave it.

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerTriggers.tsx
karthikjeeyar: > You made this trigger.template?.name optional, but it will fail to render properly if the triggerTemplateName is undefined. So we have to do something about the iterations of this map that do NOT have a triggerTemplateName.

I am filtering out the tirggers based on tempate.name on the[ top of this component](https://github.com/openshift/console/pull/7889/files#diff-b7d6724fd77af9ccf450c9cee092536c0175d346fb63fe6ce93fce8874112eebR21-R24) now. so the map will only contain the set of triggers that consists of template.name.

karthikjeeyar:
> You made this trigger.template?.name optional, but it will fail to render properly if the triggerTemplateName is undefined. So we have to do something about the iterations of this map that do NOT have a triggerTemplateName.

Removed optional chaining as I have checks in the top of the component to early return null, if the template.name is not available in the triggers.

andrewballantyne (approved):
Thanks @karthikjeeyar Looks good 👍

spadgett:

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerDetails.tsx
```diff
@@ -11,6 +11,7 @@ export interface EventListenerDetailsProps {
 
 const EventListenerDetails: React.FC<EventListenerDetailsProps> = ({ obj: eventListener }) => {
   const { t } = useTranslation();
+  const triggers = eventListener.spec.triggers?.filter((trigger) => trigger?.template?.name) || [];
```
invincibleJai, line 14: nit: 
```suggestion
  const triggers = eventListener.spec.triggers?.filter((trigger) => trigger.template?.name) || [];
```

invincibleJai:
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/EventListenerDetails.tsx
andrewballantyne: I saw that too @invincibleJai... I didn't call it out because technically the item could be `null` in an array... but it's probably unlikely. Still probably best to drop the optionality for ease of reading all the same 👍

PR #7890, opened 2021-01-21
Bug 1891766: Stop non-positive values from being entered by users for min disk size

SanjalKatiyar:
Before:
Screenshot from 2021-01-21 16-09-02
Screenshot from 2021-01-21 16-08-58

After:
Screenshot from 2021-01-21 16-11-49
Screenshot from 2021-01-21 16-11-53

### DIFF
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
@@ -38,7 +38,7 @@
   flex-direction: column;
 }
 
-.lso-create-lvs__disk-size-form-group-max-input {
+.lso-create-lvs__disk-input {
   max-width: 100px;
 }
 
--- a/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
+++ b/frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
@@ -53,8 +53,10 @@ export const LocalVolumeSetInner: React.FC<LocalVolumeSetInnerProps> = ({
     dispatch({ type: 'setShowNodesListOnLVS', value: !state.showNodesListOnLVS });
   };
 
-  const validMaxDiskSize = /^\+?([1-9]\d*)$/.test(state.maxDiskSize || '1');
-  const validMaxDiskLimit = /^\+?([1-9]\d*)$/.test(state.maxDiskLimit || '1');
+  const INTEGER_REGEX = /^\+?([1-9]\d*)$/;
+  const validMinDiskSize = INTEGER_REGEX.test(state.minDiskSize || '1');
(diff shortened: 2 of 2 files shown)


frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/local-volume-set-inner.tsx
```diff
@@ -53,6 +53,7 @@ export const LocalVolumeSetInner: React.FC<LocalVolumeSetInnerProps> = ({
     dispatch({ type: 'setShowNodesListOnLVS', value: !state.showNodesListOnLVS });
   };
 
+  const validMinDiskSize = /^\+?([1-9]\d*)$/.test(state.minDiskSize || '1');
   const validMaxDiskSize = /^\+?([1-9]\d*)$/.test(state.maxDiskSize || '1');
   const validMaxDiskLimit = /^\+?([1-9]\d*)$/.test(state.maxDiskLimit || '1');
```
bipuladh, line 58:
```suggestion
  const INTEGER_REGEX = /^\+?([1-9]\d*)$/;
  const validMinDiskSize = INTEGER_REGEX.test(state.minDiskSize || '1');
  const validMaxDiskSize = INTEGER_REGEX.test(state.maxDiskSize || '1');
  const validMaxDiskLimit = INTEGER_REGEX.test(state.maxDiskLimit || '1');
```

bipuladh (changes requested):

cloudbehl:
I think the minimum size for a disk can be 0 as well.

Also, I think we need to put one more validation where the value to max disk size should be bigger than the value of min disk size.

@bipuladh ^^

bipuladh:
I agree with you @cloudbehl . @SanjalKatiyar I guess we will need two regex patterns. One that accepts 0 and the other one that doesn't. 
Also we should definitely give a warning in max value/min value based on which one is active so that user doesn't shoot himself in the foot. 
@yuvalgalanti wdyt?

bipuladh (changes requested):

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
```diff
@@ -10,6 +10,10 @@
   margin-top: 0;
 }
 
+.lso-create-lvs__max-disk-limit {
+  max-width: 100px;
+}
+
```
bipuladh, line 16: Let's remove this as well and use the other class.

bipuladh (changes requested):

frontend/packages/local-storage-operator-plugin/src/components/local-volume-set/create-local-volume-set.scss
SanjalKatiyar: Okay... using single class for min disk size, max disk size and max disk limit...

bipuladh (approved):
/lgtm

bipuladh:
/hold
Backend tests are failing

PR #7891, opened 2021-01-21
Bug 1918648: Replace deprecated 'installplan' Subscription status field with new 'installPlanRef'

jhadvig:
Based on the info from the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1918648), we are querying subscription.status.installPlan.name for subscription's installplan, but it has changed in 4.7, now it is subscription.status.installplan.name

Checked other occurrences where we are calling this field and this should be the only one.

/assign @TheRealJon

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx
@@ -103,7 +103,7 @@ export const installPlanForSubscription = (
   installPlans: InstallPlanKind[] = [],
   subscription: SubscriptionKind,
 ): InstallPlanKind =>
-  installPlans.find((ip) => ip?.metadata?.name === subscription?.status?.installPlan?.name);
+  installPlans.find((ip) => ip?.metadata?.name === subscription?.status?.installPlanRef?.name);
 
 export const upgradeRequiresApproval = (subscription: SubscriptionKind): boolean =>
   subscription?.status?.state === SubscriptionState.SubscriptionStateUpgradePending &&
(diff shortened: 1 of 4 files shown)


spadgett:
/hold

We should confirm with the OLM team whether the API has changed from previous releases and if that's intentional.

cc @kevinrizza

spadgett:
I suspect we were always using the wrong value since it's `installplan.name` here:

https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx#L129

jhadvig:
@spadgett yes it looks like we have been using wrong one. Also it looks liked it's being deprecated if favor of `installPlanRef` field.
Screenshot 2021-01-21 at 14 45 59

spadgett:
@jhadvig Thanks for changing this 👍. If we backport to 4.6, we might go with the simpler `installPlan` -> `installplan` fix.

frontend/packages/operator-lifecycle-manager/src/components/subscription.tsx
```diff
@@ -103,7 +103,7 @@ export const installPlanForSubscription = (
   installPlans: InstallPlanKind[] = [],
   subscription: SubscriptionKind,
 ): InstallPlanKind =>
-  installPlans.find((ip) => ip?.metadata?.name === subscription?.status?.installPlan?.name);
```
spadgett, line 106: Unfortunately, TypeScript didn't catch this since we used `& K8sResourceKind` in the type definition. We should try to move to `& K8sResourceCommon` for our types, which has more strict checking.

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

TheRealJon (approved):

PR #7894, opened 2021-01-21
Bug 1918785: Use kube_pod_resource_(request|limit) metric

smarterclayton:
The new kube_pod_resource_request and _limit metrics simplify the
visualization of the resources a running or pending pod is consuming
and perform init container calculations correctly. The console should
use them instead.

### DIFF
--- a/frontend/packages/console-app/src/components/nodes/node-dashboard/queries.ts
+++ b/frontend/packages/console-app/src/components/nodes/node-dashboard/queries.ts
@@ -85,40 +85,16 @@ const top25Queries = {
 
 const resourceQuotaQueries = {
   [NodeQueries.POD_RESOURCE_LIMIT_CPU]: _.template(
-    `sum(
-      max by (namespace, pod, container) (
-          kube_pod_container_resource_limits_cpu_cores{node='<%= node %>', job="kube-state-metrics"}
-      ) * on(namespace, pod) group_left() max by (namespace, pod) (
-          kube_pod_status_phase{phase=~"Pending|Running"} == 1
-      )
-    )`,
+    `sum(kube_pod_resource_limit{node='<%= node %>',resource='cpu'})`,
   ),
   [NodeQueries.POD_RESOURCE_LIMIT_MEMORY]: _.template(
-    `sum(
-      max by (namespace, pod, container) (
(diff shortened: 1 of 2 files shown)


spadgett:
cc @openshift/openshift-team-monitoring

frontend/packages/console-app/src/components/nodes/node-dashboard/queries.ts
```diff
@@ -85,40 +85,16 @@ const top25Queries = {
...
 const resourceQuotaQueries = {
   [NodeQueries.POD_RESOURCE_LIMIT_CPU]: _.template(
-    `sum(
-      max by (namespace, pod, container) (
-          kube_pod_container_resource_limits_cpu_cores{node='<%= node %>', job="kube-state-metrics"}
-      ) * on(namespace, pod) group_left() max by (namespace, pod) (
-          kube_pod_status_phase{phase=~"Pending|Running"} == 1
-      )
-    )`,
+    `sum(kube_pod_resource_limit{node='<%= node %>',resource='cpu'})`,
```
spadgett, line 88: nit: You shouldn't need backticks now since the strings no longer span lines and there's no variable replacement. (The `<%= node %>` variable is handled by lodash `_.template`.)

```suggestion
    "sum(kube_pod_resource_limit{node='<%= node %>',resource='cpu'})",
```

pod.tsx should still have the backticks, though.

spadgett (commented):
JS changes lgtm, but it would be good to get the monitoring team to review the queries.

Thanks @smarterclayton !

frontend/packages/console-app/src/components/nodes/node-dashboard/queries.ts
smarterclayton: Oops

smarterclayton: Won't that make this line inconsistent with all other lines in the file?  I.e. do you want these two lines to be inconsistent with all the others?

spadgett: Ah, I see we use template literals everywhere in the file. Don't worry about it then. Sorry.

(I thought we have a eslint rule for unnecessary template literals, but I guess not.)

spadgett (approved):
/lgtm

PR #7895, opened 2021-01-21
Fix to show CRW icon when CRW operator is installed

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-4939

**Analysis/Root cause:**
CHE icon is shown instead of CRW icon even when CRW operator is installed because a static CHE icon was being returned in both the cases.

**Solution:**
Use ConsoleLink CR to get the correct icon.

**Screenshot:**
Screenshot from 2021-01-21 20-56-43

**Unit test Coverage:**
image

**PR test setup:**
Install CRW operator.
Go to CRW operator details page and create CRW cluster.
After all the pods are running check if the ConsoleLink CR has been created.
Once the CR is created, you should be able to see the icon.

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/import/__tests__/render-utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/import/__tests__/render-utils.spec.ts
@@ -1,6 +1,7 @@
 import { TFunction } from 'i18next';
 import { BitbucketIcon, GitAltIcon, GithubIcon, GitlabIcon } from '@patternfly/react-icons';
 import { routeDecoratorIcon } from '../render-utils';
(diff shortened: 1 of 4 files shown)


rohitkrai03 (approved):
/approve

jerolimov (approved):
Tested this on a cluster 4.7.0-0.nightly-2021-01-21-215614 with CodeReady Workspaces 2.5.1.

Needed to accept an insecure https connection to see the icon which is a test-cluster issue. Everything else works as expected and code looks also fine.

Great to see some additional tests here :+1:

/lgtm

PR #7898, opened 2021-01-21
Bug 1921227: Improve code reference handling for Console dynamic plugins

vojtechszocs:
This PR improves the handling of `CodeRef` functions and ensures that extensions declared via `console-extensions.json` files have their code references marked with the right symbol.

cc @glekner @christianvogt

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-utils.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-utils.ts
@@ -1,7 +1,8 @@
 /* eslint-disable no-console */
 
 import { CodeRef } from '../types';
-import { codeRefSymbol } from './coderef-resolver';
+
+export { applyCodeRefSymbol as getExecutableCodeRef } from './coderef-resolver';
 
 /**
  * Execute function referenced by the `CodeRef` with given arguments.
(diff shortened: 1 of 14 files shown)


frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-utils.ts
```diff
@@ -1,7 +1,8 @@
 /* eslint-disable no-console */
 
 import { CodeRef } from '../types';
-import { codeRefSymbol } from './coderef-resolver';
+
+export { applyCodeRefSymbol as getExecutableCodeRef } from './coderef-resolver';
```
vojtechszocs, line 5: cc @rohitkrai03 @christianvogt

The `applyCodeRefSymbol` function replaces the existing `getExecutableCodeRef` helper.

spadgett:

frontend/packages/topology/src/extensions/topology.ts
```diff
@@ -51,7 +51,7 @@ namespace ExtensionProperties {
     id: string;
     priority: number;
     quadrant: TopologyDecoratorQuadrant;
-    decorator: CodeRef<TopologyDecoratorGetter>;
+    decorator: () => Promise<TopologyDecoratorGetter>;
```
christianvogt, line 54: nit: I'm not sure why you are making this change?
Seems to communicate better the intention by using `CodeRef`

christianvogt:
@vojtechszocs lgtm except my one nit.

frontend/packages/topology/src/extensions/topology.ts
vojtechszocs: Above `CodeRef` type was imported from `packages/console-plugin-sdk/src/typings/base.ts` which got modified as well:

```diff
-/**
- * Code reference, resolved to a function that returns the object `T`.
- *
- * TODO: Remove this once https://github.com/openshift/console/pull/7163 gets merged that adds support for dynamic extensions in static plugins.
- */
-export type CodeRef<T> = () => Promise<T>;
```

The `CodeRef` type is declared in `packages/console-dynamic-plugin-sdk/src/types.ts` since it's related to dynamic plugins.

I'll update code to reuse the right `CodeRef` type.

vojtechszocs: @christianvogt Suggestion addressed in the second commit, please have a look.

PR #7900, opened 2021-01-21
Bug 1917372: Monitoring: Change Prometheus UI link text to "Platform Prometheus UI"

kyoto:
To reflect that this link is to the UI for the platform (cluster-level) Prometheus.

"Platform" is the same wording as used on the Alerting pages.

FYI @sichvoge, @cshinn 

screenshot


spadgett (approved):
/lgtm
/hold for input from @cshinn

kyoto:
@spadgett I already discussed with @cshinn offline, so no need to hold, I think

/hold cancel

PR #7902, opened 2021-01-22
Bug 1919231: fix topology quick search modal layout

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5278

**Analysis / Root cause**: 
The quick search modal contents aren't scrollable and on small screens the contents is displayed beyond the bounds of the modal container.

**Solution Description**: 
Set content height and add properties `overflow-y: auto`.

Kapture 2021-01-22 at 17 48 21

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
@@ -1,9 +1,15 @@
 .odc-quick-search-content {
+  height: calc(100% - 60px);
   &__list {
     width: 40%;
+    height: 100%;
+    overflow-y: auto;
   }
   &__details {
+    background: var(--pf-global--palette--white);
     width: 60%;
+    overflow-y: auto;
+    height: 100%;
     padding-right: var(--pf-global--spacer--md);
   }
 }
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
@@ -9,6 +9,7 @@ import {
   removeQueryArgument,
   setQueryArgument,
 } from '@console/internal/components/utils';
+import './QuickSearchModalBody.scss';
 
 interface QuickSearchModalBodyProps {
   allCatalogItemsLoaded: boolean;
@@ -139,7 +140,11 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
   }, [closeModal, onCancel, onEnter, selectNext, selectPrevious]);
 
   return (
-    <div ref={ref}>
+    <div
+      ref={ref}
+      className="odc-quick-search-modal-body"
+      style={{ height: catalogItems?.length > 0 ? 468 : 60 }}
+    >
       <QuickSearchBar
         searchTerm={searchTerm}
         onSearch={onSearch}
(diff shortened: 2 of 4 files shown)


christianvogt:
Looks good.

One question for @bgliwa01 @openshift/team-devconsole-ux 
Should the `view all` be sticky to the bottom? Or can we leave it as shown in Vikram's gif?

bgliwa01:
If we could make the view all sticky that would be great! Thank you!

andrewballantyne:
> If we could make the view all sticky that would be great! Thank you!

@bgliwa01 My only concern here is if we sticky it, it may cover most of a single item. Although not entirely readable anyways, it is something we should be aware of. In Vikram's screenshot that may be almost the entire side part.

jerolimov:
> If we could make the view all sticky that would be great! Thank you!

But that means also that the scrollable area will be really small and you will only see 1 to 1 1/2 items and then "View all". Up to you but hmm :thinking:

jerolimov:
> If we could make the view all sticky that would be great! Thank you!

@bgliwa01  But that means also that the scrollable area will be really small and you will only see 1 to 1 1/2 items and then "View all". Up to you but hmm :thinking:

christianvogt:
IMO if the view is so small to begin with, the user is better off going to the catalog anyways.
I'm fine either way.
I'm also ok with getting the current fix in and following up in another issue for the sticky portion.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -139,7 +140,11 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
   }, [closeModal, onCancel, onEnter, selectNext, selectPrevious]);
 
   return (
-    <div ref={ref}>
+    <div
+      ref={ref}
+      className="odc-quick-search-modal-body"
+      style={{ height: catalogItems?.length > 0 ? '466px' : '60px' }}
```
christianvogt, line 146: Something wrong with 466px as it always shows a scrollbar when there are 5 results.
On my browser the height is 467px.
Instead of 60px you can also use undefined.
Lastly, in react `'466px'` is equivalent to the number `466`. Prefer using numbers instead of strings.

frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
```diff
@@ -1,9 +1,15 @@
 .odc-quick-search-content {
+  height: calc(100% - 60px);
   &__list {
     width: 40%;
+    height: calc(100%);
```
christianvogt, line 5:
```suggestion
    height: 100%;
```

bgliwa01:
"I'm also ok with getting the current fix in and following up in another issue for the sticky portion." +1

vikram-raj:
@christianvogt @bgliwa01 created an [issue](https://issues.redhat.com/browse/ODC-5402) for sticky footer.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
vikram-raj: I kept the `60px` height because the height of the input field is `60px` and it is a child of modal body.

debsmita1:
verified this locally
/lgtm

christianvogt:
Works to solve the issue.
Firefox still always shows a scrollbar.
There's also a hover issue on mobile that's odd.

The fix is an improvement and I'll log a separate issue for the remaining issues as they aren't caused by this change.

PR #7907, opened 2021-01-22
Bug 1919396: Fixed a11y violation for Deployment Config Details Pod Status ring

dtaylor113:


frontend/packages/console-shared/src/components/pod/PodStatus.tsx
```diff
@@ -97,6 +97,7 @@ const PodStatus: React.FC<PodStatusProps> = ({
           duration: ANIMATION_DURATION,
           onEnd: updateOnEnd ? forceUpdate : undefined,
         }}
+        ariaTitle={`${truncTitle} ${truncSubTitle}`} // ex: '3 pods' or '3 Scaling to 2'
```
spadgett, line 100: Should we use the untruncated value for screen readers? Presumably it's truncated because it won't fit inside the donut. Here we don't have that problem.

dtaylor113: fixed, thanks

frontend/packages/console-shared/src/components/pod/PodStatus.tsx
```diff
@@ -124,19 +125,21 @@ const PodStatus: React.FC<PodStatusProps> = ({
       />
     );
   }, [
+    updateOnEnd,
```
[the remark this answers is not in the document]
dtaylor113: fixed, thanks

frontend/packages/console-shared/src/components/pod/PodStatus.tsx
```diff
@@ -129,6 +130,8 @@ const PodStatus: React.FC<PodStatusProps> = ({
     outerRadius,
     size,
     standalone,
+    title,
+    subTitle,
```
spadgett, line 134: nit: alpha, this should go below `truncSubTitle`

dtaylor113: arrg, sorry, thanks again!  I hate Mondays :-)

PR #7910, opened 2021-01-25
Bug 1919851: fix pipeline visualization with same pipeline & task name

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5357

**Analysis / Root cause**: 

Sharing the same name for a pipeline and one of its tasks, does not render the visualization and adding additional tasks breaks the pipeline details page.

**Solution Description**: 

Pipeline visualization graph ID must be different from the node(task) ID. Id's used in the topology should be unique.

---
Before fix (_pipeline with same name as task name_):
---

Screen Shot 2021-01-20 at 3 44 33 PM
---
After fix (_pipeline with same name as task name_):
---
image
---
_Pipeline with an extra task_
---
image
---

**Test setup:**

1. Create a pipeline with below yaml
```
kind: Pipeline
metadata:
  name: myapp-install
spec:
  tasks:
    - name: myapp-install
      taskRef:
        kind: Task
        name: myapp-install
```
2. Add an extra task using edit pipeline feature. (shouldn't throw Javascript error )

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
@@ -49,7 +49,7 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
   } else {
     content = (
       <PipelineTopologyGraph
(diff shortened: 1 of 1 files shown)


andrewballantyne (approved):
/lgtm

PR #7913, opened 2021-01-25
Update gitops backend service namespace

chetan-rns:
The GitOps operator will deploy the gitops backend service in openshift-gitops namespace. This PR will update the backend service namespace from openshift-pipelines-app-delivery to openshift-gitops.

Change in the operator: https://github.com/redhat-developer/gitops-operator/pull/32

Edit: Adding the complete list of PRs which enabled this feature in the operator:
https://github.com/redhat-developer/gitops-operator/pull/43
https://github.com/redhat-developer/gitops-operator/pull/29
https://github.com/redhat-developer/gitops-operator/pull/46

cc @bigkevmcd @wtam2018 @sbose78

### DIFF
--- a/cmd/bridge/main.go
+++ b/cmd/bridge/main.go
@@ -52,7 +52,7 @@ const (
 	openshiftMeteringHost = "reporting-operator.openshift-metering.svc:8080"
 
 	// Well-known location of the GitOps service. This is only accessible in-cluster
-	openshiftGitOpsHost = "cluster.openshift-pipelines-app-delivery.svc:8080"
+	openshiftGitOpsHost = "cluster.openshift-gitops.svc:8080"
 )
 
 func main() {
--- a/contrib/oc-environment.sh
+++ b/contrib/oc-environment.sh
@@ -31,7 +31,7 @@ export BRIDGE_K8S_MODE_OFF_CLUSTER_THANOS
 BRIDGE_K8S_MODE_OFF_CLUSTER_ALERTMANAGER=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.alertmanagerPublicURL}')
 export BRIDGE_K8S_MODE_OFF_CLUSTER_ALERTMANAGER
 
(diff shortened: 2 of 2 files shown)


bigkevmcd (approved):
Thanks!

sbose78 (approved):
This is for 4.7, right?

chetan-rns:
> This is for 4.7, right?

Yeah @sbose78

jhadvig (approved):
/lgtm
/approve

spadgett (commented):
What happens when upgrading from 4.6 to 4.7?

spadgett:
Will this break things if I upgrade the GitOps operator without upgrading to 4.7, or if I upgrade to 4.7 without upgrading the GitOps operator?

chetan-rns:
Hi @spadgett, It won't break since it's a single productized bundle for both 4.6 and 4.7. The operator can determine the namespace based on the version in which it is installed. cc @wtam2018 @bigkevmcd

spadgett:
> Hi @spadgett, It won't break since it's a single productized bundle for both 4.6 and 4.7. The operator can determine the namespace based on the version in which it is installed. cc @wtam2018 @bigkevmcd

I guess I'm unclear on that because I don't see any of that logic in https://github.com/redhat-developer/gitops-operator/pull/32. Also it looks like both the old and new namespace will remain after upgrading the operator?

Namespace names are typically considered APIs that shouldn't change release-to-release.

wtam2018:
The GitOps Operator that we are releasing will support 4.6 and above.

spadgett:

chetan-rns:
@spadgett This is the expected operator behavior

```
4.6 Cluster: Create both openshift-pipelines-app-delivery and openshift-gitops namespace
4.7 Cluster: Create only openshift-gitops namespace 
```
This won't break since the old namespace is still created when the operator is installed in a 4.6 cluster.

spadgett (approved):
/lgtm

PR #7916, opened 2021-01-25
Update .gitignore to include all cypress-a11y-report.json files.

TheRealJon:


spadgett (approved):
/lgtm

.gitignore
```diff
@@ -17,7 +18,6 @@
 /frontend/@types/gql
 /frontend/gui_test_screenshots
 /frontend/packages/integration-tests-cypress/cypress-a11y-report.json
```
spadgett, line 20: @TheRealJon Sorry, I just noticed this, but are we able to remove this line and line 21 below as well now?

PR #7917, opened 2021-01-25
Bug 1917585: Follow-on for NsDropdown placeholder i18n.

TheRealJon:
Placeholder for NsDropdown needs to be translated.


spadgett (approved):
/lgtm

PR #7918, opened 2021-01-25
Bug 1920200: Throw error if ConfigMap creation fails with unexpected response

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5332

**Analysis / Root cause**: 
When the console backend cannot be reached, the console front end enters a loop trying to fetch user-settings resulting in high CPU usage for the browser.

**Solution Description**: 
The loop seems to be caused by `useK8sWatchResource` on the userSettings ConfigMap, which keeps repeatedly polling if the request fails. That triggers an effect which tries to create or update the ConfigMap.

~The fix in this PR makes the effect always fall back to local storage if the ConfigMap fetch fails.~
Throw an exception preventing any further rendering of the hook.

(No UI change)

**Test setup:**
- Load up a page
- Disconnect the bridge
- Navigate to the +Add page

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -84,8 +84,7 @@ export const useUserSettings = <T>(
           if (err?.response?.status === 403 || err?.response?.status === 404) {
             setFallbackLocalStorage(true);
           } else {
-            setSettings(defaultValueRef.current);
-            setLoaded(true);
+            throw err;
           }
         }
       })();


christianvogt:
I'm leaning towards throwing an exception instead and let the error boundary display an error. While we do have local storage as a fallback, it was intended to handle specific use cases like our dev env. A normal user who encounters a network issue and enters this state would lose any settings that happen to make it to local storage since the next time we will load from the configmap again.

@spadgett What are your thoughts?

jerolimov:
+1 for using localStorage only when the backend doesn't support it, not if there is an network error.

My recommendation here is to save (in a useState?) if we already tried to create a ConfigMap. If this doesn't work do not try this again (or at least do not try it again for n seconds)?

spadgett:
We definitely need to back-off to avoid thrashing when this happens.

We're already handling 403/404 status. Are there other common or expected errors? If the problem is transient or unexpected, I'd lean to showing an error message and backing off retries for the reasons @christianvogt mentions.

christianvogt:
The expected use case with 403 /404 is already handled. Anything else I believe is unexpected.
We'll go with throwing the exception which will stop the thrashing so long as the parent stop rendering the hook. Which should be the case with the error boundary.

rottencandy:
> The expected use case with 403 /404 is already handled. Anything else I believe is unexpected.
> We'll go with throwing the exception which will stop the thrashing so long as the parent stop rendering the hook. Which should be the case with the error boundary.

Updated to throw an error for the unexpected cases.

spadgett:

PR #7919, opened 2021-01-25
Bug 1921227: Avoid require()'ing modules when generating @console/active-plugins

vojtechszocs:
Depends on #7898

When building Console via webpack, static plugins have their dynamic extensions validated by `ConsoleActivePluginsModule` as part of `@console/active-plugins` virtual module generation.

This virtual webpack module exports all static plugins (and their extensions) for use in the Console app.

The idea was to keep dynamic extension validation logic consistent between [static plugins](https://github.com/openshift/console/tree/master/frontend/packages/console-plugin-sdk) and [dynamic plugins](https://github.com/openshift/console/tree/master/frontend/packages/console-dynamic-plugin-sdk).

In case of static plugins, however, current implementation of test cases
- `code ref must reference a valid module`
- `code ref must reference a valid export of the given module`

relies on Node.js `require()` which fails on code that isn't strictly JavaScript (including ES module import syntax).

:memo: Note: _It is possible to re-implement ^^ test cases to utilize webpack `compilation` object similar to [`ExtensionValidator.ts`](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/src/validation/ExtensionValidator.ts) (or adapt it to support both static and dynamic plugin build context) but this effort is non-trivial, so out of this PR's scope._

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
@@ -12,7 +12,12 @@ import {
 
 // TODO(vojtech): support code refs at any level within the properties object
 
-export const codeRefSymbol = Symbol('CodeRef');
+const codeRefSymbol = Symbol('CodeRef');
+
+export const applyCodeRefSymbol = <T = any>(ref: CodeRef<T>) => {
+  ref[codeRefSymbol] = true;
+  return ref;
+};
 
 export const isEncodedCodeRef = (obj): obj is EncodedCodeRef =>
   _.isPlainObject(obj) &&
@@ -98,8 +103,7 @@ export const resolveEncodedCodeRefs = (
       const executableCodeRef: CodeRef = async () =>
         loadReferencedObject(ref, entryModule, pluginID, errorCallback);
 
-      executableCodeRef[codeRefSymbol] = true;
-      e.properties[propName] = executableCodeRef;
+      e.properties[propName] = applyCodeRefSymbol(executableCodeRef);
     });
 
     return e;
--- a/frontend/packages/console-plugin-sdk/src/store.ts
+++ b/frontend/packages/console-plugin-sdk/src/store.ts
@@ -62,22 +62,10 @@ export class PluginStore {
       ),
     );
     this.registry = new ExtensionRegistry(plugins);
-    this.updateDynamicExtensions = _.debounce(this.updateDynamicExtensions, 1000);
   }
 
   public getAllExtensions() {
     return [...this.staticPluginExtensions, ...this.dynamicPluginExtensions];
-  }
-
-  private updateDynamicExtensions() {
-    this.dynamicPluginExtensions = Array.from(this.dynamicPlugins.values()).reduce(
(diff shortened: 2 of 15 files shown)


spadgett:

frontend/packages/console-plugin-sdk/src/store.ts
```diff
@@ -62,24 +62,12 @@ export class PluginStore {
       ),
     );
     this.registry = new ExtensionRegistry(plugins);
-    this.updateDynamicExtensions = _.debounce(this.updateDynamicExtensions, 1000);
```
vojtechszocs, line 65: FYI @spadgett 

The motivation was to debounce re-computations of Console extensions & subsequent invocation of all listeners, e.g. when a user fast-toggles a dynamic plugin's enabled status via UI.

However, reading up on debounce & throttle methods, I think it's best to debounce the actual UI event (e.g. click) handler, instead of doing it in plugin infra code.

glekner:
experiencing some typescript errors:

```
ERROR in /Users/glekner/coding/console/frontend/public/components/factory/details.tsx(50,28):
TS2349: This expression is not callable.
  Type '{}' has no call signatures.
ERROR in /Users/glekner/coding/console/frontend/public/components/factory/details.tsx(96,11):
TS2322: Type 'unknown' is not assignable to type 'DetailsPageBreadCrumbsHook'.
  Type '{}' provides no match for the signature '(kind: K8sKind, urlMatch: match<any>): { name: string; path: string; }[]'.
```

vojtechszocs:
> experiencing some typescript errors:
> 
> ```
> ERROR in /Users/glekner/coding/console/frontend/public/components/factory/details.tsx(50,28):
> TS2349: This expression is not callable.
>   Type '{}' has no call signatures.
> ERROR in /Users/glekner/coding/console/frontend/public/components/factory/details.tsx(96,11):
> TS2322: Type 'unknown' is not assignable to type 'DetailsPageBreadCrumbsHook'.
>   Type '{}' provides no match for the signature '(kind: K8sKind, urlMatch: match<any>): { name: string; path: string; }[]'.
> ```

This is related to recent changes in `master` branch. Rebase & update of #7898 will fix this.

frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
```diff
@@ -12,7 +12,12 @@ import {
 
 // TODO(vojtech): support code refs at any level within the properties object
 
-export const codeRefSymbol = Symbol('CodeRef');
+const codeRefSymbol = Symbol('CodeRef');
+
+export const applyCodeRefSymbol = <T = any>(ref: CodeRef<T>) => {
```
rawagner, line 17: nit: would be nice to specify return type explicitly
```suggestion
export const applyCodeRefSymbol = <T = any>(ref: CodeRef<T>): CodeRef<T> => {
```

vojtechszocs: @rawagner OK, I can do that in a follow-up PR.

TypeScript infers the return type to `CodeRef<T>` anyway, due to `return ref;` being the single exit point of the function.

For simple functions (e.g. low cyclomatic complexity) I prefer to omit return type declaration and rely on type inference.

PR #7920, opened 2021-01-25
Move baseCRDs definitions to console-app/console-extensions.json

vojtechszocs:
- [x] CRD to feature flag mapping defaults in `public/reducers/features.ts` removed
- [x] Equivalent `console.flag/model` extensions added to Console app's `console-extensions.json`


christianvogt:
I didn't run it. But the change looks ok. Good to see us moving to the new extensions now.

spadgett:

PR #7923, opened 2021-01-25
Bug 1920248: fix pipeline rerun when pipelineSpec is embedded

nemesis09:
Fixes:
https://issues.redhat.com/browse/ODC-5395

Analysis / Root cause:
While re-running the pipelineRun with embedded pipelinespec causes the UI to crash.

Solution Description:
update util getPipelineRunData and kebab actions to handle the scenario when pipeline is passed in pipelineSpec instead of a reference in pipelineRef

Screen shots / Gifs for design review:
pipelineSpec

Test Coverage:
Screenshot from 2021-01-30 20-55-07

Browser Conformance:
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/utils.spec.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/utils.spec.ts
@@ -1,5 +1,5 @@
 import { PipelineRun, Pipeline } from '../../../../../utils/pipeline-augment';
-import { TektonResourceLabel } from '../../../const';
+import { TektonResourceLabel, preferredNameAnnotation } from '../../../const';
 import {
   pipelineTestData,
   PipelineExampleNames,
@@ -7,6 +7,7 @@ import {
 } from '../../../../../test-data/pipeline-data';
 import {
   convertPipelineToModalData,
+  getPipelineName,
   getPipelineRunData,
   getPipelineRunFromForm,
   migratePipelineRun,
@@ -28,6 +29,83 @@ const pipelineRunData = (pipeline: Pipeline): PipelineRun => ({
   spec: {
     pipelineRef: { name: pipeline.metadata.name },
   },
+});
+
+describe('PipelineAction testing migratePipelineRun', () => {
+  it('expect migratePipelineRun to do nothing when there is no migration needed', () => {
+    // Same instance should be returned if there was no need for a migration
+    expect(migratePipelineRun(samplePipelineRun)).toEqual(samplePipelineRun);
+  });
+
+  it('expect migratePipelineRun to handle serviceAccount to serviceAccountName migration (Operator 0.9.x)', () => {
+    type OldPipelineRun = PipelineRun & {
+      spec: {
+        serviceAccount: string;
+      };
+    };
+    const serviceAccountValue = 'serviceAccountValue';
+    const plr: OldPipelineRun = {
+      ...samplePipelineRun,
+      spec: {
+        ...samplePipelineRun.spec,
+        serviceAccount: serviceAccountValue,
+      },
+    };
+
+    const result: PipelineRun = migratePipelineRun(plr);
+
+    // Should be a new instance
+    expect(result).not.toEqual(plr);
+
+    // The value should have moved
+    expect(result.spec.serviceAccountName).toEqual(serviceAccountValue);
+    expect((result as OldPipelineRun).spec.serviceAccount).toBeUndefined();
+
+    // Should still have other spec properties
+    expect(result.spec.pipelineRef).toEqual(samplePipelineRun.spec.pipelineRef);
+  });
+});
+
+describe('getPipelineName', () => {
+  it('should return null if no argument is provided', () => {
+    expect(getPipelineName()).toBeNull();
+  });
+
+  it('should return the name from the pipeline metadata if both pipeline and latestRun are provided', () => {
+    expect(getPipelineName(samplePipeline, samplePipelineRun)).toEqual(
+      samplePipeline.metadata.name,
+    );
+  });
+
+  it('should return the name from the pipeline metadata if only pipeline is provided', () => {
+    expect(getPipelineName(samplePipeline)).toEqual(samplePipeline.metadata.name);
+  });
+
+  it('should return the name from the pipelineRef, if only latestRun is provided and pipelineRef exists', () => {
+    const pipelineRunWithPipelineRef =
+      pipelineTestData[PipelineExampleNames.EMBEDDED_PIPELINE_SPEC].pipelineRuns[DataState.SKIPPED];
+    expect(getPipelineName(undefined, pipelineRunWithPipelineRef)).toEqual(
+      pipelineRunWithPipelineRef.spec.pipelineRef.name,
+    );
+  });
+
+  it('should return the name from the latestRun metadata, if only latestRun is provided and annotation does not include the name', () => {
+    const pipelineRunWithPipelineSpec =
+      pipelineTestData[PipelineExampleNames.EMBEDDED_PIPELINE_SPEC].pipelineRuns[
+        DataState.IN_PROGRESS
+      ];
+    expect(getPipelineName(undefined, pipelineRunWithPipelineSpec)).toEqual(
+      pipelineRunWithPipelineSpec.metadata.name,
+    );
+  });
+
+  it('should return the name from the annotations, if only latestRun is provided and annotation includes the name', () => {
+    const pipelineRunWithAnnotations =
+      pipelineTestData[PipelineExampleNames.EMBEDDED_PIPELINE_SPEC].pipelineRuns[DataState.SUCCESS];
+    expect(getPipelineName(undefined, pipelineRunWithAnnotations)).toEqual(
+      pipelineRunWithAnnotations.metadata.annotations?.[preferredNameAnnotation],
+    );
+  });
 });
 
 describe('PipelineAction testing getPipelineRunData', () => {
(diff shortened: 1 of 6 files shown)


frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
```diff
@@ -43,8 +43,9 @@ export const reRunPipelineRun: KebabAction = (kind: K8sKind, pipelineRun: Pipeli
   labelKey: 'pipelines-plugin~Rerun',
   callback: () => {
     const namespace = _.get(pipelineRun, 'metadata.namespace');
-    const pipelineRef = _.get(pipelineRun, 'spec.pipelineRef.name');
-    if (namespace && pipelineRef) {
+    const pipelineRef = pipelineRun.spec.pipelineRef?.name;
+    const { pipelineSpec } = pipelineRun.spec;
+    if (namespace && (pipelineRef || pipelineSpec)) {
```
andrewballantyne, line 48:
```suggestion
    const { pipelineSpec, pipelineRef } = pipelineRun.spec;
    if (namespace && (pipelineRef?.name || pipelineSpec)) {
```

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -84,13 +91,18 @@ export const getPipelineRunData = (
...
+          'tekton.dev/pipeline': pipelineName,
+        },
+      ),
     },
     spec: {
       ...(latestRun?.spec || {}),
-      pipelineRef: {
+      pipelineRef: latestRun.spec.pipelineRef && {
         name: pipelineName,
       },
```
andrewballantyne, line 107: Perhaps avoid adding it if it's not there.
```suggestion
      ...(latestRun.spec.pipelineRef && { pipelineRef: {
        name: pipelineName,
      } }),
```

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -43,6 +43,12 @@ export const migratePipelineRun = (pipelineRun: PipelineRun): PipelineRun => {
   return newPipelineRun;
 };
 
+export const getPipelineName = (pipeline?: Pipeline, latestRun?: PipelineRun) =>
```
andrewballantyne, line 46: Needs tests! 😜

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -43,6 +43,12 @@ export const migratePipelineRun = (pipelineRun: PipelineRun): PipelineRun => {
...
+  pipeline
+    ? pipeline.metadata.name
+    : latestRun.spec.pipelineRef?.name ??
```
andrewballantyne, line 49: if you don't have a pipeline or a latestRun, you'll get an NPE 🙂  -- when you write tests you will see the logic shift.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -68,6 +74,7 @@ export const getPipelineRunData = (
     {},
     pipeline?.metadata?.annotations,
     latestRun?.metadata?.annotations,
+    !latestRun?.metadata?.annotations?.preferredName && { preferredName: pipelineName },
```
andrewballantyne, line 77: Only real question here is do we want to call this something else.

cc @vdemeester 

Essentially what's happening here is we have a PipelineRun without a reference to a Pipeline. However, we build our PipelineRun names off of the Pipeline name + 6 random characters. Without that, we sorta need some clean reference otherwise we'll end up with PipelineRun + 6 random characters + 6 random... etc. So we are going with the first one we set in the annotations, "preferredName" was off the top of my head. Thoughts?

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx
nemesis09: done

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: done

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: added tests

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: We wouldn't reach the call to `getPipelineName` if pipeline and `latestRun` both are absent, since it is handled at the very beginning of the definition of `getPipelineRunData`

andrewballantyne: This function knows nothing of where it was called. If you write fully comprehensive tests for this function, you'd pass null for both values and see it crashes. Please add more tests.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/utils.spec.ts
```diff
@@ -239,3 +240,36 @@ describe('convertPipelineToModalData', () => {
     expect(workspaces.filter((workspace) => workspace.type === 'PVC')).toHaveLength(3);
   });
 });
+
+describe('getPipelineName', () => {
```
andrewballantyne, line 244: Initial look, you're missing a couple tests here:

- what if nothing is passed?
- what if nulls are passed for one / both?
- what if both pipeline AND pipeline run are passed?

Remember, Unit Tests are to catch all uses of a function, not just the ones from the expected path.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
vdemeester: Do we want to *namespace* the annotation name, like `io.openshift.pipeline/preferredName` ?
Otherwise, seems good to me.

What happens if the pipelinerun doesn't have this annotation ? Will it then add the annotation for the next one ?

nemesis09: Thanks for the input @vdemeester 
yes, if the pipeline run doesn't have this annotation, it will be added to the next one.
I think we could have the annotation namespaced.  cc @andrewballantyne

andrewballantyne: Yup. Namespacing it sounds like a good idea

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/utils.spec.ts
nemesis09: added tests for following scenarios
- nothing being passed to the function, i.e. both values falsy
- both pipeline and pipeline run being passed to the function

added a check in the function to return null if both arguments are falsy. So do we need another test for null values, since it will treat null as falsy value?

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: added a check in the function to handle the scenario when both arguments are falsy

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -43,6 +43,16 @@ export const migratePipelineRun = (pipelineRun: PipelineRun): PipelineRun => {
...
 
+export const getPipelineName = (pipeline?: Pipeline, latestRun?: PipelineRun): string => {
+  if (!pipeline && !latestRun) {
+    return null;
+  }
+  return pipeline
+    ? pipeline.metadata.name
+    : latestRun.spec.pipelineRef?.name ??
+        (latestRun.metadata.annotations?.preferredName || latestRun.metadata.name);
+};
```
andrewballantyne, line 54:
```suggestion
export const getPipelineName = (pipeline?: Pipeline, latestRun?: PipelineRun): string => {
  if (pipeline) {
    return pipeline.metadata.name
  }
  if (pipelineRun) {
    return latestRun.spec.pipelineRef?.name ??
        (latestRun.metadata.annotations?.preferredName || latestRun.metadata.name)
  }
  return null;
};
```
Nit: You could write this in a fashion where it's flatter and easier to compartmentalize the logic.

Not required, just a thought.

andrewballantyne (commented):
Still need the annotation change, and then I can run it through the paces.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: this does seem to be easier to read and compartmentalize the logic.
updated.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
```diff
@@ -18,6 +18,9 @@ import { CREATE_PIPELINE_RESOURCE, initialResourceFormValues } from './const';
 import { CommonPipelineModalFormikValues, PipelineModalFormResource } from './types';
 import { TektonResourceLabel, VolumeTypes } from '../../const';
 
+// Annotation for referencing pipeline name in case of PipelineRun with no reference to a Pipeline (embedded pipeline)
+export const preferredNameAnnotation = 'io.openshift.pipeline/preferredName';
```
karthikjeeyar, line 22: We have an  **started-by** annotations already, it is like this  `pipeline.openshift.io/started-by`. We should be consistent here
```suggestion
export const preferredNameAnnotation = 'pipeline.openshift.io/preferredName';
```
and also move this to `src/components/pipelines/const.ts`

frontend/packages/pipelines-plugin/src/test-data/pipeline-data.ts
```diff
@@ -25,6 +25,7 @@ export enum PipelineExampleNames {
   CONDITIONAL_PIPELINE = 'conditional-pipeline',
   INVALID_PIPELINE_MISSING_TASK = 'missing-task-pipeline',
   INVALID_PIPELINE_INVALID_TASK = 'invalid-task-pipeline',
+  EMBEDDED_PIPELINESPEC = 'embedded-spec-pipeline-run',
```
karthikjeeyar, line 28: nit:
```suggestion
  EMBEDDED_PIPELINE_SPEC = 'embedded-pipeline-spec',
```

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts
nemesis09: done

frontend/packages/pipelines-plugin/src/test-data/pipeline-data.ts
nemesis09: done

spadgett:

frontend/packages/pipelines-plugin/src/test-data/pipeline-data.ts
```diff
@@ -1885,6 +1886,115 @@ export const pipelineTestData: PipelineTestData = {
       },
     },
   },
+  [PipelineExampleNames.EMBEDDED_PIPELINE_SPEC]: {
+    dataSource: 'embedded-pipelineSpec',
+    pipeline: {
```
andrewballantyne, line 1891:
```suggestion
    pipeline: null,
```
You're in conflict here. You crafted a Pipeline for a test chunk that says it has an embedded pipeline 🙂

andrewballantyne (commented):
Code looks pretty good. Going to test it on a cluster now before approving.

frontend/packages/pipelines-plugin/src/test-data/pipeline-data.ts
nemesis09: it indeed is conflicting. Thanks for the catch.
updated the test data

andrewballantyne (approved):
Nice work with the tests Aritra!

nemesis09:
> Nice work with the tests Aritra!

thank you :)

karthikjeeyar:
Thanks @nemesis09  Verified locally, works fine. 
/lgtm

karthikjeeyar:
@nemesis09 Since pipeline has its own package in 4.7, we need to manually cherrypick this code to 4.6

PR #7925, opened 2021-01-26
Bug 1919387: Wait till the DOM is loaded when writing to the body localStorage issue

jhadvig:
/assign @spadgett

### DIFF
--- a/frontend/public/tokener.html
+++ b/frontend/public/tokener.html
@@ -2,33 +2,35 @@
 <html lang="en">
   <head>
     <script type="text/javascript">
-      // eslint-disable-next-line no-var
-      var json = [[.]];
-      var error;
-      try {
-        ['exp', 'email', 'name', 'userID'].forEach(function (key) {
-          localStorage.setItem(key, json[key]);
-        });
-      } catch (e) {
-        error = e;
-        console.error(e);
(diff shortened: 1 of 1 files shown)


frontend/public/tokener.html
```diff
@@ -2,33 +2,35 @@
...
+          ['exp', 'email', 'name', 'userID'].forEach(function (key) {
+            localStorage.setItem(key, json[key]);
+          });
+        } catch (e) {
+          error = e;
+          console.error(e);
+          document.body.append('localStorage must be enabled:');
+          document.body.append(e.message || e.toString());
+        }
+        if (!error) {
```
spadgett, line 19: Digging into the code, I don't believe the `localStorage` items are strictly required to continue, particularly for OpenShift where we user information is not stored there. We track login state in a cookie, and user preferences have moved to a config map. We might just continue and not fail on an error.

jhadvig: @spadgett so do we want to keep the `onload()` method? Or just check if the `document.body` is not `null` when trying to write into it ?

spadgett:

spadgett (approved):
/lgtm

Let's go ahead with this fix for now, and we can consider trying to continue even after the error in a follow up.

PR #7930, opened 2021-01-26
Bug 1920524: Fix topology crash on install of ODH operator to project

jeff-phillips-18:
**Fixes**: 
Fixes https://issues.redhat.com/browse/ODC-5398

**Analysis / Root cause**: 
Operator backed service groups with multiple workload dependencies got added to the model prior to completion of the reconciliation. This caused workloads that reconciled later to create a new OBS group parented to itself causing a stack overflow when drawing the groups.

**Solution Description**: 
Do not add the OBS groups to the model until all workloads have been reconciled.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/topology/src/operators/operatorsDataModelReconciler.ts
+++ b/frontend/packages/topology/src/operators/operatorsDataModelReconciler.ts
@@ -57,6 +57,7 @@ export const operatorsDataModelReconciler = (
   }
   const defaultIcon = getImageForIconClass(`icon-openshift`);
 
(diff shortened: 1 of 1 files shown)


jerolimov (approved):
Could reproduce the crash without this change; and not with this change. :)

/lgtm

rohitkrai03 (approved):
/approve

PR #7931, opened 2021-01-26
Bug 1920551: Should not allow to edit boot order of common templates

yaacov:
Should not allow to edit boot order of common templates

Screenshot:
before:
screenshot-localhost_9000-2021 01 26-16_46_13

after:
screenshot-localhost_9000-2021 01 26-16_44_51


yaacov:
cc:// @lmandavi

PR #7932, opened 2021-01-26
Bug 1919750: Fix exception on InstallPlan list page

TheRealJon:
The Table component Row prop is not defined as a React component, so we were getting an exception for using React hooks in InstallPlanTableRow component and passing it directly to the Table Row prop. I worked around this by wrapping the InstallPlanTableRow component in a callback. This fixes the bug, but at some point we should also refactor the Table Row prop to be a real React component.


spadgett (approved):
/lgtm

PR #7935, opened 2021-01-26
adding yaml templates extension types to dynamic-plugin-sdk

glekner:
Adding extension types for YAMLTemplate extension.
This Depends on #7919 #7898﻿

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
@@ -1,0 +1,27 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+
+export namespace ExtensionProperties {
+  export type YAMLTemplate = {
+    /** Model associated with the template. */
+    model: {
+      group: string;
+      version: string;
+      kind: string;
+    };
+    /** The YAML template. */
+    template: string;
+    /** The name of the template. Use the name `default` to mark this as the default template. */
+    name: string | 'default';
+  };
+}
+
+// Extension types
+
+export type YAMLTemplate = Extension<ExtensionProperties.YAMLTemplate> & {
+  type: 'console.yaml-template';
+};
+
+// Type guards
+
+export const isYAMLTemplate = (e: Extension): e is YAMLTemplate =>
+  e.type === 'console.yaml-template';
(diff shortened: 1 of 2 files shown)


spadgett:

frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
```diff
@@ -0,0 +1,24 @@
...
+    template: string;
+    /** The name of the template. If not specified, use the `default` value. */
+    templateName?: string;
+  };
+}
+
+// Extension types
+
+export type YAMLTemplate = Extension<ExtensionProperties.YAMLTemplate> & {
+  type: 'console.yaml/template';
```
christianvogt, line 18: Since this isn't a sub type of the `console.yaml` extension, it should probably as follows:
```suggestion
  type: 'console.yaml-template';
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
```diff
@@ -0,0 +1,24 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+import { K8sKind } from 'public/module/k8s/types';
+
+export namespace ExtensionProperties {
+  export type YAMLTemplate = {
+    /** Model associated with the template. */
+    model: K8sKind;
```
christianvogt, line 7: To be consistent with other model references in the new extension format:
```suggestion
model: {
  group: string,
  version: string,
  kind: string,
}
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
```diff
@@ -0,0 +1,24 @@
...
+    /** The YAML template to use. */
```
christianvogt, line 8:
```suggestion
    /** The YAML template. */
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/yaml-templates.ts
```diff
@@ -0,0 +1,24 @@
...
+    template: string;
+    /** The name of the template. If not specified, use the `default` value. */
+    templateName?: string;
```
christianvogt, line 11: Simplify to `name`.
I prefer to force the user to mark the template as the `default` instead of simply omitting the name.
```suggestion
    /** The name of the template. Use the name `default` to mark this as the default template. */
    name: string | 'default';
```

glekner:
@christianvogt fixed suggestions

PR #7936, opened 2021-01-26
Bug 1920684: Update language preferences modal

rebeccaalpert:
Updated text in language preferences modal since all text doesn't actually update.

CC @spadgett.

### DIFF
--- a/frontend/public/components/modals/language-preferences-modal.tsx
+++ b/frontend/public/components/modals/language-preferences-modal.tsx
@@ -36,13 +36,8 @@ const LanguagePreferencesModal = (props: LanguagePreferencesModalProps) => {
 
   return (
     <form onSubmit={submit} name="form" className="modal-content modal-content--no-inner-scroll">
-      <ModalTitle>{t('language-preferences-modal~Edit language preferences')}</ModalTitle>
+      <ModalTitle>{t('language-preferences-modal~Edit language preference')}</ModalTitle>
       <ModalBody>
-        <p>
-          {t(
-            'language-preferences-modal~All interface content will appear in your selected language.',
-          )}
-        </p>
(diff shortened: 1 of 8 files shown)


dtaylor113 (approved):
/lgtm

frontend/public/components/modals/language-preferences-modal.tsx
```diff
@@ -38,13 +38,10 @@ const LanguagePreferencesModal = (props: LanguagePreferencesModalProps) => {
...
       <ModalBody>
-        <p>
-          {t(
-            'language-preferences-modal~All interface content will appear in your selected language.',
-          )}
-        </p>
         <div className="form-group">
-          <label htmlFor="language_dropdown">{t('language-preferences-modal~Language')}</label>
+          <label htmlFor="language_dropdown">
+            {t('language-preferences-modal~Select language:')}
```
spadgett, line 43: I'd probably leave this just as "Language"

We might change the user menu item and dialog title to the simpler "Select language" though. "Edit language preferences" makes it sound like there are several preferences. You're really just picking a language.

cc @beanh66 for input

spadgett:

frontend/public/components/modals/language-preferences-modal.tsx
rebeccaalpert: I spoke to Brie and Meghan and Meghan is going to mock some things up and share them @spadgett. (Meghan previously worked on the design for this component.)

rebeccaalpert:
Putting on hold because text is under review.

rebeccaalpert:
@spadgett - I updated the text per Megan's comments.

spadgett (approved):
/lgtm

PR #7942, opened 2021-01-27
Fix flakiness in OCS Operator Installation test

bipuladh:
- Add a reload before going to the storage Cluster Creation Page


frontend/packages/ceph-storage-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -50,7 +50,10 @@ Cypress.Commands.add('install', (mode: 'Internal' | 'Attached' = 'Internal', enc
 
       // Make changes to this once we add annotation
       cy.log(`Install OCS in ${mode} Mode`);
-      commonFlows.navigateToOCS(true);
+      // Reload because StorageCluster CRD is not registered in the UI; hence getting 404 Error
+      cy.visit('/');
+      cy.reload(true);
+      commonFlows.navigateToOCS();
```
cloudbehl, line 56: I have seen this happening after reload as well lot of times. Can we add some 20sec and do a reload?

PR #7945, opened 2021-01-27
Bug 1921087: Next quick start link update

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-5349

# Issue
'start next quick start' link doesn't work and is unintuitive
If the next quick start is not installed, then the link doesn't actually work.
It's also not clear to the user as to what the next quick start is since it shows the ID of the quick start and not the quick start label.

# Screenshots
quickstartnext

# Tests
Updates `QuickStartContent` and `QuickStartConclusion`

# Browser conformance
Chrome

### DIFF
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
@@ -6,12 +6,14 @@ import { QuickStartContext, QuickStartContextValues } from './utils/quick-start-
 
 type QuickStartControllerProps = {
   quickStart: QuickStart;
+  nextQuickStarts?: QuickStart[];
   footerClass: string;
   contentRef: React.Ref<HTMLDivElement>;
 };
 
 const QuickStartController: React.FC<QuickStartControllerProps> = ({
   quickStart,
+  nextQuickStarts,
   contentRef,
   footerClass,
 }) => {
@@ -66,6 +68,7 @@ const QuickStartController: React.FC<QuickStartControllerProps> = ({
     <>
       <QuickStartContent
         quickStart={quickStart}
+        nextQuickStarts={nextQuickStarts}
         taskNumber={taskNumber}
         allTaskStatuses={allTaskStatuses}
         onTaskSelect={handleTaskSelect}
--- a/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
@@ -32,6 +32,9 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   const shadows = useScrollShadows(contentRef);
 
   const quickStart = quickStarts.find((qs) => qs.metadata.name === activeQuickStartID);
+  const nextQuickStarts: QuickStart[] = quickStarts.filter((qs: QuickStart) =>
+    quickStart?.spec.nextQuickStart?.includes(qs.metadata.name),
+  );
 
   const headerClasses = classNames({
     'pf-u-box-shadow-sm-bottom': shadows === Shadows.top || shadows === Shadows.both,
@@ -68,6 +71,7 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
         <AsyncComponent
           loader={() => import('./QuickStartController').then((c) => c.default)}
           quickStart={quickStart}
+          nextQuickStarts={nextQuickStarts}
           footerClass={footerClass}
           contentRef={setContentRef}
         />
--- a/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
@@ -8,6 +8,7 @@ import './QuickStartContent.scss';
 
 type QuickStartContentProps = {
   quickStart: QuickStart;
+  nextQuickStarts?: QuickStart[];
   taskNumber: number;
   allTaskStatuses: QuickStartTaskStatus[];
   onTaskSelect: (selectedTaskNumber: number) => void;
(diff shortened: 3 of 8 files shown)


frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -67,7 +72,7 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
       <DrawerPanelBody hasNoPadding className="co-quick-start-panel-content__body">
         <AsyncComponent
           loader={() => import('./QuickStartController').then((c) => c.default)}
-          quickStart={quickStart}
+          quickStart={{ ...quickStart, nextQuickStarts }}
```
jerolimov, line 75: Hmmmm, why you don't added a second prop `nextQuickStarts` here and in the `QuickStartController`?

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -5,7 +5,7 @@ import QuickStartFooter from './controller/QuickStartFooter';
 import { QuickStartContext, QuickStartContextValues } from './utils/quick-start-context';
 
 type QuickStartControllerProps = {
-  quickStart: QuickStart;
+  quickStart: QuickStart & { nextQuickStarts?: QuickStart[] };
```
jerolimov, line 8: Why not a second `nextQuickStarts` prop here? Any benefit of reusing a QuickStart object which is now a common QuickStart object?

```suggestion
  quickStart: QuickStart;
  nextQuickStarts?: QuickStart[];
```

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
```diff
@@ -7,7 +7,7 @@ import QuickStartConclusion from './QuickStartConclusion';
 import './QuickStartContent.scss';
 
 type QuickStartContentProps = {
-  quickStart: QuickStart;
+  quickStart: QuickStart & { nextQuickStarts?: QuickStart[] };
```
jerolimov, line 10:
```suggestion
  quickStart: QuickStart;
  nextQuickStarts?: QuickStart[];
```

frontend/packages/console-app/src/components/quick-starts/controller/QuickStartContent.tsx
```diff
@@ -21,10 +21,11 @@ const QuickStartContent = React.forwardRef<HTMLDivElement, QuickStartContentProp
     ref,
   ) => {
     const {
-      spec: { introduction, tasks, conclusion, nextQuickStart = [] },
+      nextQuickStarts = [],
```
jerolimov, line 24: Extract it from the props instead

jerolimov (changes requested):
I don't see the benefit of not adding a second prop for nextQuickStart. Wdyt?

spadgett:

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
```diff
@@ -32,6 +32,11 @@ const QuickStartPanelContent: React.FC<QuickStartPanelContentProps> = ({
   const shadows = useScrollShadows(contentRef);
 
   const quickStart = quickStarts.find((qs) => qs.metadata.name === activeQuickStartID);
+  const nextQuickStarts: QuickStart[] = [];
+  quickStart?.spec.nextQuickStart?.forEach((name) => {
+    const next = quickStarts.find((qs) => qs.metadata.name === name);
+    next && nextQuickStarts.push(next);
+  });
```
invincibleJai, line 39: can we have somethings as below?

```suggestion
  const nextQuickStarts: QuickStart[] = quickStarts.filter((qs:QuickStart) => 
     quickStart?.spec.nextQuickStart?.includes(qs.metadata.name));
```

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
```diff
@@ -6,12 +6,14 @@ import { QuickStartContext, QuickStartContextValues } from './utils/quick-start-
 
 type QuickStartControllerProps = {
   quickStart: QuickStart;
+  nextQuickStarts?: QuickStart[];
```
invincibleJai, line 9: do we need this to be a list or just `QuickStart` ?

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
debsmita1: +1

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
abhinandan13jan: A quickStart could have multple `nextQuickStarts` as dependencies. Although we only make use of the first one as of today. But, lets not lose this data for future needs.

frontend/packages/console-app/src/components/quick-starts/QuickStartPanelContent.tsx
abhinandan13jan: updated

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
rohitkrai03: We should show all the next quick starts. Can you make that change and just loop over all the next quick starts to render all of them?

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/quick-starts/QuickStartController.tsx
andrewballantyne: 💯 agree... if we have it we should show it. Maybe that's a 4.8 thing and not part of this bug though.

abhinandan13jan: PR is ready for re-review with single `nextQuickStart` @rohitkrai03 @invincibleJai

jerolimov (approved):
/lgtm

PR #7947, opened 2021-01-27
Bug 1878086: Disables filesystem name input field in StorageClassForm

vbnrh:
when creating a storageclass with cephfs provisoner

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/utils/ocs-storage-class-params.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/utils/ocs-storage-class-params.tsx
@@ -105,7 +105,8 @@ export const StorageClassFormProvisoners: ExtensionSCProvisionerProp = Object.fr
         fsName: {
           name: 'Filesystem Name',
           hintText: 'CephFS filesystem name into which the volume shall be created',
-          required: true,
+          value: 'ocs-storagecluster-cephfilesystem',
+          visible: () => false,
         },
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/utils/ocs-storage-class-params.tsx
```diff
@@ -105,7 +105,9 @@ export const StorageClassFormProvisoners: ExtensionSCProvisionerProp = Object.fr
         fsName: {
           name: 'Filesystem Name',
           hintText: 'CephFS filesystem name into which the volume shall be created',
-          required: true,
+          required: false,
```
cloudbehl, line 108:
```suggestion
```

cloudbehl: not required flag.

vbnrh: removed

cloudbehl:
@vbnrh Can you be concise with the commit message description and not exceed the limit. if you need to provide more data then try to use the summary.

vbnrh:
> @vbnrh Can you be concise with the commit message description and not exceed the limit. if you need to provide more data then try to use the summary.

Sorry, I was not aware of any commit msg limits. Used to updating with long commit msgs in the operator =) I've fixed it now.

PR #7948, opened 2021-01-27
Bug 1921774: fix delete app modal error when resource not found

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5360

**Root analysis:**
before deleting a resource it is not being checked if the resource exists at all

**Solution description:**
ignoring the 404 error(if occurs) while deleting the resource that doesn't exist

**GIF:**
dele

**Browser conformance:** 
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/topology/src/utils/application-utils.ts
+++ b/frontend/packages/topology/src/utils/application-utils.ts
@@ -2,7 +2,6 @@ import * as _ from 'lodash';
 import i18next from 'i18next';
 import {
   K8sKind,
-  k8sGet,
   k8sList,
   k8sPatch,
   k8sKill,
@@ -131,7 +130,14 @@ const safeKill = async (model: K8sKind, obj: K8sResourceKind) => {
     namespace: obj.metadata.namespace,
   });
   if (resp.status.allowed) {
-    return k8sKill(model, obj);
+    try {
+      return await k8sKill(model, obj);
+    } catch (error) {
+      // 404 when resource is not found
+      if (error?.response?.status !== 404) {
+        throw error;
+      }
+    }
   }
   return null;
 };
@@ -171,13 +177,6 @@ export const cleanUpWorkload = async (
   isKnativeResource: boolean,
 ): Promise<K8sResourceKind[]> => {
   const reqs = [];
-  const isImageStreamPresent = await k8sGet(
-    ImageStreamModel,
-    resource.metadata.name,
-    resource.metadata.namespace,
-  )
-    .then(() => true)
-    .catch(() => false);
   const buildConfigs = await k8sList(BuildConfigModel, { ns: resource.metadata.namespace });
   const builds = await k8sList(BuildModel, { ns: resource.metadata.namespace });
   const resources = {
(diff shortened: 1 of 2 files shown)


frontend/packages/topology/src/utils/application-utils.ts
```diff
@@ -180,6 +180,21 @@ export const cleanUpWorkload = async (
     .catch(() => false);
   const buildConfigs = await k8sList(BuildConfigModel, { ns: resource.metadata.namespace });
   const builds = await k8sList(BuildModel, { ns: resource.metadata.namespace });
+  const isServicePresent = await k8sGet(
+    ServiceModel,
+    resource.metadata.name,
+    resource.metadata.namespace,
+  )
+    .then(() => true)
+    .catch(() => false);
```
christianvogt, line 189: This is a bad pattern to get into. I see that this was a continuation of what was previously implemented. A GET followed by DELETE is still racey.
We should simply delete the resource and ignore 404 errors because the resource doesn't exist.

frontend/packages/topology/src/utils/application-utils.ts
```diff
@@ -131,7 +130,11 @@ const safeKill = async (model: K8sKind, obj: K8sResourceKind) => {
     namespace: obj.metadata.namespace,
   });
   if (resp.status.allowed) {
-    return k8sKill(model, obj);
+    try {
+      return await k8sKill(model, obj);
+    } catch (ignored) {
+      // 404 expected
+    }
```
jerolimov, line 137: If would expect if this would check for the real status code. Maybe something like that? (untested code!)

```suggestion
    } catch (error) {
      if (error.response?.status === 404) {
        // ignore resource which could not be found.
      } else {
        throw error;
      }
    }
```

frontend/packages/topology/src/utils/application-utils.ts
```diff
@@ -195,16 +193,14 @@ export const cleanUpWorkload = async (
   const resourceBuildConfigs = getBuildConfigsForResource(resource, resources);
   const isBuildConfigPresent = !_.isEmpty(resourceBuildConfigs);
 
-  const deleteModels = [ServiceModel, RouteModel];
+  const deleteModels = [ServiceModel, RouteModel, ImageStreamModel];
   const knativeDeleteModels = [KnativeServiceModel];
```
jerolimov, line 197: You dropped below the `isImageStreamPresent` condition where `ImageStreamModel` was added to `deleteModels` and `knativeDeleteModels`. So I expect that it's required that you add this here also to the `knativeDeleteModels`, right?

jerolimov (changes requested):
See change requests above.

frontend/packages/topology/src/utils/application-utils.ts
debsmita1: I missed adding the `ImageStreamModel` to the knativeDeleteModels. Thanks !

invincibleJai:
/approve 

jerolimov (approved):
Works fine

/lgtm

PR #7949, opened 2021-01-27
Bug 1920320: Show link to common template in vm details view

yaacov:
We now consider common templates as first class template citizens, we need to link to them from details page.

Screenshot:
Before:
screenshot-localhost_9000-2021 01 27-17_30_49

After:
screenshot-localhost_9000-2021 01 27-17_30_15


frontend/packages/kubevirt-plugin/src/components/vms/vm-resource.tsx
```diff
@@ -91,11 +101,25 @@ export const VMResourceSummary: React.FC<VMResourceSummaryProps> = ({
   kindObj,
 }) => {
   const { t } = useTranslation();
+  const templatesResource: WatchK8sResource = {
```
rawagner, line 104: We should load the templates together with user templates at `vm-details-page`

frontend/packages/kubevirt-plugin/src/components/vms/vm-resource.tsx
```diff
@@ -91,11 +101,25 @@ export const VMResourceSummary: React.FC<VMResourceSummaryProps> = ({
...
 
   const isVM = kindObj === VirtualMachineModel;
   const vmiLike = isVM ? vm : vmi;
 
-  const template = getVMTemplate(vm, templates);
+  const isCommonTemplate =
+    getLabel(vmiLike, LABEL_USED_TEMPLATE_NAMESPACE) === TEMPLATE_VM_COMMON_NAMESPACE;
+  const template = isCommonTemplate
+    ? getVMCommonTemplate(vmiLike, commonTemplates)
+    : getVMTemplate(vmiLike, templates);
```
rawagner, line 122: this change is not needed when you fetch common templates with user templates

rawagner (changes requested):

frontend/packages/kubevirt-plugin/src/components/vms/vm-resource.tsx
yaacov: removed the `WatchK8sResource` hook and now using the firehose mechanism for both `templats` and `commonTemplates`

frontend/packages/kubevirt-plugin/src/components/vms/vm-resource.tsx
yaacov: in the testing environment we have VMs that do not have the template name label when created from common template, so the common template can't use the same method, using the  `getVMCommonTemplate` to catch this cases.

yaacov:
@rawagner please re review

frontend/packages/kubevirt-plugin/src/selectors/vm-template/selectors.ts
```diff
@@ -26,6 +33,25 @@ export const getVMTemplate = (
     : undefined;
 };
 
+export const getVMCommonTemplate = (
```
rawagner, line 36: you should still find the template by name and namespace. If these name/namespace labels are missing than you can fallback to label/workload/os.

rawagner (changes requested):

frontend/packages/kubevirt-plugin/src/selectors/vm-template/selectors.ts
yaacov: added the regular check first

yaacov:
@rawagner please re review

spadgett:

frontend/packages/kubevirt-plugin/src/components/vms/vmDetailsItemTemplate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const templatesResource: WatchK8sResource = {
+    isList: false,
+    optional: true,
+    kind: TemplateModel.kind,
+    name,
+    namespace,
+    selector: {
+      matchLabels: {
+        [TEMPLATE_TYPE_LABEL]:
+          namespace === TEMPLATE_VM_COMMON_NAMESPACE ? TEMPLATE_TYPE_BASE : TEMPLATE_TYPE_VM,
```
rawagner, line 30: since you have name and namespace, there's no need to match by label too. BTW in theory you can have user template in openshift namespace too.

frontend/packages/kubevirt-plugin/src/components/vms/vmDetailsItemTemplate.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from '../../constants';
+import { getBasicID, prefixedID } from '../../utils';
+import { VMTemplateLink } from '../vm-templates/vm-template-link';
+import VMDetailsItem from './vmDetailsItem';
+
+const VMDetailsItemTemplate: React.FC<VMDetailsItemTemplateProps> = ({ name, namespace }) => {
+  const { t } = useTranslation();
+  const templatesResource: WatchK8sResource = {
+    isList: false,
+    optional: true,
```
rawagner, line 23: no need for optional. The hook doesnt really care.

frontend/packages/kubevirt-plugin/src/components/vms/vmDetailsItemTemplate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [template, loadedTemplates, errorTemplates] = useK8sWatchResource<TemplateKind>(
+    templatesResource,
+  );
+  const id = getBasicID(template);
+
+  return (
+    <VMDetailsItem
+      title={t('kubevirt-plugin~Template')}
+      idValue={prefixedID(id, 'template')}
+      isNotAvail={!name || !namespace || !loadedTemplates || errorTemplates}
```
rawagner, line 43: you need some loading state. Showing `Not available` if `!loadedTemplates` is not good

rawagner (changes requested):

frontend/packages/kubevirt-plugin/src/components/vms/VMDetailsItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </dt>
+      <dd id={idValue} className={valueClassName}>
+        {isNotAvail && (
+          <span className="text-secondary">
+            {isNotAvailMessage || t('kubevirt-plugin~Not available')}
+          </span>
+        )}
+        {isLoading && !isNotAvail && <LoadingInline />}
+        {!isNotAvail && !isLoading && children}
+      </dd>
```
rawagner, line 46: I find these conditions hard to follow. Can you do something like

```suggestion
if (isNotAvail) {
  body = (
    <span className="text-secondary">
      {isNotAvailMessage || t('kubevirt-plugin~Not available')}
    </span>
  );
} else if (isLoading) {
  body = <LoadingInline />;
} else {
  body = children;
}



....
      <dd id={idValue} className={valueClassName}>
         {body}
      </dd>
```

frontend/packages/kubevirt-plugin/src/components/vms/VMDetailsItemTemplate.tsx
```diff
@@ -0,0 +1,43 @@
...
+    templatesResource,
+  );
+  const id = getBasicID(template);
+
+  return (
+    <VMDetailsItem
+      title={t('kubevirt-plugin~Template')}
+      idValue={prefixedID(id, 'template')}
+      isLoading={!loadedTemplates}
+      isNotAvail={!name || !namespace || !template || errorTemplates}
```
rawagner, line 31:
```suggestion
      isNotAvail={!name || !namespace || (loadedTemplates && !template) || errorTemplates}
```

rawagner (changes requested):

PR #7951, opened 2021-01-27
Bug 1920771: Internationalize Delete Persistent Volume Claim

rebeccaalpert:
This is a special case kebab item not addressed as part of the Storage i18n PR.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1920771.


spadgett (approved):
/lgtm

PR #7952, opened 2021-01-27
Bug 1919823: Update zh-cn translation for Home

rebeccaalpert:
The translation team recommends updating the current translation for "Home" of 家 to 主页. I'm updating it manually so we can close out the associated bug earlier. They will update the files on their end so the old translation won't creep back in with the next batch of translations.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1919823.


spadgett (approved):
/lgtm

PR #7955, opened 2021-01-27
Bug 1921267: i18n ResourceListDropdown

rebeccaalpert:
It looks like the ResourceListDropdown wasn't internationalized previously. It doesn't seem to be included in the IBM clean-up PRs.

There are some related issues (chip groups, resource labels, etc.) that need to be updated as well if this goes through as-is. I'll do those as a follow-on.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921267.


frontend/public/components/resource-dropdown.tsx
```diff
@@ -106,7 +113,7 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
       return false;
     }
 
-    return fuzzy(_.toLower(text), _.toLower(model.kind));
+    return fuzzy(_.toLower(text), model.labelKey ? t(model.labelKey) : _.toLower(model.kind));
```
spadgett, line 116: You'll want to `toLower` the translated value too. Otherwise it will *never* match since we're calling `_.toLower` on `text`. It might be good to use `toLocaleLowerCase`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase

spadgett (changes requested):

frontend/public/components/resource-dropdown.tsx
rebeccaalpert: Good call; thanks for the tip.

frontend/public/components/resource-dropdown.tsx
```diff
@@ -106,7 +114,12 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
       return false;
     }
 
-    return fuzzy(_.toLower(text), _.toLower(model.kind));
+    return fuzzy(
+      _.toLower(text),
+      model.labelKey
+        ? t(model.labelKey).toLocaleLowerCase(i18next.language)
```
spadgett, line 120: We should use `toLocaleLowerCase` instead of `_.toLower` consistently here

spadgett:
Looks like we have a build error. (Also PR needs rebase.)

```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/resource-dropdown.tsx(121,48):
TS2551: Property 'langugae' does not exist on type 'i18n'. Did you mean 'language'?```

Thanks @rebeccaalpert

**Comment by rebeccaalpert** [comment]
Should be all set.

**Comment by rebeccaalpert** [comment]
Putting on hold since I realized we can't rely on i18next.language. In cases where the language is a subset of a language or a language we don't support, i18next.language won't return the active language. We need to key off of the values in i18next.languages instead. This will have a tree of the user's language as well as the default tree.

I'll rework this to use more accurate logic.

**Comment by rebeccaalpert** [comment]
Will move forward once an approach is agreed on for #8009.

**Comment by rebeccaalpert** [comment]
I updated the ResourceListDropdown and language selector modal to use a shared language finding function.

File: frontend/public/components/resource-dropdown.tsx [hunk 3]
```diff
@@ -102,11 +109,17 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
 
   const autocompleteFilter = (text, item) => {
     const { model } = item.props;
+    const language = getActiveLanguage();
```
spadgett, line 112: I'm debating if this is the right way to set the language. In other contexts like `Intl.NumberFormat`, you wouldn't want to initialize the locale this way since we want to show the numbers based on the user's locale regardless of whether we have translations for it or now. I guess this context is a little different, however.

My other concern is that this helper has a loop in it. The logic in this function is going to run a lot (once for every resource and for every keystroke), so adding this here could have an impact on the performance.

I would probably just not pass the language to `toLocaleLowerCase` here and let it initialize from the user's system locale or just go back to `toLowerCase`.

rebeccaalpert: I'm just going to go back to toLowerCase. toLocaleLowerCase says the output will be similar in most cases except for things like Turkish, which we don't currently support anyway.

rebeccaalpert:
Feedback all set.

spadgett (approved):
/lgtm

PR #7957, opened 2021-01-27
Bug 1913969: Fix edge case exception for fieldDependency spec descriptor

TheRealJon:
The form field sort ordering logic was causing a stack overflow when the control field an dependent field had the same name. Resolved by keeping track of the current path and control field path to differentiate between the two. This also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1890180

### DIFF
--- a/frontend/packages/console-shared/src/components/dynamic-form/utils.ts
+++ b/frontend/packages/console-shared/src/components/dynamic-form/utils.ts
@@ -8,6 +8,12 @@ import { THOUSAND, MILLION, BILLION } from './const';
 
 const UNSUPPORTED_SCHEMA_PROPERTIES = ['allOf', 'anyOf', 'oneOf'];
 
+// Transform a path string to a JSON schema path array
+export const stringPathToUISchemaPath = (path: string): string[] =>
+  (_.toPath(path) ?? []).map((subPath) => {
+    return /^\d+$/.test(subPath) ? 'items' : subPath;
+  });
+
 export const useSchemaLabel = (schema: JSONSchema6, uiSchema: UiSchema, defaultLabel?: string) => {
   const options = getUiOptions(uiSchema ?? {});
   const showLabel = options?.label ?? true;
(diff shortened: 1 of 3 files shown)


spadgett:

frontend/packages/console-shared/src/components/dynamic-form/utils.ts
```diff
@@ -135,9 +156,16 @@ const getJSONSchemaPropertySortWeight = (
   // Total offset to be added to base tier
   const offset = controlFieldOffset + uiSortOrder;
 
-  // If this property is a dependent, it's weight is based on it's control field
-  if (controlFieldName) {
-    return getJSONSchemaPropertySortWeight(controlFieldName, jsonSchema, uiSchema) + offset;
+  // If this property or it's children have a control field at the current path, it's weight is
```
spadgett, line 159:
```suggestion
  // If this property or its children have a control field at the current path, it's weight is
```

frontend/packages/console-shared/src/components/dynamic-form/utils.ts
```diff
@@ -175,15 +210,21 @@ export const getJSONSchemaOrder = (jsonSchema, uiSchema) => {
...
       .toJS();
 
     return {
       ...(uiOrder.length > 1 && { 'ui:order': uiOrder }),
       ..._.reduce(
         jsonSchema?.properties ?? {},
         (orderAccumulator, propertySchema, propertyName) => {
-          const descendantOrder = getJSONSchemaOrder(propertySchema, uiSchema?.[propertyName]);
+          const descendantOrder = getJSONSchemaOrder(
+            propertySchema as JSONSchema6,
```
spadgett, line 224: Can you simply declare the type above in the function definition like

```ts
(orderAccumulator, propertySchema: JSONSchema6, propertyName) => {
```

This is more type safe than using `as`

spadgett (approved):
Two minor nits, but I wouldn't hold the PR on either comment.

/lgtm

TheRealJon:
Fixes #8120

PR #7958, opened 2021-01-27
Bug 1921258: use sessionStorage for user-settings when impersonating

christianvogt:
**Analysis / Root cause**: 
When logged in as kubeadmin and impersonating a user, the access to the user-settings configmap for kubeadmin returns a 403 which puts the user settings hook in a bad state and will update whenever another tab is opened for the kubeadmin user and their settings are updated.

**Solution Description**: 
When impersonating a user, use session storage as the backing store instead of configmap or local storage.

No UI change.

Added unit test which checks for session storage usage when impersonating.

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/__tests__/useUserSettings.spec.ts
+++ b/frontend/packages/console-shared/src/hooks/__tests__/useUserSettings.spec.ts
@@ -1,4 +1,7 @@
-import * as redux from 'react-redux';
+// FIXME upgrading redux types is causing many errors at this time
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useSelector } from 'react-redux';
 import { act } from 'react-dom/test-utils';
 import { ConfigMapKind } from '@console/internal/module/k8s';
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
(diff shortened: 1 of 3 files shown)


invincibleJai:
@christianvogt  Verified it i no longer observed Switch in perspective when tried on one window with impersonate and another with kubeadmin. 

Storing in session works as expected in chrome however in safari it doesn't seems to work as expected. That looks like a separate issue LMK if needed will log one


Jan-29-2021 17-17-32

christianvogt:
@invincibleJai thanks. I cannot test safari until i update my OS since my safari is too old and doesn't support `ResizeObserver` and the console errors out ;/

invincibleJai:
> @invincibleJai thanks. I cannot test safari until i update my OS since my safari is too old and doesn't support `ResizeObserver` and the console errors out ;/

i verified again and it works fine in Safari as well earlier i was trying to look in storage and din't seem to get it reflect there then but seems fine now

image

christianvogt:
@invincibleJai that's for the review and test.
I did just notice from your screenshot that you are using the dev setup where you choose to store settings in local storage instead of configmaps. When we enable this setting, the user-settings key in the local storage is always set to `console-user-settings` as oppose to `console-user-settings-<userid>`. I fixed the key when impersonating.

PR #7964, opened 2021-01-28
Bug 1921644: [e2e][automation] tests errors with wrong cloudInit new line syntax

glekner:


spadgett:

PR #7965, opened 2021-01-28
Bug 1921713: Fix migration tests

rawagner:


rawagner:
/hold

I need to update the PR

rawagner:
/hold cancel
@gouyang please take a look again. There was another problem where `waitForStatus` uses conditional waiting which is not always correct. `until.textToBePresentInElement` which was used before this change checks if the text is _included_ in the status. But we need to check if status _startsWith_ which is more precise

spadgett:

PR #7966, opened 2021-01-28
Bug 1919876: Updated accessMode and volumeMode for common-template without source pvc from storage class if exist

metalice:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1919876

**Analysis / Root cause**: 
The advanced wizard didn't take values from the storage class config map (only default if there isn't any storage class)

**Solution Description**: 
The advanced wizard is now taking values of volumeMode and accessMode from storage class config map if its not source PVC and storage class is present.


metalice:
@rawagner @yaacov please have a look

rawagner:
doesnt work for me.
When you're creating from common template, you dont want to get storageclass from existing datavolue, it is always default. You want to get the storageclass from the actual PVC (base image or PVC that user chose to clone)

/lgtm cancel

spadgett:

PR #7967, opened 2021-01-28
Bug 1921780: Clean up Search i18n

rebeccaalpert:
Many Search page-specific components were not internationalized. This includes chips, resource accordion headings, the label/name filter, and the "clear all filters" text. I updated these items.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921780.

### DIFF
--- a/frontend/public/components/search-filter-dropdown.tsx
+++ b/frontend/public/components/search-filter-dropdown.tsx
@@ -1,10 +1,13 @@
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
 import { CaretDownIcon, FilterIcon } from '@patternfly/react-icons';
 import { TextFilter } from './factory';
 
 export enum searchFilterValues {
+  // t('public~Label')
   Label = 'Label',
+  // t('public~Name')
   Name = 'Name',
 }
 
@@ -14,6 +17,8 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
 
   const { onChange, nameFilterInput, labelFilterInput } = props;
 
+  const { t } = useTranslation();
+
   const onToggle = (open: boolean) => setOpen(open);
   const onSelect = (event: React.SyntheticEvent) => {
     setSelected((event.target as HTMLInputElement).name as searchFilterValues);
@@ -21,10 +26,10 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
   };
   const dropdownItems = [
     <DropdownItem key="label-action" name={searchFilterValues.Label} component="button">
-      {searchFilterValues.Label}
+      {t(searchFilterValues.Label)}
     </DropdownItem>,
     <DropdownItem key="name-action" name={searchFilterValues.Name} component="button">
-      {searchFilterValues.Name}
+      {t(searchFilterValues.Name)}
     </DropdownItem>,
   ];
   const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
@@ -45,7 +50,7 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
         toggle={
           <DropdownToggle id="toggle-id" onToggle={onToggle} toggleIndicator={CaretDownIcon}>
             <>
-              <FilterIcon className="span--icon__right-margin" /> {selected}
+              <FilterIcon className="span--icon__right-margin" /> {t(selected)}
             </>
           </DropdownToggle>
         }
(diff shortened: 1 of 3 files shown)


frontend/public/locales/en/public.json
```diff
@@ -419,6 +419,10 @@
   "ReplicationController details": "ReplicationController details",
   "Phase": "Phase",
   "{{statusReplicas}} of {{specReplicas}} pods": "{{statusReplicas}} of {{specReplicas}} pods",
+  "Label": "Label",
+  "app=frontend": "app=frontend",
```
spadgett, line 423: This shouldn't be externalized since it's an actual label selector. It can't be translated.

frontend/public/locales/en/public.json
```diff
@@ -419,6 +419,10 @@
...
+  "my-resource": "my-resource",
```
spadgett, line 424: This is an example resource name that shouldn't be translated.

frontend/public/components/search-filter-dropdown.tsx
```diff
@@ -14,17 +17,19 @@ export const SearchFilterDropdown: React.SFC<SearchFilterDropdownProps> = (props
...
+
   const onToggle = (open: boolean) => setOpen(open);
   const onSelect = (event: React.SyntheticEvent) => {
     setSelected((event.target as HTMLInputElement).name as searchFilterValues);
     setOpen(!isOpen);
   };
   const dropdownItems = [
     <DropdownItem key="label-action" name={searchFilterValues.Label} component="button">
-      {searchFilterValues.Label}
+      {t(searchFilterValues.Label)}
```
spadgett, line 29: Can you confirm that this doesn't break selection of these items? How do we key on these values internally if we're translating the dropdown values?

spadgett:

frontend/public/components/search-filter-dropdown.tsx
rebeccaalpert: It keys off of the name value, which I didn't touch.

```
const onSelect = (event: React.SyntheticEvent) => {
    setSelected((event.target as HTMLInputElement).name as searchFilterValues);
    setOpen(!isOpen);
  };
```

rebeccaalpert: Works fine on my end.

frontend/public/components/search.tsx
```diff
@@ -183,17 +183,26 @@ const SearchPage_: React.FC<SearchProps> = (props) => {
...
-        {labelPlural}{' '}
+        {labelPluralKey ? t(labelPluralKey) : labelPlural}{' '}
         <span className="text-muted show small">
           {apiGroup || 'core'}/{apiVersion}
         </span>
       </span>
     );
   };
 
+  const getChipText = (item: string) => {
```
spadgett, line 197:
```suggestion
  const getChipText = (reference: K8sResourceKindReference) => {
```

frontend/public/components/search.tsx
```diff
@@ -183,17 +183,26 @@ const SearchPage_: React.FC<SearchProps> = (props) => {
...
+    const model = modelFor(item);
```
spadgett, line 198: Don't we already have the model? We shouldn't need to find it again.

frontend/public/components/search.tsx
```diff
@@ -183,17 +183,26 @@ const SearchPage_: React.FC<SearchProps> = (props) => {
...
+    // API discovery happens asynchronously. Avoid runtime errors if the model hasn't loaded.
+    if (!model || !model.labelKey) {
+      return kindForReference(item);
```
spadgett, line 201: I would fall back to `model.label` since we have a null check for the model above. It should always be defined.

frontend/public/components/search.tsx
```diff
@@ -214,7 +224,7 @@ const SearchPage_: React.FC<SearchProps> = (props) => {
                   node: (
                     <>
                       <ResourceIcon kind={resourceKind} />
-                      {kindForReference(resourceKind)}
+                      {getChipText(resourceKind)}
```
spadgett, line 227: It might be easier to just inline since the logic pretty simple.

```suggestion
                      {model.labelKey ? t(model.labelKey) : model.label}
```

spadgett (changes requested):

rebeccaalpert:
Thanks; addressed PR feedback.

frontend/public/components/search.tsx
rebeccaalpert: We don't have the model; it's hidden in a helper function. I can try refactoring tomorrow.

frontend/public/components/search.tsx
rebeccaalpert: We don't store it; it's in a helper function. I can try refactoring tomorrow.

frontend/public/components/search.tsx
rebeccaalpert: I looked into refactoring, but it would a lot of complexity to this component. This component pretty much just deals with a reference rather than the kind. (The name is not correct.) Like getToggleText, you need to convert the reference into the model to do this.

frontend/public/components/search.tsx
rebeccaalpert: I looked into refactoring, but I think it would a lot of complexity to this component. This component pretty much just deals with a reference rather than the kind. Like getToggleText, you need to convert the reference into the model to get the correct value.

spadgett (approved):
/lgtm

PR #7969, opened 2021-01-28
Bug 1921806: i18n APIResourceLinks

rebeccaalpert:
APIResourceLinks in the Explore page may not match navigation and other references to kinds. I updated them to use modal labelKeys where available instead of the kind. The fallback is the kind as originally laid out.

Note that the schema information has not been internationalized, so the descriptions that are closely associated to these in the Explore section will reference the English terms. That content is not hard-coded in the front-end, but we should look into getting that internationalized if we're externalizing Kubernetes kinds.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921806.


spadgett (approved):
/lgtm

PR #7971, opened 2021-01-28
Bug 1921836: Internationalize AccessTableRows

rebeccaalpert:
AccessTableRows didn't internationalize User, Group, or ServiceAccount even though the filter bar above it does. This could lead to a poor user experience since translated terms could differ from non-internationalized terms. I added logic to check for and internationalize the type so they're consistent.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921836.


frontend/public/components/api-explorer.tsx
```diff
@@ -548,6 +534,34 @@ const APIResourceAccessReview: React.FC<APIResourceTabProps> = ({
       transforms: [sortable],
     },
   ];
+
+  const getSubjectType = (type: string) => {
```
spadgett, line 538: I'd call this `getSubjectTypeLabel` to make it clear it's the display value and not the actual value from the resource.

```suggestion
  const getSubjectTypeLabel = (type: string) => {
```

spadgett:

rebeccaalpert:
All set @spadgett.

spadgett (approved):
/lgtm

spadgett:
I opened #7990 to fix the backend tests.

PR #7972, opened 2021-01-28
Bug 1921880: Load all dynamic plugins enabled on the cluster upon Console startup

vojtechszocs:
All dynamic plugins currently enabled on the cluster (`window.SERVER_FLAGS.consolePlugins`) will get loaded upon Console startup via `/api/plugins/<pluginName>` endpoint. Once loaded successfully, such plugins will be automatically enabled.

Non-production builds can still use `window.loadPluginFromURL` global function for development or testing purposes.

### DIFF
--- a/frontend/public/plugins.ts
+++ b/frontend/public/plugins.ts
@@ -33,15 +33,18 @@ const loadPluginFromURL = async (baseURL: string) => {
 };
 
 // Load all dynamic plugins which are currently enabled on the cluster
-Promise.all(
-  window.SERVER_FLAGS.consolePlugins.map((pluginName) =>
-    loadPluginFromURL(`${window.SERVER_FLAGS.basePath}api/plugins/${pluginName}`).then(
-      (pluginID) => {
-        pluginStore.setDynamicPluginEnabled(pluginID, true);
-      },
-    ),
-  ),
-);
+window.SERVER_FLAGS.consolePlugins.forEach((pluginName) => {
+  const url = `${window.SERVER_FLAGS.basePath}api/plugins/${pluginName}`;
+
+  loadPluginFromURL(url)
+    .then((pluginID) => {
+      pluginStore.setDynamicPluginEnabled(pluginID, true);
+    })
+    .catch((error) => {
+      // eslint-disable-next-line no-console
+      console.error(`Error while loading plugin from ${url}`, error);
+    });
+});
 
 if (process.env.NODE_ENV !== 'production') {
   // Expose Console plugin store for debugging
--- a/frontend/public/plugins.ts
+++ b/frontend/public/plugins.ts
@@ -35,9 +35,11 @@ const loadPluginFromURL = async (baseURL: string) => {
 // Load all dynamic plugins which are currently enabled on the cluster
 Promise.all(
   window.SERVER_FLAGS.consolePlugins.map((pluginName) =>
-    loadPluginFromURL(`/api/plugins/${pluginName}`).then((pluginID) => {
-      pluginStore.setDynamicPluginEnabled(pluginID, true);
-    }),
+    loadPluginFromURL(`${window.SERVER_FLAGS.basePath}api/plugins/${pluginName}`).then(
+      (pluginID) => {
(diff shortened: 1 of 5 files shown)


frontend/public/plugins.ts
```diff
@@ -30,13 +25,29 @@ export const initConsolePlugins = _.once((reduxStore: Store<RootState>) => {
...
 
+const loadPluginFromURL = async (baseURL: string) => {
+  const manifest = await fetchPluginManifest(baseURL);
+  loadDynamicPlugin(baseURL, manifest);
+};
+
+// Load all dynamic plugins which are currently enabled on the cluster
+Promise.all(
+  window.SERVER_FLAGS.consolePlugins.map((pluginName) =>
+    loadPluginFromURL(`/api/plugins/${pluginName}`),
```
spadgett, line 36: There can technically be a context root we need to take into account, so we should use `window.SERVER_FLAGS.basePath`

```suggestion
    loadPluginFromURL(`${window.SERVER_FLAGS.basePath}api/plugins/${pluginName}`),
```

spadgett (commented):
Thanks @vojtechszocs 👍

frontend/public/plugins.ts
vojtechszocs: Makes sense, I'll do the code change.

frontend/public/plugins.ts
```diff
@@ -30,13 +26,34 @@ export const initConsolePlugins = _.once((reduxStore: Store<RootState>) => {
...
 });
 
+const loadPluginFromURL = async (baseURL: string) => {
+  const manifest = await fetchPluginManifest(baseURL);
+  loadDynamicPlugin(baseURL, manifest);
+  return getPluginID(manifest);
+};
+
+// Load all dynamic plugins which are currently enabled on the cluster
+Promise.all(
```
spadgett, line 36: Do we need to call `Promise.all` if we aren't using the return value?

vojtechszocs: We don't :smiley: loading a dynamic plugin is an async operation, but such operations are not dependent on one another so we can have a simple `window.SERVER_FLAGS.consolePlugins.forEach()` call instead. I'll make the change.

vojtechszocs: Technically we could use `Promise.allSettled` which resolves after all plugins are either loaded or failed to load, but what we really want is to enable given plugin once it's loaded so a simple `then()` attached to each plugin's `Promise` should be sufficient.

frontend/public/plugins.ts
```diff
@@ -30,13 +26,30 @@ export const initConsolePlugins = _.once((reduxStore: Store<RootState>) => {
...
+  const manifest = await fetchPluginManifest(baseURL);
+  loadDynamicPlugin(baseURL, manifest);
+  return getPluginID(manifest);
+};
+
+// Load all dynamic plugins which are currently enabled on the cluster
+window.SERVER_FLAGS.consolePlugins.forEach((pluginName) => {
+  loadPluginFromURL(`${window.SERVER_FLAGS.basePath}api/plugins/${pluginName}`).then((pluginID) => {
+    pluginStore.setDynamicPluginEnabled(pluginID, true);
+  });
```
spadgett, line 39: I assume we should have a `catch` here with error handling? Maybe just a `console.error`, and we can follow up later and show something in the UI if a plugin doesn't load.

vojtechszocs: Good idea, I'll add a `catch` handler.

Loading a dynamic plugin is essentially a two-step async operation:

```ts
// 1. async, need to await the result
const manifest = await fetchPluginManifest(baseURL);

// 2. also async, but without a result, instead we expect a side effect
// (plugin entry script should call window.loadPluginEntry global function)
loadDynamicPlugin(baseURL, manifest);
```

`fetchPluginManifest` already logs any validation error details to browser console.

`loadDynamicPlugin` and the `window.loadPluginEntry` callback already report any errors to browser console as well.

But having a `catch` handler is still useful since it can report on issues like `coFetch` failing or other async operations which don't log error details.

spadgett:

spadgett (approved):
/lgtm

PR #7973, opened 2021-01-28
Bug 1920655:  only show Cluster autoscalar on Cluster Settings if Clu…

rhamilto:
…sterAutoscalar CRD exists


dtaylor113 (approved):
/lgtm

frontend/public/components/cluster-settings/cluster-settings.tsx
```diff
@@ -875,24 +878,28 @@ export const ClusterVersionDetailsTable: React.FC<ClusterVersionDetailsTableProp
...
-                  <div key={autoscaler.metadata.uid}>
-                    <ResourceLink
-                      kind={clusterAutoscalerReference}
-                      name={autoscaler.metadata.name}
-                    />
-                  </div>
-                ))
-              )}
-            </dd>
+            {clusterAutoscalerFlag && (
```
spadgett, line 881: I'd also adjust the list of resources we pass to `Firehose` so we don't try to fetch the autoscalers when the CRD is not there.

spadgett (commented):
Thanks @rhamilto 👍

spadgett:
@csrwng FYI

frontend/public/components/cluster-settings/cluster-settings.tsx
rhamilto: Updated

rhamilto: But I broke the unit test.  Hold a min.

rhamilto: Unit test fixed.

rhamilto:
Updated.  Please take another look, @spadgett

frontend/public/components/cluster-settings/cluster-settings.tsx
```diff
@@ -961,10 +968,17 @@ export const ClusterOperatorTabPage: React.FC<ClusterOperatorTabPageProps> = ({
 export const ClusterSettingsPage: React.FC<ClusterSettingsPageProps> = ({ match }) => {
   const { t } = useTranslation();
   const title = t('cluster-settings~Cluster Settings');
-  const resources = [
+  const resources: FirehoseResource[] = [
     { kind: clusterVersionReference, name: 'version', isList: false, prop: 'obj' },
-    { kind: clusterAutoscalerReference, isList: true, prop: 'autoscalers', optional: true },
   ];
+  if (useFlag(FLAGS.CLUSTER_AUTOSCALER)) {
```
spadgett, line 974: I'd pull `useFlag` onto its own line outside the `if` since hooks always need to run in the same order every render. It's not a problem now since the `if` will always evaluate, but if someone added another condition later like `isOpenShift && useFlag(FLAGS.CLUSTER_AUTOSCALER)`, it could break things. It's a good practice to have hooks on their own near the top of the function.

spadgett (commented):
Looks good, just one nit. Thanks @rhamilto

frontend/public/components/cluster-settings/cluster-settings.tsx
rhamilto: LOL.  I had it on its own line and then optimized.  That'll learn me.

rhamilto:
Comment addressed.  Thanks, @spadgett.

spadgett (approved):
/lgtm

👍

spadgett:

The test should only run on PRs with ceph changes. A bad config change caused the failure.

PR #7974, opened 2021-01-28
Bug 1921921:  Fix "Global configuration" breadcrumb casing

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921921

After:
Screen Shot 2021-01-28 at 3 54 33 PM

Which aligns it with the casing on the Cluster Settings > Global configuration tab.
Screen Shot 2021-01-28 at 4 00 14 PM


spadgett (approved):
/lgtm

PR #7977, opened 2021-01-29
Bug 1920390: Query Browser: Fix X-axis being updated when no new data fetched

kyoto:
Changing the "Stacked" option or toggling a data series on or off could
cause the X-axis to be updated to the current time even though no new
data was being fetched. This resulted in a gap on the right side of the
graph.

Fix by ensuring the X-axis only updates when necessary.


spadgett (approved):
/lgtm

spadgett:

PR #7978, opened 2021-01-29
Bug 1922075: Fix kubevirt flavor string

gouyang:
Fix the common template namespace in this as well.


spadgett:

PR #6689, opened 2020-09-21
Bug 1880946: Fix crash when Deployment is removed (pod data are not available)

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4295
https://bugzilla.redhat.com/show_bug.cgi?id=1880946

**Analysis / Root cause**: 
When deleting the Deployment in another tab (or with the cli) the PodRing crashes because there are no data available for the pods.

**Solution Description**: 
Check if the data is available and skip PodRing rendering otherwise.

**Screen shots / Gifs for design review**: 
odc-4295 webm

**Unit test coverage report**: 
Not touched

**Test setup:**
Not touched

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #7985, opened 2021-01-29
[release-4.6] Bug 1921603: Fix crash when loaded is true but the data set doesn't include the pods

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/OCPBUGSM-23911
https://bugzilla.redhat.com/show_bug.cgi?id=1921603

**Analysis / Root cause**: 
The console components fetches two sets of data to show a DeploymentConfig (or DeamonSet or StatefulSet) with the correlated Pods. The Pod status donut was shown as soon as the data are "loaded". But in some situations the pod data are not available in this moment. This can happen if the Pod is already deleted (see PR #6689) or if the loading of the pods needs to fetch more then "one page". With the [default page size](https://github.com/openshift/console/blob/release-4.6/frontend/public/actions/k8s.ts#L39) this means more than 250 Pods.

In PR #6689 we fixed this already for Deployments but didn't changed this in the three similar code based for the resource types mention above.

**Solution Description**: 
Similar to PR #6689:
* Show the loading indicator only as long as the first page is loaded and loaded is true.
* Do not render the Pod donut when there are no data are available.

Nothing changed.

Nothing changed.

**Test setup:**
1. Create a relevant list of DeploymentConfigs (6+, as mentioned Deployments are already fixed!)
```
#!/bin/bash
for ((i = 1; i <= 40; i++)); do
    echo Create nodeinfo-$i
    oc new-app jerolimov/nodeinfo --name nodeinfo-$i
    sleep 10
done
```

2. Change the paginationLimit in `frontend/public/actions/k8s.ts` to 5.
3. Open a DeploymentConfig in the developer perspective which needs to fetch at least two pages for fetch the right Pods.

If you are unsure open all detail pages for all ~10 DeploymentConfigs on your local development environment.

To completely reproduce this on a shared cluster you will need more then 250 DeploymentConfigs... :-/

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/daemon-set.tsx
+++ b/frontend/public/components/daemon-set.tsx
@@ -146,7 +146,12 @@ const DaemonSetDetails: React.FC<DaemonSetDetailsProps> = ({ obj: daemonset }) =
         namespace={daemonset.metadata.namespace}
         kind={daemonset.kind}
         render={(d) => {
-          return d.loaded ? (
+          if (!d.loaded) {
+            return <LoadingInline />;
+          } else if (!d.data[daemonset.metadata.uid].pods) {
+            return null;
+          }
+          return (
             <PodRing
               key={daemonset.metadata.uid}
               pods={d.data[daemonset.metadata.uid].pods}
@@ -154,8 +159,6 @@ const DaemonSetDetails: React.FC<DaemonSetDetailsProps> = ({ obj: daemonset }) =
               resourceKind={DaemonSetModel}
               enableScaling={false}
             />
-          ) : (
-            <LoadingInline />
           );
         }}
       />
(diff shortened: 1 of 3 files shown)


rohitkrai03 (approved):
/lgtm

jerolimov:
Backend crash depends on #8025. Hold this until the other PR got merged to reduce unnecessary test loops here.

/hold

jerolimov:
@knobunc this PR and other depends on  #8025, can you `cherry-pick-approve` this as well? :)

knobunc:
[As patch-manager] Pulling the cherry-pick-approval since the PR didn't make the merge window... sorry. Please resolve the CI problems so it can make next week's window.

PR #7997, opened 2021-02-01
Bug 1923003: Update Insights widget state

gkarat:
It turned out that we still show incorrect widget's state for some cases. 

When a customer opts out of the remote health reporting (IO Insights Operator won't send any report), then the OCP WebConsole shouldn't show the "Issues pending" state, but likely some other state.

The patch fixes the issue and aligns states to IO metrics as close as possible. See bug's details at https://issues.redhat.com/browse/CCXDEV-3862. 

Illustrations of reworked states:
1) for the case when an error occurred while receiving metrics (e.g., Insights Operator is manually turned off) Screenshot from 2021-01-29 16-37-45
2) for the case when a cluster is opted-out of health reporting (fixes the bug issue) or Insights Operator has been just initialized and waiting for the first results.
Screenshot from 2021-01-29 16-39-18

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -7,7 +7,8 @@ import {
   legendColorScale,
   riskSorting,
   mapMetrics,
-  isInitialized,
+  isWaitingOrDisabled as _isWaitingOrDisabled,
+  isError as _isError,
 } from './mappers';
 import { PrometheusHealthPopupProps } from '@console/plugin-sdk';
 import { K8sResourceKind } from '@console/internal/module/k8s';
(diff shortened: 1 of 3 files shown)


gkarat:
@spadgett, could you please take a look? Does it have a chance to be delivered before the code freeze for 4.7?

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -7,7 +7,8 @@ import {
   legendColorScale,
   riskSorting,
   mapMetrics,
-  isInitialized,
+  isWaitingOrDisabled as _isWaitingOrDisabled,
+  isError as _isError,
```
spadgett, line 11: Avoid variable names with leading or trailing underscores. We want to get away from these in general in console and add an eslint rule that disables it in the future. I would call the evaluated values `waitingOrDisabled` and `error` below to avoid the name conflicts.

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -27,24 +28,18 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
...
-        </div>
-      )}
-      {isInit && (
-        <div className="co-status-popup__section">
-          Your cluster has been identified, and Insights analyzes your cluster. The results will be
-          displayed shortly.
-        </div>
+      {isError && <div className="co-status-popup__section">Temporary unavailable.</div>}
+      {isWaitingOrDisabled && (
+        <div className="co-status-popup__section">Disabled or waiting for results.</div>
```
spadgett, line 39: This is better, but it would be nice if there was a way to differentiate these two states.

spadgett (approved):
/lgtm

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
gkarat: @spadgett, unfortunately, there is no easy way how to deal with it right now. We are planning to do a spike and rethink the whole state logic for the widget. But for now, we need to get rid of the incorrect state in the scope of this bug.

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
gkarat: @spadgett thanks for pointing it out. I will keep these lines in mind for the next patches. Right now I don't want to cancel your approval by pushing again :sweat_smile:

PR #7998, opened 2021-02-01
Bug 1910492: Fix KMS details getting auto popluated due to incorrect state manipulation.

bipuladh:
Remove direct manipulation of the state object. 
/assign @cloudbehl @afreen23


cloudbehl:
@GowthamShanmugam please have a look once.

GowthamShanmugam:
LGTM, I have tested this in my local also. works fine in different cases

PR #8006, opened 2021-02-01
Bug 1921949: Fix source code URL for apps created using self-hosted repositories

divyanshiGupta:
**Fixes:** https://bugzilla.redhat.com/show_bug.cgi?id=1921949

**Analysis/Root cause:**
After parsing the git url `source` attribute (provides only the git provider name) was being used instead of `resource` (which provides the url domain including the subdomains) and hence the wrong source code url was being generated.

**Solution:**
Use `resource` for creating source code url.

**Gifs:**
Peek 2021-02-01 22-11

/kind bug

### DIFF
--- a/frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
+++ b/frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
@@ -19,7 +19,7 @@ import { getWorkloadResources } from '../transform-utils';
 const namespace = 'test-project';
 
 jest.mock('git-url-parse', () => ({
-  default: jest.fn(() => ({ source: 'github.com', owner: 'openshift', name: 'console' })),
+  default: jest.fn(() => ({ resource: 'github.com', owner: 'openshift', name: 'console' })),
 }));
 
 function getTransformedTopologyData(
--- a/frontend/packages/topology/src/utils/topology-utils.ts
+++ b/frontend/packages/topology/src/utils/topology-utils.ts
@@ -49,7 +49,7 @@ export const getEditURL = (vcsURI?: string, gitBranch?: string, cheURL?: string)
     return null;
   }
   const parsedURL = GitUrlParse(vcsURI);
-  const gitURL = `https://${parsedURL.source}/${parsedURL.owner}/${parsedURL.name}`;
+  const gitURL = `https://${parsedURL.resource}/${parsedURL.owner}/${parsedURL.name}`;
   const fullGitURL = gitBranch ? `${gitURL}/tree/${gitBranch}` : gitURL;
   return cheURL ? `${cheURL}/f?url=${fullGitURL}&policies.create=peruser` : fullGitURL;
 };


jerolimov (changes requested):
2 characters to the rescue. :+1: 

@divyanshiGupta  Can you create a new test case for this in `topology-utils.spec.ts`? Thanks :smirk:

andrewballantyne (approved):

andrewballantyne:
/hold

Added approval so it's not blocked later on me not being available 😄 

Definitely good to have the test so we don't accidentally reintroduce this scenario. Remove hold once test is there.

andrewballantyne:
Tests didn't agree with this change 😄 

Need some `.spec` changes!

andrewballantyne:
Tests didn't agree with this change 😄 

Need some `.spec` changes! Let's make sure they are intentionally desired changes... we don't want to create another bug.

/approve cancel
/hold cancel

divyanshiGupta:
> 2 characters to the rescue. +1
> 
> @divyanshiGupta Can you create a new test case for this in `topology-utils.spec.ts`? Thanks smirk

@jerolimov We already have unit tests for `getEditUrl` in `data-transformer.spec.ts`. And hence those test cases were failing since we are mocking `git-url-parse` with `source` attribute but now we are using `resource` instead. Changed `source` with `resource` and updated the PR.  PTAL.

andrewballantyne:
> > 2 characters to the rescue. +1
> > @divyanshiGupta Can you create a new test case for this in `topology-utils.spec.ts`? Thanks smirk
> 
> @jerolimov We already have unit tests for `getEditUrl` in `data-transformer.spec.ts`. And hence those test cases were failing since we are mocking `git-url-parse` with `source` attribute default value but now we are using `resource` instead. Changed `source` with `resource` and updated the PR. PTAL.
> 
> cc: @andrewballantyne

Ah yeah, the mock is where all the power here is 🤔 

We _could_ write a test for the scenario of the sub-domain... but it would be static test against a 3rd party lib result. So not overly helpful. 

I'm going to approve this as I am unsure how a new test would actually strengthen our code here. At least it would provide an example of a url with a custom (unknown) domain and a sub-domain... but not sure that has a lot of value.

/approve

/assign @jerolimov
Please see if this scenario agrees with you Christoph.

sahil143 (commented):
@divyanshiGupta 
Seems like BitBucket git URL is also broken.
I tested with this repo `https://sbudhwar@bitbucket.org/sbudhwar/nodejs.git`
When clicked on the decorator on the node it open the repo with url `https://bitbucket.org/sbudhwar/nodejs/tree/master` but it should be `https://bitbucket.org/sbudhwar/nodejs/src/master/`
bitbucket


Let me know if we need a separate issue for this.

jerolimov (approved):
I'm fine with that.

/lgtm

jerolimov:
@sahil143 @divyanshiGupta I also think this is another issue which we have because we try to link to the exact branch. Sahil, can you create another issue for that?

/lgtm

sahil143 (approved):
Created Jira ticket https://issues.redhat.com/browse/ODC-5459 for the above-mentioned issue

PR #8007, opened 2021-02-01
Bug 1923721: Handle pipeline svg icon spin using a wrapped <g> element

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5452

**Analysis / Root cause**: 
Chromium does not animate nested SVGs.

**Solution Description**: 
Wrap the SVG in a `<g>` element and animate that instead.

https://user-images.githubusercontent.com/20013884/106496448-e8706a80-64e2-11eb-8976-8ce179f6c488.mp4

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.scss
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.scss
@@ -14,6 +14,9 @@ $border-color: var(--pf-global--BorderColor--light-100);
   &.is-linked {
     cursor: pointer;
   }
+  &--icon-spin {
+    transform-origin: 16.7px 15.1px;
+  }
 }
 
 .odc-pipeline-vis-task-text {
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
@@ -180,18 +180,24 @@ const TaskComponent: React.FC<TaskProps> = ({
         {truncatedVisualName}
       </text>
       {isPipelineRun && showStatusState && (
-        <svg
-          width={30}
-          height={30}
-          viewBox="-5 -5 20 20"
-          style={{
-            color: status
-              ? getRunStatusColor(status.reason, t).pftoken.value
-              : getRunStatusColor(runStatus.Cancelled, t).pftoken.value,
-          }}
+        <g
+          className={cx({
+            'fa-spin odc-pipeline-vis-task--icon-spin': status.reason === runStatus.Running,
+          })}
         >
-          <StatusIcon status={status.reason} shiftOrigin />
-        </svg>
+          <svg
+            width={30}
+            height={30}
+            viewBox="-5 -4 20 20"
+            style={{
+              color: status
+                ? getRunStatusColor(status.reason, t).pftoken.value
+                : getRunStatusColor(runStatus.Cancelled, t).pftoken.value,
+            }}
+          >
+            <StatusIcon status={status.reason} disableSpin />
+          </svg>
+        </g>
       )}
       {showStatusState && (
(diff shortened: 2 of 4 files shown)


karthikjeeyar:
Thanks @rottencandy  for this fix. works fine now
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.scss
```diff
@@ -14,6 +14,9 @@ $border-color: var(--pf-global--BorderColor--light-100);
   &.is-linked {
     cursor: pointer;
   }
+  &--icon-spin {
+    transform-origin: 3.5% 14%;
+  }
```
jerolimov, line 19: Sorry for the late reply. I was yesterday already confused by this "magic numbers" here. It looks like this are the % width and height of the parent container for the icon.

image

You can set this also to the width and height (-0.1, don't know why :confused: ) if you want.

I tested this in the dev tools with this changes and it looks good for me.

PipelineVisualizationTask.scss:

```suggestion
  &--icon-spin {
    transform-origin: 16.61px 16.61px;
  }
```

When I also changed the viewBox in `PipelineVisualizationTask.tsx` to `viewBox="-5 -5 20 21"` to vertical align the Icon a little bit, I need to change this to this values:

```suggestion
  &--icon-spin {
    transform-origin: 16.61px 15.61px;
  }
```

Wdyt?

I tested this because I'm interessted into this magic number. If Andrew gives an approval here I'm also fine with that.

rottencandy: Updated `transform-origin` to `16.61px`. I think values in `px` would be better than percentages. It's confusing, I thought that it would be calculated for the element that gets the animation, and not it's parent.
And `fa-spin` seems to animate simply using `rotate`.

nemesis09:
verified the changes locally
works as expected.
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
```diff
@@ -180,18 +180,24 @@ const TaskComponent: React.FC<TaskProps> = ({
...
+          className={cx({
+            'fa-spin odc-pipeline-vis-task--icon-spin': status.reason === runStatus.Running,
+          })}
         >
-          <StatusIcon status={status.reason} shiftOrigin />
-        </svg>
+          <svg
+            width={30}
+            height={30}
+            viewBox="-5 -5 20 20"
```
andrewballantyne, line 191:
```suggestion
            viewBox="-5 -4 20 20"
```
The icon sits a bit too low, let's bring it more to the center of the node.

rottencandy: Updated. Had to also change `transform-origin` to re center the icon.

andrewballantyne (approved):
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
jerolimov: transform-origin could be more into the direction of `16.7px 15.1px` . But don't want block the PR.

Still a confusing solution we have here. :confused:

rottencandy: Updated.

jerolimov (approved):
Looks fine. Thanks for the extra rounds :smile: 

/lgtm

christianvogt:
@rottencandy please create an issue on PF to add support for embedding react icons inside an existing SVG without PF nesting a new `<svg>` element. For whatever reason, adding the css animation to the nested svg has no affect. It would be good if we can use PF icons and the animation without having to create this wrapper.

PR #8009, opened 2021-02-01
Bug 1924188: Update language selector modal dropdown

rebeccaalpert:
This was a quick tech-check for Megan Hall on a usability bug. The language selector dropdown now pre-selects the language. This previously only worked if the user was using zh, en, or ja exactly. It can now handle unsupported languages and subsets of languages we support, such as en-us and zh-cn.

**en-US (new behavior - example of subset of language)**
Screen Shot 2021-02-01 at 2 31 20 PM

**es (new behavior - example of language we don't support)**
Screen Shot 2021-02-01 at 2 34 08 PM

**zh (existing behavior still works)**
Screen Shot 2021-02-01 at 2 33 44 PM

I had to use i18next.languages since i18next.language and i18n.language.lang will return the detected language regardless of whether we support it. From the docs, i18next.languages is set to an array of language codes that will be used to look up the translation value. When the language is set, this array is populated with the new language codes. Unless overridden, this array is populated with less-specific versions of that code for fallback purposes, followed by the list of fallback languages. I filtered i18next.languages against our supported list of languages to get the correct display language.

CC @rhamilto @spadgett. Are you guys ok with this approach? Let me know if you have thoughts.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1924188.

### DIFF
--- a/frontend/public/components/modals/language-preferences-modal.tsx
+++ b/frontend/public/components/modals/language-preferences-modal.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { TFunction } from 'i18next';
+import i18next, { TFunction } from 'i18next';
 
 import { Dropdown } from '../utils';
 import {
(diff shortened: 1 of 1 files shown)


rhamilto:
Tested and this makes sense to me.  Will wait for @spadgett's input.

frontend/public/components/modals/language-preferences-modal.tsx
```diff
@@ -24,7 +24,10 @@ const LanguagePreferencesModal = (props: LanguagePreferencesModalProps) => {
   }));
   const initLang =
     localStorage.getItem('bridge/language') ||
-    langOptions.find((lang) => lang.lang === i18n.language)?.lang;
+    // handles languages we don't support, as well as subsets of languages we support (such as en-us, zh-cn, etc.)
+    i18next.languages.filter((lang) =>
+      langOptions.find((langOption) => langOption.lang === lang),
+    )[0];
```
spadgett, line 30:
```suggestion
    i18next.languages.find((lang) =>
      langOptions.some((langOption) => langOption.lang === lang),
    );
```

rebeccaalpert: Thanks, all set.

spadgett (approved):
/lgtm

PR #4765, opened 2020-03-18
Add label filtering mechanism and updates row filters.

bipuladh:
Todo: 

- [x] Fix CSS

captured (3)

PR #8010, opened 2021-02-01
Bug 1915971: Role and Role Binding breadcrumbs do not work as expected

dtaylor113:
From BZ: "Today if you click into a Cluster Role from the Roles list view, the breadcrumb from that details page brings you to the generic list view of Cluster Roles. Instead, the breadcrumb should be "Roles" and should redirect you to the Roles list view from which you clicked into the Role.

The same goes for Role Bindings and Cluster Role Bindings. The breadcrumb on any Cluster Role Bindings details page should be "Role Bindings" and should redirect you to the Role Bindings list view from which you clicked into the Role Binding."

### ClusterRole Before
image
<hr/>

image
<hr/>

image

### ClusterRole After
image
<hr/>

image
<hr/>

image

### ClusterRoleBinding Before
image
<hr/>

image

### ClusterRoleBinding After
image


frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -14,18 +21,34 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
...
-        {
-          name: i18next.t('details-page~{{kind}} details', {
-            kind: i18next.t(kindObj.labelKey) || kindObj.label,
-          }),
-          path: `${match.url}`,
-        },
-      ];
+export const breadcrumbsForDetailsPage = (kindObj: K8sKind, match: any) => () => {
+  if (kindObj.apiGroup === 'config.openshift.io' && match.params.name === 'cluster') {
+    return breadcrumbsForGlobalConfig(i18next.t(kindObj.labelKey) || kindObj.label, match.url);
```
spadgett, line 26: I wouldn't have the logic here. Just use `breadcrumbsFor` in the role and role bindings pages.

dtaylor113: Ok, I can use `breadcrumbsFor` for the Role and ClusterRole details pages which use [/frontend/public/components/RBAC/role.jsx](https://github.com/openshift/console/blob/997a61229e6a910210f7861ed23a85d339f2735b/frontend/public/components/RBAC/role.jsx#L244).

However, RoleBindings and ClusterRoleBindings use the [DefaultDetailsPage](https://github.com/openshift/console/blob/f8ffd52597c86737547b2935063f71acd0ca59c7/frontend/public/components/default-resource.jsx#L143).  I can move the 'Bindings' breadcrumb logic to DefaultDetailsPage pass in a `breadcrumbsFor`, or create a new details page for RoleBindings and ClusterRoleBindings, or leave `Binding breadcrumb` logic in `breadcrumbsForDetailsPage()`.

dtaylor113: Hi @spadgett, added `breadcrumbsFor` in `/frontend/public/components/RBAC/role.jsx` for the ClusterRole details pages.
Updated logic in `breadcrumbs.ts` for ClusterRoleBinding breadcrums.

spadgett: I think it would be better to create a custom details page for ClusterRoleBindings, even if only to set the breadcrumbs.

dtaylor113: Added custom details page for ClusterRoleBindings

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +244,29 @@ export const BindingsForRolePage = (props) => {
   );
 };
 
+const getBreadcrumbPath = (plural) => {
+  const lastNamespace = sessionStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
+  return lastNamespace === ALL_NAMESPACES_KEY || lastNamespace === null
```
spadgett, line 249: I realize you can't use `getBreadcrumbPath` from breadcrumbs.ts because the path is different, but we should factor out the common logic for determining the correct namespace to avoid duplicating it. The knowledge of where it is in `sessionStorage` and how the breadcrumb namespace is determined should only be in one place.

frontend/public/components/RBAC/role.jsx
```diff
@@ -249,12 +278,37 @@ export const RolesDetailsPage = (props) => {
         { href: 'bindings', name: t('role~RoleBindings'), component: BindingsForRolePage },
       ]}
       menuActions={menuActions}
+      breadcrumbsFor={
+        props.kind === 'ClusterRole'
+          ? () => getBreadcrumbs('Role', props.kindObj, props.match)
+          : null
```
spadgett, line 284: I'm confused about this check. Shouldn't we go to the same page for both cluster roles and roles?

frontend/public/components/RBAC/role.jsx
```diff
@@ -249,12 +278,37 @@ export const RolesDetailsPage = (props) => {
...
+
+  return (
+    <DetailsPage
+      {...props}
+      menuActions={actions}
+      pages={pages}
+      breadcrumbsFor={
+        props.kind === 'ClusterRoleBinding'
+          ? () => getBreadcrumbs('RoleBinding', props.kindObj, props.match)
+          : null
```
spadgett, line 304: Again, I'm unsure why the logic is different for a cluster role binding.

spadgett (changes requested):

frontend/public/components/RBAC/role.jsx
dtaylor113: Yes, and that same page will be the Roles list, as opposed to the previous ClusterRoles list view.

frontend/public/components/RBAC/role.jsx
dtaylor113: In this case your are right, I don't need this check since `ClusterRoleBinding` is the only resource mapped to this resource page.  Note: there is no custom `RoleBinding` page, it uses `DefaultDetailsPage`.

frontend/public/components/RBAC/role.jsx
spadgett: Actually, can you use `getBreadcrumbPath` from breadcrumbs.ts and pass a `customPlural` here?

dtaylor113: Ha!  "great minds...", yes, but I also had to add create/mock a `match.params.ns` since cluster resources don't have it set:
```
const getBreadcrumbs = (modelRef, kindObj, match) => {
  const model = modelFor(modelRef);
  return [
    {
      name: i18next.t(model.labelPluralKey) || model.labelPlural,
      path: getBreadcrumbPath(
        { params: { ns: sessionStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY) } },
        model.plural,
      ),
```

dtaylor113: Added WIP label, noticed latest changes not quite right.

dtaylor113:
Hi @spadgett, should be ready for review.  -thanks

frontend/public/components/RBAC/role.jsx
spadgett: OK, I see how it works. But would it be simpler to just make this...?

```ts
breadcrumbsFor({() => getBreadcrumbs('Role', props.kindObj, props.match)}
```

If I'm not mistaken, the behavior is the same, but with less logic. You're also sure in that case that both cases always go to the same place. (If the default breadcrumb changes for some reason, that would no longer be true.)

frontend/public/components/utils/breadcrumbs.ts
```diff
@@ -10,7 +10,14 @@ export const getBreadcrumbPath = (match: any, customPlural?: string) => {
     return lastNamespace === ALL_NAMESPACES_KEY
       ? `/k8s/all-namespaces/${customPlural || match.params.plural}`
       : `/k8s/ns/${match.params.ns}/${customPlural || match.params.plural}`;
+  } else if (customPlural === 'roles' || customPlural === 'rolebindings') {
```
spadgett, line 13: I'm unsure why we need this logic now. Can't this all be handled in the details page with `breadcrumbsFor`?

dtaylor113: No, for cluster resources (ClusterRole & ClusterRoleBinding) there is no `match.param.ns`, and without this logic would return `/k8s/cluster/...` instead of a namespaced Role/Rolebinding.

frontend/public/components/RBAC/role.jsx
dtaylor113: agree, logic is the same, updated -thanks

frontend/public/components/utils/breadcrumbs.ts
spadgett: But can't you get the namespace from `sessionStorage` in `breadscrumbFor` in those details pages?

We want to avoid special handling of specific types here.

dtaylor113: HI @spadgett, I kinda had that previously, but your previous [comment](https://github.com/openshift/console/pull/8010#discussion_r570251847) seemed to indicate that `sessionStorage` should be consolidated in `breadcrumbs.jsx`:

"I realize you can't use getBreadcrumbPath from breadcrumbs.ts because the path is different, but we should factor out the common logic for determining the correct namespace to avoid duplicating it [in role.jsx]. The knowledge of where it is in sessionStorage and how the breadcrumb namespace is determined should only be in one place [breadcrumbs.ts]."

Reverted so no changes to `breadcrumbs.ts` and updated `role.jsx` to:
<pre>
  const getBreadcrumbs = (modelRef, kindObj, match) => {
  const model = modelFor(modelRef);
  <b>if (!match.params.ns) {
    match.params.ns = sessionStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
  }</b>
  return [
    {
      name: i18next.t(model.labelPluralKey) || model.labelPlural,
      path: getBreadcrumbPath(match, model.plural),
    },
</pre>

dtaylor113: Code is cleaner this way :-)

spadgett: I think this is better, although we shouldn't set the `match.params.ns` value to something else.

> I kinda had that previously, but your previous comment seemed to indicate that sessionStorage should be consolidated in breadcrumbs.ts

Ideally, we do both. We keep the breadcrumb logic for role bindings in its own page, but then have a helper like `getBreadcrumbNamespace` in `breadcrumbs.ts` that keeps the session storage logic in one place there.

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +244,25 @@ export const BindingsForRolePage = (props) => {
   );
 };
 
+const getBreadcrumbs = (modelRef, kindObj, match) => {
+  const model = modelFor(modelRef);
+  if (!match.params.ns) {
+    match.params.ns = sessionStorage.getItem(LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY);
```
spadgett, line 250: We shouldn't be overriding any values in `match`. I see why you're doing it, but that value is set by react-router should be treated as immutable. I'm not sure what consequences changes it could have. I would skip calling `getBreadcrumbPath` and just build the path yourself.

frontend/public/components/utils/breadcrumbs.ts
dtaylor113: > I think this is better, although we shouldn't set the `match.params.ns` value to something else.

For ClusterRole and ClusterRoleBinding there is no `match.params.ns`.

spadgett: Right, but we can't change the value of the `match` object that is passed to us from react-router. Treat it as read-only.

dtaylor113: Can I clone and then modify cloned match obj?

frontend/public/components/RBAC/role.jsx
dtaylor113: @spadgett fixed, used `resourceListPathFromModel` with `lastBreadcrumb` to determine correct breadcrumb path.

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +245,26 @@ export const BindingsForRolePage = (props) => {
...
 
+const getBreadcrumbs = (modelRef, kindObj, match) => {
+  const lastNamespace = getLastNamespace();
+  const model = modelFor(modelRef);
+  return [
+    {
+      name: i18next.t(model.labelPluralKey) || model.labelPlural,
+      path: resourceListPathFromModel(
+        model,
+        !lastNamespace || lastNamespace === ALL_NAMESPACES_KEY ? null : lastNamespace,
```
spadgett, line 256: Why is the `!lastNamespace` check needed? Won't this pass a `boolean` sometimes to `resourceListPathFromModel` which expects a string when `lastNamespace` is not set?

frontend/public/components/RBAC/role.jsx
```diff
@@ -249,12 +276,27 @@ export const RolesDetailsPage = (props) => {
         { href: 'bindings', name: t('role~RoleBindings'), component: BindingsForRolePage },
       ]}
       menuActions={menuActions}
+      breadcrumbsFor={() => getBreadcrumbs('Role', props.kindObj, props.match)}
```
spadgett, line 279: Can't we use the default here? I don't think we need to override `breadcrumbsFor`.

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +245,26 @@ export const BindingsForRolePage = (props) => {
   );
 };
 
+const getBreadcrumbs = (modelRef, kindObj, match) => {
```
spadgett, line 248: Why not just take the `model` instead of `modelRef`? We should know it in all cases this is called, correct?

As a rule, we should avoid `modelFor` when possible since it only works with static models.

frontend/public/components/RBAC/role.jsx
dtaylor113: lastNamespace is null before user specifically selects a project.  Basically test is saying "if lastNamespace is null OR not null and equals 'all namespaces', then pass in a `null` namespace string  which `resourceListPathFromModel` will then default to 'all namespaces' when no namespace is passed in.  
If lastNamespace is not null but doesn't equal 'all namespaces', then pass it in to be used by `resourceListPathFromModel`

frontend/public/components/RBAC/role.jsx
dtaylor113: umm...we seem to be going in [circles](https://github.com/openshift/console/pull/8010#discussion_r570569030).

frontend/public/components/RBAC/role.jsx
dtaylor113: Is there another method I should call instead of `modelFor`?

frontend/public/components/RBAC/role.jsx
spadgett: I guess I'm confused. Is this component only used for Roles, or is it used for both Roles and ClusterRoles? I don't see where it's registered as the details page for ClusterRoles.

frontend/public/components/RBAC/role.jsx
spadgett: Don't you already know the model in all cases where this is called? Can you just pass `ClusterRoleBindingModel` instead of passing the name and having to find it with `modelFor`?

frontend/public/components/RBAC/role.jsx
dtaylor113: It is used for Both [here](https://github.com/openshift/console/blob/cb0e42882c6ea3bb187f70bd96fd1c600c2c520c/frontend/public/components/RBAC/role.jsx#L284) and [here](https://github.com/openshift/console/blob/cb0e42882c6ea3bb187f70bd96fd1c600c2c520c/frontend/public/components/resource-pages.ts#L228).

frontend/public/components/RBAC/role.jsx
dtaylor113: No, for `ClusterRoleBindingModel` we really want to use the breadcrumb for `RoleBinding` Model.   Same for `ClusterRoleModel` we really want to use breadcrumb for `Role` model.

frontend/public/components/RBAC/role.jsx
spadgett: Thanks, that's what I missed. I was looking for it in resource-pages.ts.

frontend/public/components/RBAC/role.jsx
spadgett: I didn't realize the page was reused for both resources. Disregard my original comment.

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +244,25 @@ export const BindingsForRolePage = (props) => {
   );
 };
 
+const getBreadcrumbs = (model, kindObj, match) => {
+  const lastNamespace = getLastNamespace();
+  return [
+    {
+      name: i18next.t(model.labelPluralKey) || model.labelPlural,
```
spadgett, line 251: Sorry, I missed this, but I wouldn't rely on the any particular behavior when `i18next.t(undefined)` is called. I have a feeling that didn't cause a problem because it's defined for both of these models, but I'd guess we'd get at least a missing key error.

```suggestion
      name: model.labelPluralKey ? i18next.t(model.labelPluralKey) : model.labelPlural,
```

frontend/public/components/RBAC/role.jsx
```diff
@@ -238,6 +244,25 @@ export const BindingsForRolePage = (props) => {
...
+      path: resourceListPathFromModel(
+        model,
+        !lastNamespace || lastNamespace === ALL_NAMESPACES_KEY ? null : lastNamespace,
+      ),
+    },
+    {
+      name: i18next.t('details-page~{{kind}} details', {
+        kind: i18next.t(kindObj.labelKey) || kindObj.label,
```
spadgett, line 259:
```suggestion
        kind: kindObj.labelKey ? i18next.t(kindObj.labelKey) : kindObj.label,
```

You might want to assign this to a variable above to avoid having to check this twice.

dtaylor113: > You might want to assign this to a variable above to avoid having to check this twice.

@spadgett it's not the same variable.  One is getting label from model, the other from kindObj.   Ie. One is `model.labelPluralKey` and the other is `kindObj.labelKey`

dtaylor113:
Hi @spadgett, added Cypress tests for these special breadcrumbs

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +54,113 @@ describe('Namespace', () => {
...
+        .breadcrumb(0)
+        .contains(rolesOrBindings)
+        .click();
+      listPage.rows.shouldBeLoaded();
+      projectDropdown.shouldContain('All Projects');
+      listPage.titleShouldHaveText(rolesOrBindings);
+
+      cy.log(`test Cluster${rolesOrBindings} & openshift-console project`);
+      projectDropdown.selectProject('openshift-console');
+      projectDropdown.shouldContain('openshift-console');
```
spadgett, line 127: `default` would be better so the tests run on native k8s where this project won't exist.

```suggestion
      projectDropdown.selectProject('default');
      projectDropdown.shouldContain('default');
```

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +54,113 @@ describe('Namespace', () => {
...
       'Nav from list page to details page should change Project from "All Projects" to resource specific project',
     );
-    cy.get(`[data-test-rows="resource-row"]`)
-      .first()
-      .find('a')
-      .first()
-      .click();
+    listPage.rows.clickFirstLinkInFirstRow();
+    detailsPage.isLoaded();
     projectDropdown.shouldNotContain('All Projects'); // after drilldown to Details page, project should be specific to resource
```
spadgett, line 65: This isn't new, but this should be "All Namespaces" if the tests are running against native k8s.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +54,113 @@ describe('Namespace', () => {
...
+    listPage.rows.shouldBeLoaded();
     projectDropdown.shouldContain('default');
   });
+
+  ['Roles', 'RoleBindings'].forEach((rolesOrBindings) => {
+    it(`tests ${rolesOrBindings} and Cluster${rolesOrBindings} detail page breadcrumbs`, () => {
+      nav.sidenav.clickNavLink(['User Management', rolesOrBindings]);
+
+      cy.log(`test ${rolesOrBindings} & All Projects`);
+      projectDropdown.selectProject('All Projects');
```
spadgett, line 111: "All Namespaces" if running against native k8s.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
spadgett: Not new, but in each of these cases, it's better to check that the project dropdown has the expected project name. (Here and in the new tests below.)

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
spadgett: We refer to the project name enough times it would be good to assign it to a const.

frontend/packages/integration-tests-cypress/views/details-page.ts
```diff
@@ -13,6 +13,7 @@ export const detailsPage = {
       .click();
   },
   isLoaded: () => cy.byTestID('skeleton-detail-view').should('not.exist'),
+  breadcrumb: (number: number) => cy.byLegacyTestID(`breadcrumb-link-${number}`),
```
spadgett, line 16: I'd probably call this `breadcrumbIndex` or similar so it's clear that's it's taking a number.

frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -26,6 +26,20 @@ export const listPage = {
...
       });
     },
+    by: (rowFilter: string) => {
+      cy.get('.pf-c-toolbar__content-section').within(() => {
+        cy.byLegacyTestID('filter-dropdown-toggle').click();
+        /* PF Filter dropdown menu items are:
+           <li id="cluster">
+             <a data-test-row-filter="cluster">
+           Tried cy.get(`[data-test-row-filter="${rowFilter}"]`).click() which found the <a /> but said not clickable due to
+           it's css having 'pointer-events: none'.  Tried ...click({force: true}) which did the click but page not reloaded with
```
spadgett, line 36: Is this a PF accessibility bug?

frontend/packages/integration-tests-cypress/views/list-page.ts
```diff
@@ -34,6 +48,13 @@ export const listPage = {
     countShouldBe: (count: number) => {
       cy.get(`[data-test-rows="resource-row"`).should('have.length', count);
     },
+    clickFirstLinkInFirstRow: () => {
+      cy.get(`[data-test-rows="resource-row"]`)
+        .first()
+        .find('a')
+        .first()
+        .click({ force: true }); // after applying row filter, resource rows detached from DOM according to cypress, need to force the click
```
spadgett, line 56: I'd like to understand why this is better. Are we sure we're clicking on the right link after the table has been filtered?

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: Due to limitations we discussed offline, I have filed a BZ to track this issue: [Cypress: Implement method to determine whether running OpenShift or K8s native](https://bugzilla.redhat.com/show_bug.cgi?id=1930964)

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +56,113 @@ describe('Namespace', () => {
...
+      projectDropdown.shouldNotExist();
+      detailsPage
+        .breadcrumb(0)
+        .contains(rolesOrBindings)
+        .click();
+      listPage.rows.shouldBeLoaded();
+      projectDropdown.shouldContain(AllProjects);
+      listPage.titleShouldHaveText(rolesOrBindings);
+
+      cy.log(`test Cluster${rolesOrBindings} & openshift-console project`);
```
spadgett, line 127: Looks like the log messages are out of date with the code now that we changed the namespace.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +56,113 @@ describe('Namespace', () => {
...
+      projectDropdown.shouldNotExist();
+      detailsPage
+        .breadcrumb(0)
+        .contains(rolesOrBindings)
+        .click();
+      listPage.rows.shouldBeLoaded();
+      projectDropdown.shouldContain(defaultProjectName);
+      listPage.titleShouldHaveText(rolesOrBindings);
+
+      cy.log(`test ${rolesOrBindings} & openshift-console project`);
```
spadgett, line 142: Here, too.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: Hi, yes, caught those as well :-) Should be pushing new code soon.

frontend/packages/integration-tests-cypress/views/list-page.ts
dtaylor113: I don't believe so, the `'pointer-events: none'` is being set in [/frontend/public/components/_filter-toolbar.scss](https://github.com/openshift/console/blob/c5299d1e80f1c802b10dda8135c2169928046e64/frontend/public/components/_filter-toolbar.scss#L3)

frontend/packages/integration-tests-cypress/views/list-page.ts
dtaylor113: Definitely clicking on the correct link.  I removed the `{force: true}` and get: 
image

Here is the [Learn More...](https://docs.cypress.io/guides/references/error-messages.html#cy-failed-because-the-element-you-are-chaining-off-of-has-become-detached-or-removed-from-the-dom) link from the above screenshot.

I think our Filter asyncronously filters and re-renders the table without reloading screen.  Cypress has documentation on it, but I haven't yet found a solution/workaround.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: Hi @spadgett, switched to clicking on `alertmanager-main` in order to specifically test for `openshift-monitoring` project/namespace.  Not sure if that helps with the k8s native support though : /

frontend/packages/integration-tests-cypress/views/details-page.ts
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: .. done.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed

frontend/public/components/RBAC/role.jsx
dtaylor113: fixed

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -20,6 +21,8 @@ describe('Namespace', () => {
   });
 
   const newName = `${testName}-ns`;
+  const defaultProjectName = 'default';
+  const AllProjects = 'All Projects';
```
spadgett, line 25: nit: This should be camelCase with a lowercase first letter (even if the value itself has an uppercase `A`). PascalCase variable names are typically only used for components and classes. I'd probably call this `allProjectsDropdownLabel` to make it more clear what it's a constant for.

frontend/packages/integration-tests-cypress/views/list-page.ts
spadgett: > Definitely clicking on the correct link.

I mean, if it's removed from the DOM, is it really the correct link? :)

Is the page updating from underneath the tests while `.first().find('a').first()` is being evaluated? Is there a way to avoid the use of `first` here?

frontend/packages/integration-tests-cypress/views/list-page.ts
spadgett: > I don't believe so, the `'pointer-events: none'` is being set in [/frontend/public/components/_filter-toolbar.scss](https://github.com/openshift/console/blob/c5299d1e80f1c802b10dda8135c2169928046e64/frontend/public/components/_filter-toolbar.scss#L3)

In that case, is this a console accessibility bug? @bipuladh Do you remember why you added `pointer-events: none` in #4765?

bipuladh: Not sure, but I think it was causing issues while working with `Dropdown` https://github.com/openshift/console/pull/4765/files#diff-db47992a42e1e70bc69c0b457340626915c70d4b2fe7ad976ad7aa080131e424R224

dtaylor113:
> Definitely clicking on the correct link.

I mean, if it's removed from the DOM, is it really the correct link? :)

Is the page updating from underneath the tests while `.first().find('a').first()` is being evaluated? Is there a way to avoid the use of `first` here?

dtaylor113:
Hi @spadgett,
> if it's removed from the DOM, is it really the correct link? :)
Is the page updating from underneath the tests while .first().find('a').first() is being evaluated? Is there a way to avoid the use of first here?

I think the detachment happens before even before the first `cy.get(..` but after filter applied/url change/page reloaded -which I don't understand.  But even this doesn't work:
```
cy.get(`[data-test-rows="resource-row"] > td a`)
  .eq(0)
  .click();
```

Might be part of a larger cypress issue.  Looking into Cypress DOM detachment issues.

dtaylor113:
Hi @spadgett,

I accidentally resolved your last comment and didn't see an option to unresolve it, so reposting:

> I don't believe so, the `'pointer-events: none'` is being set in [/frontend/public/components/_filter-toolbar.scss](https://github.com/openshift/console/blob/c5299d1e80f1c802b10dda8135c2169928046e64/frontend/public/components/_filter-toolbar.scss#L3)

In that case, is this a console accessibility bug? @bipuladh Do you remember why you added `pointer-events: none` in #4765?

> @bipuladh 2 hours ago 
Not sure, but I think it was causing issues while working with `Dropdown` https://github.com/openshift/console/pull/4765/files#diff-db47992a42e1e70bc69c0b457340626915c70d4b2fe7ad976ad7aa080131e424R224

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed to `allProjectsDropdownlabel`

dtaylor113:
Struggling to find Role and/or ClusterRoles available in our CI cluster.

spadgett:
> Struggling to find Roles and/or ClusterRoles available in our CI cluster. Wondering if I should just create them, but that would be 4 objects though to create and delete just to test a breadcrumb.

All clusters should have ClusterRoles for sure. You might need to create a Role if using `default`. If we're creating resources, we should use the test namespace, though.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +77,127 @@ describe('Namespace', () => {
...
+        .breadcrumb(0)
+        .contains(rolesOrBindings)
+        .click();
+      listPage.rows.shouldBeLoaded();
+      projectDropdown.shouldContain(defaultProjectName);
+
+      cy.log(`test ${rolesOrBindings} & All Projects`);
+      projectDropdown.selectProject(allProjectsDropdownLabel);
+      if (rolesOrBindings === 'Roles') {
+        createExampleRole();
```
spadgett, line 184: If we're creating resources, you should create a test project instead of using `default`. (I assumed there would be a `Role` already defined when I suggested that.)

dtaylor113: sorry, my bad, you had mentioned that in a previous comment.  fixed.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +83,122 @@ describe('Namespace', () => {
...
+  ['Roles', 'RoleBindings'].forEach((rolesOrBindings) => {
+    it.only(`tests ${rolesOrBindings} and Cluster${rolesOrBindings} detail page breadcrumbs`, () => {
+      nav.sidenav.clickNavLink(['User Management', rolesOrBindings]);
+
+      cy.log(`test Cluster${rolesOrBindings} & All Projects`);
+      projectDropdown.selectProject(allProjectsDropdownLabel);
+      listPage.rows.shouldBeLoaded();
+      listPage.filter.by('cluster');
+      listPage.filter.numberOfActiveFiltersShouldBe(1);
+      listPage.filter.byName('alertmanager-main');
```
spadgett, line 146: This won't work on a non-OpenShift cluster since this ClusterRole won't exist. We should create a test project / test roles as needed.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
```diff
@@ -53,54 +83,122 @@ describe('Namespace', () => {
...
+        .breadcrumb(0)
+        .contains(rolesOrBindings)
+        .click();
+      listPage.rows.shouldBeLoaded();
+      projectDropdown.shouldContain(defaultProjectName);
+      listPage.titleShouldHaveText(rolesOrBindings);
+
+      cy.log(`test ${rolesOrBindings} & default project`);
+      listPage.filter.by('namespace');
+      listPage.filter.byName('prometheus-k8s');
```
spadgett, line 176: Same here. I would not assume this role exists on the cluster (or the hard-coded names below).

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed, created example Role, RoleBinding, ClusterRole, and ClusterRoleBinding instances using test namespace/project.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed, created example Role, RoleBinding, ClusterRole, and ClusterRoleBinding instances using test namespace/project.

frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts
dtaylor113: fixed, created example Role, RoleBinding, ClusterRole, and ClusterRoleBinding instances using test namespace/project.

frontend/public/components/utils/status-box.tsx
```diff
@@ -41,7 +41,10 @@ export const LoadError: React.FC<LoadErrorProps> = ({
 LoadError.displayName = 'LoadError';
 
 export const Loading: React.FC<LoadingProps> = ({ className }) => (
-  <div className={classNames('co-m-loader co-an-fade-in-out', className)} data-test="co-m-loader">
+  <div
+    className={classNames('co-m-loader co-an-fade-in-out', className)}
+    data-test="loading-indicator"
```
spadgett, line 46: 👍

spadgett (commented):
Looks good, just a few small nits. Thanks @dtaylor113

frontend/packages/integration-tests-cypress/tests/crud/roles-rolebindings.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+import { detailsPage } from '../../views/details-page';
+import * as yamlEditor from '../../views/yaml-editor';
+
+const roleName = 'example-role';
+const clusterRoleName = 'example-cluster-role';
+const roleBindingName = 'example-rolebinding';
+const clusterRoleBindingName = 'example-cluster-rolebinding';
+
+const createExampleRoles = () => {
+  cy.log(`create Role instance`);
```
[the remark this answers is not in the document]
dtaylor113: Hi @spadgett, this seemes to be the correct rule:
```
"rules": {
     ...
    "quotes": ["error", "single", { "allowTemplateLiterals": false }]
```
But it shows 25+ files would need to be updated for this new rule (double quotes in `{t("..` functions, etc...  I'd prefer to do this in a seperate PR.

For now, I replaced backticks with single quotes in this file.

frontend/packages/integration-tests-cypress/tests/crud/roles-rolebindings.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+  projectDropdown.selectProject(testName);
+  projectDropdown.shouldContain(testName);
+  listPage.clickCreateYAMLbutton();
+  // sidebar needs to be fully loaded, else it sometimes overlays the Create button
+  cy.byTestID('resource-sidebar').should('exist');
+  yamlEditor.isLoaded();
+  let newContent;
+  yamlEditor.getEditorContent().then((content) => {
+    newContent = _.defaultsDeep({}, { metadata: { name: roleName } }, safeLoad(content));
+    yamlEditor.setEditorContent(safeDump(newContent, { sortKeys: true })).then(() => {
```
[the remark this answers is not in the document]
dtaylor113: fixed

spadgett (approved):
/lgtm

dtaylor113:
/retest
Hit OLM failure last 2 tries

dtaylor113:
/test all
passed: 1

dtaylor113:
/test all
passed: 2

dtaylor113:
/test all
passed: 2
failed : 1 ("certificate signed by unknown authority")

PR #8012, opened 2021-02-02
Bug 1922911: Query Browser: Show empty state when all series are disabled

kyoto:
Rather than showing a graph with no data, it is better UX to show the
`GraphEmpty` empty state.

This also fixes a bug where disabling all series, then enabling one of
them again caused an error.

This change impacts the Metrics page for both admin and dev, since they
are the only pages that allow disabling all series.

screenshot


spadgett (approved):
/lgtm

PR #8015, opened 2021-02-02
Bug 1923939: add missing trigger types and update the optional fields - 4.6

karthikjeeyar:
This is a manual cherry pick of https://github.com/openshift/console/pull/7889 

**Fixes**: https://issues.redhat.com/browse/ODC-5350 in 4.6

**Analysis / Root cause**:
Event listeners has an optional triggerRef field which is not compatible in UI code causes the UI to crash

**Solution Description**:
Add the missing trigger types such as triggerRef and update the checks for optional fields.

**Test setup**:
```
kind: EventListener
metadata:
  name: vote-app
spec:
  serviceAccountName: pipeline
  triggers:
    - triggerRef: vote-app
```

This fixes the UI Crashing when above yaml is used.
cc: @andrewballantyne


spadgett:
You will need #8025 for the backend tests to pass.

karthikjeeyar:
Thanks @spadgett for the information.

/hold

invincibleJai:
/lgtm
/approve

Verified on PR build and am not seeing issue with Event listener with triggerRef creation

PR #8018, opened 2021-02-02
Bug 1923894: Listing Subscriptions in OLM should also show them when 'All Projects' namespace is picked

jhadvig:
The issue is the when the `All Projects` namespace is picked, the `props.namespace` is set to `undefined`. When flattening the Subscriptions we should match that.

/assign @TheRealJon


TheRealJon (approved):
/lgtm

PR #8022, opened 2021-02-02
Bug 1924046: Avoid checking user settings if url has namespace

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5451

**Analysis / Root cause**: 
Loading userSettings from localStorage is synchronous and loads instantaneously, so dispatching of namespace from url fails since we check if userSettings are loaded.

**Solution Description**: 
Avoid checking for the status of the userSettings fetch, if url contains a namespace.

Unchanged.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


rohitkrai03 (approved):
/approve

dtaylor113 (approved):
/lgtm

PR #8023, opened 2021-02-02
Moved api types files to types/api folder, added index file

metalice:


metalice:
@suomiy @rawagner @yaacov @glekner @pcbailey have a look :)

suomiy:
looks ok, let's wait for the CI

glekner:
convention is naming the folder `@types`. please consider it

metalice:
> convention is naming the folder `@types`. please consider it

in a project? or package? I think it might be confusing as developers are using @ prefix for packages.. so to do something like ./@types/api can be confusing? 

I think that if we could auto-generate the rest of the API and then pack it like a real npm package it will be much better, but this is the next step :)

yaacov:
> I think that if we could auto-generate the rest of the API and then pack it like a real npm package it will be much better, but this is the next step :)

+1 for auto generation

> added index file

I have some reservations about added index file:
a - we can't enforce (*) users that add a new class to add it to the index file.
b - we can't enforce users that consume a class to import it using the index file and not directly.
c - it violate our agreement to export classes as defaults
d - we can't force users that do add a new class to the index file to add it in the correct place alphabetically,

My vote is not use index files, do we all ( except me :-) )  think it's a good idea ?

(*) we can but it will require more overhead

metalice:
> > I think that if we could auto-generate the rest of the API and then pack it like a real npm package it will be much better, but this is the next step :)
> 
> +1 for auto generation
> 
> > added index file
> 
> I have some reservations about added index file:
> a - we can't enforce (*) users that add a new class to add it to the index file.
> b - we can't enforce users that consume a class to import it using the index file and not directly.
> c - it violate our agreement to export classes as defaults
> d - we can't force users that do add a new class to the index file to add it in the correct place alphabetically,
> 
> My vote is not use index files, do we all ( except me :-) ) think it's a good idea ?
> 
> (*) we can but it will require more overhead

u have some solid points, but.. :) 
c- it's talking about react components import.

all the rest will be figured out once we split it for npm package, so currently I don't see that index.js as a bad practice, it will be useful for nice paths and quick understanding. its a step in the auto generate types. :)

PR #8026, opened 2021-02-02
Add .dockerignore file so that all node_modules are ignored

TheRealJon:
Greatly improves performance of building an image locally.


spadgett (approved):
/lgtm

PR #8027, opened 2021-02-02
Bug 1922992: i18n missing items in YAML sidebar

rebeccaalpert:
Resource kind labels and button text were not internationalized. I updated this. The rest of the schema details isn't available for internationalization at this time unless we hard-code it.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1922992.


frontend/public/components/sidebars/explore-type-sidebar.tsx
```diff
@@ -49,8 +52,9 @@ export const ExploreType: React.FC<ExploreTypeProps> = (props) => {
     ? currentSelection.description
     : currentDefinition.description;
   const required = new Set(currentDefinition.required || []);
+  const kindLabel = kindObj.labelKey ? t(kindObj.labelKey) : kindObj.kind;
   const breadcrumbs = drilldownHistory.length
-    ? [kindObj ? kindObj.kind : 'Schema', ..._.map(drilldownHistory, 'name')]
+    ? [kindObj ? kindLabel : t('public~Schema'), ..._.map(drilldownHistory, 'name')]
```
spadgett, line 57: We check if `kindObj` is set above (line 31), so I don't think this fallback is needed.

frontend/public/components/sidebars/resource-sidebar-samples.tsx
```diff
@@ -97,7 +101,7 @@ const ResourceSidebarSnippet: React.FC<ResourceSidebarSnippetProps> = ({
       <p>{description}</p>
       <Button type="button" variant="link" isInline onClick={insertSnippet}>
         <PasteIcon className="co-icon-space-r" />
-        Insert Snippet
+        {t('public~Insert Snippet')}
```
spadgett, line 104:
```suggestion
        {t('public~Insert snippet')}
```

frontend/public/components/sidebars/resource-sidebar.tsx
```diff
@@ -74,7 +74,8 @@ export const ResourceSidebar = (props) => {
     return null;
   }
 
-  const { label = sidebarLabel } = kindObj ?? {};
+  const kindLabel = kindObj?.labelKey ? t(kindObj.labelKey) : kindObj?.label;
+  const label = sidebarLabel ? sidebarLabel : kindLabel ?? {};
```
spadgett, line 78: Why are we falling back to `{}` when the value should be a string?

spadgett (changes requested):

frontend/public/components/sidebars/explore-type-sidebar.tsx
rebeccaalpert: You can have no kindObj as long as you have a schema, so you do need the fallback.

spadgett (approved):
/lgtm

PR #8028, opened 2021-02-02
Bug 1924143: Create application edit url based on git provider

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5459

**Analysis / Root cause**: 
There is no standard way for git providers to display a repository's branch.

**Solution Description**: 
- Craft the git url based on the git provider, and resort to not using branch name if the git provider is unsupported.
  ```
  GitHub: {repo}/tree/{branch}
  GitLab: {repo}/-/tree/{branch}
  Bitbucket: {repo}/src/{branch}
  ```
- Avoid adding `master` in import flow if branch is not provided, as this fails for branches whose default branch is not master.

Added tests.
```yarn run v1.22.10
$ jest /data-transforms/__tests__/data-transformer
 PASS  packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts (8.066s)
Test Suites: 1 passed, 1 total
Tests:       1 skipped, 25 passed, 26 total
Snapshots:   0 total
Time:        9.822s
Ran all test suites matching /\/data-transforms\/__tests__\/data-transformer/i.
Done in 11.39s.
```

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/utils/resource-label-utils.ts
+++ b/frontend/packages/dev-console/src/utils/resource-label-utils.ts
@@ -44,10 +44,9 @@ export const getAppLabels = ({
 };
 
 export const getGitAnnotations = (gitURL: string, gitRef?: string) => {
-  const ref = gitRef || 'master';
   return {
     'app.openshift.io/vcs-uri': gitURL,
-    'app.openshift.io/vcs-ref': ref,
+    'app.openshift.io/vcs-ref': gitRef || '',
   };
 };
 
--- a/frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
+++ b/frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
@@ -18,10 +18,6 @@ import { getWorkloadResources } from '../transform-utils';
 
 const namespace = 'test-project';
 
-jest.mock('git-url-parse', () => ({
-  default: jest.fn(() => ({ resource: 'github.com', owner: 'openshift', name: 'console' })),
-}));
-
 function getTransformedTopologyData(
   mockData: TopologyDataResources,
   transformByProp: string[] = WORKLOAD_TYPES,
@@ -160,6 +156,39 @@ describe('data transformer ', () => {
     const mockGitURI = 'git@github.com:openshift/console';
     const editUrl = getEditURL(mockGitURI, '', '');
     expect(editUrl).toBe('https://github.com/openshift/console');
+  });
+
+  it('should return full git url for GitHub URI', () => {
+    const mockGitURI = 'https://github.com/openshift/console';
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://github.com/openshift/console/tree/branch1');
+  });
+
+  it('should return full git url for GitLab URI', () => {
+    const mockGitURI = 'https://gitlab.com/example/reponame';
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://gitlab.com/example/reponame/-/tree/branch1');
+  });
+
+  it('should return full git url for Bitbucket URI', () => {
+    const mockGitURI = 'https://bitbucket.org/username/examplerepo';
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://bitbucket.org/username/examplerepo/src/branch1');
+  });
+
+  it('should return only git url for repo from non-supported git provider', () => {
+    const mockGitURI = 'https://example.com/username/examplerepo';
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://example.com/username/examplerepo');
+  });
+
+  it('should return full git url for supported providers with custom domains', () => {
+    let editUrl = getEditURL('https://github.example.com/uname/repo', 'branch1');
+    expect(editUrl).toBe('https://github.example.com/uname/repo/tree/branch1');
+    editUrl = getEditURL('https://gitlab.example.com/uname/repo', 'branch1');
+    expect(editUrl).toBe('https://gitlab.example.com/uname/repo/-/tree/branch1');
+    editUrl = getEditURL('https://bitbucket.example.com/uname/repo', 'branch1');
+    expect(editUrl).toBe('https://bitbucket.example.com/uname/repo/src/branch1');
   });
 
   it('should return builder image icon for nodejs', () => {
--- a/frontend/packages/topology/src/utils/topology-utils.ts
+++ b/frontend/packages/topology/src/utils/topology-utils.ts
@@ -1,5 +1,5 @@
 import * as _ from 'lodash';
-import GitUrlParse from 'git-url-parse';
+import * as GitUrlParse from 'git-url-parse';
 import i18next from 'i18next';
 import {
   K8sResourceKind,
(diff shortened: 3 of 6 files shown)


frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -44,10 +44,9 @@ export const getAppLabels = ({
 };
 
 export const getGitAnnotations = (gitURL: string, gitRef?: string) => {
-  const ref = gitRef || 'master';
   return {
     'app.openshift.io/vcs-uri': gitURL,
-    'app.openshift.io/vcs-ref': ref,
+    'app.openshift.io/vcs-ref': gitRef || '',
```
jerolimov, line 49: Why do we not default here to master (or main?) anymore in the annotations? This is not required as part of this change, or?

frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
```diff
@@ -162,6 +158,30 @@ describe('data transformer ', () => {
...
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://bitbucket.org/username/examplerepo/src/branch1');
+  });
+
+  it('should return only git url for repo from non-supported git provider', () => {
+    const mockGitURI = 'https://example.com/username/examplerepo';
+    const editUrl = getEditURL(mockGitURI, 'branch1');
+    expect(editUrl).toBe('https://example.com/username/examplerepo');
+  });
+
```
jerolimov, line 184: Can you please add also cases for custom domains like `github.example.com`, `gitlab.example.com` and `bitbucket.example.com`? In this cases I also expect the full GitHub branch pathname.

I expect we can not determinate `git.example.com`, but that's acceptable for me at the moment.

frontend/packages/topology/src/utils/topology-utils.ts
```diff
@@ -44,13 +44,28 @@ export const getCheDecoratorData = (consoleLinks: K8sResourceKind[]): CheDecorat
   };
 };
 
+const getFullGitURL = (gitUrl: GitUrlParse.GitUrl, branch?: string) => {
+  const baseUrl = `https://${gitUrl.source}/${gitUrl.owner}/${gitUrl.name}`;
```
jerolimov, line 48: Based on the learnings in #8006 we should use gitUrl.resource here instead, or?

```suggestion
  const baseUrl = `https://${gitUrl.resource}/${gitUrl.owner}/${gitUrl.name}`;
```

frontend/packages/topology/src/utils/topology-utils.ts
```diff
@@ -44,13 +44,28 @@ export const getCheDecoratorData = (consoleLinks: K8sResourceKind[]): CheDecorat
...
+  switch (gitUrl.resource) {
+    case 'github.com':
+      return `${baseUrl}/tree/${branch}`;
+    case 'gitlab.com':
+      return `${baseUrl}/-/tree/${branch}`;
+    case 'bitbucket.org':
+      return `${baseUrl}/src/${branch}`;
+    default:
+      return baseUrl;
+  }
```
jerolimov, line 61: To support `github.example.com` and so on I would use `gitUrl.resource.includes("gitlab")` etc. here instead. Wdyt?

jerolimov (changes requested):
I like the refactoring and direction but have some small ideas here. Esp. we should support also custom domains when possible.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
rottencandy: This creates problems for repos where the default branch is not master. For ex: https://github.com/redhat-developer/s2i-dotnetcore-ex or https://github.com/eggheadio/egghead-next
The edit icon then gets this incorrect branch name and adds it to the url, which leads to 404. I think that not having a branch in url is better than assuming an incorrect branch.
But I can undo this if it is not needed.

frontend/packages/topology/src/data-transforms/__tests__/data-transformer.spec.ts
rottencandy: Updated.

frontend/packages/topology/src/utils/topology-utils.ts
rottencandy: Almost forgot this :P thanks.

frontend/packages/topology/src/utils/topology-utils.ts
rottencandy: Updated.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
jerolimov: Sounds valid for me. You can keep this.

I thought the gitRef was set in these cases. Can you check if we "lost" the branch info in the add flow for these repos?

rottencandy: Looks like we do not try to fetch the the default branch if the ref isn't provided, so it never gets set in the first place:
https://github.com/openshift/console/blob/master/frontend/packages/git-service/src/services/github-service.ts#L83
The git APIs do not require branch to be supplied, and will operate on the default branch.

So unless a ref is provided by user, it will not be set.

frontend/packages/topology/src/utils/topology-utils.ts
```diff
@@ -44,13 +44,28 @@ export const getCheDecoratorData = (consoleLinks: K8sResourceKind[]): CheDecorat
   };
 };
 
+const getFullGitURL = (gitUrl: GitUrlParse.GitUrl, branch?: string) => {
```
nemesis09, line 47: it would be great if we could add some tests to this util.

frontend/packages/topology/src/utils/topology-utils.ts
```diff
@@ -44,13 +44,28 @@ export const getCheDecoratorData = (consoleLinks: K8sResourceKind[]): CheDecorat
...
+  if (gitUrl.resource.includes('gitlab')) {
+    return `${baseUrl}/-/tree/${branch}`;
+  }
+  if (gitUrl.resource.includes('bitbucket')) {
+    return `${baseUrl}/src/${branch}`;
+  }
+  return baseUrl;
+};
+
 export const getEditURL = (vcsURI?: string, gitBranch?: string, cheURL?: string) => {
```
nemesis09, line 64: We could also add tests for this util under the scope of this PR.

frontend/packages/dev-console/src/utils/resource-label-utils.ts
nemesis09: ParseBitbucketUrl() seems to provide branch as well which is used for default branch if a ref is not provided (https://github.com/openshift/console/blob/a07aa74a966fb75a16ecbe86a73fe0e01a86d497/frontend/packages/git-service/src/services/bitbucket-service.ts#L37)

rottencandy: It appears to extract branch name from the URL, and return 'master' if there is no branch.

https://github.com/codefresh-io/parse-bitbucket-url/blob/master/index.js#L189

jerolimov (approved):
Tested it with GitHub and GitLab and works fine for me. Also know that the URL schema is valid for BitBucket. I noticed that it would also be cool at attach the contextDir if used, but this is another issue. ;)

/lgtm

rohitkrai03 (approved):
/approve

PR #8029, opened 2021-02-02
Bug 1924078: Make Topology quick search View all results footer sticky

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5402

**Analysis / Root cause**: 
The link to all results was added to the data list rather than being a separate component and styling the list to be scrollable.

**Solution Description**: 
Make the all results link a component separate from the list. Update the layout to allow the list to scroll rather than the entire left side area. Use flex display for modal rather than setting an adjusted percentage height on the search content.

image

image

image

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
@@ -1,9 +1,9 @@
 .odc-quick-search-content {
-  height: calc(100% - 60px);
+  flex: 1;
+  overflow-y: hidden;
   &__list {
     width: 40%;
     height: 100%;
-    overflow-y: auto;
   }
   &__details {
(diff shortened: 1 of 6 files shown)


frontend/packages/topology/src/components/quick-search/QuickSearchContent.scss
```diff
@@ -1,9 +1,9 @@
 .odc-quick-search-content {
-  height: calc(100% - 60px);
+  flex: 1;
```
andrewballantyne, line 2: 🎉

invincibleJai:
/approve

Verified changes , works as expected

debsmita1:
verified the changes
/lgtm

PR #8032, opened 2021-02-02
Bug 1920677: Fix for i18n for dev console guided tour

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/OCPBUGSM-23788

**Analysis / Root cause**: 
Guided tour plugin used `i18next.t` rather than the `%<string>%` format for getting localized strings.

**Solution Description**: 
Update to use the `%<string>%` format.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/guided-tour/index.ts
+++ b/frontend/packages/dev-console/src/components/guided-tour/index.ts
@@ -1,5 +1,4 @@
 import { TourDataType } from '@console/app/src/components/tour';
-import i18n from '@console/internal/i18n';
 import {
   perspectiveSwitcherTourText,
   searchTourText,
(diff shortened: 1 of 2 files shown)


invincibleJai:
@jeff-phillips-18 we have some miss in `cloudShellConfirmationModal` where `i18n.t` is used  as well,  it falls under console-app.

and these seems issue with frontend CI

frontend/packages/dev-console/src/components/guided-tour/index.ts
```diff
@@ -11,50 +10,47 @@ const getSelector = (id: string): string => `[data-tour-id="${id}"]`;
 
 export const getGuidedTour = (): TourDataType => ({
   intro: {
-    heading: i18n.t('devconsole~Welcome to the Developer Perspective!'),
+    heading: '%devconsole~Welcome to the Developer Perspective!%',
```
invincibleJai, line 13: in plugin we even have comments along with `%` , is that not needed

```suggestion
     // t('devconsole~Welcome to the Developer Perspective!')
    heading: '%devconsole~Welcome to the Developer Perspective!%',
```

PR #8041, opened 2021-02-03
Cypress: fix annotations test flake

dtaylor113:
- moved create & delete example config map out of before/after hooks, so failure won't affect login/logout.
- used more exact assertions, Ex: `.should('have.text', '2 annotations')` test

### DIFF
--- a/frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
@@ -53,6 +53,12 @@ const nameValueEquals = (row: JQuery<HTMLElement>, name: string, value: string) 
   });
 };
 
+const numberOfAnnotationsEquals = (count: number) => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000); // wait a couple of seconds for watcher to update page
+  cy.byTestID('edit-annotations').contains(`${count} annotations`);
+};
+
 describe('Annotations', () => {
   before(() => {
     cy.login();
@@ -60,7 +66,6 @@ describe('Annotations', () => {
     nav.sidenav.switcher.changePerspectiveTo('Administrator');
     nav.sidenav.switcher.shouldHaveText('Administrator');
     cy.createProject(testName);
-    createExampleConfigMapInstance();
   });
 
   afterEach(() => {
@@ -68,7 +73,6 @@ describe('Annotations', () => {
   });
 
   after(() => {
-    deleteExampleConfigMapInstance();
     cy.deleteProject(testName);
     cy.logout();
   });
@@ -91,10 +95,12 @@ describe('Annotations', () => {
       },
     ];
 
+    createExampleConfigMapInstance();
+
     cy.url().should('include', `/k8s/ns/${testName}/configmaps/example`);
     detailsPage.isLoaded();
     detailsPage.titleShouldContain('example');
-    cy.byTestID('edit-annotations').contains('0 annotations');
+    numberOfAnnotationsEquals(0);
 
     cy.log('Add annotations');
     detailsPage.clickPageActionFromDropdown('Edit annotations');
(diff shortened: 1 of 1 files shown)


frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
```diff
@@ -115,7 +115,7 @@ describe('Annotations', () => {
     modal.shouldBeClosed();
 
     cy.log('Verify first two saved annotations');
-    cy.byTestID('edit-annotations').contains('2 annotation'); // 3rd annotation without key should not have been saved
+    cy.byTestID('edit-annotations').contains('2 annotations'); // 3rd annotation without key should not have been saved
```
spadgett, line 118: This isn't a bad change, but I'm not sure it fixes the flake? If `2 annotations` matches, `2 annotation` would also match since it's a substring check

dtaylor113: From what I've read it seems like `cy.contains` should work that way.  But clearly it didn't.   Changed assertion method, it may help.
image

spadgett: @dtaylor113 Could it be a race where we check the annotation text before the watch updates the page (then the page updates after the failure but before the screenshot)?

If it were the substring problem, I'd expect the test to fail every time.

dtaylor113: I'm not sure what cypress can wait upon if that is the case, other than an arbitary `wait(5000)`.   I thought `.contains()` continually checks the DOM as it is returning a DOM element.   I was leaning more towards perhaps `.contains()` doesn't do substring, rather full word 'contains', or perhaps the space in the middle, idk.

dtaylor113: `.should('have.text', '2 annotations')` is more precise.

dtaylor113:
/retest
1 passed, 2nd run

frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
spadgett: > I was leaning more towards perhaps `.contains()` doesn't do substring, rather full word 'contains', or perhaps the space in the middle, idk.

@dtaylor113 If that were the cause, wouldn't it fail every time? It only fails about 1 time in 20, which suggests a race to me.

spadgett:
> /test all
> 1 passed, 2nd run

@dtaylor113 It only flakes about 5% of the time, so this won't be a great way to check. It would be better to run just this scenario in a loop locally and see if you can't get it to fail.

frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
dtaylor113: Might have something to do with https://docs.cypress.io/guides/core-concepts/retry-ability.html#Only-the-last-command-is-retried

frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
```diff
@@ -53,22 +53,26 @@ const nameValueEquals = (row: JQuery<HTMLElement>, name: string, value: string)
   });
 };
 
+const numberOfAnnotationsEquals = (count: number) => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000); // wait a couple of seconds for watcher to update page
+  cy.byTestID('edit-annotations').contains(`${count} annotations`);
```
dtaylor113, line 59: Hi @spadgett, added a wait

frontend/packages/integration-tests-cypress/tests/crud/annotations.spec.ts
```diff
@@ -53,22 +53,26 @@ const nameValueEquals = (row: JQuery<HTMLElement>, name: string, value: string)
   });
 };
 
+const numberOfAnnotationsEquals = (count: number) => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000); // wait a couple of seconds for watcher to update page
```
spadgett, line 58: This might be necessary in the end, but I feel like we don't really understand the problem. Were you able to reproduce outside of CI by running just annotations spec in a loop until it fails?

My understanding is that the entire `cy.byTestID('edit-annotations').contains(`${count} annotations`)` statement should be retried until it's true. With our long timeouts, I'd expect it to work (particularly when the screenshot shows the value we expect).

dtaylor113: I agree, I'll keep looking, but I think this is a workaround until a better solution is understood.

dtaylor113:
Closing due to inactivity, will reinvestigate if need arises

PR #8046, opened 2021-02-04
Bug 1924955: Fix that image containers are fetched from external container registries (which doesn't work for private image containers)

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3062
https://issues.redhat.com/browse/OCPBUGSM-24359
https://bugzilla.redhat.com/show_bug.cgi?id=1924955

**Analysis / Root cause**:
When the user provides a Image Pull Secret (a Secret with type `kubernetes.io/dockerconfigjson`) to access private image registry the Add/Import page validates the image name but the Pod doesn't start later.

The "Deploy Image" pages makes an API call to `.../$namespace/imagestreamimports` to verify if the cluster could connect to the external image registry and which image tags are available. The form could be only submit if this was successful. The imagestreamimports endpoint uses the shared Image Pull Secrets of the namespace.

But after creating this Deployment (or DeploymentConfig) the container image could not be pulled from the private registry. All Pods are failing with `ErrImagePull` and `ImagePullBackOff` errors. The reason is that the cluster tries to load the container image directly from the external URL without the provided Secret.

As workaround it was possible to assign the Image Pull Secret as `imagePullSecrets:` in the "default" ServiceAccount of the name or provide it (also as `imagePullSecrets:` in the Deployment/DeploymentConfig template section for the Pod.

**Solution Description**: 
Because the `imagestreamimports` endpoint uses all provided Secrets we decided not to update the page (yet) to select a Secret and assign it to the Deployment.

Instead of that, this PR changes the URL of a Deployment from the external image registry to the internal image registry. We already created a ImageStream which pulls the container image successfully.

For this the console needs just to set the image container name (as it does already). Whenever this `image:` was changed in a Deployment it was automatically overridden by an ImageStream admission webhook. This webhook supports a `referencePolicy` which is `Source` by default. This PR changes this value to `Local` so that the cluster internal image registry was used instead of the external image registry. Whenever the Deployment or ImageStream was changed or triggered, the Deployment contains now a link to the internal image registry.

See also:
* https://docs.openshift.com/container-platform/4.6/rest_api/image_apis/imagestream-image-openshift-io-v1.html
* https://docs.openshift.com/container-platform/4.6/openshift_images/using-imagestreams-with-kube-resources.html
* https://issues.redhat.com/browse/BUILD-100
* https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
* https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

None. UI is not changed.

Added some tests for `createOrUpdateImageStream`

**Test setup:**
1. You need a private container image on an external registry like Docker Hub or Quay with credentials to read the image. (Feel free to ask me in Slack for some credentials for a code review.)
2. Open the developer perspective
3. Select Secrets > Create > Image Pull Secret and add enter your credentials. Use `docker.io` for the Docker Hub and `quay.io` for Quay as Registry Server Address.
4. Go to Add > Container Image and import your private container image as Deployment or DeploymentConfig. (If you have the Serverless Operator installed you can also verify that the Knative Services works already before this change.)
5. Check in Topology that the Pods can start successful.
6. I also verified that an updated image in the external image registry could be updated via `oc import-image nodeinfo-private`.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


rottencandy (approved):
Tested. Works fine
/lgtm

frontend/packages/dev-console/src/components/import/__tests__/deployImage-submit-utils.spec.ts
```diff
@@ -63,6 +64,215 @@ describe('DeployImage Submit Utils', () => {
...
+                'openshift.io/generated-by': 'OpenShiftWebConsole',
+                'openshift.io/imported-from': 'myimage',
+              },
+              from: { kind: 'DockerImage', name: 'myimage' },
+              importPolicy: { insecure: false },
+              referencePolicy: { type: 'Local' },
+            },
+          ],
+        },
+      };
```
invincibleJai, line 112: nit: this data is used even in below it blocks , we can have it in separate mock data file or before each and use it (update only particular spec field if needed) WDYT?

invincibleJai (commented):
Thanks @jerolimov 🎉 

Looks good and have verified works as expected, one nit see if it make sense

frontend/packages/dev-console/src/components/import/__tests__/deployImage-submit-utils.spec.ts
jerolimov: +:100: Absolutely. Extracted this into the mock file which contains also the form data. PTAL

PR #8047, opened 2021-02-04
Bug 1925087: Add i18n support to the Secret modals and add new hint for the servername

jerolimov:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1925087

**Analysis / Root cause**: 
While investigating why private container images doesn't start successfully (fix for this here #8046) I notice that I started with an invalid domainname for the external image registry. I used `docker.com` instead of `docker.io` for Docker Hub. For this I would recommend a small hint under the "Registry Server Address" field.

**Solution Description**: 
Before I want to add the hint I added i18n support to the Create and Edit Secret page. This component is also used in the Secret modal in the Add flow and the Start Pipeline modal.

After that I added this hint under the Registry Server Address:

> For example quay.io or docker.io

Secret page Create drop down
image

Create Key/Value Secret
image

Create Image Pull Secret
image

Create Source Secret
image

Create Webhook Secret
image

On the Add Page > Import from Git > Create Image source secret:
image

On the Add Page > Import from Container Image > Create Image Pull Secret modal
image

With installed OpenShift Pipelines Operator
Pipelines > Select a Pipeline > Start modal > Show Credential options > Add Secret feature
image

2. The hint itself:
Screenshot from 2021-02-22 00-59-54

Unchanged

**Test setup:**
* Create a secret in the Secret list (from the admin or developer perspective)
* Open the secret modal from the Import Git flow in the developer perspective
* Open the secret modal from the Import image container flow in the developer perspective
* Open the secret modal from the Start Pipeline dialog in the developer perspective with installed OpenShift Pipelines operator.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


jerolimov:
@openshift/team-devconsole-ux What do you think about adding this hint on the Create/Edit Secret page?

Screenshot from 2021-02-04 12-49-06

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -55,13 +59,17 @@ export type BasicAuthSubformState = {
...
-  [SecretTypeAbstraction.generic]:
-    'Key/value secrets let you inject sensitive data into your application as files or environment variables.',
-  [SecretTypeAbstraction.source]: 'Source secrets let you authenticate against a Git server.',
-  [SecretTypeAbstraction.image]:
-    'Image pull secrets let you authenticate against a private image registry.',
-  [SecretTypeAbstraction.webhook]: 'Webhook secrets let you authenticate a webhook trigger.',
+const secretFormExplanation: Record<SecretTypeAbstraction, (t: TFunction) => string> = {
+  [SecretTypeAbstraction.generic]: (t) =>
+    t(
+      'public~Key/Value Secrets let you inject sensitive data into your application as files or environment variables.',
```
rebeccaalpert, line 65:
```suggestion
      'public~Key/value secrets let you inject sensitive data into your application as files or environment variables.',
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -55,13 +59,17 @@ export type BasicAuthSubformState = {
...
+    ),
+  [SecretTypeAbstraction.source]: (t) =>
+    t('public~Source Secrets let you authenticate against a Git server.'),
+  [SecretTypeAbstraction.image]: (t) =>
+    t('public~Image Pull Secrets let you authenticate against a private image registry.'),
+  [SecretTypeAbstraction.webhook]: (t) =>
+    t('public~Webhook Secrets let you authenticate a webhook trigger.'),
```
rebeccaalpert, line 72:
```suggestion
    t('public~Webhook secrets let you authenticate a webhook trigger.'),
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -55,13 +59,17 @@ export type BasicAuthSubformState = {
...
-  [SecretTypeAbstraction.webhook]: 'Webhook secrets let you authenticate a webhook trigger.',
+const secretFormExplanation: Record<SecretTypeAbstraction, (t: TFunction) => string> = {
+  [SecretTypeAbstraction.generic]: (t) =>
+    t(
+      'public~Key/Value Secrets let you inject sensitive data into your application as files or environment variables.',
+    ),
+  [SecretTypeAbstraction.source]: (t) =>
+    t('public~Source Secrets let you authenticate against a Git server.'),
+  [SecretTypeAbstraction.image]: (t) =>
+    t('public~Image Pull Secrets let you authenticate against a private image registry.'),
```
rebeccaalpert, line 70:
```suggestion
    t('public~Image pull secrets let you authenticate against a private image registry.'),
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -55,13 +59,17 @@ export type BasicAuthSubformState = {
...
-  [SecretTypeAbstraction.image]:
-    'Image pull secrets let you authenticate against a private image registry.',
-  [SecretTypeAbstraction.webhook]: 'Webhook secrets let you authenticate a webhook trigger.',
+const secretFormExplanation: Record<SecretTypeAbstraction, (t: TFunction) => string> = {
+  [SecretTypeAbstraction.generic]: (t) =>
+    t(
+      'public~Key/Value Secrets let you inject sensitive data into your application as files or environment variables.',
+    ),
+  [SecretTypeAbstraction.source]: (t) =>
+    t('public~Source Secrets let you authenticate against a Git server.'),
```
rebeccaalpert, line 68:
```suggestion
    t('public~Source secrets let you authenticate against a Git server.'),
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -27,14 +29,16 @@ export enum SecretTypeAbstraction {
...
 
-const secretDisplayType = (abstraction: SecretTypeAbstraction) => {
+const secretDisplayType = (isCreate: boolean, abstraction: SecretTypeAbstraction, t: TFunction) => {
   switch (abstraction) {
     case 'generic':
-      return 'Key/Value';
+      return isCreate ? t('public~Create Key/Value Secret') : t('public~Edit Key/Value Secret');
     case 'image':
-      return 'Image Pull';
+      return isCreate ? t('public~Create Image Pull Secret') : t('public~Edit Image Pull Secret');
```
rebeccaalpert, line 37:
```suggestion
      return isCreate ? t('public~Create image pull secret') : t('public~Edit image pull secret');
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -27,14 +29,16 @@ export enum SecretTypeAbstraction {
   webhook = 'webhook',
 }
 
-const secretDisplayType = (abstraction: SecretTypeAbstraction) => {
+const secretDisplayType = (isCreate: boolean, abstraction: SecretTypeAbstraction, t: TFunction) => {
   switch (abstraction) {
     case 'generic':
-      return 'Key/Value';
+      return isCreate ? t('public~Create Key/Value Secret') : t('public~Edit Key/Value Secret');
```
rebeccaalpert, line 35:
```suggestion
      return isCreate ? t('public~Create key/value secret') : t('public~Edit key/value secret');
```

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -27,14 +29,16 @@ export enum SecretTypeAbstraction {
...
     case 'image':
-      return 'Image Pull';
+      return isCreate ? t('public~Create Image Pull Secret') : t('public~Edit Image Pull Secret');
     default:
-      return _.upperFirst(abstraction);
+      return isCreate
+        ? t('public~Create {{secretType}} Secret', { secretType: _.upperFirst(abstraction) })
+        : t('public~Edit {{secretType}} Secret', { secretType: _.upperFirst(abstraction) });
```
rebeccaalpert, line 41: The use of upperFirst probably needs to be updated to match the capitalization standards.

frontend/packages/dev-console/src/components/import/CreateSecretModal.tsx
```diff
@@ -21,15 +22,16 @@ type Props = CreateSecretModalProps & ModalComponentProps;
 const getSecretForm = (type: SecretTypeAbstraction) => {
   switch (type) {
     case SecretTypeAbstraction.source:
-      return withSecretForm(SourceSecretForm, true);
+      return withSecretForm(withTranslation()(SourceSecretForm), true);
```
christianvogt, line 25: The components should be exported with translation support built in and not require the consumers to wrap them.

rebeccaalpert (changes requested):
Thanks for the PR! A few comments mostly related to word case and using the TFunction as a prop. I didn't call out every instance of TFunction being used as a prop or argument, but we typically avoid doing that in management console. It's better to move the code into a wrapped class or function component with the hook.

@spadgett has more insight into our code coverage at this point since he's the primary code reviewer for the team. I'm not sure if this is being worked on by anyone else.

frontend/packages/dev-console/src/components/import/CreateSecretModal.tsx
```diff
@@ -42,8 +44,8 @@ const CreateSecretModal: React.FC<Props> = ({ close, namespace, save, secretType
       onSave={handleSave}
       fixed={{ metadata: { namespace } }}
       secretTypeAbstraction={secretType}
-      titleVerb="Create"
       isCreate
+      t={t}
```
christianvogt, line 48: You shouldn't have to pass in `t`. The component should handle it themselves using the hook or the hoc.

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -444,27 +456,32 @@ class ConfigEntryForm extends React.Component<ConfigEntryFormProps, ConfigEntryF
...
+              aria-describedby={`${this.props.id}-address-help`}
               type="text"
               name="address"
               onChange={this.onAddressChanged}
               value={this.state.address}
               required
             />
           </div>
+          <p className="help-block" id={`${this.props.id}-address-help`}>
+            {t('public~Registry servername like quay.io or docker.io')}
```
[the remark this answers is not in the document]
christianvogt: The label says `Registry Server Address` but here you use `name`.

maybe the hint can be simplified to `eg. quay.io or docker.io`

Maybe get a suggestion from @openshift/team-devconsole-ux

jerolimov:
Oh wow, big thanks @rebeccaalpert . Will go over all this feedback step by step :)

jerolimov:
@rebeccaalpert @christianvogt Thanks for the feedback.

I refactored my change to use `withTranslation` for the old class components. I will tried to do not refactor too much at the same time. But I need to replaced the `withSecretForm` wrapper around SecretFormComponent with a pure class "SecretFormWrapper" which now calls `secretFormFactory` when render the `<Subform />`

I also changed some old strings to your mentioned "sentence case", I hope this is also fine. You can checkout all three commits separably to verify this step by step.

jerolimov:
Force pushed a small update when editing a `Secret`.

jerolimov:
>   1) Interacting with a single namespace install mode Operator (Couchbase Operator)
>        displays Couchbase Operator CouchbaseCluster creation form:
>      AssertionError: Timed out retrying after 30000ms: Expected to find element: `[data-test-id="dropdown-button"]`, but never found it.
>       at Context.eval (https://console-openshift-console.apps.ci-op-2rcwfjx1-75d12.origin-ci-int-gce.dev.openshift.com/__cypress/tests?p=tests/operator-install-single-namespace.spec.ts:13639:57)
> 
>   2) Interacting with a single namespace install mode Operator (Couchbase Operator)
>        creates a Couchbase Operator CouchbaseCluster instance via the form:
>      AssertionError: Timed out retrying after 30000ms: Expected to find element: `[data-test="create-dynamic-form"]`, but never found it.
>       at Context.eval (https://console-openshift-console.apps.ci-op-2rcwfjx1-75d12.origin-ci-int-gce.dev.openshift.com/__cypress/tests?p=tests/operator-install-single-namespace.spec.ts:13635:8)
> 
>   3) Interacting with a single namespace install mode Operator (Couchbase Operator)
>        displays details about Couchbase Operator CouchbaseCluster instance on the "Details" tab:
>      AssertionError: Timed out retrying after 30000ms: Expected to find element: `[data-test-operand-link="cb-example"]`, but never found it.
>       at Context.eval (https://console-openshift-console.apps.ci-op-2rcwfjx1-75d12.origin-ci-int-gce.dev.openshift.com/__cypress/tests?p=tests/operator-install-single-namespace.spec.ts:13646:60)

/retest

jerolimov:
> rateLimitExceeded

/retest

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -27,14 +29,16 @@ export enum SecretTypeAbstraction {
   webhook = 'webhook',
 }
 
-const secretDisplayType = (abstraction: SecretTypeAbstraction) => {
+const secretDisplayType = (isCreate: boolean, abstraction: SecretTypeAbstraction, t: TFunction) => {
```
rebeccaalpert, line 32: It's better to move code like this into a block that already has useTranslation or withTranslation rather than passing in a TFunction.

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -55,13 +59,17 @@ export type BasicAuthSubformState = {
...
 };
 
-const secretFormExplanation = {
-  [SecretTypeAbstraction.generic]:
-    'Key/value secrets let you inject sensitive data into your application as files or environment variables.',
-  [SecretTypeAbstraction.source]: 'Source secrets let you authenticate against a Git server.',
-  [SecretTypeAbstraction.image]:
-    'Image pull secrets let you authenticate against a private image registry.',
-  [SecretTypeAbstraction.webhook]: 'Webhook secrets let you authenticate a webhook trigger.',
+const secretFormExplanation: Record<SecretTypeAbstraction, (t: TFunction) => string> = {
```
rebeccaalpert, line 62: It's better to move code like this into a block that already has useTranslation or withTranslation rather than passing in a TFunction.

frontend/public/components/secrets/create-secret.tsx
```diff
@@ -393,10 +414,14 @@ type ConfigEntryFormProps = {
   id: number;
   entry: Object;
   onChange: Function;
+  t: TFunction;
 };
```
rebeccaalpert, line 419: Instead of doing this, you can also import the type WithTranslation from react-i18next and do something like 
```
type ConfigEntryFormState = WithTranslation & {...};
```

You have to disable the linter when you import it: 
```
/* eslint-disable-next-line */
import { withTranslation, WithTranslation } from 'react-i18next';
```

rebeccaalpert (changes requested):
Looking good! Thank you for making the casing changes; it's a small thing, but it makes a big difference for our overall consistency and quality. I just had a few comments about passing the TFunction around, as well as a suggestion to use the existing WithTranslation type instead of rolling your own.

frontend/public/components/secrets/create-secret.tsx
jerolimov: But this is before and with this PR a utils function, not a component ?! Hmm, don't know how I could change this without refactoring much more code.

frontend/public/components/secrets/create-secret.tsx
jerolimov: But this is before and with this PR a utils function, not a component ?! Hmm, don't know how I could change this without refactoring much more code.

frontend/public/components/secrets/create-secret.tsx
jerolimov: Removed the `t: TFunction` prop here, missed that. It looks now like this https://github.com/openshift/console/blob/ce0918bdd9c249aa6056ba90b7b40894d6618000/frontend/public/components/secrets/create-secret.tsx#L413-L422

I could not use `WithT` from `react-i18n` because I need an optional here to make the t not required for the component consumers.

jerolimov:
@rebeccaalpert updated again and added some responses above. I don't know how I could workaround this without changing to much. Any ideas? :)

frontend/public/components/secrets/create-secret.tsx
jerolimov: I tested it again and it works fine now. So I replaced `{ t?: TFunction }` with predefined type `WithT`:

https://github.com/openshift/console/blob/8c8ca211efcdb1c4bd9f71a15211ad998ef1b309/frontend/public/components/secrets/create-secret.tsx#L413-L422

frontend/public/components/secrets/create-secret.tsx
rebeccaalpert: Oh, I see, `secretFormExplanation` is being used in two components in this file. Let me see if I can find how we've handled this in other PRs. I know we've come across this before.

frontend/public/components/secrets/create-secret.tsx
rebeccaalpert: I know this isn't new code you added. :) This is just something that's come up in other reviews; I've been dinged for it myself.

It looks like `secretDisplayType` is only called in `SecretFormWrapper.` Could we move this const into that component so we avoid passing the TFunction around?

frontend/public/components/secrets/create-secret.tsx
rebeccaalpert: That's fine by me! Thanks so much. This is something that's come up when Sam's looked at my code before. Appreciate the extra work on your end.

rebeccaalpert (approved):
I'm going to go ahead and approve this. I'm sure Christian or Sam will chime in if there's stuff I missed. :) Thanks for all your work on this.

rottencandy (approved):
/lgtm

rohitkrai03 (changes requested):
@jerolimov Code looks good to me overall but there's one unresolved comment by Rebecca i think.

frontend/public/components/secrets/create-secret.tsx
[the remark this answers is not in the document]
jerolimov: Changed it to `For example quay.io or docker.io`

jerolimov:
> @jerolimov Code looks good to me overall but there's one unresolved comment by Rebecca i think.

@rohitkrai03 Hmm, I think they are all resolved.

frontend/public/components/secrets/create-secret.tsx
rohitkrai03: @jerolimov This was the unresolved comment that I was talking about.

jerolimov: @rohitkrai03 @rebeccaalpert  Ah thanks for pointing this out. I'm not a big fan of including this inline to an (already long) render function. Esp. where I tried to also keep the diff understandable. In this case I would prefer a "private" local helper function like this one. Also because the other one (one comment below this) has a similar problem but is fine. I think it makes sense that both "translate helpers" work similar here.

rohitkrai03 (approved):
/approve

PR #8048, opened 2021-02-04
Bug 1924968: Monitoring: Add some missing translations

kyoto:
Mostly for the list row filtering options.

### DIFF
--- a/frontend/public/components/monitoring/metrics.tsx
+++ b/frontend/public/components/monitoring/metrics.tsx
@@ -700,7 +700,7 @@ const QueryTable_: React.FC<QueryTableProps> = ({
 
   let columns, rows;
   if (data.resultType === 'scalar') {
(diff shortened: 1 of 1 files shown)


kyoto:
FYI @rebeccaalpert

rebeccaalpert (approved):
Looks good to me. I have one comment about translating the filter values - we should probably internationalize the row values to match since they may differ significantly where they're translated into a different language. I've used switch statements to return translations for this kind of stuff elsewhere.

Screen Shot 2021-02-04 at 10 30 35 AM

kyoto:
This was mostly implemented by https://github.com/openshift/console/pull/8571, so I changed this PR to just add a couple of missing translations and rebased.

andrewballantyne:
/lgtm

Looks like a tiny PR, let's get it merged.

PR #8051, opened 2021-02-04
Bug 1925024: removes form validation from net(sasl/tls) section for kafkaSource

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5458

**Analysis / Root cause**: 
In form view for kafksSource if sasl/tls are enebled then other fields i.e user , password or caCert, cert, key  where shown as required field which shouldn't be the case

**Solution Description**: 
Removed required validation from net section and rely on API validation for same

image

**Test:**

image

**Test setup:**
- Install Serverless operator
- create CR for Knative Kafka (enable source) along with serving and eventing

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/utils/__tests__/create-eventsources-utils.spec.ts
+++ b/frontend/packages/knative-plugin/src/utils/__tests__/create-eventsources-utils.spec.ts
@@ -18,6 +18,8 @@ import {
   getEventSourceData,
   sanitizeKafkaSourceResource,
   sanitizeSourceToForm,
+  isSecretKeyRefPresent,
+  getKafkaSourceResource,
 } from '../create-eventsources-utils';
 import { getDefaultEventingData, Kafkas } from './knative-serving-data';
 import { EventSourceFormData, EventSources } from '../../components/add/import-types';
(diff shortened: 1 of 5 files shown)


karthikjeeyar:
We could add a test spec against the net section
```
    it('should not throw error if net section is empty', async () => {
      const defaultEventingData = getDefaultEventingData(EventSources.KafkaSource);
      const mockData = _.cloneDeep(defaultEventingData);
      delete mockData.formData.data.KafkaSource.net;
      await eventSourceValidationSchema(t)
        .resolve({ value: mockData })
        .isValid(mockData)
        .then((valid) => expect(valid).toEqual(true));
    });
   ```

File: frontend/packages/knative-plugin/src/utils/__tests__/create-eventsources-utils.spec.ts [hunk 1]
```diff
@@ -297,4 +299,83 @@ describe('sanitizeSourceToForm always returns valid Event Source', () => {
     expect(formData.application.name).toEqual('');
     expect(formData.application.selectedKey).toEqual(UNASSIGNED_KEY);
   });
+
```
karthikjeeyar, line 302: Pls Add the null test as well
```suggestion
  expect(isSecretKeyRefPresent(null)).toBe(false);
```

invincibleJai: added

karthikjeeyar:
Verified locally, works fine
/lgtm

PR #8052, opened 2021-02-04
Bug 1925020: Console demo plugin deployment image should not point to dockerhub

jhadvig:
Points to https://quay.io/repository/jhadvig/console-demo-plugin

/assign @spadgett @vojtechszocs


spadgett (approved):
/lgtm

PR #8055, opened 2021-02-04
Bug 1922264: Fix rendering of access-mode selector

vbnrh:
This commit fixes an issue where the radio buttons of the access mode selector are not rendered properly when parent pvc of the snapshot is deleted.
The commit removes the dependency of rendering based on the pvc load error.
 
below screenshots of the modal after removing the parent pvcs

### DIFF
--- a/frontend/packages/console-app/src/components/access-modes/access-mode.tsx
+++ b/frontend/packages/console-app/src/components/access-modes/access-mode.tsx
@@ -24,8 +24,7 @@ export const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: stri
   );
 
 export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) => {
-  const { formClassName, pvcResource, onChange, loadError, loaded, provisioner } = props;
-
+  const { formClassName, pvcResource, onChange, loaded, provisioner } = props;
   const pvcInitialAccessMode = pvcResource ? getPVCAccessModes(pvcResource, 'value') : '';
 
   const [allowedAccessModes, setAllowedAccessModes] = React.useState<string[]>();
@@ -69,24 +68,19 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) => 
       {loaded &&
         allowedAccessModes &&
         accessModeRadios.map((radio) => {
-          let radioObj = null;
           const disabled = !allowedAccessModes.includes(radio.value);
-          allowedAccessModes.forEach((mode) => {
-            const checked =
-              !disabled && !loadError ? radio.value === accessMode : radio.value === mode;
-            radioObj = (
-              <RadioInput
-                {...radio}
-                key={radio.value}
-                onChange={onAccessModeChange}
-                inline
-                disabled={disabled}
-                checked={checked}
-                name="accessMode"
-              />
-            );
-          });
-          return radioObj;
+          const checked = radio.value === accessMode;
+          return (
+            <RadioInput
+              {...radio}
+              key={radio.value}
+              onChange={onAccessModeChange}
+              inline
+              disabled={disabled}
+              checked={checked}
+              name="accessMode"
+            />
+          );
         })}
       {(!loaded || !allowedAccessModes) && <div className="skeleton-text" />}
     </FormGroup>
--- a/frontend/public/components/utils/storage-class-dropdown.tsx
+++ b/frontend/public/components/utils/storage-class-dropdown.tsx
@@ -4,6 +4,7 @@ import * as fuzzy from 'fuzzysearch';
 
 import { Firehose, LoadingInline, Dropdown, ResourceName, ResourceIcon } from '.';
 import { isDefaultClass } from '../storage-class';
+import * as classNames from 'classnames';
 
 /* Component StorageClassDropdown - creates a dropdown list of storage classes */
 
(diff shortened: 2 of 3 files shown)


frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -69,24 +68,19 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) =>
...
+          return (
+            <RadioInput
+              {...radio}
+              key={radio.value}
+              onChange={onAccessModeChange}
+              inline
+              disabled={disabled}
+              checked={checked}
+              name="accessMode"
+            />
```
cloudbehl, line 82: `accessMode` is `""` then what will be checked?

vbnrh: the 'ReadWriteOnce' will be checked..

cloudbehl: sounds good!

frontend/public/components/utils/storage-class-dropdown.tsx
```diff
@@ -157,6 +157,7 @@ export class StorageClassDropdownInner extends React.Component<
                   ? `${this.props.hideClassName} control-label`
                   : 'control-label'
               }
+              hidden={this.props?.hideLabel}
```
spadgett, line 160: This seems like it would break accessibility of this field. What is the purpose of the label change?

cloudbehl: @spadgett We want to add an asterisk(required field) on the storage class label.

spadgett: Why not add a `required` prop to the storage class dropdown instead of hiding the label and rendering your own? This is a simpler approach and doesn't break accessibility.

vbnrh: @spadgett i have tried the required prop, it does not make the label have the asterisk as expected. hence I wanted to make the label hide-able through prop and  render our required label through the <FormGroup />

spadgett: We should fix the bug with the `required` prop so that the label has the asterisk instead.

vbnrh: fixed.. should be working as expected now

frontend/public/components/utils/storage-class-dropdown.tsx
```diff
@@ -105,6 +105,18 @@ export class StorageClassDropdownInner extends React.Component<
...
+  getLabelClassName = () => {
+    const classNames: string[] = [];
+    this.props.hideClassName
+      ? classNames.push(`${this.props.hideClassName} control-label`)
+      : classNames.push('control-label');
+
+    if (this.props.required) {
+      classNames.push('co-required');
+    }
+    return classNames.join(' ');
```
spadgett, line 117: Use the `classNames` utility, which is already a dependency.

https://github.com/JedWatson/classnames

You can probably just inline it in `render` using `classNames`:

```tsx
const labelClassName = classNames('control-label', this.props.hideClassName, { 'co-required': this.props.required });
```

frontend/public/components/utils/storage-class-dropdown.tsx
```diff
@@ -105,6 +105,18 @@ export class StorageClassDropdownInner extends React.Component<
     }
   }
 
+  getLabelClassName = () => {
+    const classNames: string[] = [];
+    this.props.hideClassName
```
spadgett, line 110: Not new, but this is a confusing prop name. What is this for?

spadgett (changes requested):

frontend/public/components/utils/storage-class-dropdown.tsx
```diff
@@ -151,14 +163,7 @@ export class StorageClassDropdownInner extends React.Component<
       <>
         {loaded && itemsAvailableToShow && (
           <div>
-            <label
-              className={
-                this.props.hideClassName
```
cloudbehl, line 156: If hideClassName is a string that is used to pass classes. Why not pass just pass co-required in hideClassName prop?

Over here[1] just add the hideClassName prop and pass co-required and it should work.
Example:
```
                <StorageClassDropdown
                  onChange={handleStorageClass}
                  filter={onlyPvcSCs}
                  id="restore-storage-class"
                  hideClassName="co-required"
                  required
                />
```

[1]. https://github.com/openshift/console/pull/8055/files#diff-34a7d40d57e261c59626bc2fc32c3a2c0450366c6e0101651a49aac7591e7e3dR183

vbnrh: name for the prop does not make sense though.... it is being used in select-capacity steps in the wizard but does not point to any css. Still not sure if thats what it is intended to do....

frontend/public/components/utils/storage-class-dropdown.tsx
vbnrh: i've changed to use this

frontend/public/components/utils/storage-class-dropdown.tsx
vbnrh: as the intention of the prop is unclear and the PR also aims to fix the required props for the dropdown. i am keeping the current implementation instead of using hideClassName to add the css.

frontend/public/components/utils/storage-class-dropdown.tsx
```diff
@@ -241,6 +246,7 @@ export type StorageClassDropdownInnerProps = {
   describedBy: string;
   defaultClass: string;
   required?: boolean;
+  hideLabel?: boolean;
```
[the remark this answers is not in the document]
vbnrh: removed.

frontend/public/components/utils/storage-class-dropdown.tsx
vbnrh: it seems thats the prop is being used in places to add class names to the label. i am still keeping the changes for the required prop as it is well recognized .

spadgett (approved):
/lgtm

PR #7408, opened 2020-12-03
Query Browser: Redesign tooltips and add stacked graph option

kyoto:
Addresses https://issues.redhat.com/browse/MON-787 and https://issues.redhat.com/browse/MON-1299.

Add a "Stacked" checkbox option to the Query Browser. Enabled for the
Metrics page of both the Admin and Dev Consoles.

Change the query browser graph tooltips to display all data series in
the graph at a single timestamp (instead of showing the data for just a
single data point).

If `formatLegendLabel` is provided, use that to generate the tooltip
string for the series names so that the names in the tooltip match those
in the graph legend. Otherwise generate a name using all of the data
series labels.

Removes the need to patch the query into Redux, so that code can be
removed from `graph.tsx`.

Fixes a bug with the tooltips where the vertical position of the tooltip
sometimes did not align to the corresponding data series line.

### Admin Console
screenshot-2

### Dev Console
screenshot-3
screenshot-1

FYI @cshinn

PR #8056, opened 2021-02-04
Bug 1922911: Query Browser: Fix graph crash after hiding and then showing a series

kyoto:
When there were multiple queries, hiding all data series and then
showing one again could result in the graph component crashing.

Bug was introduced by https://github.com/openshift/console/pull/7408


spadgett (approved):
/lgtm

PR #8058, opened 2021-02-04
Bug 1925200: show warning when alert type cannot be recognized

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5477

**Analysis / Root cause:**
The severity label can be anything (ex: bad-label) and the backend doesn't stop us from creating PrometheusRule with such a label. Currently, the logic for showing Monitoring alert decorator on topology doesn't handle that. Also, in the Monitoring Overview tab, in the topology sidebar, such unidentified labels are treated as `Info` alerts.
bad-label

**Solution Description:**
We currently support labels `critical` , `warning` , `info` , `none` , anything apart from these is shown as `warning`. 

**Screen shot:** 
bad-alert-1

Updated existing tests

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/alerts/AlertSeverityIcon.tsx
+++ b/frontend/packages/console-shared/src/components/alerts/AlertSeverityIcon.tsx
@@ -22,6 +22,7 @@ const AlertSeverityIcon: React.FC<AlertSeverityIconProps> = ({
         />
       );
     case AlertSeverity.Warning:
+    default:
       return (
         <ExclamationTriangleIcon
           style={{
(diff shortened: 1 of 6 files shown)


kyoto:
This doesn't seem right to me. I think the `severity` labels should always be lower case in the alert definition.

debsmita1:
As severity label can be anything

> This doesn't seem right to me. I think the `severity` labels should always be lower case in the alert definition.

@kyoto I have updated the logic for displaying monitoring alert decorator on topology node. Now it will show `warning` alert for labels that we do not support

PR #8059, opened 2021-02-04
Bug 1925163: Add info about dev catalog to boot source column

rawagner:
Screenshot from 2021-02-04 15-07-04

@matthewcarleton


matthewcarleton:
you're a wizard! LGTM!

PR #8061, opened 2021-02-04
Migrate obcTest.scenario.ts

rexagod:
Migrated `obcTest.scenario.ts` to use the Cypress API.

Fixes: https://issues.redhat.com/browse/RHSTOR-1558


frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+    cy.clickNavLink(['Storage', 'Object Bucket Claims'], { timeout: 2 * MINUTE });
+    projectNameSpace.selectOrCreateProject(this.namespace);
+    cy.clickNavLink(['Storage', 'Object Bucket Claims']);
+    cy.get('[data-test-id="namespace-bar-dropdown"] div div button').click();
+    cy.contains(this.namespace);
+    cy.contains('Create Object Bucket Claim', { timeout: 6 * SECOND })
+      .should('be.visible')
+      .click();
+    cy.get('#obc-name').type(this.name);
+    cy.get('div.co-storage-class-dropdown div button')
```
rexagod, line 52: Requires the OCS operator and a storage instance (to provide noobaa) installed.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
+import {
```
bipuladh, line 1: Lets rename this file too `obc-test.spec.ts`.
Scenarios are used in protractor tests.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+  OBC_RESOURCE_PATH,
+  OBC_RESOURCE_PATH,
+  OBC_STORAGE_CLASS,
+  OBC_STORAGE_CLASS_EXACT,
+  SECOND,
+} from '../../../noobaa-storage-plugin/integration-tests/utils/consts';
+import {
+  testDeployment,
+  deployment,
+} from '../../../noobaa-storage-plugin/integration-tests/mocks/deploymentData';
```
bipuladh, line 14: Move these to `ceph-storage-plugin/integration-tests-cypress/mocks`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
+import {
+  BOUND,
+  MINUTE,
+  OBC_NAME,
+  OBC_RESOURCE_PATH,
+  OBC_RESOURCE_PATH,
+  OBC_STORAGE_CLASS,
+  OBC_STORAGE_CLASS_EXACT,
+  SECOND,
+} from '../../../noobaa-storage-plugin/integration-tests/utils/consts';
```
bipuladh, line 10: Move these to `ceph-storage-plugin/integration-tests-cypress/utils/consts.ts`.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+        .contains(this.bucketclass)
+        .click();
+    }
+    cy.contains('button', 'Create').click();
+  }
+
+  verifySecretDataIsAvailable() {
+    cy.contains('Reveal Values').should(($el) => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      expect($el).to.exist;
```
bipuladh, line 69:
```suggestion
      expect($el).to.exist;
```
I don't think this would work. This is a chai assertion. I don't think we support chai in console.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+
+const MASK = '•••••';
+const NOOBAA_LABEL = 'app=noobaa';
+const NO_ANNOTATIONS = '0 annotations';
+
+const appHost = `${process.env.BRIDGE_BASE_ADDRESS || 'http://localhost:9000'}${(
+  process.env.BRIDGE_BASE_PATH || '/'
+).replace(/\/$/, '')}`;
+
+class CreateOBCHandler {
```
bipuladh, line 26: Create a view file and move this class over there. We don't need it in the test specs.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/noobaa.obcTest.scenario.spec.ts
rexagod: @bipuladh Cypress supports Chai assertions internally (https://docs.cypress.io/guides/references/assertions.html#Chai). Also as you can see below, the tests pass locally.

image

bipuladh: Although Cypress supports it, https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console/8061/pull-ci-openshift-console-master-frontend/1358690371428683776#1:build-log.txt%3A1555
we have no typescript support for it in our console, which is the reason it's failing `frontend` tests.
Please use jest assertions.

rexagod:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console/8061/pull-ci-openshift-console-master-ceph-storage-plugin/1358845820408958976#1:build-log.txt%3A720

5 of 7 tests failing, including this one, possibly due to the fact that `kubeadmin` is already logged in before `cy.login` is called, although I'm unclear as to why this happens in only 5/7 tests?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,144 @@
...
+  attachToDeployment(deploymentName: string) {
+    cy.clickNavLink(['Storage', 'Object Bucket Claims']);
+    cy.get('tr')
+      .contains(this.name)
+      .should('be.visible')
+      .byLegacyTestID('kebab-button')
+      .first()
+      .click();
+    cy.contains(ATTACH_TO_DEPLOYMENT).click();
+    cy.get('#dropdown-selectbox').click();
```
bipuladh, line 110: Use `data-test`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,144 @@
...
+    cy.byTestSelector('details-item-value__Created at')
+      .scrollIntoView()
+      .should('exist')
+      .invoke('text')
+      .should('eq', 'a few seconds ago');
+  }
+
+  verifyOBPage(obName: string) {
+    cy.contains(obName).click();
+    cy.get('.co-m-loader').should('not.exist');
```
bipuladh, line 98: Use `data-test`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    cy.byLegacyTestID('resource-title')
+      .invoke('text')
+      .should('eq', OBC_NAME);
+    cy.get('.co-resource-item__resource-status', { timeout: 1 * MINUTE })
+      .should('be.visible')
+      .invoke('text')
+      .should('eq', BOUND);
+  });
+
+  it('Test if owner and creation date are shown correctly', () => {
```
bipuladh, line 53: Does this really need to be a different `it` block? Every `it` block should be able to execute independently.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    obcHandler.getOwnerAndDate();
+  });
+
+  it('Test if secret data is available', () => {
+    obcHandler.verifySecretDataIsAvailable();
+  });
+
+  it('Test if secret data is masked', () => {
+    cy.get('.co-copy-to-clipboard__text').should(($el) => {
```
bipuladh, line 62: Use `data-test`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      secrets.forEach((arrEl) => {
+        expect(arrEl).to.equal(MASK);
+      });
+      expect(secrets.length).to.equal(childCount);
+    });
+  });
+
+  it('Test if secret data can be revealed', () => {
+    obcHandler.revealHiddenValues();
+    cy.get('dl.secret-data > dt').as('headings');
```
bipuladh, line 76: Use `data-test`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    cy.get('pre.co-copy-to-clipboard__text').then(($el) => {
```
bipuladh, line 77: Use `data-test`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      cy.get('@headings').should(($h) => {
```
bipuladh, line 78: use `data-test`

bipuladh (changes requested):
Need to do more in-depth review.

rexagod:
@bipuladh Should I squash these `fixup!` commits now?

frontend/packages/noobaa-storage-plugin/integration-tests/utils/consts.ts
```diff
@@ -2,6 +2,8 @@ export const OBC_NAME = 'test-obc';
 export const BOUND = 'Bound';
 export const ATTACH_TO_DEPLOYMENT = 'Attach to Deployment';
 export const OBC_RESOURCE_PATH = 'objectbucket.io~v1alpha1~ObjectBucketClaim';
+export const MINUTE = 60 * 1000;
+export const SECOND = 1000;
```
GowthamShanmugam, line 6: export const SECOND = 1000;
export const MINUTE = SECOND * 60;

rexagod: I did it the other way around as it guarantees that both of them are defined under strict mode conditions. For eg., after a rebase, the order could change which would lead `MINUTE` to be undefined. WDYT?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: I tried to stick as close to the pattern defined in the `obcTest.scenario` file (the one we migrated from) as possible, for better readability between the two in the future. It defines a similar `it` block as you can see below.

https://github.com/openshift/console/blob/8524169bea23554b6e54c669dea9f6d55ec1fa7e/frontend/packages/noobaa-storage-plugin/integration-tests/tests/obcTest.scenario.ts#L36

rexagod: Also, just to add to this, this `it` block should be able to execute independently, as it checks for information that's independent  of the other `it` blocks.

bipuladh: Doesn't it depend upon https://github.com/openshift/console/pull/8061/files#diff-bc97b04fc2ffc9319572ed03433dc859d1ffacac875f13eb19c0df1874650ffaR42 
Which is getting executed in the above test block.
The goal of this migration is to write tests in Cypressy manner and not only port code. All our tests should adhere to Cypress best practices. PTAL https://docs.cypress.io/guides/references/best-practices.html

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      .should('be.visible')
+      .invoke('text')
+      .should('eq', BOUND);
+  });
+
+  it('Test if owner and creation date are shown correctly', () => {
+    obcHandler.getOwnerAndDate();
+  });
+
+  it('Test if secret data is available', () => {
```
bipuladh, line 57: Does this need to be a different `it` block?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    obcHandler.verifySecretDataIsAvailable();
+  });
+
+  it('Test if secret data is masked', () => {
```
bipuladh, line 61: Does this need to be a different `it` block?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
+import { expect } from 'chai';
```
bipuladh, line 1: `it` blocks should be able to execute on its own. There shouldn't be dependency of one `it` block on the other.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+import { CreateOBCHandler } from '../views/obcPage';
+import { testName } from '../../../integration-tests-cypress/support/index';
+
+describe('Test Object Bucket Claim resource', () => {
+  let obcHandler;
+  let obcUrl;
+
+  before(() => {
+    cy.url().then((url) => {
+      if (url.match(/oauth/)) {
```
bipuladh, line 23: I am not really a fan of this approach. As conditional flows reflects that we don't know the actual state of the application, which violates Cypress principles. PTAL https://docs.cypress.io/guides/core-concepts/conditional-testing.html#Definition

rexagod: +1. This is something I did to "workaround" the flakiness of the `cy.login` method, but I'm not a fan as well if it introduces anti-patterns in the tests. I'll replace this with something that simply checks for this case and fails if the condition is not met, instead of incorporating a workaround.

```
    cy.url().should((url) => {
      expect(url).to.match(/oauth/);
    });
    cy.login();
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: I have removed the `cy.visit` part from the test above, it seems it was redundant at this point as the before block already takes us to the required page. I commented out every `it` other than the one we're talking about here and it seems work independently.

image

I'm refactoring the test structure based on the best practice link you quoted. Sending a patch soon.

rexagod:
Please check the latest commit for the aforementioned review changes.

cc @bipuladh @GowthamShanmugam

image

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: Merged this in `Test if secret data is masked` block.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    obcHandler.createBucketClaim();
+    cy.exec(`echo '${JSON.stringify(deployment)}' | kubectl create -n ${testName} -f -`);
+    cy.url().then((url) => {
+      obcUrl = url;
+    });
+  });
+
+  beforeEach(() => {
+    cy.wait(10 * SECOND);
+  });
```
bipuladh, line 39:
```suggestion
```
I am not sure why we need the arbitrary wait. We should rather wait on a condition.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  it('Test if Object Bucket Claim is created sucessfully', () => {
+    cy.byLegacyTestID('resource-title')
+      .should('be.visible')
+      .invoke('text')
+      .should('eq', OBC_NAME);
```
bipuladh, line 45:
```suggestion
      .contains(OBC_NAME);
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('resource-status')
+      .should('be.visible', { timeout: 1 * MINUTE })
+      .invoke('text')
+      .should('eq', BOUND);
```
bipuladh, line 49:
```suggestion
      .contains(BOUND, {timeout: MINUITE});
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  it('Test if owner and creation date are shown correctly', () => {
```
bipuladh, line 52: This doesn't need to be a different it block. 
```suggestion
cy.log('Test if owner and creation date are shown correctly');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestSelector('details-item-value__Owner')
+      .scrollIntoView()
+      .should('exist')
+      .invoke('text')
+      .should('eq', 'No owner');
```
bipuladh, line 57:
```suggestion
      .scrollIntoView()
      .contains('No owner');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestSelector('details-item-value__Created at')
+      .scrollIntoView()
+      .should('exist')
+      .invoke('text')
+      .should('eq', 'a few seconds ago');
```
bipuladh, line 62:
```suggestion
```
Let's not test for Created at. As this is something that can be affected by delays. Chances are it might take a while to get bound.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.url().then((url) => {
+      obcUrl = url;
+    });
+  });
+
+  beforeEach(() => {
+    cy.wait(10 * SECOND);
+  });
+
+  it('Test if Object Bucket Claim is created sucessfully', () => {
```
bipuladh, line 41:
```suggestion
  it('Test if Object Bucket Claim details page is rendered correctly', () => {
     cy.log("Test if OBC is bound");
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      .should('eq', 'No owner');
+    cy.byTestSelector('details-item-value__Created at')
+      .scrollIntoView()
+      .should('exist')
+      .invoke('text')
+      .should('eq', 'a few seconds ago');
+  });
+
+  it('Test if secret data is masked', () => {
+    cy.contains('Reveal Values').should('exist');
```
bipuladh, line 66:
```suggestion
    cy.contains('Reveal Values');
```
`contains` asserts that the value indeed exists.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('co-copy-to-clipboard__text').as('secrets');
+    cy.get('@secrets').should(($el) => {
```
bipuladh, line 68:
```suggestion
    cy.byTestID('co-copy-to-clipboard__text').should(($el) => {
```
I don't think an alias is needed here.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      .invoke('text')
+      .should('eq', 'No owner');
+    cy.byTestSelector('details-item-value__Created at')
+      .scrollIntoView()
+      .should('exist')
+      .invoke('text')
+      .should('eq', 'a few seconds ago');
+  });
+
+  it('Test if secret data is masked', () => {
```
bipuladh, line 65:
```suggestion
cy.log('Test if secret data is masked');
```
We are still in details page.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      const elText = $el.text();
+      const secrets = elText.split(hiddenText).slice(1);
+      secrets.forEach((arrEl: string) => {
+        expect(arrEl).to.equal(MASK);
+      });
+      expect(secrets.length).to.equal(childCount);
+    });
+  });
+
+  it('Test if secret data can be revealed', () => {
```
bipuladh, line 80:
```suggestion
cy.log(Test if secret data can be revealed')
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    obcHandler.revealHiddenValues();
+    cy.byTestID('co-copy-to-clipboard__text').then(($el) => {
+      cy.byTestID('secret-data__dt').should(($h) => {
```
bipuladh, line 83: Do we really need to nest these two?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        expect($el[0].innerText).to.include(NS);
+        expect($h[2].innerText).to.equal('Access Key');
+        expect($el[2].innerText).to.match(new RegExp(ACCESS_KEY));
+        expect($h[3].innerText).to.equal('Secret Key');
+        expect($el[3].innerText).to.match(new RegExp(SECRET_KEY));
+      });
+    });
+  });
+
+  it('Test if secret data can be hidden again', () => {
```
bipuladh, line 94:
```suggestion
cy.log('Test if secret data can be hidden again');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    obcHandler.hideValues();
+    cy.byTestID('co-copy-to-clipboard__text').as('secrets');
+    cy.get('@secrets').should(($el) => {
```
bipuladh, line 97:
```suggestion
    cy.byTestID('co-copy-to-clipboard__text').should(($el) => {
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      const elText = $el.text();
+      const secrets = elText.split(hiddenText).slice(1);
+      secrets.forEach((arrEl) => {
+        expect(arrEl).to.equal(MASK);
+      });
+      expect(secrets.length).to.equal(childCount);
+    });
+  });
+
+  it('Test if labels and annotations are shown correctly', () => {
```
bipuladh, line 109:
```suggestion
cy.log('Test if labels and annotations are shown correctly')
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('label-list').contains(NOOBAA_LABEL);
+    cy.byTestSelector('details-item-value__Annotations').contains(NO_ANNOTATIONS);
+  });
+
+  it('Test if namespace and secret are shown correctly', () => {
```
bipuladh, line 114:
```suggestion
  cy.log('Test if namespace and secret are shown correctly');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    obcHandler.assertNamespaceExists();
+    cy.get('[title="Secret"]')
```
bipuladh, line 116: Use data-test

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      .siblings()
```
bipuladh, line 117: Why do we need to access it via siblings can't we add data-test to it and directly access it?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      .contains(OBC_NAME);
+  });
+
+  it('Test if status and storage class are shown correctly', () => {
```
bipuladh, line 121:
```suggestion
cy.log(Test if status and storage class are shown correctly');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('status-text').contains(BOUND);
+    cy.byLegacyTestID('openshift-storage.noobaa.io').contains(OBC_STORAGE_CLASS_EXACT);
+  });
+
+  it('Test if Object Bucket is created', () => {
```
bipuladh, line 126:
```suggestion
 cy.log('Test if Object Bucket is created');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    const obName = `obc-${testName}-${OBC_NAME}`;
+    cy.contains(obName).click();
```
bipuladh, line 128:
```suggestion
  cy.byTestID('obc-link').click();
```
Let's a data-test to OBC name. I don't want our test to be dependent upon naming conventions.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('co-m-loader').should('not.exist');
```
bipuladh, line 129:
```suggestion
    detailsPage.isLoaded();
```
Present in `integration-tests-cypress/views/`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byLegacyTestID('resource-title').as('title');
+    cy.get('@title')
+      .should('be.visible')
+      .then(($el) => {
+        expect($el.text()).to.equal(obName);
+      });
```
bipuladh, line 135:
```suggestion
    cy.byLegacyTestID('resource-title').should('be.visible');
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.byTestID('resource-status').as('status');
+    cy.get('@status')
+      .should('be.visible')
+      .then(($el) => {
+        expect($el.text()).to.equal(BOUND);
+      });
+  });
```
bipuladh, line 142:
```suggestion
    cy.byTestID('resource-status').contains(BOUND);
```
No need to use aliases.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      .click();
+    cy.contains(ATTACH_TO_DEPLOYMENT).click();
+    cy.byTestID('dropdown-selectbox').click();
+    cy.contains(testDeployment).click();
+    cy.contains('button', 'Attach').click();
+    obcHandler.deploymentReady(testDeployment);
+  });
+
+  after(() => {
+    cy.visit(obcUrl);
```
bipuladh, line 160:
```suggestion
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  it('Test attachment to a Deployment', () => {
+    cy.clickNavLink(['Storage', 'Object Bucket Claims']);
+    cy.get('tr')
+      .contains(OBC_NAME)
+      .should('be.visible')
+      .byLegacyTestID('kebab-button')
+      .first()
+      .click();
+    cy.contains(ATTACH_TO_DEPLOYMENT).click();
```
bipuladh, line 152:
```suggestion
   listPage.rows.shouldBeLoaded();
   listPage.rows.clickKebabAction(OBC_NAME, ATTACH_TO_DEPLOYMENT);
```
if `ATTACH_TO_DEPLOYMENT` is only used here then directly use it here without declaring it as a constant. 
Import `listPage` from `integration-tests-cypress/views/list-page`

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      });
+    cy.byTestID('resource-status').as('status');
+    cy.get('@status')
+      .should('be.visible')
+      .then(($el) => {
+        expect($el.text()).to.equal(BOUND);
+      });
+  });
+
+  it('Test attachment to a Deployment', () => {
```
bipuladh, line 144: Let's localize creation and deletion of the deployment to this `it` block as it's not a per-requisite for the other one.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,89 @@
...
+export class CreateOBCHandler {
+  name: string;
+
+  namespace: string;
+
+  storageclass: string;
+
+  bucketclass: string;
+
+  constructor(name: string, namespace: string, storageclass: string, bucketclass: string) {
```
bipuladh, line 14:
```suggestion
  constructor(name: string, namespace: string, storageclass: string) {
```
Since we are not using it anywhere.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.name = name;
+    this.namespace = namespace;
+    this.storageclass = storageclass;
+    this.bucketclass = bucketclass;
```
bipuladh, line 18:
```suggestion
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,89 @@
...
+      .should('be.visible')
+      .click();
+    cy.contains('openshift-storage.noobaa.io').click();
+    if (this.bucketclass) {
+      cy.byTestID('bc-dropdown')
+        .should('be.visible')
+        .click()
+        .contains(this.bucketclass)
+        .click();
+    }
```
bipuladh, line 41:
```suggestion
```

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+      },
+    );
+    cy.byLegacyTestID('resource-title')
+      .should('be.visible')
+      .contains(deploymentName);
+  }
+
+  deleteBucketClaim() {
+    cy.byTestID('co-m-loader').should('exist');
```
bipuladh, line 73:
```suggestion
```
Loading states are tricky. Chances are the state changes so quickly that we miss it. This can be a source of flake. I would rather prefer an opposite validation where we check that it doesn't exist.

frontend/public/components/utils/copy-to-clipboard.tsx
```diff
@@ -19,7 +19,12 @@ export const CopyToClipboard: React.FC<CopyToClipboardProps> = React.memo((props
 
   return (
     <div className="co-copy-to-clipboard">
-      <pre className="co-pre-wrap co-copy-to-clipboard__text">{visibleValue}</pre>
+      <pre
+        className="co-pre-wrap co-copy-to-clipboard__text"
+        data-test="co-copy-to-clipboard__text"
```
bipuladh, line 24:
```suggestion
        data-test="copy-to-clipboard"
```
Data test doesn't need to adhere BEM naming convention.

frontend/packages/noobaa-storage-plugin/src/components/object-bucket-claim-page/secret.tsx
```diff
@@ -63,7 +63,11 @@ export const GetSecret: React.FC<GetSecretProps> = ({ obj }) => {
   const dl = secretValues.length
     ? secretValues.reduce((acc, datum) => {
         const { field, value } = datum;
-        acc.push(<dt key={`${field}-k`}>{field}</dt>);
+        acc.push(
+          <dt key={`${field}-k`} data-test="secret-data__dt">
```
bipuladh, line 67:
```suggestion
          <dt key={`${field}-k`} data-test="secret-data">
```
data-test doesn't need to adhere to BEM naming convention.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  SECOND,
+  SECRET_KEY,
+} from '../utils/consts';
+import { testDeployment, deployment } from '../mocks/deploymentData';
+import { CreateOBCHandler } from '../views/obcPage';
+import { testName } from '../../../integration-tests-cypress/support/index';
+
+describe('Test Object Bucket Claim resource', () => {
+  let obcHandler;
+  let obcUrl;
```
bipuladh, line 23:
```suggestion
```
Let's divide this test into only two `it` block. 
The first localized to the details page for the OBC we created. And the second one to the list page(where we only attach to the deployment for now, we will add more tests in the future if required).

frontend/packages/ceph-storage-plugin/integration-tests-cypress/mocks/deploymentData.ts
```diff
@@ -0,0 +1,37 @@
...
+  metadata: {
+    name: testDeployment,
+  },
+  spec: {
+    selector: {
+      matchLabels: {
+        app: 'hello-openshift',
+      },
+    },
+    replicas: 3,
```
bipuladh, line 15:
```suggestion
    replicas: 1,
```
More replicas take more time to get ready.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    cy.get('tr')
+      .contains(OBC_NAME)
+      .should('be.visible')
+      .byLegacyTestID('kebab-button')
+      .first()
+      .click();
+    cy.contains(ATTACH_TO_DEPLOYMENT).click();
+    cy.byTestID('dropdown-selectbox').click();
+    cy.contains(testDeployment).click();
+    cy.contains('button', 'Attach').click();
```
bipuladh, line 155: Could we use `modal.submit()` ?

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,89 @@
...
+  deploymentReady(deploymentName: string) {
+    cy.exec(`kubectl get deployment ${deploymentName} -o json -n ${this.namespace}`).then(
+      ($info) => {
+        const info = JSON.parse($info.stdout.toString());
+        const { replicas, readyReplicas } = info.status;
+        if (replicas < readyReplicas) {
+          cy.wait(2 * MINUTE);
+        }
+      },
+    );
```
bipuladh, line 66:
```suggestion
```
This is not asserting that the deployment is indeed in a ready state. Instead let's wait for it to get to ready state via UI.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: Can you please rephrase the "Let's a data-test to OBC name." part?

bipuladh: edited. apologies.

frontend/public/components/utils/resource-link.tsx
```diff
@@ -99,11 +100,16 @@ export const ResourceLink = connectToModel(
...
+            data-test={dataTest}
           >
             {value}
           </Link>
         ) : (
-          <span className="co-resource-item__resource-name" data-test-id={value}>
+          <span
+            className="co-resource-item__resource-name"
+            data-test-id={value}
+            data-test={dataTest}
```
rexagod, line 111: I've added a `data-test` attribute here so that the tests that depend on `data-test-id` don't fail.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: I'm not sure about this but it seems we need to capture the output when working with `byTest___` methods inside a `get` to get the correct type definition, if we need to operate on the passed down element inside the `thenable`. If this is in fact a bug in `Cypress`, it's probably stemming from the `add` wrapper we use to define new commands.

image
image

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: Removing this fails the `after` block as the page that deletes the OBC is not visited. This is because the second `it` block (not sure about the first one) navigates to a different page during its execution. Let me look a bit further into this and get back to you.

rexagod: Okay so I looked it up and the lines below navigate the respective tests to a different URL, which is why we need to ensure that we are back at the OBC page when we enter the `after` block so as to delete the Bucket Claim.

First test: https://github.com/openshift/console/pull/8061/files#diff-bc97b04fc2ffc9319572ed03433dc859d1ffacac875f13eb19c0df1874650ffaR102

Second test: https://github.com/openshift/console/pull/8061/files#diff-bc97b04fc2ffc9319572ed03433dc859d1ffacac875f13eb19c0df1874650ffaR114

rexagod:
Tests seem to pass on my local where it fails on the CI. Looking into this.

#### Local
image

#### CI
image

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
bipuladh: Please address this comment as well...

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  let obcHandler;
+  let obcUrl;
+
+  before(() => {
+    cy.login();
+    cy.visit('/');
+    cy.install();
+    obcHandler = new CreateOBCHandler(OBC_NAME, testName, OBC_STORAGE_CLASS);
+    obcHandler.createBucketClaim();
+    cy.exec(`echo '${JSON.stringify(deployment)}' | kubectl create -n ${testName} -f -`);
```
bipuladh, line 32: Move this to the second it block.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    listPage.rows.shouldBeLoaded();
+    listPage.rows.clickKebabAction(OBC_NAME, 'Attach to Deployment');
+    cy.byTestID('dropdown-selectbox').click();
+    cy.contains(testDeployment).click();
+    modal.submit();
+    obcHandler.deploymentReady(testDeployment);
+    cy.exec(`echo '${JSON.stringify(deployment)}' | kubectl delete -n ${testName} -f -`);
+  });
+
+  after(() => {
```
bipuladh, line 119: nit: move the after block right below the before block.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    cy.visit('/');
+    cy.install();
+    obcHandler = new CreateOBCHandler(OBC_NAME, testName, OBC_STORAGE_CLASS);
+    obcHandler.createBucketClaim();
+    cy.exec(`echo '${JSON.stringify(deployment)}' | kubectl create -n ${testName} -f -`);
+    cy.url().then((url) => {
+      obcUrl = url;
+    });
+  });
+
```
bipuladh, line 37: Add an after each block and call `checkErrors`.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/obc-test.spec.ts
rexagod: Do you mean the first one (as it depends on this code block)? Also, pointing out that after we move this to the first it block, the second it block will be dependant on the first one.

rexagod: My bad I thought you were referring to lines 29-32.

rexagod:
CI throwing same error: https://github.com/openshift/console/pull/8061#issuecomment-781811282

rexagod:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/8061/pull-ci-openshift-console-master-ceph-storage-plugin/1363795881098219520/artifacts/ceph-storage-plugin/test/artifacts/gui_test_screenshots/cypress/videos/ocs-presistent-dashboard.spec.ts.mp4

rexagod:
image

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_console/8061/pull-ci-openshift-console-master-ceph-storage-plugin/1363825214391062528/artifacts/ceph-storage-plugin/test/artifacts/gui_test_screenshots/cypress/videos/obc-test.spec.ts.mp4

rexagod:
Checked this once again locally. All tests pass.

image

rexagod:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console/8061/pull-ci-openshift-console-master-e2e-gcp-console/1364181118386966528#1:build-log.txt%3A1530

rexagod:
Finally!

frontend/packages/ceph-storage-plugin/integration-tests-cypress/views/obcPage.ts
```diff
@@ -0,0 +1,74 @@
...
+import { modal } from '../../../integration-tests-cypress/views/modal';
+
+export class CreateOBCHandler {
+  name: string;
+
+  namespace: string;
+
+  storageclass: string;
+
+  bucketclass: string;
```
bipuladh, line 12:
```suggestion
```

bipuladh (changes requested):

bipuladh:
/assign @rawagner 
Added some data-test to generic components. PTAL.

bipuladh:
looks good, please squash the commits.

bipuladh (approved):
/lgtm

PR #8062, opened 2021-02-04
Bug 1925516: Pipeline metrics tooltips

andrewballantyne:
For https://issues.redhat.com/browse/ODC-5311

A hacky implementation to try and make tooltips expand upwards from a chart. They have PF examples ([css-overflow](https://www.patternfly.org/v4/charts/tooltip-chart#css-overflow)) of this but they neglect to mention that it doesn't stop the tooltip from going off the side of the browser and get chopped off. Which is very easily done on the Pipeline Metrics page.

The solution is adding height to the chart, a negative margin to bring it back & constraining the tooltip so it doesn't go outside of the side bounds of the chart.

pipeline-metrics-tooltips

### Browser Support

Works on...

- Chrome, v88.0.4324.96
- Firefox, v78.7.0esr
- Safari, v14.0


bgliwa01 (approved):
lgtm!

karthikjeeyar:
Verified it works fine. Will add lgtm once the conflict is resolved.

PR #8063, opened 2021-02-04
Use dompurify instead of sanitize-html

jschuler:
To reduce the bundle size a bit.
I ran the `yarn analyze` target and the reported treemap stat size changed as follows:
(Before | After)
All: 44.71MB | 44.14MB
Chunk: 599.06KB | 221.01KB

The impacted file is consumed by the quickstarts.


frontend/public/components/markdown-view.tsx
```diff
@@ -18,37 +18,40 @@ const markdownConvert = (markdown, extensions?: string[]) => {
...
     ],
-    allowedAttributes: {
-      a: ['href', 'target', 'rel', 'data-*'],
-      button: ['class', 'data-*'],
-    },
-    allowedSchemes: ['http', 'https', 'mailto'],
-    transformTags: {
-      a: sanitizeHtml.simpleTransform('a', { rel: 'noopener noreferrer' }, true),
-    },
+    ALLOWED_ATTR: ["href", "target", "rel", "class"],
```
spadgett, line 53: Won't this restrict the attributes on all elements, not just `a` elements? I'm wondering if that can break some markup.

spadgett (commented):
Thanks, cc @christianvogt @andrewballantyne

frontend/public/components/markdown-view.tsx
jschuler: Hi @spadgett, so from my understanding the previous sanitize-html actually restricted ALL attributes except for 
```
a: ['href', 'target', 'rel', 'data-*'],
button: ['class', 'data-*'],
```
So for example, even though the `code` tag is allowed, the previous config would strip out the attributes.
Markdown:
```
```javascript
alert(s);
```
```

Result HTML before (language class is lost):
```
<code>alert(s);</code>
```

After this change:
```
<code class="javascript language-javascript">alert(s);</code>
```

So technically this is now a little more permissive across all the allowed tags that have been defined.
I didn't see a way to have per-tag rules like sanitize-html had.

Comment (jschuler) [reply frontend/public/components/markdown-view.tsx]: Looking at which attributes are possible on which elements (https://www.w3schools.com/tags/ref_attributes.asp):
```
href: <a>, <area>, <base>, <link>
target: <a>, <area>, <base>, <form>
rel: <a>, <area>, <form>, <link>
class: all
```
Of these elements the only applicable one based on the currently allowed tags is `<a>`.

So really the only change I can see is that `class` is now allowed on all the tags :)

File: frontend/public/components/markdown-view.tsx [hunk 2]
```diff
@@ -18,8 +18,17 @@ const markdownConvert = (markdown, extensions?: string[]) => {
     extensions,
   }).makeHtml(markdown);
 
-  return sanitizeHtml(unsafeHtml, {
-    allowedTags: [
+  // add hook to transform anchor tags
+  DOMPurify.addHook('beforeSanitizeElements', function(node) {
+    // nodeType 1 = element type
+    if (node.nodeType === 1 && node.nodeName.toLowerCase() === 'a') {
```
andrewballantyne, line 24: Best avoid magic constants...
```suggestion
    if (node.nodeType === Node.ELEMENT_NODE && node.nodeName.toLowerCase() === 'a') {
```
Although, not sure if we need to check the node type 🤔 

Also, in the upstream example they provided on the DOMPurify readme, it appears the way to check a `tagName` would be via the second param? eg: [for "script" nodes](https://github.com/cure53/DOMPurify/blob/main/demos/hooks-mentaljs-demo.html#L33)

PR #8069, opened 2021-02-05
Bug 1920677: fixes missing i18n for create a project

invincibleJai:
Fixes:
https://issues.redhat.com/browse/ODC-5533
https://issues.redhat.com/browse/OCPBUGSM-23788

refactored  CreateProjectListPage Component and all caller component  in order to have whole sentence together.

image

Analysis / Root cause:
missing i18n key


jeff-phillips-18 (approved):

jeff-phillips-18 (approved):

jeff-phillips-18 (approved):

frontend/packages/dev-console/locales/en/devconsole.json
```diff
@@ -362,12 +362,13 @@
   "Role": "Role",
   "Select a role": "Select a role",
   "Reload": "Reload",
-  "<0>{children} or <3>create a Project.</3></0>": "<0>{children} or <3>create a Project.</3></0>",
+  "or": "or",
+  "create a Project.": "create a Project.",
```
spadgett, line 366: We aren't able to break up sentences like this. It's problematic for translation. We should find a way to make this a single message (using the `Trans` component if we need to add a button).

spadgett (changes requested):

frontend/packages/dev-console/locales/en/devconsole.json
invincibleJai: Thanks @spadgett , have updated PTAL

frontend/packages/dev-console/locales/en/devconsole.json
```diff
@@ -362,12 +362,12 @@
   "Role": "Role",
   "Select a role": "Select a role",
   "Reload": "Reload",
-  "<0>{children} or <3>create a Project.</3></0>": "<0>{children} or <3>create a Project.</3></0>",
+  "<0>{{children}} or <3>create a Project.</3></0>": "<0>{{children}} or <3>create a Project.</3></0>",
```
spadgett, line 365: Hi, this is better, but we can't plugin any text as `children` here. Translators often have to move words around in a sentence depending on the language and need the full string as one message. I know it means some rework for `CreateProjectListPage`.

invincibleJai: Hi Sam , i looked into it and based on my understanding it will need refactoring in multiple components (around 10) where `CreateProjectListPage` is called from and problem with that is `create a Project` is button text which has action associated so it'll be duplicity in order to have whole sentence together.

another option is just moving `or` to the caller and have `button` and action in `CreateProjectListPage` but even then sentence will be incomplete.  i.e 

```
<CreateProjectListPage title={t('devconsole~Add')}>
     {t('devconsole~Select a Project to start adding to it or')}
 </CreateProjectListPage>
```

and `CreateProjectListPage` list as 

```
<ProjectListPage {...props} title={title}>
        {{ children }}
        <Button isInline variant="link" onClick={openProjectModal}>
          {t('devconsole~create a Project.')}
        </Button>
 </ProjectListPage>
```

Let me know your thoughts

frontend/packages/dev-console/src/components/projects/CreateProjectListPage.tsx
```diff
@@ -17,24 +15,16 @@ const CreateProjectListPage: React.FC<CreateProjectListPageProps> = ({
...
-    <Trans t={t} ns="devconsole">
-      <ProjectListPage {...props} title={title}>
-        {children} {'or '}
-        <Button isInline variant="link" onClick={openProjectModal}>
-          {'create a Project.'}
-        </Button>
-      </ProjectListPage>
-    </Trans>
+    <ProjectListPage {...props} title={title}>
+      {typeof children === 'function' ? children(openProjectModal) : children}
```
christianvogt, line 26: Please type the props accordingly.

invincibleJai: have updated , PTAL https://github.com/openshift/console/compare/489dec3448aa728bc655a4fdc235ccc56a195e37..7e5dd40d7ba1e5f6bc78b0af5cfb942873ad4f14

frontend/packages/dev-console/src/components/projects/ProjectListPage.tsx
```diff
@@ -4,12 +4,13 @@ import { ProjectsTable } from '@console/internal/components/namespace';
 import { PageHeading } from '@console/internal/components/utils';
 import './ProjectListPage.scss';
 
-export interface ProjectListPageProps {
+export type LazySubTitleRender = (props: () => void) => React.ReactNode;
+export type ProjectListPageProps = {
   title: string;
   listComponent?: React.ComponentType<any>;
-  children?: React.ReactNode;
+  children?: React.ReactNode | LazySubTitleRender;
```
christianvogt, line 11: `ProjectListPage` children should still be `React.ReactNode` 
It's the props of `CreateProjectListPage` that need to change.

frontend/packages/dev-console/src/components/projects/CreateProjectListPage.tsx
```diff
@@ -17,24 +15,18 @@ const CreateProjectListPage: React.FC<CreateProjectListPageProps> = ({
...
-      <ProjectListPage {...props} title={title}>
-        {{ children }} {'or '}
-        <Button isInline variant="link" onClick={openProjectModal}>
-          {'create a Project.'}
-        </Button>
-      </ProjectListPage>
-    </Trans>
+    <ProjectListPage {...props} title={title}>
+      {typeof children === 'function'
+        ? (children as LazySubTitleRender)(openProjectModal)
```
christianvogt, line 27: If your types are correct, you shouldn't need a cast here. I think...

frontend/packages/dev-console/src/components/projects/ProjectListPage.tsx
invincibleJai: yeah correct, did the mistake here. Have updated

karthikjeeyar:
Verified and the changes looks good to me.

PR #8071, opened 2021-02-05
ODC-5487: Temp fix for search-input bug

jcaianirh:
Fixes: https://issues.redhat.com/browse/ODC-5487


andrewballantyne (approved):
/lgtm

Good override for 4.7 GA... we can take a proper upgrade in 4.8 and drop this hack.

PR #8075, opened 2021-02-05
Bug 1924585: Fix translation for edit annotation

rebeccaalpert:
The Chinese translation was incorrect. I updated it with the one provided by QE.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1924585.


spadgett (approved):
/lgtm

PR #8076, opened 2021-02-05
Fix issues related to loading plugins via SERVER_FLAGS.consolePlugins

vojtechszocs:
This PR is meant to address issues related to loading [dynamic plugins](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/README.md) via the standard Console plugin mechanism.

- Plugin name and version must be declared in `package.json` explicitly via `consolePlugin` object. Plugin name should be the same as `metadata.name` in the corresponding `ConsolePlugin` resource.
- Plugin webpack config's `output.publicPath` option defaults to `/api/plugins/<PluginName>/` Bridge endpoint if not specified. Plugin developers are encouraged to use this default endpoint.
- Plugin development process involves running Bridge locally and instructing it to proxy `/api/plugins/<PluginName>/` requests directly to the local plugin asset server.
- `loadDynamicPlugin` function is now async and resolves/rejects as expected.
- Plugin name and version validation improved and unified.
- Demo plugin scripts: `yarn build` = production build, `yarn build-dev` = development build.

:memo: _Note:_ existing [static plugins](https://github.com/openshift/console/blob/master/frontend/packages/console-plugin-sdk/README.md) are unaffected by this change.

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/README.md
+++ b/frontend/packages/console-dynamic-plugin-sdk/README.md
@@ -9,6 +9,10 @@ released, installed and upgraded independently from each other. To ensure compat
 other plugins, each plugin must declare its dependencies using [semantic version](https://semver.org/)
 ranges.
 
+See the
+[OpenShift Console Dynamic Plugins feature page](https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md)
+for a high level overview of dynamic plugins in relation to OLM operators and cluster administration.
+
 Example project structure:
 
 ```
@@ -22,8 +26,7 @@ dynamic-demo-plugin/
 
 ## `package.json`
 
-`name` and `version` are used to identify the plugin instance at runtime. Additional plugin metadata is
-declared via the `consolePlugin` object.
+Plugin metadata is declared via the `consolePlugin` object.
 
 ```jsonc
 {
@@ -32,7 +35,9 @@ declared via the `consolePlugin` object.
   "private": true,
   // scripts, dependencies, devDependencies, ...
   "consolePlugin": {
-    "displayName": "Dynamic Demo Plugin",
+    "name": "console-demo-plugin",
+    "version": "0.0.0",
+    "displayName": "Console Demo Plugin",
     "description": "Plasma reactors online. Initiating hyper drive.",
     "exposedModules": {
       "barUtils": "./utils/bar"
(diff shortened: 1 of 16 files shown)


frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -131,9 +140,31 @@ Provides asynchronous access to specific modules exposed by the plugin.
...
+```
+
+Your plugin should start loading automatically upon Console application startup. Inspect the value of
+`window.SERVER_FLAGS.consolePlugins` to see the list of plugins which Console loads upon its startup.
+
+## Plugin detection and management
+
+- [Console operator](https://github.com/openshift/console-operator) detects available plugins through
+  `ConsolePlugin` resources on the cluster. It also maintains a cluster-wide list of currently enabled
+  plugins via `spec.plugins` field in its `ConfigMap` object (`console-operator-config`).
```
rawagner, line 159: According to https://github.com/openshift/console/pull/8076/files#diff-76eb2371c0e3f846dd818cb2510ab8f9c3d8315cd192aabd781e290970f6793bR57 Console operator uses `Console` object kind (name `cluster`) to store the enabled plugins, not `ConfigMap`

frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -131,9 +140,31 @@ Provides asynchronous access to specific modules exposed by the plugin.
...
 `utils_bar_ts-chunk.js`: webpack chunk corresponding to the exposed `barUtils` module.
 
+## Plugin development
+
+Run Bridge locally and instruct it to proxy e.g. `/api/plugins/example-plugin` requests directly to your
+local plugin asset server (web server hosting the plugin's generated assets):
+
+```sh
+./bin/bridge -plugins example-plugin=http://localhost:9001/
+```
```
rawagner, line 150: since we have a demo plugin, can you change the exaples to use `console-demo-plugin` instead of `example-plugin` ? 
IMO its just easier to start if we show the whole process on demo plugin.

```suggestion
Run Bridge locally and instruct it to proxy e.g. `/api/plugins/console-demo-plugin` requests directly to your
local plugin asset server (web server hosting the plugin's generated assets):

```sh
./bin/bridge -plugins console-demo-plugin=http://localhost:9001/
```
```

rawagner (changes requested):

frontend/packages/console-dynamic-plugin-sdk/README.md
vojtechszocs: Agreed, will do.

frontend/packages/console-dynamic-plugin-sdk/README.md
vojtechszocs: OK, maybe @spadgett or @jhadvig can clarify here?

(I thought that enabling a plugin manually involved the editing of `ConfigMap` instance named `console-operator-config` which I thought was the place where the list of enabled plugins was maintained.)

spadgett: The console operator config is a separate CRD in the `operator.openshift.io` group as @rawagner mentions. This is where an admin configures the enabled plugins.

The operator does pass the list in a config map to the console backend, but that's not a public API.

See https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

vojtechszocs:
Rebased and updated READMEs according to review comments.

PR #8080, opened 2021-02-07
Fix pipeline types

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-2533

**Analysis / Root cause**: 
Types were placed with code and poorly named.

**Solution Description**: 
Moved the types to dedicated files and unified the overlapping types.


frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
```diff
@@ -1,12 +1,12 @@
 import * as _ from 'lodash';
 import { pipelineParameters, pipelineParametersWithoutDefaults } from './utils-data';
-import { PipelineParam } from '../../../../utils/pipeline-augment';
+import { TektonParam } from '../../../../types';
```
abhinandan13jan, line 3: Is there any specific reason we are using `Tekton` in type name here? I only ask as I don't see it anywhere else. 
Could we make use of something more relatable like `CommonParam` WDYT?

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/utils.ts
```diff
@@ -1,11 +1,9 @@
 import * as _ from 'lodash';
-import { PipelineParam } from '../../../utils/pipeline-augment';
+import { TektonParam } from '../../../types';
```
abhinandan13jan, line 2: similar comment as above

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/utils.spec.ts
andrewballantyne: Because it's generic. I am open to another name... but both Task and Pipeline use the same Parameter type... so I suppose I could called it TektonParam and then create stand alone types for Pipelines and Tasks for "their params" but it seemed kinda needless.

abhinandan13jan: cool!!

abhinandan13jan:
Ran locally and tested basic pipeline related pages. Works fine
/lgtm

PR #8082, opened 2021-02-08
odc-5475-updated setting.json in frontend folder

makambalaji:
**Description**:
This pr includes below changes
 Add the Automation QE related settings to .vscode/settings.json file
 updated .gitignore file under console folder
 Added integration-tests/.OWNERS file under gitops, pipelines and helm plugin folders  
 Removed the indentation settings from .gherkin-lintrc file, will include it in sub folders in next pr's

Jira Id: https://issues.redhat.com/browse/ODC-5475

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,6 @@ cypress-a11y-report.json
 /frontend/**/dist
 /frontend/@types/gql
 /frontend/gui_test_screenshots
-/frontend/packages/integration-tests-cypress/cypress-a11y-report.json
-/frontend/packages/ceph-storage-plugin/integration-tests-cypress/cypress-a11y-report.json
 /frontend/@types
 /frontend/package-lock.json
 /frontend/po-files
--- a/frontend/.vscode/settings.json
+++ b/frontend/.vscode/settings.json
@@ -61,8 +61,10 @@
   },
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  }
-
+  },
+  "cucumberautocomplete.steps": ["./packages/*/integration-tests/support/step-definitions/*/*.ts"],
+  "cucumberautocomplete.strictGherkinCompletion": true,
+  "editor.quickSuggestions": true
   // TODO support prettier + stylelint
   // "prettier.stylelintIntegration": true,
   // "css.validate": false,
--- a/frontend/.gherkin-lintrc
+++ b/frontend/.gherkin-lintrc
@@ -5,7 +5,7 @@
   "no-dupe-scenario-names": ["on", "in-feature"],
   "no-dupe-feature-names": "on",
   "no-partially-commented-tag-lines": "on",
-  "indentation" : ["on", {"Feature": 0, "Background": 4, "Scenario": 4, "Step": 8, "Examples": 8, "example": 12, "given": 8, "when": 8, "then": 8, "and": 8, "but": 8, "feature tag": 0, "scenario tag": 4}],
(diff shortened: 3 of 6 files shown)


frontend/.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.tabSize": 4,
```
rohitkrai03, line 3: Don't think i agree with this change. Usually all across our project we follow `tabSize: 2`.

rohitkrai03 (changes requested):

frontend/.vscode/settings.json
andrewballantyne: 💯  agree... let's not make this change.

makambalaji: Removed the line, will included it in lower tree

spadgett: > Removed the line, will included it in lower tree

We should have uniform tab size settings for the same file types across the repo. What files have a 4 space tab? For JavaScript and TypeScript, we should use prettier.

andrewballantyne (approved):
LGTM. Thanks for the updates.

frontend/.vscode/settings.json
andrewballantyne: This is for Gherkin files `.feature` -- BDD test files that QE creates @spadgett -- Which is why we wanted to avoid the change here.

.gitignore
```diff
@@ -18,6 +18,8 @@ cypress-a11y-report.json
 /frontend/@types/gql
 /frontend/gui_test_screenshots
 /frontend/packages/integration-tests-cypress/cypress-a11y-report.json
+/frontend/packages/dev-console/integration-tests/cypress-a11y-report.json
+/frontend/packages/knative-plugin/integration-tests/cypress-a11y-report.json
```
andrewballantyne, line 22: Ah this change is what stopped my approval. Will need @spadgett to approve likely. There are few with global permissions.

andrewballantyne: Perhaps we change this to just say:

```
/frontend/packages/**/cypress-a11y-report.json
```
I believe that works... we don't want to commit any of these reports I assume.

makambalaji: updated

frontend/.vscode/settings.json
makambalaji: @spadgett - I think approval is required from you, As this pr contains modifications related to console/.git-ignore file

.gitignore
```diff
@@ -18,6 +18,7 @@ cypress-a11y-report.json
 /frontend/@types/gql
 /frontend/gui_test_screenshots
 /frontend/packages/integration-tests-cypress/cypress-a11y-report.json
+/frontend/packages/*/integration-tests/cypress-a11y-report.json
```
spadgett, line 21: I don't think this is needed after #7916, but we should confirm. We can probably remove the other `cypress-a11y-report.json` lines as well.

andrewballantyne: Interesting, perhaps we remove line 20 as well unless it's being handled elsewhere...

makambalaji: Yeah, now I am able to see the cypress-a11y-report.json files are ignored. I can remove 3 lines [20,21,22]

andrewballantyne (approved):

.gitignore
```diff
@@ -17,8 +17,6 @@ cypress-a11y-report.json
 /frontend/**/dist
 /frontend/@types/gql
 /frontend/gui_test_screenshots
-/frontend/packages/integration-tests-cypress/cypress-a11y-report.json
```
rhamilto, line 20: I believe we want to keep these changes.  @TheRealJon, agreed?

rhamilto: Never mind.  I see line 6 now.

rhamilto (approved):

PR #8085, opened 2021-02-08
Add storage.bind.immediate.requested annotation to golden image dvs

yaacov:
When adding a golden image, we want it to get created immediately without waiting for first consumer


frontend/packages/kubevirt-plugin/integration-tests-cypress/support/index.ts
```diff
@@ -34,6 +34,11 @@ Cypress.Commands.add('createDataVolume', (name: string, namespace: string) => {
     cy.fixture('data-volume').then((dv) => {
       dv.metadata.name = name;
       dv.metadata.namespace = namespace;
+      dv.metadata = dv?.metadata || {};
+      dv.metadata.annotations = {
+        ...(dv?.metadata?.annotations || {}),
+        'cdi.kubevirt.io/storage.bind.immediate.requested': 'true',
```
rawagner, line 40: `dv` fixture already has the annotation https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/integration-tests-cypress/fixtures/data-volume.json#L8

PR #8087, opened 2021-02-08
adding global config extension types to dynamic-plugin-sdk

glekner:
- Improve code reference handling for Console dynamic plugins
- Fix type related errors
- Improve unit test for applyCodeRefSymbol
- Avoid require()'ing modules when generating @console/active-plugins
- adding global config extension types to dynamic-plugin-sdk

Depends on #7919

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/extensions/cluster-settings.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/extensions/cluster-settings.ts
@@ -1,0 +1,29 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+
+export namespace ExtensionProperties {
+  export type ClusterGlobalConfig = {
+    /** Unique identifier for the cluster config resource instance. */
+    id: string;
+    /** The name of the cluster config resource instance. */
+    name: string;
+    /** The model which refers to a cluster config resource. */
+    model: {
+      group: string;
+      version: string;
+      kind: string;
+    };
+    /** The namespace of the cluster config resource instance. */
+    namespace: string;
+  };
+}
+
+// Extension types
+
+export type ClusterGlobalConfig = Extension<ExtensionProperties.ClusterGlobalConfig> & {
+  type: 'console.global-config';
+};
+
+// Type guards
+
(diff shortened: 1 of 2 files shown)


frontend/packages/console-dynamic-plugin-sdk/src/extensions/global-configs.ts
```diff
@@ -0,0 +1,29 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+
+export namespace ExtensionProperties {
+  export type GlobalConfig = {
```
vojtechszocs, line 4: This file could be named `cluster-settings.ts` and its extension & properties types could be named `ClusterGlobalConfig` to better reflect the fact that it's meant to extend core components under `public/components/cluster-settings`.

I also suggest grouping related bits of data together, while also reducing the amount of data to a necessary minimum. For example, referring to a k8s resource strictly via `group/version/kind` instead of using the wrapper `K8sKind` type.

```ts
/** The model which refers to a cluster config resource. */
model: {
  group: string;
  version: string;
  kind: string;
}
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/global-configs.ts
```diff
@@ -0,0 +1,29 @@
...
+    /** Unique identifier for this item. */
+    id: string;
+    /** Kind of this item. */
```
vojtechszocs, line 7: `The name of the cluster config resource instance.`

frontend/packages/console-dynamic-plugin-sdk/src/extensions/global-configs.ts
```diff
@@ -0,0 +1,29 @@
...
+    name: string;
+    /** Model for this item. */
+    model: {
+      group: string;
+      version: string;
+      kind: string;
+    };
+    /** Namespace of this item. */
```
vojtechszocs, line 15: `The namespace of the cluster config resource instance.`

glekner:
rebased and updated @vojtechszocs

christianvogt:
/hold

stop tests until rebase

glekner:
rebased @christianvogt

PR #8088, opened 2021-02-08
converting global config extension to dynamic-plugin-sdk

glekner:
depends on #7919

### DIFF
--- a/frontend/packages/knative-plugin/console-extensions.json
+++ b/frontend/packages/knative-plugin/console-extensions.json
@@ -64,5 +64,21 @@
     "flags": {
       "required": ["FLAG_CAMEL_KAMELETS"]
     }
+  },
+  {
+    "type": "console.global-config",
+    "properties": {
+      "id": "knative-serving",
+      "name": "knative-serving",
+      "model": {
+        "group": "operator.knative.dev",
+        "version": "v1alpha1",
+        "kind": "KnativeServing"
+      },
+      "namespace": "knative-serving"
+    },
+    "flags": {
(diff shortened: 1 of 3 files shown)


frontend/packages/console-dynamic-plugin-sdk/src/extensions/cluster-settings.ts
```diff
@@ -0,0 +1,29 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+
+export namespace ExtensionProperties {
+  export type ClusterGlobalConfig = {
+    /** Unique identifier for the cluster config resource instance. */
+    uid: string;
```
christianvogt, line 6: While we currently use `uid` in the static extension, we should update to `id` to be consistent with other extensions where `id` denotes a unique identifier.

glekner:
please review @christianvogt

rawagner:
/hold
compilation error

```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/cluster-settings/global-config.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/cluster-settings/global-config.tsx(105,22):
TS2339: Property 'id' does not exist on type 'ClusterGlobalConfig | { model: any; uid: any; name: any; namespace: any; kind: any; }'.
  Property 'id' does not exist on type '{ model: any; uid: any; name: any; namespace: any; kind: any; }'.
```

christianvogt:
@glekner ^^

glekner:
fixed @rawagner @christianvogt

PR #8092, opened 2021-02-08
Bug 1891766: Disable submit button for incorrect validations (LSO/OCS local volume set)

SanjalKatiyar:
Before:
"Next" button is enabled even if user has entered a wrong disk size/limit value.
Screenshot from 2021-02-08 18-23-17

After:
"Next" button will be disabled if user enters a incorrect disk size/limit value.
Screenshot from 2021-02-08 18-19-18


afreen23 (changes requested):
You don't need to create another state.
We have the disablement control here : https://github.com/openshift/console/blob/db9138bca35526459516f36c7300a450ed3a1aaf/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/create-sc/create-sc.tsx#L222

We might want to add more here.

afreen23 (approved):
looks good just nits

SanjalKatiyar:
> looks good just nits

thanks for the review... I have made the changes...

PR #8100, opened 2021-02-08
i18n: avoid 404 errors attempting to load en-US

spadgett:
Configure i18next to only the language instead of the full locale
to avoid 404 errors on console load. This improves the load time of
console since we have many namespaces.

https://www.i18next.com/overview/configuration-options

/assign @rebeccaalpert


rebeccaalpert (approved):
Looks good to me locally.

PR #8070, opened 2021-02-05
Bug 1924387: Fix malformed ja/zh modal delete message

spadgett:
/assign @rebeccaalpert 

Screen Shot 2021-02-05 at 10 43 59 AM
Screen Shot 2021-02-05 at 10 43 43 AM

PR #8101, opened 2021-02-08
Bug 1928875: Update translations

rebeccaalpert:
Translation sent us an updated file set including fixes for inconsistent Object Bucket/Object Bucket Claims, PersistentVolume/PersistentVolumeClaim, and other K8sKind translations, inconsistent parenthesis usage mentioned by Andy Pickering, and updated translations for missing items such as true/false and Add, Edit, and Remove HorizontalPodAutoscaler that were left out of the final batch for 4.7.

Fixes https://issues.redhat.com/browse/CONSOLE-2755 and https://bugzilla.redhat.com/show_bug.cgi?id=1928875.

**Modifications I made:**
- In the last translation update PR, QE had asked "If you wish to use the Arbiter stretch cluster, a minimum of 4 nodes (2 different zones, 2 nodes per zone) and 1 additional zone with 1 node is required. All nodes must be pre-labeled with zones in order to be validated on cluster creation." was correct. (The text was originally 2 zones per node). I thought it was a typo I'd made and I had updated all the keys. I just noticed the translations are now off (I can't read Chinese, but it looks like Japanese is definitely off). I'm removing both translations for now since they don't accurately reflect the English text. We'll get those two fixed in the next translation handoff.
- Updated a test I noticed had the wrong text
- Small grammar updates in English files/related translation keys
- Fixed "Network attachement" vs. "Network attachment" in Japanese files. It seems to be on the translators' side. I'll mention these to them so we don't regress.
- Chinese translation for "Edit annotations" needs to be updated on the translators' side. We had fixed these locally for QE in a recent bug. I maintained the fix here in these files so we don't regress.
- Addressed fixes made in https://github.com/openshift/console/pull/8070 to avoid a regression; we will also need to maintain this for upcoming Korean files, which are probably using the older data set affected by this bug.

**Caveats:**
- I didn't touch English K8sKind casing because I didn't want this PR to be huge. I opened [CONSOLE-2756](https://issues.redhat.com/browse/CONSOLE-2756) to track this.
- Japanese translations are inconsistently humanizing some K8sKinds; we should bring this up before we hand off the next round of files. I don't want to maintain a shadow translation database, so I'm not manually updating these here.


spadgett:
Thanks, @rebeccaalpert. Small request: It would be good to commit the translations as-is in one commit, and then make any edits in separate commits. This will make the changes easier to review, and it will be easier for the globalization team to see what changes we've made as well.

Let's open a Bugzilla bug so we can backport the updated translations to 4.7.z.

frontend/packages/ceph-storage-plugin/locales/en/ceph-storage-plugin.json
```diff
@@ -271,7 +271,7 @@
...
   "Create new storage class": "Create new storage class",
   "All required fields are not set": "All required fields are not set",
-  "In order to create the storage cluster you must set the storage class select at least 3 nodes (preferably in 3 different zones) and meet the minimum or recommended requirement": "In order to create the storage cluster you must set the storage class select at least 3 nodes (preferably in 3 different zones) and meet the minimum or recommended requirement",
+  "In order to create the storage cluster you must set the storage class, select at least 3 nodes (preferably in 3 different zones), and meet the minimum or recommended requirement": "In order to create the storage cluster you must set the storage class, select at least 3 nodes (preferably in 3 different zones), and meet the minimum or recommended requirement",
```
spadgett, line 274: The second comma added here is not needed.

```suggestion
  "In order to create the storage cluster you must set the storage class, select at least 3 nodes (preferably in 3 different zones) and meet the minimum or recommended requirement": "In order to create the storage cluster you must set the storage class, select at least 3 nodes (preferably in 3 different zones) and meet the minimum or recommended requirement",
```

spadgett (commented):
LGTM, just one small nit on the punctuation of one of the changed messages.

frontend/packages/ceph-storage-plugin/locales/en/ceph-storage-plugin.json
rebeccaalpert: Should be all set!

rebeccaalpert:
@spadgett - Added Korean files and applied changes to key text from this and the last translation PR in a separate commit.

spadgett:
@rebeccaalpert We'll need to update the language dropdown to include Korean as an option.

spadgett:
/assign @yapei 
for QE review. We're planning to treat this as a bug so we can add any missing translations to 4.7.z.

spadgett (approved):
/lgtm
/hold

frontend/public/locales/zh/machine-config-pools.json
```diff
@@ -5,8 +5,8 @@
   "Current configuration": "当前配置",
   "Current configuration source": "当前配置源",
   "Paused": "暂停",
-  "True": "True",
-  "False": "False",
+  "True": "真",
+  "False": "假",
```
yanpzhan, line 9: There are some other "True/False" on the machineconfigpool list page, not sure if all of them should be translated:
Screenshot from 2021-02-19 17-28-23

yanpzhan:
Screenshot from 2021-02-19 17-53-51
"Developer" is not translated in "Administrator" dropdown list.

spadgett:
> Screenshot from 2021-02-19 17-53-51
> "Developer" is not translated in "Administrator" dropdown list.

Can you try refreshing the page after switching languages? We have a bug where these items don't update immediately.

frontend/public/locales/zh/machine-config-pools.json
spadgett: I think we probably need to translate true/false for conditions. We're showing the status value as-is currently.

yanpzhan:
> > Screenshot from 2021-02-19 17-53-51
> > "Developer" is not translated in "Administrator" dropdown list.
> 
> Can you try refreshing the page after switching languages? We have a bug where these items don't update immediately.

"Developer" is still not translated after several refreshing.

rebeccaalpert:
Thanks @yanpzhan. It looks like they didn't provide a translation for "Developer." I'll reach out to translation and see if I can modify the conditions table for true/false.

rebeccaalpert:
I internationalized the True/False conditions in the Conditions component. I haven't heard back from translation about the "Developer" translation.

rebeccaalpert:
Still haven't heard back from Terry about the "Developer" translation.

yanpzhan:
Tested the pr again, do we need wait for "Developer" translation in this pr or leave it in a following pr?

rebeccaalpert:
@spadgett - Do you want to wait or handle any "Developer" string changes as a follow on? It sounds like Terry is on it, but I'm not sure when we'll know about the string.

rebeccaalpert:
Hey @yanpzhan - the Korean translator sent us an updated string for "Developer." Can you take another look when you get a chance? Thank you!

spadgett (approved):
/lgtm
/hold for QE approval

rebeccaalpert:
Rebased.

jhadvig (approved):
@rebeccaalpert thanks for this fix, looks great 👍 

/lgtm
/test ceph-storage-plugin

PR #8105, opened 2021-02-09
Bug 1926072: Fix close button in the new 'Storage cluster exists' warning alert modal

SanjalKatiyar:
Before:
Close Button for the alert modal was not working.
Screenshot from 2021-02-08 12-41-45

After:
Fixed the issue, it now redirects the user back to Storage Cluster Tab page.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
@@ -24,6 +24,17 @@ const ExistingClusterModal: React.FC<ExistingClusterModalProps> = ({ match, stor
     OCSServiceModel,
   )}/${clusterName}`;
 
+  const storageClusterListPage = `${resourcePathFromModel(
+    ClusterServiceVersionModel,
+    appName,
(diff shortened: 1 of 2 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
```diff
@@ -24,6 +24,17 @@ const ExistingClusterModal: React.FC<ExistingClusterModalProps> = ({ match, stor
     OCSServiceModel,
   )}/${clusterName}`;
 
+  const storageClusterTabPath = `${resourcePathFromModel(
```
afreen23, line 27: Nit:
```suggestion
  const storageClusterListPage = `${resourcePathFromModel(
```

afreen23 (approved):
Looks good just a nit

frontend/packages/ceph-storage-plugin/src/components/ocs-install/existing-cluster-modal.tsx
SanjalKatiyar: ack...

PR #8106, opened 2021-02-09
Add support for taskSpec in pipeline visualization component in the pipeline details page.

debsmita1:
**JIRA Story:**
https://issues.redhat.com/browse/ODC-5334

**Solution Description**: 
- removed the `Pipeline taskSpec is not supported` alert in the pipeline details page
- if the pipeline has inline taskSpec, then passing `{ data: { spec: task.taskSpec } }` to the task prop in the TaskComponent
- removed the `sortTasksByRunAfterAndFrom` util

taskSpec
p4
p3
p2
p1

updated test cases

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualization.tsx
```diff
@@ -20,19 +19,6 @@ const PipelineVisualization: React.FC<PipelineTopologyVisualizationProps> = ({
...
   const { t } = useTranslation();
   let content: React.ReactElement;
-  if (hasInlineTaskSpec(pipeline.spec.tasks)) {
-    // TODO: Inline taskSpec is not yet supported feature
-    content = (
-      <Alert
-        variant="info"
-        isInline
-        title={t(
-          'pipelines-plugin~This Pipeline cannot be visualized. Pipeline taskSpec is not supported.',
```
karthikjeeyar, line 30: please remove this entry from pipeline-plugin.json

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
```diff
@@ -133,6 +132,7 @@
   "Add resource": "Add resource",
   "Error loading the tasks.": "Error loading the tasks.",
   "Unable to locate any tasks.": "Unable to locate any tasks.",
+  "This Pipeline cannot be visualized. Pipeline taskSpec is not supported.": "This Pipeline cannot be visualized. Pipeline taskSpec is not supported.",
```
karthikjeeyar, line 135: run `yarn i18n`

debsmita1: This is being used in the PipelineBuilderVisualization

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
```diff
@@ -100,6 +100,22 @@ export const PipelineVisualizationTask: React.FC<PipelineVisualizationTaskProp>
...
+        namespace={namespace}
+        status={taskStatus}
+        isPipelineRun={!!pipelineRunStatus}
+        disableTooltip={disableTooltip}
+        selected={selected}
+        width={width}
+        height={height}
+      />
+    );
+  }
```
karthikjeeyar, line 118: PipelineVisualizationTask accepts `disableTooltip` and if that is passed, then the control will early return and it will not reach here at all. I think this will be problem when the edit pipeline option is enabled. Better return this one first based on taskSpec's availability.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
```diff
@@ -100,6 +100,22 @@ export const PipelineVisualizationTask: React.FC<PipelineVisualizationTaskProp>
   }
 
   let resources;
+  if (task.taskSpec) {
+    return (
+      <TaskComponent
+        pipelineRunName={pipelineRunName}
+        task={{ data: { spec: task.taskSpec } }}
```
rottencandy, line 107: Since `task` is the only prop that differs from `taskComponent`'s props, maybe it would be easier to check and conditionally pass it to `taskComponent` itself.

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
karthikjeeyar: Ah okay :+1:.

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
```diff
@@ -133,13 +133,16 @@ const sortTasksByRunAfterAndFrom = (
     let flag = -1;
     if (conditions.hasRunAfterDependency(output[i])) {
       for (let j = 0; j < output.length; j++) {
-        if (i < j && output[j].taskRef.name === output[i].runAfter[output[i].runAfter.length - 1]) {
+        if (
+          i < j &&
+          output[j].taskRef?.name === output[i].runAfter[output[i].runAfter.length - 1]
```
karthikjeeyar, line 138: This sort function seems to support only taskRef, to handle taskSpec, we could  add  this fallback ` || t.name  === output[i].runAfter[output[i].runAfter.length - 1]` ? 
but I am wondering, regardless of this sorting function's output the final structure of the pipeline is already correct :thinking:

debsmita1: yes right, I tried with different yamls but nowhere did the visualization break
cc @andrewballantyne ^

karthikjeeyar:
Verified locally, visualization works fine. I assume there edit-pipeline kebab action will be handled as part of taskspec pipeline builder support story.
/lgtm

andrewballantyne (changes requested):
Sorry, could you just delete the sorting function? We have some legacy code left over from the <4.3 version of the Pipeline Visualization where it rendered in columns.

We need to severely clean this code up at some point but at least we can remove code we don't need as we touch it.

andrewballantyne (approved):
Definitely should look to trim out that function more, but this will work for now 👍

PR #8107, opened 2021-02-09
Bug 1926735: add template support check if use it separately

gouyang:
RHEL6 is not supported by Red Hat, it pops up a warning window for user to confirm the dialog, need to address this pop-up window when open it separately during tests.


gouyang:
Verified in `https://cnv-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/test-kubevirt-console-4.7-cnv-2.6/46/console`

PR #8110, opened 2021-02-09
Bug 1926843: Consider finally tasks when calculating task status

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5504

**Analysis / Root cause**: 
Finally tasks are not considered when calculating task statuses.

**Solution Description**: 
Consider finally tasks when calculating status

(No UI change)
scrt-screenshot-16128811333

**Test setup:**
Example pipeline with finally: https://raw.githubusercontent.com/tektoncd/pipeline/v0.19.0/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/utils/pipeline-augment.ts
+++ b/frontend/packages/pipelines-plugin/src/utils/pipeline-augment.ts
@@ -210,9 +210,18 @@ export const getPipelineFromPipelineRun = (pipelineRun: PipelineRunKind): Pipeli
   };
 };
 
-export const getTaskStatus = (pipelinerun: PipelineRunKind): TaskStatus => {
+export const totalPipelineRunTasks = (pipelinerun: PipelineRunKind): number => {
   const executedPipeline = getPipelineFromPipelineRun(pipelinerun);
-  const totalTasks = (executedPipeline?.spec?.tasks || []).length ?? 0;
+  if (!executedPipeline) {
+    return 0;
+  }
+  const totalTasks = (executedPipeline.spec?.tasks || []).length ?? 0;
+  const finallyTasks = (executedPipeline.spec?.finally || []).length ?? 0;
+  return totalTasks + finallyTasks;
+};
+
+export const getTaskStatus = (pipelinerun: PipelineRunKind): TaskStatus => {
+  const totalTasks = totalPipelineRunTasks(pipelinerun);
   const plrTasks =
     pipelinerun && pipelinerun.status && pipelinerun.status.taskRuns
       ? Object.keys(pipelinerun.status.taskRuns)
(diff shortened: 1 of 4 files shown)


frontend/packages/pipelines-plugin/src/utils/pipeline-augment.ts
```diff
@@ -210,9 +210,15 @@ export const getPipelineFromPipelineRun = (pipelineRun: PipelineRunKind): Pipeli
   };
 };
 
-export const getTaskStatus = (pipelinerun: PipelineRunKind): TaskStatus => {
+const totalPipelineRunTasks = (pipelinerun: PipelineRunKind): number => {
   const executedPipeline = getPipelineFromPipelineRun(pipelinerun);
   const totalTasks = (executedPipeline?.spec?.tasks || []).length ?? 0;
+  const finallyTasks = (executedPipeline?.spec?.finally || []).length ?? 0;
```
karthikjeeyar, line 216: `spec` is not an optional field in the `pipeline`, but this `getPipelineFromPipelineRun` may return `null` when the pipelinerun status is not present initially when it is created. So for better readability, we could be explicit here instead of optional chaining, What do you think?
eg: 
```suggestion
  if ( !executedPipeline ) {
   return 0;
} else {
  const totalTasks = (executedPipeline.spec?.tasks || []).length ?? 0;
  const finallyTasks = (executedPipeline.spec?.finally || []).length ?? 0;
  return totalTasks + finallyTasks;
}
```

rottencandy: Updated.

karthikjeeyar:
Verified locally. It works fine
/lgtm

frontend/packages/pipelines-plugin/src/utils/pipeline-augment.ts
```diff
@@ -210,9 +210,18 @@ export const getPipelineFromPipelineRun = (pipelineRun: PipelineRunKind): Pipeli
   };
 };
 
-export const getTaskStatus = (pipelinerun: PipelineRunKind): TaskStatus => {
+const totalPipelineRunTasks = (pipelinerun: PipelineRunKind): number => {
```
andrewballantyne, line 213: No tests?

rottencandy: Added `totalPipelineRunTasks` tests for all the test data.

andrewballantyne (approved):

debsmita1:
verified changes locally
/lgtm

PR #8112, opened 2021-02-09
Remove dead code related to developer catalog and service catalog

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5509

**Analysis / Root cause**: After new catalog was written and moved to dev-console package, the old dead code needed to be removed. Also, the old service catalog has been deprecated since OCP 4.5 which needed to be removed as well. This also fixes one bad dependency issue where code in console/internal was importing from dev-console package.

**Solution Description**: Remove dead code not being used anywhere. 

**Screen shots / Gifs for design review**: No UI changes.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge


jerolimov (approved):
Looks good. Operator Hub and Developer Catalog still works :)

/lgtm

PR #8114, opened 2021-02-09
Use dry run options directly to remove bad dependency

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5510

**Analysis / Root cause**: `dryRunOpt` object was being used by console/internal package imported from dev-console package creating a bad dependency.

**Solution Description**: Use dry run options directly and remove bad dependency.

**Screen shots / Gifs for design review**: NO UI change.


jerolimov (approved):
:+1: 

/lgtm

PR #8116, opened 2021-02-09
Bug 1920699: retry co-fetch on 409 POST or 429

christianvogt:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1920699

**Analysis / Root cause**: 
When multiple requests are made in parallel to create k8s resources where there is a quota in place which counts the number of resources of a specific type, it's possible to receive a `409` error code back from the API server.

**Solution Description**: 
Modified `co-fetch` utility to retry automatically `fetch` requests which result in status code `429` or `409 + specific response`.
The reason for specifically checking for `POST` and the response json when status `409` is received is because a k8s call to update a resource which fails with status code `409` will most likely continue to fail unless the request is modified. Eg. 409 when a resource with the same name exists. There is however the case specified in the bugzilla where reattempting the same request after a `409` will succeed.

no UI change

Added new unit tests to test retry attempts.

**Test setup:**
See https://bugzilla.redhat.com/show_bug.cgi?id=1920699#c11

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

cc @spadgett

### DIFF
--- a/frontend/__tests__/public/co-fetch.spec.js
+++ b/frontend/__tests__/public/co-fetch.spec.js
@@ -1,6 +1,15 @@
-import { shouldLogout } from '../../public/co-fetch';
+import { coFetch, shouldLogout, validateStatus, RetryError } from '../../public/co-fetch';
 
 describe('coFetch', () => {
+  const json = async () => ({
+    details: {
+      kind: 'clusterresourcequotas',
+    },
+  });
+  const emptyHeaders = new Headers();
+  const headers = new Headers();
+  headers.set('content-type', 'application/json');
+
   it('logs out users who get a 401 from k8s', () => {
     expect(shouldLogout('/api/kubernetes/api/v1/pods')).toEqual(true);
   });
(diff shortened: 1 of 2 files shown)


frontend/__tests__/public/co-fetch.spec.js
```diff
@@ -1,6 +1,16 @@
...
 describe('coFetch', () => {
+  const json = async () => ({
+    details: {
+      group: 'quota.openshift.io',
+      kind: 'clusterresourcequotas',
+    },
+  });
+  const emptyHeaders = new Headers();
+  const headers = new Headers();
+  headers.set('content-type', 'json');
```
spadgett, line 12: I know this is just test code, but this should probably be `application/json`.

```suggestion
  headers.set('content-type', 'application/json');
```

frontend/public/co-fetch.js
```diff
@@ -56,6 +62,17 @@ const validateStatus = (response, url) => {
...
 
   return response.json().then((json) => {
+    // retry 409 conflict errors due to ClustResourceQuota
+    // https://bugzilla.redhat.com/show_bug.cgi?id=1920699
+    if (
+      retry &&
+      method === 'POST' &&
+      response.status === 409 &&
+      _.get(json, 'details.group') === 'quota.openshift.io' &&
+      _.get(json, 'details.kind') === 'clusterresourcequotas'
```
spadgett, line 72: I believe this can happen with `resourcequotas` as well. It's just less likely.

Do we need to check group/kind? I'd think a 409 on a `POST` request is always an update to an adjacent resource whether it's quota or something else. In which case, it makes sense to retry regardless of the specific kind.

christianvogt: Well interestingly enough I did try with just a `ResourceQuota` but couldn't reproduce. It is an intermittent issue so perhaps I just need to keep trying. I'll investigate further but it was quite reproducible with `ClusterResourceQuota`.
As for retrying only every 409: If you attempt to POST a new resource with a name that already exists, you get a `409`. There is no point in blindly retrying because each subsequent attempt will always fail. Because of this use case, I chose to be more specific.

christianvogt: After some more testing I was able to reproduce it with `ResourceQuota`. I will add it as well.

frontend/__tests__/public/co-fetch.spec.js
christianvogt: updated

frontend/public/co-fetch.js
christianvogt: @spadgett i've updated the code to also check for `resourcequotas` and dropped the `group` check.

spadgett (approved):
/lgtm

👍

PR #8122, opened 2021-02-10
Fix flakes in Ceph CI

bipuladh:
- Change execution flow of add-capacity test
- Remove validation for progress state in Pool test


bipuladh:
/test ceph-storage-plugin
Retesting to check if it flakes.

bipuladh:
/test ceph-storage-plugin
Retesting to check if it flakes.

bipuladh:
/test ceph-storage-plugin
Retesting to check if it flakes.

PR #8127, opened 2021-02-10
Add default value to the runtime VERSION parameter in add flow pipelines

karthikjeeyar:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3786

**Analysis / Root cause**: 

Builder Image runtime version is not considered during add flow pipeline creation, so choosing a different version will still use the same base image.

**Solution Description**: 

In Add flow pipeline creation, pass appropriate VERSION param value  based on the user selection.

Version_param

image

 ---
 
**Test setup:**

 VERSION parameter will be included in the pipeline templates only in 1.4 pipelines operator.
 
**workaround:** Manually replace MAJOR_VERSION param with  VERSION param in the pipeline templates in `openshift` namespace.
 
 1. In `openshift` ns , find `s2i-nodejs` pipeline and change the MAJOR_VERSION param with VERSION
 2. now,  from GIT add flow, add a `nodejs`  Git Repo URL 
 3. choose a Builder Image version
 4. add pipeline option before submitting the form.
 
 
**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @andrewballantyne @pradeepitm12 @vdemeester

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
@@ -37,6 +37,7 @@ export const getPipelineParams = (
   gitRef: string,
   gitDir: string,
   dockerfilePath: string,
+  tag: string,
 ) => {
   return params.map((param) => {
     switch (param.name) {
@@ -52,6 +53,8 @@ export const getPipelineParams = (
         return { ...param, default: getImageUrl(name, namespace) };
       case 'DOCKERFILE':
         return { ...param, default: dockerfilePath };
+      case 'VERSION':
+        return { ...param, default: tag || param.default };
       default:
         return param;
     }
@@ -66,6 +69,7 @@ export const createPipelineForImportFlow = async (
   gitDir: string,
   pipeline: PipelineData,
   dockerfilePath: string,
+  tag: string,
 ) => {
   const template = _.cloneDeep(pipeline.template);
 
@@ -85,6 +89,7 @@ export const createPipelineForImportFlow = async (
       gitRef,
       gitDir,
       dockerfilePath,
+      tag,
     );
 
   return k8sCreate(PipelineModel, template, { ns: namespace });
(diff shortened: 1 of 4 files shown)


frontend/packages/pipelines-plugin/src/components/import/pipeline/pipeline-template-utils.ts
```diff
@@ -52,6 +53,8 @@ export const getPipelineParams = (
         return { ...param, default: getImageUrl(name, namespace) };
       case 'DOCKERFILE':
         return { ...param, default: dockerfilePath };
+      case 'VERSION':
```
andrewballantyne, line 56: @karthikjeeyar  How difficult would it be to support both VERSION and the MAJOR/MINOR counterparts that are going to be removed in 1.4?

karthikjeeyar: @andrewballantyne  Setting the MAJOR & MINOR values is not a big deal, but the template selection process is tricky.
1.3 will have multiple templates for same builder image. eg: s2i-java-8 and s2i-java-9 so we need to fetch based on the annotations `pipeline.openshift.io/supported-version`.

In 1.4, there will be only one template per builder Image, so  when to use multiple fetching vs default fetching of templates is not straight forward.
On top of that all this needs to be removed when 1.4 is available. What are your thoughts on this ?

andrewballantyne: Yeah.... I was worried that was going to be your answer. That was my thoughts too. Let's just trust upstream will come and deliver and we don't need to put in a fallback in case it doesn't land. I retract my request 🙂

andrewballantyne (approved):
Looks good from our end... will have to make sure when the templates are updated as part of 1.4 we give it another test. `image.tag` will send `latest` if it's chosen and that'll have to be considered in the templates that upstream provides us.

cc @karthikjeeyar perhaps log a task and link the two issues together (upstream and ours) to make sure we have a working situation in the future.

debsmita1:
verified the changes
/lgtm

PR #8142, opened 2021-02-11
Bug 1901207: Pipeline resources table not immediately updated after Name filter applied or removed

debsmita1:
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=1901207

**Analysis / Root cause:**
The name filter is not passed down to Firehose resources

**Solution Description:**
Passed nameFilter to Firehose resources

**GIF:** 
pipeline-list

**Browser conformance:**
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx
@@ -14,7 +14,7 @@ interface PipelinesResourceListProps extends React.ComponentProps<typeof FireMan
 
 const PipelinesResourceList: React.FC<PipelinesResourceListProps> = (props) => {
   const { t } = useTranslation();
-  const { namespace, showTitle = true, selector, name } = props;
+  const { namespace, showTitle = true, selector, name, nameFilter } = props;
 
   const resources = [
     {
@@ -22,7 +22,7 @@ const PipelinesResourceList: React.FC<PipelinesResourceListProps> = (props) => {
       kind: referenceForModel(PipelineModel),
       namespace,
       prop: PipelineModel.id,
-      filters: { ...filters },
+      filters: { ...filters, name: nameFilter },
       selector,
       name,
     },


jerolimov (approved):
@debsmita1 Interesting, I missed the title and tried this first with Pipelines and we have the same issue here and your change fixed this for Pipelines and Pipeline Resources.

Also if you only touched the PipelineResourceList.tsx.  `packages/pipelines-plugin/src/components/pipelines/list-page/PipelinesList.tsx` has a similar code, if you want you can check why we don't need the change their.

But at least, this works for both cases.

/lgtm

jerolimov:
@debsmita1  @andrewballantyne I was a little but confused, it looks like that `packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx` renders the `Pipeline` list and the `PipelineResource` list?

I now found that the `plugin.tsx` connects the `Pipeline` list to this component. `Resource` is used here as generic k8s term, right? And and it also contains a filter for `kind: referenceForModel(PipelineModel),` which makes sense. But I don't get why this fixes also the issue for "real" `PipelineResource`s. Maybe you can help me here. :)

https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/plugin.tsx#L289-L300

andrewballantyne:
> @debsmita1 @andrewballantyne I'm still a little bit confused, it looks like that `packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx` renders the `Pipeline` list and the `PipelineResource` list? 🤔
> 
> I now found that the `plugin.tsx` connects the `Pipeline` list to this component. `Resource` is used here as generic k8s term, right? And and it also contains a filter for `kind: referenceForModel(PipelineModel),` which makes sense. But I don't get why this fixes also the issue for "real" `PipelineResource`s. Maybe you can help me here. :)
> 
> https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/plugin.tsx#L289-L300

Is there a question here @jerolimov or are you just stating what you know and asking us to validate?

debsmita1:
> @debsmita1 @andrewballantyne I'm still a little bit confused, it looks like that `packages/pipelines-plugin/src/components/pipelines/PipelinesResourceList.tsx` renders the `Pipeline` list and the `PipelineResource` list? 
> 
> I now found that the `plugin.tsx` connects the `Pipeline` list to this component. `Resource` is used here as generic k8s term, right? And and it also contains a filter for `kind: referenceForModel(PipelineModel),` which makes sense. But I don't get why this fixes also the issue for "real" `PipelineResource`s. Maybe you can help me here. :)
> 
> https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/plugin.tsx#L289-L300

@jerolimov my PR only fixes the issue in the Pipeline list page 
the list page for Pipeline is `PipelinesResourceList.tsx`
the list page for PipelineResources is `PipelineResourcesList.tsx`

andrewballantyne (approved):

PR #8145, opened 2021-02-11
Bug 1927941: Fix frontend descriptor API edge cases

TheRealJon:
Add better validation for status and spec descriptor details items.
Log warnings in console when x-descriptors are incompatible with the data they point to.
Fix warning about missing key prop in DescriptorDetailsItemList


frontend/packages/console-shared/src/components/status/Status.tsx
```diff
@@ -98,7 +99,7 @@ export const Status: React.FC<StatusProps> = ({
       return <StatusIconAndText {...statusProps} icon={<NotStartedIcon />} />;
 
     default:
-      return <>{status || DASH}</>;
+      return <>{status && _.isString(status) ? status : DASH}</>;
```
spadgett, line 102: This changes the output when the value is the empty string, `''`. Is that intentional?

spadgett (commented):
Do we have an error boundary around the descriptor?

frontend/packages/console-shared/src/components/status/Status.tsx
TheRealJon: Hmm, no. It also changes if status was a number or boolean value. Let me think about this line a little more.

TheRealJon:
@spadgett I've updated this PR to focus more on validating details page descriptors rather than updating the Status component. All descriptors on operand details pages now run through the same validation as the creation form. So, each x-descriptors is validated against the referenced resource property, and any that are incompatible with the property are ignored and a warning is printed to console. I opted to leave in deprecated descriptors and just log a warning for now, but that could also be adjusted.

TheRealJon:
> Do we have an error boundary around the descriptor?

Yes. Descriptor-backed details items are surrounded by an error boundary. Previous to this PR, the only console output when encountering this bug was the error message from the uncaught exception. This PR adds console warnings that should help troubleshoot when a descriptor is not behaving as expected.

spadgett (approved):
/lgtm

TheRealJon:
Looks like the performance test is failing because my changes put the OLM package size just above the 150 Kb limit.

TheRealJon:
/hold
Should we adjust max olm chunk size?

spadgett:
> Should we adjust max olm chunk size?

Let's just remove the performance scenario. They've been flakey, and they're not testing a bottleneck (as shown when profiling console load). IMO, the best places to improve performance are 

* Keeping our dependency bundle size low
* Improving caching
* Reducing the number of network requests (in particular for i18next namespaces)

The bundle sizes of these files are negligible in comparison.

@rhamilto FYI, since I know you're running into this, too.

TheRealJon:
> Let's just remove the performance scenario. They've been flakey, and they're not testing a bottleneck (as shown when profiling console load). IMO, the best places to improve performance are
> 
> * Keeping our dependency bundle size low
> * Improving caching
> * Reducing the number of network requests (in particular for i18next namespaces)
> 
> The bundle sizes of these files are negligible in comparison.
> 
> @rhamilto FYI, since I know you're running into this, too.

@spadgett Do we want to disable the performance tests as part of this PR?

spadgett (approved):
/lgtm

spadgett:
@TheRealJon It looks like the test failure might be legitimate.

TheRealJon:
@spadgett Updated. Should pass e2e now.

TheRealJon:
/hold

CI issues

spadgett (approved):
/lgtm

TheRealJon:
Network issues during build...

/retest

PR #8150, opened 2021-02-12
remove gap in dashboard for empty quickstart card

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-5221

# Issue
Gap in right hand panel occurs as a Card is rendered even when no Quickstarts are visible to a user

# Solution
Only render card when some Quickstart is visible to a user.

# How to test
Remove `ConsoleQuickStarts` CRD, it will take some time for the operator to restore them, meanwhile the scenario can be tested. Otherwise hardcode the output of `QuickStartLoader` to `null` or `[]`

# Screenshot
Screenshot from 2021-02-12 17-12-23
Screenshot from 2021-02-12 17-03-07

# Tests
No tests altered

# Browser Conformance
Chrome 88

### DIFF
--- a/frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
+++ b/frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
@@ -32,12 +32,14 @@ type QuickStartsCatalogCardProps = {
   quickStarts: QuickStart[];
   storageKey: string;
   userSettingsKey: string;
+  shouldShowQSTile?: (boolean) => void;
 };
 
 const QuickStartsCatalogCard: React.FC<QuickStartsCatalogCardProps> = ({
   quickStarts,
   storageKey,
   userSettingsKey,
+  shouldShowQSTile,
 }) => {
   const { t } = useTranslation();
   const { allQuickStartStates, setActiveQuickStart } = React.useContext<QuickStartContextValues>(
@@ -49,6 +51,10 @@ const QuickStartsCatalogCard: React.FC<QuickStartsCatalogCardProps> = ({
     true,
   );
   const [isOpen, setOpen] = React.useState<boolean>(false);
+
+  React.useEffect(() => {
+    shouldShowQSTile && shouldShowQSTile(showTile);
+  }, [showTile, shouldShowQSTile]);
 
   const onRemove = () => {
     setShowTile(false);
--- a/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
+++ b/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
-import { useUserSettings } from '@console/shared';
 import Dashboard from '@console/shared/src/components/dashboard/Dashboard';
 import DashboardGrid from '@console/shared/src/components/dashboard/DashboardGrid';
-import QuickStartsLoader from '@console/app/src/components/quick-starts/loader/QuickStartsLoader';
 import { HIDE_QUICK_START_DASHBOARD_TILE_STORAGE_KEY } from '@console/shared/src/components/quick-starts/quick-starts-catalog-card-constants';
 import QuickStartsCatalogCard from '@console/shared/src/components/quick-starts/QuickStartsCatalogCard';
+import { QuickStart } from '@console/app/src/components/quick-starts/utils/quick-start-types';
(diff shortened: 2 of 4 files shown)


debsmita1:
I am not able to remove the Quick Starts card from the dashboard
could you please pass sync as true in the useUserSetting utils in the cluster-dashboard ?

jerolimov (approved):
Tested this against a 4.7 cluster with quick starts and a 4.6 cluster without quick starts. Works fine.

abhinandan13jan:
> I am not able to remove the Quick Starts card from the dashboard
> could you please pass sync as true in the useUserSetting utils in the cluster-dashboard ?

@debsmita1 I've updated the PR, I tried with sync: true but dint seem to make much difference so I am using a callBack to parent instead

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
```diff
@@ -50,6 +52,10 @@ const QuickStartsCatalogCard: React.FC<QuickStartsCatalogCardProps> = ({
   );
   const [isOpen, setOpen] = React.useState<boolean>(false);
 
+  React.useEffect(() => {
+    setShowQSTile && setShowQSTile(showTile);
```
debsmita1, line 56: why not add this line inside `onRemove`

```suggestion
    shouldShowQSTile && shouldShowQSTile(false);
```

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
```diff
@@ -32,12 +32,14 @@ type QuickStartsCatalogCardProps = {
   quickStarts: QuickStart[];
   storageKey: string;
   userSettingsKey: string;
+  setShowQSTile?: (boolean) => void;
```
debsmita1, line 35:
```suggestion
  shouldShowQSTile?: (boolean) => void;
```

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
```diff
@@ -21,38 +20,35 @@ const rightCards = [{ Card: ActivityCard }];
...
-                      storageKey={HIDE_QUICK_START_DASHBOARD_TILE_STORAGE_KEY}
-                      userSettingsKey={HIDE_QUICK_START_DASHBOARD_TILE_USER_SETTINGS_KEY}
-                    />
-                  )}
-                </QuickStartsLoader>
+                <QuickStartsCatalogCard
+                  quickStarts={quickStarts}
+                  storageKey={HIDE_QUICK_START_DASHBOARD_TILE_STORAGE_KEY}
+                  userSettingsKey={HIDE_QUICK_START_DASHBOARD_TILE_USER_SETTINGS_KEY}
+                  setShowQSTile={setShowQSTile}
```
debsmita1, line 44:
```suggestion
                  shouldShowQSTile={setShowQSTile}
```

debsmita1 (changes requested):

frontend/public/components/dashboard/dashboards-page/cluster-dashboard/cluster-dashboard.tsx
abhinandan13jan: updated

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
abhinandan13jan: update

frontend/packages/console-shared/src/components/quick-starts/QuickStartsCatalogCard.tsx
abhinandan13jan: update

abhinandan13jan: Well, `onRemove` will only be fired on button click.
So in a situation, where the Tile is already in a removed state, the `showTile` in `Child` is set to `false` but the `shouldShowTile` is set to `true` in parent as `onRemove` was not called.

The idea here is to keep `shouldShowTile` in parent and `showTile` in Child to be in sync

debsmita1:
verified the changes
/lgtm

abhinandan13jan:
closing as the same will be followed up in https://issues.redhat.com/browse/ODC-5013

PR #8156, opened 2021-02-14
Converting context provider extensions to dynamic

glekner:


glekner:
@vojtechszocs the promises seems to be resolved, but the value of the context is undefined in the consumers, we need to take a deeper look.

glekner:
please review @christianvogt @vojtechszocs

frontend/packages/console-app/src/plugin.tsx
```diff
@@ -282,27 +271,6 @@ const plugin: Plugin<ConsumedExtensions> = [
...
-      useValueHook: useValuesForQuickStartContext,
-    },
-  },
-  {
-    type: 'ContextProvider',
-    properties: {
-      Provider: FileUploadContext.Provider,
-      useValueHook: useValuesFileUploadContext,
-    },
-  },
```
christianvogt, line 305: This one is new. You need to convert it as well to json.

frontend/packages/kubevirt-plugin/console-extensions.json
```diff
@@ -69,5 +69,26 @@
...
+    "type": "console.page/route/standalone",
+    "properties": {
+      "exact": false,
+      "path": "/k8s/ns/:ns/virtualmachineinstances/:name/standaloneconsole",
+      "component": { "$codeRef": "standaloneConsole" }
+    },
+    "flags": {
+      "required": ["KUBEVIRT"]
+    }
+  },
```
christianvogt, line 83: Looks like this snuck in from another PR?

frontend/public/components/app.jsx
```diff
@@ -202,13 +204,15 @@ class App_ extends React.PureComponent {
   }
 }
 
-initConsolePlugins(store);
+const AppWithExtensions = withTranslation()((props) => {
+  const [contextProviderExtensions] = useResolvedExtensions(isContextProvider);
 
-const App = withExtensions({ contextProviderExtensions: isContextProvider })(App_);
```
christianvogt, line 207: Can we add a 2nd param to `withExtensions` which if `true` will resolve all extensions with `useResolvedExtensions` ?

cc @vojtechszocs

glekner: hooks can't be conditional? @christianvogt

glekner:
fixed comments @christianvogt

frontend/public/components/app.jsx
vojtechszocs: @glekner `withExtensions` is [HOC](https://reactjs.org/docs/higher-order-components.html) counterpart to `useExtensions` hook.

Long-term expectations:
- `useExtensions` hook gets replaced by `useResolvedExtensions` hook
- `withExtensions` HOC gets removed by refactoring related class components into functional ones

@christianvogt In the short-term, yes - I can post a PR for that; however, the 2nd HOC param needs to have an impact on `TExtensionProps` type param, similar to `ResolvedExtension` type used by `useResolvedExtensions` hook.

frontend/packages/console-app/package.json
```diff
@@ -29,6 +29,14 @@
     "@console/topology": "0.0.0-fixed"
   },
   "consolePlugin": {
-    "entry": "src/plugin.tsx"
+    "entry": "src/plugin.tsx",
+    "dependencies": {
```
vojtechszocs, line 33: You don't need `consolePlugin.dependencies` in a static plugin.

frontend/packages/kubevirt-plugin/package.json
```diff
@@ -44,7 +44,8 @@
     "exposedModules": {
       "standaloneConsole": "src/components/vms/vm-console/StandaloneVMConsolePage.tsx",
       "icons": "src/utils/icons.tsx",
-      "createVM": "src/components/create-vm"
+      "createVM": "src/components/create-vm",
```
vojtechszocs, line 47:
```suggestion
      "createVM": "src/components/create-vm.tsx",
```

frontend/public/components/app.jsx
```diff
@@ -202,13 +204,15 @@ class App_ extends React.PureComponent {
   }
 }
 
-initConsolePlugins(store);
+const AppWithExtensions = withTranslation()((props) => {
```
vojtechszocs, line 207: @glekner We should utilize the second tuple element (`resolved: boolean`) of `useResolvedExtensions` hook result.

This tells us whether the `CodeRef` resolution for all matching extensions is complete or still in progress. We generally don't want to re-render unless *all* extensions have their `CodeRef` properties resolved.

vojtechszocs (changes requested):

frontend/public/components/app.jsx
christianvogt: @vojtechszocs i'm fine with a `withResolvedExtensions` solution as well. I think the typing would be easier.

@vojtechszocs not sure I agree with this statement.
> useExtensions hook gets replaced by useResolvedExtensions hook
Why would we remove `useExtensions`? It serves a different purpose.

I also think having an HOC is valid. It also will be useful as part of the SDK.
> withExtensions HOC gets removed by refactoring related class components into functional ones

vojtechszocs: @christianvogt 

Regarding the removal of `useExtensions` hook - I thought we want all code to use `useResolvedExtensions` hook as the official API to consume Console extensions as part of React render flow? (Once we upgrade to TypeScript 4.x we can use [variadic tuple types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types) feature to add `options` arg to `useResolvedExtensions` hook in order to customize its behavior.)

Regarding the removal of `withExtensions` HOC - I thought the long-term strategy is moving away from React class components in favor of functional components & hooks?

christianvogt: We don't always want to resolve extension coderefs. It's up to the consumer when to resolve them. Sometimes we want to use the code ref with `React.lazy` or do something else lazily instead of resolving up front.

frontend/dynamic-demo-plugin/src/utils/bar.ts
```diff
@@ -1,8 +1,8 @@
-import { FeatureFlagHandler } from '@console/dynamic-plugin-sdk/src/extensions/feature-flags';
+import { SetFeatureFlag } from '@console/dynamic-plugin-sdk/src/extensions/feature-flags';
 
 export default (label: string) => `Hello ${label} Function!`;
 
-export const testHandler: FeatureFlagHandler = function() {
+export const testHandler: SetFeatureFlag = function() {
```
vojtechszocs, line 5: This is already fixed in #8224 - please skip this change.

Console web app will build fine even with dynamic demo plugin containing errors.

frontend/packages/kubevirt-plugin/package.json
vojtechszocs: Can be fixed later.

frontend/public/components/app.jsx
vojtechszocs: Right, the `React.lazy` is a valid use case, agreed.

vojtechszocs:
Rebased and resolved merge conflicts.

PR #8159, opened 2021-02-15
Improve fetching LSO status during cluster installation

SanjalKatiyar:
URL: https://issues.redhat.com/browse/RHSTOR-1596
Earlier:
1) We were only considering subscription object to check the LSO status.
2) We were looking for specific LSO subscription by using "metadata.name", which can be changed when installing the operator from the CLI.

After:
1) We are considering both subscription object and status of corresponding CSV object to check for the LSO status.
2) We are looking for specific LSO subscription using "spec.name" instead of "metadata.name" ( we are filtering the LSO operator from all the installed operators and this filtering is being done on the UI side )
3) Edge Case - where multiple LSOs are installed, we are considering the first one only (0th index from a list of LSOs) and CSV corresponding to it ( this issue will be addressed when a drop-down menu will be introduced and user will specifically be able to choose among multiple installed LSOs )

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
@@ -1,20 +1,28 @@
 import * as React from 'react';
 import { Trans, useTranslation } from 'react-i18next';
-import * as _ from 'lodash';
 import { match as RouterMatch } from 'react-router';
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore
 import { Alert, Button } from '@patternfly/react-core';
-import { StorageClassResourceKind, K8sResourceKind, k8sList } from '@console/internal/module/k8s';
+import {
+  StorageClassResourceKind,
+  K8sResourceKind,
+  k8sList,
+  referenceForModel,
+} from '@console/internal/module/k8s';
 import { history, LoadingBox } from '@console/internal/components/utils';
-import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import {
+  useK8sWatchResource,
+  WatchK8sResource,
+} from '@console/internal/components/utils/k8s-watch-hook';
 import { StorageClassModel } from '@console/internal/models';
-import { fetchK8s } from '@console/internal/graphql/client';
-import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager';
+import { ClusterServiceVersionModel, SubscriptionModel } from '@console/operator-lifecycle-manager';
+import { getNamespace } from '@console/shared';
 import CreateStorageClusterWizard from './install-wizard';
-import { LSOSubscriptionResource } from '../../../resources';
 import { NavUtils } from '../../../types';
 import { filterSCWithNoProv } from '../../../utils/install';
+import { getOperatorVersion } from '../../../selectors';
+import { LSO_OPERATOR } from '../../../constants';
 import './attached-devices.scss';
 
 const goToLSOInstallationPage = () =>
@@ -31,34 +39,34 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
 
   const { appName, ns } = match.params;
   const [hasNoProvSC, setHasNoProvSC] = React.useState(false);
-  const [isLsoPresent, setIsLsoPresent] = React.useState(false);
-  const [allDataLoaded, setAllDataLoaded] = React.useState(false);
-  const [lsoNs, setLsoNs] = React.useState('');
-  const [subscription, subscriptionLoaded, subscriptionLoadError] = useK8sWatchResource<
+  const [lsoSubscription, setLsoSubscription] = React.useState<K8sResourceKind>();
+  const isDataLoaded = React.useRef(false);
+  const lsoNs = getNamespace(lsoSubscription);
+  const lsoVersion = getOperatorVersion(lsoSubscription);
+
+  const subscriptionResource: WatchK8sResource = {
+    kind: referenceForModel(SubscriptionModel),
+    isList: true,
+  };
+  const [subscriptions, subscriptionsLoaded, subscriptionsLoadError] = useK8sWatchResource<
     K8sResourceKind[]
-  >(LSOSubscriptionResource);
+  >(subscriptionResource);
+
+  const csvResource = {
+    kind: referenceForModel(ClusterServiceVersionModel),
+    name: lsoVersion,
+    namespaced: true,
+    namespace: lsoNs,
+    isList: false,
+  };
+  const [csv, csvLoaded, csvLoadError] = useK8sWatchResource<K8sResourceKind>(csvResource);
+
+  if (subscriptionsLoaded && csvLoaded) isDataLoaded.current = true;
+  const isLsoCsvSucceeded = !!lsoNs && !!lsoVersion && csv?.status?.phase === 'Succeeded';
 
   React.useEffect(() => {
-    if (subscriptionLoadError || (!subscription.length && subscriptionLoaded)) {
-      setIsLsoPresent(false);
-      setAllDataLoaded(true);
-    } else if (subscriptionLoaded && !_.isEmpty(subscription[0])) {
-      fetchK8s(
-        ClusterServiceVersionModel,
-        subscription[0]?.status?.installedCSV,
-        subscription[0]?.metadata?.namespace,
-      )
-        .then(() => {
-          setIsLsoPresent(true);
-          setLsoNs(subscription[0]?.metadata?.namespace);
-          setAllDataLoaded(true);
-        })
-        .catch(() => {
-          setIsLsoPresent(false);
-          setAllDataLoaded(true);
-        });
-    }
-  }, [subscription, subscriptionLoaded, subscriptionLoadError]);
+    setLsoSubscription(subscriptions.find((item) => item?.spec?.name === LSO_OPERATOR));
+  }, [subscriptions]);
 
   React.useEffect(() => {
     /* this call can't be watched here as watching will take the user back to this view
@@ -73,9 +81,9 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
       .catch(() => setHasNoProvSC(false));
   }, [appName, ns]);
 
-  return !allDataLoaded && !subscriptionLoadError ? (
+  return !isDataLoaded.current && !subscriptionsLoadError && !csvLoadError ? (
     <LoadingBox />
-  ) : subscriptionLoadError || !isLsoPresent ? (
+  ) : subscriptionsLoadError || csvLoadError || !isLsoCsvSucceeded ? (
     <Alert
       className="co-alert ceph-ocs-install__lso-install-alert"
       variant="info"
--- a/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
+++ b/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
@@ -61,13 +61,16 @@ export const getCephSC = (scData: K8sResourceKind[]): K8sResourceKind[] =>
     );
   });
 
+export const getOperatorVersion = (operator: K8sResourceKind): string =>
+  operator?.status?.installedCSV;
+
 export const getOCSVersion = (items: FirehoseResult): string => {
   const itemsData: K8sResourceKind[] = _.get(items, 'data');
   const operator: K8sResourceKind = _.find(
     itemsData,
     (item) => _.get(item, 'spec.name') === OCS_OPERATOR,
   );
-  return _.get(operator, 'status.installedCSV');
+  return getOperatorVersion(operator);
 };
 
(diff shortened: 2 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install.tsx
```diff
@@ -36,27 +41,41 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
     K8sResourceKind[]
   >(LSOSubscriptionResource);
 
+  const csvResource = {
+    kind: referenceForModel(ClusterServiceVersionModel),
+    fieldSelector: `metadata.namespace=${getLSONamespace(
```
cloudbehl, line 46: use getNamespace from shared folder.
```suggestion
    fieldSelector: `metadata.namespace=${getNamespace(
```

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -64,6 +64,14 @@ export const getOCSVersion = (items: FirehoseResult): string => {
   return _.get(operator, 'status.installedCSV');
 };
 
+export const getLSOVersion = (operator: K8sResourceKind): string => {
```
cloudbehl, line 67:
```suggestion
export const getOperatorVersion = (operator: K8sResourceKind): string => {
```

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -64,6 +64,14 @@ export const getOCSVersion = (items: FirehoseResult): string => {
   return _.get(operator, 'status.installedCSV');
```
cloudbehl, line 64:
```suggestion
  return getOperatorVersion(operator);
```

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -64,6 +64,14 @@ export const getOCSVersion = (items: FirehoseResult): string => {
...
 };
 
+export const getLSOVersion = (operator: K8sResourceKind): string => {
+  return _.get(operator, 'status.installedCSV');
+};
+
+export const getLSONamespace = (operator: K8sResourceKind): string => {
+  return _.get(operator, 'metadata.namespace');
+};
```
cloudbehl, line 73:
```suggestion
```

SanjalKatiyar: thanks for the review... I have made the required changes...

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices/install.tsx
```diff
@@ -36,27 +42,41 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
   >(LSOSubscriptionResource);
 
+  const csvResource = {
+    kind: referenceForModel(ClusterServiceVersionModel),
+    fieldSelector: `metadata.namespace=${getNamespace(
+      subscription[0],
+    )},metadata.name=${getOperatorVersion(subscription[0])}`,
+    namespaced: true,
+    isList: true,
+  };
```
afreen23, line 52:
```suggestion
    kind: referenceForModel(ClusterServiceVersionModel),
    fieldSelector: metadata.name=${getOperatorVersion(subscription[0])},
    namespace: lsoNs
```

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -55,13 +55,17 @@ export const getCephSC = (scData: K8sResourceKind[]): K8sResourceKind[] =>
     );
   });
 
+export const getOperatorVersion = (operator: K8sResourceKind): string => {
+  return _.get(operator, 'status.installedCSV');
```
afreen23, line 59:
```suggestion
  return operator?.status?.installedCSV
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
afreen23, line 59:
```suggestion
export const getOperatorVersion = (operator: K8sResourceKind): string => operator?.status?.installedCSV;
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
SanjalKatiyar: made the required changes... PTAL...

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -55,13 +55,17 @@ export const getCephSC = (scData: K8sResourceKind[]): K8sResourceKind[] =>
     );
   });
 
+export const getOperatorVersion = (operator: K8sResourceKind): string => {
+  return operator?.status?.installedCSV;
+};
+
```
afreen23, line 61:
```suggestion
export const getOperatorVersion = (operator: K8sResourceKind): string => operator?.status?.installedCSV;

```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/selectors/index.ts
SanjalKatiyar: yeah... okay... I missed this in your previous suggestion...

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -32,31 +41,52 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
+    namespace: lsoNs,
+    isList: true,
+  };
+  const [csv, csvLoaded, csvLoadError] = useK8sWatchResource<K8sResourceKind[]>(csvResource);
 
   React.useEffect(() => {
-    if (subscriptionLoadError || (!subscription.length && subscriptionLoaded)) {
+    const lsoSubscription = subscription.filter((item) => item?.spec?.name === LSO_OPERATOR);
+    setLsoNs(getNamespace(lsoSubscription[0]));
+    setLsoVersion(getOperatorVersion(lsoSubscription[0]));
```
afreen23, line 65: Since its just one element only hence use `find`.
```suggestion
    const lsoSubscription = subscription.find((item) => item?.spec?.name === LSO_OPERATOR);
    setLsoNs(getNamespace(lsoSubscription));
    setLsoVersion(getOperatorVersion(lsoSubscription));
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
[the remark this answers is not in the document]
SanjalKatiyar: -> I am doing filtering for a LSO subscription from the list of all installed subscriptions and I am doing this inside `useEffect` hook (line 63) so that we do not have to loop over all the subscriptions again in case component re-renders, we will only do so if subscription/csv resource changes.
-> Once we get the first LSO operator (line 63) , now when the `CreateAttachedDevicesCluster` FC re-renders this will be passed to CSV resource (line 56 & 57) and we will get the CSV corresponding to this specific filtered LSO operator.
-> Once we get the LSO subscription and the corresponding CSV, `useEffect` will not be called any more and also value of `lsoNs` and `lsoVersion` will not change.
-> If we extract ns and version outside the hook then I have to declare `lsoSubscription` variable of type either `let` or `var` outside of `CreateAttachedDevicesCluster` FC otherwise value of the variable ` lsoSubscription` will reset to default (initialised during variable declaration) every time FC will re-render because of the watch-hook. 
Or, I have to use `useState` hook for `lsoSubscription` instead of `lsoNs` and `lsoVersion`.

This is why I used set state. I might have missed something, so, I am looking forward to your feedback on my comment.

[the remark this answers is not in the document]
SanjalKatiyar: I guess there was some misunderstanding from my end... I did not see `var` and `let` variables used at that many places in our code so I thought may be we only use `const` in practice as much as possible...
Just to confirm, is it okay to use `var` or `let`?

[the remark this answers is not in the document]
SanjalKatiyar: okay... will fix that...

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -31,34 +39,34 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
-          setAllDataLoaded(true);
-        })
-        .catch(() => {
-          setIsLsoPresent(false);
-          setAllDataLoaded(true);
-        });
-    }
-  }, [subscription, subscriptionLoaded, subscriptionLoadError]);
+    setLsoSubscription(subscription.find((item) => item?.spec?.name === LSO_OPERATOR));
+  }, [subscription]);
```
afreen23, line 69: This will lower the renders
```suggestion
React.useEffect(() => {
    const subscription = subscriptions.find((item) => item?.spec?.name === LSO_OPERATOR);
    if (lsoSubscription?.spec?.name !== LSO_OPERATOR) setLsoSubscription(subscription);
  }, [lsoSubscription, subscriptions]);
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -31,34 +39,34 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
+    kind: referenceForModel(ClusterServiceVersionModel),
+    name: lsoVersion,
+    namespaced: true,
+    namespace: lsoNs,
+    isList: false,
+  };
+  const [csv, csvLoaded, csvLoadError] = useK8sWatchResource<K8sResourceKind>(csvResource);
+
+  if (subscriptionLoaded && csvLoaded) isDataLoaded.current = true;
+  const isLsoCsvSucceeded = (lsoNs && lsoVersion && csv?.status?.phase === 'Succeeded') || false;
```
afreen23, line 65:
```suggestion
  const isLsoCsvSucceeded = (lsoNs && lsoVersion && csv?.status?.phase === 'Succeeded') || false;
```
```suggestion
   const isLsoCsvSucceeded = !!lsoNs && !!lsoVersion && csv?.status?.phase === 'Succeeded';

```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -73,9 +81,9 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
       .catch(() => setHasNoProvSC(false));
   }, [appName, ns]);
 
-  return !allDataLoaded && !subscriptionLoadError ? (
+  return !isDataLoaded.current && !subscriptionLoadError && !csvLoadError ? (
     <LoadingBox />
-  ) : subscriptionLoadError || !isLsoPresent ? (
+  ) : subscriptionLoadError || csvLoadError || !lsoSubscription || !isLsoCsvSucceeded ? (
```
afreen23, line 86: This should be suffice
```suggestion
  ) : subscriptionLoadError || csvLoadError || !isLsoCsvSucceeded ? (
```
```suggestion
  ) : subscriptionLoadError || csvLoadError || !lsoSubscription || !isLsoCsvSucceeded ? (
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -31,34 +39,34 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
+  const csvResource = {
+    kind: referenceForModel(ClusterServiceVersionModel),
+    name: lsoVersion,
+    namespaced: true,
+    namespace: lsoNs,
+    isList: false,
+  };
+  const [csv, csvLoaded, csvLoadError] = useK8sWatchResource<K8sResourceKind>(csvResource);
+
+  if (subscriptionLoaded && csvLoaded) isDataLoaded.current = true;
```
afreen23, line 64: why you need ref ?
can't we use subscriptionLoaded && csvLoaded directly ?

afreen23 (changes requested):

afreen23:
fix the frontend tests

afreen23:
Thanks @SanjalKatiyar

afreen23:
/lgtm cancel
rebase with master, run yarn i18n @SanjalKatiyar

PR #8167, opened 2021-02-15
Bug 1927882:  allow choice of binding type when creating RoleBinding

rhamilto:
Note this reverses part of changes in https://github.com/openshift/console/pull/6035 since that change caused bug 1927882.

Will require back porting to 4.6.

### DIFF
--- a/frontend/public/components/RBAC/bindings.jsx
+++ b/frontend/public/components/RBAC/bindings.jsx
@@ -285,10 +285,8 @@ export const RoleBindingsPage = ({
   staticFilters = undefined,
   name,
   kind,
-  createPath = `/k8s/${
-    namespace
-      ? `ns/${namespace}/rolebindings/~new?namespace=${namespace}`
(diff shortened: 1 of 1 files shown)


frontend/public/components/RBAC/bindings.jsx
```diff
@@ -283,13 +283,7 @@ export const RoleBindingsPage = ({
   showTitle = true,
   mock = false,
   staticFilters = undefined,
-  name,
-  kind,
```
spadgett, line 287: Do we no longer need `name` and `kind`?

spadgett (approved):

frontend/public/components/RBAC/bindings.jsx
rhamilto: Whoops.  Good catch, @spadgett.  I omitted the fixes from #7546 and #7581 that were using `name` and `kind` and didn't notice.  Updated.

spadgett (approved):
/lgtm

PR #8170, opened 2021-02-16
Add dynamic extension for standalone pages

rawagner:


frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  export type StandaloneRoutePageCodeRefs = {
+    component: CodeRef<React.FC<RouteComponentProps>>;
+  };
+}
+
+// Extension types
+
+export type StandaloneRoutePage = Extension<ExtensionProperties.StandaloneRoutePage> & {
+  type: 'console.page/standalone-route';
```
vojtechszocs, line 23:
```suggestion
  type: 'console.page/route/standalone';
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type ResolvedStandaloneRoutePage = UpdateExtensionProperties<
+  StandaloneRoutePage,
+  ExtensionProperties.StandaloneRoutePageCodeRefs
+>;
+
+// Type guards
+
+export const isStandaloneRoutePage = (e: Extension): e is ResolvedStandaloneRoutePage =>
+  e.type === 'console.page/standalone-route';
```
vojtechszocs, line 34:
```suggestion
  e.type === 'console.page/route/standalone';
```

frontend/public/components/app.jsx
```diff
@@ -213,6 +210,38 @@ render(<LoadingBox />, document.getElementById('app'));
 
 const AppWithTranslation = withTranslation()(App);
 
+const StandaloneRouter = () => {
+  const [extensions, loaded] = useResolvedExtensions(isStandaloneRoutePage);
```
vojtechszocs, line 214:
```suggestion
  const [extensions, extensionsResolved] = useResolvedExtensions(isStandaloneRoutePage);
```

rawagner: renamed

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
rawagner: renamed

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
rawagner: renamed

frontend/packages/kubevirt-plugin/package.json
```diff
@@ -40,6 +40,9 @@
         "integration-tests/tests/vm.tab.snapshots.scenario.ts",
         "integration-tests/tests/vmi.actions.scenario.ts"
       ]
+    },
+    "exposedModules": {
+      "standaloneConsole": "src/components/connected-vm-console/vm-console-page"
```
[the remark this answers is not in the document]
rawagner: Ok, thanks for pointing that out. The path is now relative & includes file extension

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+import { CodeRef, EncodedCodeRef, UpdateExtensionProperties } from '../types';
+
+export namespace ExtensionProperties {
+  export type StandaloneRoutePage = {
+    /** The component to be rendered when route matches */
+    component: EncodedCodeRef;
+    /** The route path to match against */
+    path: string;
+    /** Defines whether the path is exact or not */
```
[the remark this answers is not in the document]
rawagner: fixed.

We also talked about `Pick`-ing specific props from react-router `RouteProps`. But as soon as I import the `RouteProps` schema generator crashes. So I kept it as-is.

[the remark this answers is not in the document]
vojtechszocs: > But as soon as I import the RouteProps schema generator crashes. So I kept it as-is.

Thanks for looking into this. We can fix this in a follow-up.

frontend/public/components/app.jsx
```diff
@@ -286,12 +315,7 @@ graphQLReady.onReady(() => {
       <Provider store={store}>
         <Router history={history} basename={window.SERVER_FLAGS.basePath}>
           <Switch>
-            <Route
-              path="/k8s/ns/:ns/virtualmachineinstances/:name/standaloneconsole"
-              render={(componentProps) => (
-                <AsyncComponent loader={consoleLoader} {...componentProps} />
-              )}
-            />
+            <Route path="/standalone" component={StandaloneRouter} />
```
christianvogt, line 318: Why force the route to start with `standalone` ?

vojtechszocs: @christianvogt Scoping under `/standalone` is a convention for all pages where we don't want the common Console page layout. Other plugin-contributed route pages will be processed by the default fallback

```ts
<Route path="/" component={AppWithTranslation} />
```

christianvogt: I see the fallback. However I don't see the reasoning behind the `/standalone` prefix.
You can see how we contribute the `/terminal` today. I'd like to use this extension to do this. However I don't want `/standalone/terminal` as the URL.

rawagner: If we allow any path as standalone, without prefix, then anytime you open the path and plugin is not resolved yet, you will fall through to App component (`/` path) which will render Masthead, Navigation and other components. After some time, plugin gets resolved, the `App` component gets unmounted and plugin's component gets rendered.

Thats why I opted for predefined `standalone` prefix, where we can tell this is a standalone page and we need to wait for plugins to be resolved if the path is not matched yet.

christianvogt: You should not be using `useResolvedExtensions` here but instead `useExtensions` because we don't want to pre-resolve all standalone pages upfront.
The declarative extensions will be available immediately without the need for loading anything extra and therefore can setup the routes immediately. The use case you describe should not happen.

PR #8171, opened 2021-02-16
adding navigation extension types to dynamic-plugin-sdk

glekner:


glekner:
@vojtechszocs @christianvogt can you review

frontend/packages/console-dynamic-plugin-sdk/src/extensions/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+    /** A unique identifier for this item. */
+    id: string;
+    /** The perspective ID to which this item belongs to. If not specified, contributes to the default perspective. */
+    perspective?: string;
+    /** Navigation section to which this item belongs to. If not specified, render this item as a top level link. */
+    section?: string;
+    /** Mark this item as active when the URL starts with one of these paths. */
+    startsWith?: string[];
+    /** Insert this item before the item referenced here. */
+    insertBefore?: string;
```
christianvogt, line 14: We should align with the new capabilities for `insertBefore` and `insertAfter`
```suggestion
    /** Insert this item before the item referenced here. For arrays, the first one found in order is used. */
    insertBefore?: string | string[];
    /** Insert this item after the item referenced here. For arrays, the first one found in order is used. `insertBefore` takes precedence. */
    insertAfter?: string | string[];
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/navigation.ts
```diff
@@ -0,0 +1,69 @@
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+
+export namespace ExtensionProperties {
```
christianvogt, line 3: Can you add `NavSection` as well. This was a recent addition in order to

`console.navigation/section`
```
  export interface NavSection {
    /** A unique identifier for this item. */
    id: string;
    /** The perspective ID to which this item belongs to. If not specified, contributes to the default perspective. */
    perspective?: string;
    /** Name of this section. If not supplied, only a separator will be shown above the section. */
    name?: string;
    /** Insert this item before the item referenced here. For arrays, the first one found in order is used. */
    insertBefore?: string | string[];
    /** Insert this item after the item referenced here. For arrays, the first one found in order is used. `insertBefore` takes precedence. */
    insertAfter?: string | string[];
  }
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+  export type NavItem = {
```
christianvogt, line 4: We need a way to supply `data-*` properties to nav items.
Perhaps add `dataAttributes?: {[key: string]: string}`. Where each attribute gets the prefix `data-`
eg.
```
dataAttributes: {
  ['test-id']: 'my-id',
}
```

Results in: `<NavItem data-test-id="my-id"`

```
    /** Adds `data-` attributes to the DOM. Each key will receive the `data-` prefix.  */
    dataAttributes?: {[key: string]: string},
```

glekner:
@christianvogt applied suggestions

christianvogt:
@glekner needs rebase

glekner:
rebased @spadgett @christianvogt

frontend/packages/console-dynamic-plugin-sdk/src/extensions/navigation.ts
```diff
@@ -0,0 +1,82 @@
...
+// Type guards
+
+export const isSeparator = (e: Extension): e is Separator =>
+  e.type === 'console.navigation/separator';
+export const isHrefNavItem = (e: Extension): e is HrefNavItem =>
+  e.type === 'console.navigation/href';
+export const isResourceNSNavItem = (e: Extension): e is ResourceNSNavItem =>
+  e.type === 'console.navigation/resource-ns';
+export const isResourceClusterNavItem = (e: Extension): e is ResourceClusterNavItem =>
+  e.type === 'console.navigation/resource-cluster-ns';
```
christianvogt, line 80: cluster resource shouldn't include namespace
```suggestion
  e.type === 'console.navigation/resource-cluster';
```

glekner:
fixed @christianvogt

PR #8175, opened 2021-02-16
CONSOLE-2380: Add UI for enabling/disabling dynamic plugins during op…

rhamilto:
…erator install

Test CatalogSource image url: `quay.io/spadgett/example-registry:latest`
Test operator from above CatalogSource:  `etcd`

**Install Operator form**
If the operator is not trusted, the value defaults to `Disabled` and a warning appears when changing to `Enabled`
Screen Shot 2021-02-17 at 9 50 46 AM
Screen Shot 2021-02-17 at 9 51 01 AM

**Installed Operators list**
If an operator has a plugin available and it is not enabled, a link with popover is added to the status column
Screen Shot 2021-02-16 at 10 15 52 AM

**Operator Details**
Screen Shot 2021-02-17 at 9 57 03 AM

If the operator is not trusted, the value defaults to `Disabled` and a warning appears when changing to `Enabled`
Screen Shot 2021-02-17 at 9 57 14 AM
Screen Shot 2021-02-17 at 9 57 24 AM

If there are more than one plugins associated with the operator, display the operator name to delineate:
Screen Shot 2021-02-17 at 11 02 50 AM
Screen Shot 2021-02-17 at 10 44 20 AM
Screen Shot 2021-02-17 at 10 50 44 AM


frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -334,6 +373,16 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
         await k8sCreate(OperatorGroupModel, operatorGroup);
       }
       await k8sCreate(SubscriptionModel, subscription);
+      if (!_.isEmpty(enabledPlugins)) {
+        const previousPlugins: string[] = consoleOperator?.spec?.plugins || [];
+        await k8sPatch(ConsoleModel, consoleOperator, [
+          {
+            path: '/spec/plugins',
+            value: _.union(previousPlugins, enabledPlugins),
+            op: 'add',
```
spadgett, line 382: I'm debating if we need to remove existing enabled plugins if we are reinstalling an operator and the user picks "Disabled."

frontend/packages/operator-lifecycle-manager/src/const.ts
```diff
@@ -8,3 +8,4 @@ export const OPERATOR_TYPE_ANNOTATION = 'operators.operatorframework.io/operator
 export const NON_STANDALONE_ANNOTATION_VALUE = 'non-standalone';
 export const INTERNAL_OBJECTS_ANNOTATION = 'operators.operatorframework.io/internal-objects';
 export const DEFAULT_SOURCE_NAMESPACE = 'openshift-marketplace';
+export const CLUSTER_SERVICE_VERSION_PLUGINS_ANNOTATION = 'console.openshift.io/plugins';
```
spadgett, line 11: We might call this `OPERATOR_PLUGINS_ANNOTATION` to match the CSV annotation names above.

frontend/packages/operator-lifecycle-manager/src/utils/consolePluginFormGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <div>
+              {csvPluginsCount > 1 && (
+                <legend className="co-legend co-legend--nested text-muted">{plugin}</legend>
+              )}
+              <ConsolePluginRadioInputs
+                name={plugin}
+                enabled={enabledPlugins.includes(plugin)}
+                onChange={(enabled: boolean) => setPluginEnabled(plugin, enabled)}
+              />
+              <ConsolePluginWarning
```
spadgett, line 48: I'd suggest we have one warning for all plugins. (You'd just need to check `enabled={enabledPlugins?.length}` in that case.

spadgett (commented):
Thanks @rhamilto

rhamilto:
@spadgett, I believe I've addressed all your comments with the exception of https://github.com/openshift/console/pull/8175#discussion_r577741880 since you are debating it.  See https://github.com/openshift/console/pull/8175/commits/e597d1e258bd2e037a2e1231d58bf2d151c4a3da

frontend/packages/operator-lifecycle-manager/src/const.ts
```diff
@@ -8,3 +8,5 @@ export const OPERATOR_TYPE_ANNOTATION = 'operators.operatorframework.io/operator
 export const NON_STANDALONE_ANNOTATION_VALUE = 'non-standalone';
 export const INTERNAL_OBJECTS_ANNOTATION = 'operators.operatorframework.io/internal-objects';
 export const DEFAULT_SOURCE_NAMESPACE = 'openshift-marketplace';
+export const OPERATOR_PLUGINS_ANNOTATION = 'console.openshift.io/plugins';
+export const CONSOLE_OPERATOR_CONFIG_NAME = 'cluster';
```
spadgett, line 12: We might move this elsewhere since it's not specific to OLM. Maybe

https://github.com/openshift/console/blob/master/frontend/packages/console-shared/src/constants/resource.ts

frontend/packages/operator-lifecycle-manager/src/utils.ts
```diff
@@ -1,6 +1,44 @@
+import * as _ from 'lodash';
 import { parseJSONAnnotation } from '@console/shared/src/utils/annotations';
-import { ObjectMetadata } from '@console/internal/module/k8s';
-import { INTERNAL_OBJECTS_ANNOTATION } from './const';
+import { K8sResourceKind, ObjectMetadata, Patch } from '@console/internal/module/k8s';
+import { INTERNAL_OBJECTS_ANNOTATION, OPERATOR_PLUGINS_ANNOTATION } from './const';
+
+export const getCatalogSourceIsTrusted = (catalogSource: string): boolean =>
```
spadgett, line 6: I'd suggest

```suggestion
export const isCatalogSourceTrusted = (catalogSource: string): boolean =>
```

frontend/packages/operator-lifecycle-manager/src/utils/consolePluginFormGroup.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
```
spadgett, line 1: nit: In this package, we've been kebab casing the file names (e.g., console-plugin-form-group.tsx)

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
spadgett: > I'm debating if we need to remove existing enabled plugins if we are reinstalling an operator and the user picks "Disabled."

I think the logic for this is pretty simple and we should go ahead and do it. Something like

```ts
value: [
  ...previousPlugins.filter((plugin: string) => !csvPlugins.includes(plugin)),
  ...enabledPlugins,
],
```

spadgett:
Looks good overall, just a few small comments. Thanks @rhamilto !

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -251,6 +263,106 @@ const ManagedNamespaces: React.FC<ManagedNamespacesProps> = ({ obj }) => {
...
+
+  return (
+    <>
+      {consoleOperatorConfig && canPatchConsoleOperatorConfig && (
+        <dl className="co-clusterserviceversion-details__field">
+          <dt>{t('olm~Console UI extension', { count: csvPluginsCount })}</dt>
+          {csvPlugins.map((plugin) => (
+            <dd key={plugin} className="co-clusterserviceversion-details__field-description">
+              {csvPluginsCount > 1 && (
+                <strong className="text-muted">{t('olm~{{plugin}}:', { plugin })} </strong>
```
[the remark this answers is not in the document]
rhamilto: Even though it includes a `:`?

frontend/packages/operator-lifecycle-manager/src/components/modals/console-plugin-modal.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const patch = getPluginPatch(consoleOperatorConfig, plugin, enabled);
+    const promise = k8sPatch(ConsoleOperatorConfigModel, consoleOperatorConfig, [patch]);
+    handlePromise(promise, close);
+  };
+
+  return (
+    <form onSubmit={submit} name="form" className="modal-content">
+      <ModalTitle>
+        {t('olm~Console UI extension')}
+        {csvPluginsCount > 1 && <>{t('olm~ - {{plugin}}', { plugin })}</>}
```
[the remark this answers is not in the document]
rhamilto: Even though it includes a `-`?

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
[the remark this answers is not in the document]
rhamilto: That is, does it make sense to externalize just the colon without the context?

frontend/packages/operator-lifecycle-manager/src/components/modals/console-plugin-modal.tsx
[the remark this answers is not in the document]
spadgett: Hm, maybe we want two separate messages:

```
olm~Console UI extension
```

and

```
olm~Console UI extension - {{plugin}}
```

That way the translators have the full context if we think they might need to change punctuation.

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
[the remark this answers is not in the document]
spadgett: Hm, you might be right that the punctuation is slightly different in different languages. We can leave it.

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
rhamilto: This does not work as anticipated since `op: 'add'`.  I think we need to `op: 'remove'` as a separate condition?

spadgett: Hm, what happens? I'd think we do want add, which works as either add or replace depending on whether the array exists. Here we're replacing the entire array. But I haven't tested.

rhamilto: I think https://github.com/openshift/console/pull/8175/commits/f4db0cb36e51b77e875f3a77d4b3c262eee52aeb is what we want, @spadgett?

rhamilto: Updated in be8d619ae2a06f013d786446acfe0fd491433c35

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
```diff
@@ -334,6 +373,23 @@ export const OperatorHubSubscribeForm: React.FC<OperatorHubSubscribeFormProps> =
...
+        ...enabledPlugins,
+      ];
+      if (!_.isEqual(previousPlugins.sort(), updatedPlugins.sort())) {
+        await k8sPatch(ConsoleOperatorConfigModel, consoleOperatorConfig, [
+          {
+            path: '/spec/plugins',
+            value: [
+              ...previousPlugins.filter((plugin: string) => !csvPlugins.includes(plugin)),
+              ...enabledPlugins,
+            ],
```
spadgett, line 388:
```suggestion
            value: updatedPlugins,
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -251,6 +260,106 @@ const ManagedNamespaces: React.FC<ManagedNamespacesProps> = ({ obj }) => {
...
+  const { t } = useTranslation();
+  const canPatchConsoleOperatorConfig = useAccessReview({
+    group: ConsoleOperatorConfigModel.apiGroup,
+    resource: ConsoleOperatorConfigModel.plural,
+    verb: 'patch',
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+  const aPluginIsDisabled =
+    !consoleOperatorConfig?.spec?.plugins?.length ||
+    csvPlugins.some((plugin) => isPluginEnabled(consoleOperatorConfig, plugin) === false);
```
spadgett, line 334: Can this simply be...?

```suggestion
    csvPlugins.some((plugin) => !isPluginEnabled(consoleOperatorConfig, plugin));
```

I assume `isPluginEnabled` always returns a boolean and never null.

spadgett (commented):
Thanks, @rhamilto. Looks good

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx
rhamilto: Done

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
rhamilto: Yes.  Good catch.

rhamilto:
Thanks, @spadgett.  Updated, rebased, squashed.

spadgett (approved):
/lgtm

spadgett:
The failing test is removed in #8145. Let's wait for that PR.

/hold

itsptk:
@spadgett You mentioned today that these resources are really called "console plugins." I was wondering if you thought the "Console UI extension" term used in this UI was perhaps ambiguous? Maybe it really should have been "Console plugin" or maybe even a mix of actual resource and its purpose, like "Console UI plugin"? cc @alimobrem

PR #8182, opened 2021-02-17
Bug 1929769: Switch to correct perspective when loading non-shared routes

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5283

**Analysis / Root cause**: 
Routes specific to a perspective are not rendered if the currently active perspective does not match, resulting in a 404 error.

**Solution Description**: 
Render the inactive Routes after the active routes, and switch to the correct perspective if they are loaded.

https://user-images.githubusercontent.com/20013884/108228442-df2f0100-7164-11eb-99cd-3b2c2b148c98.mp4

Unchanged.

**Test setup:**
1. Needs two browser tabs, both with open developer perspective
2. Tab 1: Open project from Sidebar, for example /project-details/ns/openshift-marketplace
3. Tab 2: Switch to Admin perspective
4. Tab 1: Browser reload via F5 / cmd-r

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/public/components/app-contents.tsx
+++ b/frontend/public/components/app-contents.tsx
@@ -80,21 +80,46 @@ const LazyRoute = (props) => (
   />
 );
 
-const getPluginPageRoutes = (activePerspective: string, routePages: RoutePage[]) =>
-  routePages.map((r) => {
-    if (r.properties.perspective && r.properties.perspective !== activePerspective) {
-      return null;
-    }
-    const Component = r.properties.loader ? LazyRoute : Route;
-    return <Component {...r.properties} key={Array.from(r.properties.path).join(',')} />;
-  });
+const getPluginPageRoutes = (
+  activePerspective: string,
+  setActivePerspective: (perspective: string) => void,
+  routePages: RoutePage[],
+) => {
+  const activeRoutes = routePages
+    .filter((r) => !r.properties.perspective || r.properties.perspective === activePerspective)
+    .map((r) => {
+      const Component = r.properties.loader ? LazyRoute : Route;
+      return <Component {...r.properties} key={Array.from(r.properties.path).join(',')} />;
+    });
+
+  const inactiveRoutes = routePages
+    .filter((r) => r.properties.perspective && r.properties.perspective !== activePerspective)
+    .map((r) => {
+      const key = Array.from(r.properties.path)
+        .concat([r.properties.perspective])
+        .join(',');
+
+      return (
+        <Route
+          {...r.properties}
+          key={key}
+          component={() => {
+            React.useEffect(() => setActivePerspective(r.properties.perspective));
+            return null;
+          }}
+        />
+      );
+    });
+
+  return [activeRoutes, inactiveRoutes];
+};
 
 const AppContents: React.FC<{}> = () => {
-  const [activePerspective] = useActivePerspective();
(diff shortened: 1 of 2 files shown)


dtaylor113:
Hi @rottencandy, just FYI on some similar work.  This sprint we are looking to add `&perspective=developer` and `&perspective=administrator` query params to OpenShift console urls which will also set the correct perspective.  This is for Adv. Cluster Management link-n-lauch-2-console integration work.  
cc @spadgett

christianvogt:
@dtaylor113 very interesting...
We originally had the dev perspective in the URL path and removed it. The hope was to never have perspective specific URLs but the UX didn't hold true to this over time.

Having the perspective in the URL has come up since but we keep dismissing it in hopes that we get to a point where we have a customized nav and dev perspective is nothing more than an opinionated navigation. But all pages are shared across the app, regardless of perspective.

cc @serenamarie125

frontend/public/components/app-contents.tsx
```diff
@@ -80,21 +80,57 @@ const LazyRoute = (props) => (
...
+    .filter((r) => !r.properties.perspective || r.properties.perspective === activePerspective)
+    .map((r) => {
+      const Component = r.properties.loader ? LazyRoute : Route;
+      return <Component {...r.properties} key={Array.from(r.properties.path).join(',')} />;
+    });
+
+  const inactiveRoutes = routePages
+    .filter((r) => r.properties.perspective && r.properties.perspective !== activePerspective)
+    .map((r) => {
+      const key = Array.from(r.properties.path).join(',');
```
christianvogt, line 98: Include the perspective as part of the `key`.

frontend/public/components/app-contents.tsx
```diff
@@ -80,21 +80,57 @@ const LazyRoute = (props) => (
...
+            return r.properties.loader();
+          }}
+        />
+      ) : (
+        <Route
+          {...r.properties}
+          key={key}
+          component={undefined}
+          render={(componentProps) => {
+            setActivePerspective(r.properties.perspective);
```
christianvogt, line 114: You should set the perspective inside an effect.

frontend/public/components/app-contents.tsx
```diff
@@ -80,21 +80,57 @@ const LazyRoute = (props) => (
...
+              <r.properties.component {...componentProps} />
+            ) : (
+              r.properties.render(componentProps)
+            );
+          }}
+        />
+      );
+    });
+
+  return [...activeRoutes, ...inactiveRoutes];
```
christianvogt, line 125: You must render all `inactiveRoutes` after all dynamic and static routes but before the fallback route.

This will ensure that we aren't pushing the user into the wrong perspective even though there's a static route available.

frontend/public/components/app-contents.tsx
rottencandy: Added.

frontend/public/components/app-contents.tsx
rottencandy: Updated.

frontend/public/components/app-contents.tsx
rottencandy: Updated.

PR #8184, opened 2021-02-17
Bug 1929803: Hide pipeline section in container image edit flow

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5538

**Analysis / Root cause**: 
Pipeline section is shown in container image edit flow.

**Solution Description**: 
PIpeline section should be hidden in container image edit flow.

asdf

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
+++ b/frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
@@ -56,10 +56,12 @@ const EditApplicationForm: React.FC<FormikProps<FormikValues> & EditApplicationF
         {createFlowType === CreateApplicationFlow.Container && <ImageSearchSection />}
         {createFlowType === CreateApplicationFlow.Container && <IconSection />}
         <AppSection project={values.project} />
-        <PipelineSection
-          builderImages={builderImages}
-          existingPipeline={appResources?.pipeline?.data}
-        />
+        {createFlowType !== CreateApplicationFlow.Container && (
+          <PipelineSection
+            builderImages={builderImages}
+            existingPipeline={appResources?.pipeline?.data}
+          />
+        )}
         <AdvancedSection values={values} appResources={appResources} />
         <FormFooter
           handleReset={handleReset}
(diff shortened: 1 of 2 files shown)


frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
```diff
@@ -56,10 +56,12 @@ const EditApplicationForm: React.FC<FormikProps<FormikValues> & EditApplicationF
         {createFlowType === CreateApplicationFlow.Container && <ImageSearchSection />}
         {createFlowType === CreateApplicationFlow.Container && <IconSection />}
         <AppSection project={values.project} />
-        <PipelineSection
-          builderImages={builderImages}
-          existingPipeline={appResources?.pipeline?.data}
-        />
+        {createFlowType !== CreateApplicationFlow.Container && (
```
andrewballantyne, line 59: We need tests to make sure the expected items are what we expect.

This may be one of those few cases where we test static (as well as conditional) renderings.

rottencandy: Added tests.

frontend/packages/dev-console/src/components/edit-application/__tests__/EditApplicationForm.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+  };
+});
+
+describe('EditApplicationForm', () => {
+  const componentProps = {
+    ...formikFormProps,
+    appResources: {},
+  };
+
+  it('should hide git & pipeline sections for container edit form', () => {
```
divyanshiGupta, line 29: Can you break this into two separate tests? This is for checking if the git and pipeline sections are hidden in container edit form so I think checking if these sections are present in git and docker forms should be done in a separate test. Similarly you can do this for other tests. Because I think what you are testing should match the test case as well as that way it becomes easier to debug if something breaks.

rottencandy: Updated.

divyanshiGupta (approved):
Verified locally, works fine.
/lgtm

invincibleJai:
/approve

Verified, works as expected!!

PR #8189, opened 2021-02-18
move route section 'after create a route to the application' option

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-1492

**Analysis / Root cause**: 
The empty routing area looks odd when the `create a route to the application` checkbox is unchecked. Seems to me like it would be better to disable the form fields or at least show a message that routing hasn't been checked.

**Solution Description**: 
move route section after  'after create a route to the application' option

image
image
image
image

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/import/route/RouteSection.tsx
```diff
@@ -1,25 +1,13 @@
...
 import CreateRoute from './CreateRoute';
 import SecureRoute from './SecureRoute';
 
-interface RouteSectionProps {
-  route: RouteData;
-}
-
-const RouteSection: React.FC<RouteSectionProps> = ({ route }) => {
-  const { t } = useTranslation();
+const RouteSection: React.FC = () => {
```
sahil143, line 5: @vikram-raj I dont think we need this wrapper component anymore, You can directly use CreateRoute, SecureRoute in the AdvancedRouteOption component

frontend/packages/dev-console/src/components/import/route/__tests__/AdvancedRouteOptions.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+  beforeEach(() => {
+    props = {
+      canCreateRoute: true,
+      resources: Resources.OpenShift,
+    };
+  });
+
+  it('Render AdvancedRoutingOptions', () => {
+    const component = shallow(<AdvancedRouteOptions {...props} />);
+    expect(component.exists()).toBe(true);
```
sahil143, line 29: component will always exist. If you want you can check for `isEmptyRender()`

frontend/packages/dev-console/src/components/import/route/RouteSection.tsx
vikram-raj: Removed this wrapper component.

sahil143 (approved):
/lgtm

PR #8191, opened 2021-02-18
Remove dead code related to old catalog extension

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5509

**Analysis / Root cause**: The new catalog uses new dynamic extensions now, so we need to remove old extension usage.

**Solution Description**: Remove dead code and old extension definition.


invincibleJai:
/lgtm

verified the changes

PR #8193, opened 2021-02-18
Bug 1917241: Format UTC date in tooltips

rebeccaalpert:
Formatted UTC date in tooltip to be human-readable. I opened a story to track the suggestion to make UTC vs. local date a user preference: https://issues.redhat.com/browse/CONSOLE-2765.

En-gb date localization appears to have been broken by a recent i18n change; I'll open a separate bug for that.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1917241.

Screen Shot 2021-02-18 at 4 27 32 PM

Screen Shot 2021-02-18 at 4 27 05 PM

Screen Shot 2021-02-18 at 4 27 53 PM

CC @openshift/team-ux-review.

### DIFF
--- a/frontend/public/components/utils/timestamp.tsx
+++ b/frontend/public/components/utils/timestamp.tsx
@@ -4,6 +4,7 @@ import { Tooltip } from '@patternfly/react-core';
 import * as classNames from 'classnames';
 import { GlobeAmericasIcon } from '@patternfly/react-icons';
 import { useTranslation } from 'react-i18next';
+import * as moment from 'moment';
 
 import * as dateTime from './datetime';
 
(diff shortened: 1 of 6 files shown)


frontend/public/components/utils/timestamp.tsx
```diff
@@ -54,7 +55,11 @@ export const Timestamp = connect(nowStateToProps)((props: TimestampProps) => {
       <Tooltip
         content={[
           <span className="co-nowrap" key="co-timestamp">
-            {mdate.toISOString()}
+            {t('{{date, MMM D, h:mm a z}}', {
```
spadgett, line 58: I'm tempted to say we should just use

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

And not rely on moment.js for dates. The problem is that Safari doesn't support date/time formats.

Not something to tackle in this PR since we are already using moment.js.

spadgett (approved):
/lgtm

kyoto:
I notice that the Japanese date format in the screenshot is somewhat unnatural. Rather than `2月 18, 11:48 午前` maybe it should be changed to `2月18日 午前11:48`?

rebeccaalpert:
Thanks @kyoto; I'll open a separate bug for that!

PR #8194, opened 2021-02-19
Update the Task Results error message to be an empty state message

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5542

**Description**: 
Update the Task Results error message to be an empty state message

image
image

image

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/shared/results/ResultsList.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/shared/results/ResultsList.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { Alert } from '@patternfly/react-core';
+import { Bullseye, EmptyState, EmptyStateBody, EmptyStateVariant } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 import { SectionHeading } from '@console/internal/components/utils';
 import { TektonResultsRun } from '../../../types';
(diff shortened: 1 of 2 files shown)


andrewballantyne (approved):

jerolimov:
@andrewballantyne @vikram-raj /cc @siamaksade 
I tested this now also in the case where some tasks are successful and only one task fails. Are you sure that the users could not profit from the results in this case where tasks depends on each other?

Example Pipeline:

```yaml
kind: Pipeline
metadata:
  name: sum-three-pipeline-with-error
spec:
  params:
    - description: the first operand
      name: first
      type: string
    - description: the second operand
      name: second
      type: string
    - description: the third operand
      name: third
      type: string
  results:
    - description: the sum of all three operands
      name: sum
      value: $(tasks.second-add.results.sum)
    - description: the sum of first two operands
      name: partial-sum
      value: $(tasks.first-add.results.sum)
    - description: the sum of everything
      name: all-sum
      value: $(tasks.second-add.results.sum)-$(tasks.first-add.results.sum)
  tasks:
    - name: first-add
      params:
        - name: first
          value: $(params.first)
        - name: second
          value: $(params.second)
      taskRef:
        kind: Task
        name: add-task
    - name: second-add
      params:
        - name: first
          value: $(tasks.first-add.results.sum)
        - name: second
          value: $(params.third)
      taskRef:
        kind: Task
        name: add-task
    - name: sometimes-failing
      taskSpec:
        metadata: {}
        steps:
          - image: node
            name: sometimes-failing
            resources: {}
            script: |
              #!/bin/bash
              exit 1
```

Example PipelineRun:

```yaml
kind: PipelineRun
metadata:
  name: sum-three-pipeline-with-error-2wtw6n
  labels:
    tekton.dev/pipeline: sum-three-pipeline-with-error
spec:
  params:
    - name: first
      value: '1'
    - name: second
      value: '23'
    - name: third
      value: '3123'
  pipelineRef:
    name: sum-three-pipeline-with-error
```

jerolimov:
Everything else looks fine here. The PR would get my lgtm and the UX is much better with this EmptyState component instead of an Alert.

andrewballantyne:
@jerolimov A valid point, but I think we would need to parse the data to find out what is and isn't real data before rendering. Not ideal. Either way, it's not for this PR. Please bring up this conversation on Slack or put it on the agenda for the meeting Pipeline tomorrow.

/retest
/lgtm

PR #8195, opened 2021-02-19
add text area for multiline task param

debsmita1:
**JIRA story:**
https://issues.redhat.com/browse/ODC-4279

**Solution Description:**
Replace `TextInput` component with `TextArea`

**GIF:**
multiline

**Browser conformance:**
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/temp-utils.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/temp-utils.tsx
@@ -5,7 +5,7 @@ import { global_disabled_color_200 as disabledColor } from '@patternfly/react-to
 import {
   Flex,
   FlexItem,
-  TextInput,
+  TextArea,
   Tooltip,
   ValidatedOptions,
   FormGroup,
@@ -30,8 +30,11 @@ export const StringParam: React.FC<ParameterProps> = (props) => {
   const { currentValue, defaultValue, isValid, name, onChange, setDirty } = props;
 
   return (
-    <TextInput
+    <TextArea
       id={name}
+      rows={1}
+      resizeOrientation="vertical"
+      style={{ minHeight: 33 }}
       validated={isValid ? 'default' : 'error'}
       isRequired={!defaultValue}
       onBlur={() => setDirty(true)}


karthikjeeyar:
Text area takes more than one line on the inital load and it doesn't look compact. 2nd point in the story's  AC says it should start with single row. 
image

use **` rows={1}`**  in the text area, so that we get a compact field like below.
image

cc: @bgliwa01

bgliwa01 (approved):
lgtm!

karthikjeeyar:
@debsmita1 I think while adding the `rows` , you have removed `minHeight:30`, you need to add that back or else the field will be minimizable even below the one line of vertical space.
minHeight

debsmita1:
> @debsmita1 I think while adding the `rows` , you have removed `minHeight:30`, you need to add that back or else the field will be minimizable even below the one line of vertical space.
> minHeight

yes, I am unsure if UX wants that
@bgliwa01 should we add a min-height to the fields ?

andrewballantyne:
My recommendation is yes, there shouldn't be any need to go below a singular line-height.

debsmita1:
I have set the min height to 30px. Updated the GIF @karthikjeeyar @andrewballantyne  @bgliwa01 
Please let me know if it looks good

karthikjeeyar:
Thanks @debsmita1, changes looks good to me. 
/lgtm

bgliwa01:
the min height works great! Thanks!

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/temp-utils.tsx
```diff
@@ -30,8 +30,10 @@ export const StringParam: React.FC<ParameterProps> = (props) => {
   const { currentValue, defaultValue, isValid, name, onChange, setDirty } = props;
 
   return (
-    <TextInput
+    <TextArea
       id={name}
+      rows={1}
+      style={{ minHeight: 30, resize: 'vertical' }}
```
andrewballantyne, line 36: Hmmm... 30px is weird. We should probably think of a more deterministic value... but after a few tries off the top of my head I'm at a bit of a loss what that would be...

- Font-size is 14px (aka 1em)
- The padding + borders gives us 12px
- Adding those two together, gets us 26 -- which feels too tight 😞 
- LineHeight is set to 21px, so instead of font-size we use that, that's 33px -- which is about what you got.

Doing a calc of all the Patternfly variables to get to the 33px turns out to be an ugly slapping together of things. I think we can probably go with the static value for now -- unlikely to cause us too much headache I would think. Sizes tend not to change too much in this area. If they say, change padding, we'll end up in an awkward height situation.

andrewballantyne, line 36: My other comment is more-or-less for posterity if we need to come back to this.

This comment however is a request for a change.

```suggestion
      style={{ minHeight: 30 }}
      resizeOrientation="vertical"
```
There is a prop for the resize style. Best to use that.

andrewballantyne (changes requested):
Glad it was a simple change, please use the prop instead of a style though.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/temp-utils.tsx
debsmita1: updated the height to 33px

andrewballantyne (approved):
/lgtm

PR #8198, opened 2021-02-19
fix validation in number type input field

nemesis09:
**Fixes:**
https://issues.redhat.com/browse/ODC-3989

**Analysis / Root cause:**
Input fields with **type="number"** allows character input in Firefox.
For an input field of type number -
- Chromium allows only numbers to be entered in the field as part of its browser validation, i.e. any non numeric character is rejected as invalid input and will not show up on the input field nor will it be passed down as value to any event for that field.
- Firefox also does similar browser validation for such fields and rejects non numeric characters as invalid input and does not pass them as value to any event. However, it allows the non numeric characters to show up on the input field which leads to confusion.
Since, under the hood, the characters are restricted by browser validation, formik doesn't receive values. Thus yup validation does not occur and also the form is submitted with initial values. This is confusing to users on firefox.

**Solution Description:**
Change input type from **number** to **text** to bypass the browser validation and handle the number validation in yup.

**Note:** The solution will strip the fields from user experience related to number fields, e.g. numeric keypad on mobile devices.

_Before -_
Firefox:
invalid-input_pre-change_firefox

Chrome:
invalid-input_pre-change_chrome

Edge:
invalid-input_pre-change_edge

_After_
Firefox:
invalid-input_post-change_firefox

Chrome:
invalid-input_post-change_chrome

Edge:
invalid-input_post-change_edge

**Test Coverage:**
NA

**Browser Conformance:**
- [x] Firefox
- [x] Chrome
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/health-checks/ProbeForm.tsx
+++ b/frontend/packages/dev-console/src/components/health-checks/ProbeForm.tsx
@@ -58,7 +58,7 @@ const ProbeForm: React.FC<ProbeFormProps> = ({ onSubmit, onClose, probeType }) =
         />
         {getRequestTypeForm(healthChecks?.[probeType]?.data?.requestType, probeType)}
         <InputField
-          type={TextInputTypes.number}
+          type={TextInputTypes.text}
           name={`healthChecks.${probeType}.data.failureThreshold`}
           label={t('devconsole~Failure threshold')}
           style={{ maxWidth: '100%' }}
@@ -68,7 +68,7 @@ const ProbeForm: React.FC<ProbeFormProps> = ({ onSubmit, onClose, probeType }) =
           isDisabled={viewOnly}
         />
         <InputField
-          type={TextInputTypes.number}
+          type={TextInputTypes.text}
           name={`healthChecks.${probeType}.data.successThreshold`}
           label={t('devconsole~Success threshold')}
           style={{ maxWidth: '100%' }}
@@ -78,7 +78,7 @@ const ProbeForm: React.FC<ProbeFormProps> = ({ onSubmit, onClose, probeType }) =
           isDisabled={viewOnly}
         />
         <InputGroupField
-          type={TextInputTypes.number}
+          type={TextInputTypes.text}
           name={`healthChecks.${probeType}.data.initialDelaySeconds`}
           label={t('devconsole~Initial delay')}
           helpText={t(
@@ -89,7 +89,7 @@ const ProbeForm: React.FC<ProbeFormProps> = ({ onSubmit, onClose, probeType }) =
           isDisabled={viewOnly}
         />
         <InputGroupField
-          type={TextInputTypes.number}
+          type={TextInputTypes.text}
           name={`healthChecks.${probeType}.data.periodSeconds`}
           label={t('devconsole~Period')}
           helpText={t('devconsole~How often to perform the probe.')}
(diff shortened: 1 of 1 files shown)


nemesis09:
cc @beaumorley @openshift/team-devconsole-ux

beaumorley:
@nemesis09 Thanks for including UX. I'm wondering if there is anywhere else in the console where we have this issue? It seems like a common scenario (number field that needs validation on submit). While I agree that having the form submit with the user thinking their value went through is a worse case, it does not seem optimal for a string field to be used since, like you said, the number pad will not be exposed on certain devices. I'd like to hear @andrewballantyne take on this.

nemesis09:
> @nemesis09 Thanks for including UX. I'm wondering if there is anywhere else in the console where we have this issue? It seems like a common scenario (number field that needs validation on submit). While I agree that having the form submit with the user thinking their value went through is a worse case, it does not seem optimal for a string field to be used since, like you said, the number pad will not be exposed on certain devices. I'd like to hear @andrewballantyne take on this.

I could find only one instance of **input type number**  other than this component in console, i.e **TrafficSplittingModal**
TrafficSplittingModal does not handle input validations as they are handled by the API. In isolation from the API error, this issue could be reproduced in the modal as well.

vikram-raj:
@nemesis09 We have the same issue with `Resource limit` in the advanced section of import flow. In Chrome resource limit field is not accept the string value but in firefox, it accepts the string.

nemesis09:
> @nemesis09 We have the same issue with `Resource limit` in the advanced section of import flow. In Chrome resource limit field is not accept the string value but in firefox, it accepts the string.

Thanks @vikram-raj i found couple of more number inputs as well e.g NumberWidget, createRoute

andrewballantyne:
> @nemesis09 Thanks for including UX. I'm wondering if there is anywhere else in the console where we have this issue? It seems like a common scenario (number field that needs validation on submit). While I agree that having the form submit with the user thinking their value went through is a worse case, it does not seem optimal for a string field to be used since, like you said, the number pad will not be exposed on certain devices. I'd like to hear @andrewballantyne take on this.

I discussed the solution with @nemesis09 before he posted this PR. I am unsure if we care a ton about the number field browser magic. It does indeed give a number pad interface to the user when they are on mobile devices, but I think we need to accept that's a loss we can live with to get immediate feedback when the user violates the validation lib we are using. Normal text fields still possess the number keys, so it's not a blocker. 

This is also appears to be a browser issue which may or may not be fixed in the future. The field as a number field does not emit an event to the validation library we are using when you type letters into a number field thus preventing our library from error'ing out on the result. 

@beaumorley I do not currently know what our percentages are around mobile support -- we seem to apply some effort in this area but it's not the chief goal of our app. Do you know what kind of numbers would support the usage of mobile devices? ie Tablets/Phones.

cc @christianvogt

andrewballantyne:
@nemesis09 Could you test each browser and let us know what the effect is without and with your change for each of the [supported browsers](https://github.com/openshift/console#supported-browsers)?

beaumorley:
Thanks for looking into this all. I agree that the solution is an improvement. Just wanted to make sure it was the best option.  I agree with what is said above. LGTM.

nemesis09:
> @nemesis09 Could you test each browser and let us know what the effect is without and with your change for each of the [supported browsers](https://github.com/openshift/console#supported-browsers)?

I've updated the description with the results from the following browsers-
- Firefox
- Safari
- Chrome
all the results are for invalid entries.
Note: since chrome and edge doesn't allow any invalid input to be shown in the field, so they will just show empty fields in the gif making it seem like nothing is being typed in the field.

christianvogt:
Not a big fan of the error message. But Beau is ok with the change
/approve
/lgtm

PR #8199, opened 2021-02-19
CONSOLE-2506: Added 'Advanced Cluster Management' to Perspective Switcher

dtaylor113:
1. Added 'Advanced Cluster Management' to Perspective Switcher
2. Filtered ACM ConsoleLink from appearing in masthead Application Launcher
3. Added support for url query params: `?perspective=dev` & `?perspective=administrator`
4. Added new ACM icon
5. Added Cypress e2e tests

image


dtaylor113:
Hi @spadgett, do we need the`...&perspective=developer` or `...&perspective=administrator` query params if this PR [Switch to correct perspective when loading non-shared routes](https://github.com/openshift/console/pull/8182) auto-selects the Perspective based on url?

frontend/public/components/nav/nav-header.tsx
```diff
@@ -1,25 +1,64 @@
...
+  if (flags[FLAGS.CONSOLE_LINK]) {
+    resources.push({
+      kind: referenceForModel(ConsoleLinkModel),
+      isList: true,
+      prop: 'consoleLinks',
+    });
+  }
+
+  return (
+    <Firehose resources={resources}>
```
spadgett, line 39: I would use `useWatchK8sList` instead. That way you don't need the wrapper component and rendering of the perspective switcher won't block while we wait for the links.

frontend/public/components/nav/nav-header.tsx
```diff
@@ -1,25 +1,64 @@
...
+    </Firehose>
+  );
+};
+
+const getACMConsoleLink = (links): K8sResourceKind => {
+  const acmLink = _.find(
+    links,
+    (link) => link.spec.location === 'ApplicationMenu' && link.metadata.name === ACM_LINK_ID,
+  );
+  return acmLink ? acmLink : null;
```
spadgett, line 50: No need for the `acmLink ? acmLink : null` check, just return the results of `file`. Prefer native `Array.find` to Lodash.

```suggestion
  return links?.find((link) => link.metadata.name === ACM_LINK_ID && link.spec.location === 'ApplicationMenu');
```

frontend/public/components/nav/nav-header.tsx
```diff
@@ -1,25 +1,64 @@
...
+  }
+
+  return (
+    <Firehose resources={resources}>
+      <NavHeaderContents onPerspectiveSelected={onPerspectiveSelected} flags={flags} />;
+    </Firehose>
+  );
+};
+
+const getACMConsoleLink = (links): K8sResourceKind => {
```
spadgett, line 45: Give the argument a type. (We should probably add a `ConsoleLinkKind` type, but not necessary for this PR.)

```suggestion
const getACMConsoleLink = (links: K8sResourceKind[]): K8sResourceKind => {
```

frontend/public/components/nav/nav-header.tsx
```diff
@@ -29,7 +68,12 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
       event.preventDefault();
       if (perspective.properties.id !== activePerspective) {
         setActivePerspective(perspective.properties.id);
-        history.push(perspective.properties.getLandingPageURL(flags));
+        if (perspective.properties.id !== ACM_LINK_ID) {
+          history.push(perspective.properties.getLandingPageURL(flags));
+        } else {
+          // For ACM link, replace entire URL
+          window.location.href = perspective.properties.getLandingPageURL(flags);
+        }
```
spadgett, line 76: This is minor, but I'd flip the logic here. I find it a little harder to think in negatives.

```suggestion
        if (perspective.properties.id === ACM_LINK_ID) {
          // For ACM link, replace entire URL
          window.location.href = perspective.properties.getLandingPageURL(flags);
        } else {
          history.push(perspective.properties.getLandingPageURL(flags));
        }
```

frontend/public/components/nav/nav-header.tsx
```diff
@@ -55,25 +99,52 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
...
+          getImportRedirectURL: acmLink.spec.href,
+        },
+      };
+      items.push(
         <DropdownItem
-          key={nextPerspective.properties.id}
+          key={acmPerspective.properties.id}
           onClick={(event: React.MouseEvent<HTMLLinkElement>) =>
-            onPerspectiveSelect(event, nextPerspective)
+            onPerspectiveSelect(event, acmPerspective)
```
spadgett, line 134: I think it's cleaner to have a separate event handler for ACM instead of having the `if` logic above in `onPerspectiveSelect`.

frontend/public/components/nav/nav-header.tsx
```diff
@@ -55,25 +99,52 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
...
           }
-          isHovered={nextPerspective.properties.id === activePerspective}
+          isHovered={acmPerspective.properties.id === activePerspective}
           component="button"
```
spadgett, line 137: Is it possible to make this an `a` element so we don't need a click handler and the browser can just treat it as a link?

frontend/public/components/nav/nav-header.tsx
dtaylor113: > Prefer native `Array.find` to Lodash.

Isn't `links?.find(...` native?  It's not lodash's `_.find(..`  :-)

mdelder:
Re: When to remove the ACM link from the Applications list -- I would keep that for a release or two. We still have to support ACM on the current versions of OCP (4.5/4.6/4.7/etc) and those versions of OCP won't support the perspective switcher to open ACM.

spadgett:
> **2. For the second 'ask' for Phase 1,** do we need the`...&perspective=developer` or `...&perspective=administrator` query params if this PR "[Switch to correct perspective when loading non-shared routes](https://github.com/openshift/console/pull/8182)" auto-selects the Perspective based on url?

I believe it's still needed. #8182 only applies to certain routes, and none of the admin routes will change the perspective.

spadgett:
> Re: When to remove the ACM link from the Applications list -- I would keep that for a release or two. We still have to support ACM on the current versions of OCP (4.5/4.6/4.7/etc) and those versions of OCP won't support the perspective switcher to open ACM.

OCP console is using the ConsoleLink resource to discover the ACM URL anyway. So from the ACM side it would be the same -- always create the `acm-console-link` ConsoleLink -- and OCP can simply render it in a different way. It would let us remove the link from the Application Launcher in 4.8, but still have it appear in earlier OpenShift versions.

frontend/public/components/nav/nav-header.tsx
dtaylor113: fixed, switched to `useK8sWatchResource`

frontend/public/components/nav/nav-header.tsx
dtaylor113: fixed

frontend/public/components/nav/nav-header.tsx
dtaylor113: fixed

frontend/public/components/nav/nav-header.tsx
dtaylor113: fixed

frontend/public/components/nav/nav-header.tsx
dtaylor113: Hi @spadgett, you sure?  The new click handler would need everything in the original handler.  I think the duplication would be greater than the `if` logic.

frontend/public/components/nav/nav-header.tsx
dtaylor113: I made it an `a` and the cursor changed to an `I` icon, other than that behaved the same.  I believe the click handler is still needed to handle setting the active perspective and `setPerspectiveDropdownOpen(false)` plus `onPerspectiveSelected()`.  

When the `?perspective=development` query params are in place, we may be able to just append the perspective query params to the URL of a `a` component, but still would need to do the other stuff in the click handler.

dtaylor113:
> OCP console is using the ConsoleLink resource to discover the ACM URL anyway. So from the ACM side it would be the same -- always create the `acm-console-link` ConsoleLink -- and OCP can simply render it in a different way. It would let us remove the link from the Application Launcher in 4.8, but still have it appear in earlier OpenShift versions.

Removed ACM link from App. Launcher

dtaylor113:
Hi @spadgett, added processing of `?perspective=dev` and `?perspective=admin` url search query params to set active perspective dropdown.

- Verified switching from `/topology/all-namespaces?view=graph` to `/k8s/all-namespaces/pods` leaves the **Developer** perspective in place.
- Verified switching from `/topology/all-namespaces?view=graph` to `/k8s/all-namespaces/pods?perspective=admin` switches to the **Administrator** perspective.
- Verified switching from Pods to Topology pages, switches to **Developer** perspective using `perspective=dev`

dtaylor113:
@tlwu2013 or @beanh66, I selected the 'cubes' icon for ACM in the Perspective Dropdown, we need to select what icon we want to use there:  
image
- thanks

frontend/public/components/nav/nav-header.tsx
spadgett: It could simply be

```
onClick={() => window.location.href = acmLink.spec.href}
```

which is really no duplication. I think it's a little cleaner this way. You might not need to create an `acmPerspective` item this way, just add the `DropdownItem` with its own handler.

frontend/public/components/nav/nav-header.tsx
spadgett: I don't think you need to call `setPerspectiveDropdownOpen` or `onPerspectiveSelected` for the ACM link since it links away from console.

frontend/public/components/nav/nav-header.tsx
```diff
@@ -16,20 +19,50 @@ export type NavHeaderProps = {
...
 
+  const urlParams = new URLSearchParams(window.location.search);
+  const perspectiveParam = urlParams.get(PerspectiveURLParam);
+  if (
+    perspectiveParam &&
+    PerspectiveParamValues.includes(perspectiveParam) &&
+    perspectiveParam !== activePerspective
+  ) {
+    setActivePerspective(perspectiveParam);
+  }
```
spadgett, line 53: I think this is the wrong place for this logic. I'm not too familiar with the code, but probably you want it in PerspectiveDetector.tsx.

frontend/public/components/nav/nav-header.tsx
dtaylor113: Hi @spadgett, oh yeah, makes sense, duh :-) fixed

frontend/public/components/nav/nav-header.tsx
dtaylor113: thanks, fixed.

frontend/public/components/nav/nav-header.tsx
dtaylor113: Ok, moved to `frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx`

frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx
```diff
@@ -7,8 +7,21 @@ type DetectPerspectiveProps = {
   children: React.ReactNode;
 };
 
+const getPerspectiveURLParam = () => {
+  const PerspectiveURLParam = 'perspective';
+  const PerspectiveParamValues = ['dev', 'admin'];
```
spadgett, line 12: Hard-coding this list could be problematic since a dynamic plugin can contribute a new perspective we don't know about. The list you have is correct for 4.8, but might not be accurate for 4.9 and later. To correctly get the list, you need to use

```ts
  const perspectiveExtensions = useExtensions<Perspective>(isPerspective);
```

See https://github.com/spadgett/console/blob/b8f030ffe33a59c430a8c618db765cda569fce7c/frontend/public/components/nav/nav-header.tsx#L22

It's potentially a problem when showing the perspectives in ACM as well. I don't have a great answer at this time since getting the complete perspective list requires loading and running a dynamic plugin. Maybe we need something on the `ConsolePlugin` resource that indicates the plugin adds a perspective, which ACM can read.

frontend/public/components/nav/nav-header.tsx
```diff
@@ -55,25 +71,43 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
...
-          onClick={(event: React.MouseEvent<HTMLLinkElement>) =>
-            onPerspectiveSelect(event, nextPerspective)
-          }
-          isHovered={nextPerspective.properties.id === activePerspective}
+          key={ACM_LINK_ID}
+          onClick={() => {
+            window.location.href = acmLink.spec.href;
+          }}
+          isHovered={ACM_LINK_ID === activePerspective}
           component="button"
```
spadgett, line 98: Should this be an `a` element instead of a `button`?

frontend/public/components/nav/nav-header.tsx
```diff
@@ -55,25 +71,43 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
...
         >
           <Title headingLevel="h2" size="md" data-test-id="perspective-switcher-menu-option">
-            <span className="oc-nav-header__icon">{nextPerspective.properties.icon}</span>
-            {nextPerspective.properties.name}
+            <span className="oc-nav-header__icon">
+              <CubesIcon />
```
spadgett, line 102: @andybraren @mdelder What icon should we use for ACM in the perspective switcher?

frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx
spadgett: Maybe it's better to have a console backend endpoint that reads the enabled plugin manifests and returns the list of perspectives. (@dtaylor113 not something to worry about for this PR.)

frontend/public/components/nav/nav-header.tsx
dtaylor113: From an earlier thread:  _"I made it an `a` and the cursor changed from a pointer to an I icon, other than that behaved the same."_

spadgett: Right, so let's make it an `a` :) We can set the pointer style if needed to make it consistent since you wouldn't normally select the text here.

frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx
dtaylor113: Updated to use `const perspectiveExtensions = useExtensions<Perspective>(isPerspective);`

frontend/public/components/nav/nav-header.tsx
dtaylor113: ok, changed to an `a` and added `cursor: pointer`

andybraren:
> @tlwu2013 or @beanh66, I selected the 'cubes' icon for ACM in the Perspective Dropdown, we need to select what icon we want to use there:

@dtaylor113 I'd suggest `pf-icon-domain` which seems to most closely represent what ACM is/does and isn't used elsewhere in OpenShift land as far as I know. FYI @jackiwakin @imjoyjean - anything within [FontAwesome Free](https://fontawesome.com/icons?d=gallery&p=2&m=free) is also viable if we find something better.

phase-1-perspectives

dtaylor113:
> I'd suggest `pf-icon-domain`

Updated
image

jackiwakin:
@andybraren Hey, sorry, I was out yesterday on PTO!

I'm not totally sold on that icon-- since we use it in such a small space, it becomes pretty hard to read or see what it is. And the line-weight thickness doesn't match the other two icons, making it look a little different and out of place.

Here's one that I made, based off of the ACM Operator Hub icon, which could be nice to visually connect back to:
([svg to download is here](https://drive.google.com/file/d/1YzxqnXbSg8gn7mvDy2pblWcvXyUT7Jex/view?usp=sharing))
Screen Shot 2021-03-02 at 5 39 03 PM




Or, here are also some that I found on Font Awesome, that could visually represent management or connections:

https://fontawesome.com/icons/project-diagram?style=solid
Screen Shot 2021-03-02 at 4 58 07 PM


https://fontawesome.com/icons/network-wired?style=solid
Screen Shot 2021-03-02 at 4 57 49 PM

https://fontawesome.com/icons/sitemap?style=solid
Screen Shot 2021-03-02 at 5 01 04 PM

andybraren:
@jackiwakin I really like the ACM-lite logo, nice!

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -247,7 +248,12 @@ class MastheadToolbarContents_ extends React.Component {
 
   _launchActions = () => {
     const { clusterID, consoleLinks, t } = this.props;
-    const launcherItems = this._getAdditionalLinks(consoleLinks?.data, 'ApplicationMenu');
+    const launcherItems = this._getAdditionalLinks(consoleLinks?.data, 'ApplicationMenu').filter(
+      (item) => {
+        // prevent ACM (Advanced Cluster Management) from appearing in App Launcher as it has moved to perspective switcher
+        return item.metadata.name !== ACM_LINK_ID;
+      },
+    );
```
TheRealJon, line 256: `consoleLinks.data` is already filtered in `_getAdditionalLinks`, we could just add another condition to that filter instead of filtering twice:

```tsx
_getAdditionalLinks(links, type) {
    return _.sortBy(
      _.filter(links, (link) => link.spec.location === type && link.metadata.name !== ACM_LINK_ID),
      'spec.text',
    );
  }
```

TheRealJon (commented):
Could we just use the dynamic plugin API to hard code a perspective extension in the pluginStore? I would assume that ACM will eventually migrate their ConsoleLink to a perspective extension, and it will be easier for us to rip out a place hoder extension at that point, rather than revert these changes to the nav header.

n our feature detection code, we could just do an API call to check for the ACM ConsoleLink, and add a hard coded perspective extension if it exists. We still need to filter it out in the masthead app launcher, but that's a one line change.

dtaylor113:
> @jackiwakin:  Here's one that I made, based off of the ACM Operator Hub icon...

@jackiwakin & @andybraren, icon imported and updated:
 
image

frontend/public/components/nav/nav-header.tsx
```diff
@@ -16,10 +20,24 @@ export type NavHeaderProps = {
   onPerspectiveSelected: () => void;
 };
 
+export const ACM_LINK_ID = 'acm-console-link';
```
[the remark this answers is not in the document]
dtaylor113: fixed

frontend/public/components/nav/nav-header.tsx
```diff
@@ -16,10 +20,24 @@ export type NavHeaderProps = {
...
+
 const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelected, flags }) => {
   const [activePerspective, setActivePerspective] = useActivePerspective();
   const [isPerspectiveDropdownOpen, setPerspectiveDropdownOpen] = React.useState(false);
   const perspectiveExtensions = useExtensions<Perspective>(isPerspective);
+  const [consoleLinks] = useK8sWatchResource<K8sResourceKind[]>({
+    isList: true,
+    kind: referenceForModel(ConsoleLinkModel),
+    optional: true,
+  });
```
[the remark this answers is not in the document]
dtaylor113: Nice!  Wasn't aware of this usage  -thanks :-)

frontend/public/components/masthead-toolbar.jsx
dtaylor113: fixed

frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx
```diff
@@ -2,13 +2,29 @@ import * as React from 'react';
...
+  return perspectiveIDs.includes(perspectiveParam) ? perspectiveParam : null;
+};
+
 const DetectPerspective: React.FC<DetectPerspectiveProps> = ({ children }) => {
   const [activePerspective, setActivePerspective, loaded] = useValuesForPerspectiveContext();
+  const perspectiveExtensions = useExtensions<Perspective>(isPerspective);
+  const perspectiveParam = getPerspectiveURLParam(perspectiveExtensions);
+  if (perspectiveParam && perspectiveParam !== activePerspective) {
+    setActivePerspective(perspectiveParam);
+  }
```
christianvogt, line 27: Redux actions should not be executed inside of the render function. Use `useEffect`

frontend/public/components/nav/nav-header.tsx
```diff
@@ -55,25 +66,46 @@ const NavHeader_: React.FC<NavHeaderProps & StateProps> = ({ onPerspectiveSelect
...
+    if (acmLink) {
+      items.push(
         <DropdownItem
-          key={nextPerspective.properties.id}
-          onClick={(event: React.MouseEvent<HTMLLinkElement>) =>
-            onPerspectiveSelect(event, nextPerspective)
-          }
-          isHovered={nextPerspective.properties.id === activePerspective}
-          component="button"
+          key={ACM_LINK_ID}
```
christianvogt, line 87: Why is this not a `Perspective` extension like we do for admin and dev perspectives?

dtaylor113: I believe that was discussed and this direction taken.  I don't recall the details.  @spadgett ?

dtaylor113: fixed, thanks

frontend/packages/console-app/src/components/detect-perspective/DetectPerspective.tsx
dtaylor113: fixed, thanks

frontend/public/components/nav/nav-header.tsx
christianvogt: I'd like to know the reason. I think going against the extension mechanism is a bad idea.

frontend/packages/integration-tests-cypress/views/guided-tour.ts
```diff
@@ -1,5 +1,7 @@
 export const guidedTour = {
   close: () => {
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000); // wait a sec for modal if it is going to be displayed
```
TheRealJon, line 4: No change required for the sake of this PR, but we should take a closer look at why we need a wait and conditional logic here.  The app should be in a known state, meaning we should know whether or not the modal is supposed to be present before we try to close it. See https://docs.cypress.io/guides/core-concepts/conditional-testing#Element-existence for more info and suggestions.

TheRealJon: Also, check out https://docs.cypress.io/guides/core-concepts/conditional-testing#Welcome-wizard as it's almost precisely this scenario.

frontend/public/components/nav/nav-header.tsx
TheRealJon: I'll +1 @christianvogt on this. As stated in my review, I also feel like we should be using the Perspective extension mechanism here.

spadgett: @christianvogt Can you add an external link using the Perspective extension? It's not a normal perspective, but a link out to another console.

The other problem is that the URL is not static, but has to be discovered from a `ConsoleLink` resource which we're watching.

christianvogt: I see... I concede then.
Is this a short term link out? It's odd that we're mixing concepts in this dropdown.

dtaylor113: Hi @spadgett, Can we move to Perspective extension for ACM phase 2 where both OCP and ACM behind same url?  Might still have to get URL from ACM's `ConsoleLink` but it wont be an external link, right?

spadgett: The current phase 2 designs remove the perspective and instead add a cluster picker with an "All Clusters" option that goes to ACM. So this is a short-term link.

frontend/packages/integration-tests-cypress/views/guided-tour.ts
spadgett: I suggest we open a bug to track removing the wait if possible.

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -216,7 +216,7 @@ class MastheadToolbarContents_ extends React.Component {
 
   _getAdditionalLinks(links, type) {
     return _.sortBy(
-      _.filter(links, (link) => link.spec.location === type),
+      _.filter(links, (link) => link.spec.location === type && link.metadata.name !== ACM_LINK_ID),
```
spadgett, line 219: Worth adding a comment here explaining why we're not showing the ACM link.

frontend/public/components/nav/_nav-header.scss
```diff
@@ -13,6 +13,7 @@
 
   .pf-c-dropdown {
     width: 100%;
+    cursor: pointer;
```
spadgett, line 16: This often indicates an accessibility issue... Is this because the existing buttons don't use a pointer cursor or another reason?

frontend/public/components/nav/nav-header.tsx
```diff
@@ -53,25 +64,46 @@ const NavHeader: React.FC<NavHeaderProps> = ({ onPerspectiveSelected }) => {
...
         >
           <Title headingLevel="h2" size="md" data-test-id="perspective-switcher-menu-option">
-            <span className="oc-nav-header__icon">{nextPerspective.properties.icon}</span>
-            {nextPerspective.properties.name}
+            <span className="oc-nav-header__icon">
+              <img
+                src={acmIcon}
+                height="12em"
+                width="12em"
+                alt="Advanced Cluster Management icon"
```
spadgett, line 97: No need for alt text here since it's decorative. `alt=""` is acceptable in that case.

```suggestion
                alt=""
```

spadgett (commented):
We should double check that the order, label, and icon match ACM's implementation exactly.

for approvals

Thanks @dtaylor113 !

yapei:
two bugs are opened and waiting for fix

dtaylor113:
> two bugs are reported and waiting for fix

Hi @yapei, can you provide bugzilla numbers?  Are they assigned to me? -thanks

frontend/packages/integration-tests-cypress/views/guided-tour.ts
dtaylor113: I was overthinking the cypress test I created for the new url query params.  The test runs succesfully even if Guided Tour modal is open.  For now I just removed the `cy.wait()`.   Long term I think we need to use [app.waitForLoad()](https://github.com/openshift/console/blob/1ef2f7e24430a74f945b6990270dedbcd507dada/frontend/packages/dev-console/integration-tests/support/pages/app.ts#L14) and switch perspectives & clode Guided Tour like dev console [does here](https://github.com/openshift/console/blob/1ef2f7e24430a74f945b6990270dedbcd507dada/frontend/packages/dev-console/integration-tests/support/pages/app.ts#L29).

frontend/public/components/masthead-toolbar.jsx
dtaylor113: > I suggest we open a bug to track removing the [cy.]wait if possible.

I was overthinking the cypress test I created for the new url query params.  The test runs succesfully even if Guided Tour modal is open.  For now I just removed the `cy.wait()`.   Long term I think we need to use [app.waitForLoad()](https://github.com/openshift/console/blob/1ef2f7e24430a74f945b6990270dedbcd507dada/frontend/packages/dev-console/integration-tests/support/pages/app.ts#L14) and switch perspectives & close Guided Tour like dev console [does here](https://github.com/openshift/console/blob/1ef2f7e24430a74f945b6990270dedbcd507dada/frontend/packages/dev-console/integration-tests/support/pages/app.ts#L29).

> Worth adding a comment

...added

frontend/public/components/nav/nav-header.tsx
dtaylor113: fixed

frontend/public/components/nav/_nav-header.scss
dtaylor113: This came about due to an [earilier comment](https://github.com/openshift/console/pull/8199#discussion_r583010755) we discussed.

frontend/public/components/masthead-toolbar.jsx
dtaylor113: > The test runs succesfully even if Guided Tour modal is open.

Guided Tour modal continues to cause failures :disappointed: 
image
image

Adding a WIP while I investigate...

dtaylor113:
> two bugs are reported and waiting for fix

Hi @yapei just tagged [BZ: 1954538](https://bugzilla.redhat.com/show_bug.cgi?id=1954538#c7) as "not a bug".
Was there another BZ? (you mentioned 2 bugs)

yapei:
Hi @dtaylor113 , sorry about late response. This is another bug [1954566](https://bugzilla.redhat.com/show_bug.cgi?id=1954566)

yapei:
Today launched a new cluster and confirmed everything is working as expected
Screen Shot 2021-05-06 at 1 40 12 PM
Screen Shot 2021-05-06 at 1 40 17 PM

dtaylor113:
Updated switcher divider from dark to light to match PF navigational dividers:
image

frontend/public/components/nav/_nav-header.scss
spadgett: @dtaylor113 If it's a link, it should have a pointer cursor already I believe. Does it have an `href`? If not, it's an accessibility problem.

frontend/public/components/nav/nav-header.tsx
```diff
@@ -53,44 +70,63 @@ const NavHeader: React.FC<NavHeaderProps> = ({ onPerspectiveSelected }) => {
...
+        data-quickstart-id="qs-perspective-switcher"
+      >
+        <Dropdown
+          isOpen={isPerspectiveDropdownOpen}
+          toggle={renderToggle(icon, name)}
+          dropdownItems={perspectiveItems}
+          data-test-id="perspective-switcher-menu"
+        />
+      </div>
+      <NavItemSeparator key={`separator-nav-header`} inset={{ default: 'insetNone' }} />
```
spadgett, line 128:
```suggestion
      <NavItemSeparator key="separator-nav-header" inset={{ default: 'insetNone' }} />
```

spadgett (approved):
/lgtm

PR #8201, opened 2021-02-19
Bug 1929577: Fix to avoid overriding of d/dc pod template container values

divyanshiGupta:
**Fixes**: https://bugzilla.redhat.com/show_bug.cgi?id=1929577
**Fixes:** https://issues.redhat.com/browse/ODC-5546

**Root Cause:**
Currently on editing d/dc using edit flow overrides the current resource containers completely with new resource containers https://github.com/openshift/console/pull/4517/files#diff-7fd530a4a1b4113f74f1e7e64beadbdb448845c1fad06d6a4d4824425583c2d6R53-R55 not taking in account the values that we dont add from UI like envFrom, imagePullPolicy etc thereby removing these values from the containers. 

In container image flow when imagestream from internal registry is selected then its name is wrongly used as container name in the trigger annotation. 

**Solution:**
Do not completely override the containers instead update only the values that we add from UI and leave other values as is.
Use correct container name in trigger annotation.

**Gif:**
Peek 2021-02-19 21-31

Peek 2021-02-22 14-53

**Unit Test Coverage:**
image

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
+++ b/frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
@@ -191,7 +191,10 @@ export const getDeploymentData = (resource: K8sResourceKind) => {
     replicas: 1,
     triggers: { image: true, config: true },
   };
-  const container = resource.spec?.template?.spec?.containers?.[0];
+  const container = _.find(
+    resource.spec?.template?.spec?.containers,
+    (c) => c.name === resource.metadata.name,
+  );
   const env = container?.env ?? [];
   switch (getResourcesType(resource)) {
     case Resources.KnativeService:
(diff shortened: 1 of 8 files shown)


spadgett:
Hi, @divyanshiGupta. Does this also address the incorrect `image.openshift.io/triggers` annotation? (See the comments in the Bugzilla.)

divyanshiGupta:
@spadgett I created a separate bug https://issues.redhat.com/browse/ODC-5546 for that issue. But if needed I can add the fix in this PR only or raise a separate PR tomorrow.

spadgett:
> @spadgett I created a separate bug https://issues.redhat.com/browse/ODC-5546 for that issue. But if needed I can add the fix in this PR only or raise a separate PR tomorrow.

OK, thanks. We'll probably want to backport both fixes. I'd recommend either fixing both in one PR or creating separate Bugzilla bugs.

rohitkrai03 (changes requested):
/approve

Code changes looks good to me. Tested locally for both env variables and trigger annotation issue. Works as expected.

rohitkrai03 (approved):
/lgtm

spadgett:
@divyanshiGupta Looks like we will need to manually backport to 4.6.z unfortunately.

PR #8202, opened 2021-02-19
Console extension processing improvements

vojtechszocs:
- [x] extension properties updates are now done via `deepMergeExtensionProperties` function
- [x] in `useResolvedExtensions` hook, if an error occurs, re-render the component and return the error

The change in `useResolvedExtensions` hook is compatible with existing code, which currently doesn't handle coderef resolution errors.


christianvogt:
changes look good but I haven't run them

PR #8204, opened 2021-02-19
Bug 1926106: i18n translation of create & delete project namespace modal

sg00dwin:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1926106


frontend/public/components/modals/create-namespace-modal.jsx
```diff
@@ -108,108 +109,115 @@ const CreateNamespaceModal = connect(
     }
 
     render() {
-      const label = this.props.createProject ? 'Project' : 'Namespace';
       const defaultNetworkPolicies = {
```
rebeccaalpert, line 112: We should internationalize these as well.

frontend/public/components/modals/create-namespace-modal.jsx
```diff
@@ -108,108 +109,115 @@ const CreateNamespaceModal = connect(
...
-            errorMessage={this.state.errorMessage}
-            inProgress={this.state.inProgress}
-            submitText="Create"
-            cancel={this.props.cancel.bind(this)}
-          />
-        </form>
+                {this.props.createProject && (
+                  <div className="form-group">
+                    <label htmlFor="input-display-name" className="control-label">
+                      {t('public~Display Name')}
```
rebeccaalpert, line 151: I think this needs to be "Display name" since we're doing sentence case.
```suggestion
                      {t('public~Display name')}
```

frontend/public/components/modals/create-namespace-modal.jsx
```diff
@@ -108,108 +109,115 @@ const CreateNamespaceModal = connect(
...
+                        onChange={this.onLabels}
+                        tags={[]}
+                      />
+                    </div>
+                  </div>
+                )}
+                {!this.props.createProject && (
+                  <div className="form-group">
+                    <label htmlFor="network-policy" className="control-label">
+                      {t('public~Default Network Policy')}
```
rebeccaalpert, line 198: I think this needs to be sentence case too.

```suggestion
                      {t('public~Default network policy')}
```

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -46,29 +48,36 @@ export const DeleteNamespaceModal: React.FC<DeleteNamespaceModalProps> = ({
   return (
     <form onSubmit={onSubmit} name="form" className="modal-content ">
       <ModalTitle className="modal-header">
-        <YellowExclamationTriangleIcon className="co-icon-space-r" /> Delete {kind.label}?
+        <YellowExclamationTriangleIcon className="co-icon-space-r" />{' '}
+        {t('public~Delete {{label}}?', { label: kind.label })}
```
rebeccaalpert, line 52: We should internationalize what we're deleting here. The consensus was to hard-code the model instead of using the kind label, but if that's not possible, we should use `kind.labelKey` instead. (`kind.labelKey` has the internationalized version of the kind.)

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -46,29 +48,36 @@ export const DeleteNamespaceModal: React.FC<DeleteNamespaceModalProps> = ({
...
       </ModalTitle>
       <ModalBody>
         <p>
-          This action cannot be undone. It will destroy all pods, services and other objects in the
-          namespace <strong className="co-break-word">{resource.metadata.name}</strong>.
+          {t(
```
rebeccaalpert, line 56: If possible, this should be one phrase using a Trans component or similar. Breaking things up can make things hard on the translators.

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -46,29 +48,36 @@ export const DeleteNamespaceModal: React.FC<DeleteNamespaceModalProps> = ({
...
         </p>
         <input
           type="text"
           data-test="project-name-input"
           className="pf-c-form-control"
           onKeyUp={onKeyUp}
-          placeholder="Enter name"
-          aria-label={`Enter the name of the ${kind.label} to delete`}
+          placeholder={t('public~Enter name')}
+          aria-label={t('public~Enter the name of the {{label}} to delete}', {
```
rebeccaalpert, line 73: We should hard-code the kind here, or if that's not possible, use kind.labelKey.

rebeccaalpert (changes requested):
Thanks for fixing this! I just had a few small comments about internationalizing the dropdown items and kind and using sentence case. The kind needs to be internationalized so it's consistent throughout the application. We tend to translate the kind or at least change the pluralization in different languages.

I don't think we're using the `Translation` render prop anywhere else (and we don't need the i18next features it adds in this instance), but that's not a hill I particular care about climbing and/or dying on. :) Seems to work fine; I don't have a big objection to using it. Others may disagree.

frontend/public/components/modals/create-namespace-modal.jsx
sg00dwin: Replaced the `label` const since it was only being used in `<ModalTitle>` and translated each string  with
```<ModalTitle>
                {this.props.createProject
                  ? t('public~Create Project')
                  : t('public~Create Namespace')}
              </ModalTitle>
```
https://github.com/openshift/console/pull/8204/files#diff-658c9e962611ba32543853d54f224c95f31993dc0f574dd7687ece2b53514a86R125-R127

frontend/public/components/modals/create-namespace-modal.jsx
sg00dwin: updated

frontend/public/components/modals/delete-namespace-modal.tsx
sg00dwin: updated

frontend/public/components/modals/delete-namespace-modal.tsx
sg00dwin: Switched to `kind.labelKey`

frontend/public/components/modals/delete-namespace-modal.tsx
sg00dwin: This uses the `<Trans>` component and `{resource.metadata.name}` isn't included in the translation. Is this correct?

rebeccaalpert: I think it might be better to include the namespace in the translation. The sentence structure we get back from translation is sometimes radically different, so we can't always rely on being able to format the sentence in a given way.

sg00dwin: Updated

sg00dwin:
@rebeccaalpert mind giving this another look

rebeccaalpert (approved):
Looks good to me! Thanks for all your work on this.

jhadvig:
@sg00dwin please check my last comments + the PR needs rebase.

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -46,29 +48,37 @@ export const DeleteNamespaceModal: React.FC<DeleteNamespaceModalProps> = ({
...
+          <Trans ns="public">
+            This action cannot be undone. It will destroy all pods, services and other objects in
+            the namespace{' '}
+            <strong className="co-break-word">{{ name: resource.metadata.name }}</strong>.
+          </Trans>
         </p>
         <p>
-          Confirm deletion by typing{' '}
-          <strong className="co-break-word">{resource.metadata.name}</strong> below:
+          <Trans ns="public">
```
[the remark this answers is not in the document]
sg00dwin: updated

frontend/public/components/modals/delete-namespace-modal.tsx
```diff
@@ -46,29 +48,37 @@ export const DeleteNamespaceModal: React.FC<DeleteNamespaceModalProps> = ({
...
       <ModalTitle className="modal-header">
-        <YellowExclamationTriangleIcon className="co-icon-space-r" /> Delete {kind.label}?
+        <YellowExclamationTriangleIcon className="co-icon-space-r" />{' '}
+        {t('public~Delete {{label}}?', { label: t(kind.labelKey) })}
       </ModalTitle>
       <ModalBody>
         <p>
-          This action cannot be undone. It will destroy all pods, services and other objects in the
-          namespace <strong className="co-break-word">{resource.metadata.name}</strong>.
+          <Trans ns="public">
```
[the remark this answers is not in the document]
sg00dwin: updated

sg00dwin:
@jhadvig I've addressed your comments

jhadvig (approved):
/lgtm
Thanks @sg00dwin 👍

PR #8210, opened 2021-02-22
Fixes volume mode when selecting RBD + RWX

vbnrh:
This commit fixes the volume mode that is used when RBD + RWX is selected and renders a read-only text
It also refactors the code to add a InfoToolTip component to shared utils

Screenshot from 2021-02-22 18-43-53

### DIFF
--- a/frontend/public/components/storage/create-pvc.tsx
+++ b/frontend/public/components/storage/create-pvc.tsx
@@ -2,9 +2,13 @@ import * as _ from 'lodash-es';
 import * as React from 'react';
 import { Helmet } from 'react-helmet';
 import { Link } from 'react-router-dom';
-import { ActionGroup, Button, Tooltip } from '@patternfly/react-core';
-import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
-import { isCephProvisioner, isObjectSC, cephStorageProvisioners } from '@console/shared/src/utils';
+import { ActionGroup, Button } from '@patternfly/react-core';
+import {
+  isCephProvisioner,
+  isObjectSC,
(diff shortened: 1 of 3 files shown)


frontend/public/components/storage/create-pvc.tsx
```diff
@@ -173,6 +165,43 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
+      <RadioInput
+        {...radio}
+        key={radio.value}
+        onChange={handleVolumeMode}
+        inline
+        checked={radio.value === volumeMode}
+        name="volumeMode"
+      />
+    ));
+  };
```
cloudbehl, line 203: Looks a bit hacky to me.

Let do it in a proper way so that moving forward any other provisioner can also leverage this functionality if they don't support any volume mode.

Let define an object with the keys pointing to the provisioner name and values being an object that says volume-mode.

cloudbehl (changes requested):

frontend/public/components/storage/create-pvc.tsx
vbnrh: i don't understand this approach.... Could you elaborate ? 
We have only two cases to look out for here... 
When provisioner is cephfs and any access mode ; we want the volume mode to always be 'Filesystem' 
When provisioner is ceph rbd and only RWX access mode; we want the volume mode to always be 'Block'

Having an map defined for only two very specific conditions seems to be convoluted approach (as per what i understand). My suggestion is to move this function to shared if we want re-usability . We can change to different approach once more conditions come up.  WDYT ?

PR #8213, opened 2021-02-22
change addForm layout to make footer full length

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-3724, https://issues.redhat.com/browse/ODC-3464

**Analysis / Root cause**: No indication that there is content out of view on add form

**Solution Description**: added shadow to footer to indicate there is content outside of view, made sticky footer full length by changing layout of form

Screenshot 2021-02-22 at 9 53 16 PM
Screenshot 2021-02-22 at 9 53 32 PM
Screenshot 2021-02-22 at 10 04 57 PM

**Test setup:**

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
+++ b/frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
@@ -5,6 +5,7 @@ import { ActionGroup, Alert, Button, ButtonVariant } from '@patternfly/react-cor
 import { ButtonBar } from '@console/internal/components/utils';
 import { FormFooterProps } from './form-utils-types';
 import './FormFooter.scss';
+import { Shadows, useScrollShadows, useScrollContainer } from '../../hooks';
 
 const FormFooter: React.FC<FormFooterProps> = ({
   handleSubmit,
@@ -24,63 +25,71 @@ const FormFooter: React.FC<FormFooterProps> = ({
   sticky,
 }) => {
   const { t } = useTranslation();
+  const [scrollContainer, footerElementRef] = useScrollContainer();
+  const shadowPosition = useScrollShadows(sticky ? scrollContainer : null);
   return (
-    <ButtonBar
+    <div
       className={cx('ocs-form-footer', {
         'ocs-form-footer__sticky': sticky,
+        'pf-u-box-shadow-sm-top':
+          sticky && (shadowPosition === Shadows.both || shadowPosition === Shadows.bottom),
       })}
-      inProgress={isSubmitting}
-      errorMessage={errorMessage}
-      successMessage={successMessage}
+      ref={footerElementRef}
     >
-      {showAlert && (
-        <Alert
-          isInline
-          className="co-alert"
-          variant="info"
-          title={infoTitle || t('console-shared~You made changes to this page.')}
-        >
-          {infoMessage ||
-            t('console-shared~Click {{submit}} to save changes or {{reset}} to cancel changes.', {
-              submit: submitLabel,
-              reset: resetLabel,
-            })}
-        </Alert>
-      )}
-      <ActionGroup className="pf-c-form pf-c-form__group--no-top-margin">
-        {!hideSubmit && (
-          <Button
-            type={handleSubmit ? 'button' : 'submit'}
-            {...(handleSubmit && { onClick: handleSubmit })}
-            variant={ButtonVariant.primary}
-            isDisabled={disableSubmit}
-            data-test-id="submit-button"
+      <ButtonBar
+        inProgress={isSubmitting}
+        errorMessage={errorMessage}
+        successMessage={successMessage}
+      >
+        {showAlert && (
+          <Alert
+            isInline
+            className="co-alert"
+            variant="info"
+            title={infoTitle || t('console-shared~You made changes to this page.')}
           >
-            {submitLabel || t('console-shared~Save')}
-          </Button>
+            {infoMessage ||
+              t('console-shared~Click {{submit}} to save changes or {{reset}} to cancel changes.', {
+                submit: submitLabel,
+                reset: resetLabel,
+              })}
+          </Alert>
         )}
-        {handleReset && (
-          <Button
-            type="button"
-            data-test-id="reset-button"
-            variant={ButtonVariant.secondary}
-            onClick={handleReset}
-          >
-            {resetLabel || t('console-shared~Reload')}
-          </Button>
-        )}
-        {handleCancel && (
-          <Button
-            type="button"
-            data-test-id="cancel-button"
-            variant={ButtonVariant.secondary}
-            onClick={handleCancel}
-          >
-            {cancelLabel || t('console-shared~Cancel')}
-          </Button>
-        )}
-      </ActionGroup>
-    </ButtonBar>
+        <ActionGroup className="pf-c-form pf-c-form__group--no-top-margin">
+          {!hideSubmit && (
+            <Button
+              type={handleSubmit ? 'button' : 'submit'}
+              {...(handleSubmit && { onClick: handleSubmit })}
+              variant={ButtonVariant.primary}
+              isDisabled={disableSubmit}
+              data-test-id="submit-button"
+            >
+              {submitLabel || t('console-shared~Save')}
+            </Button>
+          )}
+          {handleReset && (
+            <Button
+              type="button"
+              data-test-id="reset-button"
+              variant={ButtonVariant.secondary}
+              onClick={handleReset}
+            >
+              {resetLabel || t('console-shared~Reload')}
+            </Button>
+          )}
+          {handleCancel && (
+            <Button
+              type="button"
+              data-test-id="cancel-button"
+              variant={ButtonVariant.secondary}
+              onClick={handleCancel}
+            >
+              {cancelLabel || t('console-shared~Cancel')}
+            </Button>
+          )}
+        </ActionGroup>
+      </ButtonBar>
+    </div>
   );
 };
 export default FormFooter;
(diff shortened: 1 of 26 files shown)


frontend/packages/console-shared/src/hooks/useScrollShadows.tsx
```diff
@@ -43,3 +43,23 @@ export const useScrollShadows = (node: HTMLElement): Shadows => {
...
+
+type FormScrollShadowsContextType = { shadow?: Shadows };
+
+export const FormScrollShadowsContext = React.createContext<FormScrollShadowsContextType>({});
+
+type FormScrollShadowProviderProps = {
+  children: (callback: (node: HTMLElement) => void) => React.ReactNode;
+};
+
+export const FormScrollShadowProvider: React.FC<FormScrollShadowProviderProps> = ({ children }) => {
```
rohitkrai03, line 55: Why do we need to keep the scroll position in a top level context here?

frontend/packages/dev-console/src/components/NamespacedPage.tsx
```diff
@@ -28,23 +29,28 @@ const NamespacedPage: React.FC<NamespacedPageProps> = ({
...
-    </NamespaceBar>
-    <div
-      className={cx('odc-namespaced-page__content', {
-        [`is-${variant}`]: variant !== NamespacedPageVariants.default,
-      })}
-    >
-      {children}
-    </div>
-  </div>
+  <FormScrollShadowProvider>
```
rohitkrai03, line 32: I don't think we need to wrap the whole namespaced page component into this scroll provider. This component will soon move out to either `console-shared` or `sdk` package. Also the name `FormScrollShadowProvider` kind of hints that this component is going to be used for just forms.

rohitkrai03 (changes requested):

frontend/packages/console-shared/src/hooks/useScrollShadows.tsx
sahil143: Ref needs to be attached to the scrollable area and useScrollShadows need the ref that's why scrollPosition is in top level context

frontend/packages/dev-console/src/components/NamespacedPage.tsx
sahil143: I think we can move it to the ImportPage component which deals with all the forms. 
We can rename it to `ScrollShadowProvider`. WDYT?

christianvogt: By wrapping every namespace page, we're going to get a bunch of react updates every time we are scrolling pages that don't even care about the scroll position.

Ideally I'd like a to let each component manage their own needs. Which means that the footer should identify the scroll container and attach its own listener.

There's a similar need which was addressed in `dom-utils` to find the parent scroll container. Perhaps we can apply the same concept to a new hook and let the footer `find` the scroll parent to listen to. The hook can return `[scrollContainerRef, setElementRef]` where `setElementRef` is applied to the footer and the hook will traverse up the dom to find the scroll container.

sahil143: used approach mentioned by @christianvogt which works fine.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
```diff
@@ -176,6 +176,7 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
                   values.formData.tasks.length === 0
             }
             resetLabel={t('pipelines-plugin~Cancel')}
+            shadow={false}
```
sahil143, line 179: shadow set to false because layout for pipelinebuilder is different from add flows and will require more changes. looking into it.

frontend/packages/console-shared/src/hooks/useScrollContainer.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+  return undefined;
+};
+
+export const useScrollContainer = (): [HTMLElement, (node: HTMLElement) => void] => {
+  const [scrollContainerRef, setScrollContainerRef] = React.useState<HTMLElement>();
+  const elementRef = React.useCallback((node: HTMLElement) => {
+    if (node) {
+      setScrollContainerRef(getParentScrollableElement(node));
+    }
```
christianvogt, line 29: For completeness, if `node ==  null`, we should probably unset the `scrollContainerRef`

frontend/packages/console-shared/src/hooks/useScrollContainer.ts
```diff
@@ -0,0 +1,32 @@
...
+        return parentNode;
+      }
+    }
+    parentNode = parentNode.parentNode;
+  }
+  return undefined;
+};
+
+export const useScrollContainer = (): [HTMLElement, (node: HTMLElement) => void] => {
+  const [scrollContainerRef, setScrollContainerRef] = React.useState<HTMLElement>();
```
christianvogt, line 25: Convention of `Ref` suffix denotes a react reference which this isn't.

```suggestion
  const [scrollContainer, setScrollContainer] = React.useState<HTMLElement>();
```

frontend/packages/dev-console/src/components/import/DeployImagePage.tsx
```diff
@@ -21,19 +21,19 @@ const DeployImagePage: React.FunctionComponent<DeployImagePageProps> = ({ match,
...
-              <DeployImage
-                forApplication={desiredApplication}
-                namespace={namespace}
-                contextualSource={params.get(QUERY_PROPERTIES.CONTEXT_SOURCE)}
-              />
-            </Firehose>
-          )}
-        </QueryFocusApplication>
-      </div>
+      {/* <div className="co-m-pane__body" style={{ paddingBottom: 0 }}> */}
```
christianvogt, line 24: Remove commented code.

frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
```diff
@@ -22,65 +23,74 @@ const FormFooter: React.FC<FormFooterProps> = ({
   hideSubmit = false,
   showAlert,
   sticky,
+  shadow = true,
```
christianvogt, line 26: Shouldn't all sticky footers have a shadow?

Best practice is to have boolean properties default to false because it allows for the shorthand jsx notation to set to true.

frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
```diff
@@ -22,65 +23,74 @@ const FormFooter: React.FC<FormFooterProps> = ({
...
 }) => {
   const { t } = useTranslation();
+  const [scrollContainer, footerElementRef] = useScrollContainer();
+  const shadowPosition = useScrollShadows(scrollContainer);
```
christianvogt, line 30: Should find a way to not pass in a `scrollContainer` unless we need to show shadows, otherwise the listeners and calculations are needlessly done.

frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
christianvogt: Perhaps this is present because of the issue you have with the pipeline builder form?

sahil143: correct. currently pipeline and helm page layout is not fixed.. that's why.
I was going to remove the shadow property altogether from the component.
Do we need a shadow prop?

frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
sahil143: listeners and calculations are only done when `scrollContainer` is available.

christianvogt: That's right but before you use to only show shadows when a prop was passed in which made the calculation conditional and therefore should only pass in a non `null` `scrollContainer` if shadows are to be drawn.

BTW I think you still want to not show shadows if the footer isn't sticky? looks like they are always shown now.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/GitImportForm.tsx
```diff
@@ -26,17 +25,19 @@ const GitImportForm: React.FC<FormikProps<FormikValues> & GitImportFormProps> =
...
-      <BuilderSection image={values.image} builderImages={builderImages} />
-      <DockerSection buildStrategy={values.build.strategy} />
-      <AppSection
-        project={values.project}
-        noProjectsAvailable={projects.loaded && _.isEmpty(projects.data)}
-      />
-      <ResourceSection />
-      <PipelineSection builderImages={builderImages} />
-      <AdvancedSection values={values} />
+    <form onSubmit={handleSubmit} data-test-id="import-git-form">
```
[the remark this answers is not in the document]
sahil143: Nope, Since these don't involve the Yaml Editor component. I didn't see the need to use flex here.

jerolimov:
Here are two more pages:

1. Add > Samples > Select a sample ==> Footer is not sticky anymore.  And the footer shadow has an unexpected padding left and right:

image

2. Add > Dev Catalog > Database > MariaDB > Instantiate Template ==> Has no sticky header before. Want we add it here also?

image

sahil143:
> Add > Samples > Select a sample ==> Footer is not sticky anymore. And the footer shadow has an unexpected padding left and right:

@jerolimov fixed this

> Add > Dev Catalog > Database > MariaDB > Instantiate Template ==> Has no sticky header before. Want we add it here also?

If we need a sticky footer for the template form as well. This needs to be taken as a separate issue.

frontend/packages/dev-console/src/components/import/GitImportForm.tsx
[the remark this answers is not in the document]
rohitkrai03: Okay, got it.

PR #8219, opened 2021-02-23
Bug 1931755: Fix test about VM start via kebab action

gouyang:
Some tests are failed because buttons in kebab action are not enabled quickly.

https://cnv-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/test-kubevirt-console-gouyang/56/testReport/


gouyang:
verified in https://cnv-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/test-kubevirt-console/13/console

frontend/packages/console-shared/src/test-utils/actions.view.ts
```diff
@@ -29,6 +29,8 @@ const selectDropdownItem = (getActionsDropdown) => async (action: string) => {
   await browser.wait(until.elementToBeClickable(getActionsDropdown()));
   await getActionsDropdown().click();
   await browser.wait(until.presenceOf($(detailViewDropdownMenu)));
+  // wait for disabled items gone
+  await browser.sleep(3000);
```
yaacov, line 33: This is a change in `console-share` that may affect other packages, can we move the wait to `kubevirt plugin` ?
If we can't can we make if parameterized and default to `0` so it will not affect other tests ?

gouyang: It's not easy to move to `kubevirt-plugin` or add parameters to it from my perspective.
The problem is that the items in action dropdown list is not enabled immediately, do you think is it a bug in the design?
The impact is it sleeps 3 seconds, does it accept if I reduce it to 1 second, just test it works with 1 second.

PR #8224, opened 2021-02-23
Reconcile resolved extension types

vojtechszocs:
- [x] move `ResolvedExtension` type to dynamic plugin SDK common types
- [x] use [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) in `useResolvedExtensions` hook
- [x] new `resolveExtension` function, can be used with `subscribeToExtensions` API

---

:thinking: Let's assume that `CodeRef<T>` resolution fails within the `useResolvedExtensions` hook.

**Before this PR**
- any ongoing resolutions are stopped
- returned tuple values
  - `resolvedExtensions` will remain `[]` (even if some extensions were resolved successfully)
  - `resolved` will remain `false`
  - `error` will be the _first_ resolution error

**After this PR**
- finish processing any ongoing resolutions
- returned tuple values
  - `resolvedExtensions` will contain extensions that were resolved successfully
  - `resolved` will be `true`
  - `errors` will contain _all_ resolution errors


frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
```diff
@@ -48,38 +40,30 @@ export const useResolvedExtensions = <E extends Extension>(
...
     let disposed = false;
 
-    Promise.all(
+    // eslint-disable-next-line promise/catch-or-return
+    Promise.allSettled(
       extensions.map(async (e) => {
-        const resolvedProperties = await resolveCodeRefProperties(e);
-        return deepMergeExtensionProperties(e, resolvedProperties) as ResolvedExtension<E>;
+        const resolvedExtension = await resolveExtension<typeof e, any, ResolvedExtension<E>>(e);
+        return resolvedExtension;
```
vojtechszocs, line 52: @christianvogt The reason for declaring above constant is to comply with ESLint [`no-return-await`](https://eslint.org/docs/rules/no-return-await) rule.

Following alternative is possible, I'm not sure which one is better:
```ts
// eslint-disable-next-line no-return-await
return await resolveExtension<typeof e, any, ResolvedExtension<E>>(e);
```

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
```diff
@@ -48,38 +40,30 @@ export const useResolvedExtensions = <E extends Extension>(
...
   const [resolvedExtensions, setResolvedExtensions] = React.useState<ResolvedExtension<E>[]>([]);
   const [resolved, setResolved] = React.useState<boolean>(false);
-  const [error, setError] = React.useState<any>(undefined);
+  const [errors, setErrors] = React.useState<any>(undefined);
 
   React.useEffect(() => {
     let disposed = false;
 
-    Promise.all(
+    // eslint-disable-next-line promise/catch-or-return
```
vojtechszocs, line 48: @christianvogt FYI the ESLint `promise/catch-or-return` rule is meaningless for `Promise.allSettled` function.

vojtechszocs:
cc @christianvogt

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
christianvogt: @vojtechszocs I think it is kind of a mute point to use `async` `await` here instead of simply returning the promise result from `resolveExtension`.

```
extensions.map((e) => resolveExtension<typeof e, any, ResolvedExtension<E>>(e));
```

vojtechszocs: Agreed, will update the code.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
christianvogt: Interesting. I wonder if there's an update or a setting for this rule.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
```diff
@@ -43,42 +35,36 @@ import {
...
-        }
-      });
+    // eslint-disable-next-line promise/catch-or-return
+    Promise.allSettled(
+      extensions.map((e) => resolveExtension<typeof e, any, ResolvedExtension<E>>(e)),
+    ).then((results) => {
+      if (!disposed) {
+        const [fulfilledValues, rejectedReasons] = unwrapPromiseSettledResults(results);
+        setResolvedExtensions(fulfilledValues);
+        setErrors(rejectedReasons);
```
rohitkrai03, line 55: Is there a way of knowing which error was caused by which unresolved extension?

vojtechszocs: @rohitkrai03 Do we have a use case for that, besides potentially logging all of these errors?

If we do, we can adapt the hook result type, e.g. an array of

```ts
{
  resolved: boolean;
  extension?: ResolvedExtension<E>; // if resolved is true
  error?: any; // if resolved is false
}
```

which is basically an analogy of [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) result type.

vojtechszocs: > Is there a way of knowing which error was caused by which unresolved extension?

I think the only association here is the order of objects within `Promise.allSettled` result.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
christianvogt: @vojtechszocs did you look into this?

frontend/public/actions/features.ts
```diff
@@ -266,16 +266,17 @@ const featureFlagController: SetFeatureFlag = (flag, enabled) => {
...
-        processedExtensions.push(e);
-        executeReferencedFunction(e.properties.handler, featureFlagController);
-      }
+      resolveExtension(e)
+        .then((resolvedExtension) => {
+          resolvedExtension.properties.handler(featureFlagController);
+        })
+        .catch((error) => {
+          // eslint-disable-next-line no-console
+          console.log(error);
```
christianvogt, line 278:
```suggestion
          console.error(error);
```

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
rohitkrai03: I don't see any use case with our current codebase. Was just thinking if there was a way we could use it for better error logging. But the current API of `useResolvedExtensions` seem simpler to me and is already being used in a lot of places. So I won't push for a change in API because of this. We can maybe later do this if there's an actual use case for it.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
vojtechszocs: @christianvogt Looking at [`promise/catch-or-return`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/catch-or-return.md) rule documentation, there's only `terminationMethod` option which isn't useful in this particular case.

vojtechszocs:
Addressed all comments, please review.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
vojtechszocs: Added a comment explaining the suppression of `promise/catch-or-return` rule.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
```diff
@@ -43,42 +35,42 @@ import {
...
+      }
+    });
 
     return () => {
       disposed = true;
     };
   }, [extensions]);
 
-  return [resolvedExtensions, resolved, error];
+  return React.useMemo(() => [resolvedExtensions, resolved, errors], [
```
rawagner, line 71: you dont need to wrap it in useMemo. All values are from state, thus they dont change unless we call setters

vojtechszocs: This was done to comply with current JSDoc comment which states

> The hook's result is guaranteed to be referentially stable across re-renders.

But you're right, people usually destructure the hook's result anyway, so no need for memo there.

rohitkrai03 (approved):
/lgtm

Changes looks good to me.

frontend/packages/console-dynamic-plugin-sdk/src/api/useResolvedExtensions.ts
vojtechszocs: Updated the hook's JSDoc comment accordingly.

PR #8225, opened 2021-02-23
Update topology quick search to include results from quick starts catalog

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5445

**Description:**
This PR adds quick starts search and start support to topology quick search feature.

**Gif:**
Peek 2021-02-23 21-52

/kind feature


frontend/packages/topology/src/components/quick-search/QuickSearch.tsx
```diff
@@ -12,11 +13,18 @@ interface QuickSearchProps {
...
-      <QuickSearchController
-        {...catalogService}
-        namespace={namespace}
-        viewContainer={viewContainer}
-      />
+      <QuickStartsLoader>
+        {(quickStarts, qsLoaded) => (
+          <QuickSearchController
+            catalogItems={catalogService.items}
+            quickStarts={quickStarts}
```
christianvogt, line 20: All items should be generalized into a single data structure. In order to distinguish the origin of the results and abstract various providers in the search you can create a structure such as the following or come up with something else. The idea is to remove all the `if devcatalog else if quick start` conditions throughout the quick search and also support more providers in the future such as our samples or add page items.

```
[
  {
    catalog: 'devcatalog',
    getViewAllURL: (searchTerm: string) => `/catalog/ns/${namespace}?keyword=${searchTerm}`,
    // t('topology~View all developer catalog items ({{count, number}})')
    viewAllLinkKey: 'topology~View all developer catalog items ({{count, number}})'
  },
  {
    catalog: 'quickStarts',
    getViewAllURL: (searchTerm: string) => `/quickstart?keyword=${searchTerm}`,
    // t('topology~View all quick starts ({{totalQcItems, number}})'
    viewAllLinkKey: 'topology~View all quick starts ({{count, number}})',
  }
]
```

By generalizing each data set as a separate catalog, we can re-introduce the `catalog` property to the catalog extensions to distinguish providers for different catalogs (consult with @rohitkrai03).
Using this data structure you can create multiple `CatalogServiceProvider` wherein each provides items from their specific catalogs. This would require you to create a quick start catalog provider.

Whatever path we decide on, quick search should act on a generic data structure which provides all the necessary items, search capabilities, and linking for each item.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -41,6 +42,18 @@ const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem })
...
+        <TextContent>
+          <Text component={TextVariants.h6} className="odc-quick-search-details__description">
+            Prerequisites
+          </Text>
+          {selectedItem.data.prerequisites.map((pq: string) => (
+            <Text component={TextVariants.p} className="odc-quick-search-details__description">
+              {pq}
+            </Text>
+          ))}
+        </TextContent>
```
christianvogt, line 55: This should be part of the quick search catalog item description itself so that the quick search doesn't have special handling for different item types.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -111,12 +129,20 @@ const QuickSearchList: React.FC<QuickSearchListProps> = ({
...
       </DataList>
       <div className="odc-quick-search-list__all-items-link">
-        <Button variant="link" onClick={goToCatalogPage}>
-          {t('topology~View all results for "{{searchTerm}}" ({{totalItems, number}})', {
-            searchTerm,
-            totalItems,
-          })}
-        </Button>
+        {totalDcItems > 0 && (
+          <Button variant="link" onClick={() => goToCatalogPage('Dev')} id="catalog">
```
christianvogt, line 133: Why the need for the `id` attribute?

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -77,12 +87,16 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
   );
 
   const onEnter = React.useCallback(() => {
-    const viewAllLink = document.getElementById('viewAll');
+    const devCatalog = document.getElementById('catalog');
+    const quickStartCatalog = document.getElementById('quickstarts');
```
christianvogt, line 91: I see now where the IDs are being used.
I think the enter key should only work when focused on the text input. The only other items to focus are the view all links and the CTA buttons. Both of which can be invoked using the keyboard already.
If your listener for the enter key is on the text input only, then you won't have to perform these checks.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -139,12 +153,17 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
     };
   }, [closeModal, onCancel, onEnter, selectNext, selectPrevious]);
 
+  const getModalHeight = (): number => {
```
christianvogt, line 156: If you create an items data structure, you can generalize this calculation to be based on the number of item contributors.

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    type: 'Quick Start',
+    uid: qs.metadata.uid,
+    cta: {
+      callbackfn: () => setActiveQuickStart(qs.metadata.name, qs.spec.tasks?.length),
+      label: 'Start',
+    },
+    icon: {
+      url: qs.spec.icon,
+    },
+    description: qs.spec.description,
```
christianvogt, line 24: Include the prerequisite `ReactNode` elements inside the description.

frontend/packages/topology/src/components/quick-search/QuickSearch.tsx
christianvogt: Discussed with @rohitkrai03 and he has concerns with the extra work necessary to use the catalog for quick starts. Where the data comes from and how we use the data are separate enough that we can address the "where" later on and continue to use the `QuickStartsLoader`. However generalizing the data should still be addressed.

rohitkrai03: Yeah, we can generalise the data structure for the quick search catalog with information about view all link and label. All this can be abstracted in the `QuickSearchController` so that we can easily move this logic to extensions and not affect other components.

frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
```diff
@@ -1,23 +1,54 @@
...
+  quickStartsLoaded,
   viewContainer,
 }) => {
   const [isQuickSearchActive, setIsQuickSearchActive] = React.useState<boolean>(
     !!getQueryArgument('catalogSearch'),
   );
+  const { setActiveQuickStart } = React.useContext<QuickStartContextValues>(QuickStartContext);
+
+  const quickStartItems: CatalogItem[] = React.useMemo(() => {
+    return quickStartsLoaded ? getTransformedQuickStarts(quickStarts, setActiveQuickStart) : [];
```
rohitkrai03, line 36: This could be handled in a custom hook like `useTransformedQuickStarts`.

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+export const getTransformedQuickStarts = (
+  quickStarts: QuickStart[],
+  setActiveQuickStart: (quickStartId: string, totalTasks?: number) => void,
+): CatalogItem[] => {
+  return quickStarts.map((qs: QuickStart) => ({
+    name: qs.spec.displayName,
+    type: 'Quick Start',
+    uid: qs.metadata.uid,
+    cta: {
+      callbackfn: () => setActiveQuickStart(qs.metadata.name, qs.spec.tasks?.length),
```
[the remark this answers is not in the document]
rohitkrai03: This is because we do not instantiate states for all the quick starts on load. We do that lazily when a quick start is actually opened, that's why we need to pass the number of tasks here. We can look at improving this for all quick starts in a separate issue and instantiate quick starts states eagerly on quick start load.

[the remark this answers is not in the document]
christianvogt: This is very odd that we need this information to init the task state to `INIT`. Can we simply assume that a task without state is in the initial state?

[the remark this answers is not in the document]
rohitkrai03: I don't exactly remember what caused me to implement it this way but maybe we can update this to assume the initial state.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -139,12 +145,17 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
     };
   }, [closeModal, onCancel, onEnter, selectNext, selectPrevious]);
 
+  const getModalHeight = () => {
+    let height: number = 60;
+    if (catalogItems?.length > 0) {
+      if (viewAll.length === 2) height = 471;
+      else if (viewAll.length === 1) height = 448;
```
rohitkrai03, line 152: Not sure how you came to this height number but it could be converted to a simple formula like this - 
```suggestion
      height += 388 + viewAll.length*23;
```

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
```diff
@@ -1,23 +1,63 @@
...
+      return quickSearchProviders.reduce(
+        (acc, quickSearchProvider) => {
+          const items = quickSearchProvider.loaded
+            ? quickSearch(quickSearchProvider.items, searchTerm)
+            : [];
+          const itemCount = items.length;
+          const viewAllLink =
+            itemCount > 0
+              ? [
+                  <Button variant="link" isInline>
```
rohitkrai03, line 40: This would prevent tab to go to the button and will directly jump to link.
```suggestion
                  <Button variant="link" tabIndex={-1} isInline>
```

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+        {qs.spec.prerequisites?.length > 0 && (
+          <>
+            <Text component={TextVariants.h6}>Prerequisites</Text>
+            {qs.spec.prerequisites.map((pq: string) => (
+              <Text component={TextVariants.p}>{pq}</Text>
+            ))}
+          </>
+        )}
+      </>
+    );
```
rohitkrai03, line 28: I think we need to update this to the new prerequisites design for quick starts. I added a comment on the design doc regarding this and got the answer from UX that we should keep it consistent. 

I think we can directly reuse `QuickStartTileDescription` component here.

```suggestion
    const description = <QuickStartTileDescription description={qs.spec.description} prerequisites={qs.spec.prerequisites} />
```

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
christianvogt: The reason for the new design in the quick start tile is because of limited available space. In the catalog sidebar we don't have this restriction.

Whatever UX decides on I guess.

rohitkrai03: Yeah, that why I asked this question in the design docs and they agreed to make it consistent everywhere.

frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
```diff
@@ -1,23 +1,68 @@
...
+          const viewAllLink =
+            itemCount > 0
+              ? [
+                  <Button
+                    variant="link"
+                    isInline
+                    tabIndex={-1}
+                    key={quickSearchProvider.catalogType}
+                  >
+                    <Link to={quickSearchProvider.getCatalogURL(searchTerm, namespace)}>
```
christianvogt, line 46: This is very odd to wrap the `Link` with a `Button`. I guess PF doesn't have good support for this. 
Ideally use `<Button component={Link} to={...}` however PF doesn't allow us to provide the proper types to the Button.

I'd prefer to work around the typescript types incompatibility while we sort out a fix with PF instead of creating the odd markup.

frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
```diff
@@ -1,23 +1,68 @@
...
+                ]
+              : [];
+          return {
+            filteredItems: [...acc.filteredItems, ...items].sort((item1, item2) =>
+              item1.name.localeCompare(item2.name),
+            ),
+            viewAllLinks: [...acc.viewAllLinks, ...viewAllLink],
+          };
+        },
+        { filteredItems: [], viewAllLinks: [] },
```
christianvogt, line 61: I don't think the search function should be responsible for generating the view all links but for now I'm not oppose to it. IMO the view rendering the results should use the data and generate the links itself.

frontend/packages/topology/src/components/quick-search/QuickSearchController.tsx
christianvogt: Raised a PF issue: https://github.com/patternfly/patternfly-react/issues/5491

divyanshiGupta: Tried the typescript workaround with sending link in component prop of button component. It worked with some style issues but the tab enter didn’t work correctly. Now using <Link> directly instead of wrapping it inside <Button>

frontend/packages/dev-console/src/components/catalog/hooks/useCtaLink.ts
```diff
@@ -1,14 +1,19 @@
...
+const useCtaLink = (cta: { label: string; href?: string }): [string, string] => {
   const queryParams = useQueryParams();
 
   if (!cta) {
     return [null, null];
   }
 
   const { href, label } = cta;
+
+  if (!href) {
```
rohitkrai03, line 13: Add a test in `useCtaLink.spec.ts` for this scenario as well.

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -77,14 +86,11 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
...
-    if (activeElement === viewAllLink) {
-      history.push(`/catalog/ns/${namespace}?keyword=${searchTerm}`);
-    } else if (selectedItem) {
-      history.push(selectedItem.cta.href);
+    if (selectedItem) {
+      const { href, callback } = selectedItem.cta;
+      callback ? callback() : history.push(href);
     }
-  }, [selectedItem, namespace, searchTerm]);
+  }, [selectedItem]);
```
christianvogt, line 93: Minor issue is that the clicking `enter` while focused on the view all links causes the user to navigate to the selected item instead of the view all link.
I think this can be fixed by stopping propagation of the event on the view all `Link`.

christianvogt:
just one minor issue with the `enter` key on view all links.
Otherwise it works well.
Thanks for all improvements @divyanshiGupta 
/approve

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
rohitkrai03: I think the stop propagation won't work here because of the way event listener is added for enter key. We'll have to add a check that distinguishes a keyDown on view all link vs the selectedItem. Maybe we can merge this PR and handle that separately considering its the last day of this sprint. @christianvogt WDYT?

christianvogt: `onEnter` should probably only be executed when the text input has focus.

christianvogt: @divyanshiGupta please create an issue / PR to followup and address this.

rohitkrai03 (approved):
/lgtm

We should handle the enter issue mentioned by Christian in a separate PR. Apart from that everything works as expected.

PR #8226, opened 2021-02-23
Consume KebabActions extension via standard API

vojtechszocs:
- [x] consume `KebabActions` extension via `subscribeToExtensions`
- [x] above change allows us to remove the legacy `ExtensionRegistry` :tada: 

Long-term, all extensions should be consumed/interpreted as part of React render flow.

All extensions are now interpreted in a way that supports gating them via Console feature flags.


vojtechszocs:
cc @christianvogt @invincibleJai

frontend/public/components/utils/kebab.tsx
```diff
@@ -398,16 +399,20 @@ kebabFactory.common = [
   kebabFactory.Delete,
 ];
 
+let kebabActionExtensions: KebabActions[] = [];
+
+subscribeToExtensions<KebabActions>((extensions) => {
```
vojtechszocs, line 404: FYI - `KebabActions` is the extension type name, and `KebabAction` is the function for creating `KebabOption` objects.

The naming is confusing, we should use a better name when migrating `KebabActions` extension to a new type.

invincibleJai:
/lgtm

Thanks!! Verified the changes, works as expected  🎉

PR #8227, opened 2021-02-23
Add single control point of all toast notifications

christianvogt:
While patternfly does support [alert group toasts](https://www.patternfly.org/v4/components/alert-group), it does not provide a mechanism for managing all toasts in the application. A leaf component which needs to throw up a toast, must manage the full lifecycle of the toast itself like any other react component. This isn't ideal when the need arises to throw up a toast in response to an async network call and then move on to a new page.

This PR adds a `ToastProvider` which wraps the root of the application similar to redux provider. A React context is setup which allows any child component to reach for the `ToastContext` and use the API:

- `context.addToast(options: ToastOptions)`
- `context.removeToast(id: string)`

Added a convenience hook `useToast` which can be used instead of `useContext(ToastContext)`;

The `ToastOptions` are flexible enough to give developers control over toasts which auto dismiss or require user interaction as well as support actions for future use cases.

image

Unfortunately there is no simple way to test right now as this PR does not the use of toasts anywhere.
To test locally I exported the toast context as follows: `(window as any).toast = controller;`
I can then use the APIs:
```
toastId = window.toast.addToast({
  variant:'success',
  title:'test',
  content:'This is the content',
  dismissible: true,
  actions: [{
    dismiss: true,
    label: 'link 1',
    callback: () => alert('action')
  }]
});

window.toast.removeToast(toastId)
```

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari (didn't test Safari because my safari is too old to support ResizeObserver until I update the OS)
- [x] Edge

cc @openshift/team-devconsole-ux @spadgett

### DIFF
--- a/frontend/__tests__/utils/hooks-utils.tsx
+++ b/frontend/__tests__/utils/hooks-utils.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { mount } from 'enzyme';
+import { mount, MountRendererProps } from 'enzyme';
 
 const useRerender = () => {
   const [, setState] = React.useState(0);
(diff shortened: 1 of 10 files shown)


rohitkrai03 (approved):
/lgtm

Verified locally. Works as expected.

rohitkrai03:
@christianvogt You have conflicts in app.jsx.

christianvogt:
@rohitkrai03 rebased and waiting on tests

PR #8230, opened 2021-02-24
Monitoring: Add `title`s to the alert and alerting rule list rows

kyoto:
For both use the `description` annotation if available or the `message`
annotation otherwise.

FYI @cshinn 

title


spadgett (approved):
/lgtm

PR #4863, opened 2020-03-31
Monitoring: Add Alerting Rules list page

kyoto:
Add an Alerting Rules list page under Monitoring > Alerting.

Remove the "Not Firing" alerts from the Alerts list page. These "Not
Firing" alerts represented alerting rules that were not firing, so we
are removing them since we now have a separate alerting rules list.

screenshot

FYI @cshinn, @sichvoge

PR #8231, opened 2021-02-24
Bug 1866322: Add prop to AlertItem to hide default link

bipuladh:
- Disable default link from Ceph and Noobaa Dashboard
 - Add support for Rook related Alerts in Ceph Dashboard
 Before: 
Screenshot from 2021-02-24 11-41-50
After:
Screenshot from 2021-02-24 13-13-38
/assign @rawagner @yuvalgalanti

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/dashboards/persistent-internal/status-card/status-card.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/dashboards/persistent-internal/status-card/status-card.tsx
@@ -41,7 +41,9 @@ export const CephAlerts = withDashboardResources(
       <AlertsBody error={!_.isEmpty(loadError)}>
         {loaded &&
           alerts.length > 0 &&
-          alerts.map((alert) => <AlertItem key={alertURL(alert, alert.rule.id)} alert={alert} />)}
+          alerts.map((alert) => (
+            <AlertItem key={alertURL(alert, alert.rule.id)} alert={alert} hideLink />
+          ))}
       </AlertsBody>
     );
   },
--- a/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
+++ b/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
@@ -11,8 +11,14 @@ const enum status {
   BOUND = 'Bound',
   AVAILABLE = 'Available',
 }
-export const filterCephAlerts = (alerts: Alert[]): Alert[] =>
-  alerts.filter((alert) => _.get(alert, 'annotations.storage_type') === 'ceph');
+export const filterCephAlerts = (alerts: Alert[]): Alert[] => {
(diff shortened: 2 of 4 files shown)


spadgett:
This feels like the wrong solution to me. Can we not improve the alert details instead? I realize this is a stop-gap, but this doesn't seem to help other contexts where the alert is displayed.

cc @cshinn @kyoto for input

kyoto:
I agree with @spadgett that removing the links seems like the wrong approach. Also, I wonder if the linked bug hasn't been addressed already? 😄 

The bug was opened against 4.4 and there have been a number of improvements to the alert details page since then. I therefore think it would be better to restore these links.

Comments on the specific issues mentioned by the bug below.

> The alert details page repeated the same information...
> The alert message didn’t help diagnose the problem...

This is a problem with the alert's message / description text, so it depends on the alert. There is certainly much room for improvement here, but some do have useful information.


> Users questioned the value of Alert details page when it showed an empty chart

Not exactly sure what situation this was, but the chart should now always have some data. Perhaps this was caused by the concept of a "Not Firing" alert which we have now removed (PR https://github.com/openshift/console/pull/4863).


> One user questioned what “Firing” meant for State

This was addressed by adding popover help text on the alert details page.
screenshot

bipuladh:
@yuvalgalanti PTAL.

andybraren:
+1 to adding back the links. Removing them doesn't really address the underlying problem, is inconsistent with other Overview Status cards (which I can imagine someone creating a future bug for) and doesn't align with the Overview page's design goal to be a "jumping off point" to other areas of the Console for more information.

The Alert Details page is the right place to send folks, and the OCS team should be made aware of this feedback about their alert messages.

yuvalgalanti:
@andybraren @spadgett the only reason we thought to currently remove this link is that users found it annoying to be taken to a page that doesn't add any more information to what they saw on the actual alert on the dashboard. this feedback came several times in the usability study we did.

andybraren:
@yuvalgalanti I agree that "the Alert Details page for some/all OCS alerts doesn't give me the extra details I'm looking for" is totally valid feedback and a good finding, but I know we've heard similar complaints about other non-OCS alerts that appear in Cluster Overview as well. A lot of alerts could be more helpful.

Helping users answer "what should I do next to resolve this alert?" is something I know the Monitoring team and @cshinn are thinking about answering within the Alert Details page somehow, and that question doesn't go away by removing the link to where that answer will eventually/hopefully be.

Even before we figure out how to provide those "suggested next steps," if what @kyoto wrote above is correct and the usability study was with 4.4, there should be at least a little more info/charts added to the alert details page since then. In today's Console I see each Alert Details page has a Metrics chart with its firing history, info about what the severity level means, the Source of the alert, the alert's labels, the relevant alerting rule, etc.

I know that extra info may not answer the user's core problem, but it's at least something that could hopefully help them more than the Overview page's alert description does.

PR #8232, opened 2021-02-24
Bug 1924536: Change link for not available state

gkarat:
Insights widget in OCP WebConsole has a state 'disabled/not available' when Insights Operator is off. There is a link 'More about Insights' that pointed to the wrong source. Preferred is `https://docs.openshift.com/container-platform/latest/support/remote_health_monitoring/using-insights-to-identify-issues-with-your-cluster.html`

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -13,7 +13,7 @@ import {
 } from './mappers';
 import { PrometheusHealthPopupProps } from '@console/plugin-sdk';
 import { K8sResourceKind } from '@console/internal/module/k8s';
-import { ExternalLink } from '@console/internal/components/utils';
+import { ExternalLink, openshiftHelpBase } from '@console/internal/components/utils';
 import './style.scss';
 
 const DataComponent: React.FC<DataComponentProps> = ({ x, y, datum }) => {
@@ -112,7 +112,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
         )}
         {!hasIssues && (isWaitingOrDisabled || isError) && (
           <ExternalLink
-            href="https://docs.openshift.com/container-platform/latest/support/getting-support.html"
+            href={`${openshiftHelpBase}support/remote_health_monitoring/using-insights-to-identify-issues-with-your-cluster.html`}
             text={t('insights-plugin~More about Insights')}
           />
         )}


gkarat:
@spadgett please take a look :)

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -112,7 +112,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
         )}
         {!hasIssues && (isWaitingOrDisabled || isError) && (
           <ExternalLink
-            href="https://docs.openshift.com/container-platform/latest/support/getting-support.html"
+            href="https://docs.openshift.com/container-platform/latest/support/remote_health_monitoring/using-insights-to-identify-issues-with-your-cluster.html"
```
spadgett, line 115: You should use `openshiftHelpBase` to link to the correct edition and version of the docs for the cluster.

https://github.com/openshift/console/blob/master/frontend/public/components/utils/documentation.tsx#L2

spadgett (changes requested):

frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
gkarat: @spadgett thanks, edited

spadgett (approved):
/lgtm

gkarat:
@spadgett I am looking at the message above... has it failed to merge? Do we need to fix anything?

spadgett:
> @spadgett I am looking at the message above... has it failed to merge? Do we need to fix anything?

Hey @gkarat, it means the commit doesn't cherry-pick to 4.7.z cleanly, so we will have to manually backport if we want to fix in 4.7. It's merged to master, which is tracking 4.8.

gkarat:
> > @spadgett I am looking at the message above... has it failed to merge? Do we need to fix anything?
> 
> Hey @gkarat, it means the commit doesn't cherry-pick to 4.7.z cleanly, so we will have to manually backport if we want to fix in 4.7. It's merged to master, which is tracking 4.8.

Could you please guide me with that manual backport? I have no idea how to do it, tbh. Do I need to open a separate PR for that?

spadgett:
Yes, if we feel this is significant enough to backport, it would require a separate PR against the release-4.7 branch. You will also need clone the original Bugzilla and set the fix version to 4.7.z. We can also decide to fix in master only if the previously link worked, just that it wasn't the best topic.

PR #8238, opened 2021-02-24
Add workspaces to Pipeline Details Page

abhinandan13jan:


abhinandan13jan:
@bgliwa01 @karthikjeeyar Unable to restore the commit in previous PR, please look into this one instead. regretting the inconvenience.

abhinandan13jan:
@karthik I've taken care og the volumeClaimTemplate thing.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
```diff
@@ -0,0 +1,88 @@
...
+          if (workspace.volumeClaimTemplate) {
+            const volumeClaimTemplate = workspace.volumeClaimTemplate as VolumeTypeClaim;
+            const { labels } = volumeClaimTemplate.metadata;
+            return (
+              labels && (
+                <VolumeClaimTemplateLink
+                  labels={labels}
+                  workspaceName={workspace.name}
+                  namespace={namespace}
+                />
```
karthikjeeyar, line 64: Unique key is missing here, causing the error in console.
```suggestion
                <VolumeClaimTemplateLink
                  key={workspace.name}
                  labels={labels}
                  workspaceName={workspace.name}
                  namespace={namespace}
                />
```

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineDetails.tsx
```diff
@@ -45,6 +46,7 @@ const PipelineDetails: React.FC<PipelineDetailsProps> = ({
             links={taskLinks}
             title={t('pipelines-plugin~Tasks')}
           />
+          <WorkspaceDefinitionList workspaces={pipeline?.spec.workspaces} />
```
karthikjeeyar, line 49: spec is not optional but workspaces is
```suggestion
          <WorkspaceDefinitionList workspaces={pipeline.spec?.workspaces} />
```

frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunCustomDetails.tsx
```diff
@@ -57,6 +58,10 @@ const PipelineRunCustomDetails: React.FC<PipelineRunCustomDetailsProps> = ({ pip
         links={renderResources}
         namespace={pipelineRun.metadata.namespace}
       />
+      <WorkspaceResourceLinkList
+        workspaces={pipelineRun?.spec.workspaces}
```
karthikjeeyar, line 62:
```suggestion
        workspaces={pipelineRun.spec?.workspaces}
```

frontend/packages/pipelines-plugin/src/components/tasks/TaskDetails.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import { SectionHeading, ResourceSummary } from '@console/internal/components/utils';
+import { TaskModel } from '../../models';
+import { TaskKind } from '../../types';
+import WorkspaceDefinitionList from '../shared/workspaces/WorkspaceDefinitionList';
+import './TaskDetails.scss';
```
karthikjeeyar, line 7: nit: new line before scss imports

frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-test-data.ts
```diff
@@ -290,3 +290,76 @@ export const mockRunDurationTest = [
   pipelineTestData[PipelineExampleNames.WORKSPACE_PIPELINE].pipelineRuns[DataState.SUCCESS],
   pipelineTestData[PipelineExampleNames.CLUSTER_PIPELINE].pipelineRuns[DataState.SUCCESS],
 ];
+
+export const pipelineWithWorkspace: PipelineKind[] = [
```
karthikjeeyar, line 294: It will be good if we reuse the existing pipeline test data from `pipelines-plugin/src/test-data/pipeline-data.ts`. This file exclusively contains different pipeline test data including one for `WORKSPACE_PIPELINE`. WDYT?

frontend/packages/pipelines-plugin/src/components/tasks/TaskDetails.tsx
```diff
@@ -0,0 +1,34 @@
...
+        text={t('pipelines-plugin~{{taskLabel}} details', {
+          taskLabel: TaskModel.label,
+        })}
+      />
+      <div className="row">
+        <div className="col-sm-6">
+          <ResourceSummary resource={task} />
+        </div>
+        <div className="col-sm-6 odc-task-details__status">
+          <WorkspaceDefinitionList workspaces={task?.spec.workspaces} />
```
karthikjeeyar, line 27:
```suggestion
          <WorkspaceDefinitionList workspaces={task.spec?.workspaces} />
```

frontend/packages/pipelines-plugin/src/components/cluster-tasks/ClusterTaskDetails.tsx
```diff
@@ -0,0 +1,34 @@
...
+        text={t('pipelines-plugin~{{clusterTaskLabel}} details', {
+          clusterTaskLabel: ClusterTaskModel.label,
+        })}
+      />
+      <div className="row">
+        <div className="col-sm-6">
+          <ResourceSummary resource={task} />
+        </div>
+        <div className="col-sm-6 odc-cluster-task-details__status">
+          <WorkspaceDefinitionList workspaces={task?.spec.workspaces} />
```
karthikjeeyar, line 27:
```suggestion
          <WorkspaceDefinitionList workspaces={task.spec?.workspaces} />
```

frontend/packages/pipelines-plugin/src/components/taskruns/TaskRunDetailsStatus.tsx
```diff
@@ -38,6 +39,10 @@ const TaskRunDetailsStatus = ({ taskRun }) => {
           </dd>
         </dl>
       )}
+      <WorkspaceResourceLinkList
+        workspaces={taskRun?.spec.workspaces}
```
karthikjeeyar, line 43:
```suggestion
        workspaces={taskRun.spec?.workspaces}
```

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
abhinandan13jan: added key

frontend/packages/pipelines-plugin/src/components/pipelineruns/detail-page-tabs/PipelineRunCustomDetails.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/tasks/TaskDetails.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-test-data.ts
abhinandan13jan: using existing PL

frontend/packages/pipelines-plugin/src/components/taskruns/TaskRunDetailsStatus.tsx
andrewballantyne: If the TaskRun is there, I'm pretty sure Spec _has_ to be there too... I don't think conditional checks are needed. You just have to support no workspaces (empty array or undefined) in the component.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
```diff
@@ -0,0 +1,89 @@
...
+  workspaces: PipelineRunWorkspace[];
+  namespace: string;
+}
+
+const WorkspaceResourceLinkList: React.FC<WorkspaceResourceLinkListProps> = ({
+  workspaces,
+  namespace,
+}) => {
+  const { t } = useTranslation();
+  return workspaces?.length > 0 ? (
```
andrewballantyne, line 24: Nit: Ideally for cleaner code, you should just return early instead of making a return into a large ternary statement.

karthikjeeyar:
@bgliwa01  For Empty Directory scenario, `workspace-name` is shown as the text in this PR, shall we add a separate bug to add the new design changes and let this PR go in. Is it fine with you?
image

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it('Should not render dd or dt heading if no workspaces are found', () => {
+    WorkspaceDefinitionListWrapperProps = {
+      workspaces: null,
+    };
+    WorkspaceDefinitionListWrapper = shallow(
+      <WorkspaceDefinitionList {...WorkspaceDefinitionListWrapperProps} />,
+    );
+    expect(WorkspaceDefinitionListWrapper.find('dd').exists()).toBe(false);
+    expect(WorkspaceDefinitionListWrapper.find('dt').exists()).toBe(false);
```
andrewballantyne, line 26: Why not look for the root element? `dl`

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it('Should render workspace list', () => {
+    const { pipeline } = pipelineTestData[PipelineExampleNames.WORKSPACE_PIPELINE];
+    WorkspaceDefinitionListWrapperProps = {
```
andrewballantyne, line 31:
```suggestion
    workspaceDefinitionListWrapperProps = {
```

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceDefinitionList.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import { PipelineWorkspace } from 'packages/pipelines-plugin/src/types';
+
+export interface WorkspaceDefinitionListProps {
+  workspaces: PipelineWorkspace[];
+}
+
+const WorkspaceDefinitionList: React.FC<WorkspaceDefinitionListProps> = ({ workspaces }) => {
```
andrewballantyne, line 9: Do we need this component? Is this not just the same component as the `WorkspaceResourceLinkList` with less bells and whistles? I _think_ they both will render a div list 🤔

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+  workspaceName,
+  namespace,
+}) => {
+  const [pvcResource, loaded, loadError] = useK8sWatchResource<K8sResourceKind>({
+    kind: PersistentVolumeClaimModel.kind,
+    isList: false,
+    selector: { matchLabels: labels },
+  });
+
+  if (loaded && !loadError && pvcResource?.metadata) {
```
andrewballantyne, line 24: Nit: for readability, you typically want to return `null`s early so that the proper render is always at the bottom. It obv doesn't impact the flow, but it helps to be consistent in a codebase this large.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+const VolumeClaimTemplateLink: React.FC<VolumeClaimTemplateLinkProps> = ({
+  labels,
+  workspaceName,
+  namespace,
+}) => {
+  const [pvcResource, loaded, loadError] = useK8sWatchResource<K8sResourceKind>({
+    kind: PersistentVolumeClaimModel.kind,
+    isList: false,
+    selector: { matchLabels: labels },
+  });
```
andrewballantyne, line 22: I think you also need to look at the owner field 🤔 

Otherwise you're going to fetch the first PVC that matches a label and it's a coin flip if you get the one you want.

Suggested new path:
- Fetch all PVCs that match the label
- Look at the metadata.ownerReferences and trim down to the PLR / TR that matches
- List only the remaining PVCs

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+      workspaces: pipeline.spec.workspaces,
+    };
+    WorkspaceDefinitionListWrapper = shallow(
+      <WorkspaceDefinitionList {...WorkspaceDefinitionListWrapperProps} />,
+    );
+    expect(WorkspaceDefinitionListWrapper.find('dd').exists()).toBe(true);
+    expect(WorkspaceDefinitionListWrapper.find('dt').exists()).toBe(true);
```
christianvogt, line 38: Testing static markup here makes the test very fragile to component changes. You can omit these tag present assertions.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceDefinitionList.tsx
abhinandan13jan: Yes, the two components need different set of props to function. They also render lists based out of different types of objects. 
There is also a need of certain additional mandatory props in ResourceLinkList which will not be required in DefinitionList

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
abhinandan13jan: thanks for the suggestion. Updated :)

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
abhinandan13jan: Updated with this logic and also added tests

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
```diff
@@ -442,3 +443,15 @@ export const getCellsFromResults = (results: TektonResultsRun[]): ResultCells =>
...
+
+type GetMatchedPVCs = (
+  pvcResources: PersistentVolumeClaimKind[],
+  ownerReference: string,
+) => PersistentVolumeClaimKind[];
+
+export const getMatchedPVCs: GetMatchedPVCs = (pvcResources, ownerReference) => {
+  return pvcResources.filter((pvc) => {
+    const { ownerReferences } = pvc.metadata;
+    return ownerReferences.find((reference) => reference.name === ownerReference);
```
karthikjeeyar, line 455: `metadata.ownerReferences` is an optional field, If you call this function with a PVC without ownerReferences chances are this will throw a error in console, so add an empty array as a fallback value and you could also add test for the same.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceResourceLinkList.spec.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react';
+import { ShallowWrapper, shallow } from 'enzyme';
+import { ResourceLink } from '@console/internal/components/utils';
+import WorkspaceResourceLinkList, {
+  WorkspaceResourceLinkListProps,
+} from '../WorkspaceResourceLinkList';
+import { taskRunWithWorkspaces } from '../../../taskruns/__tests__/taskrun-test-data';
+import { SecretModel, PersistentVolumeClaimModel, ConfigMapModel } from '@console/internal/models';
```
karthikjeeyar, line 8: nit: sort imports

abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
abhinandan13jan: updated. checking Kind as well, doing a null check and added tests

karthikjeeyar:
Verified locally, works fine. We need to match the Empty directory scenario once we have the design for it in a different story.
/lgtm

andrewballantyne (changes requested):
/lgtm cancel

I see a couple issues that will generate bugs if we leave them.

1. You make labels required for the fetching of the volumeClaimTemplate PVCs... which is not valid since a user could create a PLR without labels and it breaks down immediately. 
2. The fetching on the PLRs today will cause duplicates in references as each workspace can render n-number of PVC links based on the number of volumeClaimTemplate workspaces there are. Ie, if I have two volumeClaimTemplates, you'll get 2 PVC links for each workspace:
image

I started a conversation with Tekton on how we could use PLRs to properly fetch them, an oversight I did not see when reviewing the designs / story -- my apologizes. They think it's a bug upstream that we don't have these values on the PLR, so this may not be fixable today -- at least not directly.

I think our best option to avoid the duplicate links is to actually go through the TRs to get the values today. We may need to fix this later if the upstream code can support PLRs knowing their PVCs that get created from volumeClaimTemplates. 

Suggested path to fix (WDYT):
```
// Collect info from the PLR
data = {
    // workspaceName: pipelineRun.status.pipelineSpec.workspaces[].name
    [workspaceName]: {
        // Loop all the tasks and find the first usage of the workspace
        taskWorkspaceName: pipelineRun.status.pipelineSpec.tasks[].workspaces[].find(workspaceName === .workspace).name
        taskName: pipelineRun.status.pipelineSpec.tasks[].name

        // Match on this.taskName === taskRun[*].pipelineTaskName
        taskRunName: pipelineRun.status.taskRuns{key} // the taskRuns key
    }
}

// Fetch TRs (data[*].taskRunName)

// Get the name of the PVC from each Task Run
pvcName: taskRun.spec.workspaces.find(data[*].taskWorkspaceName === .name).persistentVolumeClaim.name
{ workspace: data{workspaceName}, pvcName }[]

// Render ResourceLinks
```

I think if we do this off of the PLR workspace name => PVC, we'll just need to fetch the first TaskRun that uses the workspace to find out what the PVC is called.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+  labels,
+  workspaceName,
+  namespace,
+  ownerReference,
+}) => {
+  const [pvcResources, loaded, loadError] = useK8sWatchResource<PersistentVolumeClaimKind[]>({
+    kind: PersistentVolumeClaimModel.kind,
+    isList: true,
+    selector: { matchLabels: labels },
+  });
```
andrewballantyne, line 25: And you need to constrain this on `namespace`

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+  let workspaceDefinitionListWrapperProps: WorkspaceDefinitionListProps;
+
+  it('Should not render root dl if no workspaces are found', () => {
+    workspaceDefinitionListWrapperProps = {
+      workspaces: null,
+    };
+    workspaceDefinitionListWrapper = shallow(
+      <WorkspaceDefinitionList {...workspaceDefinitionListWrapperProps} />,
+    );
+    expect(workspaceDefinitionListWrapper.find('dl').exists()).toBe(false);
```
andrewballantyne, line 25: I think based on Christian's response, we want this:

```suggestion
    workspaceDefinitionListWrapper = shallow(
      <WorkspaceDefinitionList {...workspaceDefinitionListWrapperProps} />,
    );
    expect(workspaceDefinitionListWrapper.isEmptyRender()).toBe(true);
```

Ref: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/isEmptyRender.html

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceResourceLinkList.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    workspaceResourceListWrapperProps = {
+      workspaces: taskRunWithWorkspaces[0].spec.workspaces,
+      namespace: 'test',
+      ownerReference: 'test',
+    };
+    workspaceResourceListWrapper = shallow(
+      <WorkspaceResourceLinkList {...workspaceResourceListWrapperProps} />,
+    );
+    expect(workspaceResourceListWrapper.find('dd').exists()).toBe(false);
+    expect(workspaceResourceListWrapper.find('dt').exists()).toBe(false);
```
andrewballantyne, line 32: You'll want to use `.isEmptyRender` here too... (and other places in this file)

andrewballantyne:
@abhinandan13jan As requested by @bgliwa01 on slack, please drop the `(workspace-name)` from all the relationships links.

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
abhinandan13jan: added namespace

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!ownerReference) return null;
+
+  if (loaded && !loadError && pvcResources?.length > 0) {
```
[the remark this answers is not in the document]
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
```diff
@@ -442,3 +443,18 @@ export const getCellsFromResults = (results: TektonResultsRun[]): ResultCells =>
...
+type GetMatchedPVCs = (
+  pvcResources: PersistentVolumeClaimKind[],
+  ownerReference: string,
+) => PersistentVolumeClaimKind[];
+
+export const getMatchedPVCs: GetMatchedPVCs = (pvcResources, ownerReference) => {
+  return pvcResources.filter((pvc) => {
+    const { ownerReferences = [] } = pvc.metadata;
+
+    return ownerReferences.find(
```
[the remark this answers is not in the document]
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-utils.spec.ts
```diff
@@ -193,6 +196,20 @@ describe('pipeline-utils ', () => {
...
+    expect(matchedPVCs[0].metadata.name).toBe(pvcWithPipelineOwnerRef[1].metadata.name);
+    expect(matchedPVCs[1].metadata.name).toBe(pvcWithPipelineOwnerRef[3].metadata.name);
+  });
+
+  it('should not match PVCs when ownerRef matches name but not kind', () => {
+    const matchedPVCs = getMatchedPVCs(pvcWithPipelineOwnerRef, 'pipeline2', PipelineRunModel.kind);
+    expect(matchedPVCs.length).toBe(2);
+    expect(matchedPVCs[0].metadata.name).toBe(pvcWithPipelineOwnerRef[1].metadata.name);
+    expect(matchedPVCs[1].metadata.name).toBe(pvcWithPipelineOwnerRef[3].metadata.name);
+  });
```
andrewballantyne, line 211: Same tests with different expectations? 😕 

image

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceResourceLinkList.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+    ...reactI18next,
+    useTranslation: () => ({ t: (key) => key }),
+  };
+});
+
+describe('WorkspaceResourceLinkList', () => {
+  let workspaceResourceListWrapper: ShallowWrapper<WorkspaceResourceLinkListProps>;
+  let workspaceResourceListWrapperProps: WorkspaceResourceLinkListProps;
+
+  it('Should not render dd or dt heading if no workspaces are found', () => {
```
andrewballantyne, line 22: Good use case for not explaining the expected output, speak to the reason for the test.

`it('should not render if there are no workspaces'`

frontend/packages/pipelines-plugin/src/components/shared/workspaces/__tests__/WorkspaceDefinitionList.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    ...reactI18next,
+    useTranslation: () => ({ t: (key) => key }),
+  };
+});
+
+describe('WorkspaceDefinitionList', () => {
+  let workspaceDefinitionListWrapper: ShallowWrapper<WorkspaceDefinitionListProps>;
+  let workspaceDefinitionListWrapperProps: WorkspaceDefinitionListProps;
+
+  it('Should not render root dl if no workspaces are found', () => {
```
andrewballantyne, line 18: `root dl`?

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+  if (!matchedPVCs || matchedPVCs.length === 0) return null;
+
+  return (
+    <>
+      {matchedPVCs.map((pvcResource) => {
+        return (
+          <ResourceLink
+            name={pvcResource.metadata.name}
+            key={pvcResource.metadata.name}
+            namespace={namespace}
```
andrewballantyne, line 41: Nit: Technically this should be `pvcResource.metadata.namespace` -- the logic around this makes the two match but your namespace needs to be that of the `pvcResource` not of your PLR/TR

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
```diff
@@ -0,0 +1,86 @@
...
+              />
+            );
+          }
+          return <div key={workspace.name}>{workspace.name}</div>;
+        })}
+        <VolumeClaimTemplatesLink
+          namespace={namespace}
+          ownerResourceName={ownerResourceName}
+          ownerResourceKind={ownerResourceKind}
+        />
```
andrewballantyne, line 80: @abhinandan13jan Please test before you request a review... You are double rendering...

image

Or if you prefer, with my two `volumeClaimTemplate` example from before:
image

andrewballantyne (changes requested):

andrewballantyne:
If you need access to my resources, I posted as part of the upstream request for them to add volumeClaimTemplate PVC names to the `.status`. See the upstream issue: https://github.com/tektoncd/pipeline/issues/3800

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
abhinandan13jan: My PLR YAML, completely missed it. I've updated and tested against these now. Thanks a ton for the check

Screenshot from 2021-03-04 02-48-29
Screenshot from 2021-03-04 02-49-04

frontend/packages/pipelines-plugin/src/components/taskruns/TaskRunDetailsStatus.tsx
```diff
@@ -7,13 +7,16 @@ import { taskRunFilterReducer } from '../../utils/pipeline-filter-reducer';
...
 import { TaskRunKind } from '../../types';
+import WorkspaceResourceLinkList from '../shared/workspaces/WorkspaceResourceLinkList';
 
 export interface TaskRunDetailsStatusProps {
   taskRun: TaskRunKind;
 }
 
 const TaskRunDetailsStatus = ({ taskRun }) => {
   const { t } = useTranslation();
+  const ownerReference = taskRun.metadata.ownerReferences?.[0] || taskRun.metadata.name;
```
karthikjeeyar, line 18: `taskRun.metadata.ownerReferences?.[0]` This one is an object which is passed to `ownerResourceName` in `WorkspaceResourceLinkList` below. Shouldn't this be `taskRun.metadata.ownerReferences?.[0].name`  ?

karthikjeeyar: Thinking about this more, do we need this at all? TaskRun's ownerRef will be PipelineRun, but for `ownerResourceKind` is tied to taskrun.kind always, so it will never match :thinking: or am i missing something here?

ie: 
```
   <WorkspaceResourceLinkList
        workspaces={taskRun.spec.workspaces}
        namespace={taskRun.metadata.namespace}
        ownerResourceName={ownerReference} // <PipelineRun-name> (if taskrun is NOT standalone)
        ownerResourceKind={taskRun.kind} // TaskRun
      />
```

karthikjeeyar:
Workspace section is displayed without any list of workspaces for standalone Taskrun uses volumeClaimTemplate (failed due to errors):

image

frontend/packages/pipelines-plugin/src/components/taskruns/TaskRunDetailsStatus.tsx
abhinandan13jan: yes, u r correct. I misplaced the oenerReference
```
      <WorkspaceResourceLinkList
        workspaces={taskRun.spec.workspaces}
        namespace={taskRun.metadata.namespace}
        ownerResourceName={ownerReference ? ownerReference.name : taskRun.metadata.name}
        ownerResourceKind={ownerReference ? ownerReference.kind : taskRun.kind}
      />
```

karthikjeeyar:
@abhinandan13jan  You still need to fix the issue from my last comment. PTAL

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <dd>
+          {matchedPVCs.map((pvcResource) => {
+            return (
+              <ResourceLink
+                name={pvcResource.metadata.name}
+                key={pvcResource.metadata.name}
+                namespace={pvcResource.metadata.namespace}
+                kind={PersistentVolumeClaimModel.kind}
+              />
+            );
```
karthikjeeyar, line 50: Instead of repeating this block twice, it would be nice if you could add it to a const and contidionally render it based on the `showHeader`.

karthikjeeyar:
Verified, everything else looks good apart from my last minor comment, I will add lgtm once that is fixed.

debsmita1:
I tried out this yaml https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/workspace-from-volumeclaimtemplate.yaml

and I can see the same PVC listed twice
workspaces

frontend/packages/pipelines-plugin/src/components/shared/workspaces/VolumeClaimTemplateLink.tsx
abhinandan13jan: updated

abhinandan13jan:
> I tried out this yaml https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/workspace-from-volumeclaimtemplate.yaml
> 
> and I can see the same PVC listed twice
> workspaces

@debsmita1 @karthikjeeyar @andrewballantyne This happens if we use pipelineRun as the ownerReference for taskRun, which I did because of a review comment. I've reverted that particular change and it works fine now.

karthikjeeyar:
Thanks for all the changes @abhinandan13jan 
/lgtm

debsmita1:
works fine. Thanks !

andrewballantyne (approved):

PR #8244, opened 2021-02-24
Add pluginType to LoadedExtension interface

vojtechszocs:
This PR modifies the `LoadedExtension` interface in the following way:
- add `pluginType` (static or dynamic)
- remove `pluginName` since it's already part of `pluginID` (note: static plugins don't have a version)

This makes unit test assertions more sensible (better distinction between extensions) and allows us to act upon extension's `pluginType` in the future.


vojtechszocs:
After discussing with @christianvogt there isn't a concrete use case for this at the moment, closing.

PR #8247, opened 2021-02-25
Bug 1931279: Bump lodash-es to latest v4.17.21, to remove security vulnerabilities

jhadvig:
Fixes both:
- https://bugzilla.redhat.com/show_bug.cgi?id=1931279
- https://bugzilla.redhat.com/show_bug.cgi?id=1931272

since both of these issues fixes include bumping `lodash-es` to version `4.17.21`

/assign @spadgett PTAL

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -144,7 +144,7 @@
     "js-base64": "^2.5.1",
     "js-yaml": "^3.13.1",
     "json-schema": "0.2.3",
-    "lodash-es": "4.17.x",
+    "lodash-es": "^4.17.21",
     "monaco-languageclient": "^0.13.0",
     "murmurhash-js": "1.0.x",
     "null-loader": "^3.0.0",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -144,7 +144,7 @@
     "js-base64": "^2.5.1",
     "js-yaml": "^3.13.1",
     "json-schema": "0.2.3",
-    "lodash-es": "4.17.x",
+    "lodash-es": "4.17.21",
```
spadgett, line 147: We should use 4.7.21 or newer

```suggestion
    "lodash-es": "^4.17.21",
```

spadgett (changes requested):

jhadvig:
@spadgett comment addressed

spadgett (commented):
It looks like we have an earlier version of lodash coming from Victory charts via Patternfly as well that we might need to override. Might be easiest to add it to the `resolutions` in package.json.

jhadvig:
@spadgett done :)

spadgett (approved):
/lgtm

PR #5831, opened 2020-06-26
Bug 1839101: Namespaced navigation item to non-resource pages

fmartingr:
## Summary
Currently the navigation from the console works well when the target comes from a resource in the cluster, but some views that depends on namespaces that are not resources (like topology) use direct links that need to handle the namespace the user is on internally.

This causes a mild bug on the console when you have two tabs open on different namespaces, since these links use the local storage to determine the namespace the user is on and that would be the last one the user accessed.

## Proposal

Created a new navigation item that operates using a name and a namespace in the form of: `<view>/ns/<namespace>` which is the format of the navigation items replaced with this.

## Pending
- [ ] Naming for the component (Used `ViewNS`).
- [ ] Paths for these kind of pages (Used `/<name>/ns/<namespace>`) but not sure all pages use this.
- [ ] Path generation requires a namespace and not all pages may support `all-namespaces`.

## Notes

- I tried to follow the guidelines as shown in the project and the docs, if I missed something please let me know.
- I have full-stack experience on several languages, not a lot with React though, but eager to learn.

PR #8249, opened 2021-02-25
Bug 1839101: Add namespaced attribute to main navigation links

jerolimov:
**Fixes**:
https://issues.redhat.com/browse/ODC-3976
https://bugzilla.redhat.com/show_bug.cgi?id=1839101

Based on PR #5831 from @fmartingr - Big thanks for this work!!

**Analysis / Root cause**: 
Currently the navigation from the console works well when the target comes from a resource in the cluster, but some views that depends on namespaces that are not resources (like topology) use direct links that need to handle the namespace the user is on internally.

This causes a mild bug on the console when you have two tabs open on different namespaces, since these links use the local storage to determine the namespace the user is on and that would be the last one the user accessed.

**Solution Description**: 
Created a new navigation item that operates using a name and a namespace in the form of: <view>/ns/<namespace> which is the format of the navigation items replaced with this.

Before:
https://user-images.githubusercontent.com/139310/109855919-29ca7600-7c59-11eb-9c82-863ab0ab445b.mp4

After:
https://user-images.githubusercontent.com/139310/109855939-3058ed80-7c59-11eb-92b4-07dbcc47b53d.mp4

Tests not touched

**Test setup:**
* Open console in two tabs
* Navigate to project A in tab A
* Navigate to another project in tab B (switch project for example)
* Open navigation links in tab A which should open project A (opens latest used project from user settings before)

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/nav-items.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/nav-items.ts
@@ -38,7 +38,8 @@ namespace ExtensionProperties {
   }
 
   export interface HrefNavItem extends NavItem {
-    componentProps: NavItem['componentProps'] & Pick<HrefLinkProps, 'href' | 'activePath'>;
+    componentProps: NavItem['componentProps'] &
+      Pick<HrefLinkProps, 'namespaced' | 'href' | 'activePath'>;
   }
 
   export interface ResourceNSNavItem extends NavItem {
--- a/frontend/public/components/nav/admin-nav.tsx
+++ b/frontend/public/components/nav/admin-nav.tsx
@@ -127,6 +127,7 @@ const AdminNav = () => {
         <HrefLink
           id="search"
           href="/search"
+          namespaced
           name={t('public~Search')}
           startsWith={searchStartsWith}
         />
@@ -268,6 +269,7 @@ const AdminNav = () => {
         <HrefLink
           id="provisionedservices"
           href="/provisionedservices"
+          namespaced
           name={t('public~Provisioned Services')}
           activePath="/provisionedservices/"
           startsWith={provisionedServicesStartsWith}
--- a/frontend/public/components/nav/items.tsx
+++ b/frontend/public/components/nav/items.tsx
@@ -11,7 +11,10 @@ import {
   isResourceNSNavItem,
   isResourceClusterNavItem,
 } from '@console/plugin-sdk';
-import { formatNamespacedRouteForResource } from '@console/shared/src/utils';
+import {
+  formatNamespacedRouteForResource,
+  formatNamespacedRouteForHref,
+} from '@console/shared/src/utils';
 import { referenceForModel, K8sKind } from '../../module/k8s';
 import { stripBasePath } from '../utils';
 import { featureReducerName } from '../../reducers/features';
@@ -142,7 +145,11 @@ export class HrefLink extends NavLink<HrefLinkProps> {
   }
 
   get to() {
-    return this.props.href;
+    const { href, namespaced, activeNamespace } = this.props;
+    if (namespaced) {
+      return formatNamespacedRouteForHref(href, activeNamespace);
+    }
+    return href;
   }
 }
 
@@ -175,6 +182,8 @@ export type ResourceClusterLinkProps = NavLinkProps & {
 
 export type HrefLinkProps = NavLinkProps & {
   href: string;
+  namespaced?: boolean;
+  activeNamespace?: string;
 };
 
 export type NavLinkComponent<T extends NavLinkProps = NavLinkProps> = React.ComponentType<T> & {
--- a/frontend/public/components/nav/perspective-nav.tsx
+++ b/frontend/public/components/nav/perspective-nav.tsx
@@ -111,12 +111,10 @@ const PerspectiveNav: React.FC<{}> = () => {
           const { id, name } = item.properties;
           return <NavSection id={id} title={name} key={id} isGrouped={!name} />;
         }
-        if (isNavItem(item)) {
-          return createLink(item, true);
-        }
         if (isSeparatorNavItem(item)) {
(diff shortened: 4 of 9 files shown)


rottencandy (commented):
Tested locally, looks good.
/lgtm

frontend/public/components/nav/items.tsx
```diff
@@ -175,6 +182,8 @@ export type ResourceClusterLinkProps = NavLinkProps & {
 
 export type HrefLinkProps = NavLinkProps & {
   href: string;
+  namespaced: boolean;
```
christianvogt, line 185: This should be optional. Defaults to `false`.
This would reduce the overall size of this PR since a lot of changes involve setting this value to `false`.
```suggestion
  namespaced?: boolean;
```

frontend/public/components/nav/admin-nav.tsx
```diff
@@ -78,6 +78,7 @@ const MonitoringNavSection_ = ({ canAccess }) => {
           id="monitoringalerts"
           href="/monitoring/alerts"
           name={t('public~Alerting')}
+          namespaced={false}
```
christianvogt, line 81: Default value for `namespaced` should be `false` and therefore this change won't be necessary.
```suggestion
```

frontend/packages/console-plugin-sdk/src/typings/nav-items.ts
```diff
@@ -87,10 +88,5 @@ export const isSeparatorNavItem = (e: Extension): e is SeparatorNavItem => {
 };
 
 export const isNavItem = (e: Extension): e is NavItem => {
-  return (
-    isHrefNavItem(e) ||
-    isResourceNSNavItem(e) ||
-    isResourceClusterNavItem(e) ||
-    isSeparatorNavItem(e)
```
christianvogt, line 94: Why are you removing this check? `isNavItem` is supposed to include all nav item extensions.

I do see an issue in `perspective-nav.tsx` though:

```
        if (isNavItem(item)) {
          return createLink(item, true);
        }
        if (isSeparatorNavItem(item)) {
          return <NavItemSeparator key={`separator-${index}`} />;
        }
```

Should be:
```
        if (isSeparatorNavItem(item)) {
          return <NavItemSeparator key={`separator-${index}`} />;
        }
        return createLink(item, true);
```

frontend/public/components/nav/admin-nav.tsx
jerolimov: Ack. Solved

frontend/packages/console-plugin-sdk/src/typings/nav-items.ts
jerolimov: We can keep this as it was before. I just take this from the origin request but don't see a difference.

jerolimov: Applied your recommendation and rebased everything.

jerolimov:
Removed all unnecessary `namespaced: false` and `namespaced={false}` values after making the new `namespaced` prop optional. Please take a new look.

frontend/public/components/nav/perspective-nav.tsx
```diff
@@ -111,12 +111,12 @@ const PerspectiveNav: React.FC<{}> = () => {
...
         }
-        if (isNavItem(item)) {
-          return createLink(item, true);
-        }
         if (isSeparatorNavItem(item)) {
           return <NavItemSeparator key={`separator-${index}`} />;
         }
+        if (isNavItem(item)) {
+          return createLink(item, true);
+        }
```
rohitkrai03, line 119: @christianvogt's suggestion was to remove the check for `isNavItem` and directly return `createLink` in the end since `isNavItem` would be true for all nav extensions.

```suggestion
        if (isNavItem(item)) {
          return createLink(item, true);
        }
```

rohitkrai03 (changes requested):

jerolimov:
Rebased and just fixed the last comment of Rohit https://github.com/openshift/console/pull/8249#pullrequestreview-616537370

rohitkrai03 (approved):
/approve

rottencandy (approved):
/lgtm

glekner:
/hold

this conflicts with https://github.com/openshift/console/pull/8438
lets not merge code thats going to be deleted anyway.

rohitkrai03:
I agree. Since we are already migrating to dynamic plugin extensions, all the static extensions and their usage are going to be deleted anyway. Let's work to get these changes in the dynamic extensions PR.

andrewballantyne:
This PR seems to have stalled for 22 days heh. Are we looking to keep this or merge it in with another PR?

cc @rohitkrai03 @jerolimov

jerolimov:
@andrewballantyne We wait for #8438. I'm super happy to close this here, but I don't want do this without the other one is merged.

@glekner will you have time to work on, or at least rebase #8438 soon or should we merge this one here and hope that your PR do not lose this feature again?

andrewballantyne:
@glekner @jerolimov We definitely should resolve this one way or another. If we are not pushing for the dependant PR to get merged, we should either close this (and make it part of the dependant PR down the road) or merge this without it. ODC shouldn't stall a PR for a month because of a dependant. We need to get better at managing our tickets and PRs, and this is going to drag our metrics down.

I'll leave it up to you guys, if you want to try to merge this in the intended order, please make concerted efforts to do so. If not, we need to look at what to do about this PR for our own team. Bugs impact the quality of where our team is at through the new SLO metrics, and this PR is weighing down the team because of its existence. So something needs to happen here.

cc @nimishamukherjee @christianvogt @debsmita1

glekner:
/unhold

sorry for the late response as I was busy with bugs. let's merge this one, ill deal with the rebases.

jerolimov:
All fine, thanks Gilad!

PR #8250, opened 2021-02-25
Bug 1852743: Use consistent labels for CPU in node list

spadgett:
The other metrics show the label for both values before and after the `/`.

Screen Shot 2021-02-25 at 1 27 35 PM

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #8252, opened 2021-02-25
Update console-extensions.json format

vojtechszocs:
### Before

```jsonc
{
  "$schema": "/path/to/console-dynamic-plugin-sdk/dist/schema/console-extensions.json",
  "data": [ /* extension declarations */ ]
}
```

### After

```jsonc
[ /* extension declarations */ ]
```

According to current [JSON Schema specification](http://json-schema.org/specification.html), the [`$schema`](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.1.1) keyword value (representing a meta-schema) must be a URI and when requested, its media type should be `application/schema+json`.

We should stick to the standards; having a top-level `$schema` property in a JSON document is apparently not the standard way to associate such document with the given schema.

@christianvogt Once we support "plugin extends plugin" scenario, the JSON schema will become dynamic, i.e. composed from core Console extension schema & extension schemas from dependent plugin(s). In this scenario, having a single static `$schema` property isn't very helpful anyway.

### DIFF
--- a/frontend/.vscode/settings.json
+++ b/frontend/.vscode/settings.json
@@ -59,6 +59,12 @@
   "files.associations": {
     "**/console-extensions.json": "jsonc"
   },
+  "json.schemas": [
+    {
+      "fileMatch": ["**/console-extensions.json"],
+      "url": "./packages/console-dynamic-plugin-sdk/dist/schema/console-extensions.json"
+    }
+  ],
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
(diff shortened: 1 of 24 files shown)


frontend/.vscode/settings.json
```diff
@@ -59,6 +59,12 @@
   "files.associations": {
     "**/console-extensions.json": "jsonc"
   },
+  "json.schemas": [
+    {
+      "fileMatch": ["**/console-extensions.json"],
+      "url": "./packages/console-dynamic-plugin-sdk/dist/schema/console-extensions.json"
```
vojtechszocs, line 65: Reference: https://code.visualstudio.com/docs/languages/json#_mapping-to-a-schema-in-the-workspace

The `./` prefix represents the workspace root.

vojtechszocs:
/hold

Need to fix failing unit test.

PR #8253, opened 2021-02-25
Bug 1929243: override normal namespace hiding for pods view in node details

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1929243

override-namespace


rhamilto:
/hold

probably want to hide `The namespace column is only shown when in "All projects"` in the columns modal for this context.

jcaianirh:
@rhamilto hid the alert text in the col management modal for the override use case

PR #8259, opened 2021-03-01
 Bug 1924641: Remove error message for "missing storage class"

rexagod:
Mentioning here that I was able to replicate this issue only a couple
of times (as is mentioned in the BZ issue body), so this patch fix
addresses some parts of the codebase that I inferred to be the cause
of this issue from that observation.

It'd be nice if someone (who can replicate the issue successfully) can
apply this patch and provide their feedback on it.

***
To apply this, do `curl https://github.com/openshift/console/commit/f12891f445c73f0041c397813e9dedbeb36b8fe8.patch | git apply`.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/locales/en/ceph-storage-plugin.json
+++ b/frontend/packages/ceph-storage-plugin/locales/en/ceph-storage-plugin.json
@@ -331,8 +331,6 @@
   "Discover Disks": "Discover Disks",
   "Storage and Nodes": "Storage and Nodes",
   "Review and Create": "Review and Create",
-  "Missing storage class": "Missing storage class",
-  "The storage cluster needs to use a storage class to consume the local storage. In order to create one you need to discover the available disks and create a storage class using the filters to select the disks you wish to use": "The storage cluster needs to use a storage class to consume the local storage. In order to create one you need to discover the available disks and create a storage class using the filters to select the disks you wish to use",
   "Can be used on any platform. It means that OCS uses attached disks, via Local Storage Operator. In this case, the infrastructure storage class is actually provided by LSO, on top of attached drives.": "Can be used on any platform. It means that OCS uses attached disks, via Local Storage Operator. In this case, the infrastructure storage class is actually provided by LSO, on top of attached drives.",
   "Before we can create a storage cluster, the local storage operator needs to be installed. When installation is finished come back to OpenShift Container Storage to create a storage cluster.<1><0>Install</0></1>": "Before we can create a storage cluster, the local storage operator needs to be installed. When installation is finished come back to OpenShift Container Storage to create a storage cluster.<1><0>Install</0></1>",
   "Role": "Role",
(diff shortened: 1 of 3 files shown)


rexagod:
Pasting this in for reference: https://bugzilla.redhat.com/show_bug.cgi?id=1931543

rexagod:
To add to the aforementioned information, I can confirm that this patch basically checks when the provSC gets filtered in the storage classes dropdown, for the noProv storage class, and invokes the noProv [detection and updation process](https://github.com/rexagod/console/blob/f12891f445c73f0041c397813e9dedbeb36b8fe8/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx#L62).

rexagod:
PS. To apply this, do `curl https://github.com/openshift/console/commit/f12891f445c73f0041c397813e9dedbeb36b8fe8.patch | git apply`.

afreen23 (changes requested):
> Mentioning here that I was able to replicate this issue only a couple
> of times (as is mentioned in the BZ issue body), so this patch fix
> addresses some parts of the codebase that I inferred to be the cause
> of this issue from that observation.
> 
> It'd be nice if someone (who can replicate the issue successfully) can
> apply this patch and provide their feedback on it.
> 
> To apply this, do `curl https://github.com/openshift/console/commit/f12891f445c73f0041c397813e9dedbeb36b8fe8.patch | git apply`.

The issue is due to the fact that when we create storage class, we dont fetch results again to update that.
On a new mount this is not the case and storage class is found and hence the correct display.
We need to fix this issue by fetching storage class in wizard and looking if a no  provsc is there, that means moving no Prov sc to wizard component.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/create-sc/wizard-pages/create-local-volume-set.tsx
```diff
@@ -167,7 +177,14 @@ const ConfirmationModal = ({ state, dispatch, setInProgress, setErrorMessage, ns
...
             variant="small"
             actions={[
-              <Button key="confirm" variant="primary" onClick={makeLVSCall}>
+              <Button
+                key="confirm"
+                variant="primary"
+                onClick={() => {
+                  makeLVSCall();
+                  sendDetectionReq(true);
+                }}
```
bipuladh, line 186: move it out toa function maybe call it `onCreate`

rexagod:
Here a visual of the changes.

https://user-images.githubusercontent.com/33557095/110671526-9d89f700-81f4-11eb-9b66-c78dba97f4a2.mp4

afreen23 (changes requested):
Run `yarn i18n` and update texts

rexagod:
Updating the i18n files.

rexagod:
Found a bug, updating this ASAP.

rexagod:
I've removed hasNoProvSC and setHasNoProvSC entirely from the codebase since we aren't using those anywhere.

rexagod:
Fixing conflicts.

afreen23 (approved):
/lgtm

PR #8260, opened 2021-03-01
Upgrade xterm v3.12.0 to v4.10.0

sahil143:
**Fixes**: https://issues.redhat.com/browse/ODC-5547

**Analysis / Root cause**: To send commands programmatically to the terminal we need feature(addons) from the xterm.js v4.10.0

**Solution Description**: upgrade xterm version

**Test setup:**

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/utils/xterm-addon-fullscreen.ts
+++ b/frontend/packages/console-shared/src/utils/xterm-addon-fullscreen.ts
@@ -1,0 +1,30 @@
+import { Terminal, ITerminalAddon } from 'xterm';
+import './xterm-addon-fullscreen.scss';
+
+export class XtermAddonFullscreen implements ITerminalAddon {
+  private terminal: Terminal | undefined;
+
+  public activate(terminal: Terminal): void {
+    this.terminal = terminal;
+  }
+
+  public dispose(): void {
+    if (this.terminal?.element.classList.contains('fullscreen')) {
+      this.terminal?.element.classList.remove('fullscreen');
+    }
+  }
+
+  public toggleFullScreen(fullscreen?: boolean): void {
+    let fn: string;
+
+    if (typeof fullscreen === 'undefined') {
+      fn = this.terminal?.element.classList.contains('fullscreen') ? 'remove' : 'add';
+    } else if (!fullscreen) {
+      fn = 'remove';
+    } else {
+      fn = 'add';
+    }
+
+    this.terminal?.element.classList[fn]('fullscreen');
+  }
+}
--- a/frontend/public/components/terminal.jsx
+++ b/frontend/public/components/terminal.jsx
@@ -1,14 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Terminal as XTerminal } from 'xterm';
-import * as fit from 'xterm/lib/addons/fit/fit';
-import * as full from 'xterm/lib/addons/fullscreen/fullscreen';
+import { FitAddon } from 'xterm-addon-fit';
+import { withTranslation } from 'react-i18next';
 import { CompressIcon } from '@patternfly/react-icons';
 import { Button } from '@patternfly/react-core';
-import { withTranslation } from 'react-i18next';
-
-XTerminal.applyAddon(fit);
-XTerminal.applyAddon(full);
+import { XtermAddonFullscreen } from '@console/shared';
 
 class Terminal_ extends React.Component {
   constructor(props) {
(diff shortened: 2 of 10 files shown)


frontend/packages/console-shared/src/utils/xterm-addon-fullscreen.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (this.terminal.element.classList.contains('fullscreen')) {
+      this.terminal.element.classList.remove('fullscreen');
+    }
+  }
+
+  public toggleFullScreen(fullscreen: boolean): void {
+    let fn: string;
+
+    if (typeof fullscreen === 'undefined') {
+      fn = this.terminal.element.classList.contains('fullscreen') ? 'remove' : 'add';
```
invincibleJai, line 21: will `this.terminal` doesn't need any check ? type of `terminal` has even `undefined`

frontend/packages/console-shared/src/utils/xterm-addon-fullscreen.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class XtermAddonFullscreen implements ITerminalAddon {
+  private terminal: Terminal | undefined;
+
+  public activate(terminal: Terminal): void {
+    this.terminal = terminal;
+  }
+
+  public dispose(): void {
+    if (this.terminal && this.terminal.element.classList.contains('fullscreen')) {
```
invincibleJai, line 12: instead of checking for `this.terminal` here and below, will it make sense to just have

```suggestion
    if (his.terminal?.element?.classList?.contains('fullscreen')) {
```

frontend/public/components/terminal.jsx
```diff
@@ -1,14 +1,11 @@
...
-import * as full from 'xterm/lib/addons/fullscreen/fullscreen';
+import { FitAddon } from 'xterm-addon-fit';
+import { withTranslation } from 'react-i18next';
 import { CompressIcon } from '@patternfly/react-icons';
 import { Button } from '@patternfly/react-core';
-import { withTranslation } from 'react-i18next';
-
-XTerminal.applyAddon(fit);
-XTerminal.applyAddon(full);
+import { XtermAddonFullscreen } from '@console/shared';
```
rohitkrai03, line 8: Why do we need custom implementation of fullscreen addon? Did xterm remove support for their fullscreen addon?

christianvogt: iirc @sahil143 said they removed this addon. I'm not sure if they replaced or why it was removed though.

jerolimov:
@sahil143 I found two issues:

a) The small one is the "black gap" on the bottom right corner below the scrollbar:

image

b) It looks like the page resize depends on the full screen height instead of the window height. I have two scrollbars. The one for the terminal is fine but the other one shows a white space below the terminal:

https://user-images.githubusercontent.com/139310/110600299-11d88200-8184-11eb-8783-dd6d315e4465.mp4

frontend/public/components/terminal.jsx
rohitkrai03: Yeah, seems like they removed fullscreen addon - https://github.com/xtermjs/xterm.js/pull/2065#issuecomment-491614794

sahil143:
@jerolimov I'm not able to reproduce this on chrome version 89. Which version are you on?
xterm-up

invincibleJai:
I was able to reproduce in chrome 88 but then I upgraded to 89 and can't see dual scroll anymore.

jerolimov:
I tested this on Chrome 88 (on Fedora 33), there is an update available for 89. Hmmm.

sahil143:
Took some time but finally figured out the issue. The main issue here is with the `xterm-helper-textarea` it has a position absolute and top, left calculated. When there is content in the terminal the text area is at the bottom but when the browser is resized, we update the height and width of the container/outerRef div then xterm updates the dimensions using the fit addon to fit into the container/outerRef but the `xterm-helper-textarea` stays at the bottom and its position doesn't update that's why the scrollbar appears. The issue fixes itself when the user starts to type again in the terminal.

Kapture 2021-03-11 at 21 43 36

cc @jerolimov @invincibleJai

sahil143:
There is already an open issue on xterm.js repo with similar behaviour https://github.com/xtermjs/xterm.js/issues/3065

jerolimov (approved):
> There is already an open issue on xterm.js repo with similar behaviour xtermjs/xterm.js#3065

Would be cool to fix it there and keep this bug in mind. I will create a new issue for this so that we can upgrade the dependency now for your depending work.

/lgtm

PR #8265, opened 2021-03-01
Revert "Bug 1866322: Add prop to AlertItem to hide default link"

spadgett:
Reverting this change based on discussion in #8231.

/assign @bipuladh @kyoto @andybraren

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
+++ b/frontend/packages/ceph-storage-plugin/src/selectors/index.ts
@@ -11,14 +11,8 @@ const enum status {
   BOUND = 'Bound',
   AVAILABLE = 'Available',
 }
-export const filterCephAlerts = (alerts: Alert[]): Alert[] => {
-  const rookRegex = /.*rook.*/;
-  return alerts?.filter(
-    (alert) =>
-      alert?.annotations?.storage_type === 'ceph' ||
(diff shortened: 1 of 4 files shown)


frontend/packages/ceph-storage-plugin/src/selectors/index.ts
```diff
@@ -11,14 +11,8 @@ const enum status {
...
-export const filterCephAlerts = (alerts: Alert[]): Alert[] => {
-  const rookRegex = /.*rook.*/;
-  return alerts?.filter(
-    (alert) =>
-      alert?.annotations?.storage_type === 'ceph' ||
-      Object.values(alert?.labels)?.some((item) => rookRegex.test(item)),
-  );
-};
+export const filterCephAlerts = (alerts: Alert[]): Alert[] =>
+  alerts.filter((alert) => _.get(alert, 'annotations.storage_type') === 'ceph');
```
andybraren, line 15: The inclusion of Ceph alerts in the Storage Overview might be one thing from the original PR that could stay?

bipuladh: Yes let's keep this part as it is.

zherman0 (approved):

bipuladh:
@spadgett https://github.com/openshift/console/pull/8569 incorporates this change. Let's continue with 8569.

PR #8273, opened 2021-03-02
Monitoring dashboards: Allow zooming all graphs and sync their ranges

kyoto:
Allow using the mouse to zoom in on any monitoring dashboard graph. When
doing this, update all other graphs on the page to display the same time
range.

### DIFF
--- a/frontend/public/components/monitoring/dashboards/graph.tsx
+++ b/frontend/public/components/monitoring/dashboards/graph.tsx
@@ -1,36 +1,50 @@
 import * as React from 'react';
-import { connect } from 'react-redux';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useDispatch, useSelector } from 'react-redux';
 
+import {
+  monitoringDashboardsSetEndTime,
+  monitoringDashboardsSetTimespan,
+} from '../../../actions/ui';
 import { RootState } from '../../../redux';
 import { FormatLegendLabel, QueryBrowser } from '../query-browser';
-
-const Graph_: React.FC<Props> = ({
-  formatLegendLabel,
-  isStack,
-  pollInterval,
-  queries,
-  timespan,
-}) => (
-  <QueryBrowser
-    defaultSamples={30}
-    formatLegendLabel={formatLegendLabel}
-    hideControls
-    isStack={isStack}
-    pollInterval={pollInterval}
-    queries={queries}
-    timespan={timespan}
-  />
-);
-const Graph = connect(({ UI }: RootState) => ({
-  timespan: UI.getIn(['monitoringDashboards', 'timespan']),
-}))(Graph_);
 
 type Props = {
   formatLegendLabel?: FormatLegendLabel;
   isStack: boolean;
   pollInterval: number;
   queries: string[];
-  timespan: number;
+};
+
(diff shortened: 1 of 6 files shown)


frontend/public/components/monitoring/dashboards/graph.tsx
```diff
@@ -1,36 +1,69 @@
...
+      onZoom={onZoom}
+      pollInterval={pollInterval}
+      queries={queries}
+      timespan={timespan}
+    />
+  );
 };
 
-export default Graph;
+export default connect<StateProps, DispatchProps, OwnProps>(
```
spadgett, line 60: Consider `useSelector` and `useDispatch` instead of `connect`.

https://react-redux.js.org/next/api/hooks

frontend/public/components/monitoring/dashboards/graph.tsx
```diff
@@ -1,36 +1,69 @@
...
-  formatLegendLabel?: FormatLegendLabel;
-  isStack: boolean;
-  pollInterval: number;
-  queries: string[];
-  timespan: number;
+}) => {
+  const onZoom = React.useCallback(
+    (from, to) => {
+      setEndTime(to);
+      setTimespan(to - from);
```
spadgett, line 40: Since these values are stored in Redux, I'm assuming they will be remembered if I leave the dashboards and come back? I'm trying to decide if that's what we want or if we should try to clear them.

frontend/public/components/monitoring/query-browser.tsx
```diff
@@ -731,10 +740,13 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
...
   }
 
-  const onZoom = (from: number, to: number) => {
+  const zoomableGraphOnZoom = (from: number, to: number) => {
     setXDomain([from, to]);
     setSpan(to - from);
     setSamples(defaultSamples || getMaxSamplesForSpan(to - from));
+    if (_.isFunction(onZoom)) {
+      onZoom(from, to);
+    }
```
spadgett, line 749: Maybe just use optional chaining. Presumably our types will ensure it's a function if it's defined.

```suggestion
    onZoom?.(from, to);
```

spadgett (commented):
Looks good, just one question on whether we should remember the zoom values after leaving the page and coming back. @kyoto @cshinn Opinion?

frontend/public/components/monitoring/query-browser.tsx
kyoto: 👍

frontend/public/components/monitoring/dashboards/graph.tsx
kyoto: Good point 😄 . Now switched to use `useDispatch` and `useSelector`.

frontend/public/components/monitoring/dashboards/graph.tsx
kyoto: Good catch! I think it's better to revert to the default time range when you change dashboard, so I updated it to do that.

kyoto:
Addressed your feedback @spadgett. Thanks!

spadgett (approved):
/lgtm
/hold for CI fix in #8289

👍 Thanks @kyoto

PR #8275, opened 2021-03-02
Bug 1930546: show error in RsourceDropdown on load error

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/OCPBUGSM-25270

**Analysis / Root cause**: 
Monitoring-dashboard-workload keep loading when the user with cluster-role cluster-monitoring-view rolebingding user login developer console and select a project

**Solution Description**: 
Show error when having load error

image

image

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
+++ b/frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
@@ -109,13 +109,13 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
       title,
       actionItems,
     } = nextProps;
-    if (!loaded) {
+    if (!loaded && !loadError) {
       this.setState({ title: <LoadingInline /> });
       return;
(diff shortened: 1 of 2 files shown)


zherman0 (approved):

divyanshiGupta:
@vikram-raj we should probably add `!this.props.loadError` here https://github.com/openshift/console/pull/8275/files#diff-5c6a5e3c4cb46bf4319021951c3394360e3a2a2b193ab639e3a9028c8e7d6ee2R118 as well since if loaded is true but there is some error instead of the error the placeholder will be shown.

vikram-raj:
@divyanshiGupta Updated the PR with your suggested change. PTAL.

debsmita1:
verified the changes
/lgtm

rohitkrai03 (approved):
/approve

PR #8279, opened 2021-03-02
CONSOLE-2772: Consolidate A-D i18n namespaces

zherman0:
[CONSOLE-2772](https://issues.redhat.com/browse/CONSOLE-2772)
Consolidate A-E i18n namespaces


spadgett:
Thanks @zherman0. @rebeccaalpert Do you know if this will conflict with #8101?

spadgett (approved):
/lgtm
/hold

I'm debating whether we should try to manually move the translations into their respective public catalogs or just wait for the next translation drop. It might be something we can skip @rebeccaalpert Any opinion?

rebeccaalpert:
@spadgett - I think there will be conflicts. They should be easy to resolve though.

For me, it comes down to how much you value developer time. I think it's a pain to have to manually update the languages; we don't have any tooling around that the way we do for English. 

We just sent over files for translation on Monday, so the next drop won't include these changes. We'll have bad data for at least another two weeks if we don't manually update these, but the translation import process is quick. I'd lean towards waiting for new translation files if there's no pressing reason to have translations up to date.

spadgett:
Let's hold for #8101 so that we don't have any problems backporting the updated translations.

/hold

spadgett:
Hey, @zherman0. This need will rebase, then we can ask QE to review. Thanks!

spadgett (approved):
/lgtm

spadgett:
for approvals

See the note here: https://github.com/openshift/console/pull/8280#pullrequestreview-615749461

yapei:
@XiyunZhao  Please help test this PR

spadgett:
/lgtm cancel

I noticed that some of messages in packages use the public namespace. Those packages should use the namespace that matches the package name.

frontend/packages/ceph-storage-plugin/src/components/dashboards/common/breakdown-card/breakdown-body.tsx
```diff
@@ -28,7 +28,7 @@ export const BreakdownCardBody: React.FC<BreakdownBodyProps> = ({
   if (!capacityUsed || !top5MetricsStats.length || hasLoadError) {
     return (
       <div className="text-secondary capacity-breakdown-card--error">
-        {t('dashboard~Not available')}
+        {t('public~Not available')}
```
spadgett, line 31: Messages in ceph-storage-plugin should use the `ceph-storage-plugin` namespace.

```suggestion
        {t('ceph-storage-plugin~Not available')}
```

frontend/packages/console-shared/src/components/dashboard/activity-card/ActivityBody.tsx
```diff
@@ -21,7 +21,7 @@ export const Activity: React.FC<ActivityProps> = ({ timestamp, children }) => {
     <div className="co-activity-item__ongoing" data-test="activity">
       {timestamp && (
         <span className="text-secondary">
-          {t('dashboard~Started')}{' '}
+          {t('public~Started')}{' '}
```
spadgett, line 24: Use `console-shared` throughout this package.

```suggestion
          {t('console-shared~Started')}{' '}
```

frontend/packages/dev-console/src/components/monitoring/dashboard/MonitoringDashboardGraph.tsx
```diff
@@ -46,7 +46,7 @@ export const MonitoringDashboardGraph: React.FC<MonitoringDashboardGraphProps> =
       <DashboardCardBody>
         <PrometheusGraphLink
           query={query}
-          ariaChartLinkLabel={t('dashboard~View metrics for {{title}}', {
+          ariaChartLinkLabel={t('public~View metrics for {{title}}', {
```
spadgett, line 49:
```suggestion
          ariaChartLinkLabel={t('dev-console~View metrics for {{title}}', {
```

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.spec.tsx
```diff
@@ -538,22 +538,22 @@ describe(ClusterServiceVersionsDetailsPage.displayName, () => {
   it('renders a `DetailsPage` with the correct subpages', () => {
     const detailsPage = wrapper.find(DetailsPage);
     expect(detailsPage.props().pagesFor(testClusterServiceVersion)[0].nameKey).toEqual(
-      'details-page~Details',
+      'public~Details',
```
spadgett, line 541:
```suggestion
      'operator-lifecycle-manager~Details',
```

frontend/packages/pipelines-plugin/src/components/pipelineruns/__tests__/PipelineRunDetailsPage.spec.tsx
```diff
@@ -13,7 +13,7 @@ import TaskRuns from '../detail-page-tabs/TaskRuns';
 const menuActions = jest.spyOn(utils, 'useMenuActionsWithUserAnnotation');
 const breadCrumbs = jest.spyOn(hookUtils, 'usePipelinesBreadcrumbsFor');
 type PipelineRunDetailsPageProps = React.ComponentProps<typeof PipelineRunDetailsPage>;
-const i18nNS = 'details-page';
+const i18nNS = 'public';
```
spadgett, line 16: This should be `pipelines-plugin`

spadgett (changes requested):
@zherman0 Sorry for missing this on previous review. I thought all of these files were under the `public` folder. For other packages, the namespace should match the package name.

zherman0:
@spadgett - so the 'dashboard' namespace was used across multiple plugins.  I converted every plugin to its own namespace now.

XiyunZhao:
Currently didn't find any issue on private build, @zherman0 could you please help to resolve the conflict, and i would like to double check on official build and cluster

zherman0:
@spadgett - Rebased and ready for review.  Thanks.

spadgett:
@cloudbehl We might need your help understanding the ceph test failures in this PR. Thanks!

frontend/packages/dev-console/locales/en/dev-console.json
```diff
@@ -0,0 +1,3 @@
+{
+  "View metrics for {{title}}": "View metrics for {{title}}"
```
spadgett, line 2: Something isn't right here... We should make sure we're using the same namespace as the rest of dev console. We could consider removing this from your PR since it's really code owned by the dev console team.

frontend/packages/ceph-storage-plugin/locales/en/dashboard.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Not available": "Not available"
```
spadgett, line 2: I'm unsure why we have a new namespace in ceph-storage-plugin. @zherman0 Was this created on purpose?

zherman0: It was on purpose... once I removed the dashboard.json from public/locales/en, the i18n auto generated that file to the /package/ceph/locales.  The complete solution would be to go find where that is used in the ceph code and change it to ceph-storage.json but I decided to leave that out of this scope since it involves 5 ceph files.  Let me know if you want me to make those changes as part of this PR.

spadgett: @zherman0 Can you open a bug against the storage plugin component to make this change?

spadgett (approved):
/lgtm

We will need docs and px approval

frontend/packages/ceph-storage-plugin/locales/en/dashboard.json
zherman0: This has been resolved by PR 8764.

spadgett (approved):
/lgtm

PR #8284, opened 2021-03-03
adding console.resource-metadata extension types to dynamic-plugin-sdk

glekner:


christianvogt:
/approve

@spadgett The idea behind this extension is to be able to associate metadata to models without the need to override each model version. It will also allow us to better make use of the models returned by the api server and then augment them with metadata.

One major use case is associating badge colors to specific models. Also tagging api groups or versions with badges.

glekner:
rebased

glekner:
@christianvogt please relgtm

PR #8149, opened 2021-02-12
[release-4.7] Bug 1928023: Consider finally tasks when calculating task status

rottencandy:
This is a cherry-pick of #8110

PR #8292, opened 2021-03-03
Bug 1934664: Consider finally tasks when calculating task status

rottencandy:
This is a manual cherry pick of #8149 / #8110 .

4.6 uses the pipeline to calculate total tasks, instead of using `pipelineSpec` from the pipeline run like in 4.7+.
This change also continues to use the pipeline, since switching to `pipelineSpec` requires too many type & file changes.

### DIFF
--- a/frontend/packages/dev-console/src/test/pipeline-data.ts
+++ b/frontend/packages/dev-console/src/test/pipeline-data.ts
@@ -24,6 +24,7 @@ export enum PipelineExampleNames {
   INVALID_PIPELINE_MISSING_TASK = 'missing-task-pipeline',
   INVALID_PIPELINE_INVALID_TASK = 'invalid-task-pipeline',
   EMBEDDED_PIPELINE_SPEC = 'embedded-pipeline-spec',
+  PIPELINE_WITH_FINALLY = 'pipeline-with-finally',
 }
 
 type CombinedPipelineTestData = {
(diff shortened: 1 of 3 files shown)


frontend/packages/dev-console/src/utils/pipeline-augment.ts
```diff
@@ -183,6 +184,7 @@ export interface PipelineRun extends K8sResourceKind {
     // Odd status value that only appears in a single case - cancelling a pipeline
     status?: 'PipelineRunCancelled';
     timeout?: string;
+    finally?: PipelineTask[];
```
andrewballantyne, line 187: This is new... but I don't see any issues with it.

andrewballantyne (approved):

andrewballantyne:
@rottencandy Your tests need updating to handle two params:

> `console/frontend/packages/dev-console/src/utils/__tests__/pipeline-augment.spec.ts` failed

frontend/packages/dev-console/src/test/pipeline-data.ts
```diff
@@ -1597,6 +1598,91 @@ export const pipelineTestData: PipelineTestData = {
...
+        metadata: {
+          name: 'finally-pipeline-3tt7aw',
+          namespace: 'tekton-pipelines',
+          labels: { [TektonResourceLabel.pipeline]: 'finally-pipeline' },
+        },
+        spec: {
+          pipelineRef: { name: 'finally-pipeline' },
+        },
+        status: {
+          pipelineSpec: {
```
karthikjeeyar, line 1638: Since we do not have this `pipelineSpec` type defined in the `pipelineRunKind.status` in 4.6, there is a type error in this file as well.

rottencandy: Updated.

frontend/packages/dev-console/src/utils/pipeline-augment.ts
rottencandy: Removed it, I think I got confused.

karthikjeeyar:
Verified and working fine.
/lgtm

PR #8289, opened 2021-03-03
test: clear the session token before login

spadgett:
Sometimes Cypress will not correctly clear the session cookie on logout
when running in CI, possibly due to a Cypress bug in cookie handling. We
have not been able to reproduce manually or running the tests outside of
CI. Work around the problem by clearing the cookie manually before login.

Enable cookie debugging in Cypress to help troubleshoot similar problems
in the future.

This also changes the login command to use `window.SERVER_FLAGS.authDisabled`
to detect a local development environment regardless of whether
`BRIDGE_KUBEADMIN_PASSWORD` is set.

PR #8294, opened 2021-03-03
Bug 1934329: Bump sanitize-html version to 2.3.2 to improperly validate the hostnames

jhadvig:
Improper validation of hostnames set by the "allowedIframeHostnames" option can lead to bypass hostname whitelist for iframe element. Fixing this even though we are not using this option in this repo.

/assign @spadgett

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -177,7 +177,7 @@
     "react-virtualized": "9.x",
     "redux": "4.0.1",
     "reselect": "4.x",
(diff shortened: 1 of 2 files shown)


spadgett (approved):
/lgtm

spadgett:
Holding until #8289 merges to unblock CI.

/hold

zherman0 (approved):

PR #8295, opened 2021-03-03
odc-5564-select the builder image when git url is not validated

makambalaji:
**problem**: "git url is not validated due to git rate limit
**solution**: Enter the git url, if it is not validated, then select the builder image manually

**JIRA Id**:  https://issues.redhat.com/browse/ODC-5564

image

image

image

### DIFF
--- a/frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
+++ b/frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
@@ -1,4 +1,4 @@
-import { gitAdvancedOptions, buildConfigOptions } from '../../constants/add';
+import { gitAdvancedOptions, buildConfigOptions, builderImages } from '../../constants/add';
 import { messages } from '../../constants/staticText/addFlow-text';
 import { gitPO } from '../../pageObjects/add-flow-po';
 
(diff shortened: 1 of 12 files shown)


frontend/packages/pipelines-plugin/integration-tests/support/step-definitions/pipelines/create-from-add-options.ts
```diff
@@ -81,10 +81,6 @@ Given('pipeline {string} is executed for 5 times', (pipelineName: string) => {
   cy.get(pipelineRunDetailsPO.pipelineRunStatus).should('not.have.text', 'Running');
 });
 
-When('user enters Git Repo url as {string}', (gitUrl: string) => {
```
VeereshAradhya, line 84: If the implementation is removed the step from feature file also should be removed right?

makambalaji: Its duplicate one, so thought to remove it

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
```diff
@@ -130,8 +130,21 @@ export const gitPage = {
       .get(gitPO.cancel)
       .should('be.enabled')
       .click(),
-  verifyValidatedMessage: () =>
-    cy.get(gitPO.gitSection.validatedMessage).should('have.text', messages.gitUrlValidated),
+  verifyValidatedMessage: (builderImageName = 'Node.js') => {
```
VeereshAradhya, line 133: the `builderImageName` will have default value of `Node.js`. What if another URL is given? The step should include the `builderImageName` as param. WDYT?

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
```diff
@@ -130,8 +130,21 @@ export const gitPage = {
...
+    cy.get(gitPO.gitSection.validatedMessage).should('not.have.text', 'Validating...');
+    cy.get('body').then(($body) => {
+      if (
+        $body
+          .find(gitPO.gitSection.validatedMessage)
+          .text()
+          .includes(messages.privateGitRepoMessage)
+      ) {
+        cy.get(`[aria-label="${builderImageName}"]`).click();
```
VeereshAradhya, line 142: how about logging `selecting builder image builderImageName` in case of fail in validation?

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
makambalaji: Even if we give different url, still the script works fine, because builder image selection doesn't impact on any screen

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
makambalaji: Added

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
VeereshAradhya: If the validation fails, how the script will know what builder image (nodejs, python, php, etc..) it has to select?

makambalaji:
Holding until #8289 merges to unblock CI.

/hold

frontend/packages/dev-console/integration-tests/support/pages/add-flow/git-page.ts
makambalaji: 1st point - 
We already used param, but the param got default value 'Node.js'. [If we are not passing any parameter. it autoselects the node.js builder image by default]
3rd point -
Based on the scenario, we can identify the test data.

makambalaji:
PR: #8289  got merged
/hold cancel

gajanan-more (approved):
Running as expected.
/lgtm

PR #8302, opened 2021-03-04
create infra to execute commands in terminal

sahil143:
**_NOTE_**: This PR depends on #8260 . For this PR review commit https://github.com/openshift/console/pull/8302/commits/88283106fe5252917d34e0307aff56cfd3e46602

**Fixes**: https://issues.redhat.com/browse/ODC-5548

**Analysis / Root cause**: missing infrastructure to execute a command outside of the terminal

**Solution Description**: created infrastructure to execute a command outside of the 

terminal

exec-term

**_admin: when cloudshell is closed_**
close-exec-term

**_developer: when cloudshell is closed_**
dev-exex-term

**Test setup:**

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
@@ -13,6 +13,7 @@ import { setCloudShellActive } from '../../redux/actions/cloud-shell-actions';
 import Terminal, { ImperativeTerminalType } from './Terminal';
 import TerminalLoadingBox from './TerminalLoadingBox';
 import useActivityTick from './useActivityTick';
+import ExecuteCommand from './ExecuteCommand';
 import {
   getCloudShellCR,
   CLOUD_SHELL_STOPPED_BY_ANNOTATION,
@@ -78,6 +79,10 @@ const CloudShellExec: React.FC<CloudShellExecProps> = ({
     },
     [tick],
   );
+
+  const onCommand = React.useCallback((command: string): void => {
+    ws.current && ws.current.send(`0${Base64.encode(`${command}\n`)}`);
+  }, []);
 
   React.useEffect(() => {
     onActivate(true);
@@ -238,9 +243,12 @@ const CloudShellExec: React.FC<CloudShellExecProps> = ({
 
   if (wsOpen) {
     return (
-      <div className="co-cloudshell-terminal__container">
-        <Terminal onData={onData} ref={terminal} />
-      </div>
+      <>
+        <div className="co-cloudshell-terminal__container">
+          <Terminal onData={onData} ref={terminal} />
+        </div>
+        <ExecuteCommand onCommand={onCommand} />
+      </>
     );
   }
 
--- a/frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
+++ b/frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
@@ -1,12 +1,16 @@
-import { RootState } from '@console/internal/redux';
 import { Dispatch } from 'react-redux';
+import { RootState } from '@console/internal/redux';
 import { action, ActionType } from 'typesafe-actions';
 import { isCloudShellExpanded, isCloudShellActive } from '../reducers/cloud-shell-selectors';
 
 export enum Actions {
   SetCloudShellExpanded = 'setCloudShellExpanded',
   SetCloudShellActive = 'setCloudShellActive',
+  SetCloudShellCommand = 'setCloudShellCommand',
 }
+
+export const setCloudShellCommand = (command: string | null) =>
+  action(Actions.SetCloudShellCommand, { command });
 
 export const setCloudShellExpanded = (isExpanded: boolean) =>
   action(Actions.SetCloudShellExpanded, { isExpanded });
@@ -31,6 +35,7 @@ export const setCloudShellActive = (isActive: boolean) =>
 const actions = {
   setCloudShellExpanded,
   setCloudShellActive,
+  setCloudShellCommand,
 };
 
 export type CloudShellActions = ActionType<typeof actions>;
--- a/frontend/packages/console-app/src/redux/reducers/cloud-shell-selectors.ts
+++ b/frontend/packages/console-app/src/redux/reducers/cloud-shell-selectors.ts
@@ -7,3 +7,6 @@ export const isCloudShellExpanded = (state: RootState): boolean =>
 
 export const isCloudShellActive = (state: RootState): boolean =>
   !!state.plugins?.console?.[cloudShellReducerName]?.isActive;
+
+export const getCloudShellCommand = (state: RootState): string | null =>
+  state.plugins?.console?.[cloudShellReducerName]?.command ?? null;
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -19877,6 +19877,11 @@ xterm@^4.10.0, xterm@^4.8.1:
   resolved "https://registry.yarnpkg.com/xterm/-/xterm-4.10.0.tgz#fc4f554e3e718aff9b83622e858e64b0953067bb"
   integrity sha512-Wn66I8YpSVkgP3R95GjABC6Eb21pFfnCSnyIqKIIoUI13ohvwd0KGVzUDfyEFfSAzKbPJfrT2+vt7SfUXBZQKQ==
 
+xterm@^4.10.0:
+  version "4.10.0"
+  resolved "https://registry.yarnpkg.com/xterm/-/xterm-4.10.0.tgz#fc4f554e3e718aff9b83622e858e64b0953067bb"
+  integrity sha512-Wn66I8YpSVkgP3R95GjABC6Eb21pFfnCSnyIqKIIoUI13ohvwd0KGVzUDfyEFfSAzKbPJfrT2+vt7SfUXBZQKQ==
+
 y18n@^3.2.1:
   version "3.2.1"
   resolved "https://registry.yarnpkg.com/y18n/-/y18n-3.2.1.tgz#6d15fba884c08679c0d77e88e7759e811e07fa41"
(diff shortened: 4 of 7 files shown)


frontend/packages/console-app/src/components/quick-starts/QuickStartMarkdownView.tsx
```diff
@@ -23,6 +32,28 @@ type QuickStartMarkdownViewProps = {
   exactHeight?: boolean;
 };
 
+const QuickStartMarkdownExecute = ({ docContext, rootSelector }) => {
```
sahil143, line 35: Any changes to this file are only for testing purposes. no need to review these

frontend/packages/console-app/src/redux/reducers/cloud-shell-selectors.ts
```diff
@@ -7,3 +10,8 @@ export const isCloudShellExpanded = (state: RootState): boolean =>
 
 export const isCloudShellActive = (state: RootState): boolean =>
   !!state.plugins?.console?.[cloudShellReducerName]?.isActive;
+
+export const getCloudShellCommand = (state: RootState): boolean =>
+  state.plugins?.console?.[cloudShellReducerName]?.command;
```
christianvogt, line 15: You probably need to return `null` to match the correct signature:
```suggestion
export const getCloudShellCommand = (state: RootState): string | null =>
  state.plugins?.console?.[cloudShellReducerName]?.command ?? null;
```

frontend/packages/console-app/src/redux/reducers/cloud-shell-selectors.ts
```diff
@@ -7,3 +10,8 @@ export const isCloudShellExpanded = (state: RootState): boolean =>
...
+
+export const useCloudShellCommandSelector = (): string | null => useSelector(getCloudShellCommand);
```
christianvogt, line 17: Not sure there's value in keeping this vs using `useSelector(getCloudShellCommand)` which is what's done with the other selectors.

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
```diff
@@ -1,13 +1,37 @@
...
 import { action, ActionType } from 'typesafe-actions';
 import { isCloudShellExpanded, isCloudShellActive } from '../reducers/cloud-shell-selectors';
 
 export enum Actions {
   SetCloudShellExpanded = 'setCloudShellExpanded',
   SetCloudShellActive = 'setCloudShellActive',
+  SetCloudShellCommand = 'SetCloudShellCommand',
 }
 
+const cloudShellCommand = (command: string) => action(Actions.SetCloudShellCommand, { command });
```
christianvogt, line 15:
```suggestion
const cloudShellCommand = (command: string | null) => action(Actions.SetCloudShellCommand, { command });
```

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
```diff
@@ -1,13 +1,37 @@
...
+
+export const useCloudShellCommandDispatch = (): {
+  setCloudShellCommand: (command: string) => void;
+  clearCloudShellCommand: () => void;
+} => {
```
christianvogt, line 20: I'd prefer a single function that accepts the command or null.

invincibleJai:
Verified locally was able to execute a command in scenarios as mentioned for kubeadmin and regular user. Looks good!!

christianvogt:
/hold
@sahil143 can you remove your test code

sahil143:
/hold cancel
removed test code

frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
```diff
@@ -238,9 +243,12 @@ const CloudShellExec: React.FC<CloudShellExecProps> = ({
...
   if (wsOpen) {
     return (
-      <div className="co-cloudshell-terminal__container">
-        <Terminal onData={onData} ref={terminal} />
-      </div>
+      <>
+        <div className="co-cloudshell-terminal__container">
+          <Terminal onData={onData} ref={terminal} />
+        </div>
+        <ExecuteCommand onCommand={onCommand} />
```
rohitkrai03, line 250: Do we really need a separate component here to run the command?

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
```diff
@@ -1,13 +1,31 @@
...
+import { useDispatch, Dispatch } from 'react-redux';
 import { RootState } from '@console/internal/redux';
-import { Dispatch } from 'react-redux';
 import { action, ActionType } from 'typesafe-actions';
 import { isCloudShellExpanded, isCloudShellActive } from '../reducers/cloud-shell-selectors';
 
 export enum Actions {
   SetCloudShellExpanded = 'setCloudShellExpanded',
   SetCloudShellActive = 'setCloudShellActive',
+  SetCloudShellCommand = 'SetCloudShellCommand',
```
rohitkrai03, line 12: Just to be consistent with all the other actions.

```suggestion
  SetCloudShellCommand = 'setCloudShellCommand',
```

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
```diff
@@ -1,13 +1,31 @@
...
 }
 
+const cloudShellCommand = (command: string | null) =>
+  action(Actions.SetCloudShellCommand, { command });
+
+export const useCloudShellCommandDispatch = (): ((command: string | null) => void) => {
```
rohitkrai03, line 18: Not sure if this should be part of this file. Its actually an action dispatcher and not an action creator.

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
sahil143: We can add the logic to execute command in the CloudShellExec component itself but every time a command is sent the component will rerender and would require extra checks(like web socket is open or not) to execute commands. This way when command is sent only the Execute component is affected and will only render when websocket is open. 🙂

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
sahil143: we can create a new file `cloud-shell-action-dispatcher` under the actions directory. WDYT?

rohitkrai03: I think if are keeping the separate component to handle execution of the command then this can be part of this component itself. I guess my question is why have a separate dispatcher utility when it is just dispatching an action. The components can handle this code directly.

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
```diff
@@ -1,13 +1,31 @@
...
 import { action, ActionType } from 'typesafe-actions';
 import { isCloudShellExpanded, isCloudShellActive } from '../reducers/cloud-shell-selectors';
 
 export enum Actions {
   SetCloudShellExpanded = 'setCloudShellExpanded',
   SetCloudShellActive = 'setCloudShellActive',
+  SetCloudShellCommand = 'SetCloudShellCommand',
 }
 
+const cloudShellCommand = (command: string | null) =>
```
rohitkrai03, line 15:
```suggestion
const setCloudShellCommand = (command: string | null) =>
```

rohitkrai03 (changes requested):

frontend/packages/console-app/src/components/cloud-shell/CloudShellExec.tsx
rohitkrai03: Okay, yeah. It makes sense to keep this logic to prevent re-renders of parent component.

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
sahil143: So it's easy to use/import the utility since we don't have redux types up to date.

rohitkrai03: First off, we should definitely upgrade the redux types. Its been like this for so long now. And secondly my concern with keeping separation of concerns because dispatching an action is usually done from a component and should be part of component logic. If not then we can have a utility hook in utils folder related to `cloud-shell` that any component can use. Keeping this here along with actions and reducers feels a bit odd.

rohitkrai03: On a second thought, I guess it doesn't really matters. You can keep this in a separate file named `cloud-shell-dispatchers.ts`. We can maybe start to use this pattern whenever we want to reuse a complex dispatch logic anywhere.

frontend/packages/console-app/src/redux/actions/cloud-shell-actions.ts
invincibleJai: @sahil143 ^^

frontend/yarn.lock
```diff
@@ -19877,6 +19877,11 @@ xterm@^4.10.0, xterm@^4.8.1:
   resolved "https://registry.yarnpkg.com/xterm/-/xterm-4.10.0.tgz#fc4f554e3e718aff9b83622e858e64b0953067bb"
   integrity sha512-Wn66I8YpSVkgP3R95GjABC6Eb21pFfnCSnyIqKIIoUI13ohvwd0KGVzUDfyEFfSAzKbPJfrT2+vt7SfUXBZQKQ==
 
+xterm@^4.10.0:
```
rohitkrai03, line 19880: Why is this change in here?

rohitkrai03, line 19880: Why is this change in here?

PR #8151, opened 2021-02-12
Customize VM template boot source

rawagner:

PR #8313, opened 2021-03-05
Add provder and support fields when customizing source

rawagner:
depends on https://github.com/openshift/console/pull/8151


yaacov:
@rawagner ping ^^

yaacov:
@rawagner looks like the tests actually failing ^^

PR #8314, opened 2021-03-05
Add info about customize flow when adding source

rawagner:
Screenshot from 2021-03-05 14-15-23


yaacov:
/hold

@rawagner hi, it looks like a `i18n` thing that needs fixing, can you take a look ?

PR #8321, opened 2021-03-08
Monitoring dashboards: Fix bug where end time was not reset after zoom

kyoto:
There was a bug where if you zoomed with the mouse and later on selected
a time range from the dropdown, then the graph end time was not reset to
the current time.

Bug was introduced by https://github.com/openshift/console/pull/8273

### DIFF
--- a/frontend/public/components/monitoring/dashboards/index.tsx
+++ b/frontend/public/components/monitoring/dashboards/index.tsx
@@ -229,12 +229,17 @@ const AllVariableDropdowns = connect(({ UI }: RootState) => ({
   variables: UI.getIn(['monitoringDashboards', 'variables']),
 }))(AllVariableDropdowns_);
 
-const TimespanDropdown_: React.FC<TimespanDropdownProps> = ({ timespan, setTimespan }) => {
+const TimespanDropdown_: React.FC<TimespanDropdownProps> = ({ timespan }) => {
+  const dispatch = useDispatch();
   const { t } = useTranslation();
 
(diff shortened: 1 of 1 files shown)


frontend/public/components/monitoring/dashboards/index.tsx
```diff
@@ -261,14 +266,9 @@ const TimespanDropdown_: React.FC<TimespanDropdownProps> = ({ timespan, setTimes
...
-  ({ UI }: RootState) => ({
-    timespan: UI.getIn(['monitoringDashboards', 'timespan']),
-  }),
-  {
-    setTimespan: UIActions.monitoringDashboardsSetTimespan,
-  },
-)(TimespanDropdown_);
+export const TimespanDropdown = connect(({ UI }: RootState) => ({
+  timespan: UI.getIn(['monitoringDashboards', 'timespan']),
+}))(TimespanDropdown_);
```
TheRealJon, line 271: IMO, if we're going to touch this, we might as well just get rid of `connect` here and switch to the `useSelector` Redux hook.

TheRealJon (approved):
/lgtm

One minor comment, I don't think it absolutely needs to be changed to merge this, just a suggestion.

frontend/public/components/monitoring/dashboards/index.tsx
kyoto: 👍 I'll address this in a follow-up PR. Thanks!

PR #8329, opened 2021-03-08
Bug 1931005:  include related objects without a name

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1931005

After:
Screen Shot 2021-03-08 at 1 55 02 PM

I debated the use of our standard empty state indicator (`-`) here, but ultimately decided to use it for the sake of consistency.

### DIFF
--- a/frontend/public/components/cluster-settings/related-objects.tsx
+++ b/frontend/public/components/cluster-settings/related-objects.tsx
@@ -19,6 +19,16 @@ const tableColumnClasses = [
   '', // NS
 ];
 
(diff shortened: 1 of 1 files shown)


yapei:
@rhamilto  Shall we show `-` when name is empty or only show `-` if no instances was found after searching(oc get xxxx --all-namespaces | oc get xxx -n {namespace})?  I agree with the most straightforward approach that we simple show what we parse from YAML

dtaylor113 (approved):
/lgtm

PR #8334, opened 2021-03-09
Query Browser: Fix bug where lines briefly appear outside graph frame

kyoto:
When changing the graph's time range, the data was not cleared until the
updated data was fetched. This could cause the data for the old
time range to be briefly rendered on the new X axis. If the old time
range was outside the new time range, this would result in the graph
lines being momentarily rendered outside of the graph frame.

Also adds a couple of missing type definitions.


TheRealJon (approved):
/lgtm

PR #8338, opened 2021-03-09
Fix to show correct internal image-stream name in container image edit flow's internal registry image-stream dropdown for kn service

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5607

**Analysis/Root cause:**
When support for selecting the runtime icon from the runtime icon dropdown was added to the edit flows the previous code for showing runtime icon for deployments created on creating an app using container image flow's internal registry option was not properly cleaned up for knative service resource. Label 'app.kubernetes.io/name' which is used for getting the internal imagestream name is set incorrectly with runtime icon name instead of the imagestream name for knative service resource.

**Solution:**
Clean up previous code for showing runtime icon for kn services created using internal registry option. Set value of lablel 'app.kubernetes.io/name' as internal image-stream name instead of runtime icon name for kn service.

**GIf:**
Peek 2021-03-09 16-36

/kind bug


invincibleJai:
/lgtm

Verified the changes works as expected.

PR #8340, opened 2021-03-09
Remove unused Dex dependency

jhadvig:
Came across this part of our codebase and after a bit of investigation I've found out that we are not using DexClient for anything, so I've removed it together with it's flags.

/assign @spadgett


spadgett (approved):
/lgtm

zherman0 (approved):

PR #8341, opened 2021-03-09
Cypress: fixed skipping login when running locally

dtaylor113:
Fixed login function so that it first visits base url which in turn sets `win.SERVER_FLAGS`; then login checks `win.SERVER_FLAGS?.authDisabled` which is disabled when bridge'ing locally.

This is a follow on to: https://github.com/openshift/console/pull/8289


spadgett (approved):
/lgtm

PR #8342, opened 2021-03-09
Rework terminal to check if the operator is installed before showing the terminal button

JPinkney:
This PR reworks the terminal availability check to make sure that the web terminal operator is installed before showing the terminal button. Right now, the check only consists of making sure that specific webhookconfigurations are available on the cluster. This approach works fine right now. However, in the future, the web terminal operator will depend on the devworkspace-operator (devworkspace-operator will provide most of the actual functionality), and we don't want to show the web terminal button when the devworkspace-operator is installed but the web-terminal operator isn't.

Related issue: https://issues.redhat.com/browse/WTO-20
Related PR: https://github.com/openshift/console-operator/pull/513

### DIFF
--- a/pkg/terminal/operator.go
+++ b/pkg/terminal/operator.go
@@ -5,17 +5,34 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-const (
-	webhookName = "controller.devfile.io"
+var (
+	OperatorAPIResource = &schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "subscriptions",
+	}
+
+	OperatorGroupVersion = &schema.GroupVersion{
+		Group:   "operators.coreos.com",
+		Version: "v1alpha1",
+	}
 )
 
-// workspaceOperatorIsRunning checks if the workspace operator is running and webhooks are enabled,
+const (
+	webhookName             = "controller.devfile.io"
+	webTerminalOperatorName = "web-terminal"
+	operatorsNamespace      = "openshift-operators"
+)
+
+// checkWebTerminalOperatorIsRunning checks if the workspace operator is running and webhooks are enabled,
 // which is a prerequisite for sending a user's token to a workspace.
-func workspaceOperatorIsRunning() (bool, error) {
+func checkWebTerminalOperatorIsRunning() (bool, error) {
 	config, err := rest.InClusterConfig()
 	if err != nil {
 		return false, err
@@ -40,5 +57,33 @@ func workspaceOperatorIsRunning() (bool, error) {
 		}
 		return false, err
 	}
+	return true, nil
+}
+
+// checkWebTerminalOperatorIsInstalled checks to see that a web-terminal-operator is installed on the cluster
+func checkWebTerminalOperatorIsInstalled() (bool, error) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return false, err
+	}
+
+	config.GroupVersion = OperatorGroupVersion
+	config.APIPath = "apis"
+
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = client.Resource(*OperatorAPIResource).Namespace(operatorsNamespace).Get(context.TODO(), webTerminalOperatorName, metav1.GetOptions{})
+	if err != nil {
+		// Web Terminal subscription is not found but it's technically not a real error so we don't want to propogate it. Just say that the operator is not installed
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+
+		return false, err
+	}
+
 	return true, nil
 }
--- a/pkg/terminal/proxy.go
+++ b/pkg/terminal/proxy.go
@@ -79,13 +79,13 @@ func (p *Proxy) HandleProxy(user *auth.User, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	operatorRunning, err := workspaceOperatorIsRunning()
+	isWebTerminalOperatorRunning, err := checkWebTerminalOperatorIsRunning()
 	if err != nil {
-		http.Error(w, "Failed to check workspace operator state. Cause: "+err.Error(), http.StatusInternalServerError)
+		http.Error(w, "Failed to check web terminal operator state. Cause: "+err.Error(), http.StatusInternalServerError)
 		return
 	}
-	if !operatorRunning {
-		http.Error(w, "Terminal endpoint is disabled: workspace operator is not deployed.", http.StatusForbidden)
+	if !isWebTerminalOperatorRunning {
+		http.Error(w, "Terminal endpoint is disabled: web terminal operator is not deployed.", http.StatusForbidden)
 		return
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	enabled, err := workspaceOperatorIsRunning()
+	isOperatorInstalled, err := operatorIsInstalled()
 	if err != nil {
-		klog.Errorf("Failed to check if workspace operator is running: %s", err)
+		klog.Errorf("Failed to check if the web-terminal operator is installed: %s", err)
```
sleshchenko, line 186: After I modified manually console cluster role, I tried to run bridge and got error:
```
E0311 14:08:29.546308   23204 proxy.go:186] Failed to check if the web-terminal operator is installed: unknown
```

When I investigated more it leaded to:
```
responseBody: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"operators.operators.coreos.com \"web-terminal.openshift-operators\" is forbidden: User \"system:serviceaccount:openshift-console:console\" cannot get resource \"operators\" in API group \"operators.coreos.com\" at the cluster scope","reason":"Forbidden","details":{"name":"web-terminal.openshift-operators","group":"operators.coreos.com","kind":"operators"},"code":403}
error: unknown
```

So, I would avoid using Rest client not to handle response manually. As alternative I can propose dynamic client, I'll share proposal a bit later today

sleshchenko: When not found, error is logged into logs, while it's not exceptional case:

```
E0311 14:27:46.292581   27133 proxy.go:186] Failed to check if the web-terminal operator is installed: the server could not find the requested resource
```

pkg/terminal/operator.go
```diff
@@ -42,3 +46,24 @@ func workspaceOperatorIsRunning() (bool, error) {
...
+		return false, err
+	}
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = client.RESTClient().
+		Get().
+		AbsPath(fmt.Sprintf("/apis/%s/%s/operators/%s", operatorsGroup, operatorsVersion, webTerminalOperatorName)).
```
sleshchenko, line 63: After I uninstalled operator with OperatorHub on DevConsole, operator seems to live on the cluster:
```
> kc get subscription -n openshift-operators
No resources found in openshift-operators namespace.
> kc get operators
NAME                               AGE
web-terminal.openshift-operators   35m
```

JPinkney: I've just pushed the changes to make it so this does a subscription check

pkg/terminal/proxy.go
sleshchenko: All of the above seems to be fixed with the new changes

sleshchenko (approved):
Works as expected after the last changes :+1:

pkg/terminal/operator.go
```diff
@@ -42,3 +59,34 @@ func workspaceOperatorIsRunning() (bool, error) {
...
+
+	client, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = client.Resource(*OperatorAPIResource).Namespace(operatorsNamespace).Get(context.TODO(), webTerminalOperatorName, metav1.GetOptions{})
+
+	// Web Terminal subscription is found
+	if err == nil {
```
jhadvig, line 81: We should be checking for the error right after the call.
```suggestion
	if err != nil ||  errors.IsNotFound(err) {
		return false, err
	}

	return true, nil
```

pkg/terminal/operator.go
```diff
@@ -42,3 +59,34 @@ func workspaceOperatorIsRunning() (bool, error) {
 	}
 	return true, nil
 }
+
+// operatorIsInstalled checks to see that a web-terminal-operator is installed on the cluster
+func operatorIsInstalled() (bool, error) {
```
jhadvig, line 64:
```suggestion
func checkWebTerminalOperatorIsInstalled() (bool, error) {
```

pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	enabled, err := workspaceOperatorIsRunning()
+	isOperatorInstalled, err := operatorIsInstalled()
```
jhadvig, line 184:
```suggestion
	isWebTerminalOperatorInstalled, err := checkWebTerminalOperatorIsInstalled()
```

pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
...
+		klog.Errorf("Failed to check if the web-terminal operator is installed: %s", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	if !enabled {
+	if !isOperatorInstalled {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+	isOperatorRunning, err := workspaceOperatorIsRunning()
```
jhadvig, line 194: I know this is not new
```suggestion
	isWorkspaceOperatorRunning, err := checkWorkspaceOperatorIsRunning()
```

pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		klog.Errorf("Failed to check if web-terminal operator is running: %s", err)
```
jhadvig, line 196: not sure if this shouldn't be:
```suggestion
		klog.Errorf("Failed to check if workspace operator is running: %s", err)
```

jhadvig (changes requested):
@JPinkney thanks for the PR. I've added couple of comments.

pkg/terminal/operator.go
sleshchenko: Sorry for getting back my :+1: 
I missed that the proposed way won't work as original. It will log error when subscription is not found.

it may be 
```go
	if err != nil {
      if errors.IsNotFound(err) {
        return false, nil
      }
	  return false, err
    }
```

JPinkney:
@jhadvig When you get a chance can you take another look, I've updated this and the console-operator PR as well

pkg/terminal/operator.go
```diff
@@ -5,17 +5,34 @@ import (
...
+	webhookName             = "controller.devfile.io"
+	webTerminalOperatorName = "web-terminal"
+	operatorsNamespace      = "openshift-operators"
 )
 
-// workspaceOperatorIsRunning checks if the workspace operator is running and webhooks are enabled,
+// webTerminalOperatorIsRunning checks if the workspace operator is running and webhooks are enabled,
 // which is a prerequisite for sending a user's token to a workspace.
-func workspaceOperatorIsRunning() (bool, error) {
+func webTerminalOperatorIsRunning() (bool, error) {
```
jhadvig, line 35:
```suggestion
func checkWebTerminalOperatorIsRunning() (bool, error) {
```

pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
...
+	isWebTerminalOperatorInstalled, err := checkWebTerminalOperatorIsInstalled()
 	if err != nil {
-		klog.Errorf("Failed to check if workspace operator is running: %s", err)
+		klog.Errorf("Failed to check if the web terminal operator is installed: %s", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	if !enabled {
+	if !isWebTerminalOperatorInstalled {
+		w.WriteHeader(http.StatusServiceUnavailable)
```
jhadvig, line 191: Let's add a corresponding `klog` msg here

pkg/terminal/proxy.go
```diff
@@ -181,13 +181,23 @@ func (p *Proxy) HandleProxyEnabled(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	isWebTerminalOperatorRunning, err := webTerminalOperatorIsRunning()
```
jhadvig, line 194:
```suggestion
	isWebTerminalOperatorRunning, err := checkWebTerminalOperatorIsRunning()
```

jhadvig (commented):
The PR looks good. A few additional comments.
Please squash and let's merge this 👍

pkg/terminal/proxy.go
JPinkney: Fixed

jhadvig (approved):
/lgtm

sleshchenko:
I wonder if it can be back-ported to openshift 4.7?
I think it makes sense since CRW is going to bring webhooks on the cluster and as side-effect enable terminals without this PR.

jhadvig:
@sleshchenko not sure if we should backport, if its not a BZ
/approve

PR #8345, opened 2021-03-10
Add support for dynamic plugins for storage class provisioner

bipuladh:


frontend/packages/ceph-storage-plugin/console-extensions.json
```diff
@@ -1 +1,154 @@
...
+    "properties": {
+      "csi": {
+        "openshift-storage.rbd.csi.ceph.com": {
+          "title": "Ceph RBD",
+          "provisioner": "rbd.csi.ceph.com",
+          "allowVolumeExpansion": true,
+          "parameters": {
+            "clusterID": {
+              "name": "Cluster ID",
+              "hintText": "The namespace where Ceph is deployed",
```
christianvogt, line 13: Localize all strings in this file:
```suggestion
              "name": "%namespace~Cluster ID%",
              "hintText": "%namespace~The namespace where Ceph is deployed%",
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/storage-class-provisioner.ts
```diff
@@ -0,0 +1,60 @@
...
+      [key: string]: {
+        title: string;
+        provisioner: string;
+        allowVolumeExpansion: boolean;
+        parameters: {
+          [key: string]: {
+            name: string;
+            hintText: string;
+            value?: string;
+            visible?: boolean | EncodedCodeRef;
```
christianvogt, line 20: Why `| EncodedCodeRef`. Looks like this is only used as a boolean.

frontend/public/components/storage-class-form.tsx
```diff
@@ -857,13 +845,32 @@ class StorageClassFormWithTranslation extends React.Component<
...
+      ) {
         return null;
       }
 
       if (parameter.Component) {
         const { Component } = parameter;
-        return (
+        return isExecutableCodeRef(Component) ? (
+          <AsyncComponent
+            loader={Component}
```
christianvogt, line 869: Component can only be a coderef, therefore the else condition can be removed.

bipuladh: We can add components directly to this file as well. In that case it wouldn't be code ref.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/storage-class-provisioner.ts
bipuladh: We are not using it as a function now. But it can be passed as a function, the function would take in storage class related items as params.

frontend/public/components/storage-class-form.tsx
christianvogt: I wasn't aware. Why not register all through extensions?

bipuladh: Well plugins could contribute storage classes as well, but there are default storage classes supported by OCP which can have these components embedded in the file itself.

christianvogt: @bipuladh but why wouldn't core console contribute these defaults through the same mechanism?

bipuladh: https://github.com/openshift/console/blob/master/frontend/public/components/storage-class-form.tsx#L149
So one way we could enforce this is by making sure that entries here in the future are lazy imports, else we need the above block. We are adding other storage class provisioners directly from the same file and not frome console-app.

christianvogt: @bipuladh you haven't answered my question still.
Why add the defaults here instead of contributed through the same extensions?

frontend/packages/ceph-storage-plugin/console-extensions.json
bipuladh: @christianvogt the default provisioners are also not translated. Do we need to translate these?

christianvogt: All strings in console should be translated. But seems like that should be a separate PR.

bipuladh:
/retest
Docker failure

frontend/packages/ceph-storage-plugin/console-extensions.json
```diff
@@ -1 +1,199 @@
...
+      "csi": {
+        "openshift-storage.rbd.csi.ceph.com": {
+          "title": "%ceph-storage-plugin~Ceph RBD%",
+          "provisioner": "rbd.csi.ceph.com",
+          "allowVolumeExpansion": true,
+          "parameters": {
+            "clusterID": {
+              // t('ceph-storage-plugin~Cluster ID')
+              "name": "%ceph-storage-plugin~Cluster ID%",
+              // t('ceph-storage-plugin~The namespace where Ceph is deployed')
```
christianvogt, line 14: Unnecessary comment. The next line is sufficient for translation.

frontend/packages/ceph-storage-plugin/src/components/ocs-storage-class-form/ocs-storage-class-form.tsx
```diff
@@ -26,7 +26,7 @@ import {
   WatchK8sResource,
 } from '@console/internal/components/utils/k8s-watch-hook';
 import { useFlag } from '@console/shared/src/hooks/flag';
-import { ProvisionerProps } from '@console/plugin-sdk';
+import { ProvisionerProps } from '@console/dynamic-plugin-sdk/src/extensions/storage-class-provisioner';
```
christianvogt, line 29: This seems wrong when referencing API from the SDK. It should be exposed at the root.

frontend/public/components/storage-class-form.tsx
```diff
@@ -81,30 +81,18 @@ class StorageClassFormWithTranslation extends React.Component<
...
+      );
+      Object.entries(item?.properties?.[Provisioner.OTHERS] ?? {})?.forEach(
+        ([key, val]) => (this.defaultStorageTypes[key] = val),
+      );
+    });
 
   // For 'csi' storage type
-  CSIStorageTypes = Object.freeze({
-    ...this.getExtensionsStorageClassProvisioners(Provisioner.CSI),
+  CSIStorageTypes = {
```
christianvogt, line 95: Do you have an issue tracking the migration of these types to be declarative?

bipuladh: The question would be for the default types do we really need to migrate them? If ever a default type is introduced which needs a custom component like the ceph storage class providers do then we can move them out to console-app. But these types right now do not require any component. Wdyt?

christianvogt: Not sure what having a component or not has to do with this.

You're maintaining 2 different methods of contribution and 2 code paths. It's simply a matter for code maintenance. One way to do something is preferred over multiple. I'm simply suggesting the migration be done now or tracked in an issue.

frontend/packages/console-dynamic-plugin-sdk/src/coderefs/coderef-resolver.ts
```diff
@@ -99,15 +123,15 @@ export const resolveEncodedCodeRefs = (
   errorCallback: VoidFunction,
 ): Extension[] =>
   _.cloneDeep(extensions).map((e) => {
-    const refs = filterEncodedCodeRefProperties(e.properties);
-
-    Object.entries(refs).forEach(([propName, ref]) => {
+    const paths = getCodeRefPaths(e.properties, isEncodedCodeRef);
```
vojtechszocs, line 126: I'd suggest the following approach:

- adapt `filterEncodedCodeRefProperties` and `filterExecutableCodeRefProperties` to deep-pick any code refs, for example:

```ts
filterEncodedCodeRefProperties({
  foo: true,
  bar: { "$codeRef": "a.b" },
  qux: {
    baz: { "$codeRef": "x.y" },
    test: 1
  }
})
// =>
{
  bar: { "$codeRef": "a.b" },
  qux: {
    baz: { "$codeRef": "x.y" }
  }
}
```

Code ref properties are ones whose value passes the given test (`isEncodedCodeRef` or `isExecutableCodeRef`).

- adapt `resolveEncodedCodeRefs` to deep-map the result of `filterEncodedCodeRefProperties` and merge it with extension's properties

- adapt `resolveCodeRefProperties` to deep-map the result of `filterExecutableCodeRefProperties` and return it

- adapt `collectCodeRefData` in extension validator

- `getDynamicExtensions` in active-plugins module uses `JSON.stringify` so this code shouldn't need any changes

And of course, adapt all affected unit tests :smiley: this is a lot of work and should be done carefully in a separate PR.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/storage-class-provisioner.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export type StorageClassProvisioner = ExtensionDeclaration<
+  'console.storage-class/provisioner',
+  {
+    [key: string]: {
+      [key: string]: {
+        title: string;
+        provisioner: string;
+        allowVolumeExpansion: boolean | CodeRef<(arg) => boolean>;
```
vojtechszocs, line 15: Why not simply `CodeRef<(arg) => boolean>` ?

I'm not strictly against these `value | CodeRef<func_to_get_value>` unions, it's just that we need to check first, if there aren't any issues from generated JSON schema, build and runtime perspectives.

vojtechszocs (changes requested):

frontend/public/components/storage-class-form.tsx
bipuladh: Creating a Jira Task to track moving these from here to console-app.

PR #8349, opened 2021-03-10
Open link configured with OLM x-descriptors:urn:alm:descriptor:org.w3:link in new window/tab

dmvolod:
It would be prefer to open link configured with following OLM descriptor in new window/tab as users always lives Console application in case of click on the Link


dmvolod:
@spadgett @TheRealJon could you please to have a look at this change.

TheRealJon:
@dmvolod IMO, we probably shouldn't change the behavior of this existing descriptor because it could have pretty wide impact. It might be worth considering a new `x-descriptors:urn:alm:descriptor:org.w3:externalLink` descriptor though.

dmvolod:
Yes, sure, thanks @TheRealJon 
I will close this PR and submit a new one with externalLink support

TheRealJon:
cc @tlwu2013

dmvolod:
Close with preparation fix for #8477

PR #8354, opened 2021-03-10
Overriding default kabab actions

GowthamShanmugam:

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/clusterserviceversions.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/clusterserviceversions.ts
@@ -9,6 +9,8 @@ namespace ExtensionProperties {
     label: string;
     /** API group of the resource */
     apiGroup: string;
+    /** override default actions */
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@bipuladh @cloudbehl, please review(the idea here is to override the default action Edit/Delete), For pool management, i am overriding default actions using my own component modal.  But still, i am blocking actions for external cluster and default pool inside code only, i am displaying a message in the modal.

cloudbehl:
@rawagner Can you review this and see if the idea fits in well enough?

bipuladh (changes requested):
The CSV Extension itself is going to get removed. 
/hold

GowthamShanmugam:
> The CSV Extension itself is going to get removed.
> /hold

Is it happening for 4.8? if yes can you please add this future in a new extention

GowthamShanmugam:
as per comment: https://github.com/openshift/console/pull/8354#pullrequestreview-621798130 closing this PR

PR #8358, opened 2021-03-10
Fix ResolvedExtension type inference

vojtechszocs:
This fixes the issue where `ResolvedExtension` return type of `useResolvedExtensions` hook would cause non-coderef extension properties to be typed as `never`.


rohitkrai03 (approved):
/lgtm

Tested the changes locally. Works as expected.

PR #7306, opened 2020-11-24
CONSOLE-2494: Bump node and yarn versions in docker builder file

zherman0:
Need to wait for [ART-2456] to be complete and for PR7304 to merge.

This bumps dependencies for nodejs and yarn in the build file.

PR #8363, opened 2021-03-10
improve i18n parser to support console-extensions.json

christianvogt:
Dynamic plugins contribute extensions through the `console-extensions.json`. To localize a string in json, the string must match the format `/%.+%/`.
eg.
```
{
  "foo": "%bar%"
}
```

This PR adds a custom lexer which parses `*.json` files and matches string values against the localization pattern.

Added unit test for custom JSON lexer.

Update the `i18n-scripts/.eslintrc.js` rules to reuse those defined in the console eslint plugin.

cc @rohitkrai03 @vojtechszocs @rebeccaalpert @spadgett 

Before this change we would require a comment to pick up the localized string. Although there is no lexer currently for json files.
```
  [
    {
      "type": "AddAction",
      "flags": {
        "required": ["OPENSHIFT_HELM"]
      },
      "properties": {
        "id": "helm",
        "url": "/catalog?catalogType=HelmChart",
        // t('helm-plugin~Helm Chart')
        "label": "%helm-plugin~Helm Chart%",
        // t('helm-plugin~Browse the catalog to discover and install Helm Charts')
        "description": "%helm-plugin~Browse the catalog to discover and install Helm Charts%"
      }
    }
  ]
```

After this change, the lexer will pickup the string without the comment:
```
  [
    {
      "type": "AddAction",
      "flags": {
        "required": ["OPENSHIFT_HELM"]
      },
      "properties": {
        "id": "helm",
        "url": "/catalog?catalogType=HelmChart",
        "label": "%helm-plugin~Helm Chart%",
        "description": "%helm-plugin~Browse the catalog to discover and install Helm Charts%"
      }
    }
  ]
```

locale file: `helm-plugin.json`
```
{
  "Helm Chart": "Helm Chart",
  "Browse the catalog to discover and install Helm Charts": "Browse the catalog to discover and install Helm Charts"
}
```

### DIFF
--- a/frontend/i18n-scripts/.eslintrc.js
+++ b/frontend/i18n-scripts/.eslintrc.js
@@ -1,61 +1,9 @@
-var fs = require('fs');
-
 module.exports = {
   root: true,
-  env: {
-    browser: true,
-    es6: true,
-    jasmine: true,
-    jest: true,
-  },
-  extends: [
-    'eslint:recommended',
-    'plugin:import/errors',
-    'plugin:import/warnings',
-    'plugin:console/prettier',
-  ],
-  parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint'],
+  extends: ['plugin:console/node-typescript-prettier'],
   rules: {
-    camelcase: [
-      'error',
-      { allow: [] },
-    ],
-    'consistent-return': 0,
-    'consistent-this': [1, 'that'],
-    curly: [2, 'all'],
-    'default-case': [2],
-    'dot-notation': [2],
-    'no-multiple-empty-lines': [2, { max: 2, maxEOF: 0 }],
-    eqeqeq: [2, 'allow-null'],
-    'guard-for-in': 2,
-    'import/no-unresolved': ['error'],
-    'import/no-duplicates': ['error'],
-    'max-nested-callbacks': [1, 4],
-    'no-alert': 2,
-    'no-caller': 2,
-    'no-console': 0,
-    'no-constant-condition': 2,
-    'no-debugger': 2,
-    'no-else-return': ['error'],
-    'no-global-strict': 0,
-    'no-irregular-whitespace': ['error'],
-    'no-prototype-builtins': 0, // Disable with exlint v6 update.
-    'no-shadow': ['error'],
-    'no-undef': 0,
-    'no-underscore-dangle': 0,
-    'no-var': 2,
-    'object-shorthand': ['error', 'properties'],
-    'prefer-const': ['error', { destructuring: 'all' }],
-    'prefer-template': 2,
-    radix: 2,
-    'require-atomic-updates': 0,
-  },
-  settings: {
-    'import/extensions': ['.js'],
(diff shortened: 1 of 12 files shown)


rebeccaalpert (approved):
Looks good to me. Just one comment about a console.log that looked like it may not be intentional. Thanks for cleaning up the linter file and making some of the other scripts easier to read.

frontend/i18n-scripts/.eslintrc.js
```diff
@@ -1,61 +1,9 @@
...
-    'require-atomic-updates': 0,
-  },
-  settings: {
-    'import/extensions': ['.js'],
-    'import/resolver': { node: { extensions: ['.js'] } },
-  },
-  globals: {
-    process: 'readonly',
+    'no-console': 'off',
+    // fs.promises requires a newer version of node however our compliance is set to node >=10
```
vojtechszocs, line 6: FYI once #7306 is merged, our compliance will be Node.js >= 14.

frontend/i18n-scripts/lexers.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Custom JSON parser for localizing keys matching format: /%.+%/
+ */
+module.exports.CustomJSONLexer = class extends EventEmitter {
+  extract(content, filename) {
+    const keys = [];
+    try {
+      JSON.parse(content, (key, value) => {
+        if (typeof value === 'string') {
+          const match = value.match(/%(.+)%/);
```
vojtechszocs, line 12:
```suggestion
          const match = value.match(/^%(.+)%$/);
```

frontend/i18n-scripts/set-english-defaults.js
```diff
@@ -55,11 +55,9 @@ function updateFile(fileName) {
...
   }
 
-  fs.writeFile(fileName, JSON.stringify(updatedFile, null, 2), function writeJSON(e) {
-    if (e) {
-      return console.error(e);
-    }
-  });
+  fs.promises
+    .writeFile(fileName, JSON.stringify(updatedFile, null, 2))
+    .catch((e) => console.error(e));
```
vojtechszocs, line 60:
```suggestion
    .catch((e) => console.error(fileName, e));
```

vojtechszocs (changes requested):

frontend/i18n-scripts/.eslintrc.js
christianvogt: right, but not there yet :P

christianvogt:
@vojtechszocs thanks for the review. I've updated all the suggestions.

frontend/i18n-scripts/lexers.js
```diff
@@ -0,0 +1,25 @@
...
+          console.log('testing:', match[1]);
```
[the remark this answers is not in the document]
christianvogt: removed

christianvogt:
updated lexer parser to use `comment-json` to properly parse json files with comments as this is supported by dynamic plugin extensions files.

PR #8368, opened 2021-03-11
Bug 1937683: Show plane image destination of output in buildConfig when the destination is a DockerImage

jhadvig:
Before:
Screenshot_2021-03-11 nodejs-echo Â· Details Â· Red Hat OpenShift Container Platform

After:
Screenshot 2021-03-11 at 12 20 20

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #8372, opened 2021-03-11
workspaces in pipeline start modal

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-4772

# Screenshot
Screencast from 03-31-2021 07_59_39 PM

# Tests
Added VolumeClaimTemplate.spec

# Browser Conformance
Chrome,


frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+                {...radio}
+                key={radio.value}
+                onChange={handleAccessMode}
+                inline
+                disabled={disabled}
+                checked={checked}
+                aria-describedby="access-mode-help"
+                name="accessMode"
+              />
+            );
```
andrewballantyne, line 131: Doing a `.forEach` in the middle of rendering is mighty weird... are you not just going to override your rendering multiple times before rendering?

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+    setAccessMode('ReadWriteOnce');
+    setAccessModeHelp(displayMessage);
+    // setting accessMode to default with the change to Storage Class selection
+    setAllowedAccessModes(modes);
+    setStorageClass(updatedStorageClass?.metadata?.name);
+    setStorageProvisioner(provisioner);
+    if (storageProvisioner.includes(cephStorageProvisioners[1])) {
+      setVolumeMode('Filesystem');
+    }
+  };
```
andrewballantyne, line 68: This feels weird... I wonder if we need some interim handler or wrapper of the StorageClass to simplify this logic. It's a dropdown after all.

andrewballantyne (commented):
Not doing a full review or anything, but just reading the code gets me these comments. If they are part of the path you're working on, ignore away 🙂

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
```diff
@@ -36,6 +38,20 @@ const getVolumeTypeFields = (volumeType: VolumeTypes, index: number, t: TFunctio
...
+      return (
+        <Alert variant="info" title="this option could not be used to share data between tasks" />
+      );
+    }
+    case VolumeTypes.VolumeClaimTemplate: {
+      return (
+        <VolumeClaimTemplateForm
+          name={`workspaces.${index}.data.volumeClaimTemplate`}
+          initialSizeValue="200"
+          initialSizeUnit="Mi"
```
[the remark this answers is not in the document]
abhinandan13jan: yes, it is 1Gi, updated accordingly

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+    requestSizeValue,
+    requestSizeUnit,
+    storageClass,
+    name,
+    setFieldTouched,
+    setFieldValue,
+  ]);
+
+  return (
+    <ExpandCollapse
```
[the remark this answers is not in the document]
abhinandan13jan: So, there is a help-block above this now. added as per UX screenshot

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
abhinandan13jan: Moved the forEach out and only applying to set the `checked` flag. I am replicating the same logic used in createPVCForm.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
abhinandan13jan: The same login applies for the CreatePVC dropdown. Hence, kept it. I dont think we can move it out as a util as it is directly changing the states of this component.

However, should we do this validation on client or should we let the server take care of this?

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
```diff
@@ -36,6 +38,20 @@ const getVolumeTypeFields = (volumeType: VolumeTypes, index: number, t: TFunctio
         />
       );
     }
+    case VolumeTypes.EmptyDirectory: {
+      return (
+        <Alert variant="info" title="this option could not be used to share data between tasks" />
```
andrewballantyne, line 43:
```suggestion
        <Alert inline variant="info" title="this option could not be used to share data between tasks" />
```
We don't use non-inline versions anywhere in the console I don't think. At least not in the DevConsole (that I have seen)

We'll also need to translate this... but I suppose we can wait on UX wording.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+    storageClass,
+    name,
+    setFieldTouched,
+    setFieldValue,
+  ]);
+
+  return (
+    <>
+      <p className="help-block">
+        {t('pipelines-plugin~This will create a PersistentVolumeClaim with a size of 1 GiB')}
```
andrewballantyne, line 102: Maybe we use the actual value here... That way when they change it, the wording doesn't say the incorrect thing.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/VolumeClaimTemplateForm.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+    setFieldValue: jest.fn(),
+    setFieldTouched: jest.fn(),
+    validateForm: jest.fn(),
+  })),
+  getFieldId: jest.fn(),
+}));
+
+describe('VolumeClaimTemplateForm', () => {
+  beforeEach(() => {
+    VolumeClaimTemplateFormProps = {
```
andrewballantyne, line 30:
```suggestion
    volumeClaimTemplateFormProps = {
```
Even in tests we should still use camelCase for variable names.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
andrewballantyne: And the wording has been figured out [here on the designs](https://docs.google.com/document/d/12ZETv528Puq41o9zg8_GCvUhhd5g8NEiGbA3bnmawjs/edit?disco=AAAALzTmAK4).

andrewballantyne:
image

Please align this with the design. Those labels are harsh standing out.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
andrewballantyne: I'll have to relook at this once you clean things up (other comments). It's possible there is a way to streamline this so it's not so heavy.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
abhinandan13jan: yes, that seems to be the better way. Updated accordingly

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/VolumeClaimTemplateForm.spec.tsx
abhinandan13jan: yes

abhinandan13jan:
> image
> 
> Please align this with the design. Those labels are harsh standing out.

@andrewballantyne Updated screenshot. Now using the same classes as that of create-pvc form

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,184 @@
...
+    const provisioner: string = updatedStorageClass?.provisioner || '';
+    // if the provisioner is unknown or no storage class selected, user should be able to set any access mode
+    const modes = provisionerAccessModeMapping[provisioner]
+      ? provisionerAccessModeMapping[provisioner]
+      : getAccessModeForProvisioner(provisioner);
+    // setting message to display for various modes when a storage class of a know provisioner is selected
+    const displayMessage =
+      provisionerAccessModeMapping[provisioner] || isCephProvisioner(provisioner)
+        ? 'Access mode is set by storage class and cannot be changed'
+        : 'Permissions to the mounted drive';
```
rottencandy, line 61: Should we translate these?

abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
andrewballantyne: Still need to check on this, but it should be fine as-is if I cannot figure out a better solution.

karthikjeeyar:
Through git import flow, we are creating default volumeClaimTemplates behind the scenes, with a pipeline label attached to VolumentClaimTemplate, this label will be attached to the newly created PVC  as well so start modal can auto select an existing PVC for that pipeline. A user who starts the pipeline for the second time will see the existing PVC selected for them, I think we should also include this labeling of VolumeClaimTemplate here. @andrewballantyne WDYT?

https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts#L133-L135

andrewballantyne:
> Through git import flow, we are creating default volumeClaimTemplates behind the scenes, with a pipeline label attached to VolumentClaimTemplate, this label will be attached to the newly created PVC as well so start modal can auto select an existing PVC for that pipeline. A user who starts the pipeline for the second time will see the existing PVC selected for them, I think we should also include this labeling of VolumeClaimTemplate here. @andrewballantyne WDYT?
> 
> https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts#L133-L135

I think that's a design question @karthikjeeyar - My thoughts are "sure, why not". @bgliwa01 Opinion?

nemesis09:
Please run "yarn i18n" to fix failing tests

frontend/public/locales/en/public.json
```diff
@@ -637,5 +637,6 @@
   "OK": "OK",
   "Last updated": "Last updated",
   "Disabled": "Disabled",
-  "Enabled": "Enabled"
+  "Enabled": "Enabled",
+  "Empty Directory doesn't support shared data between tasks": "Empty Directory doesn't support shared data between tasks"
```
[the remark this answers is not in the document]
abhinandan13jan: I missed the prefix. Updated now

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PiplelineWorkspacesSection.tsx
```diff
@@ -36,6 +38,24 @@ const getVolumeTypeFields = (volumeType: VolumeTypes, index: number, t: TFunctio
         />
       );
     }
+    case VolumeTypes.EmptyDirectory: {
+      return (
+        <Alert
+          isInline
+          variant="info"
+          title={t("Empty Directory doesn't support shared data between tasks")}
```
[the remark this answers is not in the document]
abhinandan13jan: updated

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,184 @@
...
+    setFieldTouched,
+    setFieldValue,
+  ]);
+
+  return (
+    <>
+      <p className="help-block">
+        {t(
+          'pipelines-plugin~This will create a PersistentVolumeClaim with a size of {{initialSizeValue}} {{initialSizeUnit}}',
+          { initialSizeValue, initialSizeUnit },
```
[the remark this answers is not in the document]
abhinandan13jan: makes sense. updated accordingly

abhinandan13jan:
@andrewballantyne @karthikjeeyar @nemesis09 Thanks for the review. Updated accordingly

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,184 @@
...
+        <label className="control-label co-required" htmlFor="access-mode">
+          {t('pipelines-plugin~Access Mode')}
+        </label>
+        <FormGroup fieldId="accessMode" data-test-id="accessModeRadio">
+          {accessModeRadios.map((radio) => {
+            const disabled = !allowedAccessModes.includes(radio.value);
+            let checked = null;
+            allowedAccessModes.forEach((mode) => {
+              checked = !disabled ? radio.value === accessMode : radio.value === mode;
+            });
```
nemesis09, line 129: in the code piece 
` allowedAccessModes.forEach((mode) => {
     checked = !disabled ? radio.value === accessMode : radio.value === mode;
});`
wouldn't the expression `radio.value === mode` always yield false when disabled is true  i.e `allowedAccessModes` does not include radio.value?

nemesis09: could we instead write it as
`const checked = disabled ? disabled : radio === accessMode;`

abhinandan13jan: What you say makes sense. But I've only replicated the logic used in create-pvc form without changing what they are trying to achieve.

I think in the case when all radios are disabled and one is selected by default. i.e user has no authority to make a change, this expression will hold true

abhinandan13jan: updated. Works fine locally

nemesis09: I'm a bit confused how this will account for a default selection when every `radio` is disabled.
As far as i could understand, when `disabled` is true, every iteration of `forEach` evaluates the expression `radio.value === mode` which would never yield true since `disabled` is true only when `allowedAccessModes` does not include `radio.value`.

debsmita1:
@abhinandan13jan you seem to have missed adding validation in the form, the `Size` field is made required but no error is shown when it is empty

Screenshot from 2021-03-19 20-21-46

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/__tests__/VolumeClaimTemplateForm.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+describe('VolumeClaimTemplateForm', () => {
+  beforeEach(() => {
+    volumeClaimTemplateFormProps = {
+      ...formikFormProps,
+      name: 'test.form',
+    };
+  });
+
+  it('should render correct HelpText and ExpandCollapse', () => {
```
nemesis09, line 36: should we be testing if a component renders unless it is rendered conditionally or the number of instances of a component rendered, changes?

abhinandan13jan:
> @abhinandan13jan you seem to have missed adding validation in the form, the `Size` field is made required but no error is shown when it is empty
> 
> Screenshot from 2021-03-19 20-21-46

I am not sure, whether we should validate anything on the client end. The patched spec should be validated against the tekton API. cc: @andrewballantyne What should be done here?

nemesis09:
> > @abhinandan13jan you seem to have missed adding validation in the form, the `Size` field is made required but no error is shown when it is empty
> > Screenshot from 2021-03-19 20-21-46
> 
> I am not sure, whether we should validate anything on the client end. The patched spec should be validated against the tekton API. cc: @debsmita1 @andrewballantyne What should be done here?

Even if the spec is validated against the API, should we allow a submit with invalid values?
Currently the form submits with an empty "size" field and an error log is shown in the Log Snippet of the PipelineRun details page.
Screenshot from 2021-03-19 22-46-00

abhinandan13jan:
> > > @abhinandan13jan you seem to have missed adding validation in the form, the `Size` field is made required but no error is shown when it is empty
> > > Screenshot from 2021-03-19 20-21-46
> > 
> > 
> > I am not sure, whether we should validate anything on the client end. The patched spec should be validated against the tekton API. cc: @debsmita1 @andrewballantyne What should be done here?
> 
> Even if the spec is validated against the API, should we allow a submit with invalid values?
> Currently the form submits with an empty "size" field and an error log is shown in the Log Snippet of the PipelineRun details page.
> Screenshot from 2021-03-19 22-46-00

validation added

nemesis09:
> > > > @abhinandan13jan you seem to have missed adding validation in the form, the `Size` field is made required but no error is shown when it is empty
> > > > Screenshot from 2021-03-19 20-21-46
> > > 
> > > 
> > > I am not sure, whether we should validate anything on the client end. The patched spec should be validated against the tekton API. cc: @debsmita1 @andrewballantyne What should be done here?
> > 
> > 
> > Even if the spec is validated against the API, should we allow a submit with invalid values?
> > Currently the form submits with an empty "size" field and an error log is shown in the Log Snippet of the PipelineRun details page.
> > Screenshot from 2021-03-19 22-46-00
> 
> validation added

I tried it out but apparently it still allows submit with empty field and invalid values

nemesis09:
I think the problem in the validation is that it is not validating "requestSize" alone and instead validating "storage" which is "requestSize"+"requestUnit"
so even with an empty "requestSize" field, storage does get a string containing only the "requestUnit", which is sufficient for a successful validation as per `yup.string().required()`

nemesis09:
Thanks @abhinandan13jan 
The validations work fine now.
I just have some observations around the `help-block` message that is generated using the `requestSize` and `requestUnit`.
The message shows invalid or no values when the `requestSize` has such values.
here are some examples-

- `requestSize` is empty
Screenshot from 2021-03-24 17-21-56
 
- `requestSize` has non numeric values
Screenshot from 2021-03-24 17-22-09

- `requestSize` has negative numeric values
Screenshot from 2021-03-24 17-22-25

cc @bgliwa01

abhinandan13jan:
> Thanks @abhinandan13jan
> The validations work fine now.
> I just have some observations around the `help-block` message that is generated using the `requestSize` and `requestUnit`.
> The message shows invalid or no values when the `requestSize` has such values.
> here are some examples-
> 
> * **`requestSize` is empty**
> 
> Screenshot from 2021-03-24 17-21-56
> 
> * **`requestSize` has non numeric values**
> 
> Screenshot from 2021-03-24 17-22-09
> 
> * **`requestSize` has negative numeric values**
> 
> Screenshot from 2021-03-24 17-22-25
> 
> cc @bgliwa01

updated please re-review

nemesis09:
tested locally
works fine
values such as -1 and 0 still show up in the help block but as discussed those could be handled as edge cases later with relevant UXD

Thanks for the changes @abhinandan13jan 
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+    setStorageClass(updatedStorageClass?.metadata?.name);
+    setStorageProvisioner(provisioner);
+    if (storageProvisioner.includes(cephStorageProvisioners[1])) {
+      setVolumeMode('Filesystem');
+    }
+  };
+
+  const handleRequestSizeChange = (size: RequestSize): void => {
+    const { value, unit } = size;
+    setRequestSizeValue(value);
```
andrewballantyne, line 83: Recommend you just handle the invalid values part of this PR as there is little reason to push it to a future issue.

Suggested change diff:
```
diff --git frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
index 8f0b6f41e7..b359eb9eb7 100644
--- frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
+++ frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
@@ -42,6 +42,7 @@ const VolumeClaimTemplateForm: React.FC<VolumeClaimTemplateFormProps> = ({
   const [allowedAccessModes, setAllowedAccessModes] = React.useState(initialAccessModes);
   const [volumeMode, setVolumeMode] = React.useState(initialVolumeMode);
   const [accessMode, setAccessMode] = React.useState('ReadWriteOnce');
+  const [requestSizeError, setRequestSizeError] = React.useState<string>(null);
   const [requestSizeValue, setRequestSizeValue] = React.useState(initialSizeValue);
   const [requestSizeUnit, setRequestSizeUnit] = React.useState(initialSizeUnit);
   const [storageProvisioner, setStorageProvisioner] = React.useState('');
@@ -80,6 +81,14 @@ const VolumeClaimTemplateForm: React.FC<VolumeClaimTemplateFormProps> = ({

   const handleRequestSizeChange = (size: RequestSize): void => {
     const { value, unit } = size;
+
+    const numberValue = parseInt(value, 10);
+    if (Number.isNaN(numberValue) || numberValue <= 0) {
+      setRequestSizeError(t('pipelines-plugin~Size must be an integer greater than 0.'));
+      setRequestSizeValue('');
+      return;
+    }
+
     setRequestSizeValue(value);
     setRequestSizeUnit(unit);
   };
@@ -171,8 +180,13 @@ const VolumeClaimTemplateForm: React.FC<VolumeClaimTemplateFormProps> = ({
           describedBy="request-size-help"
           inputID="request-size-input"
           data-test-id="pvc-size-input"
-        />
-        <p className="help-block">{t('pipelines-plugin~Desired storage capacity')}</p>
+        >
+          {requestSizeError ? (
+            <p className="pf-c-form__helper-text pf-m-error">{requestSizeError}</p>
+          ) : (
+            <p className="help-block">{t('pipelines-plugin~Desired storage capacity')}</p>
+          )}
+        </RequestSizeInput>
         <label className="control-label" htmlFor="volume-mode">
           {t('pipelines-plugin~Volume Mode')}
         </label>
```

andrewballantyne (changes requested):
Design has a gap between the dropdown and the alert for Empty Directory:
Screen Shot 2021-03-25 at 9 38 07 AM

Yours:
Screen Shot 2021-03-25 at 9 39 45 AM

There is also a `.` at the end of the Alert sentence.

---

The dropdown items should match the design. You have an odd gap in your `VolumeClaim Template`:
Screen Shot 2021-03-25 at 9 39 36 AM

Design:
Screen Shot 2021-03-25 at 9 39 13 AM

Could you also convert the PVC to the fully qualified `PersistentVolumeClaim`? Also if it's easy to do, please rearrange the order to match designs, I think that's cleaner imo.

---

It's very tight on your form, can you space out the fields a little more? See designs:
Screen Shot 2021-03-25 at 9 43 20 AM

Yours for comparison:
Screen Shot 2021-03-25 at 9 43 10 AM

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
abhinandan13jan: This will not disable the form unless the form field in set to an invalid value, I am doing that in line 98

abhinandan13jan:
@andrewballantyne I've changed to make the form aligned to designs and updated the latest screenshots. please revisit
Screenshot from 2021-03-30 16-16-05
Screenshot from 2021-03-30 16-16-13

andrewballantyne:
@abhinandan13jan I don't see a new push? I think you forgot to push the changes...

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  return (
+    <FormGroup fieldId={name}>
+      {errors[name] && <Alert isInline variant="danger" title={t('pipelines-plugin~Required')} />}
+      <p className="help-block">{helpText}</p>
+      <ExpandCollapse
+        textExpanded={t('pipelines-plugin~Hide VolumeClaimTemplate options')}
+        textCollapsed={t('pipelines-plugin~Show VolumeClaimTemplate options')}
+      >
+        <p>
```
andrewballantyne, line 135: You shouldn't use `p` tags as a means to just force content separation. It's not the intended use of the paragraph tag. If you elements were inline, then it could possibly be used, but still not in the case of a form.

Recommend you just change them to be `div` and add a className of `opp-VolumeClaimTemplateForm--element` or something along that lines. Create a bottom margin/padding of 8px or 4px or whatever is needed to make it similar to the design.

andrewballantyne (changes requested):
@abhinandan13jan You didn't address any of these...

Design has a gap between the dropdown and the alert for Empty Directory:
Screen Shot 2021-03-25 at 9 38 07 AM

Yours:
Screen Shot 2021-03-25 at 9 39 45 AM

There is also a `.` at the end of the Alert sentence.

---

The dropdown items should match the design. You have an odd gap in your `VolumeClaim Template`:
Screen Shot 2021-03-25 at 9 39 36 AM

Design:
Screen Shot 2021-03-25 at 9 39 13 AM

Could you also convert the PVC to the fully qualified `PersistentVolumeClaim`? Also if it's easy to do, please rearrange the order to match designs, I think that's cleaner imo.

abhinandan13jan:
> @abhinandan13jan You didn't address any of these...
> 
> Design has a gap between the dropdown and the alert for Empty Directory:
> Screen Shot 2021-03-25 at 9 38 07 AM
> 
> Yours:
> Screen Shot 2021-03-25 at 9 39 45 AM
> 
> There is also a `.` at the end of the Alert sentence.
> 
> The dropdown items should match the design. You have an odd gap in your `VolumeClaim Template`:
> Screen Shot 2021-03-25 at 9 39 36 AM
> 
> Design:
> Screen Shot 2021-03-25 at 9 39 13 AM
> 
> Could you also convert the PVC to the fully qualified `PersistentVolumeClaim`? Also if it's easy to do, please rearrange the order to match designs, I think that's cleaner imo.

@andrewballantyne I've updated. need re-review

andrewballantyne:
We have a Enum problem that finally caught up with us. Please accept my PR on your fork/branch: https://github.com/abhinandan13jan/console/pull/5

It'll fix this issue:
Screen Shot 2021-03-31 at 11 17 14 AM

abhinandan13jan:
> We have a Enum problem that finally caught up with us. Please accept my PR on your fork/branch: [abhinandan13jan#5](https://github.com/abhinandan13jan/console/pull/5)
> 
> It'll fix this issue:

Merged your PR @andrewballantyne

andrewballantyne (approved):

frontend/packages/pipelines-plugin/src/components/pipelines/const.ts
```diff
@@ -28,13 +28,21 @@ export const pipelineResourceTypeSelections = {
...
+  VolumeClaimTemplate = 'volumeClaimTemplate',
 }
 
+export const volumeTypeOptions: { [type in VolumeTypes]: string } = {
+  [VolumeTypes.EmptyDirectory]: 'Empty Directory',
+  [VolumeTypes.ConfigMap]: 'Config Map',
+  [VolumeTypes.Secret]: 'Secret',
+  [VolumeTypes.PVC]: 'PersistentVolumeClaim',
+  [VolumeTypes.VolumeClaimTemplate]: 'VolumeClaimTemplate',
+};
```
christianvogt, line 44: What about localization?

andrewballantyne: Whoops... guess that should be done too!

abhinandan13jan: updated

rottencandy (approved):
/lgtm

PR #8373, opened 2021-03-12
Update Dev perspective to show `Add` page on first visit

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5613

**Solution Description**: 
Add a flag to the perspective extension for the `GetLandingPage` indicating if the user has visited the perspective before. On a nav to the perspective, update the user settings to flag that the user has visited the perspective.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind feature

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useActivePerspective.ts
+++ b/frontend/packages/console-shared/src/hooks/useActivePerspective.ts
@@ -4,6 +4,12 @@ import {
   PerspectiveContextType,
   PerspectiveType,
 } from '@console/app/src/components/detect-perspective/perspective-context';
+import { USERSETTINGS_PREFIX } from '../constants';
+
+const PERSPECTIVE_VISITED_FEATURE_KEY = 'perspective.visited';
+
+export const getPerspectiveVisitedKey = (perspective: PerspectiveType): string =>
+  `${USERSETTINGS_PREFIX}.${PERSPECTIVE_VISITED_FEATURE_KEY}.${perspective}`;
 
 export const useActivePerspective = (): [
   PerspectiveType,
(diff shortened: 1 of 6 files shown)


frontend/public/components/app-contents.tsx
```diff
@@ -56,15 +61,29 @@ type DefaultPageProps = {
...
+  const firstVisit = React.useRef<boolean>();
+
+  if (Object.keys(flags).some((key) => flagPending(flags[key])) || !visitedLoaded) {
     return <LoadingBox />;
   }
   const perspective = perspectiveExtensions.find((p) => p.properties.id === activePerspective);
+
+  if (!visited) {
+    // Mark perspective as visited
+    setVisited(true);
```
christianvogt, line 77: Side effects should be in `useEffect`

frontend/packages/console-shared/src/hooks/useActivePerspective.ts
```diff
@@ -5,6 +5,11 @@ import {
   PerspectiveType,
 } from '@console/app/src/components/detect-perspective/perspective-context';
 
+const PERSPECTIVE_VISITED_SUFFIX = '_VISITED';
+
+export const getPerspectiveVisitedKey = (perspective: PerspectiveType): string =>
+  `${perspective}${PERSPECTIVE_VISITED_SUFFIX}`;
```
christianvogt, line 11: User settings keys should use reverse domain notation with the `console` prefix
```suggestion
export const getPerspectiveVisitedKey = (perspective: PerspectiveType): string =>
  `console.perspective.visited.${perspective}`;
```

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -210,7 +210,7 @@ const plugin: Plugin<ConsumedExtensions> = [
       name: '%devconsole~Developer%',
       icon: <CodeIcon />,
       defaultPins: [referenceForModel(ConfigMapModel), referenceForModel(SecretModel)],
-      getLandingPageURL: () => '/topology',
+      getLandingPageURL: (flags, isFirstVisit) => (isFirstVisit ? 'add' : '/topology'),
```
christianvogt, line 213:
```suggestion
      getLandingPageURL: (flags, isFirstVisit) => (isFirstVisit ? '/add' : '/topology'),
```

PR #8378, opened 2021-03-15
CONSOLE-2379: Add build config for dynamic plugin SDK

rawagner:
This PR enables us to build & publish dynamic plugin sdk.

### Bundling SDK
Sources are transpiled via `tsc`.
After the tsc build we also copy `LICENSE` and `README` files and generate `package.json` into `dist` folder.

### Publishing SDK
A new script `publish-sdk` handles publishing of the package into npmjs. The script bundles the sources and runs `yarn publish`. Only the content of `dist` folder is published. 
Currently the package is published as `@openshift-console/dynamic-plugin-sdk` but there's ongoing work to acquire `@openshift` acc/org on npmjs so we can publish there once available.

### Exposing Console's API
Console's API is exposed via `window.api`. Consuming plugin needs to configure webpack externals according to https://github.com/rawagner/console-dynamic-foo/tree/wp_externals#using-consoles-api
For now only k8s watch hooks are exposed.

### Real dynamic plugin using published SDK
I've also created dynamic demo plugin https://github.com/rawagner/console-dynamic-foo which lives outside the console repo & uses the released sdk package from npmjs. The project structure is the same as https://github.com/openshift/console/tree/master/frontend/dynamic-demo-plugin


frontend/packages/console-dynamic-plugin-sdk/src/extensions/index.ts
```diff
@@ -0,0 +1,12 @@
+export * from './add-actions';
```
rawagner: I re-added the index file to have an entry point for `webpack` build and single file .d.ts which is generated by `DtsBundleWebpack`

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pvc.ts
```diff
@@ -1,6 +1,5 @@
-import { Extension } from '@console/plugin-sdk/src/typings/base';
-import { ExtensionDeclaration, CodeRef } from '../types';
-import { ExtensionCommonK8sResource as PVCResource } from '../utils/common';
```
rawagner: there were issues with single `.d.ts` file genetated by `DtsBundleWebpack` when `foo as bar` syntax is used. The `PVCResource` was missing in the generated file

frontend/packages/console-dynamic-plugin-sdk/src/types.ts
```diff
@@ -1,4 +1,46 @@
-import { Extension } from '@console/plugin-sdk/src/typings/base';
+/**
+ * Console feature flags used to gate extension instances.
+ */
+export type ExtensionFlags = Partial<{
```
rawagner, line 4: I copied these types from `@console/plugin-sdk` to get rid of the dependency

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
...
+const EXTENSIONS_EXPORT = 'lib/extensions/index.js';
+const WEBPACK_PLUGIN_EXPORTS = './lib/webpack/ConsoleRemotePlugin.js';
+
+const createPackageJson = () => {
+  const packageJson = readPkg.sync({ normalize: false });
+  packageJson.name = PKG_NAME;
+  delete packageJson.private;
+  packageJson.license = 'Apache-2.0';
+  packageJson.main = EXTENSIONS_EXPORT;
+  packageJson.exports = {
```
rawagner, line 14: FYI, TS does not yet support `exports` in `package.json`. But it will be useful once added. For now, consumer of `dynamic-plugin-sdk` has to add `paths` into his `tsconfig.json`.

vojtechszocs (changes requested):

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
+import * as fs from 'fs';
+import * as readPkg from 'read-pkg';
+
+const PKG_NAME = '@openshift-console/dynamic-plugin-sdk';
```
christianvogt, line 4: Eventually we'll want to make sure that we omit `dynamic` from the package name. `plugin-sdk` is sufficient`

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
...
+const EXTENSIONS_EXPORT = 'lib/extensions/index.js';
+const WEBPACK_PLUGIN_EXPORTS = './lib/webpack/ConsoleRemotePlugin.js';
+
+const createPackageJson = () => {
```
christianvogt, line 8: We should look into maintaining a static package.json with proper values and up to date version.
This could be done by separating out everything that shouldn't be in the SDK into a separate `plugin-runtime` package. The idea being that everything in this package is only for consumption as public API.

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
...
+  const packageJson = readPkg.sync({ normalize: false });
+  packageJson.name = PKG_NAME;
+  delete packageJson.private;
+  packageJson.license = 'Apache-2.0';
+  packageJson.main = EXTENSIONS_EXPORT;
+  packageJson.exports = {
+    './': `./${EXTENSIONS_EXPORT}`,
+    './webpack': WEBPACK_PLUGIN_EXPORTS,
```
christianvogt, line 16: We should create a `public` folder convention for all our packages such that everything in the `public` dir is exposed api so it is very clear when we change API.
```
"exports": {
  "./*": "./src/public/*"
}
```

frontend/packages/console-dynamic-plugin-sdk/src/types.ts
christianvogt: +1 for reconciling as soon as we can.

frontend/packages/console-dynamic-plugin-sdk/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "moduleResolution": "node",
+    "target": "es2016",
+    "declaration": true,
+    "outDir": "dist/lib",
+    "experimentalDecorators": true,
+    "sourceMap": true,
+    "noUnusedLocals": true,
+    "skipLibCheck": true
+  },
+  "files": ["src/webpack/ConsoleRemotePlugin.ts", "src/extensions/index.ts"]
```
christianvogt, line 13: If we expose a `src/public/*` can we glob the public files ?

frontend/packages/console-dynamic-plugin-sdk/src/extensions/index.ts
rawagner: we need 3 different entries
 - 1. webpack (ConsoleRemotePlugin) so dynamic plugin wont bundle webpack in the build
 - 2. api - so we can replace the import with webpack externals
 - 3. extensions/index - the rest

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const PKG_NAME = '@openshift-console/dynamic-plugin-sdk';
+const EXTENSIONS_EXPORT = 'lib/extensions/index.js';
+const WEBPACK_PLUGIN_EXPORTS = './lib/webpack/ConsoleRemotePlugin.js';
+
+const createPackageJson = () => {
+  const packageJson = readPkg.sync({ normalize: false });
+  packageJson.name = PKG_NAME;
+  delete packageJson.private;
+  packageJson.license = 'Apache-2.0';
```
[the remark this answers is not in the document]
rawagner: overkill IMO :)

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
+import * as fs from 'fs';
+import * as readPkg from 'read-pkg';
+
+const PKG_NAME = '@openshift-console/dynamic-plugin-sdk';
+const EXTENSIONS_EXPORT = 'lib/extensions/index.js';
+const WEBPACK_PLUGIN_EXPORTS = './lib/webpack/ConsoleRemotePlugin.js';
+
+const createPackageJson = () => {
+  const packageJson = readPkg.sync({ normalize: false });
+  packageJson.name = PKG_NAME;
```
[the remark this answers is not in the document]
rawagner: ok, inlined

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
+import * as fs from 'fs';
```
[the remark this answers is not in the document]
rawagner: renamed

frontend/packages/console-dynamic-plugin-sdk/package.json
```diff
@@ -7,18 +7,23 @@
...
     "build": "yarn clean && yarn validate && yarn generate",
+    "bundle": "yarn build && yarn tsc && yarn prepare-dist",
     "generate": "yarn generate-schema && yarn generate-doc",
     "generate-schema": "yarn ts-node ./scripts/generate-schema.ts",
     "generate-doc": "yarn ts-node ./scripts/generate-doc.ts",
     "validate": "yarn ts-node ./scripts/validate-extensions.ts",
-    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}'"
+    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}'",
+    "prepare-dist": "yarn ts-node ./scripts/create-pkg.ts && cp ./README.md ./dist && cp ../../../LICENSE ./dist",
+    "publish-sdk": "yarn bundle && yarn publish dist --no-git-tag-version"
```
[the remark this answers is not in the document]
rawagner: renamed

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
```diff
@@ -0,0 +1,26 @@
...
+  };
+  packageJson.readme = './README.md';
+  packageJson.peerDependencies = packageJson.dependencies;
+  delete packageJson.dependencies;
+  delete packageJson.devDependencies;
+  delete packageJson.scripts;
+  fs.writeFileSync(`dist/package.json`, JSON.stringify(packageJson, null, 2));
+};
+
+createPackageJson();
```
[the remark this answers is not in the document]
rawagner: Script updated.
Now it generates package.json, copies LICENCE, README and schema to dist dir.
Readme is sanitized too

frontend/packages/console-dynamic-plugin-sdk/src/utils/schema.ts
```diff
@@ -0,0 +1,7 @@
+import * as path from 'path';
+import * as fs from 'fs';
+
+export const getSchemaPath = (fileName: string): string =>
```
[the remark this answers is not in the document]
rawagner: `schema` is generated to sdk package root now and copied during the build to `dist`. The utility function is not needed anymore, thus removed.
I've also added the `schema` dir to gitignore and eslint ignore

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
rawagner: I agree. Once we acquire @openshift acc on npmjs, I think the package name should be @openshift/console-plugin-sdk

frontend/packages/console-dynamic-plugin-sdk/tsconfig.json
rawagner: we can use `include` instead of `files` in that case

frontend/packages/console-dynamic-plugin-sdk/scripts/create-pkg.ts
rawagner: OK, that will simplify things. I will look into it in a followup

rawagner:
rebased on latest master

jelkosz:
just adding the 
/lgtm
back

jelkosz:
a push just to try to get rid of the needs-rebase label. Adding the
/lgtm
back

jelkosz:
hmmm will
/test all
run also a test for the needs-rebase?

PR #8380, opened 2021-03-15
Bug 1936788: Fix for RBD + RWX Access Mode  UI issue

GowthamShanmugam:


GowthamShanmugam:
image

GowthamShanmugam:
I feel the correct approach would be based on VolumeMode we have to enable and disable access mode. But i have to stick with design, volumeMode is the bottom-most option, so changing any option based on this is not correct.

GowthamShanmugam:
duplicate of: https://github.com/openshift/console/pull/8210

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -279,9 +284,14 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-        {storageProvisioner.includes(cephStorageProvisioners[1]) ? (
+        {allowedVolumeModes.length === 1 ? (
           <>
-            Filesystem <InfoToolTip />
+            {volumeMode}{' '}
+            <InfoToolTip
+              accessMode={accessMode}
+              volumeMode={volumeMode}
+              storageClass={storageClass}
+            />
```
GowthamShanmugam, line 294: Generic for all provisioners

spadgett:
Hi, this PR has build errors you'll need to address:

```

ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx(40,3):
TS2305: Module '"../../../../../../public/components/storage/shared"' has no exported member 'snapshotPVCStorageClassAnnotation'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx(41,3):
TS2305: Module '"../../../../../../public/components/storage/shared"' has no exported member 'snapshotPVCAccessModeAnnotation'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx(42,3):
TS2305: Module '"../../../../../../public/components/storage/shared"' has no exported member 'snapshotPVCVolumeModeAnnotation'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx(42,3):
TS2305: Module '"../../../../../../public/components/storage/shared"' has no exported member 'snapshotPVCStorageClassAnnotation'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx(43,3):
TS2305: Module '"../../../../../../public/components/storage/shared"' has no exported member 'snapshotPVCAccessModeAnnotation'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx(44,3): 
```

frontend/public/components/storage/shared.ts
```diff
@@ -63,3 +63,13 @@ export const getAccessModeForProvisioner = (provisioner: string) => {
     ? ['ReadWriteOnce', 'ReadWriteMany']
     : ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'];
 };
+
+export const getVolumeModeForProvisioner = (provisioner: string, accessMode: string): string[] => {
+  return provisioner && isCephProvisioner(provisioner)
+    ? Object.keys(cephProvisionerAccessModes[provisioner])
+        .filter((volumeMode) =>
+          cephProvisionerAccessModes[provisioner][volumeMode].includes(accessMode),
```
cloudbehl, line 71:
```suggestion
          provisionerAccessModes[provisioner][volumeMode].includes(accessMode),
```

frontend/public/components/storage/shared.ts
```diff
@@ -63,3 +63,13 @@ export const getAccessModeForProvisioner = (provisioner: string) => {
     ? ['ReadWriteOnce', 'ReadWriteMany']
     : ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'];
 };
+
+export const getVolumeModeForProvisioner = (provisioner: string, accessMode: string): string[] => {
+  return provisioner && isCephProvisioner(provisioner)
+    ? Object.keys(cephProvisionerAccessModes[provisioner])
```
cloudbehl, line 69:
```suggestion
    ? Object.keys(provisionerAccessModes[provisioner])
```

cloudbehl (commented):
This can be simplified as well as we have already declared `provisionerAccessModes`.

frontend/packages/console-shared/src/utils/storage-utils.ts
```diff
@@ -20,3 +20,23 @@ export const isCephProvisioner = (scProvisioner: string): boolean => {
...
 
 export const isObjectSC = (sc: StorageClass) => objectStorageProvisioners.includes(sc.provisioner);
+
+export enum volumeModes {
+  // ceph supported volum modes
+  Block,
+  Filesystem,
+}
+
+export const cephProvisionerAccessModes: {
```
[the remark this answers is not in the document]
GowthamShanmugam: ack

frontend/packages/console-shared/src/utils/storage-utils.ts
```diff
@@ -20,3 +20,23 @@ export const isCephProvisioner = (scProvisioner: string): boolean => {
 };
 
 export const isObjectSC = (sc: StorageClass) => objectStorageProvisioners.includes(sc.provisioner);
+
+export enum volumeModes {
+  // ceph supported volum modes
```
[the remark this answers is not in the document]
GowthamShanmugam: ack

frontend/public/components/storage/shared.ts
GowthamShanmugam: ack

frontend/public/components/storage/shared.ts
GowthamShanmugam: ack

vbnrh (changes requested):

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -13,19 +13,24 @@ import { Checkbox } from '../checkbox';
...
   initialAccessModes,
   accessModeRadios,
   volumeModeRadios,
   dropdownUnits,
   getAccessModeForProvisioner,
+  getVolumeModeForProvisioner,
 } from './shared';
 
-const InfoToolTip = () => (
+const InfoToolTip = (props) => (
```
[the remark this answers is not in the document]
GowthamShanmugam: InfoToolTip is our own function, and it uses <Tooltip> which already comes from the library only, It is not necessary to call this function, we can directly use <Tooltip>

frontend/packages/console-shared/src/utils/storage-utils.ts
```diff
@@ -20,3 +20,23 @@ export const isCephProvisioner = (scProvisioner: string): boolean => {
...
+  Filesystem,
+}
+
+export const cephProvisionerAccessModes: {
+  [key: string]: {
+    [key in keyof typeof volumeModes]?: string[];
+  };
+} = {
+  'openshift-storage.rbd.csi.ceph.com': {
+    Block: ['ReadWriteOnce', 'ReadWriteMany'],
```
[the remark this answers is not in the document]
vbnrh: Also can we create an enum for these access modes. Used in many places.

frontend/public/components/storage/create-pvc.tsx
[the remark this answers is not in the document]
vbnrh: Yes but still it would best to make it accessible from a common point. We are using it in bucket class page, pvc page and will most likely be using it in restore pvc modal also. It would avoid the trouble of creating it in every other page.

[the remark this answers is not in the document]
GowthamShanmugam: No, it does not make any sense, the way we are using it here wrong. we don't have to add one more layer on top of  Tooltop, no need to call this InfoToolTip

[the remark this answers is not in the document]
GowthamShanmugam: if we are combining more than one element and come up with the single jsx element then it makes sense, for example checkbox (label + input)

frontend/packages/console-shared/src/utils/storage-utils.ts
```diff
@@ -20,3 +20,23 @@ export const isCephProvisioner = (scProvisioner: string): boolean => {
...
+    Filesystem: ['ReadWriteOnce'],
+  },
+  'openshift-storage.cephfs.csi.ceph.com': {
+    Filesystem: ['ReadWriteOnce', 'ReadWriteMany'],
```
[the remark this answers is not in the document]
GowthamShanmugam: ack

frontend/packages/console-shared/src/utils/storage-utils.ts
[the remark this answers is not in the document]
GowthamShanmugam: ack

frontend/packages/console-shared/src/utils/storage-utils.ts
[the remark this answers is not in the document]
GowthamShanmugam: agreed, right now what we are doing is a hack, need to avoid this for creating PVC and need to include it for clone and snapshot.

frontend/public/components/storage/create-pvc.tsx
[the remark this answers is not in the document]
vbnrh: Yes exactly ! We are using `<InfoToolTip> = <Tooltip /> + <OutlinedQuestionCircleIcon />` so that when someone hovers over the `?` icon, i can view the information.  We need to try to make this information customizable in other areas. 

If we don't make it generic, We will have to add the icon and tooltip specifically wherever we want to use it . I hope it makes sense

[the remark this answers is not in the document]
GowthamShanmugam: Ah! ok i never saw OutlinedQuestionCircleIcon

GowthamShanmugam:
image

frontend/public/components/storage/shared/shared.ts
```diff
@@ -0,0 +1,164 @@
...
+  return possibleAccessModes;
+};
+
+export const getAccessModeForProvisioner = (
+  provisioner: string,
+  ignoreReadOnly?: boolean,
+  volumeMode: string = '',
+): string[] => {
+  const accessModeMapping = provisionerAccessModeMapping[provisioner] || {};
+  const selectedAccessModes: string[] = accessModeMapping[volumeMode]
```
bipuladh, line 141:
```suggestion
  const selectedAccessModes: AccessMode[] = accessModeMapping[volumeMode]
```

frontend/public/components/storage/shared/shared.ts
```diff
@@ -0,0 +1,164 @@
...
+    ? accessModeMapping[volumeMode]
+    : accessModeMapping?.All ?? [];
```
bipuladh, line 143:
```suggestion
  const selectedAccessModes: AccessMode[] = accessModeMapping[volumeMode]
   || accessModeMapping?.All || [];
```

frontend/public/components/storage/shared/volume-modes/volume-mode.tsx
```diff
@@ -0,0 +1,104 @@
...
+    label,
+  } = props;
+
+  const { t } = useTranslation();
+
+  const pvcInitialVolumeMode: string = pvcResource
+    ? pvcResource?.spec?.volumeMode
+    : availableVolumeMode ?? initialVolumeModes[0];
+
+  const [volumeMode, setVolumeMode] = React.useState('');
```
bipuladh, line 29:
```suggestion
  const [volumeMode, setVolumeMode] = React.useState<VolumeMode>(null);
```

bipuladh (changes requested):

frontend/public/components/storage/shared/shared.ts
```diff
@@ -0,0 +1,164 @@
...
+  Filesystem = 'Filesystem',
+  Block = 'Block',
+}
+
+export enum accessModes {
+  ReadWriteOnce = 'ReadWriteOnce',
+  ReadWriteMany = 'ReadWriteMany',
+  ReadOnlyMany = 'ReadOnlyMany',
+}
+
```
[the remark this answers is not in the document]
GowthamShanmugam: Why this needs to be a type, we are not going to initialize or not going to create multiple var for this.

frontend/public/components/storage/info-tooltip.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import { Tooltip, TooltipProps } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
```
GowthamShanmugam, line 3: Do i need to move this file inside console/shared or public/component?

GowthamShanmugam:
clone-pvc
create-pvc
Restore

frontend/public/components/storage/shared.ts
```diff
@@ -58,8 +123,62 @@ export const dropdownUnits = {
...
+  // AccessMode or All
+  const accessModeList: string[] = combinationsForProvisioner.map((combination) => combination[0]);
+
+  return ignoreReadOnly
+    ? accessModeList.filter((modes) => modes !== AccessModes.ROX)
+    : accessModeList;
+};
+
+export const getVolumeModeForProvisioner = (provisioner: string, accessMode: string): string[] => {
+  let combinationsForProvisioner: AllowedCombinations = getAllowedCombiationsForProvisioner(
```
GowthamShanmugam, line 172: I feel let's return this as string[] for easy use,

GowthamShanmugam:
@cloudbehl @vbnrh Changes done

frontend/public/components/storage/shared.ts
```diff
@@ -1,55 +1,120 @@
...
+    allowedCombinations[1],
+    allowedCombinations[3],
+    allowedCombinations[5],
+  ],
 };
-export const initialAccessModes = ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'];
+
+export const initialAccessModes: string[] = Object.values(AccessModes);
+export const initialVolumeModes: string[] = Object.values(VolumeModes);
+
```
GowthamShanmugam, line 96: The reason I am still keeping string[] here and below function is, a lot of plugins are using this, if I want to return as enum i have to change their plugin also, that is a little bit difficult task for n

GowthamShanmugam:
@cloudbehl Please review

GowthamShanmugam:
Replaced Tooltip with FieldLevelHelp as per: https://github.com/openshift/console/commit/7eecdd559e84565b8f785b76056040488ce2c5f9#diff-471f45f63e6144addaf2694c56d1698efaeeb1f74f649907cdc3d7f70068f9f4

pvc_fix_1
pvc_fix

frontend/packages/console-app/src/components/modals/clone/clone-pvc-modal.tsx
```diff
@@ -139,9 +142,11 @@ const ClonePVCModal = withHandlePromise((props: ClonePVCModalProps) => {
             />
           </FormGroup>
           <AccessModeSelector
+            label={t('console-app~Access mode')}
             onChange={handleAccessMode}
             formClassName="co-clone-pvc-modal__form--space"
             pvcResource={resource}
+            volumeMode={resource?.spec?.volumeMode}
```
spadgett, line 149: Is it possible for `AccessModeSelector` to read the `volumeMode` from `resource` directly since we're already passing the resource?

frontend/public/components/storage/shared.ts
```diff
@@ -1,55 +1,120 @@
...
+  ['All'],
+];
+
+type ProvisionerAccessModeMapping = {
+  [provisioner: string]: AllowedCombinations;
+};
+
+// See https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes for more details
+export const provisionerAccessModeMapping: ProvisionerAccessModeMapping = {
+  'kubernetes.io/no-provisioner': [allowedCombinations[0], allowedCombinations[1]],
```
spadgett, line 36: Values like `allowedCombinations[0]` are really opaque and have no meaning outside of where they're defined. This really obscures the code to me. I prefer the previous implementation where the values were inline. If we're worried about typos, we should simply create constants.

If we need separate values for Filesystem and Block, create separate maps instead of nested arrays.

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -102,8 +99,7 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
       setUpdatedProvisioner(updatedStorageClass?.provisioner);
     };
 
-    const handleVolumeMode: React.ReactEventHandler<HTMLInputElement> = (event) =>
-      setVolumeMode(event.currentTarget.value);
+    const handleVolumeMode = (value: string) => setVolumeMode(value);
     const handleAccessMode = (value: string) => setRestoreAccessMode(value);
```
spadgett, line 103: These helpers aren't necessary. Just use `setVolumeMode` and `setRestoreAccessMode` directly since they take the same parameter values.

frontend/packages/console-app/src/components/modals/clone/clone-pvc-modal.tsx
```diff
@@ -139,9 +142,11 @@ const ClonePVCModal = withHandlePromise((props: ClonePVCModalProps) => {
             />
           </FormGroup>
           <AccessModeSelector
+            label={t('console-app~Access mode')}
```
spadgett, line 145: Do we ever use anything other than "Access mode" for the label? If not, do we need a `label` prop?

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -144,148 +140,140 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
...
+          <VolumeModeSelector
+            onChange={handleVolumeMode}
+            formClassName="co-restore-pvc-modal__input"
+            provisioner={updatedProvisioner}
+            pvcResource={pvcResource}
+            accessMode={restoreAccessMode}
+            storageClass={pvcSC}
+            loaded={pvcResourceLoaded}
+            availableVolumeMode={volumeSnapshotAnnotations?.[snapshotPVCVolumeModeAnnotation]}
+            label={t('console-app~Volume mode')}
```
spadgett, line 202: Is `label` necessary or can we always use "Volume mode"?

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -144,148 +140,140 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
       );
     };
     return (
-      <div className="modal-content modal-content--no-inner-scroll">
+      <form onSubmit={submit} name="form" className="modal-content modal-content--no-inner-scroll">
```
spadgett, line 143: Why do we need to use `form` in place of the Patternfly `Form` component?

frontend/public/components/storage/access-modes/access-mode.tsx
```diff
@@ -28,33 +22,28 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) =>
...
     formClassName,
     pvcResource,
+    volumeMode,
     onChange,
     loaded,
     provisioner,
     availableAccessModes = [],
+    label,
+    description,
+    ignoreReadOnly,
```
spadgett, line 32: Can you describe what `ignoreReadOnly` does and why we need it?

frontend/public/components/storage/access-modes/access-mode.tsx
```diff
@@ -78,10 +67,11 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) =>
 
   return (
     <FormGroup
-      label={t('console-app~Access Mode')}
+      label={label}
       isRequired
       fieldId="access-mode"
       className={formClassName}
+      htmlFor="access-mode"
```
spadgett, line 74: I'm a little unsure if this is right. Is there an element with an `access-mode` ID?

frontend/public/components/storage/access-modes/access-mode.tsx
```diff
@@ -100,6 +90,11 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) =>
             />
           );
         })}
+      {allowedAccessModes && allowedAccessModes && description && (
+        <p className="help-block" id="access-mode-help">
```
spadgett, line 94: You need an `aria-described-by="access-mode-help"` on the input.

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -156,13 +131,8 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
 
-  const handleAccessMode: React.ReactEventHandler<HTMLInputElement> = (event) => {
-    setAccessMode(event.currentTarget.value);
-  };
-
-  const handleVolumeMode: React.ReactEventHandler<HTMLInputElement> = (event) => {
-    setVolumeMode(event.currentTarget.value);
-  };
+  const handleAccessMode = (value: string) => setAccessMode(value);
+  const handleVolumeMode = (value: string) => setVolumeMode(value);
```
spadgett, line 135: These helpers are unnecessary as they have the same signature as the functions they wrap.

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -199,35 +169,17 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-        })}
-        <p className="help-block" id="access-mode-help">
-          {accessModeHelp}
-        </p>
+        <AccessModeSelector
+          label={t('public~Access mode')}
+          formClassName="control-label"
+          onChange={handleAccessMode}
+          provisioner={storageProvisioner}
+          loaded={true}
```
spadgett, line 178:
```suggestion
          loaded
```

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -199,35 +169,17 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
+          availableAccessModes={initialAccessModes}
+          description={accessModeHelp}
+          ignoreReadOnly={true}
```
spadgett, line 181:
```suggestion
          ignoreReadOnly
```

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -270,27 +222,16 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-              checked={radio.value === volumeMode}
-              name="volumeMode"
-            />
-          ))
-        )}
+        <VolumeModeSelector
+          onChange={handleVolumeMode}
+          provisioner={storageProvisioner}
+          accessMode={accessMode}
+          formClassName="control-label"
```
spadgett, line 230: This seems like the wrong class to put on a form group

frontend/public/components/storage/shared.ts
spadgett: > The reason I am still keeping string[] here and below function is, a lot of plugins are using this, if I want to return as enum i have to change their plugin also, that is a little bit difficult task for n

I might just remove the use of enum and keep it as a string for now. We could create a type for type checking like

```
export type AccessMode = 'ReadWriteOnce' | 'ReadWriteMany' | 'ReadOnlyMany';
```

frontend/public/components/storage/shared.ts
```diff
@@ -58,8 +123,64 @@ export const dropdownUnits = {
...
+
+const filterAllowedCombiations = (
+  combinationsForProvisioner: AllowedCombinations,
+  mode: string,
+  value: string,
+): AllowedCombinations => {
+  return value
+    ? mode === 'volumeMode'
+      ? combinationsForProvisioner.filter((combination) => combination[1] === value)
+      : combinationsForProvisioner.filter((combination) => combination[0] === value)
```
spadgett, line 143: Avoid nested ternaries as they are hard to understand.

I'm a little unclear on the logic here. I think this is confusing because we're trying to combine too many things in a single nested array instead of having separate arrays.

frontend/public/components/storage/volume-modes/volume-mode.tsx
```diff
@@ -0,0 +1,99 @@
...
+    : availableVolumeMode ?? initialVolumeModes[0];
+
+  const [volumeMode, setVolumeMode] = React.useState('');
+  const [allowedVolumeModes, setAllowedVolumeModes] = React.useState<string[]>([]);
+
+  React.useEffect(() => {
+    if (loaded) {
+      setAllowedVolumeModes(getVolumeModeForProvisioner(provisioner, accessMode));
+    }
+  }, [loaded, provisioner, accessMode]);
```
spadgett, line 34: This doesn't need to be stored in state. If you want to avoid recomputing the value on every render, you can use `useMemo` instead.

frontend/public/components/storage/volume-modes/volume-mode.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  React.useEffect(() => {
+    // Makesure the default or already checked radio button value is from any one of allowed the access mode
+    if (allowedVolumeModes.length > 0) {
```
spadgett, line 38: Prefer early returns like this which are easier to understand.

```
if (!allowedVolumeModes.length) {
  return;
}

// ... remaining logic ...

File: frontend/public/components/storage/volume-modes/volume-mode.tsx [hunk 32]
```diff
@@ -0,0 +1,99 @@
...
+      if (volumeMode === '' && allowedVolumeModes.includes(pvcInitialVolumeMode)) {
```
spadgett, line 39: Unless you really need to distinguish between `''` and `null` or `undefined` use

```suggestion
      if (!volumeMode && allowedVolumeModes.includes(pvcInitialVolumeMode)) {
```

frontend/public/components/storage/volume-modes/volume-mode.tsx
```diff
@@ -0,0 +1,99 @@
...
+      )}
+    </FormGroup>
+  );
+};
+
+type VolumeModeSelectorProps = {
+  formClassName?: string;
+  pvcResource?: PersistentVolumeClaimKind;
+  accessMode: string;
+  onChange: Function;
```
spadgett, line 93: Specify the actual function signature.

frontend/public/components/storage/volume-modes/volume-mode.tsx
```diff
@@ -0,0 +1,99 @@
...
+    onChange(event.currentTarget.value);
+  };
+
+  return (
+    <FormGroup
+      fieldId="volume-mode"
+      className={formClassName}
+      label={label}
+      isRequired
+      htmlFor="volume-mode"
```
spadgett, line 62: Is there an element with this ID?

spadgett (changes requested):

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
GowthamShanmugam: ack

frontend/public/components/storage/shared.ts
GowthamShanmugam: Here the expected logic is for each provisioner we need to allow some volume mode and access mode combinations, we can't create a map for this volume mode these are the access modes or for these access modes these are the volume modes. 

it should be a combination of both, that why I created a array of supported combinations.

instead of index values 0 or 1 ... can i use combinations directly here like 

[[AccessModes.RWO, VolumeModes.Filesystem], ...]

frontend/public/components/storage/shared.ts
GowthamShanmugam: ack

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
GowthamShanmugam: ack

frontend/public/components/storage/shared.ts
GowthamShanmugam: For now, we dont have a complex use case, so let me make it simple:
```
as per UI we select access mode first and then volume mode :

type ProvisionerAccessModeMapping = {
  [provisioner: string]: {
    [accessMode in AccessModes]?: VolumeModes[]
  };
};

export const provisionerAccessModeMapping: ProvisionerAccessModeMapping = {
  'kubernetes.io/no-provisioner': {
    ReadWriteOnce: [VolumeModes.Filesystem, VolumeModes.Block],
  },
}


```

is this make sense?

frontend/public/components/storage/access-modes/access-mode.tsx
GowthamShanmugam: Let assume the provisioner supports three access modes: RWO, RWX, ROX, we have a common function to return all supported accessModes. 
There is no point in returning ROX (Read-only) access mode for create-pvc page. so from there we can pass ignore read-only flag the return other excess mode except ROX

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -144,9 +137,9 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
       );
     };
     return (
-      <div className="modal-content modal-content--no-inner-scroll">
-        <ModalTitle>{t('console-app~Restore as new PVC')}</ModalTitle>
-        <Form onSubmit={submit} name="form">
```
GowthamShanmugam, line 149: moved Form to fix some glitch in UI

frontend/public/components/storage/shared.ts
```diff
@@ -58,8 +133,48 @@ export const dropdownUnits = {
...
+      : Object.keys(modeMapping)
+          .map((mode) => modeMapping[mode])
+          .flat();
+  } else {
+    accessModes = initialAccessModes;
+  }
+
+  // remove duplicate in accessModes
+  accessModes = [...new Set(accessModes)];
+
```
GowthamShanmugam, line 161: volumeMode wise access mode may repeatt

frontend/public/components/storage/shared.ts
```diff
@@ -58,8 +133,48 @@ export const dropdownUnits = {
...
+): AccessMode[] => {
+  let accessModes: AccessMode[];
+  const modeMapping: ModeMapping = getProvisionerAccessModeMapping(provisioner);
+
+  if (!_.isEmpty(modeMapping)) {
+    accessModes = volumeMode
+      ? modeMapping[volumeMode]
+      : Object.keys(modeMapping)
+          .map((mode) => modeMapping[mode])
+          .flat();
```
GowthamShanmugam, line 154: {"Block": ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'], ....}

GowthamShanmugam:
@spadgett Changes done, Please review

GowthamShanmugam:
@spadgett Please review

GowthamShanmugam:
@spadgett Please review, all changes done

GowthamShanmugam:
@spadgett Yes, this PR has few conflicts with  #8505 related to i18n

frontend/public/components/storage/shared.ts
```diff
@@ -58,8 +133,48 @@ export const dropdownUnits = {
...
+  } else {
+    accessModes = initialAccessModes;
+  }
+
+  // remove duplicate in accessModes
+  accessModes = [...new Set(accessModes)];
```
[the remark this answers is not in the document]
GowthamShanmugam: Yes, i am always using getAccessModeRadios, getVolumeModeRadios function to display radios buttons. These are hardcoded list and always returns same order. 

I am just finding allowed volume and access modes for each provisioner and comparing that with above hardcoded list to decide which radio button to be enabled and disabled.

GowthamShanmugam:
@spadgett Changes done

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
```diff
@@ -61,14 +60,11 @@ const VolumeClaimTemplateForm: React.FC<VolumeClaimTemplateFormProps> = ({
...
-    const modes = provisionerAccessModeMapping[provisioner]
-      ? provisionerAccessModeMapping[provisioner]
-      : getAccessModeForProvisioner(provisioner);
+    const modes = getAccessModeForProvisioner(provisioner);
     // setting message to display for various modes when a storage class of a know provisioner is selected
-    const displayMessage =
-      provisionerAccessModeMapping[provisioner] || isCephProvisioner(provisioner)
-        ? t('pipelines-plugin~Access mode is set by storage class and cannot be changed')
-        : t('pipelines-plugin~Permissions to the mounted drive');
+    const displayMessage = getAccessModeForProvisioner(provisioner)
```
spadgett, line 65:
```suggestion
    const displayMessage = modes
```

frontend/public/components/storage/access-modes/access-mode.tsx
```diff
@@ -26,62 +22,64 @@ export const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: stri
...
-      let currentModes: string[];
-      provisioner?.includes(cephRBDProvisionerSuffix) && pvcResource?.spec.volumeMode !== 'Block'
-        ? (currentModes = ['ReadWriteOnce', 'ReadOnlyMany'])
-        : (currentModes = updateAllowedAccessModes(provisioner));
-      setAllowedAccessModes(currentModes);
+      setAllowedAccessModes(
+        getAccessModeForProvisioner(
+          provisioner,
+          ignoreReadOnly,
+          filterByVolumeMode && pvcResource?.spec?.volumeMode,
```
spadgett, line 58: Won't think sometime return a boolean and sometime a string? It would be better to do something like

```suggestion
          filterByVolumeMode ? pvcResource?.spec?.volumeMode : null,
```

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -271,29 +218,15 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-              checked={radio.value === volumeMode}
-              name="volumeMode"
-            />
-          ))
-        )}
+        <VolumeModeSelector
+          onChange={setVolumeMode}
+          provisioner={storageProvisioner}
+          accessMode={accessMode}
+          className=""
```
spadgett, line 226: You should be able to remove this prop instead of setting it to an empty string. (Presumably `className` is optional.)

frontend/public/components/storage/access-modes/access-mode.tsx
```diff
@@ -1,15 +1,11 @@
+import * as _ from 'lodash';
```
spadgett, line 1: Can you explain why this was moved to the public folder? We're actually trying to migrate the components the other way (from public to console-shared or console-app).

frontend/public/components/storage/create-pvc.tsx
GowthamShanmugam: ack

frontend/public/components/storage/access-modes/access-mode.tsx
GowthamShanmugam: let me check this

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
GowthamShanmugam: let me check this

frontend/public/components/storage/access-modes/access-mode.tsx
GowthamShanmugam: These two files can reused by other if it is in public, If it is not correct, can we move create-pvc also somewhere else with these two files?

GowthamShanmugam:
@spadgett I addressed all review changes, please review

frontend/public/components/storage/access-modes/access-mode.tsx
spadgett: console-shared is the right package for reusable components.

frontend/packages/console-app/locales/en/console-app.json
```diff
@@ -28,18 +28,17 @@
...
   "Restore as new PVC": "Restore as new PVC",
   "When restore action for snapshot <1>{{snapshotName}}</1> is finished a new crash-consistent PVC copy will be created.": "When restore action for snapshot <1>{{snapshotName}}</1> is finished a new crash-consistent PVC copy will be created.",
-  "Volume Mode": "Volume Mode",
   "Size should be equal or greater than the restore size of snapshot": "Size should be equal or greater than the restore size of snapshot",
   "{{resourceKind}} details": "{{resourceKind}} details",
   "Created at": "Created at",
   "Status": "Status",
   "API version": "API version",
   "Restore": "Restore",
+  "Only {allowedVolumeModes[0]} volume mode is available for {storageClass} with {accessMode} access mode": "Only {allowedVolumeModes[0]} volume mode is available for {storageClass} with {accessMode} access mode",
```
spadgett, line 41: The first variable going to be a little confusing for the translators. It would be better if we can avoid in-lining the logic in the message.

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -26,62 +25,64 @@ export const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: stri
...
+      setAllowedAccessModes(
+        getAccessModeForProvisioner(
+          provisioner,
+          ignoreReadOnly,
+          filterByVolumeMode ? pvcResource?.spec?.volumeMode : null,
+        ),
+      );
     }
-  }, [loaded, provisioner, pvcResource]);
+  }, [filterByVolumeMode, ignoreReadOnly, loaded, provisioner, pvcResource]);
```
spadgett, line 65: It would be better if we used `pvcResource?.spec?.volumeMode` as the dependency instead of `pcvResource`. That way the hook isn't reevaluated if other `pvcResource` values change, only the one we're using.

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
spadgett: > moved Form to fix some glitch in UI

Can you explain what the glitch is you needed to fix? It's unclear to me why moving the `Form` should have been necessary.

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -200,35 +166,16 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
...
-            );
-          });
-
-          return radioObj;
-        })}
-        <p className="help-block" id="access-mode-help">
-          {accessModeHelp}
-        </p>
+        <AccessModeSelector
+          className=""
```
spadgett, line 171: You should be able to remove this prop instead of setting it to `""`

GowthamShanmugam:
@spadgett glitch in current snapshot restore page
image

frontend/packages/console-app/locales/en/console-app.json
GowthamShanmugam: ack

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
GowthamShanmugam: ack

frontend/public/components/storage/create-pvc.tsx
GowthamShanmugam: Ah! i missed this twice, let me fix this correctly this time

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -26,62 +25,64 @@ export const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: stri
...
+      setAllowedAccessModes(
+        getAccessModeForProvisioner(
+          provisioner,
+          ignoreReadOnly,
+          filterByVolumeMode ? pvcResource?.spec?.volumeMode : null,
+        ),
+      );
     }
-  }, [loaded, provisioner, pvcResource]);
+  }, [filterByVolumeMode, ignoreReadOnly, loaded, provisioner, pvcResource?.spec?.volumeMode]);
```
GowthamShanmugam, line 65: I think lint will modify  this to pvcResource from pvcResource?.spec?.volumeMode

spadgett (commented):
> @spadgett glitch in current snapshot restore page(failing when long storage class dropdown item)
> 
> image

That seems like a problem with the dropdown styles we should fix so it doesn't grow like that rather than moving the `Form`.

cc @sg00dwin @rhamilto

GowthamShanmugam:
After fix

image

GowthamShanmugam:
@spadgett  But what i observed is everywhere it used like this
<Form onSubmit={submit}>
      <div className="modal-content modal-content--no-inner-scroll">
      
  nowhere Form is used inside <div className="modal-content modal-content--no-inner-scroll">

GowthamShanmugam:
@spadgett All changes are done

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
spadgett, line 65: Hi, it looks like you still have `pvcResource` in the access list here. You might need to assign `pvcResource?.spec?.volumeMode` to a var outside of the `useEffect` hook and use that as the dependency.

spadgett:
@GowthamShanmugam I'm proposing we make a CSS fix to the dropdown so that it doesn't grow outside of the dialog rather than moving the `Form` component to a different part of the DOM. @rhamilto or @sg00dwin might be able to help here since they're much better at CSS than I am :)

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
GowthamShanmugam: no no, linit is changing form pvcResource?.spec?.volumeMode to pvcResource, i tired

GowthamShanmugam:
@spadgett i fixed pvcResource?.spec?.volumeMode  also, for CSS i will talk to @rhamilto and @sg00dwin

GowthamShanmugam:
> > @spadgett glitch in current snapshot restore page(failing when long storage class dropdown item)
> > image
> 
> That seems like a problem with the dropdown styles we should fix so it doesn't grow like that rather than moving the `Form`.
> 
> cc @sg00dwin @rhamilto

@sg00dwin @rhamilto Need help here,  <StorageClassDropdown> behaves weird when dropdown item is long, ideally, we don't need to display description and when expand it should be folded. Any idea how to fix this

rhamilto:
> @sg00dwin @rhamilto Need help here, behaves weird when dropdown item is long, ideally, we don't need to display description and when expand it should be folded. Any idea how to fix this

Cursory investigation suggests this bug only occurs when `<StorageClassDropdown>` is inside a modal.  Struggling to figure out how to get the modal to render.  Tips on getting a `VolumeSnapshot` set up?

GowthamShanmugam:
@rhamilto 

PersistentVolumeClaims-> click kabab menu -> Create snapshot

image

rhamilto:
> @spadgett But what i observed is everywhere it used like this
> 
> nowhere <Form> is used inside "<div className="modal-content modal-content--no-inner-scroll">"

Yep.  We've got a bug resulting in the use of `<Form>` with our modal styles [1].  The quickest and safest fix for now is to not use `<Form>` and use a straight `<form>` instead. 

For example:
``` 
      <form onSubmit={submit} name="form" className="modal-content modal-content--no-inner-scroll">
        <ModalTitle>{t('console-app~Restore as new PVC')}</ModalTitle>
        <ModalBody>
          <p>
            <Trans t={t} ns="console-app">
              When restore action for snapshot <strong>{{ snapshotName }}</strong> is finished a new
              crash-consistent PVC copy will be created.
            </Trans>
          </p>
          <FormGroup
            label={t('console-app~Name')}
            isRequired
            fieldId="pvc-name"
            className="co-restore-pvc-modal__input"
          >
            <TextInput
              isRequired
              type="text"
              id="pvc-name"
              data-test="pvc-name"
              name="restore-pvc-modal__name"
              value={restorePVCName}
              onChange={setPVCName}
            />
          </FormGroup>
          <FormGroup fieldId="restore-storage-class" className="co-restore-pvc-modal__input">
            {!snapshotClassResourceLoaded ? (
              <div className="skeleton-text" />
            ) : (
              <StorageClassDropdown
                onChange={handleStorageClass}
                filter={onlyPvcSCs}
                id="restore-storage-class"
                required
                selectedKey={volumeSnapshotAnnotations?.[snapshotPVCStorageClassAnnotation]}
              />
            )}
          </FormGroup>
          <AccessModeSelector
            onChange={handleAccessMode}
            formClassName="co-restore-pvc-modal__input"
            provisioner={updatedProvisioner}
            loaded={pvcResourceLoaded}
            loadError={pvcResourceLoadError}
            pvcResource={pvcResource}
            availableAccessModes={volumeSnapshotAnnotations?.[
              snapshotPVCAccessModeAnnotation
            ]?.split(',')}
          />
          <FormGroup
            fieldId="pvc-volume-mode"
            className="co-restore-pvc-modal__input"
            label={t('console-app~Volume Mode')}
            isRequired
          >
            {getVolumeModeRadios().map((radio) => (
              <RadioInput
                {...radio}
                key={radio.value}
                onChange={handleVolumeMode}
                inline
                checked={radio.value === volumeMode}
                name="volumeMode"
              />
            ))}
          </FormGroup>
          <FormGroup
            label={t('console-app~Size')}
            isRequired
            fieldId="pvc-size"
            className="co-restore-pvc-modal__input co-restore-pvc-modal__ocs-size"
            helperTextInvalid={t(
              'console-app~Size should be equal or greater than the restore size of snapshot',
            )}
            validated={validSize ? 'default' : 'error'}
          >
            {snapshotClassResourceLoaded ? (
              <RequestSizeInput
                name="requestSize"
                onChange={requestedSizeInputChange}
                defaultRequestSizeUnit={requestedUnit}
                defaultRequestSizeValue={requestedSize}
                dropdownUnits={dropdownUnits}
                isInputDisabled={
                  snapshotClassResourceLoadError || isCephProvisioner(snapshotClassResource?.driver)
                }
                required
              />
            ) : (
              <div className="skeleton-text" />
            )}
          </FormGroup>
          <div className="co-restore-pvc-modal__details-section">
            <p className="text-muted">
              {t('console-app~{{resourceKind}} details', {
                resourceKind: VolumeSnapshotModel.label,
              })}
            </p>
            <Grid hasGutter>
              <GridItem span={6}>
                <div className="co-restore-pvc-modal__pvc-details">
                  <strong>{t('console-app~Created at')}</strong>
                  <span>
                    <Timestamp timestamp={resource?.metadata?.creationTimestamp} />
                  </span>
                </div>
                <div className="co-restore-pvc-modal__pvc-details">
                  <strong>{t('console-app~Status')}</strong>
                  <Status status={resource?.status?.readyToUse ? 'Ready' : 'Not Ready'} />
                </div>
                <div className="co-restore-pvc-modal__pvc-details">
                  <strong>{t('console-app~Size')}</strong>
                  <p>{pvcRequestedSize}</p>
                </div>
              </GridItem>
              <GridItem span={6}>
                <div className="co-restore-pvc-modal__pvc-details">
                  <strong>{t('console-app~Namespace')}</strong>
                  <div>
                    <ResourceIcon kind={NamespaceModel.kind} />
                    <span>{namespace}</span>
                  </div>
                </div>
                <div className="co-restore-pvc-modal__pvc-details">
                  <strong>{t('console-app~API version')}</strong>
                  <p>{resource?.apiVersion}</p>
                </div>
              </GridItem>
            </Grid>
          </div>
        </ModalBody>
        <ModalSubmitFooter
          submitDisabled={!pvcSC || !validSize}
          inProgress={inProgress}
          errorMessage={errorMessage}
          submitText={t('console-app~Restore')}
          cancel={cancel}
        />
      </form>
```

[1] This is another bug resulting from https://github.com/patternfly/patternfly/blob/master/src/patternfly/base/_shield-inheritable.scss.  Another reason to continue work removing PF3 from the code base.

rhamilto:
I opened https://bugzilla.redhat.com/show_bug.cgi?id=1968690 to track the form modal layout issues.

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
rhamilto: Our existing pattern is to wrap `<ModalTitle>`, `<ModalBody>` and `<ModalFooter>` with `<form>`.  Note we have some bugs when trying to use `<Form className="modal-content modal-content--no-inner-scroll"`, so it's best to stick with `<form className="modal-content modal-content--no-inner-scroll"`

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
```diff
@@ -144,9 +137,9 @@ const RestorePVCModal = withHandlePromise<RestorePVCModalProps>(
...
+      <form onSubmit={submit} name="form">
+        <div className="modal-content modal-content--no-inner-scroll">
```
spadgett, line 141: It looks like this `div` should be changed to the `form` element (keeping the classnames) based on @rhamilto's comment.

GowthamShanmugam: mm i thought both are same, anyway let me try without div

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -26,62 +25,65 @@ export const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: stri
...
   const { t } = useTranslation();
-
   const pvcInitialAccessMode = pvcResource
     ? getPVCAccessModes(pvcResource, 'value')
     : availableAccessModes;
 
   const [allowedAccessModes, setAllowedAccessModes] = React.useState<string[]>();
-  const [accessMode, setAccessMode] = React.useState<string>('');
+  const [accessMode, setAccessMode] = React.useState<string>();
+  const volumeMode: string = React.useMemo(() => pvcResource?.spec?.volumeMode, [pvcResource]);
```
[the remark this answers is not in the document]
GowthamShanmugam: ack, try to assign a value directly

frontend/packages/console-app/src/components/modals/restore-pvc/restore-pvc-modal.tsx
rhamilto: It may render the same (I did not test), but the `<div>` is unnecessary, so it's better without.

GowthamShanmugam: ack

GowthamShanmugam:
After Fix:

image

spadgett (approved):
/lgtm

PR #8381, opened 2021-03-15
Bug 1938140: Add 'Roll Back' action for Replication Controllers and Replica Sets

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5594

**Analysis / Root cause**: 
Functionality to allow `Roll Back` replication controllers was dropped from version 4 (from 3.x).

**Solution Description**: 
Add the action `Roll Back` to replication controllers. When initiated display a modal allowing the user to select the roll back options and start the roll back.

image

image

image

image

**Browser conformance**: 
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

FYI @openshift/team-devconsole-ux

/kind bug


andrewballantyne:
/retitle Bug 1938140: Add 'Roll Back' action for Replication Controllers

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react';
+import { Trans, useTranslation } from 'react-i18next';
+import { createModalLauncher, ModalTitle, ModalBody, ModalSubmitFooter } from '../factory/modal';
+import { withHandlePromise } from '../utils';
+import { DeploymentConfigModel } from '../../models';
+import { k8sCreate, k8sUpdate } from '../../module/k8s';
+
+const BaseRollBackRCModal = withHandlePromise((props) => {
+  const { t } = useTranslation();
+  const dcName = props.resource.metadata.ownerReferences[0].name;
```
spadgett, line 10: I'd use optional chaining here just in case and handle the scenario where the DC name is not set.

To be strictly correct, we should find the owner reference to the correct `kind` in case there are multiple owner references. Do have a utility for this logic somewhere used by the topology view?

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+        // update the deployment config based on the one returned by the rollback
+        return k8sUpdate(DeploymentConfigModel, updatedDC);
+      }),
+      props.close,
+    );
+  };
+
+  return (
+    <form onSubmit={submit} name="form" className="modal-content">
+      <ModalTitle>{t('modal~Roll Back')}</ModalTitle>
```
spadgett, line 55: Should "Rollback" be one word? (FWIW, that's what's used in the both the OpenShift and k8s documentation.)

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+    const req = {
+      kind: 'DeploymentConfigRollback',
+      apiVersion: 'apps.openshift.io/v1',
+      name: dcName,
+      latest: true,
+      force: true,
+      spec: {
+        from: {},
+        revision: parseInt(
+          props.resource?.metadata?.annotations?.['openshift.io/deployment-config.latest-version'],
```
spadgett, line 28: We might want to check that this is set as well before performing the action.

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+          </Trans>
+        </p>
+        <div className="checkbox">
+          <label className="control-label">
+            <input
+              type="checkbox"
+              onChange={() => setChangeScaleSettings(!changeScaleSettings)}
+              checked={!!changeScaleSettings}
+            />
+            {t('modal~replica count and selector')}
```
spadgett, line 73: Should these be sentence case?

frontend/public/components/replication-controller.jsx
```diff
@@ -60,7 +60,25 @@ const CancelAction = (kind, obj) => ({
   accessReview: asAccessReview(kind, obj, 'patch'),
 });
 
-const menuActions = [CancelAction, ...replicaSetMenuActions];
+const RollBackAction = (kind, obj) => ({
+  // t('public~Roll Back')
+  labelKey: 'public~Roll Back',
+  // disabled if the DC is not Active
+  isDisabled: _.includes(
```
spadgett, line 67: This should be disabled for replication controllers that aren't owned by a deployment config.

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,107 @@
...
+          </label>
+        </div>
+        <div className="checkbox">
+          <label className="control-label">
+            <input
+              type="checkbox"
+              onChange={() => setChangeStrategy(!changeStrategy)}
+              checked={!!changeStrategy}
```
spadgett, line 81: These values should already be booleans, correct? I don't think the `!!` is necessary.

frontend/public/components/replication-controller.jsx
spadgett: nit: prefer native `Array.includes` to lodash

frontend/public/components/modals/roll-back-rc-modal.jsx
jeff-phillips-18: Was going by the 3.x images. I'll find out for sure.

jeff-phillips-18: We could use `Rollback` but would we still, then, use `when rolling back` ?

frontend/public/components/modals/roll-back-rc-modal.jsx
serenamarie125: don't think so since there's a colon in initial message @bmignano could you verify?

frontend/public/components/modals/roll-back-rc-modal.jsx
jeff-phillips-18: Updated. Also updated screen shots above

frontend/public/components/modals/roll-back-rc-modal.jsx
bmignano: I consulted with Abi on the content side and she recommends sentence case for _everything_. So:

**Roll back**

Use the following settings from **dancer-ex-git-4** when rolling back:
[ ] Replica count and selector
[ ] Deployment strategy
[ ] Deployment trigger

Cancel | Roll back

jeff-phillips-18: Great. @bmignano Can you confirm the screen shots above?

bmignano: In terms of capitalization, looks good! No need to combine **Roll back** into one word though (Modal header and confirmation button), unless that's more accurate from the technical side.

jeff-phillips-18: I believe we decided above that it should be one word based on existing documentation.

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -168,11 +169,14 @@ const getDeploymentRevision = (obj: K8sResourceKind): number => {
...
 };
 
-const getDeploymentConfigVersion = (obj: K8sResourceKind): number => {
+export const getDeploymentConfigVersion = (obj: K8sResourceKind): number => {
   const version = getAnnotation(obj, DEPLOYMENT_CONFIG_LATEST_VERSION_ANNOTATION);
   return version && parseInt(version, 10);
 };
 
+export const getDeploymentConfigNameForRC = (obj: K8sResourceKind): string =>
+  obj.metadata.ownerReferences.find((ref) => ref.kind === DeploymentConfigModel.kind)?.name;
```
spadgett, line 178: I'd use optional chaining here to be safe. `ownerReferences` is not a required field and not all replication controllers are managed by a deployment config.

```suggestion
  obj.metadata.ownerReferences?.find?.((ref) => ref.kind === DeploymentConfigModel.kind)?.name;
```

frontend/public/components/modals/index.ts
```diff
@@ -55,6 +55,11 @@ export const podSelectorModal = (props) =>
     m.podSelectorModal(props),
   );
 
+export const rollBackRCModal = (props) =>
```
spadgett, line 58: nit: I prefer `rollbackRCModal` if "Rollback" is one word. That also matches the case of the API resource `DeploymentConfigRollback`. (Same for some of the other names below.)

frontend/public/components/modals/roll-back-rc-modal.jsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react';
```
spadgett, line 1: `rollback-rc-modal.jsx`?

frontend/public/components/replication-controller.jsx
```diff
@@ -27,14 +26,15 @@ import {
...
 
+const INACTIVE_STATUSES = ['New', 'Pending', 'Running'];
+
 const CancelAction = (kind, obj) => ({
   // t('public~Cancel rollout')
   labelKey: 'public~Cancel rollout',
-  hidden: !_.includes(
-    ['New', 'Pending', 'Running'],
+  hidden: !INACTIVE_STATUSES.includes(
     obj?.metadata?.annotations?.['openshift.io/deployment.phase'],
```
spadgett, line 38: We should make sure this action is hidden if the annotation is missing or the replication controller isn't owned by a deployment config.

frontend/public/components/replication-controller.jsx
```diff
@@ -60,7 +60,24 @@ const CancelAction = (kind, obj) => ({
...
+    obj?.metadata?.annotations?.['openshift.io/deployment.phase'],
+  ),
+  // Hidden if RC is active or does not belong to a deployment config
+  hidden: obj?.status?.replicas > 0 || !getDeploymentConfigNameForRC(obj),
+  callback: () =>
+    rollBackRCModal({
+      resourceKind: kind,
+      resource: obj,
+    }),
+  accessReview: asAccessReview(kind, obj, 'patch'),
```
spadgett, line 77: Sorry, I missed this on last review, but the RBAC check looks wrong here. We need to be able to update the deployment config rather than patching the replication controller.

frontend/public/components/replication-controller.jsx
jeff-phillips-18: @spadgett Do you want this change for `Cancel Rollout` or for `Rollback` (or both)?

spadgett: @jeff-phillips-18 You can use the owner references to see if it's owned by a deployment config or a standalone replication controller.

The Rollback action might be OK since we're already checking the deployment phase annotation (which should only be there if it's owned), but it's probably better to check both places.

jeff-phillips-18: @spadgett I've added the check for `deployment.phase` annotation to both the cancel and rollback actions.

frontend/public/components/replication-controller.jsx
```diff
@@ -27,40 +26,67 @@ import {
...
+            },
+          ]),
+      }),
+    accessReview: asAccessReview(kind, obj, 'patch'),
+  };
+};
+
+const RollbackAction = (kind, obj) => {
+  const deploymentPhase = obj?.metadata?.annotations?.['openshift.io/deployment.phase'];
+  const dcObj = {
```
spadgett, line 67: I wouldn't create a fake object... You can specify the access review yourself instead of using the `asAccessReview` helper.

```ts
accessReview: {
  group: DeploymentConfigModel.apiGroup,
  resource: DeploymentConfigModel.plural,
  name: getDeploymentConfigNameForRC(obj),
  namespace: obj?.metadata?.namespace,
}
```

frontend/public/components/replication-controller.jsx
```diff
@@ -27,40 +26,67 @@ import {
...
+      rollbackRCModal({
+        resourceKind: kind,
+        resource: obj,
+      }),
+    accessReview: {
+      group: DeploymentConfigModel.apiGroup,
+      resource: DeploymentConfigModel.plural,
+      name: dcName,
+      namespace: obj?.metadata?.namespace,
+    },
```
spadgett, line 85: Sorry, I forgot to specify the verb in my example, which is needed :(

```suggestion
    accessReview: {
      group: DeploymentConfigModel.apiGroup,
      resource: DeploymentConfigModel.plural,
      name: dcName,
      namespace: obj?.metadata?.namespace,
      verb: 'update',
    },
```

jeff-phillips-18: thanks! fixed.

spadgett (approved):
/lgtm

yanpzhan:
@jeff-phillips-18 After tested against the pr, I have a doubt about this scenario: when create a dc which will fail fore sure due to wrong image, all the rcs are "Failed", now there are "Roll Back" items in all rc's action list. Is this normal?

jeff-phillips-18:
> @jeff-phillips-18 After tested against the pr, I have a doubt about this scenario: when create a dc which will fail fore sure due to wrong image, all the rcs are "Failed", now there are enabled "Roll Back" items in all rc's action list. Is this normal?

I suppose it would depend on the reason for failure. I was able to rollback to a failed item that failed due to another rollback being created when that build was still running.

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -160,21 +162,27 @@ const sortByRevision = (
...
 
+export const getDeploymentConfigNameForRC = (obj: K8sResourceKind): string =>
+  obj?.metadata?.ownerReferences?.find((ref) => ref.kind === DeploymentConfigModel.kind)?.name;
+
+export const getDeploymentNameForRS = (obj: K8sResourceKind): string =>
+  obj?.metadata?.ownerReferences?.find((ref) => ref.kind === DeploymentModel.kind)?.name;
+
 const getDeploymentConfigName = (obj: K8sResourceKind): string => {
-  return _.get(obj, 'metadata.ownerReferences[0].name', null);
+  return obj?.metadata?.ownerReferences?.[0]?.name;
```
spadgett, line 185: It looks like this might be the same as `getDeploymentConfigNameForRC`? (Except it assumes the first owner reference is the DC)

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -160,21 +162,27 @@ const sortByRevision = (
 };
 
 const getAnnotation = (obj: K8sResourceKind, annotation: string): string => {
```
spadgett, line 164: nit: Not new, but these utilities should probably use `K8sResourceCommon` which has more strict types.

```suggestion
const getAnnotation = (obj: K8sResourceCommon, annotation: string): string => {
```

frontend/public/components/modals/rollback-modal.jsx
```diff
@@ -0,0 +1,215 @@
...
+  const { t } = useTranslation();
+  const isDCRollback = props.resource.kind === ReplicationControllerModel.kind;
+  const dName = isDCRollback
+    ? getDeploymentConfigNameForRC(props.resource)
+    : getDeploymentNameForRS(props.resource);
+  const [changeScaleSettings, setChangeScaleSettings] = React.useState(false);
+  const [changeStrategy, setChangeStrategy] = React.useState(false);
+  const [changeTriggers, setChangeTriggers] = React.useState(false);
+  const deploymentResource = {
+    kind: isDCRollback ? DeploymentConfigModel.kind : DeploymentModel.kind,
```
spadgett, line 35: Do we need an API group here? I'm a little surprised this works.

frontend/public/components/replicaset.jsx
```diff
@@ -25,11 +25,68 @@ import {
...
+      rollbackModal({
+        resourceKind: kind,
+        resource: obj,
+      }),
+    accessReview: {
+      group: DeploymentModel.apiGroup,
+      resource: DeploymentModel.plural,
+      name: deploymentName,
+      namespace: obj?.metadata?.namespace,
+      verb: 'update',
```
spadgett, line 83: Deployment rollback should check the `patch` verb since we use `k8sPatch`

```suggestion
      verb: 'patch',
```

frontend/public/components/modals/rollback-modal.jsx
jeff-phillips-18: It really doesn't like when I use `referenceForModel(DeploymentModel)` or `referenceForModel(DeploymentConfigModel)`

spadgett: I think there's probably an existing bug here somewhere :/ I'd expect that to work, and we should really be using the API group for anything that has one. But we can look at that separately from this PR.

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -159,23 +161,22 @@ const sortByRevision = (
...
+export const getDeploymentConfigVersion = (obj: K8sResourceCommon): number => {
   const version = getAnnotation(obj, DEPLOYMENT_CONFIG_LATEST_VERSION_ANNOTATION);
   return version && parseInt(version, 10);
 };
 
-const getDeploymentConfigName = (obj: K8sResourceKind): string => {
-  return _.get(obj, 'metadata.ownerReferences[0].name', null);
-};
+export const getOwnerNameByKind = (obj: K8sResourceCommon, kind: string): string =>
+  obj?.metadata?.ownerReferences?.find((ref) => ref.kind === kind)?.name;
```
spadgett, line 179: If we're making this a common utility, we should really require the API group, not just the kind. The kind on its own doesn't identify a resource.

frontend/packages/console-shared/src/utils/resource-utils.ts
```diff
@@ -159,22 +163,33 @@ const sortByRevision = (
...
+  if (!reference) {
+    return null;
+  }
+  const groupVersionKind = getGroupVersionKind(reference);
+  if (!groupVersionKind) {
+    return null;
+  }
+  const apiVersion = `${groupVersionKind[0]}/${groupVersionKind[1]}`;
+  return obj?.metadata?.ownerReferences?.find(
+    (ref) => ref.kind === kind.kind && ref.apiVersion === apiVersion,
```
spadgett, line 191: I think we should ignore the actual version if possible since the same resource can have multiple versions. Maybe something like this (not tested)

```suggestion
    (ref) => ref.kind === kind.kind && ((!kind.apiGroup && ref.apiVersion === 'v1') || ref.apiVersion?.startsWith(`${kind.apiGroup}/`)),
```

spadgett (approved):
/lgtm

serenamarie125 (approved):
Looks great!

yanpzhan:
/label qe-approved
QE have tested "Rollback" action for both Deployments and Deployment Configs, features work well as designed, no issue is found.

PR #8383, opened 2021-03-15
(bug-fix): Fixes donut chart loading

afreen23:
-  fixes available disk info
 -  used node names for request on localvolumediscoveryresults

Before:
before

After:
after


bipuladh (approved):
/lgtm

PR #8384, opened 2021-03-15
Bug 1905627: Warn using an unsupported IE browser

jhadvig:
/assign @rhamilto

### DIFF
--- a/pkg/serverutils/utils.go
+++ b/pkg/serverutils/utils.go
@@ -3,6 +3,7 @@ package serverutils
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"k8s.io/klog"
 )
@@ -23,6 +24,32 @@ func SendResponse(rw http.ResponseWriter, code int, resp interface{}) {
 	if err != nil {
 		klog.Errorf("Failed sending HTTP response body: %v", err)
 	}
+}
+
+func IsUnsupportedBrowser(r *http.Request) bool {
+	userAgentHeader := r.Header.Get("User-Agent")
+	isUnsupported := false
+	unsupportedBrowserHeadersIdentifier := []string{
+		"Trident/", // IE 11
+		"MSIE ",    // IE 10 or older
+	}
+	for _, identifier := range unsupportedBrowserHeadersIdentifier {
+		if strings.Contains(userAgentHeader, identifier) {
+			isUnsupported = true
+			break
+		}
+	}
+	return isUnsupported
+}
+
+func SendUnsupportedBrowserResponse(rw http.ResponseWriter, brand string) {
+	content := `<p>Internet Explorer 11 and earlier are not supported, and the ` + brand + ` console will not function with your current browser.
+Please use a supported browser such as the latest version of Google Chrome, Mozilla Firefox, Microsoft Edge, or Safari.
+The current Firefox ESR (Extended Support Release) is also supported. To see the browsers we test, refer to the
+<a href="https://access.redhat.com/articles/4763741">OpenShift Container Platform 4.x Tested Integrations (for x86_x64)</a>.</p>`
+	rw.Header().Set("Content-Type", "text/html")
+	rw.WriteHeader(http.StatusFailedDependency)
+	rw.Write([]byte(content))
 }
 
 type ApiError struct {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -490,6 +490,11 @@ func (s *Server) handleKnativeChannelCRDs(w http.ResponseWriter, r *http.Request
 }
 
 func (s *Server) indexHandler(w http.ResponseWriter, r *http.Request) {
+	if serverutils.IsUnsupportedBrowser(r) {
+		serverutils.SendUnsupportedBrowserResponse(w, s.Branding)
+		return
+	}
+
 	jsg := &jsGlobals{
 		ConsoleVersion:        version.Version,
 		AuthDisabled:          s.authDisabled(),


frontend/public/components/app.jsx
```diff
@@ -143,7 +144,22 @@ class App_ extends React.PureComponent {
     }
   }
 
+  _isIE() {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf('MSIE '); // IE 10 or older
+    const trident = ua.indexOf('Trident/'); //IE 11
```
rhamilto, line 150: Nit
```suggestion
    const trident = ua.indexOf('Trident/'); // IE 11
```

rhamilto (commented):
Just a nit.  Otherwise LGTM.

rhamilto (approved):
/lgtm

TheRealJon (approved):

rhamilto:
/hold 

as this does not appear to be rendering in IE

zherman0 (approved):

jhadvig:
@spadgett @rhamilto I've found out why the change was not working.  There are several issues, one is the IE it self and it's inability to just work. For that reason we need to check for the `User-Agent` header on the backend level, before parsing the `index.html` GO template. There we have two options how to handle unsupported browser:
1. render a black page that with the msg that the browser is unsupported and return prematurely, which will result into skip setting `jsGlobals` and parsing the `index.html` with them.
2. extend the `index.html` with case of unsupported browser statement

thoughts ?

rhamilto:
> @spadgett @rhamilto I've found out why the change was not working. There are several issues, one is the IE it self and it's inability to just work. For that reason we need to check for the `User-Agent` header on the backend level, before parsing the `index.html` GO template. There we have two options how to handle unsupported browser:
> 
> 1. render a black page that with the msg that the browser is unsupported and return prematurely, which will result into skip setting `jsGlobals` and parsing the `index.html` with them.

`black` should be `blank`?

> 2. extend the `index.html` with case of unsupported browser statement
> 
> thoughts ?

It seems like option 1 is simpler, and given very few users should every hit this, prefer it.

jhadvig:
> black should be blank?

yes 😄

pkg/serverutils/utils.go
```diff
@@ -25,6 +26,28 @@ func SendResponse(rw http.ResponseWriter, code int, resp interface{}) {
...
+			break
+		}
+	}
+	return isUnsupported
+}
+
+func SendUnsupportedBrowserResponse(rw http.ResponseWriter) {
+	rw.Header().Set("Content-Type", "text/html")
+	rw.WriteHeader(http.StatusFailedDependency)
+	rw.Write([]byte("<div>Unsupported browser detected. Please use a supported browser such as Google Chrome, Mozilla Firefox, or Microsoft Edge.</div>"))
```
spadgett, line 48: Safari is also supported. I would add some more detail here with a link to the tested integrations. Maybe

```html
<p>Internet Explorer 11 and earlier are not supported, and the {brand} console will not function with your current browser.
Please use a supported browser such as the latest version of Google Chrome, Mozilla Firefox, Microsoft Edge, or Safari.
The current Firefox ESR (Extended Support Release) is also supported. To see the browsers we test, refer to the
<a href="https://access.redhat.com/articles/4763741">OpenShift Container Platform 4.x Tested Integrations (for x86_x64)</a>.
```

pkg/serverutils/utils.go
```diff
@@ -25,6 +26,28 @@ func SendResponse(rw http.ResponseWriter, code int, resp interface{}) {
 	}
 }
 
+func CheckUnsupportedBrowser(r *http.Request) bool {
```
spadgett, line 29: I would prefix it with `Is` to make it clear it's a function that returns a boolean.

```suggestion
func IsUnsupportedBrowser(r *http.Request) bool {
```

pkg/serverutils/utils.go
spadgett: It might be good to link to the download pages for the various browsers as well.

rhamilto: I'm not sure I'd mention Safari given it is no longer available for Windows (which anyone seeing this message will be using since they're using IE11).

I favor the message as it is for its brevity and maintainability (there is little that can become outdated).  Very few users should be seeing this message.  When they do, the important part  to communicate is the browser they are using is not supported.  As has been discussed in Slack, we need better documentation on supported browsers, but putting more of that here seems like overkill.  Maybe we link to that doc when it exists?

jhadvig:
@spadgett comments addressed.

>Maybe we link to that doc when it exists?

@rhamilto would definitely make sense 👍

pkg/serverutils/utils.go
```diff
@@ -25,6 +26,32 @@ func SendResponse(rw http.ResponseWriter, code int, resp interface{}) {
...
+		}
+	}
+	return isUnsupported
+}
+
+func SendUnsupportedBrowserResponse(rw http.ResponseWriter, brand string) {
+	content := `<p>Internet Explorer 11 and earlier are not supported, and the ` + brand + ` console will not function with your current browser.
+Please use a supported browser such as the latest version of Google Chrome, Mozilla Firefox, Microsoft Edge, or Safari.
+The current Firefox ESR (Extended Support Release) is also supported. To see the browsers we test, refer to the
+<a href="https://access.redhat.com/articles/4763741">OpenShift Container Platform 4.x Tested Integrations (for x86_x64)</a>.</p>`
```
spadgett, line 49: This looks good to me, although I know @rhamilto had concerns about being so specific in case what we support changes. I'm OK either way. @rhamilto ?

spadgett (commented):
Thanks, @jhadvig. Were you able to determine why this failed QE validation?

https://bugzilla.redhat.com/show_bug.cgi?id=1905627#c5

pkg/serverutils/utils.go
rhamilto: I'm good with the consensus, whatever that may be.

jhadvig: right, but your formulation is reasonable and make sense to me, so went with that.
@rhamilto thoughts?

jhadvig:
@spadgett I think that there might be testing the previous change.
Xiyun Zhao
@XiyunZhao could you please retest the change, since on Friday I pushed new change. Thanks :)

spadgett (approved):
/lgtm

Leaving the hold until QE can verify (thanks!)

pkg/serverutils/utils.go
rhamilto: My take is less is more.  We know users don't read every word, and brevity is important when trying to relay something important.  So I'm a fan of significantly less text here, but am happy to roll with the majority opinion.

spadgett:
/hold cancel

QE has verified the fix: https://bugzilla.redhat.com/show_bug.cgi?id=1905627#c6

jhadvig:
@spadgett do we wanna backport this to 4.7 since the the BZ is set to that Version?

PR #8385, opened 2021-03-15
add support for finally tasks in pipeline builder

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-4313
**Analysis / Root cause**: 

As a user, I'd like to be able to compose finally tasks in the Pipeline Builder.

**Solution Description**: 

Add support to finally section in the pipeline builder.

finally tasks in pipeline builder:
image

Finally tasks while creating pipeline:
create_pipeline

Editing pipeline
editing_pipeline

```
 packages/pipelines-plugin/src/components/pipelines/pipeline-topology/__tests__/utils.spec.ts

  getTopologyNodesEdges
    ✓ expect to return nodes and edges for a pipeline without finally tasks (25ms)
    ✓ expect to return nodes and edges for a pipeline with finally tasks (2ms)
  getLastRegularTasks
    ✓ expect to handle the empty array input
    ✓ expect to return the list of task names that will be executed at the end (3ms)
  getFinallyTaskHeight
    ✓ expect to return dynamic height for finally task based on tasks length and builder options  (1ms)
```

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @bgliwa01 @andrewballantyne @nemesis09

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
@@ -83,14 +83,21 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
     [setStatus],
   );
 
-  const onTaskSelection = (task: PipelineVisualizationTaskItem, resource: TaskKind) => {
+  const onTaskSelection = (
+    task: PipelineVisualizationTaskItem,
+    resource: TaskKind,
+    isFinallyTask: boolean,
+  ) => {
+    const builderNodes = isFinallyTask ? values.formData.finallyTasks : values.formData.tasks;
     setSelectedTask({
-      taskIndex: values.formData.tasks.findIndex(({ name }) => name === task.name),
+      isFinallyTask,
+      taskIndex: builderNodes.findIndex(({ name }) => name === task.name),
       resource,
     });
   };
 
   useResourceValidation(
+    values.formData.finallyTasks,
     values.formData.tasks,
     values.formData.resources,
     values.formData.workspaces,
@@ -98,20 +105,25 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
   );
 
   const updateTasks = (changes: CleanupResults): void => {
-    const { tasks, listTasks, errors: taskErrors } = changes;
+    const { tasks, listTasks, finallyTasks, finallyListTasks, errors: taskErrors } = changes;
 
     setFieldValue('formData.tasks', tasks);
     setFieldValue('formData.listTasks', listTasks);
+    setFieldValue('formData.finallyTasks', finallyTasks);
+    setFieldValue('formData.finallyListTasks', finallyListTasks);
     updateErrors(taskErrors);
   };
 
-  const selectedId = values.formData.tasks[selectedTask?.taskIndex]?.name;
+  const nodeType = selectedTask?.isFinallyTask ? 'finallyTasks' : 'tasks';
+  const selectedId = values.formData[nodeType][selectedTask?.taskIndex]?.name;
   const selectedIds = selectedId ? [selectedId] : [];
 
   const taskGroup: PipelineBuilderTaskGroup = {
     tasks: values.formData.tasks,
     listTasks: values.formData.listTasks,
     highlightedIds: selectedIds,
+    finallyTasks: values.formData.finallyTasks,
+    finallyListTasks: values.formData.finallyListTasks,
   };
 
   const closeSidebarAndHandleReset = React.useCallback(() => {
@@ -141,6 +153,8 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
       ...newFormData.spec,
       name: newFormData.metadata?.name,
       listTasks: values.formData.listTasks,
+      finallyTasks: newFormData.spec.finally,
+      finallyListTasks: values.formData.finallyListTasks,
     };
     return _.merge({}, initialPipelineFormData, formData);
   };
@@ -230,6 +244,7 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
                   t,
                 );
               }}
+              isFinallyTask={selectedTask.isFinallyTask}
               selectedPipelineTaskIndex={selectedTask.taskIndex}
               taskResource={selectedTask.resource}
             />
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import { useTranslation } from 'react-i18next';
+import { getRandomChars } from '@console/shared';
 import { referenceForModel } from '@console/internal/module/k8s';
 import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
 import { ClusterTaskModel, TaskModel } from '../../../models';
@@ -15,12 +16,16 @@ import { PipelineVisualizationTaskItem } from '../../../utils/pipeline-utils';
 import { AddNodeDirection } from '../pipeline-topology/const';
 import {
   PipelineBuilderTaskNodeModel,
+  PipelineBuilderFinallyNodeModel,
   PipelineMixedNodeModel,
   PipelineTaskListNodeModel,
 } from '../pipeline-topology/types';
 import {
+  createBuilderFinallyNode,
   createInvalidTaskListNode,
   createTaskListNode,
+  getFinallyTaskHeight,
+  getLastRegularTasks,
   handleParallelToParallelNodes,
   tasksToBuilderNodes,
 } from '../pipeline-topology/utils';
@@ -30,6 +35,7 @@ import {
   TaskErrorMap,
   UpdateErrors,
   UpdateOperationAddData,
+  UpdateOperationConvertToFinallyTaskData,
   UpdateOperationConvertToTaskData,
   UpdateOperationFixInvalidTaskListData,
   UpdateTasksCallback,
@@ -77,6 +83,75 @@ export const useTasks = (namespace?: string): UseTasks => {
   };
 };
 
+const useConnectFinally = (
+  namespace,
+  nodes,
+  taskGroup: PipelineBuilderTaskGroup,
+  onTaskSelection: SelectTaskCallback,
+  onUpdateTasks: UpdateTasksCallback,
+  tasksInError: TaskErrorMap,
+): PipelineMixedNodeModel => {
+  const { clusterTasks, namespacedTasks } = useTasks(namespace);
+  const taskGroupRef = React.useRef(taskGroup);
+  taskGroupRef.current = taskGroup;
+  const addNewFinallyListNode = () => {
+    const data: UpdateOperationConvertToFinallyTaskData = {
+      listTaskName: `finally-list-${getRandomChars(6)}`,
+    };
+    onUpdateTasks(taskGroupRef.current, { type: UpdateOperationType.ADD_FINALLY_LIST_TASK, data });
+  };
+  // TODO: Cleanup in ODC-3165
+  const getTask = (taskRef: PipelineTaskRef) => {
+    if (taskRef?.kind === ClusterTaskModel.kind) {
+      return clusterTasks?.find((task) => task.metadata.name === taskRef?.name);
+    }
+    return namespacedTasks?.find((task) => task.metadata.name === taskRef?.name);
+  };
+
+  const convertListToFinallyTask = (resource: TaskKind, name: string) => {
+    const data: UpdateOperationConvertToTaskData = { resource, name };
+    onUpdateTasks(taskGroupRef.current, {
+      type: UpdateOperationType.CONVERT_LIST_TO_FINALLY_TASK,
+      data,
+    });
+  };
+  const allTasksLength = taskGroup.finallyTasks.length + taskGroup.finallyListTasks.length;
+  const finallyNodeName = `finally-node-${taskGroup.finallyTasks.length}-${taskGroup.finallyListTasks.length}`;
+  const regularRunAfters = getLastRegularTasks(nodes);
+
+  const finallyGroupNode: PipelineBuilderFinallyNodeModel = createBuilderFinallyNode(
+    getFinallyTaskHeight(allTasksLength, false),
+  )(finallyNodeName, {
+    isFinallyTask: true,
+    namespace,
+    namespaceTaskList: namespacedTasks,
+    clusterTaskList: clusterTasks,
+    task: {
+      isFinallyTask: true,
+      name: finallyNodeName,
+      runAfter: regularRunAfters,
+      addNewFinallyListNode,
+      finallyTasks: taskGroup.finallyTasks.map((ft) => ({
+        ...ft,
+        onTaskSelection: () => onTaskSelection(ft, getTask(ft.taskRef), true),
+        error: getErrorMessage(nodeTaskErrors, tasksInError)(ft.name),
+        selected: taskGroup.highlightedIds.includes(ft.name),
+        disableTooltip: true,
+      })),
+      finallyListTasks: taskGroup.finallyListTasks.map((flt) => ({
+        ...flt,
+        convertList: (resource: TaskKind) => convertListToFinallyTask(resource, flt.name),
+        onRemoveTask: () => {
+          onUpdateTasks(taskGroupRef.current, {
+            type: UpdateOperationType.DELETE_LIST_TASK,
+            data: { listTaskName: flt.name },
+          });
+        },
+      })),
+    },
+  });
+  return finallyGroupNode;
+};
 type UseNodes = {
   nodes: PipelineMixedNodeModel[];
   tasksCount: number;
@@ -175,7 +250,7 @@ export const useNodes = (
       ? tasksToBuilderNodes(
           validTaskList,
           onNewListNode,
-          (task) => onTaskSelection(task, getTask(task.taskRef)),
+          (task) => onTaskSelection(task, getTask(task.taskRef), false),
           getErrorMessage(nodeTaskErrors, tasksInError),
           taskGroup.highlightedIds,
         )
@@ -193,16 +268,24 @@ export const useNodes = (
 
   const localTaskCount = namespacedTasks?.length || 0;
   const clusterTaskCount = clusterTasks?.length || 0;
-
+  const finallyNode = useConnectFinally(
+    namespace,
+    nodes,
+    taskGroup,
+    onTaskSelection,
+    onUpdateTasks,
+    tasksInError,
+  );
   return {
     tasksCount: localTaskCount + clusterTaskCount,
     tasksLoaded: !!namespacedTasks && !!clusterTasks,
     loadingTasksError: errorMsg,
-    nodes,
+    nodes: [...nodes, finallyNode],
   };
 };
 
 export const useResourceValidation = (
+  finallyTasks: PipelineTask[],
   tasks: PipelineTask[],
   resourceValues: TektonResource[],
   workspaceValues: PipelineWorkspace[],
@@ -213,7 +296,7 @@ export const useResourceValidation = (
   React.useEffect(() => {
     const resourceNames = resourceValues.map((r) => r.name);
 
-    const errors = tasks.reduce((acc, task) => {
+    const errors = [...tasks, ...finallyTasks].reduce((acc, task) => {
       const output = task.resources?.outputs || [];
       const input = task.resources?.inputs || [];
       const missingResources = [...output, ...input].filter(
@@ -262,5 +345,13 @@ export const useResourceValidation = (
       }
       onError(outputErrors);
     }
-  }, [tasks, resourceValues, workspaceValues, onError, previousErrorIds, setPreviousErrorIds]);
+  }, [
+    tasks,
+    resourceValues,
+    workspaceValues,
+    onError,
+    previousErrorIds,
+    setPreviousErrorIds,
+    finallyTasks,
+  ]);
 };
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebar.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebar.tsx
@@ -29,6 +29,7 @@ type TaskSidebarProps = {
   workspaceList: PipelineWorkspace[];
   selectedPipelineTaskIndex: number;
   taskResource: TaskKind;
+  isFinallyTask: boolean;
   onClose: () => void;
 };
 
@@ -40,10 +41,12 @@ const TaskSidebar: React.FC<TaskSidebarProps> = (props) => {
     workspaceList,
     selectedPipelineTaskIndex,
     taskResource,
+    isFinallyTask,
     onClose,
   } = props;
   const { t } = useTranslation();
-  const formikTaskReference = `formData.tasks.${selectedPipelineTaskIndex}`;
+  const taskType = isFinallyTask ? 'finallyTasks' : 'tasks';
+  const formikTaskReference = `formData.${taskType}.${selectedPipelineTaskIndex}`;
   const [taskField] = useField<PipelineTask>(formikTaskReference);
 
   const updateTask = (newData: Partial<UpdateOperationUpdateTaskData>) => {
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/types.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/types.ts
@@ -21,6 +21,8 @@ export type PipelineBuilderListTask = PipelineBuilderTaskBase;
 export type PipelineBuilderTaskGrouping = {
   tasks: PipelineTask[];
   listTasks: PipelineBuilderListTask[];
+  finallyTasks: PipelineTask[];
+  finallyListTasks: PipelineBuilderListTask[];
 };
 
 export type PipelineBuilderTaskGroup = PipelineBuilderTaskGrouping & {
@@ -45,6 +47,7 @@ export type PipelineBuilderFormikValues = FormikValues & PipelineBuilderFormYaml
 export type SelectedBuilderTask = {
   resource: TaskKind;
   taskIndex: number;
+  isFinallyTask: boolean;
 };
 
 export type TaskErrorMap = {
@@ -54,6 +57,7 @@ export type TaskErrorMap = {
 export type SelectTaskCallback = (
   task: PipelineVisualizationTaskItem,
   taskResource: TaskKind,
+  isFinallyTask: boolean,
 ) => void;
 
 export type UpdateOperation<D extends UpdateOperationBaseData = UpdateOperationBaseData> = {
@@ -76,6 +80,9 @@ export type UpdateOperationConvertToTaskData = UpdateOperationBaseData & {
   name: string;
   resource: TaskKind;
   runAfter?: string[];
+};
+export type UpdateOperationConvertToFinallyTaskData = {
+  listTaskName: string;
 };
 export type UpdateOperationFixInvalidTaskListData = UpdateOperationBaseData & {
   existingName: string;
@@ -118,11 +125,12 @@ export type UpdateOperationUpdateTaskData = UpdateOperationBaseData & {
 export type CleanupResults = {
   tasks: PipelineTask[];
   listTasks: PipelineBuilderListTask[];
+  finallyTasks: PipelineTask[];
+  finallyListTasks: PipelineBuilderListTask[];
   errors?: TaskErrorMap;
 };
 
 export type UpdateOperationAction<D extends UpdateOperationBaseData> = (
-  tasks: PipelineTask[],
-  listTasks: PipelineBuilderListTask[],
+  taskGrouping: PipelineBuilderTaskGrouping,
   data: D,
 ) => CleanupResults;
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
@@ -9,10 +9,12 @@ import {
   PipelineBuilderListTask,
   PipelineBuilderTaskBase,
   PipelineBuilderTaskGroup,
+  PipelineBuilderTaskGrouping,
   TaskErrorMap,
   UpdateOperation,
   UpdateOperationAction,
   UpdateOperationAddData,
+  UpdateOperationConvertToFinallyTaskData,
   UpdateOperationConvertToTaskData,
   UpdateOperationDeleteListTaskData,
   UpdateOperationFixInvalidTaskListData,
@@ -153,15 +155,16 @@ const getErrors = (task: PipelineTask, resource: TaskKind): TaskErrorMap => {
   return { [task.name]: errorListing.length > 0 ? errorListing : null };
 };
 
-const addListNode: UpdateOperationAction<UpdateOperationAddData> = (tasks, listTasks, data) => {
+const addListNode: UpdateOperationAction<UpdateOperationAddData> = (taskGrouping, data) => {
   const { direction, relatedTask } = data;
 
   const newTaskName = `${direction}-${getRandomChars(6)}`;
   const relatedTaskName = relatedTask.name;
-
+  const { tasks, listTasks } = taskGrouping;
   switch (direction) {
     case AddNodeDirection.BEFORE:
       return {
+        ...taskGrouping,
         tasks: tasks.map((pipelineTask) =>
           mapBeRelated<PipelineTask>(newTaskName, relatedTaskName, pipelineTask),
         ),
@@ -174,6 +177,7 @@ const addListNode: UpdateOperationAction<UpdateOperationAddData> = (tasks, listT
       };
     case AddNodeDirection.AFTER:
       return {
+        ...taskGrouping,
         tasks: tasks.map((pipelineTask) =>
           mapReplaceRelatedInOthers<PipelineTask>(newTaskName, relatedTaskName, pipelineTask),
         ),
@@ -190,6 +194,7 @@ const addListNode: UpdateOperationAction<UpdateOperationAddData> = (tasks, listT
       };
     case AddNodeDirection.PARALLEL:
       return {
+        ...taskGrouping,
         tasks: tasks.map((pipelineTask) =>
           mapAddRelatedToOthers<PipelineTask>(newTaskName, relatedTaskName, pipelineTask),
         ),
@@ -208,15 +213,15 @@ const addListNode: UpdateOperationAction<UpdateOperationAddData> = (tasks, listT
 const getTaskNames = (tasks: PipelineTask[]) => tasks.map((t) => t.name);
 
 const convertListToTask: UpdateOperationAction<UpdateOperationConvertToTaskData> = (
-  tasks,
-  listTasks,
+  taskGrouping,
   data,
 ) => {
   const { name, resource, runAfter } = data;
-  const usedNames = getTaskNames(tasks);
+  const { tasks, listTasks, finallyTasks } = taskGrouping;
+  const usedNames = getTaskNames([...tasks, ...finallyTasks]);
   const newPipelineTask: PipelineTask = convertResourceToTask(usedNames, resource, runAfter);
-
   return {
+    ...taskGrouping,
     tasks: [
       ...tasks.map((pipelineTask) =>
         mapReplaceRelatedInOthers(newPipelineTask.name, name, pipelineTask),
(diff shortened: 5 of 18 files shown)


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -223,14 +229,87 @@ export const getEdgesFromNodes = (nodes: PipelineMixedNodeModel[]): PipelineEdge
...
+    } = regularTask;
+    if (runAfter && runAfter.length) {
+      runAfter.forEach((ra) => {
+        if (!acc.includes(ra)) {
+          acc.push(ra);
+        }
+      });
+    }
+    return acc;
+  }, []);
```
andrewballantyne, line 258: You could just hold onto them in an object as keys and it'll distinct it for you. 
```suggestion
     acc[runAfter] = true;
     ...
  }, {});
  Object.keys(...)
```

Alternatively Lodash has a `uniq` utility: https://lodash.com/docs/3.10.1#uniq


Line 251's `.forEach` should be a `.filter` if you choose to keep it the current design.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -223,14 +229,87 @@ export const getEdgesFromNodes = (nodes: PipelineMixedNodeModel[]): PipelineEdge
...
     getPipelineTasks(pipeline, pipelineRun),
   );
-  const nodes: PipelineMixedNodeModel[] = tasksToNodes(taskList, pipeline, pipelineRun);
+  const taskNodes: PipelineMixedNodeModel[] = tasksToNodes(taskList, pipeline, pipelineRun);
+
+  const finallyTaskNodes: PipelineMixedNodeModel[] = finallyTasksToNodes(
+    taskNodes,
+    pipeline,
+    pipelineRun,
+  );
```
andrewballantyne, line 310: Curious, why did you over-reach and do [ODC-4312](https://issues.redhat.com/browse/ODC-4312)? @nemesis09 FYI.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -223,14 +229,87 @@ export const getEdgesFromNodes = (nodes: PipelineMixedNodeModel[]): PipelineEdge
...
+
+  const nodes = [...taskNodes, ...finallyTaskNodes];
```
andrewballantyne, line 312: Also you could have just done a singular node and placed it as the last item... you seemed to have wrapped it in an array for some reason I'm not seeing...

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebar.tsx
```diff
@@ -40,10 +41,12 @@ const TaskSidebar: React.FC<TaskSidebarProps> = (props) => {
     workspaceList,
     selectedPipelineTaskIndex,
     taskResource,
+    isFinallyTask,
     onClose,
   } = props;
   const { t } = useTranslation();
-  const formikTaskReference = `formData.tasks.${selectedPipelineTaskIndex}`;
+  const builderTask = isFinallyTask ? 'finallyTasks' : 'tasks';
```
andrewballantyne, line 48:
```suggestion
  const taskType = isFinallyTask ? 'finallyTasks' : 'tasks';
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -439,21 +539,74 @@ export const applyChange = (
...
 
   switch (type) {
     case UpdateOperationType.ADD_LIST_TASK:
-      return addListNode(tasks, listTasks, data as UpdateOperationAddData);
+      return addListNode(
+        tasks,
+        listTasks,
+        data as UpdateOperationAddData,
+        finallyTasks,
+        finallyListTasks,
```
andrewballantyne, line 551: Hmmm... concerned that we ended up adding a bunch of properties that are just pass-thru... 

Maybe wrap the return statements that don't use a specific set of data?

Something like:
```
return withUnusedData(addListNode(tasks, listTasks, data as UpdateOperationAddData))
```
Where `withUnusedData` simply just adds it onto the end of the data:
```
const withUnusedData = (updatedValues) => ({ tasks, listTasks, finallyTasks, finallyListTasks, ...updatedValues })
// switch statement
```

Ideally this would allow us to only modify what we need (and the following comment will allow all the data to still be available)... power of spread will give us a nice clean fallback without stepping on anything / being redundant.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -439,21 +539,74 @@ export const applyChange = (
   op: UpdateOperation,
 ): CleanupResults => {
   const { type, data } = op;
-  const { tasks, listTasks } = taskGroup;
+  const { tasks, listTasks, finallyTasks, finallyListTasks } = taskGroup;
 
   switch (type) {
     case UpdateOperationType.ADD_LIST_TASK:
-      return addListNode(tasks, listTasks, data as UpdateOperationAddData);
+      return addListNode(
```
andrewballantyne, line 546: Maybe we also just pass in `taskGroup` at this point. I know it has the selectedIds but that's immaterial I think. Having two data to each method makes it easier than managing multiple separate lists that all came from the same object; it's getting a bit unruly. 

ie `return methodName(taskGroup, data as Whatever)`

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -360,42 +430,49 @@ const updateTask: UpdateOperationAction<UpdateOperationUpdateTaskData> = (
...
-
-  const updatedResourceIndex = tasks.findIndex(
+  const allTasks = [...tasks, ...finallyTasks];
+  const updatedResourceIndex = allTasks.findIndex(
     (pipelineTask) => pipelineTask.name === thisPipelineTask.name,
   );
-  const updatedTasks = tasks.map((pipelineTask) => {
-    if (pipelineTask.name !== thisPipelineTask.name) {
+
+  const getUpdatedTasks = (Updatetasks) =>
```
andrewballantyne, line 444: Type and camelCase the variable please.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -431,6 +512,25 @@ const fixInvalidListTask: UpdateOperationAction<UpdateOperationFixInvalidTaskLis
...
+  data,
+  finallyTasks,
+  finallyListTasks,
+) => {
+  const newTaskName = `finally-list-${getRandomChars(6)}`;
+
+  return {
+    finallyTasks: [...finallyTasks],
+    finallyListTasks: [...finallyListTasks, { name: newTaskName }],
+    tasks: [...tasks],
```
andrewballantyne, line 532: Not sure you need to re-wrap the items... omit spread? (and if you do one of my previous changes, omit entirely)

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const { t } = useTranslation();
+  const { finallyTasks = [], finallyListTasks = [] } = task;
+  const FINALLY_TASK_NODE_HEIGHT = NODE_HEIGHT + FINALLY_NODE_VERTICAL_SPACING;
+  const allTasksLength = finallyTasks.length + finallyListTasks.length;
+  return (
+    <Layer id="groups">
+      <>
+        <rect
+          className="odc-finally-node"
+          key={'rect'}
```
andrewballantyne, line 39: key?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+    pipeline,
+    pipelineRun,
+    disableBuilder,
+  } = element.getData();
+  const { t } = useTranslation();
+  const { finallyTasks = [], finallyListTasks = [] } = task;
+  const FINALLY_TASK_NODE_HEIGHT = NODE_HEIGHT + FINALLY_NODE_VERTICAL_SPACING;
+  const allTasksLength = finallyTasks.length + finallyListTasks.length;
+  return (
+    <Layer id="groups">
```
andrewballantyne, line 35: Perhaps I am missing something RE: Topology... why the Layer? Nothing else in Pipeline Topology has or needs a layer, this "node" is a glorified container for sure but it's self contained and there is no drag/drop sorta functionality...

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+          fill="transparent"
+          strokeWidth={1}
+          stroke="var(--pf-global--BorderColor--light-100)"
+          width={NODE_WIDTH + FINALLY_NODE_PADDING * 2}
+          height={
+            NODE_HEIGHT +
+            (allTasksLength - 1) * FINALLY_TASK_NODE_HEIGHT +
+            (!disableBuilder ? NODE_HEIGHT : 0) +
+            FINALLY_NODE_PADDING * 2
+          }
```
andrewballantyne, line 49: Curious, you calculated all of this in other spots... no reuse?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+  } = element.getData();
+  const { t } = useTranslation();
+  const { finallyTasks = [], finallyListTasks = [] } = task;
+  const FINALLY_TASK_NODE_HEIGHT = NODE_HEIGHT + FINALLY_NODE_VERTICAL_SPACING;
+  const allTasksLength = finallyTasks.length + finallyListTasks.length;
+  return (
+    <Layer id="groups">
+      <>
+        <rect
+          className="odc-finally-node"
```
andrewballantyne, line 38: ClassName?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+        />
+
+        {finallyTasks.map((ft, i) => (
+          <g
+            key={ft.name}
+            className="odc-finally-builder-node"
+            transform={`translate(${FINALLY_NODE_PADDING}, ${FINALLY_TASK_NODE_HEIGHT * i +
+              FINALLY_NODE_PADDING})`}
+          >
+            <rect width={NODE_WIDTH} height={NODE_HEIGHT} fill="transparent" />
```
andrewballantyne, line 61: A node sized transparent rectangle? What is this aiming to do?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+            FINALLY_NODE_PADDING * 2
+          }
+          rx="20"
+          ry="20"
+        />
+
+        {finallyTasks.map((ft, i) => (
+          <g
+            key={ft.name}
+            className="odc-finally-builder-node"
```
andrewballantyne, line 57: ClassName?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+            transform={`translate(${FINALLY_NODE_PADDING}, ${FINALLY_TASK_NODE_HEIGHT * i +
+              FINALLY_NODE_PADDING})`}
+          >
+            <rect width={NODE_WIDTH} height={NODE_HEIGHT} fill="transparent" />
+            <g onClick={ft.onTaskSelection && !disableBuilder ? () => ft.onTaskSelection() : null}>
```
andrewballantyne, line 62: Could this onClick not go on the parent group (`g`) node?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+          rx="20"
+          ry="20"
+        />
+
+        {finallyTasks.map((ft, i) => (
+          <g
+            key={ft.name}
+            className="odc-finally-builder-node"
+            transform={`translate(${FINALLY_NODE_PADDING}, ${FINALLY_TASK_NODE_HEIGHT * i +
+              FINALLY_NODE_PADDING})`}
```
andrewballantyne, line 59: Suggest the same fully qualifying y
```suggestion
            transform={`translate(${FINALLY_NODE_PADDING}, ${NODE_HEIGHT * i + 
              FINALLY_NODE_VERTICAL_SPACE * (i - 1) + FINALLY_NODE_PADDING})`}
```
I understand it's more long winded, but it's easier to read imo.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+              fill={blueColor.value}
+              x={FINALLY_ADD_LINK_TEXT_HEIGHT}
+              className="odc-add-finally-list-text"
+              style={{ cursor: 'pointer' }}
+              onClick={() => {
+                element.getData().task.addNewFinallyListNode(`test-${finallyTasks.length}`);
+              }}
+              textAnchor="middle"
+            >
+              Add finally tasks
```
andrewballantyne, line 122: i18n

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+            </g>
+          </g>
+        ))}
+        {!disableBuilder &&
+          finallyListTasks.map((flt, i) => (
+            <g
+              key={flt.name}
+              transform={`translate(${FINALLY_NODE_PADDING}, ${FINALLY_TASK_NODE_HEIGHT *
+                (i + finallyTasks.length) +
+                FINALLY_NODE_PADDING})`}
```
andrewballantyne, line 89: Same suggestion here about expanding the calculation

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+          ))}
+        {!disableBuilder && (
+          <g
+            transform={`translate(${(NODE_WIDTH + FINALLY_NODE_PADDING * 2) / 2}, ${allTasksLength *
+              FINALLY_TASK_NODE_HEIGHT +
+              FINALLY_ADD_LINK_TEXT_HEIGHT +
+              FINALLY_NODE_PADDING})`}
+          >
+            <PlusNodeDecorator
+              x={(FINALLY_NODE_PADDING + FINALLY_ADD_LINK_TEXT_HEIGHT + FINALLY_ADD_LINK_SIZE) * -1}
```
andrewballantyne, line 108: `FINALLY_ADD_LINK_TEXT_HEIGHT` in a horizontal calculation? I'm confused. Height is a vertical construct and rarely does it make sense in a horizontal calculation. Can I get a little help with understanding?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+              y={FINALLY_ADD_LINK_SIZE * -1}
+              tooltip={t('pipelines-plugin~Add finally tasks')}
+            />
+            <text
+              fill={blueColor.value}
+              x={FINALLY_ADD_LINK_TEXT_HEIGHT}
+              className="odc-add-finally-list-text"
+              style={{ cursor: 'pointer' }}
+              onClick={() => {
+                element.getData().task.addNewFinallyListNode(`test-${finallyTasks.length}`);
```
andrewballantyne, line 118: Needs to be on the group otherwise clicking on the "+" won't do anything.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+            <PlusNodeDecorator
+              x={(FINALLY_NODE_PADDING + FINALLY_ADD_LINK_TEXT_HEIGHT + FINALLY_ADD_LINK_SIZE) * -1}
+              y={FINALLY_ADD_LINK_SIZE * -1}
+              tooltip={t('pipelines-plugin~Add finally tasks')}
+            />
+            <text
+              fill={blueColor.value}
+              x={FINALLY_ADD_LINK_TEXT_HEIGHT}
+              className="odc-add-finally-list-text"
+              style={{ cursor: 'pointer' }}
```
andrewballantyne, line 116: I think this also could be on the `g`

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+          >
+            <PlusNodeDecorator
+              x={(FINALLY_NODE_PADDING + FINALLY_ADD_LINK_TEXT_HEIGHT + FINALLY_ADD_LINK_SIZE) * -1}
+              y={FINALLY_ADD_LINK_SIZE * -1}
+              tooltip={t('pipelines-plugin~Add finally tasks')}
+            />
+            <text
+              fill={blueColor.value}
+              x={FINALLY_ADD_LINK_TEXT_HEIGHT}
+              className="odc-add-finally-list-text"
```
andrewballantyne, line 115: Another className... also to note here, no `odc` prefix should be found in pipelines anymore (sadly there are a few)

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,131 @@
...
+                onNewTask={flt.convertList}
+              />
+            </g>
+          ))}
+        {!disableBuilder && (
+          <g
+            transform={`translate(${(NODE_WIDTH + FINALLY_NODE_PADDING * 2) / 2}, ${allTasksLength *
+              FINALLY_TASK_NODE_HEIGHT +
+              FINALLY_ADD_LINK_TEXT_HEIGHT +
+              FINALLY_NODE_PADDING})`}
```
andrewballantyne, line 105: Curious if we need to translate this at all (horizontally)... maybe the padding around the left edge? Is the goal here to centre it?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -22,13 +24,14 @@ import {
...
 } from './types';
 
-const createGenericNode: NodeCreatorSetup = (type, width?) => (name, data) => ({
+const createGenericNode: NodeCreatorSetup = (type, width?, height?) => (name, data) => ({
   id: name,
   data,
-  height: NODE_HEIGHT,
-  width: width != null ? width : NODE_WIDTH,
+  height: height || NODE_HEIGHT,
+  width: width || NODE_WIDTH,
```
andrewballantyne, line 34: If width is === 0 you will take `NODE_WIDTH` instead... undo that logic change.

andrewballantyne (changes requested):
Overall well done. I have a slew of nits and some required changes for you. Let me know if you disagree with anything, I'm open to discussion / explaining things more if you are unsure about any change.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/PipelineBuilderForm.tsx
```diff
@@ -83,35 +83,48 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
...
 
     setFieldValue('formData.tasks', tasks);
     setFieldValue('formData.listTasks', listTasks);
+    setFieldValue('formData.finallyTasks', finallyTasks);
+    setFieldValue('formData.finallyListTasks', finallyListTasks);
     updateErrors(taskErrors);
   };
 
-  const selectedId = values.formData.tasks[selectedTask?.taskIndex]?.name;
+  const builderNodes = selectedTask?.isFinallyTask ? 'finallyTasks' : 'tasks';
```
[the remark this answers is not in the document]
karthikjeeyar: yes, nodeType makes more sense, will update it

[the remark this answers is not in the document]
karthikjeeyar: yes, nodeType makes more sense, will update it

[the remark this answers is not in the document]
karthikjeeyar: yes, nodeType makes more sense, will update it

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
karthikjeeyar: Unlike pipeline builder, visualization in details page will not always have `finally section` (depending on pipeline definition),  so i had to conditionally handle the graph width and height based on the finally task availability, during the process I had to add this. But there are some more things that needs to handled in ODC-4312 like `Finally links` , `Finally task status`, `Tooltips` apart from this shared visualization component.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
```diff
@@ -77,6 +82,79 @@ export const useTasks = (namespace?: string): UseTasks => {
...
+  const convertListToFinallyTask = (resource: TaskKind, name: string) => {
+    const data: UpdateOperationConvertToTaskData = { resource, name };
+    onUpdateTasks(taskGroupRef.current, {
+      type: UpdateOperationType.CONVERT_LIST_TO_FINALLY_TASK,
+      data,
+    });
+  };
+
+  const finallyNodeName = `finally-node-${taskGroup.finallyTasks.length +
+    taskGroup.finallyListTasks.length}`;
```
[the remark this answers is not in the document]
karthikjeeyar: new nodes are pushed inside finally node, due to this dynamic nature of the height of the node  changing the finally-node name causes the layout to recalculate the x & y coordinates. Static name is causing the nodes and edges to dangle in the graph.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
karthikjeeyar: Used uniq from lodash here.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
karthikjeeyar: Should be singular node, updated

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
karthikjeeyar: Ah nice, it makes sense. updated

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
karthikjeeyar: Not required, It was leftover from my initial layout grouping approach, I removed it now.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
karthikjeeyar: Having a utility function for this, now i am making use of it here.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
karthikjeeyar: not required at all, removed it.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
karthikjeeyar: This one is to place the link elements ( icon + text) in the right spot in the finally node.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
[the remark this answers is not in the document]
andrewballantyne: Fair enough, any desire to do `finally-node-#-#` instead? That way when you convert one of the nodes it changes the id

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
andrewballantyne: Not sure you actually had to do this at all... since the graph would not need resizing if it did not render the finally outside of the builder. But it is what it is. Thanks for the clarification.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -248,37 +280,53 @@ const removeAndUpdateTasks = <
...
   >(listTaskName, listTasks, tasks);
+  const {
+    updateOnlyTasks: updateFinallyTasks,
+    updateAndRemoveTasks: updateFinallyListTasks,
+  } = removeAndUpdateTasks<PipelineBuilderListTask, PipelineTask>(
+    listTaskName,
+    finallyListTasks,
+    finallyTasks,
+  );
   return {
```
andrewballantyne, line 300: For consistency, please spread `taskGroup` here.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/update-utils.ts
```diff
@@ -434,26 +486,40 @@ const fixInvalidListTask: UpdateOperationAction<UpdateOperationFixInvalidTaskLis
...
+    finallyListTasks: [...finallyListTasks, { name: newTaskName }],
+  };
+};
+
 export const applyChange = (
   taskGroup: PipelineBuilderTaskGroup,
   op: UpdateOperation,
 ): CleanupResults => {
   const { type, data } = op;
-  const { tasks, listTasks } = taskGroup;
```
andrewballantyne, line 442: You went a different direction, so let us make it more pure than this currently.

```
const { tasks, listTasks, finallyTasks, finallyListTasks } = taskGroup;
const taskGrouping: PipelineBuilderTaskGrouping = { tasks, listTasks, finallyTasks, finallyListTasks };
// and pass `taskGrouping` into all the methods.
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/types.ts
```diff
@@ -118,11 +126,12 @@ export type UpdateOperationUpdateTaskData = UpdateOperationBaseData & {
...
   listTasks: PipelineBuilderListTask[];
+  finallyTasks: PipelineTask[];
+  finallyListTasks: PipelineBuilderListTask[];
   errors?: TaskErrorMap;
 };
 
 export type UpdateOperationAction<D extends UpdateOperationBaseData> = (
-  tasks: PipelineTask[],
-  listTasks: PipelineBuilderListTask[],
+  taskGroup: PipelineBuilderTaskGroup,
```
andrewballantyne, line 135:
```suggestion
  taskGroup: PipelineBuilderTaskGrouping,
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <TaskList
+              width={NODE_WIDTH}
+              height={NODE_HEIGHT}
+              listOptions={[...clusterTaskList, ...namespaceTaskList]}
+              onRemoveTask={flt.onRemoveTask}
+              onNewTask={flt.convertList}
+            />
+          </g>
+        ))}
+      {!disableBuilder && (
```
andrewballantyne, line 91: What do you think about removing `disableBuilder` prop and making two components with reusable sections from both. I'm not sure you're saving maintenance effort here, you seem to just condition a bunch of stuff and your rendering logic is almost all within the return statement, so it would be easy to break up into two components. `BuilderFinallyNode` and `FinallyNode`

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -22,13 +24,14 @@ import {
...
 } from './types';
 
-const createGenericNode: NodeCreatorSetup = (type, width?) => (name, data) => ({
+const createGenericNode: NodeCreatorSetup = (type, width?, height?) => (name, data) => ({
   id: name,
   data,
-  height: NODE_HEIGHT,
-  width: width != null ? width : NODE_WIDTH,
+  height: height ?? NODE_HEIGHT,
+  width: width ?? NODE_WIDTH,
```
andrewballantyne, line 34: Technically still a change in logic... but I think it gets the spirit of what I was after.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/utils.ts
```diff
@@ -223,14 +229,74 @@ export const getEdgesFromNodes = (nodes: PipelineMixedNodeModel[]): PipelineEdge
...
+  );
+};
+
+export const getLastRegularTasks = (regularTasks: PipelineMixedNodeModel[]): string[] => {
+  const runAfters = regularTasks.reduce((acc, { data: { task: { runAfter } } }) => {
+    if (runAfter) {
+      acc.push(...runAfter);
+    }
+    return _.uniq(acc);
+  }, []);
```
andrewballantyne, line 247: Not that it really matters, but this is inefficient. `uniq` after the fact.
```suggestion
  const runAfters = _.uniq(regularTasks.reduce((acc, { data: { task: { runAfter } } }) => {
    // more simple? 
    return runAfter ? acc.concat(runAfter) : acc;
  }, []));
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('getLastRegularTasks', () => {
+  it('expect to handle the empty array input', () => {
+    expect(getLastRegularTasks([])).toHaveLength(0);
+  });
+
+  it('expect to return the list of task names that will be executed at the end', () => {
+    const { nodes } = getTopologyNodesEdges(pipeline);
+    expect(getLastRegularTasks(nodes)).toHaveLength(1);
+    expect(getLastRegularTasks(nodes)).toEqual(['verify']);
+  });
```
andrewballantyne, line 39: Although a good test, it's not comprehensive. Could we get a test where there are at least two end points?

andrewballantyne (changes requested):
Some final notes. I'll swing back later today to run the builder through more comprehensive tests.

bgliwa01 (approved):
LGTM! Thanks Karthik!

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/__tests__/utils.spec.ts
karthikjeeyar: Updated the assertion to test multiple leaf nodes.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
karthikjeeyar: Separated into two different components.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/BuilderFinallyNode.tsx
```diff
@@ -0,0 +1,117 @@
...
+  FINALLY_ADD_LINK_TEXT_HEIGHT,
+  BUILDER_NODE_ERROR_RADIUS,
+  FINALLY_ADD_LINK_SIZE,
+} from './const';
+import PlusNodeDecorator from './PlusNodeDecorator';
+import ErrorNodeDecorator from './ErrorNodeDecorator';
+import TaskList from './TaskList';
+import { getFinallyTaskHeight } from './utils';
+
+const BuilderFinallyNode: React.FC<{ element: Node }> = ({ element }) => {
```
andrewballantyne, line 22: Best type Element better... ie [BuilderNode](https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/BuilderNode.tsx#L17) has the data typed.

You'll need to expand your types to have both expected props for the `FinallyNode` as well as the `BuilderFinallyNode`.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/types.ts
```diff
@@ -17,6 +17,42 @@ export type PipelineRunAfterNodeModelData = {
...
 };
+
+type finallyTask = {
+  onTaskSelection?: () => void;
+  error?: string;
+  selected?: boolean;
+  disableTooltip?: boolean;
+};
+
+export type FinallyNodeModalData = {
```
andrewballantyne, line 28:
```suggestion
export type FinallyNodeModelData = {
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { PipelineVisualizationTask } from '../detail-page-tabs/pipeline-details/PipelineVisualizationTask';
+import {
+  NODE_WIDTH,
+  NODE_HEIGHT,
+  FINALLY_NODE_PADDING,
+  FINALLY_NODE_VERTICAL_SPACING,
+} from './const';
+import { getFinallyTaskHeight } from './utils';
+
+const FinallyNode: React.FC<{ element: Node }> = ({ element }) => {
```
andrewballantyne, line 13: Need to type Element.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/FinallyNode.tsx
```diff
@@ -0,0 +1,51 @@
...
+          transform={`translate(${FINALLY_NODE_PADDING}, ${NODE_HEIGHT * i +
+            FINALLY_NODE_VERTICAL_SPACING * i +
+            FINALLY_NODE_PADDING})`}
+        >
+          <PipelineVisualizationTask
+            pipelineRunName={pipelineRun?.metadata?.name}
+            task={ft}
+            pipelineRunStatus={pipelineRun && pipelineRunFilterReducer(pipelineRun)}
+            namespace={pipeline?.metadata?.namespace}
+            disableTooltip={false}
```
andrewballantyne, line 40:
```suggestion
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    expect(edges).toHaveLength(20);
+  });
+});
+
+describe('getLastRegularTasks', () => {
+  it('expect to handle the empty array input', () => {
+    expect(getLastRegularTasks([])).toHaveLength(0);
+  });
+
+  it('expect to return the list of task names that will be executed at the end', () => {
```
andrewballantyne, line 35: Not sure why you dropped the other test; we can have two tests 😅

andrewballantyne (changes requested):
You have some unused/unneeded code in your Builder node... it doesn't address status at all so it doesn't need those details. Plus you have some gaps in your types for the new nodes. Let me know if you need any help with the types.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/__tests__/utils.spec.ts
karthikjeeyar: Added Types to the new components and brought back the previous test.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/types.ts
```diff
@@ -17,6 +17,49 @@ export type PipelineRunAfterNodeModelData = {
...
+  name: string;
+  runAfter: string[];
+  selected?: boolean;
+  isFinallyTask: boolean;
+  finallyTasks: FinallyTask[];
+};
+export type FinallyNodeData = {
+  task: FinallyNodeTask;
+};
+export type FinallyBuilderNodeData = {
```
andrewballantyne, line 44: Consistency would be helpful for future lookup of files/types. `Finally + Builder` or `Builder + Finally`. Your component is `BuilderFinallyNode`, your type is `FinallyBuilderNode`

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/types.ts
```diff
@@ -45,6 +88,8 @@ export type PipelineMixedNodeModel = PipelineNodeModel<PipelineRunAfterNodeModel
 export type PipelineTaskNodeModel = PipelineNodeModel<TaskNodeModelData>;
 export type PipelineBuilderTaskNodeModel = PipelineNodeModel<BuilderNodeModelData>;
 export type PipelineTaskListNodeModel = PipelineNodeModel<TaskListNodeModelData>;
+export type PipelineFinallyNodeModel = PipelineNodeModel<FinallyNodeModel>;
+export type PipelineFinallyBuilderNodeModel = PipelineNodeModel<BuilderFinallyNodeModel>;
```
andrewballantyne, line 92: Do a quick search for all `FinallyBuilder` and `BuilderFinally` usages, and align under one banner please 😄

andrewballantyne (changes requested):
Some inconsistencies in your naming that will irk me something fierce if I had to look up a type or node in the future 😉 

Please update to one or the other for my grey-hair prevention plan 🙂

andrewballantyne (approved):

nemesis09:
verified changes locally
/lgtm

PR #8391, opened 2021-03-16
Bug 1937018: Fix null handling in FilterToolbar_ to prevent possible exceptions

TheRealJon:
- Remove initializer for deconstructed prop
- Add optional chaining/null coalescence wherever the rowFilters prop is used in FilterToolbar_
- Update ProvidedAPIs component so that it does not pass 'null' into this prop


spadgett (approved):
/lgtm

PR #8392, opened 2021-03-16
Bug 1932281:  inform user when channels are unknown

rhamilto:
if `!clusterVersion.spec.channel`, add alert instructing to configure a channel and improve Channel empty state indicator.
Screen Shot 2021-03-29 at 3 26 24 PM

if `clusterVersion.status.desired.channels`
Screen Shot 2021-03-29 at 3 24 20 PM

if `!clusterVersion.status.desired.channels`
Screen Shot 2021-03-29 at 3 23 25 PM
Screen Shot 2021-03-29 at 3 23 39 PM

if version is not found in `clusterVersion.spec.channel`
Screen Shot 2021-03-29 at 3 26 14 PM


zherman0 (approved):

frontend/public/components/modals/cluster-channel-modal.tsx
```diff
@@ -22,34 +29,67 @@ const ClusterChannelModal = withHandlePromise((props: ClusterChannelModalProps)
     o[val] = val;
     return o;
   }, {});
+  const version = semver.parse(getLastCompletedUpdate(cv));
   const submit: React.FormEventHandler<HTMLFormElement> = (e) => {
     e.preventDefault();
     const patch = [{ op: 'add', path: '/spec/channel', value: channel }];
     return handlePromise(k8sPatch(ClusterVersionModel, cv, patch), close);
   };
+  const channelsExist = cv.status.desired.channels;
```
spadgett, line 38: I'd be careful and use optional chaining.

```suggestion
  const channelsExist = cv.status?.desired?.channels?.length;
```

spadgett (commented):
LGTM, just one nit to avoid a runtime error if the `status` stanza isn't yet defined. I'm not sure if this would ever happen in a real cluster, but I would code defensively just in case since it's not a required field in the API definition.

frontend/public/components/modals/cluster-channel-modal.tsx
rhamilto: Thanks, @spadgett.  Updated.

rhamilto:
Added an integration test fix.

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

PR #8400, opened 2021-03-17
converting yaml-templates extension types to dynamic-plugin-sdk

glekner:
JIRA: https://issues.redhat.com/browse/CONSOLE-2540


frontend/packages/console-dynamic-plugin-sdk/src/utils/k8s.ts
```diff
@@ -0,0 +1,17 @@
+import { K8sKind } from '@console/internal/module/k8s/types';
+
+/** get K8sKind from extension's model */
+export const getK8sKindFromExtension = ({
```
christianvogt, line 4: This shouldn't be in the SDK. It's specific to our current implementation of `K8sKind`.

Maybe even change the signature of `referenceForModel`:

```
export const referenceForModel = (model: K8sKind | { group: string, version: string, kind: string}): GroupVersionKind =>
```

frontend/public/components/create-yaml.tsx
```diff
@@ -27,7 +31,7 @@ export const CreateYAML = connectToPlural((props: CreateYAMLProps) => {
   } = props;
   const { params } = match;
   const { t } = useTranslation();
-  const templateExtensions = useExtensions<YAMLTemplate>(isYAMLTemplate);
+  const [templateExtensions] = useResolvedExtensions<YAMLTemplate>(isYAMLTemplate);
```
christianvogt, line 34: Why change? There are no code references to resolve.

frontend/public/components/edit-yaml.jsx
```diff
@@ -50,13 +50,20 @@ const stateToProps = ({ k8s, UI }) => ({
...
+  };
+
 /**
  * This component loads the entire Monaco editor library with it.
  * Consider using `AsyncComponent` to dynamically load this component when needed.
  */
 /** @augments {React.Component<{obj?: any, create: boolean, kind: string, redirectURL?: string, resourceObjPath?: (obj: K8sResourceKind, objRef: string) => string}, onChange?: (yaml: string) => void>} */
 export const EditYAML_ = connect(stateToProps)(
-  withExtensions({ templateExtensions: isYAMLTemplate })(
+  WithYamlTemplates(
```
christianvogt, line 66: Why change this? You should be fine without the need to resolve.

frontend/packages/kubevirt-plugin/console-extensions.json
```diff
@@ -30,5 +30,29 @@
...
+        "group": "kubevirt.io",
+        "kind": "VirtualMachine",
+        "version": "v1alpha3"
+      },
+      "template": "apiVersion: kubevirt.io/v1alpha3 kind: VirtualMachine metadata: name: vm-example labels: app: vm-example os.template.kubevirt.io/fedora31: 'true' workload.template.kubevirt.io/server: 'true' annotations: name.os.template.kubevirt.io/fedora31: Fedora 31 description: VM example spec: running: false template: metadata: labels: kubevirt.io/domain: vm-example vm.kubevirt.io/name: vm-example os.template.kubevirt.io/fedora31: 'true' workload.template.kubevirt.io/server: 'true' spec: domain: cpu: cores: 1 sockets: 1 threads: 1 devices: disks: - bootOrder: 1 disk: bus: virtio name: containerdisk - disk: bus: virtio name: cloudinitdisk interfaces: - masquerade: {} name: default networkInterfaceMultiqueue: true rng: {} resources: requests: memory: 1G hostname: vm-example networks: - name: default pod: {} terminationGracePeriodSeconds: 0 volumes: - containerDisk: image: 'kubevirt/fedora-cloud-container-disk-demo:latest' name: containerdisk - cloudInitNoCloud: userData: |- #cloud-config password: fedora chpasswd: { expire: False } name: cloudinitdisk"
```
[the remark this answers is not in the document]
christianvogt: Ah! maybe we should use `console-extensions.yaml` instead of json lol (....hmmm....+1)

You could have used: https://onlineyamltools.com/convert-yaml-to-json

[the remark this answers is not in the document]
glekner: this tool converts yaml to json, not a multiline string to a single line string with spaces/line breaks

frontend/packages/console-dynamic-plugin-sdk/src/utils/k8s.ts
glekner: but `k8sKind` has different names for the fields. `apiGroup` and `apiVersion`

glekner: we can maybe create another `referenceForExtensionModel`

frontend/packages/kubevirt-plugin/console-extensions.json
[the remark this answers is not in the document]
christianvogt: whoops. you're right.
So convert all new lines to `\n` and all `\` to `\\`. Then decode when you read.

glekner:
- rebased and updated `YAMLTemplate` extension: the template is now a CodeRef that resolves into a string to avoid writing multiline strings in the json file.

@christianvogt I addressed your comments, but now we do need `useResolvedExtensions` :)

frontend/public/components/create-yaml.tsx
christianvogt, line 34: You only need to resolve the templates which will actually be used. That could be done by augmenting the guard.
```suggestion
  const [templateExtensions] = useResolvedExtensions<YAMLTemplate>((e) => isYAMLTemplate(e) && ...);
```

You only want to resolve code refs for the extensions that are actually being used.

Resolution of templates is now asynchronous. To prevent flashing, you should handle the loading case and show the `LoadingBox` otherwise there's potential to show a different template before the extensions load.

frontend/public/components/edit-yaml.jsx
```diff
@@ -50,13 +50,20 @@ const stateToProps = ({ k8s, UI }) => ({
   models: k8s.getIn(['RESOURCES', 'models']),
 });
 
+const WithYamlTemplates = (Component) =>
+  function Comp(props) {
+    const [templateExtensions] = useResolvedExtensions(isYAMLTemplate);
```
christianvogt, line 55: Ideally we don't resolve all templates when we only actually need a specific one.
Again, you can augment the filter function.

frontend/packages/knative-plugin/src/yaml-templates.ts
```diff
@@ -16,3 +16,5 @@ spec:
       - image: openshift/hello-openshift
 `,
 );
+
+export const defaultYamlTemplate = yamlTemplates.getIn([ServiceModel, 'default']);
```
christianvogt, line 20: We don't need the `ImmutableMap` here anymore. We can export the template directly:
```suggestion
export const defaultServiceYamlTemplate = `...`
```

frontend/public/components/edit-yaml.jsx
```diff
@@ -50,13 +50,26 @@ const stateToProps = ({ k8s, UI }) => ({
   models: k8s.getIn(['RESOURCES', 'models']),
 });
 
+const WithYamlTemplates = (Component) =>
+  function Comp(props) {
+    const [templateExtensions, resolvedTemplates] = useResolvedExtensions(
+      (e) => isYAMLTemplate(e) && e?.properties?.model?.kind === props?.obj?.kind,
```
christianvogt, line 56: Do you need the `?` optional checks?
Definitely the extension itself must conform and have the model and kind.

```suggestion
      (e) => isYAMLTemplate(e) && e.properties.model.kind === props.obj.kind,
```

Looks like this only checks the `kind` and not `group` and `version`. Why not?

frontend/public/components/create-yaml.tsx
```diff
@@ -27,7 +31,9 @@ export const CreateYAML = connectToPlural((props: CreateYAMLProps) => {
   } = props;
   const { params } = match;
   const { t } = useTranslation();
-  const templateExtensions = useExtensions<YAMLTemplate>(isYAMLTemplate);
+  const [templateExtensions, resolvedTemplates] = useResolvedExtensions<YAMLTemplate>(
+    (e): e is YAMLTemplate => isYAMLTemplate(e) && e?.properties?.model?.kind === kindObj?.kind,
```
christianvogt, line 35: Same comment here as the one made for `WithYamlTemplates `

frontend/packages/knative-plugin/src/yaml-templates.ts
glekner: fixed

frontend/public/components/edit-yaml.jsx
glekner: `group` and `version` are not consistent. 
for PVC for example the `apiVersion` is `apiVersion: v1`
for VMTemplate its `kubevirt.io/v1alpha3`

some use only version, some use group/version
why kind is not enough here? AFAIK we don't have multiple templates for the same kind

glekner:
@christianvogt I added some comments.
the filtering inside `useResolvedExtensions` causes infinite re-renders, so I did the filtering outside. we can file a bug on useResolvedExtensions and deal with it in another PR

frontend/public/components/edit-yaml.jsx
christianvogt: Oh that's interesting. We should track this as a separate issue then because that seems like a bug somewhere.
At the end of the day, this filtering at the extension level is only the first step. Even if we resolve all extensions, it still works. So for now just checking kind will work and will already reduce the number of code refs we need to resolve.

christianvogt:
> @christianvogt I added some comments.
> the filtering inside `useResolvedExtensions` causes infinite re-renders, so I did the filtering outside. we can file a bug on useResolvedExtensions and deal with it in another PR

cc @vojtechszocs 
@glekner the whole point of filtering with the typeguard function is to limit the number of extension code refs that are resolved. Filtering outside is no different than what was done previous; resolves ALL template code refs (even those that are unrelated). If you didn't do the filtering at all, the correct template would still be chosen because they all get put into a map with the model reference string as the key.

glekner:
@christianvogt @vojtechszocs should this issue block us?
there are currently 9 YAML extensions, which are strings, so not very heavy.

christianvogt:
@glekner it shouldn't block. But please add a comment with a TODO to add the proper filtering. I just thought we would have a simple change to make it right.

glekner:
added comments @christianvogt

christianvogt:
Tested all yaml template contributed through the new extension to be functioning as expected.

Thanks @glekner  for fixing the render loop issue and addressing the filtering issue.

/approve
/lgtm
/hold

spadgett:
/assign @sferich888 @ahardin-rh @yapei
for approvals

There are no user visible changes with this PR. @glekner, can you give QE some guidance on what areas to regression test?

spadgett:
/label px-approved

We have general PX approval for the dynamic plugin SDK PRs.

frontend/public/components/edit-yaml.jsx
spadgett: I haven't reviewed the PR closely so I don't know the context, but `kind` is not unique on its own. We should be checking the group everywhere as well. The core k8s types that existed before there were API groups only have the version in `apiVersion` without a group (because it is not in a group). New types will all have groups, however.

yapei:
Communicated with @glekner on slack about the regression testing steps, no issues found, approving

PR #8403, opened 2021-03-17
quick search modal close on starting quickstart

abhinandan13jan:
# Adresses
https://issues.redhat.com/browse/ODC-5600

# Issue
quick search modal doesn't close on starting quickstart

# Screenshot
qss2
qss1

# Tests
No tests changed

# Browser conformance
Chrome

### DIFF
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchContent.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchContent.tsx
@@ -17,6 +17,7 @@ interface QuickSearchContentProps {
   selectedItem: CatalogItem;
   onSelect: (itemId: string) => void;
   viewAll?: CatalogLinkData[];
+  closeModal: () => void;
 }
 
 const QuickSearchContent: React.FC<QuickSearchContentProps> = ({
@@ -27,6 +28,7 @@ const QuickSearchContent: React.FC<QuickSearchContentProps> = ({
   selectedItem,
   selectedItemId,
   onSelect,
+  closeModal,
 }) => {
   return (
     <Split className="odc-quick-search-content">
@@ -43,11 +45,12 @@ const QuickSearchContent: React.FC<QuickSearchContentProps> = ({
           searchTerm={searchTerm}
           namespace={namespace}
           onSelectListItem={onSelect}
+          closeModal={closeModal}
         />
       </SplitItem>
       <Divider component="div" isVertical />
       <SplitItem className="odc-quick-search-content__details">
-        <QuickSearchDetails selectedItem={selectedItem} />
+        <QuickSearchDetails selectedItem={selectedItem} closeModal={closeModal} />
       </SplitItem>
     </Split>
   );
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
@@ -1,15 +1,16 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { history } from '@console/internal/components/utils';
 import { CatalogItem } from '@console/dynamic-plugin-sdk';
 import { Button, ButtonVariant, TextContent, Title } from '@patternfly/react-core';
 import './QuickSearchDetails.scss';
+import { handleCta } from './utils/quick-search-utils';
 
 interface QuickSearchDetailsProps {
   selectedItem: CatalogItem;
+  closeModal: () => void;
 }
 
-const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem }) => {
+const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem, closeModal }) => {
   const { t } = useTranslation();
 
   return (
@@ -23,9 +24,8 @@ const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem })
       <Button
         variant={ButtonVariant.primary}
         className="odc-quick-search-details__form-button"
-        onClick={() => {
-          const { href, callback } = selectedItem.cta;
-          callback ? callback() : history.push(href);
+        onClick={(e) => {
+          handleCta(e, selectedItem, closeModal);
         }}
       >
         {selectedItem.cta.label}
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import cx from 'classnames';
-import { history } from '@console/internal/components/utils';
 import { CatalogItem } from '@console/dynamic-plugin-sdk';
 import {
   DataList,
@@ -20,6 +19,7 @@ import { getIconProps } from '@console/dev-console/src/components/catalog/utils/
 import { getImageForIconClass } from '@console/internal/components/catalog/catalog-item-icon';
 import './QuickSearchList.scss';
 import { CatalogLinkData } from './utils/quick-search-types';
+import { handleCta } from './utils/quick-search-utils';
 import { Link } from 'react-router-dom';
 
 interface QuickSearchListProps {
@@ -29,6 +29,7 @@ interface QuickSearchListProps {
   searchTerm: string;
   namespace: string;
   onSelectListItem: (itemId: string) => void;
+  closeModal: () => void;
 }
 
 const QuickSearchList: React.FC<QuickSearchListProps> = ({
@@ -36,14 +37,9 @@ const QuickSearchList: React.FC<QuickSearchListProps> = ({
   viewAll,
   selectedItemId,
   onSelectListItem,
+  closeModal,
 }) => {
   const { t } = useTranslation();
-
-  const handleCta = (e: React.SyntheticEvent, item: CatalogItem) => {
-    e.preventDefault();
-    const { href, callback } = item.cta;
-    callback ? callback() : history.push(href);
-  };
 
   const getIcon = (item: CatalogItem) => {
     const { iconImg, iconClass } = getIconProps(item);
@@ -73,7 +69,7 @@ const QuickSearchList: React.FC<QuickSearchListProps> = ({
             className={cx('odc-quick-search-list__item', {
               'odc-quick-search-list__item--highlight': item.uid === selectedItemId,
             })}
-            onDoubleClick={(e: React.SyntheticEvent) => handleCta(e, item)}
+            onDoubleClick={(e: React.SyntheticEvent) => handleCta(e, item, closeModal)}
           >
             <DataListItemRow className="odc-quick-search-list__item-row">
               <DataListItemCells
--- a/frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
+++ b/frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { CatalogItem } from '@console/dynamic-plugin-sdk';
 import {
-  history,
   getQueryArgument,
   removeQueryArgument,
   setQueryArgument,
@@ -12,6 +11,7 @@ import './QuickSearchButton.scss';
 
 import './QuickSearchModalBody.scss';
 import { CatalogLinkData, QuickSearchData } from './utils/quick-search-types';
+import { handleCta } from './utils/quick-search-utils';
 
 interface QuickSearchModalBodyProps {
   allCatalogItemsLoaded: boolean;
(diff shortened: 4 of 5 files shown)


christianvogt:
Good catch. That looks like there's also an existing bug.
How to reproduce:

- open quick search
- enter a term
- click away from the modal to dismiss it
- Observe: URL still contains query param: `catalogSearch=`
- refresh the page
- Observe: topology loads with quick search open

That being said. I think clicking the `start` button of a quick start should clear the search string and close the modal. This should be fixed as part of this issue.

christianvogt:
Please make sure not to break the use case that when the user clicks on the create button of a dev-catalog item, which navigates the user to a new page, if they press the browser back button they should return to topology with the quick search open.

rottencandy (approved):
Thanks. Tested and looks good.
/lgtm

divyanshiGupta:
@abhinandan13jan you will need to remove the search and close the modal on enter and double click as well since by pressing enter or doing double click on a selected quick-start item also opens the quickstart. Update `handleCta` in `QuickSearchModalList` and use it for all of these actions.

abhinandan13jan:
> @abhinandan13jan you will need to remove the search and close the modal on enter and double click as well since by pressing enter or doing double click on a selected quick-start item also opens the quickstart. Update `handleCta` in `QuickSearchModalList` and use it for all of these actions.

@divyanshiGupta I've updated to handle cases for enter and double click. Could you please take another look

divyanshiGupta:
@abhinandan13jan in my above comment I asked you to update handleCta and use it for all these actions.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -25,7 +26,13 @@ const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem })
         className="odc-quick-search-details__form-button"
         onClick={() => {
           const { href, callback } = selectedItem.cta;
-          callback ? callback() : history.push(href);
+          if (callback) {
+            callback();
+            removeQueryArgument('catalogSearch');
```
rohitkrai03, line 31: You should not remove query arguments directly like this because it breaks the use case where if user clicks on back button on browser the modal will not open with previous search result.

frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -29,20 +29,26 @@ interface QuickSearchListProps {
...
 }) => {
   const { t } = useTranslation();
 
   const handleCta = (e: React.SyntheticEvent, item: CatalogItem) => {
     e.preventDefault();
     const { href, callback } = item.cta;
-    callback ? callback() : history.push(href);
+    if (callback) {
+      callback();
+      removeQueryArgument('catalogSearch');
```
rohitkrai03, line 49: Same as above.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -88,9 +88,15 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
   const onEnter = React.useCallback(() => {
     if (selectedItem) {
       const { href, callback } = selectedItem.cta;
-      callback ? callback() : history.push(href);
+      if (callback) {
+        callback();
+        removeQueryArgument('catalogSearch');
```
rohitkrai03, line 93: Same as above.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -132,6 +138,7 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
 
     const onOutsideClick = (e: MouseEvent) => {
       if (!ref.current?.contains(e.target as Node)) {
+        removeQueryArgument('catalogSearch');
```
rohitkrai03, line 141: You should only remove the query argument if there's no search term. We want to save the query argument if user clicks outside in the middle of search and comes back to the search.

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
abhinandan13jan: > Good catch. That looks like there's also an existing bug.
> How to reproduce:
> 
> * open quick search
> * enter a term
> * click away from the modal to dismiss it
> * Observe: URL still contains query param: `catalogSearch=`
> * refresh the page
> * Observe: topology loads with quick search open
> 
> That being said. I think clicking the `start` button of a quick start should clear the search string and close the modal. This should be fixed as part of this issue.

@rohitkrai03 The idea is to remove the QueryArg in case of a click/callback of QuickStart and not for the case of going for an installation / catalog operation. It will still contain the queryparam

abhinandan13jan:
> @abhinandan13jan in my above comment I asked you to update handleCta and use it for all these actions.

@divyanshiGupta I am unable to understand what you mean? Do you want me to move the function handleCta up one level, or do you want it to pass a callback to it's parent. As the operation to be done is in multiple components in different levels of the hierarchy.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
divyanshiGupta: We are already removing query argument when there is no search term.

divyanshiGupta:
> > @abhinandan13jan in my above comment I asked you to update handleCta and use it for all these actions.
> 
> @divyanshiGupta I am unable to understand what you mean? Do you want me to move the function handleCta up one level, or do you want it to pass a callback to it's parent. As the operation to be done is in multiple components in different levels of the hierarchy or as a util?

@abhinandan13jan it simply means that create a utility/update `handleCta` according to the new use case so that you don't have to make same changes at 3 different places.

abhinandan13jan:
> > > @abhinandan13jan in my above comment I asked you to update handleCta and use it for all these actions.
> > 
> > 
> > @divyanshiGupta I am unable to understand what you mean? Do you want me to move the function handleCta up one level, or do you want it to pass a callback to it's parent. As the operation to be done is in multiple components in different levels of the hierarchy or as a util?
> 
> @abhinandan13jan it simply means that create a utility/update `handleCta` according to the new use case so that you don't have to make same changes at 3 different places.

Updated. please revisit

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
divyanshiGupta: @abhinandan13jan as @rohitkrai03 mentioned don't remove query argument on outside click since the user should be able to come back to their incomplete search after the modal got closed on outside click.

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -53,3 +54,12 @@ export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogIte
     [t, quickStarts, setActiveQuickStart],
   );
 };
+
+export const handleCta = (e: React.SyntheticEvent, item: CatalogItem, clodemodal: () => void) => {
```
divyanshiGupta, line 58:
```suggestion
export const handleCta = (e: React.SyntheticEvent, item: CatalogItem, closeModal: () => void) => {
```

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -53,3 +54,12 @@ export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogIte
...
+  e.preventDefault();
+  const { href, callback } = item.cta;
+  if (callback) {
+    clodemodal();
```
divyanshiGupta, line 62:
```suggestion
    closeModal();
```

divyanshiGupta (changes requested):

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
abhinandan13jan: updated

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
abhinandan13jan: updated

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
rohitkrai03: @abhinandan13jan ^^

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -53,3 +54,12 @@ export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogIte
     [t, quickStarts, setActiveQuickStart],
   );
 };
+
+export const handleCta = (e: React.SyntheticEvent, item: CatalogItem, clodeModal: () => void) => {
```
rohitkrai03, line 58:
```suggestion
export const handleCta = (e: React.SyntheticEvent, item: CatalogItem, closeModal: () => void) => {
```

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -53,3 +54,12 @@ export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogIte
...
+  e.preventDefault();
+  const { href, callback } = item.cta;
+  if (callback) {
+    clodeModal();
```
rohitkrai03, line 62:
```suggestion
    closeModal();
```

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
rohitkrai03: @abhinandan13jan The UX designs had this use case that if user goes away to some other page clicking on some action on the quick search modal and clicks back button on the browser, he should be brought back to topology with the modal open with the same search term as before. This use case will be broken by your changes.

divyanshiGupta: @rohitkrai03 in this case it makes sense since the query argument will only be removed when there is a callback.

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -53,3 +54,12 @@ export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogIte
...
+    removeQueryArgument('catalogSearch');
+  } else history.push(href);
```
christianvogt, line 64: What am I missing? Looks to me like `callback` is never called.

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
abhinandan13jan: updated

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
abhinandan13jan: updated

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
abhinandan13jan: updated

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
abhinandan13jan: added callback

frontend/packages/topology/src/components/quick-search/QuickSearchContent.tsx
```diff
@@ -17,6 +17,7 @@ interface QuickSearchContentProps {
   selectedItem: CatalogItem;
   onSelect: (itemId: string) => void;
   viewAll?: CatalogLinkData[];
+  closeModal?: () => void;
```
divyanshiGupta, line 20: Why is `closeModal` optional here and required in list and details component?
```suggestion
  closeModal: () => void;
```

divyanshiGupta:
@abhinandan13jan I verified this locally and it works as expected. Resolve the above comment and then I will add lgtm.

frontend/packages/topology/src/components/quick-search/QuickSearchContent.tsx
abhinandan13jan: fixed

abhinandan13jan:
@divyanshiGupta Thanks for the review. I've fixed as per the comment above.

rohitkrai03 (approved):
/approve

PR #8407, opened 2021-03-17
Bug 1940206: i18n details page text

rebeccaalpert:
A couple of empty states and options weren't internationalized yet.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1940206.


frontend/public/components/volumes-table.tsx
```diff
@@ -209,6 +144,78 @@ export const VolumesTable = (props) => {
...
     },
   ];
+  const VolumesTableRows = ({ componentProps: { componentData } }) => {
+    return _.map(componentData, (volume: RowVolumeData) => {
+      const {
+        container,
+        mountPath,
+        name,
+        readOnly,
+        volumeResource,
```
spadgett, line 154: I realize you're trying to avoid shadowing variables from the outer scope, but `volumeResource` and `volumePod` are confusing names. It sounds like they refer to a `PersistentVolume` or container storage. I would pick something different.

rebeccaalpert: All set.

spadgett (approved):
/lgtm

spadgett:
/lgtm cancel

Looks like there's a build error:

```
ERROR in /go/src/github.com/openshift/console/frontend/public/components/volumes-table.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/volumes-table.tsx(154,9):
TS2339: Property 'tableRowResource' does not exist on type 'RowVolumeData'.
```

rebeccaalpert:
My bad; sorry. Should be all set.

spadgett (approved):
/lgtm

spadgett:
/lgtm cancel

@rebeccaalpert It looks like this is consistently failing CI. Can you take a look?

rebeccaalpert:
@spadgett - I see the test failure, but I'm not sure how any changes I made are causing it.

spadgett:
> @spadgett - I see the test failure, but I'm not sure how any changes I made are causing it.

@rebeccaalpert I would try to run the test with headless mode disabled to see what's happening. You could also try to step through the test scenario manually to make sure we haven't introduced a regression with the volumes table.

rebeccaalpert:
Yup, I was doing that last week. The actions button isn't being clicked so the test fails. I don't see how that has anything to do with my changes.

zherman0 (approved):

rebeccaalpert:
I think I figured out what was wrong with the tests. (I pulled the row back out of the calling component and they're now passing locally.) I'll see what CI thinks.

spadgett (approved):
/lgtm

PR #8408, opened 2021-03-17
Cypress:  numerically naming cypress screenshot files

dtaylor113:
Prepends cypress screenshot filenames with a number, ex '1_', '2_', etc..  The CI artifacts cypress screenshot directory (/artifacts/e2e-gcp-console/test/artifacts/gui_test_screenshots/cypress/screenshots/) seems to sort screenshot filenames alphanumerically, so adding this will order the screenshots from first to last (which doesn't seem to correspond to `Modified` datetime).

This should order the files by: `before all, before each, test suite, retries, after each, after all,` and should help us quickly determine what failed first.

image

### DIFF
--- a/frontend/packages/integration-tests-cypress/plugins/index.js
+++ b/frontend/packages/integration-tests-cypress/plugins/index.js
@@ -2,6 +2,7 @@
 // the project's config changing)
 const wp = require('@cypress/webpack-preprocessor');
 const fs = require('fs');
+const path = require('path');
 
 module.exports = (on, config) => {
   const options = {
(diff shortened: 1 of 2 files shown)


frontend/packages/integration-tests-cypress/plugins/index.js
```diff
@@ -41,6 +41,23 @@ module.exports = (on, config) => {
       return null;
     },
   });
+  on('after:screenshot', (details) => {
+    // Prepend "1_", "2_", etc. to screenshot filenames because they are sorted alphanumerically in CI's artifacts dir
+    const screenshotFolder = details.path.substr(0, details.path.lastIndexOf('/'));
+    const fileName = details.path.substr(details.path.lastIndexOf('/') + 1);
```
spadgett, line 47: This probably breaks if you run the tests on Windows. It would be better to use a node utility for parsing the path.

dtaylor113:
Another example:

image

frontend/packages/integration-tests-cypress/plugins/index.js
dtaylor113: updated to use `path.parse`

dtaylor113:
Added back the WIP label as I have to investigate if other packages (olm, ceph, etc..) utilize `frontend/packages/integration-tests-cypress/plugins/index.js` in order to get the cypress screenshot renaming.

dtaylor113:
Hi @spadgett, @rhamilto, @TheRealJon, pls review when you can.  Would like to get this in to help e2e debugging.  -thanks

spadgett (approved):
/lgtm

PR #8146, opened 2021-02-11
Adding aggregated services to catalog, topology and various views.

wtrocki:
## Motivation

This PR adds Firepig OpenShift Community plugin that will put base for discovering services in the OpenShift console. This PR is still containing some elements not aligned with the design. There are also couple issues that we need to solve. Creating for early preview.

## Verification

To verify changes we need to install firepig operator 
The operator can be added as a separate catalog source.
Currently, we offer only alpha stream that represents the latest master changes.

### Operator installation Procedure

1. Create catalog.yml file
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: rhoas-operators
  namespace: openshift-marketplace
spec:
  displayName: RHOAS Operators
  icon:
    base64data: ""
    mediatype: ""
  image: quay.io/rhoas/service-operator-registry:autolatest
  priority: -400
  publisher: RHOAS
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 260s
```

1. Ad catalog source

```
oc apply -f catalog.yml
```

1. Wait 1 minute and go to openshift catalog. 
Type rhoas and install operator.

### Verification procedure

1. Go to the Add screen and look for `Managed Service section`.
2. Follow instructions on the wizard until redirected to the topology screen.

## Notes

1. Discovered services are now static (we do not have dynamic capabilities to populate the catalog yet). 
We can create follow up issues for solving those

2. Icons are still temporary

3. Designs are still in progress. 

4. There are still some issues with topology that we would love to get help

5. Catalog integration needs improvements

## Progress

- [x] Base for the plugin 
- [x] Initial design in topology
- [x] Catalog plugin
- [x] Error handling
- [x] Applying proper designs (currently in works)
- [x] Service binding connections (currently hacked - no visuals)

PR #8415, opened 2021-03-18
show kafka source connector

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5573

** Description**: 
show visual connector between KS and KC.

image

ToDo

**Test setup:**
- apply this catalogSource.yaml
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: rhoas-operators
  namespace: openshift-marketplace
spec:
  displayName: RHOAS Operators
  icon:
    base64data: ""
    mediatype: ""
  image: quay.io/rhoas/service-operator-registry:autolatest
  priority: -400
  publisher: RHOAS
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 260s
```
- Wait 1 minute and go to openshift catalog.
- Type rhoas and install operator.
- create KC resource


invincibleJai:
Thanks @vikram-raj, I tried the PR and observed deployments for Kafka source is shown explicitly in topology. We would need to handle that.

image

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -826,6 +839,33 @@ export const getSubscriptionTopologyEdgeItems = (
   return edges;
 };
 
+export const getKnSourceKafkaTopologyEdgeItems = (
+  resource: K8sResourceKind,
+  managedKafkaConnections,
+): EdgeModel[] => {
+  const edges = [];
+  _.forEach(managedKafkaConnections.data, (managedKafkaConnection) => {
```
invincibleJai, line 847: you can reducer here i think

vikram-raj:

wtrocki:
@vikram-raj I would like to verify/try this change as there were a lots of modifications in KafkaConnection recently.

how this can be run together with changes from #8146 ? Is this PR assuming that the other PR is merged first?

wtrocki:
My take would be to drop changes that are common with #8146 and merge this into this PR.

I would like to help to add this to the original PR and verify it end to end but that might be not desired.

frontend/packages/knative-plugin/src/components/overview/KnativeTopologyEdgePanel.tsx
```diff
@@ -24,6 +28,8 @@ const connectorTypeToTitle = (type: string): string => {
       return 'Traffic distribution connector';
     case TYPE_EVENT_SOURCE_LINK:
       return 'Event source connector';
+    case TYPE_KAFKA_CONNECTION_LINK:
+      return 'Kafka source connector';
```
wtrocki, line 32: Technically we can also add support for Strimzi Kafka that have different CR type. Would this be generic Kafka source that can use couple other models or it would be dedicated to KafkaConnection Crs?

I think generic source will be very good and actually simple to add but might be outside scope of the PR.

frontend/packages/knative-plugin/src/models.ts
```diff
@@ -367,3 +367,16 @@ export const CamelIntegrationPlatformModel: K8sKind = {
   color: knativeEventingColor.value,
   badge: BadgeType.TECH,
 };
+
+export const ManagedKafkaConnectionModel: K8sKind = {
```
wtrocki, line 371: This could be removed once we integrate both PR togethe

frontend/packages/knative-plugin/src/topology/components/edges/KafkaConnectionLink.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const KafkaConnectionLink: React.FC<KafkaConnectionLinkProps> = ({
+  element,
+  targetDragRef,
+  children,
+  ...others
+}) => {
+  const resourceObj = getResource(element.getSource());
+  const resourceModel = modelFor(referenceFor(resourceObj));
+  const editAccess = useAccessReview({
```
wtrocki, line 29: Note to myself - our topology SBO integration also need to use accessReview

frontend/packages/knative-plugin/src/topology/components/nodes/KafkaNode.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
wtrocki: Definietly. Other component was updated to use different styles.
We also have drop target defined for SBI

I would see this PR still adding targets for Kafka but in generic way.
Since both plugins might be moving to dynamic plugins at some time I think it is good to apply pattern 
of each plugin having shared API that can be published to NPM.

This way knative plugin can use and import shared API even after we move outside console mono repo.
In this situation we can even benefit from it as we could publish our plugin as package and it can be just imported in this PR. WDYT?

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -826,6 +838,34 @@ export const getSubscriptionTopologyEdgeItems = (
...
 };
 
+export const getKnSourceKafkaTopologyEdgeItems = (
+  kafkaSource: K8sResourceKind,
+  managedKafkaConnections,
+): EdgeModel[] => {
+  const { data } = managedKafkaConnections;
+  const edges = data.reduce((acc, managedKafkaConnection) => {
+    const mkcServiceAccountSecretName =
+      managedKafkaConnection?.spec?.credentials?.serviceAccountSecretName;
```
wtrocki, line 848: I think that we should check state of managedkafka CR before creating or we could show edges as pending if we do not want to do it. Both will be trivial to do once we integrate with original PR

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -826,6 +838,34 @@ export const getSubscriptionTopologyEdgeItems = (
...
+    const kafkaSourcePasswordSecretKeyRefName =
+      kafkaSource.spec?.net?.sasl?.password?.secretKeyRef?.name;
+    const kafkaSourceUserSecretKeyRefName = kafkaSource.spec?.net?.sasl?.user?.secretKeyRef?.name;
+    if (
+      mkcServiceAccountSecretName === kafkaSourcePasswordSecretKeyRefName &&
```
wtrocki, line 853: I think this check for connection can be simplified. 
Maybe use CR metadata instead?

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
wtrocki: Ignore. In current implementation we will always make sure that ManagedKafka is valid

wtrocki:
Screenshot 2021-03-29 at 11 42 01

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
wtrocki: https://user-images.githubusercontent.com/981838/112825544-cc70eb80-9083-11eb-8570-d88985f6528d.png

frontend/packages/knative-plugin/src/components/overview/KnativeTopologyEdgePanel.tsx
vikram-raj: We are not showing `Strimzi Kafka` in the topology. So it is just for `Kafka Source`.

wtrocki:
Let me know when it will be possible to try this

invincibleJai:
I verified the changes and overall it works fine, I observed an edge case
- on drag connector from KS (create flow shouldn't be shown)

image

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
invincibleJai: @wtrocki yeah this is something if we do, will definitely be neat in UI logic but it may not support the use case where we would like to show a connection between KS and KC even if the user created `Kafka source` via CLI with bootstrap server/secrets ( SASL plain) as they may not add the annotations and thus connections won't be shown UI.

@vikram-raj will be good to check if the value is there as well before comparison and bootstrap server matches

wtrocki: Knowing this requirement I would have made those secrets less generic. Something like CloudKafkaConnectionSecret etc. I think we can leave that now

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -944,14 +973,57 @@ const getOwnedEventSourceData = (
   };
 };
 
+const sinkURIDataModel = (resource, resources, data, knDataModel) => {
```
invincibleJai, line 976: nit: we can add type here

jeff-phillips-18 (approved):

vikram-raj:
> I verified the changes and overall it works fine, I observed an issue
> 
> * on drag connector from KS (create flow shouldn't be shown)
> 
> image

I handle this scenario and updated the PR. PTAL again. Thanks.

frontend/packages/knative-plugin/src/topology/components/nodes/EventSource.tsx
```diff
@@ -59,6 +64,24 @@ const EventSource: React.FC<EventSourceProps> = ({
   const { width, height } = element.getBounds();
   const size = Math.min(width, height);
   const { data, resources } = element.getData();
+  const allowEdgeCreation = useAllowEdgeCreation();
+  const isBootstrapServersPresent = resources?.obj?.spec?.bootstrapServers?.length > 0;
```
invincibleJai, line 68: we need to improve this instead of checking just for `bootstrapServers , should check if `bootstrapServers` is of `KC` i.e element is connected to `KC` via Kafka Connector

frontend/packages/knative-plugin/src/topology/const.ts
```diff
@@ -1,7 +1,10 @@
 import { DEFAULT_GROUP_PAD, GROUP_WIDTH } from '@console/topology/src/const';
 
 export const TYPE_EVENT_SOURCE = 'event-source';
+export const TYPE_EVENT_SOURCE_KAFKA = 'knative-source-kafka';
 export const TYPE_EVENT_SOURCE_LINK = 'event-source-link';
+export const TYPE_KAFKA_CONNECTION_LINK = 'event-source-kafka-link';
+export const TYPE_KAFKA_CONNECTION = 'knative-kafka';
```
invincibleJai, line 7: Do we need `TYPE_KAFKA_CONNECTION` now?

frontend/packages/knative-plugin/src/topology/const.ts
```diff
@@ -1,7 +1,10 @@
 import { DEFAULT_GROUP_PAD, GROUP_WIDTH } from '@console/topology/src/const';
 
 export const TYPE_EVENT_SOURCE = 'event-source';
+export const TYPE_EVENT_SOURCE_KAFKA = 'knative-source-kafka';
```
invincibleJai, line 4:
```suggestion
export const TYPE_EVENT_SOURCE_KAFKA = 'event-source-kafka';
```

frontend/packages/knative-plugin/src/topology/create-connector-utils.ts
```diff
@@ -62,7 +68,21 @@ const createPubSubConnector = (source: Node, target: Node) => {
   ).then(() => null);
 };
 
+const createKafkaConnection = (source, target) =>
```
invincibleJai, line 71: we can add type here

frontend/packages/knative-plugin/src/topology/topology-types.ts
```diff
@@ -9,12 +9,15 @@ export enum NodeType {
   Revision = 'knative-revision',
   PubSub = 'event-pubsub',
   SinkUri = 'sink-uri',
+  KnSourceKafka = 'knative-source-kafka',
```
invincibleJai, line 12:
```suggestion
  EventSourceKafka = 'Event-source-kafka',
```

frontend/packages/rhoas-plugin/src/topology/components/rhoasComponentUtils.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '@console/topology/src/components/graph-view';
+
+const getKafkaConnectionTooltip = (monitor): string => {
+  return monitor.getOperation()?.type === MOVE_EV_SRC_KAFKA_CONNECTOR_OPERATION ||
+    monitor.getOperation()?.type === CREATE_EV_SRC_KAFKA_CONNECTOR_OPERATION
+    ? 'Create a Kafka connector'
+    : 'Create a binding connector';
+};
+
+export const obsORKafkaConnectionDropTargetSpec = (
```
invincibleJai, line 27:
```suggestion
export const obsOrKafkaConnectionDropTargetSpec = (
```

frontend/packages/rhoas-plugin/src/topology/components/rhoasComponentUtils.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (isEdge(item)) {
+      return canDropEdgeOnNode(monitor.getOperation()?.type, item, props.element);
+    }
+    if (item === props.element) {
+      return false;
+    }
+    return (
+      !props.element.getTargetEdges().find((e) => e.getSource() === item) ||
+      item.getType() === TYPE_KAFKA_CONNECTION_LINK ||
+      monitor.getOperation()?.type === CREATE_EV_SRC_KAFKA_CONNECTOR_OPERATION
```
invincibleJai, line 46: do we need this?

frontend/packages/topology/src/components/graph-view/components/componentUtils.ts
```diff
@@ -226,7 +227,9 @@ const graphDropTargetSpec: DropTargetSpec<
   collect: (monitor) => {
     const operation = monitor.getOperation() as EditableDragOperationType;
     const dragEditInProgress =
-      monitor.isDragging() && (operation?.type === CREATE_CONNECTOR_OPERATION || operation?.edit);
+      monitor.isDragging() &&
+      (operation?.type === CREATE_CONNECTOR_OPERATION || operation?.edit) &&
+      operation?.type !== CREATE_EV_SRC_KAFKA_CONNECTOR_OPERATION;
```
invincibleJai, line 232: @jeff-phillips-18  Does this looks fine, we added it here as we needed to hide + actin on the drag of connector cc @vikram-raj

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -1136,6 +1200,35 @@ export const createEventingPubSubSink = (subObj: K8sResourceKind, target: K8sRes
...
+  }
+  const sourceObj = getResource(source);
+  const targetObj = getResource(target);
+  const mkcBoostrapServer = targetObj?.status?.bootstrapServerHost;
+  const mkcServiceAccountSecretName = targetObj?.spec?.credentials?.serviceAccountSecretName;
+  const knKafkaSourceObj = _.omit(sourceObj, 'status');
+  const updatedObjPayload = {
+    ...knKafkaSourceObj,
+    spec: {
+      bootstrapServers: [mkcBoostrapServer],
```
invincibleJai, line 1218:
```suggestion
      ...knKafkaSourceObj.spec,
      bootstrapServers: [mkcBoostrapServer],
```

frontend/packages/topology/src/components/graph-view/components/componentUtils.ts
jeff-phillips-18: I would have expected a `canDrop` check or something here. No point setting `dragEditInProgress` if you can't drop on the graph. Should be a better way than checking specifically for that type.

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -826,6 +827,37 @@ export const getSubscriptionTopologyEdgeItems = (
   return edges;
 };
 
+export const getKnSourceKafkaTopologyEdgeItems = (
+  kafkaSource: K8sResourceKind,
+  kafkaConnections,
```
karthikjeeyar, line 832: Do we have a type for kafkaConnection that can be used here?

karthikjeeyar:
Verified the PR locally, works as expected

frontend/packages/knative-plugin/src/topology/__tests__/knative-topology-utils.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    const expectedEdges = [
+      {
+        id: '1317f615-9636-11e9-b134-06a61d886b689_1_85ffdf52-59f5-4120-b2cd-e90c991845e0',
+        type: 'event-source-kafka-link',
+        source: '1317f615-9636-11e9-b134-06a61d886b689_1',
+        target: '85ffdf52-59f5-4120-b2cd-e90c991845e0',
+      },
+    ];
+    const edges = getKnSourceKafkaTopologyEdgeItems(
+      getEventSourceResponse(EventSourceCamelModel).data[0],
```
invincibleJai, line 18: this should be kafka model

invincibleJai:
/lgtm

Thanks @vikram-raj , Verified below scenarios works as expected

- Create KC with the complete flow in NS (if possible have one more KC)
- Create 2 KSVC with logger service image

**Scenario1:**
- now Create Kafka Source create form and provide bootstrap host, topics, SASL secret, TLS enabled (based on what's in KC) and sink to KSVC. On create should see KC connected to KS and  KS to KSVC
- Hover on KS should not show create connector as KS already has been connected to KC however user should be able to perform drag and drop action on another KC for Kafka Connector
- In ksvc pod logs can see events emitted in logs

**Scenario 2:**
- Create Kafka Source create form and provide bootstrap host, topics (any random value) and sink to KSVC. On create should see KS connected to KSVC  and not to KC
- Now hover on KC should see the connector, should be able to drag the connector and valid drop only be KC.
- On drop verify the network call going properly i.e KS spec getting updated as expected. The call will fail now KS is not mutable (which will be fixed in upcoming release 1.14 of the serverless operator)

**Scenario 3:**
- Verify existing SBO with KC is not impacted
- Provision Node.js Sample
- Draw an arrow to KC while having KafkaSource connection
- Do opposite flow - have SBO first and connect to KafkaSource

image

PR #8416, opened 2021-03-18
Cypress: Fix OLM test flakes

dtaylor113:
- consolidated non-independant `it` tests, into single test
- Added check for loading indicators to not exist
- Add check for url to contain '~new`

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
@@ -9,7 +9,6 @@ const operatorID = 'couchbase-enterprise-console-e2e-openshift-marketplace';
 const operatorRow = 'Couchbase Operator';
 const operatorPkgName = 'couchbase-enterprise';
 const operatorInstallFormURL = `/operatorhub/subscribe?pkg=${operatorPkgName}&catalog=${catalogSourceName}&catalogNamespace=openshift-marketplace&targetNamespace=${testName}`;
-const operatorInstance = 'CouchbaseCluster';
 const operandLink = 'cb-example';
 
 describe(`Interacting with a single namespace install mode Operator (${operatorName})`, () => {
@@ -43,8 +42,8 @@ describe(`Interacting with a single namespace install mode Operator (${operatorN
   });
 
   it(`creates the single namespace install mode ClusterServiceVersion for ${operatorName}`, () => {
-    cy.visit(operatorInstallFormURL);
     cy.log('configure Operator install form for single namespace');
+    cy.visit(operatorInstallFormURL);
     cy.byTestID('A specific namespace on the cluster-radio-input').check();
     cy.log(`verify the dropdown selection shows the ${testName} namespace`);
     cy.byTestID('dropdown-selectbox').should('contain', `${testName}`);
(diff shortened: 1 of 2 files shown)


spadgett:
Based on the video I'm looking at for at least one failure, the form is not loading. I think this might be a legitimate bug rather than simply a test flake.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/batch/pull-ci-openshift-console-master-e2e-gcp-console/1372889154475528192/artifacts/e2e-gcp-console/test/artifacts/gui_test_screenshots/cypress/videos/operator-install-single-namespace.spec.ts.mp4

rhamilto:
> * consolidated non-independant `it` tests, into single test

we should also do this for https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-global.spec.ts so they're consistent?

dtaylor113:
> Based on the video I'm looking at for at least one failure, the form is not loading.

Might be related to this: https://stackoverflow.com/questions/51327713/why-cypress-not-open-modal-page
I'll check if our modals are using annimations/transitions

dtaylor113:
@rhamilto & @spadgett, I noticed from the video and running it locally that after clicking on the 'Create ...' the loading dots appear for a split second which might be causing the issue.  Added a check to wait for loading dots to not exist.

Also updated `operator-install-global.spec` to reduce to independant `it` tests.

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
```diff
@@ -58,42 +58,31 @@ describe(`Interacting with a single namespace install mode Operator (${operatorN
     cy.byTestID('view-installed-operators-btn').click();
     cy.log(`verify the ClusterServiceVersion row for ${operatorRow} exists`);
     cy.byTestOperatorRow(operatorRow, { timeout: 60000 }).should('exist');
-  });
-
-  it(`displays details about ${operatorName} ClusterServiceVersion on the "Details" tab`, () => {
```
rhamilto, line 63: In the global test, you migrated these to a `cy.log`, which I like.  Do the same here?

dtaylor113: Yes, but in most cases the `it` was followed immediately by a `cy.log`, ie:
```
it(`displays empty message on the ${operatorName} ClusterServiceVersion "All Instances" tab`, () => {
    cy.log('navigate to the "All Instances" tab');
```
In these cases, I tended to use the `cy.log`, because it was concise, whereas the `it` I feel is verbose, and a new overall `it` had been created.

dtaylor113:
Hmm...another type of failure/flake?  cc @rhamilto -thanks
**Global**
image
**Single**
image

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
rhamilto: If that's the case, you don't need `operatorInstance` as a `const` since it is only used once.

dtaylor113:
/test all
passed: 1

dtaylor113:
/test all
passed: 2

frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/operator-install-single-namespace.spec.ts
dtaylor113: fixed, ty

dtaylor113:
@rhamilto & @spadgett, still seeing [OLM flakes](https://search.ci.openshift.org/?search=AssertionError&maxAge=168h&context=1&type=junit&name=e2e-gcp-console&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job).   I'd like to merge this PR to see if it cuts down on the flakes.

rhamilto:
Ok by me.  @spadgett?

spadgett:
> @rhamilto & @spadgett, still seeing [OLM flakes](https://search.ci.openshift.org/?search=AssertionError&maxAge=168h&context=1&type=junit&name=e2e-gcp-console&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job). I'd like to merge this PR to see if it cuts down on the flakes. Consolidating `it` tests will at least do a 'retry' from a correct start state.

Hey, @dtaylor113. My concern is: I don't think they are flakes. It looks like a legitimate bug to me. If we improve the retries so that the tests stop failing, that just covers up the problem instead of fixing it, and we lose the only way we have to reproduce currently (CI). I think we need to debug what's happening here.

If we're sure it's a test flake, OK. But I'm not convinced that's the case.

dtaylor113:
> If we're sure it's a test flake, OK. But I'm not convinced that's the case.

I've never been able to reproduce in local cypress runs, nor have I've seen this happen in the actual concole.

dtaylor113:
closing in favor of addressing [BZ 1976939](https://bugzilla.redhat.com/show_bug.cgi?id=1976939) and [BZ 1966499](https://bugzilla.redhat.com/show_bug.cgi?id=1966499) which deal with rewriting the OLM tests to use Red Hat Operators instead of Community Operators.

PR #8418, opened 2021-03-19
Query Browser: Fix cutoff for when to show X axis seconds

kyoto:
https://github.com/openshift/console/pull/8365 changed the number of X axis ticks, so we need to change this to match.


spadgett (approved):
/lgtm

PR #8423, opened 2021-03-19
Fix for creating visual connectors to operator workload nodes

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5621

**Analysis / Root cause**: 
The create service binding connection function was accepting any targets as valid as long as the drop hints included `serviceBinding`. It has an old check for a hint over a operator workload but those are not used any more.

**Solution Description**: 
Pass source and target to the create connection functions and check target type in the create service binding connection function.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


rohitkrai03 (approved):
/approve
/lgtm

PR #8429, opened 2021-03-22
Monitoring: Fix list page default sort

kyoto:
The default sort was not always being applied.

Also changes the component to use `useSelector` instead of `connect` and replaces uses of `store.dispatch` with `useDispatch`.


spadgett (approved):
/lgtm

PR #8432, opened 2021-03-22
Bug 1935176: Indicate single node deployment on cluster overview page

rawagner:
Screenshot from 2021-03-22 11-43-15

### DIFF
--- a/frontend/packages/console-app/locales/en/console-app.json
+++ b/frontend/packages/console-app/locales/en/console-app.json
@@ -11,6 +11,7 @@
   "Components of the Control Plane are responsible for maintaining and reconciling the state of the cluster.": "Components of the Control Plane are responsible for maintaining and reconciling the state of the cluster.",
   "Components": "Components",
   "Response rate": "Response rate",
+  "Single master": "Single master",
   "Incompatible file type": "Incompatible file type",
   "{{fileName}} cannot be uploaded. Only {{fileExtensions}} files are supported currently. Try another file.": "{{fileName}} cannot be uploaded. Only {{fileExtensions}} files are supported currently. Try another file.",
   "Clone": "Clone",
--- a/frontend/packages/console-app/src/components/dashboards-page/status.ts
+++ b/frontend/packages/console-app/src/components/dashboards-page/status.ts
@@ -25,6 +25,7 @@ import { PrometheusResponse } from '@console/internal/components/graphs';
 import { humanizePercentage } from '@console/internal/components/utils/units';
 import { getOperatorsStatus } from '@console/shared/src/components/dashboard/status-card/state-utils';
 import { pluralize } from '@patternfly/react-core';
+import { isSingleNode } from '@console/shared/src/selectors/infrastructure';
 
 export const fetchK8sHealth = async (url: string) => {
   const response = await coFetch(url);
@@ -89,7 +90,12 @@ export const getWorstStatus = (
   };
 };
 
-export const getControlPlaneHealth: PrometheusHealthHandler = (responses, t) => {
+export const getControlPlaneHealth: PrometheusHealthHandler = (
+  responses,
+  t,
+  resource,
+  infrastructure,
+) => {
   const componentsHealth = responses.map(({ response, error }) =>
     getControlPlaneComponentHealth(response, error, t),
   );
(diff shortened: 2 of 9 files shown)


frontend/packages/console-app/src/plugin.tsx
```diff
@@ -136,8 +136,12 @@ const plugin: Plugin<ConsumedExtensions> = [
   {
     type: 'Dashboards/Overview/Health/Prometheus',
     properties: {
+      // t('console-app~Control Plane (single master)')
       // t('console-app~Control Plane')
-      title: '%console-app~Control Plane%',
+      getTitle: (infrastructure) =>
+        infrastructure?.status?.infrastructureTopology === 'SingleReplica'
```
romfreiman, line 142: ok, it was bad example. 
there is also this part: 
https://github.com/openshift/origin/pull/25936/files#diff-cb149e746da7ca315f8c9b852dd6f407e04e16a7199587be3ae745fa57fe681eR64

I guess for now just check both, and if both equals to 1, then SNO, or, maybe both equals 1 and master=1?

romfreiman: or, maybe, if u write about control plane, so jsut controlTopology = 'SingleReplica' is good enough?

nirfarkas:
LGTM! Thanks!

frontend/packages/console-app/src/plugin.tsx
rawagner: @romfreiman added check for both `infrastructureTopology` && `controlPlaneTopology`. If you think we need to check the count of master nodes too, let me know.

eranco74: I think the only check we need is `controlPlaneTopology == 'SingleReplica'`.
The number of workers isn't relevant for the control plane high availability indication.

andybraren:
Reading through [the Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1935176) and [Slack discussion](https://coreos.slack.com/archives/C018KQE33MF/p1616061395018800?thread_ts=1614197690.120300&cid=C018KQE33MF), this small tweak might A) reduce redundancy between the Details, Inventory, and Status cards B) preserve a bit more horizontal space in the Status card (which is at a premium these days 😬) and C) convey what the user needs to know more clearly and with fewer words.

console-8432-control-plane

@nirfarkas WDYT? This is similar to some of your earlier ideas in the thread to use/enhance what exists already, but utilizes the existing Status subtext pattern. Ultimately it seems like the thing users need to know is that their cluster's Control Plane isn't highly-available in an SNO (or whatever) configuration, and this puts that info in one spot.

Side note: if we ever combine the Cluster and Control Plane statuses someday (I think we could?) this subtext could probably apply to the combined `Cluster` status as well.

nirfarkas:
@andybraren Thanks for that! Didn't know the prices have gone high for “pixel-state” in the status card.
I think that the real question here is the hierarchy: Single-Node vs. non-high availability. If the availability is more prominent, then we should go with your first design, but if SNO is more important, then it's either the second design or my suggestion.

If both are important, I guess we might need to keep the non-high availability in the details card as well and tweak the space the "Single master" takes in the status card.

andybraren:
> I think that the real question here is the hierarchy: Single-Node vs. non-high availability. If the availability is more prominent, then we should go with your first design, but if SNO is more important, then it's either the second design or my suggestion.

Agreed! I suspect the outcome/result is the thing that's more key to know. It's kinda like saying:

"The control plane is not highly available... because you're in a SNO configuration so there's only one master node and blah blah..."

which starts with the outcome, vs:

"The control plane is on a single master... what that means is that it's not highly available, which we mentioned over there in the Details card too"

> If both are important, I guess we might need to keep the non-high availability in the details card as well and tweak the space the "Single master" takes in the status card.

I've seen recorded demos of SNO where folks point their mouse at the Inventory card to show that there's only 1 node. It's cool and exciting now for sure, but may eventually become pretty common and boring. Giving SNO extra ad space in the Status/Details card doesn't quite seem needed to me, at least. 🤷‍♂️

nirfarkas:
@andybraren the control plane is a single master has also other implications:
image
I think that Moran & Tomas wanted to emphasize here the in-a-glance identification that this is SNO (I'll ask again) & also the non-high availability of the control plane. I'll discuss this with Moran and Tomas and present your feedback (👍 ) Thanks!

jelkosz:
after an offline discussion with Moran it seems the current implementation is fits the requirements better since "control plane: single master" is clearer and more positive than "not ha" and that the details are explained in the details card.

jtomasek (approved):

rawagner:
the `Single master` message is now under the `Control plane` text

Screenshot from 2021-04-06 15-19-35

nirfarkas:
Perfect!

PR #8437, opened 2021-03-22
Bug 1941995: fix backwards incompatible trigger api changes

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5405

**Analysis / Root cause**: 

Current logic of adding trigger from UI using `template.name` in Eventlisteners is a BREAKING change in 1.4 Pipelines operator, therefore update the logic to support new trigger Api and use `triggerRef` while creating the Trigger resources.

**Solution Description**: 

New flow for creating the eventlistener via triggerRef is added based on the availability of `Trigger` resource in the cluster (availability of the resource is determined using `dryRun` option), in case if the trigger resource is not available the control will fallback to the old flow (template.name).

NO UI changes

packages/pipelines-plugin/src/components/pipelines/modals/triggers/__tests__/submit-utils.spec.ts (8.609s)
  submitTrigger
  
    ✓ expect to run the new flow when trigger resource dry run results in success (20ms)
    ✓ expect to the eventlistener to contain triggerRef (3ms)
    ✓ expect to fallback to the old flow when trigger resource dry run results in not found (2ms)

packages/pipelines-plugin/src/components/pipelines/utils/__tests__/pipeline-operator.spec.ts

    ✓ should return the installed version for the old name of pipeline ClusterServiceVersion (1ms)
    ✓ should return installed version for the new name of pipeline ClusterServiceVersion (1ms)
    
**Test setup:**

Install 1.3.1 or 1.4 OSP operator.

1. Add a pipeline through git import flow
2. Add trigger from pipelines list view
3. Check if the Trigger, Trigger Template, EventListener is created and in pipeline's details page the url should be listed inside copy to clipboard component.

In Operators < 1.3.1,
1. Adding Trigger should create TriggerTemplate and Eventlistener only.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @andrewballantyne @jerolimov @vdemeester

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
@@ -1,7 +1,7 @@
 import { getRandomChars } from '@console/shared';
-import { apiVersionForModel, RouteKind } from '@console/internal/module/k8s';
+import { apiVersionForModel, k8sCreate, RouteKind } from '@console/internal/module/k8s';
 import { RouteModel } from '@console/internal/models';
-import { EventListenerModel, TriggerTemplateModel } from '../../../../models';
+import { EventListenerModel, TriggerModel, TriggerTemplateModel } from '../../../../models';
 import { PipelineKind, PipelineRunKind } from '../../../../types';
 import { PIPELINE_SERVICE_ACCOUNT } from '../../const';
 import { getPipelineOperatorVersion } from '../../utils/pipeline-operator';
@@ -9,6 +9,7 @@ import {
   EventListenerKind,
   EventListenerKindBindingReference,
   TriggerBindingKind,
+  TriggerKind,
   TriggerTemplateKind,
   TriggerTemplateKindParam,
 } from '../../resource-types';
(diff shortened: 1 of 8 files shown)


frontend/packages/pipelines-plugin/src/components/pipelines/utils/pipeline-operator.ts
```diff
@@ -13,7 +13,8 @@ export const getPipelineOperatorVersion = async (namespace: string): Promise<Sem
   });
   const matchingCSVs = allCSVs.filter(
     (csv) =>
-      csv.metadata?.name?.startsWith('openshift-pipelines-operator') &&
+      (csv.metadata?.name?.startsWith('redhat-openshift-pipelines') ||
+        csv.metadata?.name?.startsWith('openshift-pipelines-operator')) &&
```
karthikjeeyar, line 17: From OSP 1.2.x onwards the CSV name is changed from  `openshift-pipelines-operator` to `redhat-openshift-pipelines`, removed `operator` suffix and added `redhat` prefix. I have updated this function so that it can be used in subsequent 1.4 fix related PRs.

frontend/packages/pipelines-plugin/src/components/pipelines/utils/triggers.ts
```diff
@@ -73,6 +85,20 @@ const useAllEventListeners = (namespace: string) => {
...
+    () => ({
+      kind: referenceForModel(TriggerModel),
+      isList: true,
+      namespace,
+    }),
+    [namespace],
+  );
+  const [resources, triggerLoaded] = useK8sWatchResource<EventListenerKind[]>(triggerResource);
+
+  return triggerLoaded ? resources : null;
```
invincibleJai, line 99: should check err as well??

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
```diff
@@ -31,6 +32,58 @@ export const createTriggerTemplate = (
...
+    .then(() => true)
+    .catch((e) => e?.response?.status !== 404);
+};
+
+export const createTrigger = (
+  namespace: string,
+  triggerTemplateRef: string,
+  triggerBindings: TriggerBindingKind[],
+): TriggerKind => {
+  return {
```
invincibleJai, line 49: nit: can do implicit return here and below

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/submit-utils.ts
```diff
@@ -68,28 +71,47 @@ export const submitTrigger = async (
...
   );
-  const eventListener: EventListenerKind = await createEventListener(
-    thisNamespace,
-    [triggerBinding.resource],
-    triggerTemplate,
-  );
-
   const metadata = { ns: thisNamespace };
   let resources: K8sResourceKind[];
   try {
```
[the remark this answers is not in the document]
karthikjeeyar: Updated the order and fixed DRY issue.

andrewballantyne (approved):
As discussed with Karthik on Slack. We are doing a bit too much in this PR to fully backport this change. We should make our 4.7.z stream fix to be a switch from `template.name` to `template.ref` in the `EventListener`'s `spec.templates[]` instead of adopting a new resource (Triggers).

Since we need a ticket for 4.8 that says "this issue" is fixed, I am okay with the BZ being tied to this. We just can't backport the change as-is.

frontend/packages/pipelines-plugin/src/components/pipelines/utils/pipeline-operator.ts
karthikjeeyar: This code block got merged in my metrics endpoint pr, so I have removed this from this PR to resolve conflicts.

vikram-raj (commented):
Verified the changes with both 1.3 and 1.4 pipeline operators and it's working as expected.

/lgtm

PR #8438, opened 2021-03-22
converting navigation extensions to dynamic-plugin-sdk

glekner:


glekner:
@christianvogt @vojtechszocs please review.

I added the field 
```js
    /** if true, adds /k8s/ns/active-namespace to the begining */
    addResourcePrefix?: boolean;
```
to `HrefNavItem`.

Kubevirt VM/VMT `/virtualization` uses it, and this page doesn't handle namespaces like the dev perspective views.
since there is no model for `virtualization` but it is namespaced, we need this field.

rohitkrai03:
> @christianvogt @vojtechszocs @rawagner please review.
> 
> I added the field
> 
> ```js
>     /** if true, adds /k8s/ns/active-namespace to the begining */
>     addResourcePrefix?: boolean;
> ```
> 
> to `HrefNavItem`.
> 
> Kubevirt VM/VMT `/virtualization` uses it, and this page doesn't handle namespaces like the dev perspective views.
> since there is no model for `virtualization` but it is namespaced, we need this field.

https://github.com/openshift/console/pull/8249 seems to be solving the same problem.

glekner:
Took changes from #8249 @jerolimov 
updated to `namespaced` and `isResource` for `HrevNavLink`

christianvogt:
@glekner please rebase and let's get this reviewed as we held up https://github.com/openshift/console/pull/8249 for this

frontend/public/components/nav/items.tsx
```diff
@@ -203,13 +234,11 @@ export const createLink = (item: PluginNavItem, rootNavLink = false): React.Reac
       Component = ResourceClusterLink;
     }
     if (Component) {
-      const { id } = item.properties;
-      const key = item.properties.componentProps.name;
-      const props = item.properties.componentProps;
+      const { id, name, ...props } = item.properties;
       if (rootNavLink) {
-        return <RootNavLink id={id} {...props} key={key} component={Component} />;
+        return <RootNavLink name={name} id={id} key={name} component={Component} {...props} />;
```
rawagner, line 239: as a `key` we should use `item.properties.uid`.

we actually pass LoadedExtension<GenericNavItem> to `createLink` so `uid` prop is accessible

rawagner (changes requested):

frontend/public/components/nav/section.tsx
```diff
@@ -192,7 +188,7 @@ export const NavSection = connect(navSectionStateToProps)(
           const childItems = sortExtensionItems(this.getNavItemExtensions(perspective, title, id));
 
           childItems.forEach((item) => {
-            const pluginChild = this.mapChild(createLink(item as NavItem));
+            const pluginChild = this.mapChild(createLink(item as LoadedExtension<NavItem>));
```
rawagner, line 191: if you change `sortExtensionItems` to
```js
export const sortExtensionItems = <E extends NavItem>(
  extensionItems: LoadedExtension<E>[],
): LoadedExtension<E>[] => {
...
};
```

you will not have to cast to `LoadedExtension<NavItem>` anymore

frontend/public/module/k8s/k8s.ts
```diff
@@ -1,6 +1,5 @@
 import * as _ from 'lodash-es';
 import { ExtensionK8sModel } from '@console/dynamic-plugin-sdk/src/api/common-types';
-
```
rawagner, line 3: lets remove this change :)

frontend/packages/metal3-plugin/src/plugin.tsx
```diff
@@ -21,7 +20,7 @@ import { referenceForModel } from '@console/internal/module/k8s';
 import { MachineModel, NodeModel, CertificateSigningRequestModel } from '@console/internal/models';
 // TODO(jtomasek): change this to '@console/shared/src/utils' once @console/shared/src/utils modules
 // no longer import from @console/internal (cyclic deps issues)
-import { formatNamespacedRouteForResource } from '@console/shared/src/utils/namespace';
+// import { formatNamespacedRouteForResource } from '@console/shared/src/utils/namespace';
```
rawagner, line 23: can be removed

rawagner (changes requested):

PR #8439, opened 2021-03-22
Bug 1939753: Track and show error messages in modals

rebeccaalpert:
Error messages weren't being passed to the footer in a few modals. I refactored them to catch and show the error messages.

Screen Shot 2021-03-22 at 1 09 53 PM

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1939753.

### DIFF
--- a/frontend/public/components/modals/delete-modal.tsx
+++ b/frontend/public/components/modals/delete-modal.tsx
@@ -27,7 +27,6 @@ import { ResourceLink } from '../utils/resource-link';
 const DeleteModal = withHandlePromise((props: DeleteModalProps) => {
   const [isChecked, setIsChecked] = React.useState(true);
   const [owner, setOwner] = React.useState(null);
-  const [errorMessage] = React.useState(null);
 
   const { t } = useTranslation();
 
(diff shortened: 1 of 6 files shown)


frontend/public/components/modals/delete-modal.tsx
```diff
@@ -24,10 +19,11 @@ import { k8sList } from '../../module/k8s/resource';
 import { ResourceLink } from '../utils/resource-link';
 
 //Modal for resource deletion and allows cascading deletes if propagationPolicy is provided for the enum
-const DeleteModal = withHandlePromise((props: DeleteModalProps) => {
```
spadgett, line 27: Is `withHandlePromise` not working? We use this in a lot of places, so I'm worried it's a problem elsewhere.

rebeccaalpert: It seems to handle the promise okay, but I couldn't get error handling to work with it.

rebeccaalpert: withHandlePromise looked ok. I think I figured out the real issue.

spadgett (approved):
/lgtm

PR #8271, opened 2021-03-01
Bug 1933857: Fix uncaught exception in DetailsPage component

TheRealJon:
Update `useBreadCrumbsForDetailPage` to use optional chaning when referencing the `kindObj` argument, as it may be undefined in some cases. Also updated `modelFor` usage to `useK8sModel` hook.

PR #8443, opened 2021-03-22
Bug 1933857: Follow on fix to ensure operand details page 404s when no model is found

TheRealJon:
Follow on for #8271:
- Adjust the `useK8sModel` hook to return `undefined` instead of `{}` when no model is found. 
- Update all uses of `useK8sModel` hook to make sure `undefined` result is handled.
- Render 404 error on operand details page when CRD doesn't exist (no model is present in the redux store).


spadgett (approved):
/lgtm

TheRealJon:
/hold

Looks like a legit test failure, looking into it.

TheRealJon:
/hold cancel

Unit test needed to be updated.

spadgett (approved):
/lgtm

PR #8445, opened 2021-03-23
Bug 1941990: fix pipeline metrics endpoint for 1.4 osp

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5667

**Analysis / Root cause**: 

Pipeline Metrics is not showing any data in 1.4 OSP as the metrics endpoint is changed in upstream operator

**Solution Description**: 

Conditionally forming the metrics endpoint using the pipeline operator installed.

Before fix:
image

After fix:
image

**Test setup:**

Pipeline Metrics should work in both 1.3.1 and as well as in 1.4 Openshift pipelines operator.

Unit tests:
```

  pipelineDetailsPage.spec.tsx
    ✓ should have the latest metrics endpoint as default queryPrefix (5ms)
    ✓ should use the new metrics endpoint if the pipeline operator is greater than 1.4.0 (3ms)
    ✓ should use the old metrics endpoint if the pipeline operator is less than 1.4.0 (3ms)
```
```
packages/pipelines-plugin/src/components/pipelines/utils/__tests__/pipeline-operator.spec.ts
  isGAVersionInstalled
    ✓ should return false if the operator is not identified (3ms)
    ✓ should return true if the installed operator is below 1.4.0 (1ms)
    ✓ should return true if the installed operator is above 1.4.0
```

/assign @jerolimov 
/kind bug
cc: @andrewballantyne @VeereshAradhya

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineDetails.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineDetails.tsx
@@ -3,22 +3,13 @@ import { useTranslation } from 'react-i18next';
 import { SectionHeading, ResourceSummary } from '@console/internal/components/utils';
 import { TriggerTemplateModel } from '../../../../models';
 import WorkspaceDefinitionList from '../../../shared/workspaces/WorkspaceDefinitionList';
-import { PipelineKind } from '../../../../types';
-import { RouteTemplate } from '../../utils/triggers';
 import DynamicResourceLinkList from '../../resource-overview/DynamicResourceLinkList';
 import TriggerTemplateResourceLink from '../../resource-overview/TriggerTemplateResourceLink';
 import PipelineVisualization from './PipelineVisualization';
 import { getPipelineTaskLinks } from '../utils';
+import { PipelineDetailsTabProps } from '../types';
 
-interface PipelineDetailsProps {
-  obj: PipelineKind;
-  customData: RouteTemplate[];
-}
-
-const PipelineDetails: React.FC<PipelineDetailsProps> = ({
-  obj: pipeline,
-  customData: routeTemplates,
-}) => {
+const PipelineDetails: React.FC<PipelineDetailsTabProps> = ({ obj: pipeline, customData }) => {
   const { t } = useTranslation();
   const { taskLinks, finallyTaskLinks } = getPipelineTaskLinks(pipeline);
 
@@ -34,7 +25,7 @@ const PipelineDetails: React.FC<PipelineDetailsProps> = ({
           <TriggerTemplateResourceLink
             namespace={pipeline.metadata.namespace}
             model={TriggerTemplateModel}
-            links={routeTemplates}
+            links={customData.templateNames}
           />
           <DynamicResourceLinkList
             namespace={pipeline.metadata.namespace}
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetrics.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetrics.tsx
@@ -6,7 +6,6 @@ import DashboardCardBody from '@console/shared/src/components/dashboard/dashboar
 import DashboardCardHeader from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardHeader';
 import DashboardCardTitle from '@console/shared/src/components/dashboard/dashboard-card/DashboardCardTitle';
 import { parsePrometheusDuration } from '@console/internal/components/utils/datetime';
-import { PipelineKind } from '../../../types';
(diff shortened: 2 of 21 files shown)


invincibleJai:
/approve

Thanks looks good and have verified the changes locally and works as expected with operator 1.3/1.4

frontend/packages/pipelines-plugin/src/components/pipelines/utils/pipeline-operator.ts
```diff
@@ -1,19 +1,21 @@
...
 
 export const getPipelineOperatorVersion = async (namespace: string): Promise<SemVer | null> => {
   const allCSVs: ClusterServiceVersionKind[] = await k8sList(ClusterServiceVersionModel, {
     ns: namespace,
   });
   const matchingCSVs = allCSVs.filter(
     (csv) =>
-      csv.metadata?.name?.startsWith('openshift-pipelines-operator') &&
+      (csv.metadata?.name?.startsWith('openshift-pipelines-operator') ||
+        csv.metadata?.name?.startsWith('redhat-openshift-pipelines')) &&
```
invincibleJai, line 18: is it required to check on name?

karthikjeeyar: Yes, this needs to be handled.

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineDetails.tsx
```diff
@@ -12,13 +12,12 @@ import { getPipelineTaskLinks } from '../utils';
 
 interface PipelineDetailsProps {
   obj: PipelineKind;
-  customData: RouteTemplate[];
+  customData: {
+    templateNames: RouteTemplate[];
+  };
```
andrewballantyne, line 17: I know Query Prefix isn't used in this component, but it would seem we have a common type that should be written somewhere and then referenced so the data remains available and common during refactors.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetrics.tsx
```diff
@@ -21,12 +21,16 @@ import './PipelineMetrics.scss';
 
 interface PipelineMeticsProps {
   obj: PipelineKind;
+  customData: {
+    queryPrefix: string;
+  };
 }
 
-const PipelineMetrics: React.FC<PipelineMeticsProps> = ({ obj }) => {
+const PipelineMetrics: React.FC<PipelineMeticsProps> = ({ obj, customData }) => {
```
andrewballantyne, line 29: The types will probably all be the same, perhaps we even go so far as going to:
```suggestion
const PipelineMetrics: React.FC<PipelineDetailsTabProps> = ({ obj, customData }) => {
```

And go with the common type:
```
type PipelineDetailsTabProps = {
  obj: PipelineKind;
  customData: {
    templateNames: RouteTemplate[];
    queryPrefix: string;
  };
};
```

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineMetrics.tsx
andrewballantyne: Could you also get all the `Pipeline` details page tabs to be this type too, even if you didn't touch them.

karthikjeeyar: Having a common type for all the pipeline details tabs.

andrewballantyne:
@karthikjeeyar You have two type errors:

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/PipelineRuns.spec.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/PipelineRuns.spec.tsx(7,3):
TS2322: Type '{ metadata: { name: string; }; }' is not assignable to type 'PipelineKind'.
  Property 'spec' is missing in type '{ metadata: { name: string; }; }' but required in type '{ spec: PipelineSpec; }'.
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/__tests__/PipelineMetrics.spec.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/__tests__/PipelineMetrics.spec.tsx(43,7):
TS2741: Property 'templateNames' is missing in type '{ queryPrefix: MetricsQueryPrefix; }' but required in type '{ templateNames: RouteTemplate[]; queryPrefix: string; }'.
```

andrewballantyne:
LGTM, once you fix your type errors we should be good to go.

karthikjeeyar:
fixed the type errors :)

PR #8446, opened 2021-03-23
Bug 1939606: Attempting to put a host into maintenance mode warns about Ceph cluster health, but no storage cluster problems are apparent

SanjalKatiyar:
**Earlier:**
In bare metal OCP cluster, attempting to put a node into maintenance mode was resulting in the warning-alert even though there were no problems with the Ceph cluster.

**After:**
There will be no alert if the status of the Ceph cluster is `HEALTH_OK`

### DIFF
--- a/frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
+++ b/frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
@@ -41,7 +41,7 @@ const StartNodeMaintenanceModal = withHandlePromise<StartNodeMaintenanceModalPro
 
   const [cephClusters, loaded] = useK8sWatchResource<K8sResourceKind[]>(cephClusterResource);
   const cephCluster = cephClusters?.[0];
-  const cephClusterHealthy = !cephCluster || cephCluster?.status?.health === 'OK';
+  const cephClusterHealthy = cephCluster?.status?.ceph?.health === 'HEALTH_OK';
 
   const action = t('metal3-plugin~Start Maintenance');
   return (
@@ -70,7 +70,7 @@ const StartNodeMaintenanceModal = withHandlePromise<StartNodeMaintenanceModalPro
               />
             </FormGroup>
           </StackItem>
-          {!cephClusterHealthy && (
+          {!!cephCluster && !cephClusterHealthy && (
             <StackItem>
               <Alert
                 variant="warning"


frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
```diff
@@ -41,7 +41,7 @@ const StartNodeMaintenanceModal = withHandlePromise<StartNodeMaintenanceModalPro
 
   const [cephClusters, loaded] = useK8sWatchResource<K8sResourceKind[]>(cephClusterResource);
   const cephCluster = cephClusters?.[0];
-  const cephClusterHealthy = !cephCluster || cephCluster?.status?.health === 'OK';
+  const cephClusterHealthy = !cephCluster || cephCluster?.status?.health === 'HEALTH_OK';
```
afreen23, line 44:
```suggestion
  const cephClusterHealthy = cephCluster?.ceph?.status?.health === 'HEALTH_OK';
```

afreen23 (changes requested):

frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
SanjalKatiyar: thanks for pointing it out... just one thing we need this `!cephCluster` condition because we do not want to show the warning -alert if Ceph cluster is not installed.

afreen23: You are using optional chaining hence that would handle itself.
No worries if you want to keep it that way too.

afreen23 (approved):

frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
```diff
@@ -41,7 +41,7 @@ const StartNodeMaintenanceModal = withHandlePromise<StartNodeMaintenanceModalPro
 
   const [cephClusters, loaded] = useK8sWatchResource<K8sResourceKind[]>(cephClusterResource);
   const cephCluster = cephClusters?.[0];
-  const cephClusterHealthy = !cephCluster || cephCluster?.status?.health === 'OK';
+  const cephClusterHealthy = !cephCluster || cephCluster?.status?.ceph?.health === 'HEALTH_OK';
```
afreen23, line 44:
```suggestion
  const cephClusterHealthy = cephCluster?.status?.ceph?.health === 'HEALTH_OK';
```

frontend/packages/metal3-plugin/src/components/modals/StartNodeMaintenanceModal.tsx
afreen23: Do the respective change below:
```suggestion
!!cephCluster && !cephClusterHealthy &&
```

afreen23 (approved):

jtomasek (approved):
/lgtm

PR #8459, opened 2021-03-24
Bug 1766287: Use httpd in workload YAML templates

spadgett:
Use the httpd sample image in place of hello-openshift in our YAML templates. This requires that the sample images are imported on the cluster to run successfully.

This is a better example as it runs a real web server and will show the Apache test page if exposed by a route.

Screen Shot 2021-03-24 at 2 49 47 PM

/cc @smarterclayton

### DIFF
--- a/frontend/integration-tests/tests/crd-extensions.scenario.ts
+++ b/frontend/integration-tests/tests/crd-extensions.scenario.ts
@@ -274,9 +274,6 @@ describe('CRD extensions', () => {
         { metadata: { name: podName, labels: { app: name } } },
         safeLoad(content),
       );
-      // TODO: Remove when the default `registry.redhat.io/openshift4/ose-hello-openshift-rhel8`
-      // image is published.
-      newContent.spec.containers[0].image = 'openshift/hello-openshift:latest';
       await yamlView.setEditorContent(safeDump(newContent));
       await yamlView.saveButton.click();
       expect(crudView.errorMessage.isPresent()).toBe(false);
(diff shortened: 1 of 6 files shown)


rhamilto:
It looks like quite a few tests are using `openshift/hello-openshift`.  Do we want to update those, too?

yselkowitz:
> It looks like quite a few tests are using `openshift/hello-openshift`. Do we want to update those, too?

Yes, we need to purge all usage of hello-openshift, but that may be okay for a separate PR.

spadgett:
> It looks like quite a few tests are using `openshift/hello-openshift`. Do we want to update those, too?

I believe it's almost all mock data. There is one test for Deploy Image that deploys hello-openshift from quay. I'll look more closely tomorrow.

spadgett:
@rhamilto @yselkowitz OK, I removed references to hello-openshift from the core console integration tests. I also updated the YAML sample for ConsoleQuickStart to no longer reference it. Hopefully this passes CI.

I'll work with the plugin teams to remove references from plugin integration tests. There are many references in mock data as well, but those should be OK. I think any additional changes will need to be in follow up PRs given the number of references.

@rhamilto Another look?

rhamilto (approved):
/lgtm

gabemontero:
hey @spadgett - would you be amenable to backporting this and the other related PRs for this BZ to 4.7? 

On the samples side, we've had some tricky edge cases come up with the CVO creating the in payload `hello-openshift` imagestream during 4.6 to 4.7 upgrades when other features like disconnected, mirroring, and global image config for allowed / blocked registries.

To help minimize the impact of ^^, I'd like to backport the removal of the hello-openshift imagestream to 4.7 (where it was introduced) with https://github.com/openshift/cluster-samples-operator/pull/393 since its introduction was already marked as a mistake so to speak.

But certainly I don't want to break the 4.7 console where it still references that imagestream.

PR #8460, opened 2021-03-25
Pipeline Builder Validation Rework

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3165

Continuation / reworking of #4805 

**Analysis / Root cause**: 
Todays Pipeline Builder suffers from a weak and over-engineered error system.

**Solution Description**: 
We already have a validation library (`yup`) for our forms that can heavily be used to manage most state in our builder. Implemented a comprehensive `yup` schema that covers all the state and handles (almost) all the errors.

Visually should be near no difference.

`validation-utils.spec.ts`:
```
  Pipeline Build validation schema
    Form/YAML switcher validation
      ✓ should fail when provided an unknown editor type (17ms)
      ✓ should fail initial values because there are no tasks (6ms)
    Base form validation
      ✓ should pass if there is at least one Task (5ms)
      ✓ should fail if there is an invalid name (4ms)
      ✓ should fail if not provided a param name (4ms)
      ✓ should pass even if params default and description are empty (4ms)
      ✓ should pass when params default and description are provided (4ms)
      ✓ should pass when resource has a valid name and type (4ms)
      ✓ should fail when provided an unknown resource type (4ms)
      ✓ should pass when provided with a valid workspace name (4ms)
      ✓ should fail when workspaces do not have a name (4ms)
    Tasks and ListTasks validation
      ✓ should fail if just has a name (5ms)
      ✓ should pass if provided a taskSpec and name (4ms)
      ✓ should fail if provided an incomplete taskRef (4ms)
      ✓ should pass if provided with a proper taskRef and name (3ms)
      Validate Run Afters
        ✓ should fail if runAfter is a single string (5ms)
        ✓ should fail if runAfter is an array of strings that do not match other tasks (5ms)
        ✓ should pass if runAfter is an array of strings that match other task names (5ms)
        ✓ should fail if runAfter in listTasks does not match other listTasks (4ms)
        ✓ should fail if runAfter is itself (3ms)
        ✓ should pass if runAfter is an array of strings that match listTasks or tasks names (4ms)
      Validate Parameters
        ✓ should fail if task params has no value and also the task lacks a default (4ms)
        ✓ should pass if task params has no value but the task has a default (4ms)
        ✓ should pass if the task params has no value but the task default is an empty string (3ms)
        ✓ should pass if the task params has no value but the task is embedded and not external (4ms)
      Validate Resources
        ✓ should fail if the task does not include resources but the task does (5ms)
        ✓ should fail if the task does not fully include the required resources (4ms)
        ✓ should fail if the task contains some of the required resources (4ms)
        ✓ should fail early if the task contains some of the required resources and is the wrong type (4ms)
        ✓ should pass if the task contains all of the required resources (4ms)
        ✓ should fail if the task contains all of the required resources but the wrong types (4ms)
        ✓ should fail if the task contains all the required resources but a mismatch in name occurs (3ms)
      Validate Workspaces
        ✓ should fail if the task does not include workspaces but the task does (3ms)
        ✓ should fail if the task contains some of the required workspaces (4ms)
        ✓ should pass if the task contains all the required workspaces (4ms)
        ✓ should fail if the task contains all the required workspaces but a mismatch in name occurs (5ms)
```

`utils.spec.ts`:
```
  findTaskFromFormikData / findTask
    ✓ should decompose formik state & handle nulls
    ✓ should handle fail states (1ms)
    ✓ should be able to find a clusterTask
    ✓ should be able to find a namespacedTask
  safeName
    ✓ should straight return with no reserved names
    ✓ should return when no reserved names match
    ✓ should return a random 3 digit suffix if already in use (1ms)
  convertResourceToTask
    task with param variety
      ✓ should find the same amount of params that was given
      ✓ should not have a runAfter since it was not provided
      ✓ should refer to the proper taskRef (1ms)
      Param "param-with-description"
        ✓ should have a typeof value equal to the same typeof default
        ✓ should have not have a value (required param) if it lacks a default
      Param "param-with-default"
        ✓ should have a typeof value equal to the same typeof default
        ✓ should have a value for a default param
      Param "param-with-neither"
        ✓ should have a typeof value equal to the same typeof default
        ✓ should have not have a value (required param) if it lacks a default
      Param "param-with-both"
        ✓ should have a typeof value equal to the same typeof default (1ms)
        ✓ should have a value for a default param
      Param "array-param-without-default"
        ✓ should have a typeof value equal to the same typeof default
        ✓ should have not have a value (required param) if it lacks a default
      Param "array-param-with-default"
        ✓ should have a typeof value equal to the same typeof default
        ✓ should have a value for a default param
    namespacedTask with runAfter
      ✓ should have a taskRef of a task
      ✓ should consist of the explicit runAfter that was given
      ✓ should match the params that was given (1ms)
  runAfter Manipulation Utils
    mapReplaceRelatedInOthers
      ✓ should handle nulls
      ✓ should do nothing if there are no runAfters
      ✓ should do nothing if the runAfters do not include the target
      ✓ should replace a related runAfter with a new runAfter (1ms)
    mapRemoveRelatedInOthers
      ✓ should handle nulls
      ✓ should do nothing if there are no runAfters
      ✓ should do nothing if the runAfters do not include the target (1ms)
      ✓ should remove a related runAfter if it matches
    mapStitchReplaceInOthers
      ✓ should handle nulls
      ✓ should just remove if there are no runAfters in removalTask (1ms)
      ✓ should do nothing if the task is not in the runAfter (1ms)
      ✓ should remove the task and consume the runAfters of the removalTask
      ✓ should remove the task and combine runAfters with the removalTask (1ms)
      ✓ should remove the task and combine runAfters with the removalTask (dupe runAfter)
    mapBeRelated
      ✓ should handle nulls
      ✓ should add runAfter if matching name is provided
      ✓ should replace runAfter if one exists when matching name is provided (1ms)
    mapAddRelatedToOthers
      ✓ should handle nulls
      ✓ should do nothing if there are no runAfters
      ✓ should do nothing if there is no matching name in the runAfters
      ✓ should add task name to runAfters if related name is found
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarName.tsx
```diff
@@ -40,25 +25,34 @@ const TaskSidebarName: React.FC<TaskSidebarNameProps> = (props) => {
...
-      </SidebarInputWrapper>
+      <TextInput
+        id="task-name"
+        validated={isValid ? 'default' : 'error'}
+        isRequired
+        onChange={(value) => {
+          setInterimName(value);
+
+          setValidating(true);
+          nameValidationSchema(t, 63)
```
karthikjeeyar, line 36: If the modified name is already taken by one of the tasks or finallyTasks, then we should throw some error to force the user to provide a unique name.

karthikjeeyar: Name duplication messes up the graph structure because there are two nodes with same name.

https://user-images.githubusercontent.com/9964343/112825107-fbe61f00-90a8-11eb-8084-dc8aa4d62a3e.mp4

andrewballantyne: Yup, good catch... this was always an issue and it was just not uncovered haha. Will fix it as part of this PR.

karthikjeeyar:
Thanks Andrew. I verified locally works fine.
/lgtm

PR #8461, opened 2021-03-25
Bug 1927423: Monitoring: Add warning to list pages when silences cannot be loaded

kyoto:
Adds a warning message to the alerts list and alerting rules list pages
when the silences cannot be loaded from Alertmanager.

screenshot

FYI @cshinn


spadgett (approved):
/lgtm

PR #8462, opened 2021-03-25
Bug 1942843: Modified description for storage class encryption - 4.7 Backport

GowthamShanmugam:


GowthamShanmugam:
@cloudbehl please review

bipuladh:
@spadgett we are not making any changes to kubevirt. Could you please override the test? Thank you.

russellb:
This is blocked because the kubevirt-plugin is broken.  #8619 is needed to resolve it

spadgett:
Hi @russellb, we've actually removed kubevirt plugin as a required test. It was enabled by accident. See https://github.com/openshift/release/pull/17660

russellb:
@spadgett I think you'll also need to add `optional: true` to the job definition to not block on it if that's your intention

spadgett:
> @spadgett I think you'll also need to add `optional: true` to the job definition to not block on it if that's your intention

It should block when it runs, but it should only run when the particular kubevirt package changes rather than all PRs.

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.70

PR #8472, opened 2021-03-25
Fix for topology crash on load when application group is selected

jeff-phillips-18:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5687

**Analysis / Root cause**: 
Topology side panel contents are attempting to render before the graph data loaded by extensions is complete. The rendering of the application group panel accesses the graph data and hence fails.

**Solution Description**: 
Check for the graph data being set before accessing it. Remove redundant code that determines if there is side panel content.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


jerolimov (approved):
Tested it locally, could verify the crash before and fix with this PR.

/approve

jeff-phillips-18:
/hold

Making some updates for the side bar.

PR #8476, opened 2021-03-26
Create KSVC with min/max scale if workload has related HPA

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5655

**Solution Description**: 
Create KSVC with min/max scale if the workload has related HPA

image

**Test setup:**
- Install serverless operator created associated CRs (KnativeServing)
- Create an ns and add workload as D/DC
- Add HPA to the created workload with min/max replicas
- In topology, the workload context menu will have the action "Create Knatve Service"
- KSVC will be created with min/max scale based on related HPA

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/utils/knatify-utils.ts
+++ b/frontend/packages/knative-plugin/src/utils/knatify-utils.ts
@@ -22,15 +22,17 @@ import {
 import { RegistryType } from '@console/dev-console/src/utils/imagestream-utils';
 import { ServiceModel } from '../models';
 import { getKnativeServiceDepResource } from './create-knative-utils';
+import { KNATIVE_MAXSCALE_ANNOTATION, KNATIVE_MINSCALE_ANNOTATION } from '../const';
 
 const PART_OF = 'app.kubernetes.io/part-of';
 const knatify = 'knatify';
 
-export const getKnatifyWorkloadData = (obj: K8sResourceKind) => {
+export const getKnatifyWorkloadData = (obj: K8sResourceKind, relatedHpa?: K8sResourceKind) => {
   const { metadata, spec } = obj || {};
   const { name = '', namespace = '', labels = {}, annotations = {} } = metadata || {};
   const { metadata: templateMetadata, spec: templateSpec } = spec?.template || {};
   const { image, ports, imagePullPolicy, env, resources } = templateSpec?.containers[0] || {};
+  const { spec: hpaSpec } = relatedHpa ?? {};
 
   const healthChecks = getHealthChecksData(obj, 0);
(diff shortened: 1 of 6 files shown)


frontend/packages/knative-plugin/src/utils/knatify-utils.ts
```diff
@@ -48,7 +49,15 @@ export const getKnatifyWorkloadData = (obj: K8sResourceKind) => {
       template: {
         metadata: {
           labels: templateMetadata?.labels ?? {},
-          annotations: templateMetadata?.annotations ?? {},
+          annotations: {
+            ...(templateMetadata?.annotations && templateMetadata.annotations),
```
karthikjeeyar, line 53:
```suggestion
            ...templateMetadata?.annotations,
```

frontend/packages/knative-plugin/src/utils/knatify-utils.ts
```diff
@@ -48,7 +49,15 @@ export const getKnatifyWorkloadData = (obj: K8sResourceKind) => {
...
+            ...(hpaSpec?.minReplicas && {
+              'autoscaling.knative.dev/minScale': `${hpaSpec.minReplicas}`,
```
karthikjeeyar, line 55: import and use const here
```suggestion
              KNATIVE_MINSCALE_ANNOTATION: `${hpaSpec.minReplicas}`,
```

frontend/packages/knative-plugin/src/utils/knatify-utils.ts
```diff
@@ -48,7 +49,15 @@ export const getKnatifyWorkloadData = (obj: K8sResourceKind) => {
...
+            }),
+            ...(hpaSpec?.maxReplicas && {
+              'autoscaling.knative.dev/maxScale': `${hpaSpec.maxReplicas}`,
```
karthikjeeyar, line 58:
```suggestion
              KNATIVE_MAXSCALE_ANNOTATION: `${hpaSpec.maxReplicas}`,
```

karthikjeeyar:
I think knative service should have its own route, sharing the routes will always point to underlying D/DC. WDYT?

invincibleJai:
> I think knative service should have its own route, sharing the routes will always point to underlying D/DC. WDYT?

for KSVC we don't create a separate Resource for Route as do for k8s Route,  hence the svc name collision happens if KSVC and k8s Route are the same and only in the same namespace. This is not connected with this ticket, am already following up on this on Serving Channels. 

Will open a separate ticket based on the discussion if needed.

PR #8481, opened 2021-03-26
Bug 1943449: Bump hosted-git-info package as part of CVE

jhadvig:
/assign @spadgett

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -314,6 +314,7 @@
     "node": ">=14.x"
   },
   "resolutions": {
+    "hosted-git-info": "^3.0.8",
     "jquery": "3.5.1",
     "lodash-es": "^4.17.21",
     "minimist": "1.2.5",
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -9783,10 +9783,12 @@ hoopy@^0.1.4:
   resolved "https://registry.yarnpkg.com/hoopy/-/hoopy-0.1.4.tgz#609207d661100033a9a9402ad3dea677381c1b1d"
   integrity sha512-HRcs+2mr52W0K+x8RzcLzuPPmVIKMSv97RGHy0Ea9y/mpcaK+xTrjICA04KAHi4GRzxliNqNJEFYWHghy3rSfQ==
 
-hosted-git-info@^2.1.4:
-  version "2.7.1"
-  resolved "https://registry.yarnpkg.com/hosted-git-info/-/hosted-git-info-2.7.1.tgz#97f236977bd6e125408930ff6de3eec6281ec047"
-  integrity sha512-7T/BxH19zbcCTa8XkMlbK5lTo1WtgkFi3GvdWEyNuc4Vex7/9Dqbnpsf4JMydcfj9HCg4zUWFTL3Za6lapg5/w==
+hosted-git-info@^2.1.4, hosted-git-info@^3.0.8:
+  version "3.0.8"
+  resolved "https://registry.yarnpkg.com/hosted-git-info/-/hosted-git-info-3.0.8.tgz#6e35d4cc87af2c5f816e4cb9ce350ba87a3f370d"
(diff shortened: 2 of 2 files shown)


frontend/package.json
```diff
@@ -314,6 +314,7 @@
     "node": ">=14.x"
   },
   "resolutions": {
+    "hosted-git-info": "^3.0.8",
```
spadgett, line 317: It would be good to track this down to the dependency that includes it and bump that if possible instead of using a resolution. It looks like it might be babel?

frontend/yarn.lock
```diff
@@ -9783,10 +9783,12 @@ hoopy@^0.1.4:
   resolved "https://registry.yarnpkg.com/hoopy/-/hoopy-0.1.4.tgz#609207d661100033a9a9402ad3dea677381c1b1d"
   integrity sha512-HRcs+2mr52W0K+x8RzcLzuPPmVIKMSv97RGHy0Ea9y/mpcaK+xTrjICA04KAHi4GRzxliNqNJEFYWHghy3rSfQ==
 
-hosted-git-info@^2.1.4:
-  version "2.7.1"
-  resolved "https://registry.yarnpkg.com/hosted-git-info/-/hosted-git-info-2.7.1.tgz#97f236977bd6e125408930ff6de3eec6281ec047"
-  integrity sha512-7T/BxH19zbcCTa8XkMlbK5lTo1WtgkFi3GvdWEyNuc4Vex7/9Dqbnpsf4JMydcfj9HCg4zUWFTL3Za6lapg5/w==
+hosted-git-info@^2.1.4, hosted-git-info@^3.0.8:
```
spadgett, line 9786: I'm a little unsure about the resolutions here since this is a major version bump that could have API incompatibilities.

frontend/package.json
jhadvig: I see that it's from `node-sass` which was [bumped](https://github.com/openshift/console/pull/7365) into `5.0.0` version by you recently.

jhadvig: also...
Screenshot 2021-03-29 at 08 05 26

frontend/yarn.lock
jhadvig: Right. Due to my previous comment I'm wondering if it even makes sense to bump it since we are not directly using the pkg anywhere, which is not making the vulnerable `fromUrl()` method exposed...
Or we need to migrate to `dart-sass`, which seems to me to be a bit of a overkill ...
@spadgett thoughts ?

frontend/package.json
spadgett: Looks like you're right

```
❯ yarn why hosted-git-info
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "hosted-git-info"...?
[2/4] 🚚  Initialising dependency graph...
warning Resolution field "jquery@3.5.1" is incompatible with requested version "jquery@~3.4.1"
warning Resolution field "jquery@3.5.1" is incompatible with requested version "jquery@~3.4.1"
warning Resolution field "minimist@1.2.5" is incompatible with requested version "minimist@0.0.8"
warning Resolution field "minimist@1.2.5" is incompatible with requested version "minimist@0.0.8"
warning Resolution field "minimist@1.2.5" is incompatible with requested version "minimist@~0.0.1"
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "hosted-git-info@2.7.1"
info Reasons this module exists
   - "_project_#normalize-package-data" depends on it
   - Hoisted from "_project_#normalize-package-data#hosted-git-info"
   - Hoisted from "_project_#node-sass#meow#normalize-package-data#hosted-git-info"
   - Hoisted from "_project_#node-sass#meow#read-pkg-up#read-pkg#normalize-package-data#hosted-git-info"
```

spadgett: Hm, it sounds like we should be using dart sass anyway, though. @jhadvig Do you mind opening a story for that?

spadgett (approved):
/lgtm

PR #8482, opened 2021-03-26
Bug 1943543: Fix to include template and selected options for RC rollbacks

jeff-phillips-18:
Fixes rollbacks for ReplicationControllers to include the template in the rollback as well as the options selected by the user.


spadgett (approved):
/lgtm

PR #8483, opened 2021-03-26
Migrate Resource List Pages from Static to Dynamic extensions

bipuladh:


frontend/public/components/resource-list.tsx
```diff
@@ -57,11 +64,9 @@ export const ResourceListPage = connectToPlural(
           <title>{kindObj.labelPlural}</title>
         </Helmet>
         <AsyncComponent
-          autoFocus={!noProjectsAvailable}
           kind={modelRef}
           loader={componentLoader}
           match={props.match}
-          mock={noProjectsAvailable}
```
christianvogt, line 64: Is this removal safe?

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -15,7 +15,32 @@ export type StandaloneRoutePage = ExtensionDeclaration<
   }
 >;
 
+type ResourcePageCommonProps = {
+  /** See https://reacttraining.com/react-router/web/api/match */
+  match: RouteComponentProps['match'];
+  /** The resource kind scope. */
+  kind: string;
```
christianvogt, line 22: Shouldn't this be expanded to:
```
 model: {
  group: string,
  version: string,
  kind: string,
 }, 
```

bipuladh: I thought we would be passing GVK identifier to the Component.

christianvogt: The format of our single GVK string is an internal detail and we can construct it internally given the expanded object.

PR #8486, opened 2021-03-26
Bug 1942117: Fix breadcrumbs, delete modal, and create for i18n in pipelines models (and other kinds).

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1942673

cc: @rebeccaalpert


rebeccaalpert (approved):
Looks perfect. Thanks for addressing the other two in addition to the breadcrumbs. I was going to do them, but now I don't have to. :)

spadgett (approved):
/lgtm

PR #8490, opened 2021-03-26
Bug 1862084: Consistent formatting of dates and times

spadgett:
https://issues.redhat.com/browse/CONSOLE-2790

/cc @rhamilto @rebeccaalpert @kyoto @alimobrem @andybraren 

@cloudbehl I have not changed any of the ceph package code since I didn't have an environment to test.

Consistently format dates with `Intl.DateTimeFormat`. This will ensure
that we honor the browser's locale setting even for languages we don't
yet support. Always use AM/PM instead of 24 hour times so that we are
consistent. Ultimately, we want to make the date and time format
a user preference.

Remove moment.js in favor of native browser date-time utilities.
moment.js is no longer receiving new features. See project status here:

https://momentjs.com/docs/

Screen Shot 2021-03-26 at 2 44 09 PM

Screen Shot 2021-03-26 at 2 44 13 PM

Screen Shot 2021-03-26 at 2 44 17 PM

Screen Shot 2021-03-26 at 2 44 33 PM

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-utils.ts
@@ -1,9 +1,12 @@
 import * as _ from 'lodash';
-import * as moment from 'moment';
 import { TFunction } from 'i18next';
 import { humanizeNumberSI } from '@console/internal/components/utils';
+import {
+  dateFormatterNoYear,
+  parsePrometheusDuration,
+} from '@console/internal/components/utils/datetime';
 import { PrometheusResponse, PrometheusResult } from '@console/internal/components/graphs';
-import { parsePrometheusDuration } from '@console/internal/components/utils/datetime';
+
 import { PipelineKind } from '../../../types';
 
 export interface GraphData {
@@ -51,7 +54,7 @@ const formatPositiveValue = (v: number): string =>
 export const formatValue = (v: number): string =>
   (v < 0 ? '-' : '') + formatPositiveValue(Math.abs(v));
 export const formatDate = (date: Date) => {
-  return `${moment(date).format('MMM DD')}`;
+  return dateFormatterNoYear.format(date);
 };
 export const formatTimeSeriesValues = (result: PrometheusResult, samples: number, span: number) => {
   const { metric, values } = result;
@@ -103,10 +106,12 @@ export const getXaxisValues = (timespan: number): number[] => {
   const xValues = [];
   if (!timespan) return xValues;
   const oneDayDuration = parsePrometheusDuration('1d');
-  const endDate = new Date(Date.now()).setHours(0, 0, 0, 0);
   const numDays = Math.round(timespan / oneDayDuration);
-  for (let m = moment(endDate); xValues.length - 1 < numDays; m.subtract(1, 'days')) {
-    xValues.push(m.unix() * 1000);
+  const d = new Date(Date.now());
+  d.setHours(0, 0, 0, 0);
+  while (xValues.length - 1 < numDays) {
+    xValues.push(d.getTime());
+    d.setDate(d.getDate() - 1);
   }
   return xValues.slice(0, numDays);
 };
--- a/frontend/public/components/monitoring/query-browser.tsx
+++ b/frontend/public/components/monitoring/query-browser.tsx
@@ -49,10 +49,12 @@ import {
   useSafeFetch,
 } from '../utils';
 import {
+  dateFormatterNoYear,
+  dateTimeFormatter,
   formatPrometheusDuration,
-  getLocaleDate,
   parsePrometheusDuration,
-  twentyFourHourTime,
+  timeFormatter,
+  timeFormatterWithSeconds,
 } from '../utils/datetime';
 import { PrometheusAPIError } from './types';
 import { ONE_MINUTE } from '@console/shared/src/constants/time';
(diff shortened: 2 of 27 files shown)


spadgett:
This is pretty close to implementing https://issues.redhat.com/browse/CONSOLE-2790

spadgett:
cc @rawagner

rhamilto:
LGTM

spadgett:
Also related to https://issues.redhat.com/browse/CONSOLE-2765

frontend/public/components/utils/datetime.ts
```diff
@@ -3,8 +3,32 @@ import * as moment from 'moment';
 
 const getLocale = () => localStorage.getItem('bridge/language');
 
-export const getLocaleDate = (date: Date, options): string =>
-  date.toLocaleDateString(getLocale() || undefined, options);
+export const timeFormatter = new Intl.DateTimeFormat(getLocale() || undefined, {
```
spadgett, line 6: Note that this will not be reactive to language preference changes. I'm debating whether we should just always use `undefined` here and use the system locale. Date formatting seems like it should be a separate preference, particularly since we can support any locale, not just the ones we have translations for.

rhamilto: It won't be reactive without refreshing, but it will still be active.  I'm thinking it's better than not having it at all.

spadgett:
Remaining references to moment:

```
packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-utils.ts
2:import * as moment from 'moment';

public/i18n.js
12:import moment from 'moment';

public/components/utils/datetime.ts
2:import * as moment from 'moment';
44:  return moment(dateTime).from(moment(now), options.omitSuffix === true);
```

frontend/public/components/utils/datetime.ts
spadgett: We can make it reactive, but we'd need a `useEffect` hook with the locale as a dependency everywhere we use it. Maybe it's worth it.

Or we could create a new formatter each render.

spadgett: A new formatter each render wouldn't perform as well when using `Timestamp` in lists, though, since we'd be creating a formatter for every row.

spadgett:
I'm still working on this, but I pushed an additional commit that completely removes moment.js.

@andrewballantyne I had to change some pipeline metrics code and haven't tested that yet. I might need your guidance on how to test.

rawagner:
great work, thank you Sam!

andrewballantyne:
> I'm still working on this, but I pushed an additional commit that completely removes moment.js.
> 
> @andrewballantyne I had to change some pipeline metrics code and haven't tested that yet. I might need your guidance on how to test.

Absolutely. @karthikjeeyar is our resident point of contact for OpenShift Pipelines (and the one who did the feature work). 

Only note that I know off the top of my head is Pipeline Metrics runs on completed Pipelines only. So we need the Pipeline Runs to fully finish before the Metrics view updates. To get a Pipeline, the easiest way would be to go through the add flow. Add a git repo (from an app that is say nodejs or java) and scroll to the bottom of the form. There is an "Add Pipeline" option to create your a template-driven Pipeline with your workload (which will update said workload on each run). That'll get you going with a Pipeline.

I'll see about checking this out and doing some actual testing sometime today. Let me (or Karthik) know if you have any specific questions about Pipelines @spadgett

frontend/public/components/monitoring/query-browser.tsx
```diff
@@ -323,10 +320,15 @@ const Graph: React.FC<GraphProps> = React.memo(
 
     const xAxisTickCount = Math.round(width / 100);
     const xAxisTickShowSeconds = span < xAxisTickCount * ONE_MINUTE;
-    const xAxisTickFormat =
-      span > parsePrometheusDuration('1d')
-        ? (d) => `${formatDate(d)}\n${twentyFourHourTime(d)}`
-        : (d) => twentyFourHourTime(d, xAxisTickShowSeconds);
+    const xAxisTickFormat = (d) => {
+      if (span > parsePrometheusDuration('1d')) {
+        return dateTimeFormatter.format(d);
```
kyoto, line 325: I notice this causes the axis labels to overlap. Maybe we can keep a `${date}\n${time}` approach here?

screenshot

kyoto:
Thanks @spadgett, this looks great!

frontend/public/components/monitoring/query-browser.tsx
spadgett: Thanks, yeah. I somehow missed that there was a newline bulk replacing these. I'll roll that change back.

spadgett:
I'm seeing some label overlap at narrower screen widths. We might have to reduce the number of x-axis ticks we use.

Screen Shot 2021-03-29 at 12 44 25 PM

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-utils.ts
```diff
@@ -51,7 +54,7 @@ const formatPositiveValue = (v: number): string =>
 export const formatValue = (v: number): string =>
   (v < 0 ? '-' : '') + formatPositiveValue(Math.abs(v));
 export const formatDate = (date: Date) => {
-  return `${moment(date).format('MMM DD')}`;
+  return dateFormatter.format(date);
```
andrewballantyne, line 57: @spadgett We lost our x-axis:

image

My quick debugging is this change.

andrewballantyne: Let me know if you need anything else here @spadgett -- I'll step away from this PR until you need a re-review or help testing it.

spadgett: Thanks @andrewballantyne, I have a fix.

Screen Shot 2021-04-15 at 10 30 18 AM

spadgett:
/assign @rhamilto 

This should be ready for review.

rhamilto:
Huzzah!  Excellent work, @spadgett!

Tested in Safari 14.0.3 and everything looks good.

/lgtm

PR #8495, opened 2021-03-29
 Bug 1943996: Bump ua-parser-js to avoid ReDoS via malicious User-Agent header

jhadvig:
This is just a dependency, which we are not using/calling directly.

/assign @spadgett


spadgett (approved):
/lgtm

PR #8498, opened 2021-03-29
Expose plugin API

rawagner:
expose k8s watch hook via window.api which can be used by dynamic plugin & webpack externals configuration.
Can be tested with Console Dynamic Foo plugin `wp_externals` branch  https://github.com/rawagner/console-dynamic-foo/tree/wp_externals

Updated README with required configuration https://github.com/rawagner/console-dynamic-foo/tree/wp_externals#configuration-required-to-use-consoles-api

Depends on https://github.com/openshift/console/pull/8378 (first 2 commits)


frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
```diff
@@ -0,0 +1,13 @@
...
+  useK8sWatchResource,
+  useK8sWatchResources,
+} from '@console/internal/components/utils/k8s-watch-hook';
+
+export const exposePluginAPI = _.once(() => {
+  window.api = {
+    useK8sWatchResource,
+    useK8sWatchResources,
+  };
+});
```
christianvogt, line 13: IMO we should be using webpack module federation for exposing API instead of externals and having to create a global API object.

Plugins who want to define their own API for other plugins will need a way to expose their API and that should be consistent with the console core.

frontend/packages/console-dynamic-plugin-sdk/src/api/api.ts
```diff
@@ -0,0 +1,10 @@
+import { UseK8sWatchResource, UseK8sWatchResources } from './api-types';
+
+export * from './api-types';
+
+const MockImpl = () => {
+  throw new Error('You need to configure webpack externals to use this function at runtime.');
+};
+
+export const useK8sWatchResource: UseK8sWatchResource = MockImpl;
+export const useK8sWatchResources: UseK8sWatchResources = MockImpl;
```
christianvogt, line 10: I'd like to see us bring the actual code into the SDK package instead of doing this work around.

christianvogt (commented):
While the current approach I don't think is ideally where we want to go, if it works, we can start here and work on improving it in future prs.

rawagner:
Thanks for the review @christianvogt . Good to hear you're ok with the current approach for the first iteration. I will look into moving the code we want to expose into `SDK` & module federation.

rawagner:
closing in favor of https://github.com/openshift/console/pull/8378

PR #8503, opened 2021-03-29
[release-4.7] Bug 1944349: fix backwards incompatible trigger api changes

karthikjeeyar:
This is a manual cherry pick of https://github.com/openshift/console/pull/8437
Problem: 
OSP 1.4 (GA version) has breaking api changes which needs to be addressed in OCP-4.7.

Solution:
Fixes the trigger API breaking changes in 1.4 OSP in OCP-4.7. Replaced `template.name` with `template.ref` and still support earlier versions operator to use template.name via Yaml.

Screenshots:
No UI changes

Test cases:
submitTrigger
    ✓ expect the eventlistener to contain template.name if the operator is below GA version (1.4.0) (23ms)
    ✓ expect the eventlistener to contain template.ref if the operator is GA version (1.4.0) (4ms)
    
/kind bug

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
@@ -1,3 +1,4 @@
+import { lt } from 'semver';
 import { getRandomChars } from '@console/shared';
 import { apiVersionForModel, RouteKind } from '@console/internal/module/k8s';
 import { RouteModel } from '@console/internal/models';
@@ -57,6 +58,14 @@ export const createEventListener = async (
       ref: triggerBinding.metadata.name,
     };
   };
+  const getTriggerTemplate = (name: string) => {
+    if (lt(pipelineOperatorVersion, '1.4.0')) {
+      return {
+        name,
+      };
+    }
+    return { ref: name };
+  };
 
   return {
     apiVersion: apiVersionForModel(EventListenerModel),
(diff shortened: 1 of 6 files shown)


frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/resource-utils.ts
```diff
@@ -69,7 +69,7 @@ export const createEventListener = async (
       triggers: [
         {
           bindings: triggerBindings.map(mapTriggerBindings),
-          template: { name: triggerTemplate.metadata.name },
+          template: { ref: triggerTemplate.metadata.name },
```
andrewballantyne, line 72: What version was it added in? Do we have any concerns with a 4.7 cluster having like a 4.6 operator install version?

karthikjeeyar: I checked the version. It was added in 0.9.x so ocp-4.6 and below wont have this change. I will conditionally add ref based on the operator version.

karthikjeeyar: Updated based on the operator version.

vikram-raj:
@karthikjeeyar I notice that Removal of Triggers has some issue and need to fix.

image (1)

karthikjeeyar:
Thanks @vikram-raj, I have fixed removal issue and pushed [here](https://github.com/openshift/console/compare/274c9b14b49e64d31f7dd108c13cc1b069031612..0288047fce7eb09fd8f510d540a6b456fad39ab5).

vikram-raj (commented):
Verify the changes with the 1.3 and 1.4 pipeline operators.

/lgtm

markmc:
(patch manager) prioritized for this release to align with pipelines GA

PR #8504, opened 2021-03-29
Bug 1936620: Fix create operand form default boolean field

TheRealJon:
Use rjsf built in BooleanField, and override checkbox widget when boolean switch descriptor is used.


spadgett (commented):
LGTM, but it looks like you need to run `yarn i18n`

spadgett (approved):
/lgtm

PR #8505, opened 2021-03-29
[WIP] Bug 1942117: Fix Workloads i18n misses

rebeccaalpert:
Addressed terminal error message, Deployments Edit Horizontal Pod Autoscaler descriptions, Deployments Add Storage form, Deployments Edit Update Strategy, and Statefulset details page Volume section remove volume modal.

Still trying to figure out where Create Knative Service form data is coming from since that's still partially in English.

Screen Shot 2021-03-29 at 3 44 27 PM

Screen Shot 2021-03-29 at 1 41 43 PM

Screen Shot 2021-03-29 at 2 02 23 PM

Screen Shot 2021-03-29 at 2 14 09 PM

Screen Shot 2021-03-29 at 3 00 46 PM

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1942117


frontend/packages/operator-lifecycle-manager/src/components/operand/const.ts
```diff
@@ -1,10 +1,13 @@
+import i18next from 'i18next';
 import { JSONSchema6 } from 'json-schema';
 import { JSONSchemaType } from '@console/shared/src/components/dynamic-form';
 
-export const YAML_HELP_TEXT =
-  'Create by manually entering YAML or JSON definitions, or by dragging and dropping a file into the editor.';
-export const FORM_HELP_TEXT =
-  'Create by completing the form. Default values may be provided by the Operator authors.';
+export const YAML_HELP_TEXT = i18next.t(
```
spadgett, line 5: These won't be reactive. We should probably remove these as constants and just inline them in the component. (If they're used in more than one place, we can create custom components that are just for the messages.)

frontend/public/components/modals/configure-update-strategy-modal.tsx
```diff
@@ -1,15 +1,18 @@
...
 import { DeploymentModel } from '../../models';
 import { createModalLauncher, ModalTitle, ModalBody, ModalSubmitFooter } from '../factory/modal';
-import { pluralize, withHandlePromise, HandlePromiseProps } from '../utils';
+import { withHandlePromise, HandlePromiseProps } from '../utils';
 import { RadioInput } from '../radio';
 
-export const UPDATE_STRATEGY_DESCRIPTION =
-  'How should the pods be replaced when a new revision is created?';
+export const UPDATE_STRATEGY_DESCRIPTION = i18next.t(
+  'public~How should the pods be replaced when a new revision is created?',
```
spadgett, line 14: This needs to be moved into a component.

frontend/public/components/modals/remove-volume-modal.tsx
```diff
@@ -70,26 +71,29 @@ export const RemoveVolumeModal: React.FC<RemoveVolumeModalProps> = (props) => {
...
-          Are you sure you want to remove volume{' '}
-          <strong className="co-break-word">{volume.name}</strong>
-          <span>
-            {' '}
-            from <strong>{kind.label}</strong>: <strong>{resource.metadata.name}</strong>?
-          </span>
+          <Trans t={t} ns="public">
+            Are you sure you want to remove volume{' '}
+            <strong className="co-break-word">{volume.name}</strong>
+            <span>
```
spadgett, line 87: Can you remove the unnecessary span?

frontend/public/components/storage/shared.ts
```diff
@@ -32,25 +33,25 @@ export const initialAccessModes = ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMa
 export const accessModeRadios = [
   {
     value: 'ReadWriteOnce',
-    title: 'Single User (RWO)',
+    title: i18next.t('public~Single User (RWO)'),
```
spadgett, line 36: This won't be reactive. We should look for a different way to handle this instead of setting it once as a top-level constant in the module.

rebeccaalpert:
I pushed a commit addressing feedback but still need to test to make sure I didn't break anything and fix the existing test failures.

frontend/packages/dev-console/src/components/hpa/HPAUtilizationField.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
```
zherman0, line 2: This file is being worked in PR8272 FYI

spadgett:
You have a number of build errors:

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx(7,3):
TS2305: Module '"../../../../../../../public/components/storage/shared"' has no exported member 'accessModeRadios'.

ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/VolumeClaimTemplateForm.tsx(8,3):
TS2305: Module '"../../../../../../../public/components/storage/shared"' has no exported member 'volumeModeRadios'.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx(193,9):
TS6133: 'InfoToolTip' is declared but its value is never read.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx(194,6):
TS2304: Cannot find name 'Tooltip'.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx(198,8):
TS2304: Cannot find name 'OutlinedQuestionCircleIcon'.

ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx
ERROR in /go/src/github.com/openshift/console/frontend/public/components/storage/create-pvc.tsx(199,7):
TS2304: Cannot find name 'Tooltip'.
```

spadgett:
I have a feeling this will conflict with #8380 (although I haven't confirmed).

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -36,6 +60,34 @@ export const AccessModeSelector: React.FC<AccessModeSelectorProps> = (props) =>
 
   const { t } = useTranslation();
 
+  const getPVCAccessModes = (resource: PersistentVolumeClaimKind, key: string) => {
+    const accessModeRadios = [
```
spadgett, line 64: Avoid introducing duplicate code. We can still translate these without copying them into each component. Create a function that returns these values and uses `i18next.t`

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -6,23 +6,47 @@ import {
...
-        res.push(mode[key]);
-      }
-      return res;
+const AccessModeRadios = (props) => {
+  const { t } = useTranslation();
+  const { accessMode, allowedAccessModes, onAccessModeChange } = props;
+  const accessModeRadios = [
+    {
+      value: 'ReadWriteOnce',
+      title: t('public~Single User (RWO)'),
```
spadgett, line 19:
```suggestion
      title: t('public~Single user (RWO)'),
```

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -6,23 +6,47 @@ import {
...
+    },
+    {
+      value: 'ReadWriteMany',
+      title: t('public~Shared Access (RWX)'),
```
spadgett, line 23:
```suggestion
      title: t('public~Shared access (RWX)'),
```

frontend/packages/console-app/src/components/access-modes/access-mode.tsx
```diff
@@ -6,23 +6,47 @@ import {
...
+    },
+    {
+      value: 'ReadOnlyMany',
+      title: t('public~Read Only (ROX)'),
```
spadgett, line 27:
```suggestion
      title: t('public~Read only (ROX)'),
```

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
```diff
@@ -56,6 +56,15 @@ export const CreateOperand: React.FC<CreateOperandProps> = ({
   model,
 }) => {
   const { t } = useTranslation();
+
+  const FORM_HELP_TEXT = t(
```
spadgett, line 60: Just inline these messages instead of creating constants for them. They're only used in one place.

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -117,6 +189,16 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
   ]);
 
   const { t } = useTranslation();
+
+  const InfoToolTip = () => (
```
spadgett, line 193: I might be missing it, but where is this used?

It would be better to make this its own component defined outside of this one.

spadgett (changes requested):

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
rebeccaalpert: It's used in two places; I'd have to duplicate it if I inline.

frontend/public/components/storage/create-pvc.tsx
rebeccaalpert: Looks like caller was removed when I rebased. I'll take this out.

rebeccaalpert:
Looks like the build errors are due to changes that crept in when I rebased. Will fix.

rebeccaalpert:
Addressed PR feedback.

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
spadgett: I only see one? You wouldn't need it in the dependency list if it's inlined inside of `useCallback`.

If we really need to keep it this way, I'd use camelCase `formHelpText`.

frontend/public/components/storage/shared.ts
```diff
@@ -29,30 +30,33 @@ export const provisionerAccessModeMapping = {
   'pd.csi.storage.gke.io': ['ReadWriteOnce'],
 };
 export const initialAccessModes = ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'];
-export const accessModeRadios = [
+
+export const accessModeRadios: any = () => [
```
spadgett, line 34: Use a better return type than `any`. If you leave the return type off, TypeScript would infer something better here.

frontend/public/components/storage/shared.ts
```diff
@@ -29,30 +30,33 @@ export const provisionerAccessModeMapping = {
...
   },
   {
     value: 'ReadOnlyMany',
-    title: 'Read Only (ROX)',
+    title: i18next.t('public~Read only (ROX)'),
   },
 ];
-export const volumeModeRadios = [
+
+export const volumeModeRadios: any = () => [
```
spadgett, line 49: Use a better return type than `any`. If you leave the return type off, TypeScript would infer something better here.

frontend/public/components/storage/shared.ts
rebeccaalpert: Unfortunately it complains about not having access to filter or map. :) I tried a fancier type and it complained about not having access to pop, etc. I can keep trying more types.

frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx
rebeccaalpert: I don't think the second one is in the diff. I can rename it.

rebeccaalpert:
I think I got everything. Let me know if you have additional feedback.

frontend/public/components/storage/shared.ts
```diff
@@ -29,30 +30,33 @@ export const provisionerAccessModeMapping = {
   'pd.csi.storage.gke.io': ['ReadWriteOnce'],
 };
 export const initialAccessModes = ['ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'];
-export const accessModeRadios = [
+
+export const accessModeRadios = () => [
```
spadgett, line 34: Let's call this `getAccessModeRadios`. It makes it more obvious it's a simple getter function.

frontend/public/components/storage/shared.ts
```diff
@@ -29,30 +30,33 @@ export const provisionerAccessModeMapping = {
...
   },
   {
     value: 'ReadOnlyMany',
-    title: 'Read Only (ROX)',
+    title: i18next.t('public~Read only (ROX)'),
   },
 ];
-export const volumeModeRadios = [
+
+export const volumeModeRadios = () => [
```
spadgett, line 49:
```suggestion
export const getVolumeModeRadios = () => [
```

rebeccaalpert:
All set!

rebeccaalpert:
Rebased and ran `yarn i18n`.

frontend/packages/dev-console/src/components/hpa/HPAUtilizationField.tsx
```diff
@@ -22,6 +23,7 @@ const HPAUtilizationField: React.FC<HPAUtilizationFieldProps> = ({
   onUpdate,
   type,
 }) => {
+  const { t } = useTranslation();
```
florkbr, line 26: @rebeccaalpert build failure, looks like `const { t }` is defined above (merge issue?)

```bash
ERROR in ./packages/dev-console/src/components/hpa/HPAUtilizationField.tsx 15:12
Module parse failed: Identifier 't' has already been declared (15:12)
File was processed with these loaders:
 * ./node_modules/cache-loader/dist/cjs.js
 * ./node_modules/thread-loader/dist/cjs.js
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
|     const thisErrorMetric = (_e = (_d = (_c = errors.formData) === null || _c === void 0 ? void 0 : _c.spec) === null || _d === void 0 ? void 0 : _d.metrics) === null || _e === void 0 ? void 0 : _e[index];
|     const error = (_g = (_f = thisErrorMetric === null || thisErrorMetric === void 0 ? void 0 : thisErrorMetric.resource) === null || _f === void 0 ? void 0 : _f.target) === null || _g === void 0 ? void 0 : _g.averageUtilization;
>     const { t } = useTranslation();
|     return (React.createElement(FormGroup, { fieldId: `${type}-utilization`, label: t('devconsole~{{label}} Utilization', { label }), helperText: t('devconsole~{{label}} request and limit must be set before {{label}} utilization can be set.', { label }), helperTextInvalid: error, validated: error ? 'error' : null },
|         React.createElement(InputGroup, null,
```

rebeccaalpert: Merge issue! :(

I'll fix it. Thanks!

spadgett (approved):
/lgtm

rebeccaalpert:
Rebased to fix conflict.

rebeccaalpert:
Rebased public.json.

spadgett (approved):
/lgtm

PR #8507, opened 2021-03-30
Query Browser: Switch colors to `ChartThemeColor.multiUnordered`

kyoto:
Also removes the `@ts-ignore`, which seems to no longer be necessary.

Before | After
-|-
multi | multiUnordered
multi2 | multiUnordered2


jcaianirh (commented):
lgtm

PR #8508, opened 2021-03-30
Bug 1944509: Add context to translation strings

yaacov:
We have one text context separated to several translatable texts, this makes it hard for the translating team to know how to correctly translate each text fragment. and may in some languages cause the text to be un translatable if order of sentance chage, e.g. if some langauge use "<vm name> for ssh" instead of "ssh for <vm name>"


yaacov:
@metalice please review

PR #8509, opened 2021-03-30
Bug 1944273: don't use latest virtio disk image for downstream

yaacov:
If you attempt to deploy a Windows VM, the following container image will be used: registry.redhat.io/container-native-virtualization/virtio-win without any version specifier (implicitly the :latest tag will be used).

At some point, the :latest tag was removed, because people should use version-stable components.

The GUI shouldn't point at the latest tag, but to a version that matches the current release.


yaacov:
@glekner please review

PR #8515, opened 2021-03-30
View shortcuts on the topology list view

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5689
https://issues.redhat.com/browse/ODC-5688

**Root analysis:**
- `View shortcuts` is not shown on the topology list view 
- It is disabled when there are no workloads

**Solution description:**
- showing `View shortcuts` in both the views
- when there are no workloads only **ctrl + spacebar** and **drag and drop** actions are shown
- In the List view, when there are workloads only  **click**, **ctrl + spacebar** and **drag and drop** actions are shown

**Test coverage:**
test

**Screenshots:**
List view with workloads
Screenshot from 2021-03-30 18-51-28

List view with no workloads
Screenshot from 2021-03-30 18-51-20

Graph view with no workloads
Screenshot from 2021-03-30 18-50-54

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


bgliwa01:
> question for UX is whether the shortcuts menu should be changed based on whether there are nodes present in the graph or not. This PR shows a reduced list based if there are no nodes present.
> cc @bgliwa01

Good question! I think it makes sense to have the reduced list.

andrewballantyne:
I think the goal of this PR was a reduced list because it's the list view not because it was empty? 🤔  [Slack convo](https://coreos.slack.com/archives/CHEE211PW/p1616601549017400?thread_ts=1616601432.016700&cid=CHEE211PW) (which is in the ticket)

andrewballantyne:
> > question for UX is whether the shortcuts menu should be changed based on whether there are nodes present in the graph or not. This PR shows a reduced list based if there are no nodes present.
> > cc @bgliwa01
> 
> Good question! I think it makes sense to have the reduced list.

@christianvogt so that does impact this ticket: https://issues.redhat.com/browse/ODC-5688

christianvogt:
@andrewballantyne looks like this PR can close both issues.

nemesis09:
@debsmita1 
The tests for TopologyPageToolbar are failing. They need to be updated.

debsmita1:
> @debsmita1
> The tests for TopologyPageToolbar are failing. They need to be updated.

Fixed

frontend/packages/topology/src/components/graph-view/TopologyShortcuts.tsx
```diff
@@ -1,23 +1,30 @@
...
 import { TFunction } from 'i18next';
 import { Shortcut, ShortcutTable } from '@console/shared';
+import { TopologyViewType } from '../../topology-types';
 
 export type Options = {
   supportedFileTypes: string[];
+  isEmptyModel: boolean;
+  viewType: TopologyViewType;
 };
 export const getTopologyShortcuts = (t: TFunction, options: Options): React.ReactElement => (
```
nemesis09, line 11: Since this is rendering components conditionally, we should have a unit test for it.

debsmita1: added TopologyShortcuts.spec.tsx

frontend/packages/topology/src/__tests__/TopologyShortcuts.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+describe('TopologyShortcuts tests', () => {
+  it('should show reduced list in view shortcuts popover on topology toolbar when there are no workloads', () => {
+    const wrapper = shallow(
+      getTopologyShortcuts(jest.fn(), {
+        supportedFileTypes: undefined,
+        isEmptyModel: true,
+        viewType: TopologyViewType.graph,
+      }),
+    );
+    const secondWrapper = shallow(wrapper.getElement());
```
nemesis09, line 15: why do we need a second shallow rendering here?

frontend/packages/topology/src/__tests__/TopologyShortcuts.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(secondWrapper.find(Shortcut).exists()).toBe(true);
+    expect(secondWrapper.find(Shortcut).props().ctrl).toBe(true);
```
nemesis09, line 17: Ideally we shouldn't be looking inside props of a component rendered while testing.
but I couldn't think of a way in which we could avoid that for testing this util.
@andrewballantyne WDYT?

frontend/packages/topology/src/__tests__/TopologyShortcuts.spec.tsx
andrewballantyne: I think because `shallow` rendering prevented her from reaching the details she wanted. `mount` would have been needed.

frontend/packages/topology/src/__tests__/TopologyShortcuts.spec.tsx
andrewballantyne: Agreed, tests that dig into components is a bad idea. These tests are really getting nitty and gritty into the components and the way they are constructed just to see if they are there.

How about `data-test-id`s? `Shortcut` does not provide that functionality yet, but it would be easy to add, perhaps just a spread of any unknown props to the `tr`.

Then to test, it would be just a selection of `data-test-id`s to see if the shortcuts are there.

andrewballantyne (changes requested):

frontend/packages/topology/src/__tests__/TopologyShortcuts.spec.tsx
nemesis09: Yes, this seems like a good alternative. We could use `data-test-id`s to facilitate the tests, since we just need to check the presence of the components based on the conditions.

frontend/packages/topology/src/components/graph-view/TopologyShortcuts.tsx
```diff
@@ -1,23 +1,30 @@
...
   <ShortcutTable>
```
nemesis09, line 12: I'm a bit confused with this being a util. It could simply be a component and to me it looks more like a component.
I could be missing something here. Is there a reason we chose to make this a util rather than a component?

nemesis09: although, it isn't really related to the problem that this bug is about.

nemesis09:
Thank you for the changes @debsmita1 
I do have some confusions regarding the util but the bug is addressed irrespective of that.
/lgtm

frontend/packages/topology/src/components/graph-view/TopologyShortcuts.tsx
debsmita1: I doubt if there's any specific reason why this is a util and not a component

frontend/packages/topology/src/components/graph-view/TopologyShortcuts.tsx
```diff
@@ -1,23 +1,30 @@
...
-    <Shortcut drag>{t('topology~Move')}</Shortcut>
-    <Shortcut shift drag>
-      {t('topology~Edit Application grouping')}
-    </Shortcut>
-    <Shortcut rightClick>{t('topology~Access context menu')}</Shortcut>
-    <Shortcut click>{t('topology~View details in side panel')}</Shortcut>
-    <Shortcut hover>{t('topology~Access create connector handle')}</Shortcut>
+    {!options.isEmptyModel && options.viewType === TopologyViewType.graph && (
+      <>
+        <Shortcut drag>{t('topology~Move')}</Shortcut>
```
andrewballantyne, line 15: You should be `data-test-id`'ing these items and checking if they are there. You're data-test ids for lower items is fine, but you should be testing for the whole item not just a bunch of loose operations. Checking to see if something has `drag` is _okay_ but you really should be writing a test to see if `Move` is there when the condition is met.

debsmita1: Done

frontend/packages/console-shared/src/components/shortcuts/Shortcut.tsx
```diff
@@ -43,37 +43,43 @@ const Shortcut: React.FC<ShortcutProps> = ({
   return (
     <tr>
```
andrewballantyne, line 44: Nit: You should add the data-test-id to this node

andrewballantyne (approved):
Minor nit, probably should be done. Snag a re-review from @nemesis09 or someone else on the team early India morning for the final LGTM.

frontend/packages/topology/src/components/graph-view/TopologyShortcuts.tsx
rohitkrai03: I think it would be a good idea to convert this to a component. But since its not related to the changes in this PR, I would let this go in without it as well.

rohitkrai03 (commented):
Overall changes looks good to me.

rohitkrai03 (approved):
/lgtm

PR #8516, opened 2021-03-30
show type header for serverless function in topology

nemesis09:
Story:
https://issues.redhat.com/browse/ODC-5658

Problem/Description:
The Topology sidebar and Service details page should show that it belongs to serverless function

Solution:
Add a "Type" header with content "Function" under it to indicate the details belong to serverless function

Screens:
_Topology Sidebar for Serverless function_
Screenshot from 2021-03-30 21-22-14

_Service details page for Serverless function_
Screenshot from 2021-03-30 21-23-05

_Topology Sidebar for Knative service_
Screenshot from 2021-03-30 21-22-32

_Service details page for Knative service_
Screenshot from 2021-03-30 21-22-42

_Secret details page_
Screenshot from 2021-04-09 22-34-24

Test Coverage:
Screenshot from 2021-03-30 21-27-00
Screenshot from 2021-03-30 21-30-05

Browser Conformance:
- [x] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/components/overview/ServerlessFunctionType.tsx
+++ b/frontend/packages/knative-plugin/src/components/overview/ServerlessFunctionType.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+
+const ServerlessFunctionType: React.FC = () => {
+  const { t } = useTranslation();
+  return (
+    <dl className="co-m-pane__details">
+      <dt>{t('knative-plugin~Type')}</dt>
+      <dd>{t('knative-plugin~Function')}</dd>
+    </dl>
+  );
+};
+
+export default ServerlessFunctionType;
--- a/frontend/packages/knative-plugin/src/topology/const.ts
+++ b/frontend/packages/knative-plugin/src/topology/const.ts
@@ -21,3 +21,6 @@ export const KNATIVE_GROUP_NODE_PADDING = [
 ];
 
 export const EVENT_MARKER_RADIUS = 6;
+
+// URI Kind
+export const URI_KIND = 'URI';
--- a/frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
+++ b/frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
@@ -37,6 +37,7 @@ import {
   KNATIVE_GROUP_NODE_HEIGHT,
   KNATIVE_GROUP_NODE_PADDING,
   KNATIVE_GROUP_NODE_WIDTH,
+  URI_KIND,
 } from './const';
 import {
   getDynamicEventSourcesModelRefs,
@@ -996,8 +997,7 @@ const sinkURIDataModel = (
           namespace: data.resources.obj.metadata.namespace || '',
         },
         spec: { sinkUri },
-        type: { nodeType: NodeType.SinkUri },
-        kind: 'URI',
+        kind: URI_KIND,
       };
       const sinkData: KnativeTopologyDataObject<KnativeServiceOverviewItem> = {
         id: sinkTargetUid,
@@ -1140,7 +1140,7 @@ export const createKnativeEventSourceSink = (
   }
   const eventSourceObj = _.omit(source, 'status');
   let sink = {};
-  if (NodeType.SinkUri === target.type?.nodeType) {
+  if (target.kind === URI_KIND) {
     sink = {
       uri: target?.spec?.sinkUri,
     };
@@ -1246,13 +1246,12 @@ export const createSinkPubSubConnection = (
   return createEventingPubSubSink(resources.obj, targetObj);
 };
 
-export const isServerlessFunction = (element: Node): boolean => {
+export const isServerlessFunction = (element: K8sResourceKind): boolean => {
   if (!element) {
     return false;
   }
-
   const {
     metadata: { labels },
-  } = getResource(element);
+  } = element;
   return !!labels?.[SERVERLESS_FUNCTION_LABEL];
 };
--- a/frontend/public/components/default-resource.tsx
+++ b/frontend/public/components/default-resource.tsx
@@ -41,7 +41,10 @@ const tableColumnClasses = [
   Kebab.columnClass,
 ];
 
-export const DetailsForKind = (kind: string) =>
+export const DetailsForKind = (
+  kind: string,
+  additionalDetailsCallback?: (obj: K8sResourceKind) => React.ReactNode,
+) =>
   function DetailsForKind_({ obj }) {
     const { t } = useTranslation();
 
@@ -49,6 +52,9 @@ export const DetailsForKind = (kind: string) =>
       const model = modelFor(item);
       return model?.labelKey ? t(model.labelKey) : kindForReference(item);
     };
+
+    const additionalDetails: React.ReactNode =
+      additionalDetailsCallback && additionalDetailsCallback(obj);
 
     return (
       <>
@@ -64,6 +70,7 @@ export const DetailsForKind = (kind: string) =>
                 showNodeSelector={false}
               />
             </div>
+            {additionalDetails && <div className="col-md-6">{additionalDetails}</div>}
           </div>
         </div>
         {_.isArray(obj?.status?.conditions) && (
--- a/frontend/public/components/secret.jsx
+++ b/frontend/public/components/secret.jsx
@@ -19,6 +19,7 @@ import {
 } from './utils';
 import { SecretType } from './secrets/create-secret';
 import { configureAddSecretToWorkloadModal } from './modals/add-secret-to-workload';
+import { DetailsItem } from './utils/details-item';
 
 export const WebHookSecretKey = 'WebHookSecretKey';
 
@@ -102,6 +103,7 @@ const SecretTableRow = ({ obj: secret, index, key, style }) => {
 
 const SecretDetails = ({ obj: secret }) => {
   const { t } = useTranslation();
+  const { data, type } = secret;
   return (
     <>
       <div className="co-m-pane__body">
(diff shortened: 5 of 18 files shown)


nemesis09:
/assign @invincibleJai 
cc @openshift/team-devconsole-ux

bgliwa01 (approved):
LGTM! thanks!

invincibleJai:
> Also I'm not sure if there is another story to take care of this, but there should be a different fx icon in topology. cc @beaumorley

@bgliwa01 yes it handled as part of https://github.com/openshift/console/pull/8525 PTAL , thanks !!

nemesis09:
> > Also I'm not sure if there is another story to take care of this, but there should be a different fx icon in topology. cc @beaumorley
> 
> @bgliwa01 yes it handled as part of #8525 PTAL , thanks !!

Thanks! @invincibleJai

frontend/public/components/default-resource.tsx
```diff
@@ -64,6 +67,9 @@ export const DetailsForKind = (kind: string) =>
                 showNodeSelector={false}
               />
             </div>
+            {additionalDetailsCallback && (
+              <div className="col-md-6">{additionalDetailsCallback(obj)}</div>
```
sahil143, line 71: If `additionalDetailsCallback` return null then an empty div will rendered.

frontend/public/components/utils/details-page.tsx
```diff
@@ -91,8 +91,6 @@ export const ResourceSummary: React.FC<ResourceSummaryProps> = ({
           />
         </DetailsItem>
       )}
-      {type ? <dt>{t('details-page~Type')}</dt> : null}
-      {type ? <dd>{type}</dd> : null}
```
sahil143, line 95: Why is this removed? If any resource had the `type` it would not work anymore

frontend/public/components/default-resource.tsx
nemesis09: added check to restrict render when function returns null or undefined.

frontend/public/components/utils/details-page.tsx
nemesis09: Type was added for "Secrets" alone. No other resources seemed to use this.
any specific type for a resource can be handled using annotations, as I have done for sinkUri.
"type" for Secret has been added to "SecretKind" which is specific to secret.
please refer to this conversation for more information
https://coreos.slack.com/archives/CHGU4P8UU/p1616684514128300

sahil143 (approved):
/lgtm

verified locally, works as expected

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -977,9 +978,9 @@ export const transformKnNodeData = (
                 metadata: {
                   uid: sinkTargetUid,
                   namespace: data.resources.obj.metadata.namespace || '',
+                  annotations: { [typeAnnotation]: NodeType.SinkUri },
                 },
                 spec: { sinkUri },
-                type: { nodeType: NodeType.SinkUri },
                 kind: 'URI',
```
christianvogt, line 984: Can we not use the fake `kind` we created here instead of a new annotation?

frontend/packages/knative-plugin/src/components/overview/ServerlessFunctionType.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+
+const ServerlessFunctionType: React.FC = () => {
+  const { t } = useTranslation();
+  return (
+    <dl className="co-m-pane__details">
+      <dt>{t('knative-plugin~Type')}</dt>
+      <dd>{t('knative-plugin~Function')}</dd>
+    </dl>
```
christianvogt, line 10: Shouldn't we be using the `DetailsItem` component?
```
<DetailsItem label={t('knative-plugin~Type')}>{t('knative-plugin~Function')}</DetailsItem>
```

frontend/public/components/utils/details-page.tsx
christianvogt: You've removed this from the details view. However I do not see that you've added it back in the right column for secrets

nemesis09:
> You've removed this from the details view. However I do not see that you've added it back in the right column for secrets
This wasn't being used by any component.

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
nemesis09: The concern was to keep it specific to SinkUri. So we chose to use type `sink-uri` and annotaion seemed to be the proper place to use it.
cc @invincibleJai

frontend/packages/knative-plugin/src/topology/const.ts
```diff
@@ -19,3 +19,6 @@ export const KNATIVE_GROUP_NODE_PADDING = [
 ];
 
 export const EVENT_MARKER_RADIUS = 6;
+
+// Annotation to identify a specific type of resource bearing the same resource "Kind"
+export const typeAnnotation = 'knative.openshift.io/type';
```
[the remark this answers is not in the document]
nemesis09: done

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
invincibleJai: @nemesis09 we can go with `kind` as suggested, second thought was if in future any resource has kind `URI` which is not the case so far

nemesis09: updated to use kind

frontend/public/components/utils/details-page.tsx
nemesis09: this was never used by any component. Using this component would result in a bug being observed since it was rendering an object.
`SecretDetails` already receives a `secret.type`. I just updated it to use `SecretKind` instead of `K8sResourceKind`.

https://github.com/openshift/console/blob/dc102a74f4bbd98607c969b67fa13e1f31dbc6b1/frontend/public/components/secret.jsx#L116

invincibleJai:
> > You've removed this from the details view. However I do not see that you've added it back in the right column for secrets
> 
> Apparently, this wasn't being used by any component.

@nemesis09 it was shown in details look like see below 

image

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
christianvogt: Since this is a made up resource, we could make a more unique kind as well and even pair it with a unique api group.

frontend/packages/knative-plugin/src/components/overview/ServerlessFunctionType.tsx
christianvogt: Ok that's fine.

nemesis09:
added type for secret details page
Thanks for bringing it to notice @invincibleJai @christianvogt

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -1140,7 +1140,7 @@ export const createKnativeEventSourceSink = (
   }
   const eventSourceObj = _.omit(source, 'status');
   let sink = {};
-  if (NodeType.SinkUri === target.type?.nodeType) {
+  if (NodeType.SinkUri === target.kind) {
```
christianvogt, line 1143: `NodeType.SinkUri` has a value of `'sink-uri'`.

```suggestion
  if (URI_KIND === target.kind) {
```

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
```diff
@@ -1246,13 +1246,15 @@ export const createSinkPubSubConnection = (
   return createEventingPubSubSink(resources.obj, targetObj);
 };
 
-export const isServerlessFunction = (element: Node): boolean => {
+export const isServerlessFunction = (element: Node | K8sResourceKind): boolean => {
```
christianvogt, line 1249: Better to create two separate functions or change it to this.
```suggestion
export const isServerlessFunction = (resource: K8sResourceKind): boolean => {
```

Then when you have a topology node in hand, always call it like this `isServerlessFunction(getResource(node))`

frontend/public/components/secret.jsx
```diff
@@ -110,10 +111,18 @@ const SecretDetails = ({ obj: secret }) => {
           <div className="col-md-6">
             <ResourceSummary resource={secret} />
           </div>
+          {type && (
+            <div className="col-md-6">
+              <dl className="co-m-pane__details">
+                <dt>{t('public~Type')}</dt>
+                <dd>{type}</dd>
```
christianvogt, line 118: Shouldn't this be using `DetailsItem` since it's value directly from the resource? `DetailsItem` does extra work to potentially pull out descriptions from metadata.
```suggestion
                <DetailsItem label={t('public~Type')} obj={secret} path="type" />
```

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
nemesis09: done

frontend/packages/knative-plugin/src/topology/knative-topology-utils.ts
nemesis09: done

frontend/public/components/secret.jsx
nemesis09: done

christianvogt: Nice!
image

christianvogt:
/hold
consistent failure of cypress test on accessibility error

nemesis09:
pushed changes to address failing cypress tests
/hold cancel

PR #8518, opened 2021-03-30
Kafka Source creation form to support bootstrap server host from KC

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5657

**Solution Description**: 
Kafka Source creation form, show type ahead with bootstrap server from (Kafka connection) KC if the project has KC and user has access to them

image

image

**Test setup:**

- Install RHOAS operator / Serverless operator
- Create KC in ns with catalog
- Create Kafka Source Form

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/components/add/event-sources/KafkaSourceSection.tsx
+++ b/frontend/packages/knative-plugin/src/components/add/event-sources/KafkaSourceSection.tsx
@@ -13,22 +13,31 @@ import { EventSources } from '../import-types';
 
 interface KafkaSourceSectionProps {
   title: string;
+  namespace: string;
   fullWidth?: boolean;
 }
 
-const KafkaSourceSection: React.FC<KafkaSourceSectionProps> = ({ title, fullWidth }) => {
+const KafkaSourceSection: React.FC<KafkaSourceSectionProps> = ({ title, namespace, fullWidth }) => {
   const { t } = useTranslation();
-  const memoResources = React.useMemo(() => strimziResourcesWatcher(), []);
-  const { kafkas, kafkatopics } = useK8sWatchResources<{
+  const memoResources = React.useMemo(() => strimziResourcesWatcher(namespace), [namespace]);
+  const { kafkas, kafkatopics, kafkaconnections } = useK8sWatchResources<{
     [key: string]: K8sResourceKind[];
   }>(memoResources);
 
   const [bootstrapServers, bsPlaceholder] = React.useMemo(() => {
     let bootstrapServersOptions: SelectInputOption[] = [];
     let placeholder: React.ReactNode = '';
-    if (kafkas.loaded && !kafkas.loadError) {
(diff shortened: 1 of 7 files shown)


frontend/packages/knative-plugin/src/models.ts
```diff
@@ -367,3 +367,16 @@ export const CamelIntegrationPlatformModel: K8sKind = {
   color: knativeEventingColor.value,
   badge: BadgeType.TECH,
 };
+
+export const ManagedKafkaConnectionModel: K8sKind = {
```
vikram-raj, line 371: We can reuse the `KafkaConnectionModel` from `rhoas-plugin`.

invincibleJai:
@vikram-raj Have updated PTAL, Thanks 

cc @wtrocki

frontend/packages/knative-plugin/src/components/add/event-sources/KafkaSourceSection.tsx
```diff
@@ -39,7 +48,7 @@ const KafkaSourceSection: React.FC<KafkaSourceSectionProps> = ({ title, fullWidt
             },
           ];
       placeholder = t('knative-plugin~Add bootstrap servers');
-    } else if (kafkas.loadError) {
+    } else if (isKafkasLoadError) {
```
wtrocki, line 51: Since this depends on 2 resources should we also check message bellow if it is not undefined to see if error comes from Kafka source?

invincibleJai: `isKafkasLoadError` will true only we get err from both i.e `kafkas.loadError && kafkaconnections.loadError` so load error won't be undefined if there is err but yes it makes sense to show err from both i.e `kafkaconnections.loadError.message` as well. Have updated the code.

wtrocki:
I have one small functional suggestion: selecting multiple bootstrap servers when picking managed Kafka should be disabled. Reason for it is that for Managed Kafka there is always single bootstrap server. 

We do not want people to select different brokers by mistake as this might cause issues with consumers.

invincibleJai:
> I have one minor functional suggestion selecting multiple bootstrap servers when picking KafkaConnection can be disabled. The reason for it is that for KafkaConnection there is always a single bootstrap server.
> 
> So when typing bootstrap servers devs can possibly put multiple ones, but selection using KafkaConnection will be typically replacing existing one or disabling update until other is removed.
> 
> We do not want people to select different brokers by mistake as this might cause issues with consumers.

Thanks @wtrocki, that's a valid point. will discuss this on Serverless call and address this as separate PR/ticket if needed

vikram-raj (commented):
Verified it. It works as expected.

/lgtm

vikram-raj (commented):
/lgtm

PR #8521, opened 2021-03-30
Fix remove trigger issue

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5700

**Analysis / Root cause**: 

While removing the trigger that has template.ref in the eventlistener fails with an NPE and not able to remove the trigger succesfully.

**Solution Description**: 

1. Handled both template.name and template.ref incase of trigger removal.
2. Cleaned up the new trigger resource flow and reverting back to template.ref / template.name based on operator version.

No UI changes.

  removeTrigger
    ✓ expect to remove the TT and EL resources when template.name is used in Eventlistener (4ms)
    ✓ expect to remove the TT and EL resources when template.ref is used in Eventlistener (1ms)

 submitTrigger
    ✓ expect to use template.name when pipeline GA operator is not installed (27ms)
    ✓ expect to use template.ref when pipeline GA operator is installed (3ms)

  EventListener Details
    ✓ should render EventListenerTriggers section if the trigger contains binding ref (20ms)
   

**Test setup:**

1. Install OSP 1.4 operator and add trigger to a pipeline

/kind bug
cc: @andrewballantyne @vikram-raj


frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/__tests__/EventListenerDetails.spec.tsx
```diff
@@ -28,9 +28,9 @@ describe('EventListener Details', () => {
     );
   });
 
-  it('should not render EventListenerTriggers section if the trigger contains binding ref', () => {
+  it('should render EventListenerTriggers section if the trigger contains binding ref', () => {
     wrapper.setProps({ obj: EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_REF] });
-    expect(wrapper.find(EventListenerTriggers).exists()).toBe(false);
+    expect(wrapper.find(EventListenerTriggers).exists()).toBe(true);
   });
```
andrewballantyne, line 35: I think we need another test here for `name` to successfully find `EventlistenerTriggers`

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/__tests__/remove-utils.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    try {
+      setupJestMocks(EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_REF]);
+      await removeTrigger(values, pipelineData.pipeline);
+      expect(jest.spyOn(k8s, 'k8sKill')).toHaveBeenCalledTimes(2);
+      expectAndDeleteElResources(EventlistenerTestData[EventlistenerTypes.BINDINGS_TEMPLATE_REF]);
+    } catch (e) {
+      fail(e);
+    }
+  });
+});
```
andrewballantyne, line 66: Need another test for when there are multiple TTs tied to the same EventListener. [We don't delete EventListeners if they had more than one TT](https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/remove-utils.ts#L27-L30).

Also maybe another test on top of that one... for two EventListeners tied to the same TT.

Thanks for writing these tests, should have done them myself when I coded it up 😅

frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/submit-utils.ts
```diff
@@ -71,47 +68,27 @@ export const submitTrigger = async (
     pipelineRun,
     triggerTemplateParams,
   );
+  const eventListener: EventListenerKind = await createEventListener(
```
andrewballantyne, line 71: I know this predates this PR, but let's not keep this an async method. Do the fetching of the operator version out here and pass it into `createEventListener` instead. Just so the method is more pure with its usage.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
```diff
@@ -1,23 +1,6 @@
 import { K8sResourceCommon, K8sResourceKind, Toleration } from '@console/internal/module/k8s';
 import { PipelineRunKind } from '../../../types';
 
-export type TriggerKind = K8sResourceCommon & {
```
andrewballantyne, line 4: Yikes, this file really should exist under types. My bad! I missed it when I was doing the type refactors.

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
andrewballantyne: I'll leave this up to you if you want to do this part of your PR. If not, we can just put up another PR to address it.

andrewballantyne:
Also, don't forget to fix the test we talked about in `frontend/packages/pipelines-plugin/src/components/pipelines/modals/triggers/__tests__/submit-utils.spec.ts`... perhaps just mock out `exposeRoute`.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-types/triggers.ts
karthikjeeyar: will do in separate PR.

vikram-raj (commented):
verified the changes. Thanks for fixing the missing i18n key error.

/lgtm

PR #8522, opened 2021-03-30
Support URLs in Result values

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5711

**Analysis / Root cause**: 
When results had URLs they were just presented like strings. This is apparently a common use-case that we would like to support.

**Solution Description**: 
When a URL is present, make it clickable.

With clickable link:
Screen Shot 2021-03-30 at 5 06 59 PM

And the trigger settings (that generated it):
Screen Shot 2021-03-30 at 5 07 39 PM

More complicated example:
Screen Shot 2021-04-01 at 12 05 20 PM

```
 PASS  packages/pipelines-plugin/src/utils/__tests__/render-utils.spec.tsx (17.368s)
  handleURLs
    ✓ should return the same value if it is not a string (2ms)
    ✓ should not do anything if there are no URLs in the string
    Test easy URL Examples
      ✓ should create an ExternalLink for the URL, test straightURL (11ms)
      ✓ should create an ExternalLink for the URL, test prefixedURL (4ms)
      ✓ should create an ExternalLink for the URL, test suffixedURL (3ms)
      ✓ should create an ExternalLink for the URL, test bothPrefixAndSuffixURL (2ms)
      verify backing RegExp finds the urls
        ✓ should find the URL, test straightURL
        ✓ should find the URL, test prefixedURL
        ✓ should find the URL, test suffixedURL
        ✓ should find the URL, test bothPrefixAndSuffixURL
    Test edge-case URL Examples
      ✓ should create multiple ExternalLinks and not lose the interim prefix/suffix values (2ms)
      ✓ should create multiple ExternalLinks when more than one url is present (3ms)
      ✓ should handle escaped URLs (such as googling for a URL) (3ms)
      ✓ should handle redirect wrapper URLs (2ms)
```

**Test setup:**

* Have OpenShift Pipelines Operator installed.
* My sample resources from the screenshots: https://raw.githubusercontent.com/andrewballantyne/pipeline-scratch-pad/master/examples/results/resources.yaml

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/utils/__tests__/render-utils.spec.tsx
+++ b/frontend/packages/pipelines-plugin/src/utils/__tests__/render-utils.spec.tsx
@@ -1,0 +1,100 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import { ExternalLink } from '@console/internal/components/utils';
+import { handleURLs, GROUP_MATCH_REGEXP } from '../render-utils';
+
+describe('handleURLs', () => {
+  it('should return the same value if it is not a string', () => {
+    // We will only likely get strings, but it shouldn't break/NPE if they are not
+    expect(handleURLs(null)).toBe(null);
+    expect(handleURLs(undefined)).toBe(undefined);
+    expect(handleURLs(true as any)).toBe(true);
+    const v = {};
+    expect(handleURLs(v as any)).toBe(v);
+  });
+
+  it('should not do anything if there are no URLs in the string', () => {
+    const stringsWithoutURLs = ['not a URL', 'redhat.com', 'http', '://something.com'];
+    stringsWithoutURLs.forEach((string: string) => {
+      expect(handleURLs(string)).toBe(string);
+    });
+  });
+
+  describe('Test easy URL Examples', () => {
+    const validStringsWithURL: { [testName: string]: string } = {
+      straightURL: 'https://redhat.com',
+      prefixedURL: 'Red Hat website: https://redhat.com',
+      suffixedURL: "https://redhat.com is Red Hat's website",
+      bothPrefixAndSuffixURL: 'This is the company website https://redhat.com for Red Hat',
+    };
+
+    Object.keys(validStringsWithURL).forEach((testName: string) => {
+      const string = validStringsWithURL[testName];
+      it(`should create an ExternalLink for the URL, test ${testName}`, () => {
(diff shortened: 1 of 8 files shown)


frontend/packages/pipelines-plugin/src/utils/render-utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const [, prefix, link, suffix] = matches || [];
+
+  if (link) {
+    return (
+      <>
+        {prefix}
+        <ExternalLink additionalClassName="opp-render-utils-external-url" href={link}>
+          {link}
+        </ExternalLink>
+        {suffix}
```
karthikjeeyar, line 22: Two things I noticed here
1. Suffix is incorrect
expected: 
`Visit http://example.com for logs`
Actual:
image

2. if there are multiple urls only one gets wrapped by external link other is a plain string
image

Note: There needs to be a space between the link and the suffix part.

frontend/packages/pipelines-plugin/src/utils/__tests__/render-utils.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  describe('verify backing RegExp finds the urls', () => {
+    Object.keys(validStringsWithURL).forEach((testName: string) => {
+      const string = validStringsWithURL[testName];
+      it(`should find the URL for, test ${testName}`, () => {
+        const [, , url] = string.match(GROUP_MATCH_REGEXP);
+        expect(url).toBe('https://redhat.com');
+      });
+    });
+  });
```
karthikjeeyar, line 48: one more test cases to test preffix & suffix value will be good.

frontend/packages/pipelines-plugin/src/utils/render-utils.tsx
andrewballantyne: Yup, this is caused by my laziness lol. I used a fake RegExp to capture a URL, then googled for another one before I submitted. Re-ran my tests but didn't visually test it 🤦 

Thanks for the review 👍

andrewballantyne:
Fixed the comments you made @karthikjeeyar. Thanks for the review! Updated the tests for the missed issues.

karthikjeeyar:
/lgtm

PR #8527, opened 2021-03-31
Create component to visualize copy/execute snippets in the Quickstarts

sahil143:
Fixes: https://issues.redhat.com/browse/ODC-5653

Created showdown extensions and components to render the inline and multiline snippets for copy to clipboard and execute snippets in cloudshell.

snippet
Screenshot 2021-03-31 at 8 58 24 PM

[TODO]:

- [x] Add Multiline snippets
- [x] Unit tests


frontend/packages/console-shared/src/components/markdown-extensions/MarkdownExecuteSnippet.tsx
```diff
@@ -0,0 +1,91 @@
...
+type MarkdownExecuteCommandProps = {
+  docContext: HTMLDocument;
+  rootSelector: string;
+};
+
+const MarkdownExecuteSnippet: React.FC<MarkdownExecuteCommandProps> = ({
+  docContext,
+  rootSelector,
+}) => {
+  const elements = docContext.querySelectorAll(`${rootSelector} [${MARKDOWN_EXECUTE_BUTTON_ID}]`);
```
christianvogt, line 56: Use memo so that you don't query in render flow.

frontend/packages/console-shared/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+type TooltipProps = {
+  reference: Element | (() => HTMLElement);
+  className?: string;
+  placement?: TooltipPlacement;
+  onShow?: (event: Event) => void;
+  onHide?: (event: Event) => void;
+  children: React.ReactNode;
+};
+
+export const Tooltip: React.FC<TooltipProps> = ({ children, className, ...rest }) => {
```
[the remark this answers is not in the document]
sahil143: PF Tooltip doesn't have hide/show events and also has an arrow positioning issue when tooltip content is updated.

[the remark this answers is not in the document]
christianvogt: Please log these as issues on patternfly Tooltip. We can use your custom tooltip in the meantime.
Add comments as to why this component exists with reference to the bugs and that we should remove this component when we migrate back to PF. Make sure the API is consistent.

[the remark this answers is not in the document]
sahil143: created issue https://github.com/patternfly/patternfly-react/issues/5620

frontend/packages/console-shared/src/components/markdown-extensions/showdown-extensions.ts
```diff
@@ -0,0 +1,168 @@
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import useCloudShellAvailable from '@console/app/src/components/cloud-shell/useCloudShellAvailable';
+import { MARKDOWN_COPY_BUTTON_ID, MARKDOWN_EXECUTE_BUTTON_ID, MARKDOWN_SNIPPET_ID } from './const';
+
+import './showdown-extension.scss';
+
+export const useInlineCopyClipboardShowdownExtension = () => {
+  const { t } = useTranslation();
+  return {
```
christianvogt, line 10: return value should use `useMemo`

frontend/packages/console-shared/src/components/markdown-extensions/showdown-extensions.ts
```diff
@@ -0,0 +1,168 @@
...
+            </div>`;
+      },
+    }),
+    [showExecuteButton, t],
+  );
+};
+
+export const useMultilineCopyClipboardShowdownExtension = () => {
+  const { t } = useTranslation();
+  return {
```
christianvogt, line 84: return value should use `useMemo`

frontend/public/components/markdown-view.tsx
```diff
@@ -9,16 +9,23 @@ import './_markdown-view.scss';
 
 const tableTags = ['table', 'thead', 'tbody', 'tr', 'th', 'td'];
 
-const markdownConvert = (markdown, extensions?: string[]) => {
-  const unsafeHtml = new Converter({
+type ShowdownExtension = {
+  type: string;
+  regex?: RegExp;
+  replace?: (...args: any[]) => string;
+};
```
christianvogt, line 16: Can we add a `render` prop to each extension such that we have a single means for contributing extensions rather than separating the rendering from the showdown extension?
```suggestion
type ShowdownExtension = {
  type: string;
  regex?: RegExp;
  replace?: (...args: any[]) => string;
  render?: (contentDocument: HTMLDocument, rootSelector: string) => React.ReactNode;
};
```

frontend/packages/console-shared/src/components/markdown-extensions/showdown-extensions.ts
```diff
@@ -0,0 +1,168 @@
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import useCloudShellAvailable from '@console/app/src/components/cloud-shell/useCloudShellAvailable';
+import { MARKDOWN_COPY_BUTTON_ID, MARKDOWN_EXECUTE_BUTTON_ID, MARKDOWN_SNIPPET_ID } from './const';
+
+import './showdown-extension.scss';
+
+export const useInlineCopyClipboardShowdownExtension = () => {
```
christianvogt, line 8: I'd prefer to have a single file per extension
`inline-clipboard-extension.ts`, `multiline-clipboard-extension.ts` etc...

frontend/packages/console-shared/src/components/markdown-extensions/const.ts
```diff
@@ -0,0 +1,3 @@
+export const MARKDOWN_COPY_BUTTON_ID = 'data-copy-for';
+export const MARKDOWN_EXECUTE_BUTTON_ID = 'data-execute-for';
+export const MARKDOWN_SNIPPET_ID = 'data-snippet-id';
```
christianvogt, line 3: These consts could move to their respective extension file.

frontend/packages/console-shared/src/components/markdown-extensions/MarkdownExecuteSnippet.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const showExecuteButton = useCloudShellAvailable();
+
+  return elements.length > 0 && showExecuteButton ? (
+    <>
+      {Array.from(elements).map((elm) => {
+        const attributeValue = elm.getAttribute(MARKDOWN_EXECUTE_BUTTON_ID);
+        return (
+          <ExecuteSnippet
+            key={`${MARKDOWN_EXECUTE_BUTTON_ID}-${attributeValue}`}
```
christianvogt, line 76: Sufficient to provide the `attributeValue` since it is already unique. No need to make the key longer with a prefix.
```suggestion
            key={attributeValue}
```

frontend/public/components/markdown-view.tsx
sahil143: I'm not sure if that's a good idea. We will be rendering the component multiple times with each extension and these extensions are directly fed into the showdown converter.

frontend/packages/console-shared/src/components/markdown-extensions/const.ts
sahil143: These consts are being shared by all of the extensions. I think it's good to keep them in a separate file.

christianvogt: Sorry when i made the comment I didn't realize they were shared. I thought I had deleted this comment afterwards :) I guess I didn't!

frontend/public/components/markdown-view.tsx
christianvogt: I want to combine `renderExtension` and `extensions` into a single prop on the markdown component.
It's odd that the showdown extension and the rendering of the extension are disconnected.

What if we did this:
```
type MarkdownExtension = {
  showdownExtension: {
    type: string;
    regex?: RegExp;
    replace?: (...args: any[]) => string;
  },
  render?: (contentDocument: HTMLDocument, rootSelector: string) => React.ReactNode;
};
```

Can you explain what you mean by:
> We will be rendering the component multiple times with each extension

Instead of accepting a `renderExtensions` prop we can now loop over the extensions and execute the `render` prop.

frontend/packages/console-shared/src/components/markdown-extensions/MarkdownExecuteSnippet.tsx
rohitkrai03: @sahil143 Seems like you missed to resolve this one.

rohitkrai03 (commented):
Overall looks good to me. Works as expected. But I agree with @christianvogt's comment about combining the API for extensions and renderers. Maybe that can be a follow up PR if we want to get this merged in this sprint? WDYT @christianvogt?

frontend/packages/console-shared/src/components/markdown-extensions/MarkdownExecuteSnippet.tsx
sahil143: Sorry missed this one. I tried using the useMemo for this but the elements are not present if I use the useMemo here because the first render document doesn't have the queried elements whereas it is present in the next renders but since the dependency doesn't know that document is changed useMemo is not triggered.

frontend/public/components/markdown-view.tsx
christianvogt: discussed with sahil and we will look into this in the future

frontend/packages/console-shared/src/components/markdown-extensions/MarkdownExecuteSnippet.tsx
christianvogt: @sahil143 please create an issue to track this. The render extensions must be rendered only once the content is present on the page. This must be a bug from when we changed from iframe to inline content. 
We should investigate.

christianvogt:
/lgtm
/approve

followup PRs will be made for improving the API for extensions with render components and another to investigate why the render component cannot find elements on first render

PR #8534, opened 2021-03-31
Bug 1916116: Pf update to fix NavExpandable component

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1916116

NavExpandable wasn't firing off expand events, this update fixes that problem and the above bug, and updates pf to the 2021-4 release.


rhamilto:
/hold 

This does appear to fix the bug at hand; however, it seems to introduce a regression where the nav section collapses when navigating using a nav section item.

https://user-images.githubusercontent.com/895728/113188875-9620a100-9228-11eb-8315-691c0ff7f969.mov

redallen:
This callback is now properly being fired:
https://github.com/openshift/console/blob/master/frontend/public/components/nav/section.tsx#L236

You can remove it to fix the regression.

rhamilto:
Tested and the regression is fixed.  🥇 Mind squashing and I'll tag?

jcaianirh:
/hold remove
Pf fix verified
pr squashed @rhamilto

jcaianirh:
/hold cancel
Pf fix verified
pr squashed @rhamilto

rhamilto (approved):
/lgtm

PR #8538, opened 2021-04-01
Embedded Tasks in Pipeline Builder

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5612

**Analysis / Root cause**: 
We currently support Embedded Tasks outside of the Pipeline Builder to a small extent (visualization and the UI doesn't crash). We need to support it in the Pipeline Builder.

**Solution Description**: 
Support Embedded Tasks in the Pipeline Builder by taking the `taskSpec` and wrapping it in a fake TaskKind resource ... thus allowing the rest of the Pipeline Builder to work unimpeded.

Also fixed some types that were overlapping / redundant.

Screen Shot 2021-03-31 at 8 36 24 PM
Screen Shot 2021-03-31 at 8 36 58 PM

cc @bgliwa01 Not a lot to show, but FYI.

New tests...
```
 PASS  packages/pipelines-plugin/src/components/pipelines/pipeline-builder/__tests__/validation-utils.spec.ts
  Pipeline Build validation schema
    Tasks and ListTasks validation
      Validate Run Afters
        ✓ should pass if runAfter is on a taskSpec task (4ms)
      Validate Parameters
        ✓ should fail if the taskSpec params are required and not provided (6ms)
        ✓ should pass if the taskSpec params have defaults (4ms)
        ✓ should pass if the taskSpec params are required but a value is provided (3ms)
      Validate Resources
        ✓ should fail if the taskSpec has resources but the task does not (4ms)
        ✓ should pass if the taskSpec has resources and the task provides it (7ms)
      Validate Workspaces
        ✓ should fail if the taskSpec requests a workspace (4ms)
        ✓ should pass if the taskSpec requests a workspace and it is provided (3ms)
```
```
 PASS  packages/pipelines-plugin/src/utils/__tests__/pipeline-augment.spec.ts
  getResourceModelFromTaskKind
    ✓ should handle null
    ✓ should be able to find ClusterTaskModel
    ✓ should be able to find TaskModel (1ms)
    ✓ should return the TaskModel for undefined
    ✓ should return null for any unknown value
  getResourceModelFromBindingKind
    ✓ should handle null (1ms)
    ✓ should be able to find ClusterTriggerBindingModel
    ✓ should be able to find TriggerBindingModel
    ✓ should return TriggerBindingModel for undefined
    ✓ should return null for any unknown value
```

Removed tests in `frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-utils.spec.ts` around `hasInlineTaskSpec` as the utility is no longer used.

**Test setup:**

Add a `taskSpec` based task in the YAML view (or edit an existing Pipeline). Then interact with it in the visualization. Everything should work as if it was powered by a stand-alone Task resource.

eg. Sample Pipeline:
```
kind: Pipeline
metadata:
  name: new-pipeline
spec:
  tasks:
    - name: echo-good-morning
      taskSpec:
        params:
          - name: name
            description: Your name
        workspaces:
          - name: input
        steps:
          - name: echo
            image: ubuntu
            script: |
              #!/usr/bin/env bash
              echo "Good Morning, $(params.name)!"
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
+++ b/frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
@@ -158,7 +158,6 @@
   "Add workspace": "Add workspace",
   "Error loading the tasks.": "Error loading the tasks.",
   "Unable to locate any tasks.": "Unable to locate any tasks.",
-  "This Pipeline cannot be visualized. Pipeline taskSpec is not supported.": "This Pipeline cannot be visualized. Pipeline taskSpec is not supported.",
   "Input resources": "Input resources",
   "Output resources": "Output resources",
   "Display Name": "Display Name",
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/pipeline-details/PipelineVisualizationTask.tsx
@@ -12,7 +12,7 @@ import {
   truncateMiddle,
 } from '@console/internal/components/utils';
 import { SvgDropShadowFilter } from '@console/topology/src/components/svg';
-import { PipelineTaskSpec, PipelineTaskRef, TaskKind } from '../../../../types';
+import { TektonTaskSpec, PipelineTaskRef, TaskKind } from '../../../../types';
 import { runStatus, getRunStatusColor } from '../../../../utils/pipeline-augment';
 import { PipelineRunModel, TaskModel, ClusterTaskModel } from '../../../../models';
 import { StatusIcon } from './StatusIcon';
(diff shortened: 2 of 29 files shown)


debsmita1:
In the task sidebar, for Tasks and Cluster Tasks we show the display name with the resource icon in the header. Wondering why we are showing `Embedded Task` and not the display name of that embedded task

andrewballantyne:
> In the task sidebar, for Tasks and Cluster Tasks we show the display name with the resource icon in the header. Wondering why we are showing `Embedded Task` and not the display name of that embedded task

@debsmita1 It's actually not the display name, it's more of a common case so it appears that way. It's the `metadata.name` of the TaskKind. If you change your alias in the `Display Name` field, it does not change the header value. For instance... Add two of the same task and look at the one we generated the suffix for.

debsmita1:
I have tried with a few pipeline resource yamls having embedded taskSpec & it works fine
will wait for UX to approve

bgliwa01 (approved):
lgtm

karthikjeeyar:
Verified locally, works as expected :clap: . Will add lgtm once the conflict is resolved.

PR #8546, opened 2021-04-02
Implement externalLink option for the org.w3 alm x-descriptor

### LINKED ISSUE #8477: Implement externalLink option for the org.w3 alm x-descriptor
It would be prefer to open link configured with following OLM descriptor in new window/tab as users always lives Console application in case of click on the provide link
However, changing default behavior for the standard configuration, i.e. `x-descriptors:urn:alm:descriptor:org.w3:link` could have an impact on other applications and configurations
As discussed in #8349 it would be nice to implement this functionality like `x-descriptors:urn:alm:descriptor:org.w3:externalLink`

dmvolod:
It would be prefer to open link configured with following OLM descriptor in new window/tab as users always lives Console application in case of click on the provide link
However, changing default behavior for the standard configuration, i.e. `x-descriptors:urn:alm:descriptor:org.w3:link` could have an impact on other applications and configurations
Implement functionality for support `x-descriptors:urn:alm:descriptor:org.w3:externalLink` option

fixes #8477

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md
+++ b/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md
@@ -27,11 +27,12 @@
 #### [statusDescriptors](#statusDescriptors-1):
   1. [podStatuses](#1-podStatuses)
   2. [w3Link](#2-w3Link)
-  3. [conditions](#3-conditions)
-  4. [text](#4-text)
-  5. [k8sPhase](#5-k8sPhase)
-  6. [k8sPhaseReason](#6-k8sPhaseReason)
-  7. [k8sResourcePrefix](#7-k8sResourcePrefix)
+  3. [w3ExternalLink](#3-w3ExternalLink)
+  4. [conditions](#4-conditions)
+  5. [text](#5-text)
+  6. [k8sPhase](#6-k8sPhase)
+  7. [k8sPhaseReason](#7-k8sPhaseReason)
+  8. [k8sResourcePrefix](#8-k8sResourcePrefix)
 ---------------------------------------
 
 #### [DEPRECATED Descriptors](#deprecated-descriptors-1)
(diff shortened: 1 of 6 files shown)


dmvolod:
@TheRealJon could you please to have a look at this PR
Thanks in advance

frontend/packages/operator-lifecycle-manager/src/components/descriptors/status/index.tsx
```diff
@@ -64,6 +64,27 @@ const StatusConditions: React.FC<StatusCapabilityProps> = ({
...
+  obj,
+  value,
+}) => {
+  const { t } = useTranslation();
+  return (
+    <DetailsItem description={description} label={label} obj={obj} path={fullPath}>
+      {!_.isNil(value) ? (
+        <a href={value} rel="noopener noreferrer" target="_blank">
+          {value.replace(/https?:\/\//, '')}
+        </a>
```
TheRealJon, line 80: Reuse existing ExternalLink component. as it has specific styling. Also, don't remove scheme from URL. https://github.com/openshift/console/blob/67128172502d43de3ece30e7b11818006ec9d752/frontend/public/components/utils/link.tsx#L68

```suggestion
        <ExternalLink href={value}>
          {value}
        </a>
```

frontend/packages/operator-lifecycle-manager/src/components/descriptors/status/index.tsx
```diff
@@ -64,6 +64,27 @@ const StatusConditions: React.FC<StatusCapabilityProps> = ({
   );
 };
 
+const ExternalLink: React.FC<StatusCapabilityProps> = ({
```
TheRealJon, line 67: You'll either need to rename this or the imported ExternalLink component

frontend/packages/operator-lifecycle-manager/src/components/descriptors/status/index.spec.tsx
```diff
@@ -89,6 +89,14 @@ describe('Status descriptors', () => {
     expect(wrapper.find('dd').text()).toEqual('example.com');
   });
 
+  it('renders an external link status', () => {
+    descriptor['x-descriptors'] = [StatusCapability.w3ExternalLink];
+    descriptor.path = 'externalLink';
+    wrapper = wrapper.setProps({ descriptor });
+
+    expect(wrapper.find('dd').text()).toEqual('example.com');
```
TheRealJon, line 97: You'll need to update this to look for the full URL with scheme once you've made the other updates I suggested.

frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md
```diff
@@ -853,7 +854,55 @@ status:
...
+    <td width="50%">
+      <img src="img/status/2-1_w3Link.png" />
+      <h6><small>Click on the link sends users to target URL.</small></h6>
+      </td>
+    <td width="50%">
+      <img src="img/status/2-2_w3Link.png" />
+      <h6><small>Hover over displayName shows descriptions on tooltip.</small></h6>
+      </td>
+  </tr>
+</table>
```
TheRealJon, line 903: We'll need to update these with new screenshots after the other suggested updates. ExternalLinks are styled differently.

TheRealJon (changes requested):
We should reuse our existing ExternalLink component and not remove the url scheme:

https://github.com/openshift/console/blob/67128172502d43de3ece30e7b11818006ec9d752/frontend/public/components/utils/link.tsx#L68

TheRealJon:
@dmvolod Looks like there is a merge conflict on this. Could you rebase on master and force push?

kberkos-public:
Hi, has this feature been implemented? It is an issue that is causing some headache for us also.

PR #8552, opened 2021-04-04
Fix to align with changes in CNV 2.6 UI

ibragins:


frontend/integration-tests/views/crud.view.ts
```diff
@@ -7,7 +7,8 @@ import { appHost, testName, waitForNone } from '../protractor.conf';
 import { waitForCount } from '@console/shared/src/test-utils/utils';
 
 export const createYAMLButton = $('#yaml-create');
-export const createItemButton = $('#item-create');
```
yaacov, line 10: are you sure this change is correct for all tests that use this const ?

frontend/integration-tests/views/crud.view.ts
```diff
@@ -68,10 +69,14 @@ export const filterForName = async (name: string) => {
 };
 
 const actionOnKind = (action: string, kind: string) => {
-  return `${action} ${kind}`;
+  const humanizedKind = (kind.includes('~') ? kind.split('~')[2] : kind)
```
yaacov, line 72: are you sure kind should be split ?
"ConfigMap" is "ConfigMap" not "Config Map" maybe there is a typo and someone wrote "Config Map"  instead of "ConfigMap" , if this is the case, we need to fix the code and not the test.

frontend/integration-tests/views/crud.view.ts
```diff
@@ -68,10 +69,14 @@ export const filterForName = async (name: string) => {
...
+    .split(/(?=[A-Z])/)
+    .join(' ');
+
+  return `${action} ${humanizedKind}`;
 };
-export const editKind = (kind: string) => actionOnKind(actions.edit, kind);
```
yaacov, line 73: are you sure no other test use this consts ?

frontend/integration-tests/views/crud.view.ts
ibragins: This is strange, cause actually I didn't do this change here

frontend/integration-tests/views/crud.view.ts
ibragins: Didn't touch this, don't know how this got to PR

yaacov:
/hold
need some code changes, ping me when fixed

frontend/integration-tests/views/crud.view.ts
ibragins: Old selector doesn't exist any more, so it is OK to change this

frontend/packages/kubevirt-plugin/integration-tests/tests/mocks/mocks.ts
```diff
@@ -49,6 +49,21 @@ export const multusNAD = {
...
+export const sriovNAD = {
+  apiVersion: 'k8s.cni.cncf.io/v1',
+  kind: 'NetworkAttachmentDefinition',
+  metadata: {
+    name: `sriov-${testName}`,
+    namespace: testName,
+    labels: { automatedTest: testName },
+  },
+  spec: {
+    // config: '{ "cniVersion": "0.3.1", "type": "cnv-bridge", "bridge": "testbridge", "ipam": {} }',
```
yaacov, line 61: if its wrong you don't need to leave it as a comment, just remove it

This also a comment for the rest of the places you left the old code as a comment - if it's wrong: delete it, if it's not wrong then don't comment it out :-)

yaacov:
/retest

@ibragins hi, is this still relevant (created Apr-4) or fixed by other PR ?

ibragins:
This is not fixed, but actually these are tests for the features that are
deprecated, so not relevant any more


Igor Braginsky

Software Quality Engineer, MTV Migration

Red Hat <https://www.redhat.com>

***@***.***
M: +972526922863
<https://www.redhat.com>

yaacov:
> tests for the features that are deprecated, so not relevant any more

/close

PR #8559, opened 2021-04-05
Bug 1945630: Fix default resource log download file name

TheRealJon:
Update download link in ResourceLog component so that the default file name will be `${podName}-${containerName}.log`


spadgett (approved):
/lgtm

PR #8560, opened 2021-04-05
Bug 1888521: Add support for adding dynamic storage providers for a Deployment

bipuladh:
Screenshot from 2021-04-15 14-45-42


frontend/public/components/storage/attach-pvc-storage.tsx
```diff
@@ -0,0 +1,436 @@
+import * as _ from 'lodash-es';
```
bipuladh, line 1: Moved `attach-storage.tsx` to `attach-pvc-storage.tsx`. Shows as a new file because `attach-storage.tsx` was then used to create the parent component which accepts extensions.

frontend/packages/ceph-storage-plugin/src/components/attach-obc/attach-obc-deployment.tsx
```diff
@@ -0,0 +1,150 @@
...
+            id="exists"
+            name="exists"
+            isChecked={!createOBC}
+          />
+          {!createOBC && (
+            <div className="ceph-attach-obc__subgroup">
+              <ListDropdown
+                resources={[{ kind: referenceForModel(NooBaaObjectBucketClaimModel), namespace }]}
+                selectedKeyKind={referenceForModel(NooBaaObjectBucketClaimModel)}
+                placeholder="Select claim"
```
christianvogt, line 86: Localize this and all your new strings.

frontend/public/components/storage/attach-storage.scss
```diff
@@ -0,0 +1,8 @@
+.co-storage {
+  &-radio {
+    margin-bottom: var(--pf-global--spacer--lg);
+    &__item {
+      margin-bottom: var(--pf-global--spacer--sm);
+    }
```
christianvogt, line 6: Why is this required? Is the default PF wrong? If we disagree with it then we should have a bug raised on PF. But otherwise we should be using their default spacing.

frontend/public/components/storage/attach-storage.tsx
```diff
@@ -1,267 +1,55 @@
...
-    if (!volumeAlreadyMounted) {
-      const existingVolumes = _.get(obj, 'spec.template.spec.volumes');
-      const volumePatch = _.isEmpty(existingVolumes)
-        ? { op: 'add', path: '/spec/template/spec/volumes', value: [volume] }
-        : { op: 'add', path: '/spec/template/spec/volumes/-', value: volume };
-      return [...patches, volumePatch];
-    }
-    return patches;
-  };
+  React.useEffect(() => {
```
christianvogt, line 39: This should be `useMemo` instead of an effect.

frontend/public/components/storage/attach-storage.scss
bipuladh: There is no spacing between radio items. 
Before:
Screenshot from 2021-04-12 17-03-57
After:
Screenshot from 2021-04-12 17-04-07

christianvogt: This seems to be the design intent from patternfly (https://patternfly-react.surge.sh/components/radio):
image

Whether we agree or not, we should be using PF styles. As I said, perhaps this is something we can bring up with PF to add spacing between adjacent radios.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/storage-provider.ts
```diff
@@ -0,0 +1,14 @@
+import { RouteComponentProps } from 'react-router';
+import { Extension } from '@console/plugin-sdk/src/typings/base';
+import { ExtensionDeclaration, CodeRef } from '../types';
+
+export type StorageProvider = ExtensionDeclaration<
+  'console.storage-provider',
+  {
+    name: string;
+    Component: CodeRef<React.FC<RouteComponentProps>>;
```
christianvogt, line 9:
```suggestion
    Component: CodeRef<React.ComponentType<RouteComponentProps>>;
```

I wonder if it should be `Partial<RouteComponentProps>` instead.
cc @vojtechszocs

christianvogt: Maybe we should re-export `RouteComponentProps` so that to use the extension, the dev doesn't need to also depend on `react-router`. This might be a better solution

christianvogt: `export interface RouteComponentProps<Params extends { [K in keyof Params]?: string } = {}, C extends StaticContext = StaticContext, S = H.LocationState> {`

Sorry... re-exporting isn't going to solve the problem I thought about in terms of eliminating the dep on react-router. Not when the type itself has so many other react-router types in it.

christianvogt: If you just go with what you originally had, we can also drop down to `Component: CodeRef<React.ComponentType<RouteComponentProps>>;` in the future.
So i think my first suggestion is the best.

cloudbehl (commented):
nits

cloudbehl:
Tested the PR and works well!

PR #8561, opened 2021-04-05
Bug 1946320: Fix empty state message for PVC dropdown

bipuladh:


frontend/public/components/utils/list-dropdown.jsx
```diff
@@ -158,7 +158,7 @@ class ListDropdown_ extends React.Component {
             isInline
             className="co-alert pf-c-alert--top-margin"
             variant="info"
-            title={`No ${desc} found or defined`}
+            title={emptyMessage ?? `No ${desc} found or defined`}
```
spadgett, line 161: We should make sure we never show this message if `desc` is not defined. Is it possible to use `desc` is this context, too? Do we need a separate empty message?

Not new, but I would remove "or defined" from this message and leave it as `No ${desc} found`.

bipuladh: empty message would allow us to give a very specific message for empty state like when PVCs are not present in particular namespace. But for now just hiding the alert if no kind or desc is present.

spadgett: > empty message would allow us to give a very specific message for empty state like when PVCs are not present in particular namespace.

We probably want that specific of a message by default with no special prop. `No {kind} found in namespace {namespace}` would be a good default message for namespaced resources.

spadgett (approved):
/lgtm

PR #8566, opened 2021-04-06
Bug 1946449: fix cloud init tests as UI changed

gouyang:
- remove workload/flavor set if it's using common template
- click cloud-init to open the cloud-init form before the tests


gouyang:
changes in protractor should not affect cypress.
/retest

PR #8568, opened 2021-04-06
Bug 1937717: Set a kebab column in the template list

yaacov:
In the template list view, when char size is large, the browser auto arrange re-order the actions and put the last action "kebab" in lower line.
This PR gives the kebab a full column so it will always be first in it's column.

Before:
screenshot-localhost_9000-2021 04 06-12_10_52

After:
screenshot-localhost_9000-2021 04 06-12_17_53


yaacov:
@glekner please review

PR #8569, opened 2021-04-06
Add support for Alert Action in Dynamic SDK

bipuladh:
- Add support for Modals in Alert Action
Modal: 
CPT2104061432-1511x913
Link: 
Screenshot from 2021-04-06 14-48-57


frontend/packages/console-dynamic-plugin-sdk/src/extensions/notification-alert.ts
```diff
@@ -0,0 +1,18 @@
...
+  'console.alert-action',
+  {
+    /* Alert name as defined by `alert.rule.name` property */
+    alert: string;
+    /* Action text */
+    text: string;
+    /* Action is a link or a function that performs side effects(Ex: Opens a modal). */
+    isLink: boolean;
+    /* function that returns the path or performs side effects */
+    actionProperty: CodeRef<(alert) => string | void>;
```
afreen23, line 14: Why don't you do `history.push` for links and keep just one path resolver ?

frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
```diff
@@ -49,20 +55,57 @@ export const StatusItem: React.FC<StatusItemProps> = ({ Icon, timestamp, message
...
       message={getAlertDescription(alert) || getAlertMessage(alert)}
     >
       {action ? (
-        <Link to={_.isFunction(action.path) ? action.path(alert) : action.path}>{action.text}</Link>
+        path ? (
+          <Link to={path}>{action.text}</Link>
+        ) : (
+          <Button variant={ButtonVariant.link} onClick={resolvedProperty?.actionProperty} isInline>
+            {action.text}
+          </Button>
```
afreen23, line 105:
```suggestion
          <Button variant={ButtonVariant.link} onClick={resolvedProperty?.actionProperty} isInline>
            {action.text}
          </Button>
```

afreen23 (changes requested):

frontend/packages/console-dynamic-plugin-sdk/src/extensions/notification-alert.ts
bipuladh: `no-implicit-globals` would not allow the history object to be directly accessed in the utils folder. We could obviously disable it. @christianvogt wdyt?

afreen23: Can you explain more about the use case here ? I am failing to interpret the above statement wrt `utils` folder.

frontend/packages/ceph-storage-plugin/console-extensions.json
```diff
@@ -1 +1,28 @@
-[]
+[
+  {
+    "type": "console.alert-action",
+    "properties": {
+      "alert": "CephOSDDiskUnavailable",
+      // t('ceph-storage-plugin~Troubleshoot')
```
christianvogt, line 6: Unnecessary for translation. The next line is sufficient.

frontend/packages/ceph-storage-plugin/package.json
```diff
@@ -23,6 +23,9 @@
       "ceph-storage": [
         "integration-tests/**/*.scenario.ts"
       ]
+    },
+    "exposedModules": {
+      "Alert": "./src/utils/alert-action-path.tsx"
```
christianvogt, line 28: Prefer modules to use lowercase names.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/notification-alert.ts
christianvogt: In general links should be rendered as a link and not a button. However I'm not opposed to consistency here where the action itself simply handles navigation or whatever else it wants to do instead. It would simplify the API.

You can import the `history` from `'@console/internal/components/utils';`

frontend/packages/console-shared/src/components/dashboard/status-card/AlertItem.tsx
```diff
@@ -49,20 +55,57 @@ export const StatusItem: React.FC<StatusItemProps> = ({ Icon, timestamp, message
...
 
-const AlertItem: React.FC<AlertItemProps> = ({ alert, hideLink }) => {
+const AlertItem: React.FC<AlertItemProps> = ({ alert }) => {
   const actionsExtensions = useExtensions<AlertAction>(isAlertAction);
   const action = getAlertActions(actionsExtensions).get(alert.rule.name);
   const { t } = useTranslation();
+  const [resolvedProperty, setResolvedProperty] = React.useState<
+    Omit<AlertAction['properties'], 'alert'> & {
+      actionProperty: string | (() => any);
+    }
```
christianvogt, line 65: This would be greatly simplified if you were to use `history.push`. You could remove the state and the effect and simply resolve the coderef and execute it `onClick`

Furthermore, why aren't these extensions already resolved using `useResolvedExtensions`?

bipuladh: `useResolvedExtensions` resolves all the extensions but we really don't need to resolve all the coderef. Just the one's for which the alert's are active. Hence I am doing it manually.

christianvogt: What about doing something like this instead by filtering only on the one action you are interested in?

```
const actionsExtensions = useResolvedExtensions(
    React.useCallback(
      (e) =>
        isAlertAction(e) && e.properties.alert === alert.rule.name,
      [alert.rule.name],
    ),
  );
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/notification-alert.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type AlertAction = ExtensionDeclaration<
+  'console.alert-action',
+  {
+    /* Alert name as defined by `alert.rule.name` property */
+    alert: string;
+    /* Action text */
+    text: string;
+    /* function that returns the path or performs side effects */
+    actionProperty: CodeRef<(alert) => () => void>;
```
christianvogt, line 12: The action should immediately execute and not be a factory function. If the implementation wants to wrap the action in a factory function, then it can do so instead of forcing the API to be this way.
```suggestion
    action: CodeRef<(alert) => void>;
```

frontend/packages/patternfly/src/components/notification-drawer/notification-entry.tsx
```diff
@@ -133,8 +155,11 @@ type NotificationIconTypes = {
 
 type NotificationActionProps = {
   onClick: (event: React.MouseEvent<HTMLElement>) => void;
-  text: string;
-  path: string;
+  text?: string;
+  path?: string;
+  alertAction?: Pick<ResolvedExtension<AlertAction>['properties'], 'text'> & {
+    actionProperty: () => void;
+  };
```
christianvogt, line 162: It's more logical to let the parent component combine the alert action with `onClick`.

christianvogt:
/approve

lgtm but I haven't run it

cloudbehl:
Tested the PR. Alert actions are working.

/lgtm

PR #8571, opened 2021-04-06
Bug 1942086: Fix Home i18n misses

rebeccaalpert:
The Alerting filter toolbar and Alerting Rules AR Watchdog details page heading appear to have been missed by the first pass of i18n work. I internationalized them and some other misses I noticed.

Screen Shot 2021-04-06 at 12 37 29 PM

Screen Shot 2021-04-06 at 12 40 37 PM

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1942086.


zherman0 (approved):

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -212,14 +232,30 @@ const SeverityIcon: React.FC<{ severity: string }> = ({ severity }) => {
   return <Icon />;
 };
 
-export const Severity: React.FC<{ severity: string }> = ({ severity }) =>
-  _.isNil(severity) ? (
+export const Severity: React.FC<{ severity: string }> = ({ severity }) => {
+  const { t } = useTranslation();
+
+  const getSeverityKey = (severityData) => {
```
spadgett, line 238: Missing type. Is this really `severityData` or simply a `severity`? :)

spadgett (commented):
/assign @kyoto 

It would be good for @kyoto to take a look since he is most familiar with this code.

frontend/public/components/monitoring/alerting.tsx
rebeccaalpert: `severity` is already defined, so I had to rename it. Open to suggestions.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -212,14 +232,30 @@ const SeverityIcon: React.FC<{ severity: string }> = ({ severity }) => {
...
+    switch (severityData) {
+      case AlertSeverity.Critical:
+        return t('public~Critical');
+      case AlertSeverity.Info:
+        return t('public~Info');
+      case AlertSeverity.Warning:
+        return t('public~Warning');
+      default:
+        return t('public~None');
```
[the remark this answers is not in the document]
rebeccaalpert: Ah, good to know! Thanks for the heads up.

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
```diff
@@ -67,6 +77,33 @@ export const MonitoringAlerts: React.FC<props> = ({ match, rules, filters, listS
...
+      { id: AlertSeverity.Warning, title: t('public~Warning') },
+      { id: AlertSeverity.Info, title: t('public~Info') },
+      { id: AlertSeverity.None, title: t('public~None') },
+    ],
+    reducer: ({ labels }: Alert | Rule) => labels?.severity,
+    type: 'alert-severity',
+  };
+
+  const alertFilters: RowFilter[] = [alertStateFilter, severityRowFilter];
+
```
kyoto, line 106: It looks like these filters are duplicates of those defined in `alerting.tsx`. I'd prefer to just define them once and leave this importing them. I guess that will mean using `i18next.t` instead of `useTranslation` when defining them.

frontend/public/components/monitoring/alerting.tsx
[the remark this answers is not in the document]
kyoto: 👍 That's right, `severity` could actually be any string

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -162,24 +162,54 @@ const alertStateIcons = {
...
+      case AlertStates.Firing:
+        return t('public~Firing');
+      case AlertStates.Pending:
+        return t('public~Pending');
+      case AlertStates.Silenced:
+        return t('public~Silenced');
+      default:
+        return t('public~Not Firing');
+    }
+  };
```
kyoto, line 180: I think it would make sense to move `getStateKey` outside of the component (and use `i18next.t`) since it will be used in at least other place.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -599,10 +624,81 @@ export const AlertsDetailsPage = withFallback(
...
+        case 'Platform':
+          return t('public~Platform');
+        case 'User':
+          return t('public~User');
+        default:
+          return source;
+      }
+    };
+
+    const SilenceTableRow: RowFunction<Silence> = ({ index, key, obj, style }) => {
```
kyoto, line 653: `SilenceTableRow` is another chunk of code that it getting duplicated with this change. I think we can leave it as a separate component that calls `useTranslation` itself.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -843,10 +897,65 @@ export const AlertRulesDetailsPage = withFallback(
       return `${nameLabel}{${_.map(otherLabels, (v, k) => `${k}="${v}"`).join(',')}}`;
     };
 
+    const ActiveAlerts = ({ activeAlerts, ruleID, activeNamespace }) => (
```
kyoto, line 900: I'd prefer to leave `ActiveAlerts` as a separate component. It was originally broken out to keep `AlertRulesDetailsPage` from becoming too large.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -599,10 +624,81 @@ export const AlertsDetailsPage = withFallback(
...
+          />
+        );
+      }
+      if (alertState && alertState.activeAt) {
+        return <StateTimestamp text={t('public~Since')} timestamp={alertData.activeAt} />;
+      }
+      return null;
+    };
+
+    const getSourceKey = (source) => {
```
kyoto, line 642: As with `getStateKey`, I think it would be preferable to move `getSourceKey` outside of the components so it can be shared.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -742,11 +838,11 @@ export const AlertsDetailsPage = withFallback(
           {silencesLoaded && !_.isEmpty(alert?.silencedBy) && (
             <div className="co-m-pane__body">
               <div className="co-m-pane__body-group">
-                <SectionHeading text="Silenced By" />
+                <SectionHeading text={t('public~Silenced by')} />
```
kyoto, line 841: 💯 Thanks

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1338,6 +1361,101 @@ const AlertsPage_: React.FC<Alerts> = ({ data, loaded, loadError }) => {
     },
   ];
 
+  const severityRowFilter: RowFilter = {
```
kyoto, line 1364: Prefer to keep this outside the components to avoid duplication (probably using `i18next.t`).

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1338,6 +1361,101 @@ const AlertsPage_: React.FC<Alerts> = ({ data, loaded, loadError }) => {
...
+      items: [
+        { id: AlertSource.Platform, title: t('public~Platform') },
+        { id: AlertSource.User, title: t('public~User') },
+      ],
+      reducer: alertSource,
+      type: 'alert-source',
+    },
+  ];
+
+  const AlertTableRow: RowFunction<Alert> = ({ index, key, obj, style }) => {
```
kyoto, line 1401: Prefer to keep this as a separate component to keep `AlertsPage` from growing too large.

kyoto:
👍 Thanks @rebeccaalpert! A few comments related to avoiding the code duplication, but otherwise generally looks good to me.

frontend/public/components/monitoring/alerting.tsx
rebeccaalpert: I tried this initially. If you use `useTranslation`, you get errors because you can only use `useTranslation` in a React function component. (This is a RowFunction typed component.) If you try to make it into a function component, the Table component that takes it as a Row prop complains that it's not a RowFunction. Hence the duplicates.

i18next.t seems to work ok, but I've gotten comments from Sam about it not being reactive before.

frontend/public/components/monitoring/alerting.tsx
rebeccaalpert: i18next.t is iffy with row filters. It's causing missingKey errors that will break the CI tests.

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
rebeccaalpert: i18next.t is iffy in row filters. Using it causes missingKey errors that will break the CI tests. Sam's against passing the TFunction around as an argument, so this was the only other option.

rebeccaalpert:
I addressed the PR feedback where it was possible and left comments where it was not given the constraints.

rebeccaalpert:
I need to update the i18n files; I'll do that tomorrow (I'm in the middle of a task and can't switch branches right now).

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
kyoto: I wasn't aware that `i18next.t` is iffy 😢 

Although we are mostly avoiding passing around `t`, it seems like it would be better than duplicating blocks of code? WDYT @spadgett?

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -585,6 +596,78 @@ const alertStateToProps = (state: RootState, { match }): AlertsDetailsPageProps
   };
 };
 
+const SilenceTableRow: RowFunction<Silence> = ({ index, key, obj, style }) => {
```
kyoto, line 599: Could you please return `SilenceTableRow` to its original place in the file? That will make it easier to read the diff / track changes in the git history.

Similarly for `AlertTableRow`, `ActiveAlerts`, `AlertStateDescription`, anything else?

rebeccaalpert:
I reordered components and updated the i18n files.

rebeccaalpert:
Looking to see if there's a way I can get around the type limitations for row filters so we can pass i18next.t keys through a function and avoid the missingKey errors.

frontend/packages/dev-console/src/components/monitoring/alerts/MonitoringAlerts.tsx
rebeccaalpert: Figured it out! I'll open a tech debt story to clean up all the filters passing the TFunction I noticed in the codebase. Thanks for your patience.

rebeccaalpert:
This should be in good shape now; let me know if you have additional feedback.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1351,47 +1431,32 @@ const AlertsPage_: React.FC<Alerts> = ({ data, loaded, loadError }) => {
...
   );
 };
 const AlertsPage = withFallback(connect(alertsToProps)(AlertsPage_));
 
-const ruleHasAlertState = (rule: Rule, state: AlertStates): boolean =>
+export const ruleHasAlertState = (rule: Rule, state: AlertStates): boolean =>
   state === AlertStates.NotFiring ? _.isEmpty(rule.alerts) : _.some(rule.alerts, { state });
 
-const ruleAlertStateFilter = (filter, rule: Rule) =>
+export const ruleAlertStateFilter = (filter, rule: Rule) =>
```
kyoto, line 1443: It looks like `ruleHasAlertState` and `ruleAlertStateFilter` don't need to be exported now.

frontend/public/components/monitoring/alerting.tsx
```diff
@@ -1458,6 +1495,49 @@ const RulesPage_: React.FC<Rules> = ({ data, loaded, loadError }) => {
...
+      items: [
+        { id: AlertSource.Platform, title: t('public~Platform') },
+        { id: AlertSource.User, title: t('public~User') },
+      ],
+      reducer: alertingRuleSource,
+      type: 'alerting-rule-source',
+    },
+  ];
+
+  const RuleTableRow: RowFunction<Rule> = ({ index, key, obj, style }) => (
```
kyoto, line 1513: Let's move `RuleTableRow` back outside this component to keep `RulesPage` smaller. `RuleTableRow` will just need to call `useTranslation` itself.

kyoto:
Thanks @rebeccaalpert! Just a couple of minor comments, but looks good.

It looks like the commits still need to be squashed too.

rebeccaalpert:
All set!

PR #8572, opened 2021-04-06
Bug 1946705: Use search terms as separate words in quick add search

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5662

**Analysis / Root cause**: 
Using multiple words to narrow down search results does not work because the catalog items are compared against the full query string at once.

**Solution Description**: 
Break down the query string into multiple words, and match all the catalog items against every word separately.

Before:

After:

https://user-images.githubusercontent.com/20013884/113751050-2418f200-9729-11eb-8b0f-3b98a202718f.mp4

Unchanged.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/utils/keyword-filter.ts
+++ b/frontend/packages/console-shared/src/utils/keyword-filter.ts
@@ -1,0 +1,23 @@
+import * as _ from 'lodash';
+
+export const keywordFilter = <T>(
+  filterText: string,
+  items: T[],
+  compareFunction: (keyword: string, item: T) => boolean,
+): T[] => {
+  if (!filterText) {
+    return items;
+  }
+  const keywords = _.uniq(filterText.match(/\S+/g)).map((w) => w.toLowerCase());
+
+  // Sort the longest keyword fist
+  keywords.sort(function(a: string, b: string) {
+    return b.length - a.length;
+  });
+
+  return items.filter((item) => {
+    return keywords.every((keyword) => {
+      return compareFunction(keyword, item);
+    });
+  });
+};
--- a/frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
+++ b/frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
@@ -11,7 +11,7 @@ import {
 } from '@console/app/src/components/quick-starts/utils/quick-start-context';
 
 export const quickSearch = (items: CatalogItem[], query: string) => {
-  return items.filter((item) => keywordCompare(query, item));
+  return keywordCompare(query, items);
 };
 
 export const useTransformedQuickStarts = (quickStarts: QuickStart[]): CatalogItem[] => {
--- a/frontend/packages/console-shared/src/utils/index.ts
+++ b/frontend/packages/console-shared/src/utils/index.ts
@@ -20,3 +20,4 @@ export * from './multiselectdropdown';
 export * from './annotations';
 export * from './xterm-addon-fullscreen';
 export * from './yup-validations';
+export * from './keyword-filter';
--- a/frontend/packages/dev-console/src/components/catalog/service/CatalogServiceProvider.tsx
+++ b/frontend/packages/dev-console/src/components/catalog/service/CatalogServiceProvider.tsx
@@ -72,7 +72,7 @@ const CatalogServiceProvider: React.FC<CatalogServiceProviderProps> = ({
 
   const searchCatalog = React.useCallback(
     (query: string) => {
-      return catalogItems.filter((item) => keywordCompare(query, item));
+      return keywordCompare(query, catalogItems);
     },
     [catalogItems],
   );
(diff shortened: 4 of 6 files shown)


frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
```diff
@@ -3,15 +3,37 @@ import { useTranslation } from 'react-i18next';
...
 import { history, removeQueryArgument } from '@console/internal/components/utils';
 import { CatalogItem } from '@console/dynamic-plugin-sdk';
-import { keywordCompare } from '@console/dev-console/src/components/catalog/utils/catalog-utils';
 import { QuickStart } from '@console/app/src/components/quick-starts/utils/quick-start-types';
 import {
   QuickStartContext,
   QuickStartContextValues,
 } from '@console/app/src/components/quick-starts/utils/quick-start-context';
 
+const keywordListCompare = (keywords: string[], item: CatalogItem): boolean => {
```
spadgett, line 12: I suggest we create a helper so we can reuse keyword logic in other places we don't want to use `fuzzy` for matching. We had similar code in 3.x with some optimizations like checking the longest keyword first (most likely not to match) to improve performance when filtering large lists of things. See

https://github.com/openshift/origin-web-common/blob/master/src/services/keyword.js

rohitkrai03, line 12: I think this should also be changed for the main dev catalog. Search experience for overall catalog should remain same.

rohitkrai03 (changes requested):

frontend/packages/topology/src/components/quick-search/utils/quick-search-utils.tsx
rottencandy: Moved the keyword logic to a helper function.

rottencandy: Updated `keywordCompare` function.

rohitkrai03 (approved):
/approve

frontend/packages/console-shared/src/utils/keyword-filter.ts
```diff
@@ -0,0 +1,19 @@
+export const keywordFilter = <T>(
+  queryString: string,
+  items: T[],
+  compareFunc: (keyword: string, item: T) => boolean,
+): T[] => {
+  if (!queryString) {
+    return items;
+  }
+  const keywords = queryString.trim().split(' ');
```
spadgett, line 9:
```suggestion
  const keywords = _.uniq(filterText.match(/\S+/g));

  // Sort the longest keyword first.
  keywords.sort(function(a: string, b: string){
    return b.length - a.length;
  });
```

frontend/packages/console-shared/src/utils/keyword-filter.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  return items.filter((item) => {
+    return keywords.every((keyword) => {
+      if (!keyword) {
```
spadgett, line 13: When would `keyword` be empty?

frontend/packages/console-shared/src/utils/keyword-filter.ts
```diff
@@ -0,0 +1,19 @@
...
+        return false;
+      }
+      return compareFunc(keyword.toLowerCase(), item);
```
spadgett, line 16: Don't you need to lowercase item, too?

It would be better to lowercase the keyword once when creating the array instead of on every comparison.

frontend/packages/console-shared/src/utils/keyword-filter.ts
rottencandy: Thanks, updated.

frontend/packages/console-shared/src/utils/keyword-filter.ts
rottencandy: I had copied this over from the old function. Doesn't apply anymore, so removed.

frontend/packages/console-shared/src/utils/keyword-filter.ts
rottencandy: Since the item could be of any type, I've moved the lowercasing to `compareFunction`.

frontend/packages/console-shared/src/utils/keyword-filter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const keywordFilter = <T>(
+  filterText: string,
+  items: T[],
+  compareFunction: (keyword: string, item: T) => boolean,
+): T[] => {
+  if (!filterText) {
+    return items;
+  }
+  const keywords = _.uniqBy(filterText.match(/\S+/g), (w) => w.toLowerCase());
```
spadgett, line 11: Is your intent to lowercase the keywords here? I don't believe this will do that, it will just use `toLowerCase` for the uniqueness check.

rottencandy: Sorry I assumed it would save the lowercase strings.
Updated to use `map`.

spadgett (approved):
/lgtm
/hold cancel

PR #8573, opened 2021-04-06
add support for workspaces

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-5559

# Screenshots
workspace support
Screenshot from 2021-04-07 19-03-37
Screenshot from 2021-04-06 22-43-33
Screenshot from 2021-04-06 23-23-00

# Tests
Updated optional check for WorkspaceDefinitionList
Updated Section check for WorkspaceResourceLinkList
Updated test Data 

# Browser conformance
Chrome


andrewballantyne:
FWIW, I added the optional workspaces here @abhinandan13jan: https://github.com/openshift/console/pull/8576

frontend/packages/pipelines-plugin/src/components/shared/workspaces/WorkspaceResourceLinkList.tsx
```diff
@@ -29,72 +29,70 @@ const WorkspaceResourceLinkList: React.FC<WorkspaceResourceLinkListProps> = ({
...
+                  name={secret.secretName}
+                  kind={SecretModel.kind}
+                  namespace={namespace}
+                  displayName={displayName}
+                />
+              );
+            }
+            if (!workspace.volumeClaimTemplate) {
+              return <div key={workspace.name}>{workspace.name}</div>;
+            }
```
andrewballantyne, line 89: What scenario is this section for?

abhinandan13jan: This is listing the `emptyDir`s

andrewballantyne: Interesting... not sure I like the if condition or how it renders... personally I think it should be removed if you are not handling the case, but I'll fix it in a follow up PR.

Thanks for the info.

andrewballantyne (approved):

PR #8574, opened 2021-04-06
update label app.kubernetes.io/name and remove unused yamls

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5741

**Analysis / Root cause**:
- Update label app.kubernetes.io/name to use the application name by default rather than the runtime image name 
- Removes unused yamls

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -21,8 +21,8 @@ export const getAppLabels = ({
     app: name,
     'app.kubernetes.io/instance': name,
     'app.kubernetes.io/component': name,
+    'app.kubernetes.io/name': name,
```
rohitkrai03, line 24: With this we have 4 labels with the same value of `name`. But not sure if we can do anything else here. @christianvogt any ideas what we can do to add `component` or `instance` label with correct values here?

invincibleJai: the default value for `component ` and `instance ` can be the same per https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc

```
app.kubernetes.io/name | REC | The name, reflecting component. | mysql
app.kubernetes.io/component | REC | This is the role/type of the component | frontend
app.kubernetes.io/instance | REC | A unique name identifying the application, usually used if different from app.kubernetes.io/name | accounts
```

PR #8576, opened 2021-04-06
Add support for MultiColumnField to support multiline

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5742

**Analysis / Root cause**: 
MultiColumnField assumed everything is a singular line height - this was problematic for the design in Pipeline Builder.

**Solution Description**: 
Supported an optional complex field setting to avoid over reaching into fields and assuming their size / layout.

Pipeline Builder Example...

Single entry:
Screen Shot 2021-04-06 at 3 28 45 PM

Multiple entries:
image

Multiple `MultiColumnField`s:
Screen Shot 2021-04-06 at 3 28 19 PM

No change.

**Test setup:**

Go to Pipeline Builder and add workspaces.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
+++ b/frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import * as _ from 'lodash';
+import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { MinusCircleIcon } from '@patternfly/react-icons';
 import {
@@ -21,6 +23,7 @@ export interface MultiColumnFieldRowProps {
   isReadOnly?: boolean;
   disableDeleteRow?: boolean;
   spans: gridItemSpanValueShape[];
+  complexFields?: boolean[];
 }
 
 const MultiColumnFieldRow: React.FC<MultiColumnFieldRowProps> = ({
@@ -32,14 +35,24 @@ const MultiColumnFieldRow: React.FC<MultiColumnFieldRowProps> = ({
   isReadOnly,
   disableDeleteRow,
   spans,
+  complexFields = [],
 }) => {
   const { t } = useTranslation();
   return (
-    <div className="odc-multi-column-field__row">
+    <div
+      className={cx('odc-multi-column-field__row', {
+        'odc-multi-column-field__row--singleLine': _.isEmpty(complexFields),
+      })}
+    >
       <Grid>
         {React.Children.map(children, (child: React.ReactElement, i) => {
-          const fieldName = `${name}.${rowIndex}.${child.props.name}`;
-          const newProps = { ...child.props, name: fieldName };
+          const fieldName = `${name}.${rowIndex}`;
+          let newProps = child.props;
+          if (complexFields[i]) {
+            newProps = { ...newProps, namePrefix: fieldName };
+          } else {
+            newProps = { ...newProps, name: `${fieldName}.${child.props.name}` };
+          }
           return (
             <GridItem span={spans[i]} key={fieldName}>
               <div className="odc-multi-column-field__col">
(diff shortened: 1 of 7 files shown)


andrewballantyne:
/hold

Some dispute on layout/design... Waiting on @bgliwa01's approval before this can merge. Any reviews can be conducted against the code design, little of the design changes will impact the overall coding result.

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
```diff
@@ -32,14 +34,24 @@ const MultiColumnFieldRow: React.FC<MultiColumnFieldRowProps> = ({
   isReadOnly,
   disableDeleteRow,
   spans,
+  complexFields = [],
```
abhinandan13jan, line 37: Why do we need to have an array and store value for every column? We could simply have this flag at the row level. WDYT?

frontend/packages/pipelines-plugin/src/components/pipelines/detail-page-tabs/OptionalableWorkspace.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <InputField
+        name={`${namePrefix}.name`}
+        type={TextInputTypes.text}
+        placeholder={t('pipelines-plugin~Name')}
+        isReadOnly={isReadOnly}
+      />
+      <div style={{ marginBottom: 'var(--pf-global--spacer--xs)' }} />
+      <CheckboxField
+        name={`${namePrefix}.optional`}
+        label={t('pipelines-plugin~Optional Workspace')}
```
abhinandan13jan, line 25: Optional workspace

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
```diff
@@ -49,6 +49,7 @@
   "Event Listener details": "Event Listener details",
   "Triggers": "Triggers",
   "URL": "URL",
+  "Optional Workspace": "Optional Workspace",
```
abhinandan13jan, line 52: workspace

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
```diff
@@ -32,14 +34,24 @@ const MultiColumnFieldRow: React.FC<MultiColumnFieldRowProps> = ({
...
 }) => {
   const { t } = useTranslation();
   return (
-    <div className="odc-multi-column-field__row">
+    <div
+      className={cx('odc-multi-column-field__row', {
+        'odc-multi-column-field__row--singleLine': !complexFields,
```
abhinandan13jan, line 43: Should we use a better check than `!complexFields`?

rohitkrai03: Agreed. This will always result in `false`. Maybe you need to use `_.isEmpty(complexFields)` here.

rohitkrai03 (changes requested):
Overall changes looks good to me. There's just one comment.

andrewballantyne:
Just to answer your question @abhinandan13jan - you deleted it but to clarify for a learning experience here.

> Why do we need to have an array and store value for every column? We could simply have this flag at the row level. WDYT?

We don't want to store it as a singular variable because you may mix regular Formik Fields with a "complex" one. If it's complex, it doesn't dig into the props looking for `name` to update the Formik name prop. So there may be cases where we want both to happen at the same time.

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
andrewballantyne: Good point, thanks guys! Easy update.

andrewballantyne:
/unhold

Brigid is checking with UX, but there is no need to hold up this PR on that conversation. We will address it via another PR if there are changes to be made.

bgliwa01 (approved):
I will confirm with UX but approving for now!

rohitkrai03 (approved):
/lgtm

PR #8583, opened 2021-04-07
[WIP] Bug 1885359: Cypress get() doesn't find the `Install` button element rendered on a side pop up in install operators page

dtaylor113:
Previously, we had to use `cy.byLegacyTestID('operator-install-btn').click({ force: true });`  -note the `force:true` in order to click on the **Install** button.

When `force:true` was removed Cypress throws: 
```
"This element `<a.pf-c-button.pf-m-primary.co-catalog-page__overlay-action>` [the Install button] is not 
visible because its parent `<div#modal-container.pf-c-backdrop__open>` has CSS property: 
`overflow: hidden` and an effective width and height of: `1000 x 0` pixels."
```
image

Updated `.pf-c-backdrop__open` to `overflow: visible;` which fixed the [Cypress 'covering' issue](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Covering).  

`<div id="modal-container" role="dialog" aria-modal="true" class="">` is the top modal div and is defined in [app.jsx](https://github.com/openshift/console/blob/cf09e61f9e19bd81e5ab0b3f2c643c14ef6ec6cd/frontend/public/components/app.jsx#L184); It has a height of 10px and a width of 1301px (as computed in browser's dev console), but nothing I can see on the screen.  -Wondering if setting `overflow: visible` really changes anything visual? 

`<div class="pf-c-backdrop">` seems to be the first really 'visible' object, it contains the semi-transparanent background behind the modal.  Checked z-indexes and they are 'stacked' correctly, so the Install button is on top.

### DIFF
--- a/frontend/public/style/_overrides.scss
+++ b/frontend/public/style/_overrides.scss
@@ -512,6 +512,10 @@ button.pf-c-dropdown__menu-item.pf-m-disabled {
   position: absolute !important;
 }
 
+#modal-container.pf-c-backdrop__open {
+  overflow: visible;
+}
+
 .text-muted {
   color: var(--pf-global--Color--200);
 }
--- a/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
+++ b/frontend/packages/operator-lifecycle-manager/integration-tests-cypress/tests/create-namespace.spec.ts
@@ -27,7 +27,7 @@ describe('Create namespace from install operators', () => {
     cy.log('test namespace creation from dropdown');
     cy.visit(`/operatorhub/ns/${testName}`);
     cy.byTestID(operatorSelector).click();
-    cy.byLegacyTestID('operator-install-btn').click({ force: true });
+    cy.byLegacyTestID('operator-install-btn').click();
 
     // configure operator install
     cy.byTestID('Select a Namespace-radio-input').check();


frontend/public/style/_overrides.scss
```diff
@@ -512,6 +512,10 @@ button.pf-c-dropdown__menu-item.pf-m-disabled {
   position: absolute !important;
 }
 
+#modal-container.pf-c-backdrop__open {
+  overflow: visible;
```
rhamilto, line 516: Probably worth reaching out to the PatternFly team to:
1. determine why this declaration exists
1. let them know it is creating this issue

spadgett:
I wonder if this is a Cypress bug if the z-index is correct. I am OK leaving `force: true` as long as we understand the problem and it isn't because a component is unmounting/remounting. It would be good to understand from the Patternfly team why the purpose of this property before we override.

cc @jcaianirh

spadgett:
Thanks @dtaylor113, really appreciate your tracking this down.

dtaylor113:
Closing this PR due to @spadgett  comment: _"I am OK leaving force: true as long as we understand the problem and it isn't because a component is unmounting/remounting."_  This does seem to be a Cypress issue as z-index's are ok, and a user can hover and click on the install button in the GUI outside of Cypress.   

Closing this PR in favor of pursuing other `.click({force: true})` usage which might be needed due to mounting and unmounting of component resulting in DOM detachment.

PR #8585, opened 2021-04-07
Use redhat-developer devfile sample repository

### LINKED ISSUE #8584: cypress devfile add flow tests shouldn't use private devfile repository
'create-from-devfile.feature' is currently referring to the private devfile repository which needs to be replaced with 'https://github.com/redhat-developer/devfile-sample'

josephca:
Fixes #8584 - cypress devfile add flow tests shouldn't use private devfile repository


makambalaji:
/lgtm
/approve

Agreed, Better to avoid private url's :+1:

PR #8587, opened 2021-04-07
Adding login template strings for translation

sg00dwin:
Related to: https://github.com/openshift/oauth-server/pull/71


frontend/public/locales/en/public.json
```diff
@@ -728,5 +728,11 @@
...
   "Disabled": "Disabled",
-  "Enabled": "Enabled"
-}
\ No newline at end of file
+  "Enabled": "Enabled",
+  "LogInToYourAccount": "Log in to your account",
+ 	"LogIn": "Log in",
+ 	"LoginTitle": "Login",
+ 	"WelcomeTo": "Welcome to {{platformTitle}}",
+ 	"LogInWith": "Log in with {{providerName}}",
```
sg00dwin, line 736: @rebeccaalpert I included `{{platformTitle}}` and `{{providerName}}` to give the translators context. Is this acceptable?

rebeccaalpert: Works for me! Thanks.

rebeccaalpert: It's better if the phrase is the key on the left. The translators only receive the key value and they're not going to be able to translate LoginTitle, etc.

sg00dwin: updated the keys
```
"Log in to your account": "Log in to your account",
"Log in": "Log in",
"Login": "Login",
"Welcome to {{platformTitle}}": "Welcome to {{platformTitle}}",
"Log in with {{providerName}}": "Log in with {{providerName}}",
"Error": "Error"
```

rebeccaalpert: Thank you!

rebeccaalpert (approved):
Looks good to me. It just looks like you need to run `yarn i18n`. I think the script allows for manual entries. If that turns out to be incorrect and these are overwritten, you may need to add some parser comments somewhere instead like `// t('public~Log in')`.

frontend/public/locales/en/public.json
```diff
@@ -728,5 +728,11 @@
...
   "Disabled": "Disabled",
-  "Enabled": "Enabled"
-}
\ No newline at end of file
+  "Enabled": "Enabled",
+  "Log in to your account": "Log in to your account",
+  "Log in": "Log in",
+  "Login": "Login",
+  "Welcome to {{platformTitle}}": "Welcome to {{platformTitle}}",
+  "Log in with {{providerName}}": "Log in with {{providerName}}",
```
spadgett, line 736: This won't necessarily work for the OAuth server login page. See https://github.com/openshift/oauth-server/pull/71#discussion_r609880583

spadgett (changes requested):
@sg00dwin Don't manually add these messages to the public.json file. IIRC `yarn i18n` will remove them when run. We need to add i18next-parser comments to one our of source files like

```
// t('Log in')
// ...etc...
```

Then `yarn i18n` will automatically add them to public.json.

frontend/public/locales/en/public.json
sg00dwin: > @sg00dwin Don't manually add these messages to the public.json file. IIRC `yarn i18n` will remove them when run. We need to add i18next-parser comments to one our of source files like

@spadgett is there a recommended file to add these parser comments? Or should I just pick one?

spadgett (approved):
/lgtm

PR #8597, opened 2021-04-09
Add support to override and hide CSV Actions

bipuladh:
- Add Edit Block Pool (static)extension in Ceph Plugin

### DIFF
--- a/frontend/packages/console-plugin-sdk/src/typings/clusterserviceversions.ts
+++ b/frontend/packages/console-plugin-sdk/src/typings/clusterserviceversions.ts
@@ -3,6 +3,8 @@ import { Extension } from './base';
 
 namespace ExtensionProperties {
   export interface ClusterServiceVersionAction {
+    /** Unique id for this Action */
+    id: string;
     /** the kind this action is for */
     kind: K8sResourceKindReference;
     /** label of action */
@@ -11,6 +13,8 @@ namespace ExtensionProperties {
     apiGroup: string;
     /** action callback */
     callback: (kind: K8sResourceKindReference, obj: any) => () => any;
+    /** Hide the action */
+    hidden?: ((kind: K8sResourceKindReference, obj: any) => boolean) | boolean;
   }
 }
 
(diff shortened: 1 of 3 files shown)


frontend/packages/console-plugin-sdk/src/typings/clusterserviceversions.ts
```diff
@@ -11,6 +13,8 @@ namespace ExtensionProperties {
     apiGroup: string;
     /** action callback */
     callback: (kind: K8sResourceKindReference, obj: any) => () => any;
+    /** Hide the action */
+    hidden?: ((kind: K8sResourceKindReference, obj: any) => boolean) | boolean;
```
rawagner, line 17: Whats the usecase for having only boolean ? Why would I contribute an action that is always hidden ? 
But that can be addressed later

christianvogt: I have the same question. It doesn't make sense to me to have `| boolean`

bipuladh: We have use cases where we don't care about the object.
For ex: In External Mode OCS Storage Cluster. You cannot delete any `Block Pool`. For such cases we could just add `false`.

christianvogt: @bipuladh please explain further. You have defined an action which is always `hidden`. How does this make sense?

bipuladh: So Pools created in RHCS Cluster( to which our external mode cluster is connected to ) can have other Clusters connected to it as well. So we should not allow deletion of Block Pool if our Cluster is an external cluster. We have a flag named `OCS_EXTERNAL_MODE` which gets activated when we are running in external mode. We would override delete with the above action and just keep the boolean value as false. So none of the Block Pools can be deleted from the UI.

christianvogt: @bipuladh I don't see where you would do the override. All extensions are loaded at once without any provision for overriding. Having a static action which is always `hidden = false` is essentially a useless contribution it seems.

PR #8599, opened 2021-04-09
Make multus behaviour common between two modes

bipuladh:
Changes to make it behave Internal Mode similar to Attached Mode.


afreen23 (approved):
/lgtm

PR #8550, opened 2021-04-04
Pool management - delete

GowthamShanmugam:
Signed-off-by: Gowtham Shanmugasundaram <gshanmug@redhat.com>

PR #8600, opened 2021-04-09
Added plugin for delete block pool modal

GowthamShanmugam:
This PR is on top of https://github.com/openshift/console/pull/8597
Also, depending on:  https://github.com/openshift/console/pull/8550

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/plugin.ts
+++ b/frontend/packages/ceph-storage-plugin/src/plugin.ts
@@ -796,6 +796,26 @@ const plugin: Plugin<ConsumedExtensions> = [
         ).then((m) => m.default),
     },
   },
+  {
+    type: 'ClusterServiceVersion/Action',
+    properties: {
+      id: 'delete',
+      kind: models.CephBlockPoolModel.kind,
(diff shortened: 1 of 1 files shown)


frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -796,6 +796,26 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+      callback: (kind, obj) => () => {
+        const props = { kind, blockPoolConfig: obj };
+        import(
+          './components/modals/block-pool-modal/delete-block-pool-modal' /* webpackChunkName: "ceph-storage-delete-block-pool-modal" */
+        )
+          .then((m) => m.deleteBlockPoolModal(props))
+          .catch((e) => {
+            // eslint-disable-next-line no-console
+            console.error('Error loading block Pool Modal', e);
+          });
```
cloudbehl, line 815: can this be part of the component itself?

GowthamShanmugam: This error is for even it can't find or any challenge in loading component

PR #8602, opened 2021-04-09
Bug 1848151: Don't poll ClusterVersion when user doesn't have authority

spadgett:
Update the notification drawer to use the `useClusterVersion` hook,
which checks the `CLUSTER_VERSION` flag before trying to watch the
`ClusterVersion` resource. The `CLUSTER_VERSION` flag will resolve to
false if the user doesn't have permission to get the `ClusterVesrion`
resource.

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #8605, opened 2021-04-09
(fix): Remove `i18nKey` for translation

afreen23:
- using `i18nKey` on `Trans` component was not detected by parser.
- as a workaround using Pf component and removing `<br>` tag to observe the same UI behavior
- Not using `<p>` tags, since they are set up as replacement tags , and causing missing key issues


afreen23:
cc @rebeccaalpert

rebeccaalpert (approved):
Looks good to me! Thank you.

zherman0 (approved):

PR #8615, opened 2021-04-12
Bug 1922065: Rootdisk size is default to 15GiB in customize wizard

glekner:


frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
```diff
@@ -225,6 +225,25 @@ export const CreateVM: React.FC<RouteComponentProps> = ({ location }) => {
     }
   }, [loaded, initData, templates, userTemplates, selectedTemplate, t]);
 
+  React.useMemo(() => {
+    const temp = selectedTemplate?.variants?.[0];
+    const requestedPvcSize =
+      temp?.objects?.[0]?.spec?.dataVolumeTemplates[0]?.spec?.pvc?.resources?.requests?.storage;
```
glekner, line 231: @rawagner do we have selectors for this data? couldn't find.

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
```diff
@@ -225,6 +225,25 @@ export const CreateVM: React.FC<RouteComponentProps> = ({ location }) => {
...
+
+    if (requestedPvcSize) {
+      const [value] = requestedPvcSize.match(/\d+/g);
+      const [unit] = requestedPvcSize.match(/[a-zA-Z]+/g);
```
glekner, line 235: @rawagner also here, do we have a better option?

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
yaacov: do we have a way to get the "rootdisk" and not just "[0]"
@metalice do you have code to find the "rootdisk" in case the first one is some other disk ?

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
yaacov: maybe `stringValueUnitSplit` can help here ?

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
rawagner: best way is to use VMWrapper.getBootDevice() and check if it is a 'disk' type.

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
rawagner: `stringValueUnitSplit` or use `getSize()` from `DataVolumeWrapper`

frontend/packages/kubevirt-plugin/src/components/create-vm/create-vm.tsx
rawagner: To get the VM from `selectedTemplate`, use `selectVM` selector

PR #8617, opened 2021-04-12
Bug 1926289: create app-content div with modal-container as sibling

sahil143:
fixes: https://issues.redhat.com/browse/OCPBUGSM-24598

Issue:
1. Catalog Sidebar overlaps the Quick Starts
2. If `#modal-container` is moved as a children of `#app` it's hidden from screen readers because `#app` is markd as aria-hidden true.

Solution:
```
<QuickStartDrawer>
  <div id="app-content"> // mark aria-hidden=true when modal is open
  ....
  </div>
  <div id="modal-container" />
</QuickStartDrawer>
```
create a wrapper `div` and move everything as children of that div except QuickStartDrawer, #modal-container. 
one the modal is open mark `#app-content` as aria-hidden true instead of `#app` so that modal and QuickStartDrawer both are visible to screen readers

overlap-modal

### DIFF
--- a/frontend/packages/integration-tests-cypress/support/project.ts
+++ b/frontend/packages/integration-tests-cypress/support/project.ts
@@ -22,7 +22,7 @@ Cypress.Commands.add('createProject', (name: string, devConsole: boolean = false
   listPage.clickCreateYAMLbutton();
   modal.shouldBeOpened();
   cy.byTestID('input-name').type(name);
-  cy.testA11y('Create Project modal', '#modal-container');
+  cy.testA11y('Create Project modal');
   modal.submit();
   modal.shouldBeClosed();
   // TODO, switch to 'listPage.titleShouldHaveText(name)', when we switch to new test id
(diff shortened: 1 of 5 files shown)


christianvogt:
/approve

lgtm

@spadgett please have a look

dtaylor113:
In console Cypress AXE testing we limit testing just the modal itself, not from root of the DOM via [cy.testA11y('Create Namespace modal', '#modal-container');](https://github.com/openshift/console/blob/bba265c49d23b7587de1e31022959c147fa6e472/frontend/packages/integration-tests-cypress/tests/crud/namespace-crud.spec.ts#L39).  We should take out the 2nd parameter to test if the entire modal passes AXE from root, in other words that the rest of the Application behind the modal passes AXE when modal is open.

dtaylor113:
Hi @sahil143, thanks for updating that Cypress test, however that was just an example.  Here is the complete list:
image

I downloaded this PR and removed `#modal-container` from the remaining tests ^^ .  `namespace-crud.spec.ts` and `monitoring.spec.ts` both passed, multiple times, locally.   

Please add these changes to this PR, or if easier, I can create a PR against this PR with the changes.
Thanks,
- Dave

PR #8623, opened 2021-04-12
Bug 1946922: Fix ingress details page to show referenced secret name

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1946922

Before:
Screen Shot 2021-04-12 at 12 43 29 PM

After:
Screen Shot 2021-04-12 at 12 45 30 PM

### DIFF
--- a/frontend/public/components/ingress.jsx
+++ b/frontend/public/components/ingress.jsx
@@ -37,7 +37,7 @@ const getHosts = (ingress) => {
   return <div className="text-muted">No hosts</div>;
 };
 
-const TLSCert = (ingress) => {
+const TLSCert = ({ ingress }) => {
   const { t } = useTranslation();
   if (!_.has(ingress.spec, 'tls')) {
     return (


spadgett (approved):
/lgtm

yapei (commented):
I think we should also return `ResourceLink` component instead of just `ResourceIcon`
```
  return (
    <div>
      <ResourceIcon kind="Secret" />
      <span>{certs.join(', ')}</span>
    </div>
  );
 ```

**Comment by yapei** [comment]
Also I have a question that `spec.tls` is a list of hosts & secret mapping,  how are we going to display them if there are more than one tls secret? such as 

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: tls-example-ingress-twotls
spec:
  tls:
  - hosts:
      - https-example.foo.com
    secretName: testsecret-tls
  - hosts:
      - http-anotherhost.com
    secretName: testsecret-another-tls
  rules:
  - host: https-example.foo.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: service1
            port:
              number: 80
  - host: http-anotherhost.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: service1
            port:
              number: 80  
```

jcaianirh:
@yapei the secret isn't present in the list of secrets, so I think thats why we've been passing back just the text of the name, instead of the resourcelink. However, I think you're right on the comma separated list of resources, so I'm creating a follow on pr for this.

spadgett:
@yapei Yeah, I debated making this comment as well. The original intent of the code was to show it as plain text, but a bug in how we were passing the props prevented that from working. Switching to a link is a bigger change since we need to handle RBAC checks for users who can't see secrets and change the display when there is more than one. I think any other improvements show be a follow up usability bug or story.

yapei:
Thank you @jcaianirh  @spadgett  for your explanation, that is very helpful

PR #8595, opened 2021-04-09
Support FormSelect for Pipeline Resources

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-3741

**Analysis / Root cause**: 
Poor UX interacting with the dropdown for Pipeline Resources

**Solution Description**: 
Make use of the FormSelect component from PF that supports this.

**Screen shots / Gifs for design review**: 
@openshift/team-devconsole-ux 
@bgliwa01 
@serenamarie125 (requested the change)

FormSelectPipelineDropdowns

**Unit test coverage report**: 
No change

**Test setup:**

Create a Resource Pipeline / View Pipeline Builder

[Sample Resource Pipeline](https://raw.githubusercontent.com/andrewballantyne/pipeline-scratch-pad/master/examples/resources/resources.yaml)

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #8624, opened 2021-04-12
Workspaces and Resource dropdowns are FormSelect

andrewballantyne:
https://issues.redhat.com/browse/ODC-3741

Continuation of https://github.com/openshift/console/pull/8595 as requested by @bgliwa01 

/cc @bgliwa01 @karthikjeeyar 
cc @serenamarie125 

pb-sidebar-form-select

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/field-types.ts
+++ b/frontend/packages/console-shared/src/components/formik-fields/field-types.ts
@@ -63,7 +63,7 @@ export interface DropdownFieldProps extends FieldProps {
   onChange?: (value: string) => void;
 }
 
-export type FormSelectFieldOptions<T = any> = {
+export type FormSelectFieldOption<T = any> = {
   label: string;
   value: T;
   isPlaceholder?: boolean;
@@ -71,8 +71,8 @@ export type FormSelectFieldOptions<T = any> = {
 };
 
 export type FormSelectFieldProps = FieldProps & {
-  disabled?: boolean;
-  options: FormSelectFieldOptions[];
+  isDisabled?: boolean;
+  options: FormSelectFieldOption[];
   onChange?: (selectedValue: any) => void;
 };
 
--- a/frontend/packages/console-shared/src/components/formik-fields/FormSelectField.tsx
+++ b/frontend/packages/console-shared/src/components/formik-fields/FormSelectField.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { useField, useFormikContext, FormikValues } from 'formik';
 import { FormGroup, FormSelect, FormSelectOption } from '@patternfly/react-core';
 import { useFormikValidationFix } from '../../hooks';
-import { FormSelectFieldProps } from './field-types';
+import { FormSelectFieldOption, FormSelectFieldProps } from './field-types';
 import { getFieldId } from './field-utils';
 
 const FormSelectField: React.FC<FormSelectFieldProps> = ({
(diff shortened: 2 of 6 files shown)


andrewballantyne:
cc @makambalaji This covers your dropdown problem in the sidebar fwiw.

bgliwa01 (approved):
lgtm! Thank you Andrew!

debsmita1:
Looks like you have missed using FormSelect for the workspaces dropdown in the pipeline start modal

Screenshot from 2021-04-15 12-40-50

andrewballantyne:
> Looks like you have missed using FormSelect for the workspaces dropdown in the pipeline start modal
> 
> Screenshot from 2021-04-15 12-40-50

@bgliwa01 Thoughts? We don't want to convert all dropdowns in the Console to be this state... and for ConfigMap/Secret/PVC the next dropdown will need a search field so it cannot be converted. Where is the line we draw? My thought was it more-so for Pipeline Builder more than anything else. Do we want to expand the use-cases?

cc @serenamarie125

bgliwa01:
> > Looks like you have missed using FormSelect for the workspaces dropdown in the pipeline start modal
> > Screenshot from 2021-04-15 12-40-50
> 
> @bgliwa01 Thoughts? We don't want to convert all dropdowns in the Console to be this state... and for ConfigMap/Secret/PVC the next dropdown will need a search field so it cannot be converted. Where is the line we draw? My thought was it more-so for Pipeline Builder more than anything else. Do we want to expand the use-cases?
> 
> cc @serenamarie125

I think for now keep it just the Pipeline builder.

debsmita1:
@andrewballantyne  Tried creating a pipeline using one of the samples, the placeholder is missing if I remove the input resource
pb

andrewballantyne:
Good catch, I'll look into that.

andrewballantyne:
> @andrewballantyne Tried creating a pipeline using one of the samples, the placeholder is missing if I remove the input resource
> pb

PF Bug: https://github.com/patternfly/patternfly-react/issues/5687

Put in a workaround for the time being. Thanks for catching that.

frontend/packages/console-shared/src/components/formik-fields/field-types.ts
```diff
@@ -63,7 +63,7 @@ export interface DropdownFieldProps extends FieldProps {
   onChange?: (value: string) => void;
 }
 
-export type FormSelectFieldOptions<T = any> = {
+export type FormSelectFieldOption<T = any> = {
```
andrewballantyne, line 66: Oops... Option types are singular not plural 🤦

andrewballantyne:
fixed a mistake I made where dropdowns didn't get disabled 🤦

frontend/packages/console-shared/src/components/formik-fields/field-types.ts
```diff
@@ -63,16 +63,16 @@ export interface DropdownFieldProps extends FieldProps {
...
   label: string;
   value: T;
   isPlaceholder?: boolean;
   isDisabled?: boolean;
 };
 
 export type FormSelectFieldProps = FieldProps & {
-  disabled?: boolean;
-  options: FormSelectFieldOptions[];
+  isDisabled?: boolean;
```
andrewballantyne, line 74: This is the disable dropdown fix.

debsmita1:
It shows an empty workspace dropdown when there are no workspaces
Screenshot from 2021-04-19 12-38-56

The input resources dropdown works fine
Screenshot from 2021-04-19 13-00-31

andrewballantyne:
Doh, that's unrelated to this PR... but a valid issue... We should trim empty items 😓

andrewballantyne:
@debsmita1 check the diff on the 2nd force push for the fix to empty workspaces showing up in the dropdown.

debsmita1:
Thanks @andrewballantyne !
verified the changes, and it works fine now
/lgtm

PR #8625, opened 2021-04-12
Bug 1924788: Not clear error message when there are no NADs available for the user

avivtur:
Not clear error message when there are no NADs available for the user

In this example, I add a pop-up for error message about unavailable NADs,
which gives the full error message to user when there is an unclear error message.

before:

bz_1924788_before_1_test_user

after:

bz_1924788_after_2_test_user_after_text_with_color
bz_1924788_after_1_test_user_after_text_with_color

 

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1924788


avivtur:
@yaacov @matthewcarleton @metalice please review

matthewcarleton:
thanks @avivtur - I wonder if we should use "more info" and use the expandable component rather than the Pop over. Help popovers are typically reserved for static text.

avivtur:
Hi @matthewcarleton thanks for your reply,
I have made also an example with expandable section:
bz_1924788_after_1_expandable
bz_1924788_after_2_expandable

What do you think about it?

yaacov:
/retitle Bug 1924788: Not clear error message when there are no NADs available for the user

matthewcarleton:
LGTM

yaacov:
@avivtur hay, the screenshots are not from this PR, cancaling lgtm

avivtur:
@yaacov Yes you right I made a local example, If the second option is preffered I'll commit the changes now

yaacov:
can you change the text from "show more" / "show less" to "Show/Hide raw error message" ?
The gap between the "show more" and the text of the error looks to big to me, can you make the error text closer to the "show less" header ?

matthewcarleton:
> can you change the text from "show more" / "show less" to "Show/Hide raw error message" ?
> The gap between the "show more" and the text of the error looks to big to me, can you make the error text closer to the "show less" header ?

@yaacov that is how the PF component does it - do we want to customize it?
If we want to write this to be more specific could we say show/hide error message? I don't think we need the word raw.

yaacov:
> > can you change the text from "show more" / "show less" to "Show/Hide raw error message" ?
> > The gap between the "show more" and the text of the error looks to big to me, can you make the error text closer to the "show less" header ?
> 
> @yaacov that is how the PF component does it - do we want to customize it?
> If we want to write this to be more specific could we say show/hide error message? I don't think we need the word raw.

ahh, didn't know that, if it's come from patternfly than I'm ok with it, 
ca we add a title/note above/before the raw error text that it is a raw error text and not just regular "more info" ?

avivtur:
Hi all
@matthewcarleton @yaacov this is the example of the new code I've pushed:
Screenshot from 2021-04-14 15-13-32
Screenshot from 2021-04-14 15-13-29

Please let me now which is preffered

avivtur:
Hi all,
These are creenshots of my last commit
@matthewcarleton @yaacov please review

Screenshot from 2021-04-18 14-00-40
Screenshot from 2021-04-18 14-00-38

PR #8628, opened 2021-04-12
Bug 1942673: Fix pipeline model label references not using keys

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1942673

Address the references to the label and labelPlural in pipeline model to use translated lableKey and labelPluralKeys. Includes detail and list pages, and overviews and sections.

cc: @rebeccaalpert

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/resource-overview/TriggerTemplateResourceLink.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/resource-overview/TriggerTemplateResourceLink.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { ResourceLink } from '@console/internal/components/utils';
 import { K8sKind, referenceForModel } from '@console/internal/module/k8s';
 import { ClipboardCopy } from '@patternfly/react-core';
@@ -15,7 +16,8 @@ const TriggerTemplateResourceLink: React.FC<TriggerTemplateResourceLinkProps> = 
   namespace,
   model,
 }) => {
-  const title = model.labelPlural;
+  const { t } = useTranslation();
+  const title = t(model.labelPluralKey);
   const kind = referenceForModel(model);
 
   if (links.length === 0) {
--- a/frontend/packages/pipelines-plugin/src/components/cluster-tasks/ClusterTaskDetails.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/cluster-tasks/ClusterTaskDetails.tsx
@@ -17,7 +17,7 @@ const ClusterTaskDetails: React.FC<ClusterTaskDetailsProps> = ({ obj: task }) =>
     <div className="co-m-pane__body">
       <SectionHeading
         text={t('pipelines-plugin~{{clusterTaskLabel}} details', {
-          clusterTaskLabel: ClusterTaskModel.label,
+          clusterTaskLabel: t(ClusterTaskModel.labelKey),
         })}
       />
       <div className="row">
--- a/frontend/packages/pipelines-plugin/src/components/pipeline-resources/list-page/PipelineResourcesList.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipeline-resources/list-page/PipelineResourcesList.tsx
@@ -10,7 +10,7 @@ const PipelineResourcesList: React.FC = (props) => {
   return (
     <Table
       {...props}
-      aria-label={PipelineResourceModel.labelPlural}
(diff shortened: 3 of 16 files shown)


rebeccaalpert (approved):
These look good to me. It looks like the Pipelines Plugin models' labelKey and labelPluralKey values don't have parser comments (i.e. `// t('pipelines-plugin~ClusterTriggerBinding')`), so they're not being picked up by the parser. We should fix that, whether that's as part of this PR or a separate bug.

andrewballantyne:
The model values also need a space... I'm handling it currently in my PR, but feel free to handle it here too. https://github.com/openshift/console/pull/8630/files?file-filters%5B%5D=.scss&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-7f8eda14258aabd95da2c6c5153fa25ee774766954a3f11b300889a178b928ddR27

rebeccaalpert:
@andrewballantyne - We dehumanized all the resource kinds on the admin console side of things (i.e. Resource Quotas to ResourceQuotas). They have been done that way intentionally.

andrewballantyne:
> @andrewballantyne - We dehumanized all the resource kinds on the admin console side of things (i.e. Resource Quotas to ResourceQuotas). They have been done that way intentionally.

Ah, interesting point. Unfortunately they were not reflected like this in the `pipelines-plugin.json` file and thus it cannot find `PipelineRun` part of the Pipeline package 😞 

However, not sure it makes sense to keep English with a space and then translations without a space. One way needs to be fully adopted. My PR is just correcting an i18n error on the page that I didn't cause.

I'll be happy to log a bug and trace down all references of `Pipeline Run` and replace them with `PipelineRun` in both i18n keys and in text (ie model.label) if that's our official stance. cc @bgliwa01 FYI

rebeccaalpert:
@andrewballantyne - We did all the kinds without a space on the admin side (there may be some mistakes remaining) for English. I think @spadgett was going to bring this up with the plugin folks at a meeting (but I'm not sure which one). Sam, do you have time to weigh in on if resource kind dehumanization is our official stance at this point, or just for admin console?

rebeccaalpert:
@andrewballantyne - I spoke to @spadgett at scrum and resource kind dehumanization (no spaces) is the goal for plugins as well. He said he brought it up in the plugin sync call last week, but dev console wasn't able to attend. It would be great if Pipeline Run, etc. could be brought in line with the others.

andrewballantyne:
> @andrewballantyne - I spoke to @spadgett at scrum and resource kind dehumanization (no spaces) is the goal for plugins as well. He said he brought it up in the plugin sync call last week, but dev console wasn't able to attend. It would be great if Pipeline Run, etc. could be brought in line with the others.

Understood. I'll do a scan and update all of them, thanks for the update. It's a bit odd still to me, but best to be consistent with the larger whole. The changes I make should not likely impact anything in this PR.

andrewballantyne:
For posterity, https://github.com/openshift/console/pull/8640 & https://github.com/openshift/console/pull/8644 are to clean up more relationship with the labelKeys and kind spaces.

andrewballantyne (changes requested):
Ah, had a chance to look into things... the problem is a lack of a mock for the additions of useTranslation.

```
jest.mock('react-i18next', () => {
  const reactI18next = require.requireActual('react-i18next');
  return {
    ...reactI18next,
    useTranslation: () => ({ t: (key) => key }),
  };
});
```

Need to add this to the test files you added the call to.

frontend/packages/pipelines-plugin/src/components/pipelines/resource-overview/TriggerTemplateResourceLink.tsx
```diff
@@ -15,7 +16,8 @@ const TriggerTemplateResourceLink: React.FC<TriggerTemplateResourceLinkProps> =
   namespace,
   model,
 }) => {
-  const title = model.labelPlural;
+  const { t } = useTranslation();
```
andrewballantyne, line 19: Breaks tests in `frontend/packages/pipelines-plugin/src/components/pipelines/resource-overview/__tests__/TriggerTemplateResourceLink.spec.tsx` needs mock.

andrewballantyne (changes requested):
Might only be the one case 🤔

frontend/packages/pipelines-plugin/src/components/pipelines/resource-overview/TriggerTemplateResourceLink.tsx
jcaianirh: thanks @andrewballantyne !

spadgett (approved):
/lgtm

PR #8629, opened 2021-04-12
Migrate helm actions to use new action extensions

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5692

**Changes in the PR**: 
- Updated the action extension types based on the design change to use providers.
- Used new extension declaration.
- Contributed helm actions using new extension mechanism. 
- Created `ActionsLoader` component that fetches and loads all the actions.
- Created `KebabMenu` components that use new `Action` type. 
- Updated the Helm list page to use new kebab actions from extensions.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/extensions/actions.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/extensions/actions.ts
@@ -1,42 +1,41 @@
 import * as React from 'react';
 import { Extension } from '@console/plugin-sdk/src/typings/base';
-import { CodeRef, EncodedCodeRef, UpdateExtensionProperties } from '../types';
+import { ExtensionDeclaration, CodeRef } from '../types';
 import { ExtensionHook } from '../api/common-types';
 import { AccessReviewResourceAttributes } from './console-types';
 
-namespace ExtensionProperties {
-  export type ActionCreator = {
-    /** ID used to identify the action creator. */
-    id: string;
-    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
-    contextId: string | 'resource';
-    /** A react hook which creates and returns action for the given scope.
-     * If contextId = `resource` then the scope will always be a K8s resource object
-     * */
-    creator: EncodedCodeRef;
-  };
-
-  export type ActionProvider = {
+/** ActionProvider contributes a hook that returns list of action ids for specific context */
+export type ActionProvider = ExtensionDeclaration<
+  'console.action/provider',
+  {
     /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
     contextId: string | 'resource';
     /** A react hook which returns action for the given scope.
      * If contextId = `resource` then the scope will always be a K8s resource object
      * */
-    provider: EncodedCodeRef;
-  };
+    provider: CodeRef<ExtensionHook<Action[]>>;
+  }
+>;
 
-  export type ResourceActionProvider = {
+/** ResourceActionProvider contributes a list of action ids for specific resource model */
+export type ResourceActionProvider = ExtensionDeclaration<
+  'console.action/resource-provider',
+  {
     /** The model for which this provider provides actions for. */
     model: {
       group: string;
       version?: string;
       kind?: string;
     };
-    /** The action ids to contribute to */
-    actions: string[];
-  };
+    /** A react hook which returns action for the given resource model */
+    provider: CodeRef<ExtensionHook<Action[]>>;
+  }
+>;
 
-  export type ActionGroup = {
+/** ActionGroup contributes an action group that can also be a submenu */
+export type ActionGroup = ExtensionDeclaration<
+  'console.action/group',
+  {
     /** ID used to identify the action section. */
     id: string;
     /** The label to display in the UI.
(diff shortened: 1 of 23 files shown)


frontend/packages/console-shared/src/components/actions/ActionsHookResolver.tsx
```diff
@@ -0,0 +1,32 @@
...
+  useValue,
+  scope,
+  onValueResolved,
+  onValueError,
+}) => {
+  const [value, loaded, loadError] = useValue(scope);
+
+  React.useEffect(() => {
+    if (loaded) onValueResolved(value, id);
+  }, [id, loaded, onValueResolved, value]);
```
christianvogt, line 23: IMO it makes sense to only execute this callback if loaded and the value changes.
Not sure you need the ID. Seems like that could be something handled by the parent? But I suppose you can keep it if needed.
```suggestion
  React.useEffect(() => {
    if (loaded) onValueResolved(value);
  }, [loaded, value]);
```

frontend/packages/console-shared/src/components/actions/useActions.ts
```diff
@@ -0,0 +1,55 @@
...
+  React.useEffect(() => {
+    if (providerExtensionsResolved && stableProviderExtensions.current.length < 1) {
+      stableProviderExtensions.current = providerExtensions;
+    }
+    // Do not upodate the stable provider ref every time the extensions change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [providerExtensionsResolved]);
+
+  stableProviderExtensions.current.forEach((extension) => {
+    const [result, resolved, error] = extension.properties.provider(scope);
```
christianvogt, line 35: Oh no... `useActions` may not be a viable option because of the lazy resolution of extensions...

frontend/packages/console-shared/src/components/kebab/KebabItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export type KebabItemProps = {
+  option: Action;
+  onClick: (event: React.MouseEvent<{}>, action: Action) => void;
+  autoFocus?: boolean;
+  onEscape?: () => void;
+};
+
+// eslint-disable-next-line no-underscore-dangle
+const KebabItem_: React.FC<KebabItemProps & { isAllowed: boolean }> = ({
```
christianvogt, line 18: Rename to avoid eslint disable
```suggestion
const KebabItemButton: React.FC<KebabItemProps & { isAllowed: boolean }> = ({
```

frontend/packages/console-shared/src/components/kebab/KebabItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const classes = classNames('pf-c-dropdown__menu-item', { 'pf-m-disabled': disabled });
+  return (
+    <button
+      type="button"
+      className={classes}
+      onClick={(e) => !disabled && onClick(e, option)}
+      // eslint-disable-next-line jsx-a11y/no-autofocus
+      autoFocus={autoFocus}
+      onKeyDown={onEscape && handleEscape}
+      data-test-action={t(option.label)}
```
christianvogt, line 41:
```suggestion
      data-test-action={t(option.id)}
```

frontend/packages/console-shared/src/components/kebab/KebabMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+        'dropdown pf-c-dropdown': true,
+        'pf-m-expanded': active,
+      })}
+    >
+      <button
+        ref={dropdownRef}
+        type="button"
+        aria-expanded={active}
+        aria-haspopup="true"
+        aria-label="Actions"
```
christianvogt, line 78: loocalize

frontend/packages/console-shared/src/components/kebab/KebabSubMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // use a callback ref because FocusTrap is old and doesn't support non-function refs
+  const subMenuCbRef = React.useCallback((node) => (subMenuRef.current = node), []);
+
+  return (
+    <>
+      <button
+        ref={nodeRef}
+        type="button"
+        className="oc-kebab__sub pf-c-dropdown__menu-item"
+        data-test-action={option.label}
```
christianvogt, line 30:
```suggestion
        data-test-action={option.id}
```

frontend/packages/helm-plugin/src/actions/creators.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Action } from '@console/dynamic-plugin-sdk';
+
+export const getHelmDeleteAction = ({
+  releaseName,
+  namespace,
+  redirect,
+}: HelmActionsScope): Action => ({
+  id: 'helm-delete-action',
+  // t('helm-plugin~Uninstall Helm Release')
+  label: 'helm-plugin~Uninstall Helm Release',
```
christianvogt, line 13: This should be a translated string.

frontend/packages/console-shared/src/components/actions/ActionsHookResolver.tsx
rohitkrai03: Yeah, the ID is needed to create a map of extension ids to the actions returned by the provider. Also, do we really need to remove `onValueResolved` from the dependencies? It's a callback from the parent which doesn't change after the first render anyway.

frontend/packages/console-shared/src/components/actions/useActions.ts
rohitkrai03: Yeah, I was trying to find a way to make it work. But it doesn't seem like it will work as expected. Will remove it from the commit.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/actions.ts
```diff
@@ -1,42 +1,41 @@
...
-    /** The context ID helps to narrow the scope of contributed actions to a particular area of the application. Ex - topology, helm */
-    contextId: string | 'resource';
-    /** A react hook which creates and returns action for the given scope.
-     * If contextId = `resource` then the scope will always be a K8s resource object
-     * */
-    creator: EncodedCodeRef;
-  };
-
-  export type ActionProvider = {
+/** ActionProvider contributes a hook that returns list of action ids for specific context */
```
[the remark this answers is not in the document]
rohitkrai03: Done.

frontend/packages/console-shared/src/components/kebab/KebabItems.tsx
rohitkrai03: Done.

frontend/packages/console-shared/src/components/kebab/KebabItems.tsx
rohitkrai03: Done

frontend/packages/console-shared/src/components/kebab/KebabMenu.tsx
rohitkrai03: Done.

frontend/packages/console-shared/src/components/kebab/KebabSubMenu.tsx
rohitkrai03: Done.

frontend/packages/helm-plugin/src/actions/creators.ts
rohitkrai03: Done.

frontend/packages/console-shared/src/components/actions/ActionsHookResolver.tsx
christianvogt: If the parent changes `onValueResolved` you don't actually want to re-run the effect. You only care to notify when `value` changes.

The reason I said ignore the `id` prop is because the parent can do this and keep the `id` in the closure.
`onValueResolved={(value) => onValueResolved(value, id)}`

rohitkrai03: Updated. Seems like it will have the same effect though, earlier `onValueResolved` never changed after first render but now it does and we removed it from the dependencies array explicitly.

christianvogt: Yes the outcome is the same :)
But this is a pattern that should be applied in other components in future. You shouldn't rely on a stable reference of the callback to ensure we don't over-notify. When used in an effect in this manner, only the change in the trigger value should result in a notification and not a change in the callback itself.

frontend/packages/console-shared/src/components/actions/ActionsHookResolver.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [value, loaded, loadError] = useValue(scope);
+
+  React.useEffect(() => {
+    if (loaded) onValueResolved(value);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loaded, value]);
+
+  React.useEffect(() => {
+    if (loadError) onValueError(loadError);
+  }, [loadError, onValueError]);
```
christianvogt, line 26: Don't include `onValueError` in the deps.
You could probably combine both effects, no? Should you be resolving if there is an error?

frontend/packages/helm-plugin/src/actions/providers.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const useHelmActionProvider = (scope: HelmActionsScope) => {
+  const { t } = useTranslation();
+  const actions = React.useMemo(
+    () => [
+      getHelmUpgradeAction(scope, t),
+      getHelmRollbackAction(scope, t),
+      getHelmDeleteAction(scope, t),
+    ],
+    [scope, t],
```
christianvogt, line 14: Can you be sure that `scope` will not be mutated by the caller?

christianvogt:
Tested and it works fine.

frontend/packages/helm-plugin/src/actions/providers.ts
rohitkrai03: That would be the assumption here since the provider and creators are all owned by the same plugin contributing these actions.

frontend/packages/console-shared/src/components/actions/ActionsHookResolver.tsx
rohitkrai03: I think keeping it in two separate effects makes sense here. We also don't want to stop the resolution when error occurs because the parent should know what to do in case of an error. The resolver should return a resolved value or `undefined` in any case to complete the loading of all resolvers.

rohitkrai03: Removed `onValueError ` from dependencies.

christianvogt:
/lgtm

No change in UI and epic has no-doc label as there is no impact to the user to refactor the underlying mechanism to implement helm UI actions.

/label px-approved 
/label doc-approved

Tested helm actions on list page to be functioning. Tested actions also from details and topology page.

/label qe-approved

PR #8632, opened 2021-04-13
fixes issue with kn resources not shown if KS is created

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5759

**Analysis / Root cause**: 
In Topology, Kntive resources were not shown but underlying Deployments were shown instead if Kafka Source is created in same ns

image

**Unit Test**: 

image

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


vikram-raj (commented):
Verified it. and it works as expected.
Thanks @invincibleJai 

/lgtm

PR #8638, opened 2021-04-13
Bug 1952697: Add Segment support for analytics

DaoDaoNoCode:
**Fixes:**
https://issues.redhat.com/browse/CONSOLE-2536

**Solution/Description:**
Add telemetry support via dynamic plugins, create a demo extension of using [segment.io](https://segment.com) to instrument the UI.

NOTE: `SEGMENT_KEY` in `frontend/dynamic-demo-plugin/src/utils/telemetry.ts` need to be provided to connect the [segment.io](https://segment.com) console.

### DIFF
--- a/Dockerfile.plugins.demo2
+++ b/Dockerfile.plugins.demo2
@@ -3,6 +3,12 @@
 # See frontend/dynamic-demo-plugin/README.md for details.
 
 # Stage 0: build the demo plugin
+# Manually pre-build the demo plugin:
+# cd frontend
+# yarn install
+# cd dynamic-demo-plugin
+# yarn install
+# yarn build
 
 # Stage 1: build the target image
 FROM node:10
--- a/Dockerfile.plugins.demo2
+++ b/Dockerfile.plugins.demo2
@@ -1,0 +1,22 @@
+# This image is used for testing OpenShift Console dynamic plugin capabilities.
+#
+# See frontend/dynamic-demo-plugin/README.md for details.
+
+# Stage 0: build the demo plugin
+
+# Stage 1: build the target image
+FROM node:10
+
+COPY ./frontend/dynamic-demo-plugin/dist /opt/console-demo-plugin/static
+COPY ./frontend/dynamic-demo-plugin/node_modules /opt/console-demo-plugin/node_modules
+COPY ./frontend/dynamic-demo-plugin/http-server.sh /opt/console-demo-plugin/http-server.sh
+
+LABEL io.k8s.display-name="OpenShift Console Demo Plugin" \
+      io.k8s.description="Sample OpenShift Console dynamic plugin used for testing purposes." \
+      io.openshift.tags="openshift" \
+      maintainer="Vojtech Szocs <vszocs@redhat.com>"
+
+USER node
+
+WORKDIR /opt/console-demo-plugin
+ENTRYPOINT [ "./http-server.sh", "./static" ]
--- a/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
+++ b/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
@@ -1,0 +1,4 @@
+export const eventListener = (eventType: string, properties: {}) => {
+  // eslint-disable-next-line no-console
+  console.log('Demo Plugin received telemetry event: ', eventType, properties);
+};
--- a/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
+++ b/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
@@ -1,4 +1,84 @@
-export const eventListener = (eventType: string, properties: {}) => {
+import { createHash } from "crypto";
+
+const SEGMENT_KEY = '';
+
+const initSegment = () => {
+  const analytics = ((window as any).analytics = (window as any).analytics || []);
+  if (!analytics.initialize)
+    if (analytics.invoked)
+      window.console && console.error && console.error('Segment snippet included twice.');
+    else {
+      analytics.invoked = !0;
+      analytics.methods = [
+        'trackSubmit',
+        'trackClick',
+        'trackLink',
+        'trackForm',
+        'pageview',
+        'identify',
+        'reset',
+        'group',
+        'track',
+        'ready',
+        'alias',
+        'debug',
+        'page',
+        'once',
+        'off',
+        'on',
+        'addSourceMiddleware',
+        'addIntegrationMiddleware',
+        'setAnonymousId',
+        'addDestinationMiddleware',
+      ];
+      analytics.factory = function(e: string) {
+        return function() {
+          var t = Array.prototype.slice.call(arguments);
+          t.unshift(e);
+          analytics.push(t);
+          return analytics;
+        };
+      };
+      for (var e = 0; e < analytics.methods.length; e++) {
+        var key = analytics.methods[e];
+        analytics[key] = analytics.factory(key);
+      }
+      analytics.load = function(key: string, e: Event) {
+        const t = document.createElement('script');
+        t.type = 'text/javascript';
+        t.async = !0;
+        t.src = 'https://cdn.segment.com/analytics.js/v1/' + key + '/analytics.min.js';
+        const n = document.getElementsByTagName('script')[0];
+        if (n.parentNode) {
+          n.parentNode.insertBefore(t, n);
+        }
+        analytics._loadOptions = e;
+      };
+      analytics.SNIPPET_VERSION = '4.13.1';
+      if (SEGMENT_KEY) {
+        analytics.load(SEGMENT_KEY);
+      }
+    }
+  return analytics;
+};
+
+initSegment();
+
+export const eventListener = (eventType: string, properties?: any) => {
   // eslint-disable-next-line no-console
   console.log('Demo Plugin received telemetry event: ', eventType, properties);
+  const analytics = initSegment();
+  switch (eventType) {
+    case 'identify':
+      const { clusterId, id, user } = properties;
+      // Use md5 hash to anonymize the user
+      const anonymousId = createHash('md5').update(`${clusterId}-${id}`).digest('hex');
+      (window as any).analytics.identify(anonymousId, user);
+      break;
+    case 'page':
+      analytics.page(undefined, properties);
+      break;
+    default:
+      analytics.track(eventType, properties);
+  }
 };
--- a/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
+++ b/frontend/dynamic-demo-plugin/src/utils/telemetry.ts
@@ -1,84 +1,90 @@
-import { createHash } from "crypto";
-
 const SEGMENT_KEY = '';
 
 const initSegment = () => {
   const analytics = ((window as any).analytics = (window as any).analytics || []);
-  if (!analytics.initialize)
-    if (analytics.invoked)
-      window.console && console.error && console.error('Segment snippet included twice.');
-    else {
-      analytics.invoked = !0;
-      analytics.methods = [
-        'trackSubmit',
-        'trackClick',
-        'trackLink',
-        'trackForm',
-        'pageview',
-        'identify',
-        'reset',
-        'group',
-        'track',
-        'ready',
-        'alias',
-        'debug',
-        'page',
-        'once',
-        'off',
-        'on',
-        'addSourceMiddleware',
-        'addIntegrationMiddleware',
-        'setAnonymousId',
-        'addDestinationMiddleware',
-      ];
-      analytics.factory = function(e: string) {
-        return function() {
-          var t = Array.prototype.slice.call(arguments);
-          t.unshift(e);
-          analytics.push(t);
-          return analytics;
-        };
+  if (analytics.initialize) {
+    return;
+  }
+  if (analytics.invoked)
+    window.console && console.error && console.error('Segment snippet included twice.');
+  else {
+    analytics.invoked = true;
+    analytics.methods = [
+      'trackSubmit',
+      'trackClick',
+      'trackLink',
+      'trackForm',
+      'pageview',
+      'identify',
+      'reset',
+      'group',
+      'track',
+      'ready',
+      'alias',
+      'debug',
+      'page',
+      'once',
+      'off',
+      'on',
+      'addSourceMiddleware',
+      'addIntegrationMiddleware',
+      'setAnonymousId',
+      'addDestinationMiddleware',
+    ];
+    analytics.factory = function(e: string) {
+      return function() {
+        let t = Array.prototype.slice.call(arguments);
+        t.unshift(e);
+        analytics.push(t);
+        return analytics;
       };
-      for (var e = 0; e < analytics.methods.length; e++) {
-        var key = analytics.methods[e];
-        analytics[key] = analytics.factory(key);
-      }
-      analytics.load = function(key: string, e: Event) {
-        const t = document.createElement('script');
-        t.type = 'text/javascript';
-        t.async = !0;
-        t.src = 'https://cdn.segment.com/analytics.js/v1/' + key + '/analytics.min.js';
-        const n = document.getElementsByTagName('script')[0];
-        if (n.parentNode) {
-          n.parentNode.insertBefore(t, n);
-        }
-        analytics._loadOptions = e;
-      };
-      analytics.SNIPPET_VERSION = '4.13.1';
-      if (SEGMENT_KEY) {
-        analytics.load(SEGMENT_KEY);
-      }
+    };
+    for (let e = 0; e < analytics.methods.length; e++) {
+      let key = analytics.methods[e];
+      analytics[key] = analytics.factory(key);
     }
-  return analytics;
+    analytics.load = function(key: string, e: Event) {
+      const t = document.createElement('script');
+      t.type = 'text/javascript';
+      t.async = true;
+      t.src = 'https://cdn.segment.com/analytics.js/v1/' + encodeURIComponent(key) + '/analytics.min.js';
+      const n = document.getElementsByTagName('script')[0];
+      if (n.parentNode) {
+        n.parentNode.insertBefore(t, n);
+      }
+      analytics._loadOptions = e;
+    };
+    analytics.SNIPPET_VERSION = '4.13.1';
+    if (SEGMENT_KEY) {
+      analytics.load(SEGMENT_KEY);
+    }
+  }
 };
 
 initSegment();
 
-export const eventListener = (eventType: string, properties?: any) => {
+export const eventListener = async (eventType: string, properties?: any) => {
   // eslint-disable-next-line no-console
   console.log('Demo Plugin received telemetry event: ', eventType, properties);
-  const analytics = initSegment();
+  const anonymousIP = {
+    context: {
+      ip: '0.0.0.0',
+    },
+  };
   switch (eventType) {
     case 'identify':
-      const { clusterId, id, user } = properties;
-      // Use md5 hash to anonymize the user
-      const anonymousId = createHash('md5').update(`${clusterId}-${id}`).digest('hex');
-      (window as any).analytics.identify(anonymousId, user);
+      const { user, ...otherProperties } = properties;
+      const id = user.metadata.uid || `${location.host}-${user.metadata.name}`;
+      // Use SHA1 hash algorithm to anonymize the user
+      const anonymousIdBuffer = await crypto.subtle.digest('SHA-1', new TextEncoder().encode(id));
+      const anonymousIdArray = Array.from(new Uint8Array(anonymousIdBuffer));
+      const anonymousId = anonymousIdArray.map((b) => b.toString(16).padStart(2, '0')).join('');
+      (window as any).analytics.identify(anonymousId, otherProperties, anonymousIP);
       break;
     case 'page':
-      analytics.page(undefined, properties);
+      (window as any).analytics.page(undefined, properties, anonymousIP);
       break;
     default:
-      analytics.track(eventType, properties);
+      (window as any).analytics.track(eventType, properties, anonymousIP);
   }
 };
--- a/frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
+++ b/frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
@@ -7,6 +7,7 @@ import {
   QuickStartStatus,
   QuickStartTaskStatus,
 } from './quick-start-types';
+import { useTelemetry } from '@console/shared/src/hooks/useTelemetry';
 
 export type QuickStartContextValues = {
   activeQuickStartID: string;
@@ -56,6 +57,7 @@ const useAllQuickStartStates = () =>
 export const useValuesForQuickStartContext = (): QuickStartContextValues => {
   const [activeQuickStartID, setActiveQuickStartID] = useActiveQuickStartId();
   const [allQuickStartStates, setAllQuickStartStates] = useAllQuickStartStates();
+  const fireTelemetryEvent = useTelemetry();
 
   const setActiveQuickStart = useCallback(
     (quickStartId: string, totalTasks?: number) => {
@@ -89,8 +91,12 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
           [quickStartId]: getDefaultQuickStartState(totalTasks, QuickStartStatus.IN_PROGRESS),
         };
       });
+      fireTelemetryEvent('Quick Start Initiated', {
+        id: quickStartId,
+        type: 'start',
+      });
     },
-    [setActiveQuickStartID, setAllQuickStartStates],
+    [setActiveQuickStartID, setAllQuickStartStates, fireTelemetryEvent],
   );
 
   const restartQuickStart = useCallback(
@@ -105,8 +111,12 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
         ...qs,
         [quickStartId]: getDefaultQuickStartState(totalTasks, QuickStartStatus.IN_PROGRESS),
       }));
+      fireTelemetryEvent('Quick Start Initiated', {
+        id: quickStartId,
+        type: 'restart',
+      });
     },
-    [setActiveQuickStartID, setAllQuickStartStates],
+    [setActiveQuickStartID, setAllQuickStartStates, fireTelemetryEvent],
   );
 
   const nextStep = useCallback(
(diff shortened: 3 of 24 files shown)


frontend/packages/topology/src/components/quick-search/QuickSearchList.tsx
```diff
@@ -69,7 +71,12 @@ const QuickSearchList: React.FC<QuickSearchListProps> = ({
             className={cx('odc-quick-search-list__item', {
               'odc-quick-search-list__item--highlight': item.uid === selectedItemId,
             })}
-            onDoubleClick={(e: React.SyntheticEvent) => handleCta(e, item, closeModal)}
+            onDoubleClick={(e: React.SyntheticEvent) => {
+              handleCta(e, item, closeModal);
+              fireTelemetryEvent('Quick Search used', {
+                item: item.name,
```
christianvogt, line 77: We should include additional info about the item:
```suggestion
              fireTelemetryEvent('Quick Search Used', {
                id: item.uid,
                type: item.type,
                name: item.name,
```

frontend/packages/topology/src/components/quick-search/QuickSearchModalBody.tsx
```diff
@@ -89,9 +91,12 @@ const QuickSearchModalBody: React.FC<QuickSearchModalBodyProps> = ({
     (e) => {
       if (selectedItem) {
         handleCta(e, selectedItem, closeModal);
+        fireTelemetryEvent('Quick Search used', {
+          item: selectedItem.name,
```
christianvogt, line 95:
```suggestion
        fireTelemetryEvent('Quick Search Used', {
          id: selectedItem.uid,
          type: selectedItem.type,
          name: selectedItem.name,
```

frontend/packages/topology/src/filters/TopologyFilterBar.tsx
```diff
@@ -78,7 +80,12 @@ const TopologyFilterBar: React.FC<TopologyFilterBarProps> = ({
     <Toolbar className="co-namespace-bar odc-topology-filter-bar">
       <ToolbarContent>
         <ToolbarItem>
-          <QuickSearchButton onClick={() => setIsQuickSearchOpen(true)} />
+          <QuickSearchButton
+            onClick={() => {
+              fireTelemetryEvent('Quick Search Accessed', null);
```
christianvogt, line 85: You can omit the `null` parameter if you make it optional.

Also, you should move this to `TopologyView` as mentioned in another comment.

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -264,6 +268,13 @@ class MastheadToolbarContents_ extends React.Component {
             externalLink: true,
             href: getOCMLink(clusterID),
             image: <img src={redhatLogoImg} alt="" />,
+            callback: () => {
+              if (this._isMounted) {
+                fireTelemetryEvent('Launcher Menu Accessed', {
```
christianvogt, line 273: Not sure why you need the `isMounted` check. Seems unnecessary because the action cannot be executed if the component isn't mounted.

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -264,6 +268,13 @@ class MastheadToolbarContents_ extends React.Component {
...
+                  item: t('masthead~OpenShift Cluster Manager'),
```
christianvogt, line 274: I suggest that this value be a constant instead of the translated string.

frontend/public/components/masthead-toolbar.jsx
DaoDaoNoCode: That was used to test the bug I met before I created `withTelemetry` HOC. I will remove that, thanks for pointing out.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
```diff
@@ -26,6 +28,11 @@ const QuickSearchDetails: React.FC<QuickSearchDetailsProps> = ({ selectedItem, c
         className="odc-quick-search-details__form-button"
         onClick={(e) => {
           handleCta(e, selectedItem, closeModal);
+          fireTelemetryEvent('Quick Search Used', {
+            id: selectedItem.uid,
+            type: selectedItem.type,
+            name: selectedItem.name,
```
christianvogt, line 34: The same event is always fired immediately after `handleCta` utility function is called. Maybe you can pass the `fireTelemtryEvent` function to `handleCta` and it can have a single implementation for firing the `Quick Search Used` event?

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -286,6 +292,11 @@ class MastheadToolbarContents_ extends React.Component {
           externalLink: true,
           href: _.get(item, 'spec.href'),
           image: <img src={_.get(item, 'spec.applicationMenu.imageURL')} alt="" />,
+          callback: () => {
+            fireTelemetryEvent('Launcher Menu Accessed', {
+              item: _.get(item, 'spec.text'),
```
christianvogt, line 297: Name here is a unique identifier from the ConsoleLink CR. Let's pass it as the ID and put the text as the name.
```suggestion
              id: item.name,
              name: _.get(item, 'spec.text'),
```

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -264,6 +265,11 @@ class MastheadToolbarContents_ extends React.Component {
             externalLink: true,
             href: getOCMLink(clusterID),
             image: <img src={redhatLogoImg} alt="" />,
+            callback: () => {
+              fireTelemetryEvent('Launcher Menu Accessed', {
+                item: 'OpenShift Cluster Manager',
```
christianvogt, line 270: Since we want to also capture an ID for the ConsoleLink contributed items, let's duplicate the constant value here as both the ID and the name.
```suggestion
                id: 'OpenShift Cluster Manager',
                name: 'OpenShift Cluster Manager',
```

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
```diff
@@ -89,8 +91,12 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
           [quickStartId]: getDefaultQuickStartState(totalTasks, QuickStartStatus.IN_PROGRESS),
         };
       });
+      fireTelemetryEvent('Quick Start Initiated', {
+        id: quickStartId,
+        type: 'start',
```
christianvogt, line 96: Please add the quick start `name` as well to make it easier to understand the event when looking at the event data.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
```diff
@@ -105,8 +111,12 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
         ...qs,
         [quickStartId]: getDefaultQuickStartState(totalTasks, QuickStartStatus.IN_PROGRESS),
       }));
+      fireTelemetryEvent('Quick Start Initiated', {
+        id: quickStartId,
+        type: 'restart',
```
christianvogt, line 116: Add quick start `name`

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
```diff
@@ -124,12 +134,19 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
         let updatedTaskStatus;
 
         if (status === QuickStartStatus.NOT_STARTED) {
+          fireTelemetryEvent('Quick Start Initiated', {
+            id: activeQuickStartID,
+            type: 'start',
```
christianvogt, line 139: Add quick start `name`

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
```diff
@@ -124,12 +134,19 @@ export const useValuesForQuickStartContext = (): QuickStartContextValues => {
...
+          });
           updatedStatus = QuickStartStatus.IN_PROGRESS;
         } else if (
           status === QuickStartStatus.IN_PROGRESS &&
           taskStatus !== QuickStartTaskStatus.INIT &&
           taskNumber === totalTasks - 1
         ) {
+          fireTelemetryEvent('Quick Start Completed', {
+            id: activeQuickStartID,
```
christianvogt, line 148: Add quick start `name`

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
DaoDaoNoCode: I tried that before I created `withTelemetry` HOC, and it has some issues. Will try it again after @vojtechszocs 's fix on dynamic-demo-plugin folder. (both `yarn build` and `yarn build-dev` don't work for me)

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
DaoDaoNoCode: @christianvogt Actually, `quickStartId` here is `name`. It seems that `name` is used to identify all the quick starts when the functions are implemented. I can change it to
```
{ name: quickStartId, type: 'start' }
```
Do we need to pass `uid` of the quick starts as `id` here? It's hardly used, so there will be some type changes in the functions because we didn't pass `uid` to this component.

christianvogt: The quick start ID is the constant while the name is going to be the human readable value. Quick starts will also be localized so we want to ensure that the ID is a constant value.
The name is useful when looking at the data in segment.

frontend/packages/topology/src/components/quick-search/QuickSearchDetails.tsx
christianvogt: It's fine for now, but it shouldn't have been an issue to pass in a function reference parameter to `handleCta`.

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+initSegment();
+
+export const eventListener = (eventType: string, properties?: any) => {
+  // eslint-disable-next-line no-console
+  console.log('Demo Plugin received telemetry event: ', eventType, properties);
+  switch (eventType) {
+    case 'identify':
+      const { clusterId, id, user } = properties;
```
christianvogt, line 72: There is no `id` property. Extract the id from the `user` object:
```suggestion
      const { clusterId, user } = properties;
      const id = user.metadata?.uid || user.metadata?.name
```

frontend/public/components/masthead-toolbar.jsx
```diff
@@ -286,6 +293,12 @@ class MastheadToolbarContents_ extends React.Component {
           externalLink: true,
           href: _.get(item, 'spec.href'),
           image: <img src={_.get(item, 'spec.applicationMenu.imageURL')} alt="" />,
+          callback: () => {
+            fireTelemetryEvent('Launcher Menu Accessed', {
+              id: item.name,
```
christianvogt, line 298: I think this should be `item.metadata.name`

frontend/packages/console-shared/src/hooks/useTelemetry.ts
```diff
@@ -0,0 +1,16 @@
...
+} from '@console/dynamic-plugin-sdk';
+
+export const useTelemetry = () => {
+  const [extensions] = useResolvedExtensions(isTelemetryListener);
+  return React.useCallback<TelemetryEventListener>(
+    (eventType, properties) => {
+      extensions.forEach((e) => e.properties.listener(eventType, properties));
+    },
+    [extensions],
+  );
```
christianvogt, line 15: I think we should include common properties in the `useTelemetry` hook.
For now the common properties should include the openshift version.
@spadgett anything else?

```suggestion
  const [extensions] = useResolvedExtensions(isTelemetryListener);
  const clusterVersion = useClusterVersion();
  const openshiftVersion = getOpenShiftVersion(clusterVersion);
  return React.useCallback<TelemetryEventListener>(
    (eventType, properties) => {
      extensions.forEach((e) => e.properties.listener(eventType, { openshiftVersion, ...properties }));
    },
    [extensions, openshiftVersion],
  );
```

spadgett: I'm not sure how these properties are used, but cluster ID comes to mind. Note that normal users won't be able to get the version or cluster ID. (We have a bug tracking that issue for version.) User agent also would be good if it's not already captured.

christianvogt: @spadgett how will clusterID be useful and isn't that considered identifying information?
If the user cannot get the OpenShift version it should then be blank in the payload.

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,86 @@
...
+  console.log('Demo Plugin received telemetry event: ', eventType, properties);
+  switch (eventType) {
+    case 'identify':
+      const { clusterId, user } = properties;
+      const id = user.metadata?.uid || user.metadata?.name;
+      // Use md5 hash to anonymize the user
+      const anonymousId = createHash('md5')
+        .update(`${clusterId}-${id}`)
+        .digest('hex');
+      (window as any).analytics.identify(anonymousId, user);
```
christianvogt, line 78: We cannot store identifying information. Do not pass through the `user` object as traits.
```suggestion
      (window as any).analytics.identify(anonymousId);
```

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
christianvogt: After investigation, the adding the name at this point would require some api refactoring and not worth the effort for the initial telemetry release.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
christianvogt: After investigation, the adding the name at this point would require some api refactoring and not worth the effort for the initial telemetry release.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
christianvogt: After investigation, the adding the name at this point would require some api refactoring and not worth the effort for the initial telemetry release.

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.ts
christianvogt: After investigation, the adding the name at this point would require some api refactoring and not worth the effort for the initial telemetry release.

christianvogt:
@DaoDaoNoCode i think the changes look good!

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+  };
+  switch (eventType) {
+    case 'identify':
+      const { user, consoleVersion } = properties;
+      const id = user.metadata?.uid || `${location.host}-${user.metadata?.name}`;
+      // Use SHA1 hash algorithm to anonymize the user
+      const anonymousIdBuffer = await crypto.subtle.digest('SHA-1', new TextEncoder().encode(id));
+      const anonymousIdArray = Array.from(new Uint8Array(anonymousIdBuffer));
+      const anonymousId = anonymousIdArray.map((b) => b.toString(16).padStart(2, '0')).join('');
+      (window as any).analytics.identify(anonymousId, { consoleVersion }, anonymousIP);
```
christianvogt, line 81: Since this is an example, I suggest grabbing all other properties as the payload. It's mo

```suggestion
      const { user, ...otherProperties } = properties;
      const id = user.metadata?.uid || `${location.host}-${user.metadata?.name}`;
      // Use SHA1 hash algorithm to anonymize the user
      const anonymousIdBuffer = await crypto.subtle.digest('SHA-1', new TextEncoder().encode(id));
      const anonymousIdArray = Array.from(new Uint8Array(anonymousIdBuffer));
      const anonymousId = anonymousIdArray.map((b) => b.toString(16).padStart(2, '0')).join('');
      (window as any).analytics.identify(anonymousId, otherProperties, anonymousIP);
```

frontend/packages/console-shared/src/hooks/useTelemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+  extensionsLoadedRef.current = extensionsLoaded;
+  const queueRef = React.useRef<QueueItem[]>([]);
+  const notifyListeners = React.useCallback<TelemetryEventListener>(
+    (eventType, properties) => {
+      if (!extensionsLoadedRef.current) {
+        queueRef.current.push({ eventType, properties });
+      } else {
+        extensionsRef.current.forEach((e) =>
+          e.properties.listener(eventType, {
+            consoleVersion,
```
christianvogt, line 28:
```suggestion
            consoleVersion: window.SERVER_FLAGS.consoleVersion
```

frontend/packages/console-shared/src/hooks/useTelemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+  TelemetryEventListener,
+} from '@console/dynamic-plugin-sdk';
+
+type QueueItem = {
+  eventType: string;
+  properties?: any;
+};
+
+export const useTelemetry = () => {
+  const { consoleVersion } = window.SERVER_FLAGS;
```
christianvogt, line 14: Use the value in place where required.
```suggestion
```

Dockerfile.plugins.demo2
```diff
@@ -0,0 +1,22 @@
+# This image is used for testing OpenShift Console dynamic plugin capabilities.
```
spadgett, line 1: Why do we need two different Dockerfiles for the demo plugin?

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
+const SEGMENT_KEY = '';
```
spadgett, line 1: Do we plan to eventually make segment its own sample plugin, separate from dynamic-demo-plugin?

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const initSegment = () => {
+  const analytics = ((window as any).analytics = (window as any).analytics || []);
+  if (!analytics.initialize) {
```
spadgett, line 5: Prefer early returns, which are easier to understand.

```suggestion
  if (analytics.initialize) {
    return;
  }
  // ...rest of logic...
```

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (analytics.invoked)
+      window.console && console.error && console.error('Segment snippet included twice.');
+    else {
+      analytics.invoked = !0;
```
spadgett, line 9: Why `!0` instead of `true`?

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+        'off',
+        'on',
+        'addSourceMiddleware',
+        'addIntegrationMiddleware',
+        'setAnonymousId',
+        'addDestinationMiddleware',
+      ];
+      analytics.factory = function(e: string) {
+        return function() {
+          var t = Array.prototype.slice.call(arguments);
```
spadgett, line 34: Prefer `let` or `const` to `var` (here and below).

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+      };
+      for (var e = 0; e < analytics.methods.length; e++) {
+        var key = analytics.methods[e];
+        analytics[key] = analytics.factory(key);
+      }
+      analytics.load = function(key: string, e: Event) {
+        const t = document.createElement('script');
+        t.type = 'text/javascript';
+        t.async = !0;
+        t.src = 'https://cdn.segment.com/analytics.js/v1/' + key + '/analytics.min.js';
```
spadgett, line 48: This is probably not an issue in practice since I wouldn't expect key to have special characters, but I'd use `encodeURIComponent` to be safe.

```suggestion
        t.src = 'https://cdn.segment.com/analytics.js/v1/' + encodeURIComponent(key) + '/analytics.min.js';
```

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,89 @@
...
+  console.log('Demo Plugin received telemetry event: ', eventType, properties);
+  const anonymousIP = {
+    context: {
+      ip: '0.0.0.0',
+    },
+  };
+  switch (eventType) {
+    case 'identify':
+      const { user, ...otherProperties } = properties;
+      const id = user.metadata?.uid || `${location.host}-${user.metadata?.name}`;
```
spadgett, line 76: If `user` is defined, I would expect `metadata` to always be defined. So does this need to be...?

```suggestion
      const id = user?.metadata.uid || `${location.host}-${user?.metadata.name}`;
```

frontend/public/components/app.jsx
```diff
@@ -238,6 +240,40 @@ const AppRouter = () => {
   );
 };
 
+const CaptureTelemetry = React.memo(() => {
+  const fireTelemetryEvent = useTelemetry();
+
+  // notify of identity change
+  const user = useSelector(({ UI }) => UI.get('user'));
+  React.useEffect(() => {
+    if (user.metadata?.uid || user.metadata?.name) {
```
spadgett, line 249:
```suggestion
    if (user?.metadata.uid || user?.metadata.name) {
```

frontend/public/components/app.jsx
```diff
@@ -238,6 +240,40 @@ const AppRouter = () => {
...
+      fireTelemetryEvent('identify', { user });
+    }
+    // Only trigger identify event when the user identifier changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [user.metadata?.uid || user.metadata?.name, fireTelemetryEvent]);
```
spadgett, line 254:
```suggestion
  }, [user?.metadata.uid || user?.metadata.name, fireTelemetryEvent]);
```

spadgett (commented):
Looks good overall 👍

Dockerfile.plugins.demo2
DaoDaoNoCode: This is created by Christian. Hey @christianvogt , do you think we still need this Dockerfile?

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
christianvogt: @spadgett yes we do. This is just an example. For devsandbox we will require creating a separate plugin that can use this example as its foundation.

Maybe what we do is create a repo with a service that can be deployed as a plugin and configured with segment.

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
christianvogt: I think we can omit `?` altogether then.
We can play it safe and check if user is not defined, we should not fire an identify event at all.

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
christianvogt: Note that this segment piece is taken as a direct copy from segment.io's provided code snippet.

However since we formatted the code, we may as well update it correctly as well.

DaoDaoNoCode: This script is provided by segment.io, I just copied it without changing anything. I will test if we could change `!0` to `true`.

Dockerfile.plugins.demo2
christianvogt: @spadgett i created this separate dockerfile because it greatly simplified testing. The normal docker file takes forever to build. Running a local build for the demo plugin is quick. This dockerfile assumes a pre-built plugin and simply packages it.

Dockerfile.plugins.demo2
```diff
@@ -0,0 +1,22 @@
...
+#
+# See frontend/dynamic-demo-plugin/README.md for details.
+
+# Stage 0: build the demo plugin
```
christianvogt, line 5: Let's update the documentation for this dockerfile to let users know that the the plugin must be pre-built:
```suggestion
# Stage 0: build the demo plugin
# Manually pre-build the demo plugin:
# cd frontend
# yarn install
# cd dynamic-demo-plugin
# yarn install
# yarn build
```

frontend/public/components/app.jsx
DaoDaoNoCode: If we make this change here we will get `TypeError: Cannot read property 'uid' of undefined`.

Dockerfile.plugins.demo2
spadgett: It's a little weird to have two Dockefiles, but I guess it's OK. Long term we should try to figure out why the build is so slow and try to fix that so it's not needed.

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
```diff
@@ -0,0 +1,90 @@
+const SEGMENT_KEY = '3Din4v9Wvo5LlHz5omATWCHrrgkliWaF';
```
spadgett, line 1: It looks like the SEGMENT_KEY is still defined here?

It would be nice if we can parameterize this somehow during the build so I don't need to change the code, but this is OK for now.

spadgett (commented):
/approve

Thanks @DaoDaoNoCode. It would be good for @christianvogt to give the final LGTM. We'll need a Bugzilla bug to backport to 4.7.z.

frontend/public/components/app.jsx
spadgett: > If we make this change here we will get `TypeError: Cannot read property 'uid' of undefined`.

That worries me to honest. `metadata` should always be defined on a user. Is user getting initialized to `{}` maybe?

frontend/dynamic-demo-plugin/src/utils/telemetry.ts
DaoDaoNoCode: Oh, sorry about forgetting remove this! I will remove it now. Thanks for reminding @spadgett .

christianvogt:
/lgtm

tested with console demo plugin

frontend/public/components/app.jsx
DaoDaoNoCode: I am not sure. @christianvogt Any thought of this?

spadgett:
/hold

This needs a Bugzilla bug before we can merge.

PR #8641, opened 2021-04-13
Consume Redux extensions in React render flow

vojtechszocs:
Depends on #8631

The primary and recommended way to consume Console extensions is via React hooks.

This fits the overall React application-centric design where changes (sourced by user interaction or other systems) cause the application to respond & re-render accordingly.


christianvogt:
@vojtechszocs you have lint errors

PR #8640, opened 2021-04-13
Bug 1949180: Parse Pipelines kinds

rebeccaalpert:
Pipelines kinds weren't being picked up the parser. I added comments so the parser will pick them up and dehumanized them to bring them in line with Admin console.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1949180.

PR #8644, opened 2021-04-13
Remove spaces between Pipeline Kinds translations

andrewballantyne:
Based off of and thus blocked by https://github.com/openshift/console/pull/8640.
/hold

Removes the spaces between Pipeline kinds. ie `Pipeline Run` is now `PipelineRun`. Applies to all Pipeline Models.

cc @rebeccaalpert 

@makambalaji - this likely will impact some Cypress tests, I'll work with you tomorrow to see about how we will address these changes.


andrewballantyne:
/hold cancel

Dependant PR merged.

frontend/packages/console-shared/src/components/multi-tab-list/MultiTabListPage.tsx
```diff
@@ -47,6 +46,18 @@ const MultiTabListPage: React.FC<MultiTabListPageProps> = ({
     }
   };
 
+  const items = Object.keys(menuActions).reduce((acc, key) => {
+    const menuAction: MenuAction = menuActions[key];
+    const label =
+      menuAction.label || menuAction.model?.labelKey ? t(menuAction.model.labelKey) : null;
```
rebeccaalpert, line 52: What we've done elsewhere is include a default to the menuAction.model.label if there's no labelKey.

frontend/packages/pipelines-plugin/src/components/pipelineruns/PipelineRunDetailsPage.tsx
```diff
@@ -16,7 +16,7 @@ const PipelineRunDetailsPage: React.FC<DetailsPageProps> = (props) => {
   const { t } = useTranslation();
   const { kindObj, match } = props;
   const menuActions: KebabAction[] = useMenuActionsWithUserAnnotation(
-    getPipelineRunKebabActions(true),
+    getPipelineRunKebabActions(t)(true),
```
rebeccaalpert, line 19: I know this is something other people are doing a lot in console, but I always get dinged for this whenever Sam reviews my PRs. I think he may have concerns about passing TFunctions around.

rebeccaalpert (changes requested):
The majority looks good to me. I think it would be good to include model.label where model.labelKey doesn't exist in common components (we do that elsewhere). Sam has also consistently objected to passing TFunction as a parameter in my PRs. I know this something widespread in console at this point, but wanted to flag that as a potential issue from his side.

frontend/packages/pipelines-plugin/src/components/pipelineruns/PipelineRunDetailsPage.tsx
andrewballantyne: I'm not a fan of it either... but we chose a hook approach in the Console. Utilities have words that need translated, they cannot use the hook. I'm open to hearing about alternatives... but unless we want to just refresh the browser, we can't really go with the static approach `i18next.t(...)`.

In some of our static models, we went with the `key` approach; Kebabs and K8sKinds... but that only gets us so far with infrastructure. We'd need to start baking into common components an approach that uses keys instead. In the above case, it's an error modal I believe. I suppose we could instead of taking a string, take a "stringKey". 

This harks back to my comment on Slack about adopting i18n fully as a first class citizen. Display strings should not be allowed anymore imo, we force the hand of the developer into i18n... but obv that only works in most cases as Sam noted the exception for k8sKinds we discover through api fetching.

frontend/packages/console-shared/src/components/multi-tab-list/MultiTabListPage.tsx
andrewballantyne: I'll add the fallback. Thanks for the suggestion!

frontend/packages/pipelines-plugin/src/components/pipelineruns/PipelineRunDetailsPage.tsx
andrewballantyne: Slack thread happened here: https://coreos.slack.com/archives/C0176H51F8V/p1618412328043000

Essentially the answer is to use `i18n.t` when inside functions as it gets re-executed each time the function is called and thus pulls the latest setting of the language at that time.

rebeccaalpert (approved):
Looks good to me! Thanks for adding the fallback and following up with Sam on the right approach for reusable components.

jerolimov (approved):
Tested this locally and it works fine :+1:

/lgtm

spadgett:
/hold

@andrewballantyne Can you rebase and see if this still builds cleanly?

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelineruns/list-page/PipelineRunList.tsx
ERROR in /go/src/github.com/openshift/console/frontend/packages/pipelines-plugin/src/components/pipelineruns/list-page/PipelineRunList.tsx(12,19):
TS2304: Cannot find name 't'.
```

spadgett:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/batch/pull-ci-openshift-console-master-e2e-gcp-console/1388126265805901824

PR #8645, opened 2021-04-13
update render of multiple secrets with resource icon/text pairs

jcaianirh:
Before:
image

After:
Screen Shot 2021-04-13 at 4 54 38 PM


yapei:
@jcaianirh Shall we render a link to the secret?

spadgett:
Hey, @jcaianirh. Let's track this as a separate bug and make this a link with RBAC checks. I've removed the reference to Bug 1946922 for now.

PR #8634, opened 2021-04-13
Bug 1949093:  fix a.pf-c-button hover issues

rhamilto:
Bonus:
* correctly alphabetize related rules
* remove duplicated `.pf-c-button.co-p-has-sidebar__sidebar-close` rule

Before:
Screen Shot 2021-04-13 at 7 56 02 AM
Screen Shot 2021-04-13 at 7 56 20 AM

After:
Screen Shot 2021-04-13 at 8 13 43 AM
Screen Shot 2021-04-13 at 8 13 37 AM

PR #8646, opened 2021-04-13
Bug 1949093: Alternate fix for css overrides

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1949093

Alternate fix for the above bug. The merged fix @rhamilto created is merged here:
https://github.com/openshift/console/pull/8634

This fix is from the pf team and it reverts the button component fully back to how it was previously configured. The suggested usage for these overrides was to have them located before any other overrides in the file.

### DIFF
--- a/frontend/public/style/_overrides.scss
+++ b/frontend/public/style/_overrides.scss
@@ -44,9 +44,6 @@
   &:disabled,
   &.pf-m-disabled,
   &.pf-m-aria-disabled {
-    &::after {
-      border-color: var(--pf-c-button--disabled--after--BorderColor);
-    }
     color: var(--pf-c-button--disabled--Color);
     background-color: var(--pf-c-button--disabled--BackgroundColor);
   }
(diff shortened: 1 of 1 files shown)


frontend/public/style/_overrides.scss
```diff
@@ -1,3 +1,57 @@
...
+  }
+  &.pf-m-plain {
+    color: var(--pf-c-button--m-plain--Color);
+    background-color: var(--pf-c-button--m-plain--BackgroundColor);
+  }
+  // default disable styles so they override the overrides above
+  &:disabled,
+  &.pf-m-disabled,
+  &.pf-m-aria-disabled {
+    &::after {
```
mcoker, line 47: Looking at this again, you shouldn't need this `&::after {}` block.

jcaianirh: updated to remove that block

rhamilto:
I think I prefer mine simply because it is less code; that said, it doesn't fix every case this one does, but we don't need to necessarily patch overrides for things we aren't using.

PR #8655, opened 2021-04-14
NETOBSERV-4 Create NetworkPolicy dialog

jotak:
New form to create network policies as an alternative to yaml
cf https://issues.redhat.com/browse/NETOBSERV-4

Checklist:
- [ ] Remove volume snapshot routes workaround when #8979 is merged


jotak:
Hey @andrew-ronaldson , @astoycos 
Concerning the 3 different options provided when a peer is added, what do you think about renaming them:
- Allow namespace-wide inbound traffic (or outbound)
- Allow cluster-wide inbound traffic (or outbound)
- Allow inbound traffic per IP block

I'm suggesting this because, especially the second option "Allow traffic from another namespace in the cluster" doesn't sound good to me, first because it can be from more than just one namespace, and secondly because it's not necessarily other namespaces, it could very well include the present namespace. What do you think?

Also, something else: since network policies with openshift-sdn work sometimes differently, what do you think about having a warning when a field might not be supported by some network types? E.g. the whole egress section could have such a warning note: "Egress network policies aren't supported with OpenShiftSDN network type." in the explanation text.

Ideally I'd like to detect automatically which network type is in use, but afaik this is a cluster configuration that requires admin role, hence normal users wouldn't have access to it (unless you see another way to get that than fetching the cluster Network config?)

astoycos:
Hey Joel! 

For starters I think we should change "Add from source'" to "Add NetworkPolicyPeer" 

Apart from that I agree the way the things are worded now is a bit confusing. Lets take a step back, The options we must be able to express when defining a single network policy peer are 

A Peer defined with.... 

1. A Pod Selector 
2. A Namespace Selector 
3. A Namespace Selector && Pod Selector
4. An IpBlock 

based on the [design](https://marvelapp.com/prototype/8592c3a/screen/79048227)  When we add a New "From Source" we just need to narrow down to

- Same Namespace (i.e 1.)
- Different Namespace (i.e 2 OR (1&&2)) 
- IPBlock (i.e 4) 

So Regarding 

>Allow namespace-wide inbound traffic (or outbound)

"namespace-wide" is a bit confusing to me here

>Allow cluster-wide inbound traffic (or outbound)
Allow inbound traffic per IP block

I think its mostly there but i'd rather see 

-- Allow traffic from/to same namespace 
-- Allow traffic from/to different namespace 
-- Allow traffic from/to IpBlock 

>Also, something else: since network policies with openshift-sdn work sometimes differently, what do you think about having a warning when a field might not be supported by some network types? E.g. the whole egress section could have such a warning note: "Egress network policies aren't supported with OpenShiftSDN network type." in the explanation text.

This is a REALLY good point and I definitely think if we don't have a dynamic warning we need to have at least a static warning somewhere on the form. 

>Ideally I'd like to detect automatically which network type is in use, but afaik this is a cluster configuration that requires admin role, hence normal users wouldn't have access to it (unless you see another way to get that than fetching the cluster Network config?)

So its a cluster configuration yes, but it can't be RE-configured after cluster installation, so I don't really see any point in hiding it from Developers... but yeah I'm not sure what the default is there

Regardless if we can't fetch the config from the CC, you can always just look to see what namespace is there.... In OCP-SDN's case I think it's just `openshift-sdn` and in OVN-K's case its `openshift-ovn-kubernetes`, if you want to be more through you can also see what pods are deployed there as well.  This is a bit hacky but would work.

jotak:
> For starters I think we should change "Add from source'" to "Add NetworkPolicyPeer"
> 

I'm fine with renaming this as well (tbh i was thinking about something like "Add allowed source" to make it more visible that it's about allowing, not restricting - although it's also mentioned elsewhere). Anyway we can discuss about all these points in the next meeting.

> I think its mostly there but i'd rather see
> 
> -- Allow traffic from/to same namespace
> -- Allow traffic from/to different namespace
> -- Allow traffic from/to IpBlock
> 

Well, it doesn't really address my point, because the second option may still include the policy's namespace, so it's not 'different' namespace. I don't think I'm nitpicking, take the example of an application than spans over several namespaces, and you want to restrict the traffic to pods belonging to that application. If I read the options that you're suggesting, my first guess would be that I have to create 2 network policies, one for the same namespace, and another for the other namespaces. But in reality I could perfectly just have 1 network policy, which IMO is more obvious to understand with a wording like "cluster-wide traffic".

Said differently, I'm trying to make apparent the different scopes that are proposed, given those scopes are not exclusive, but inclusive. Unless I'm wrong, it looks like that:

```

 ┌───────────────────────────────────────┐
 │Scope: all (ipblock)                   │
 │                                       │
 │  ┌───────────────────────────────┐    │
 │  │Scope: cluster (pod+ns selec)  │    │
 │  │                               │    │
 │  │  ┌───────────────────────┐    │    │
 │  │  │Scope: namespace       │    │    │
 │  │  │                       │    │    │
 │  │  │  (pod selec)          │    │    │
 │  │  │                       │    │    │
 │  │  └───────────────────────┘    │    │
 │  │                               │    │
 │  └───────────────────────────────┘    │
 │                                       │
 └───────────────────────────────────────┘

```

> Regardless if we can't fetch the config from the CC, you can always just look to see what namespace is there.... In OCP-SDN's case I think it's just `openshift-sdn` and in OVN-K's case its `openshift-ovn-kubernetes`, if you want to be more through you can also see what pods are deployed there as well. This is a bit hacky but would work.

That's a good trick, thanks! (I just need also to check if the console can get these names despite logged in user not seeing them)

andrew-ronaldson:
I think its mostly there but i'd rather see

-- Allow traffic from/to same namespace
-- Allow traffic from/to different namespace
-- Allow traffic from/to IpBlock

==> Well, it doesn't really address my point, because the second option may
still include the policy's namespace, so it's not 'different' namespace.

Sorry if I'm not understanding this Joel but if you want to use the Policy
namespace would you just add source "*Allow traffic from same namespace"*
then the dropdown option "*All pods in the namespace"*? That's how I see it
but maybe that's just me.

I like "Add allowed source" instead of from sources. In the description
area below the heading we could explain the Peers so it is clearer. I agree
the wording needs work and I will work with writers on the team.

[image: Screen Shot 2021-05-07 at 8.43.20 AM.png]

I'm trying to read through SDN policy now to understand that better. I'll
get some questions ready for our meeting.

andrew-ronaldson:
Sorry if I'm not understanding this Joel but if you want to use the Policy namespace would you just add source "**Allow traffic from same namespace**" then the dropdown option "**All pods in the namespace**"? That's how I see it but maybe that's just me.

I like "Add allowed source" instead of from sources. In the description area below the ingress rule heading we could explain the Peers so it is clearer. I agree the wording needs work and I will work with writers on the team.

jotak:
> Sorry if I'm not understanding this Joel but if you want to use the Policy namespace would you just add source "**Allow traffic from same namespace**" then the dropdown option "**All pods in the namespace**"? That's how I see it but maybe that's just me.

We can discuss this in our meeting if needed. My point is if you want to create a rule that is both for the same namespace and other namespaces (since a rule can target multiple namespaces at once)

astoycos:
Hey Joel, thanks for the descriptive feedback! And no worries it's good to nitpick here since it's pretty confusing, I see what you mean... 

>Said differently, I'm trying to make apparent the different scopes that are proposed, given those scopes are not exclusive, but inclusive. Unless I'm wrong, it looks like that:

```
 ┌───────────────────────────────────────┐
 │Scope: all (ipblock)                   │
 │                                       │
 │  ┌───────────────────────────────┐    │
 │  │Scope: cluster (pod+ns selec)  │    │
 │  │                               │    │
 │  │  ┌───────────────────────┐    │    │
 │  │  │Scope: namespace       │    │    │
 │  │  │                       │    │    │
 │  │  │  (pod selec)          │    │    │
 │  │  │                       │    │    │
 │  │  └───────────────────────┘    │    │
 │  │                               │    │
 │  └───────────────────────────────┘    │
 │                                       │
 └───────────────────────────────────────┘
```

This scope diagram is definitely correct, and well done btw! It's counter intuitive but you're 100% right you could allow/deny traffic from pods in the SAME namespace and DIFFERENT namespaces with a single rule. I like this suggestion 

>What about something like:

>Allow peers from the same namespace
Allow peers from the cluster
Allow peers by IP block

The only thing I would advocate against is using the word `cluster` simply because I would't want any user getting Network Policy (Which is ultimately namespaced scoped) with Cluster Network Policy (which will be cluster scoped) I would just follow what they do in the [K8's docs](https://kubernetes.io/docs/concepts/services-networking/network-policies/) and say like 

```
Allow peers from (Specified, Selected, particular, etc) namespaces
```
Whatever word there would work ^

jotak:
It's ready for review.

(Side note: I also did some tries to have multiple views (form / yaml / split) as discussed with @andrew-ronaldson , enabling to switch between views without loosing context, this is pretty promising IMO, but leaving it aside for the moment so that it can be discussed in a more global context)
<details><summary>see preview</summary>

branch: https://github.com/jotak/console/tree/netpolicy-form-edit
</details>

frontend/public/components/network-policies/network-policy-form.tsx
```diff
@@ -0,0 +1,370 @@
...
+        <input
+          className="pf-c-form-control"
+          type="text"
+          onChange={handleNamespaceChange}
+          value={networkPolicy.namespace}
+          id="namespace"
+          name="namespace"
+          aria-describedby="namespace-help"
+          required
+        />
```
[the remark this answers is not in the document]
jotak: ping @andrew-ronaldson 
(I made the same remark previously). If I remember correctly, the reason to repeat the namespace field was that users may want to change the namespace of their policy without impacting globally selected namespace, correct?

frontend/public/components/network-policies/create-network-policy.tsx
```diff
@@ -0,0 +1,47 @@
...
+            data-test="yaml-link"
+            replace
+          >
+            {t('public~Edit YAML')}
+          </Link>
+        </div>
+      </h1>
+      <p className="co-m-pane__explanation">
+        {t('public~NetworkPolicy describes what network traffic is allowed for a set of Pods')}
+      </p>
```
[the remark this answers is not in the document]
jotak: I'd like to have some input from the @openshift/team-ux-review team ; for instance I'm not sure if using 2nd person "you" would be consistent with other texts in the console.

For information, the current text that I'm using is taken from our generated doc for NetworkPolicy that we can see as well in the YAML editor side panel.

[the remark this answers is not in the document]
yapei: > I'd like to have some input from the @openshift/team-ux-review team ; for instance I'm not sure if using 2nd person "you" would be consistent with other texts in the console.

Fair enough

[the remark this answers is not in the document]
jotak: I've changed to `NetworkPolicy can specify how Pods are allowed to communicate with various network entities.`
But still up to suggestions

frontend/public/components/network-policies/network-policy-form.tsx
```diff
@@ -0,0 +1,368 @@
...
+                variant="link"
+                isDisabled={networkPolicy.ingress.rules.length === 0}
+                onClick={removeAllIngress}
+              >
+                {t('public~Remove all')}
+              </Button>
+            </div>
+            <Button variant="secondary" onClick={addIngressRule}>
+              {t('public~Add ingress rule')}
+            </Button>
```
yapei, line 287: Found a usability issue during the testing,  by clicking `Add ingress rule` or `Add egress rule` button, user can add as many ingress/egress rules as wanted or accidentally which will generate some YAML files like
```
  ingress:
    - ports:
        - protocol: TCP
          port: 443
      from:
        - podSelector:
            matchLabels:
              role: frontend
    - ports:
        - protocol: SCTP
          port: 80
      from:
        - podSelector:
            matchLabels:
              role: frontend
          namespaceSelector:
            matchLabels:
              project: myproject
```

However for `spec.ingress.from` we should better generate them in a way 
```
spec:
  podSelector:
    matchLabels:
      role: db
  policyTypes:
  - Ingress
  - Egress
  ingress:
  - from:
    - ipBlock:   //only one occurance
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - namespaceSelector:  //only one occurance
        matchLabels:
          project: myproject
    - podSelector:  //only one occurance
        matchLabels:
          role: frontend
    ports:  //only one occurance
    - protocol: TCP
      port: 6379
```
Screen Shot 2021-05-20 at 4 51 51 PM

yapei: User can add or update rules only in one ingress rule editor, shall we?

jotak: As far as I know the spec allows to have any number of rules, there's no restriction about the number of occurrences of ipblocks / pods/namespaces selectors. It allows to build advanced configurations, e.g. several ip blocks or several namespaces with different labels. Though in terms of "good practices", to be honest, I don't know if it's ok, or if it's more recommended to create multiple policies. This was discussed with the networking team and we decided to allow any number of rules and peers within a policy. (cc @astoycos )

frontend/public/components/network-policies/_create-network-policy.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.co-create-networkpolicy__header .pf-m-secondary {
+  color: var(--pf-global--primary-color--100);
+  padding-left: var(--pf-global--spacer--md);
+  padding-right: var(--pf-global--spacer--md);
+}
+
+.co-create-networkpolicy__header .pf-m-secondary .pf-c-dropdown__toggle-icon {
+  margin-right: 0;
+}
```
yapei, line 63: another user experience issue is several text font size and bold format seems unmatched

Screen Shot 2021-05-20 at 5 13 12 PM

frontend/public/components/network-policies/network-policy-form.tsx
andrew-ronaldson: @jotak The expand/collapse headings on Ingress, Egress and the peers shouldn't be blue. The carrot is the only clickable action and not the title. Thanks!

frontend/public/components/network-policies/network-policy-form.tsx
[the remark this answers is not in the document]
andrew-ronaldson: Thanks @yapei @jotak 
I kept this in the form in its pre-filled state for a few reasons:

- The current YAML-only version puts it into the code in the same way so this is a continuation of that convention
- Because it is part of the Policy and this form is for users learning Network Policy I didn't want to leave it out of the form
- Having the field and supporting text below is helpful to explain each step of the process. I have a request out with our UXD writing team to finesse the micro-copy on this form.

frontend/public/components/network-policies/network-policy-form.tsx
jotak: @andrew-ronaldson this is defined in patternfly's expandable: https://www.patternfly.org/v4/components/expandable-section
You think I should use something different or create a new component?

andrew-ronaldson: There is a component called field groups that should work.
https://www.patternfly.org/v4/components/form#field-groups

jotak: I've replaced ExpandableSection with FormFieldGroupExpandable, here's how it looks:

- Peers collapsed:
Capture d’écran de 2021-05-20 18-17-59

- Peers expanded:
Capture d’écran de 2021-05-20 18-18-06

At the moment I left the default styles for these form groups, which is quite different from what there was in the mockup (especially for Ingress and Egress). Considering what @yapei commented earlier (having several different font sizes), should we change them or not?

frontend/public/components/network-policies/network-policy-rule-config.tsx
```diff
@@ -0,0 +1,166 @@
...
+    case 'allPodsSameNS':
+    case 'somePodsSameNS':
+      return t('public~Peers in the same namespace');
+    case 'somePodsAllNS':
+    case 'somePodsAnyNS':
+    case 'allPodsAnyNS':
+      return t('public~Peers in specified namespaces');
+    case 'ipBlock':
+    default:
+      return t('public~IP block');
```
yapei, line 36: does it look better if we use `Peers by IP block` text here ?

jotak:
renamed label & rebased

frontend/public/components/network-policies/network-policy-form.tsx
andrew-ronaldson: On the ingress rule expandable section the description text below is further down than the PF component. This is a snippet from the PF site. 
`<div class="pf-c-form__field-group-header-description">Field group 1 description text.</div>`

As for the text sizing I'd like to adjust the default components so we have some hierarchy from: 

- Ingress
- - ingress rule
- - - peers

jotak:
After talking with @andrew-ronaldson , there's a couple of changes to do wrt. UX:
New mockup reference: https://marvelapp.com/prototype/2a03af26/screen/79630977

- Increase Ingress / Egress font size (same as Policy type)
- Missing description text for peers
- Add new items (rules, peers) to the top of the list
- Remove rules numbering
- Make dropdowns full width
- Warning text: use https://www.patternfly.org/v4/components/alert/#inline-types (also I'll change warning to info, because warning looks quite aggressive).

jotak:
@yapei @andrew-ronaldson I've implemented the changes described above (last commit).

frontend/public/components/network-policies/network-policy-form.tsx
```diff
@@ -0,0 +1,390 @@
...
+      </div>
+      {showSDNAlert && (
+        <Alert
+          variant="info"
+          title={t(
+            'public~When using the OpenShift SDN cluster network provider, egress network policy is not supported.',
+          )}
+          actionClose={<AlertActionCloseButton onClose={() => setShowSDNAlert(false)} />}
+        />
+      )}
```
yapei, line 265: Shall we move this alert to `Egress` rules when user tries to add `Egress` rules?

yapei: Screen Shot 2021-06-01 at 4 10 32 PM

jotak: The first place where "egress" is mentioned, hence where this information is useful, is the checkbox "Deny all egress traffic", so that's why I think it's preferable to have this alert prior to it. The top of the "Policy type" section made sense to me.
For that reason I think we shouldn't remove it from there ; eventually we could duplicate it so that's it's also visible in Egress. The downside is that it adds more redundant information on screen, which some users could perceive as noise. So, I'm not well decided. @andrew-ronaldson , any thought about that?

jotak: Or perhaps, we could have it displayed as a tooltip on the "deny all egress" checkbox + as this alert in the egress section. wdyt?

jotak:
@yapei the design changed again a little bit, after discussing with Andrew, all pods/ns selectors are now not visible by default and will turn visible when clicking on a link.
This way of doing makes obsolete the second dropdown that we previously had to refine which kind of peer we wanted, when configuring peers.

jotak:
@kyoto @spadgett this is waiting for approval

frontend/public/components/network-policies/network-policy-peer-ipblock.tsx
```diff
@@ -0,0 +1,104 @@
...
+              className="pf-c-form-control"
+              type="text"
+              onChange={(event) => handleExceptionsChange(idx, event.currentTarget.value)}
+              placeholder="10.2.1.0/12"
+              aria-describedby="ports-help"
+              name={`exception-${idx}`}
+              id={`exception-${idx}`}
+              value={exc}
+            />
+            <Button
```
spadgett, line 65: This button needs an aria-label for screen readers.

frontend/public/components/network-policies/network-policy-peer-selectors.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const handleNamespaceSelectorChange = (updated: string[][]) => {
+    onChange(podSelector, updated);
+  };
+
+  const helpTextPodSelector =
+    direction === 'ingress'
+      ? namespaceSelector
+        ? t(
```
spadgett, line 20: Avoid nested ternaries, they're hard to follow.

frontend/public/components/network-policies/network-policy-ports.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  onChange={(event) =>
+                    onSingleChange({ ...port, port: event.currentTarget.value }, idx)
+                  }
+                  placeholder="443"
+                  aria-describedby="ports-help"
+                  name={`${key}-port`}
+                  id={`${key}-port`}
+                  value={port.port}
+                />
+                <Button
```
spadgett, line 82: Missing aria-label

frontend/public/components/network-policies/network-policy-rule-config.tsx
```diff
@@ -0,0 +1,183 @@
...
+    onChange(rule);
+  };
+
+  const removePeer = (idx) => {
+    rule.peers = [...rule.peers.slice(0, idx), ...rule.peers.slice(idx + 1)];
+    onChange(rule);
+  };
+
+  return (
+    <Card style={{ marginBottom: 15 }}>
```
spadgett, line 86: Avoid inline styles

frontend/public/components/network-policies/network-policy-rule-config.tsx
```diff
@@ -0,0 +1,183 @@
...
+                toggleAriaLabel="Peer"
+                isExpanded
+                header={
+                  <FormFieldGroupHeader
+                    titleText={{
+                      text: getPeerRuleTitle(t, direction, peer),
+                      id: `peer-header-${idx}`,
+                    }}
+                    actions={
+                      <Button
```
spadgett, line 147: Missing aria-label

spadgett (commented):
Thanks @jotak. Great work on this overall.

frontend/public/components/network-policies/network-policy-form.tsx
```diff
@@ -0,0 +1,354 @@
...
+        <Title headingLevel="h2">{t('public~Policy type')}</Title>
+      </div>
+      {showSDNAlert && (
+        <Alert
+          variant="info"
+          title={t(
+            'public~When using the OpenShift SDN cluster network provider, egress network policy is not supported.',
+          )}
+          actionClose={<AlertActionCloseButton onClose={() => setShowSDNAlert(false)} />}
+        />
```
sg00dwin, line 228: Screen Shot 2021-06-30 at 12 44 10 PM

Seems this should be a [warning alert](https://www.patternfly.org/v4/components/alert#inline-types)?
In other forms we display `<Alert>` inline so they aren't floating.

```suggestion
        <Alert
          variant="warning"
          isInline
          className="co-alert"
          title={t(
            'public~When using the OpenShift SDN cluster network provider, egress network policy is not supported.',
          )}
          actionClose={<AlertActionCloseButton onClose={() => setShowSDNAlert(false)} />}
        />
```

frontend/public/components/network-policies/_create-network-policy.scss
```diff
@@ -0,0 +1,56 @@
+.co-create-networkpolicy {
+  &__alert {
+    margin-bottom: 0;
+  }
+}
+
+.co-create-networkpolicy__deny .checkbox {
```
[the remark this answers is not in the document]
jotak: Ok, I'm adding the `className` props to `Checkbox` then

frontend/public/components/network-policies/network-policy-form.tsx
jotak: I was using a warning at first but changed to info. The reason is that this banner is always displayed, even when there's no problem because the user uses ovn-k cni, and we don't want to sound too aggressive or worrying while there are chances that there's no issue at all. Also, having it as a floating box allows the user to dismiss it if desired (if it's perceived as annoying).

This situation isn't perfect for sure, but it's temporary, there's a task with the SDN team to expose information about the CNI plugin in use; when it's done, we will be able to remove this message and adapt the display to whatever is relevant.

frontend/public/components/network-policies/network-policy-form.tsx
```diff
@@ -0,0 +1,354 @@
...
+                updateEgressRules(newRules);
+              }}
+              onRemove={() => removeEgressRule(idx)}
+            />
+          ))}
+        </FormFieldGroupExpandable>
+      )}
+      <ButtonBar errorMessage={error} inProgress={inProgress}>
+        <ActionGroup className="pf-c-form">
+          <Button type="submit" isDisabled={!isValid()} id="save-changes" variant="primary">
```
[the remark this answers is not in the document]
jotak: good point, I'll rework that part

frontend/public/components/network-policies/network-policy-model.ts
```diff
@@ -0,0 +1,175 @@
...
+    return {
+      matchLabels: obj,
+    };
+  }
+  return emptyValue;
+};
+
+type Rule = { from?: K8SPeer[]; to?: K8SPeer[]; ports?: K8SPort[] };
+
+const ruleToK8s = (
```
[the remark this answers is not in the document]
jotak: Yes I thought about it :-) there are pros and cons.
True, it's a lot of code that could be just wiped out.
But I don't think it's more error prone, quite the opposite, because the k8s model is not always very consistent, and I don't like to see code to deal with inconsistencies leaking in all the various components, I prefer to define "clean" types and have a more solid base to work with in the underlying components.

To give some examples of the inconsistencies:
- The way a policy affects "ingress" or "egress" rules are different depending on presence of fields and content of policyType field (probably for legacy reasons). In the rewritten model, they obey the same rule.
- Selector field being present or not, empty or not doesn't have exactly the same consequence depending on where it is in the resource definition (parent or peer).

It's also easier to cover how these inconsistencies are handled in tests by calling pure / stateless typescript functions rather than through react component introspection.

And last point, the rewritten model can be adapted to better match the form structure, e.g. a `denyAll` bool to map to checkboxes, instead of a playing with the combination of `policyType` and `ingress` / `egress` fields.

So to sum up, even if it adds hundred lines of code, I think it makes the react component code easier to read and to reason about in the end.

frontend/public/components/network-policies/network-policy-form.tsx
jotak: BTW I could add something to the message like "If your network provider is OVN-Kubernetes, please ignore this message"

frontend/public/components/network-policies/create-network-policy.tsx
```diff
@@ -0,0 +1,49 @@
...
+    ingress: {
+      denyAll: false,
+      rules: [],
+    },
+    egress: {
+      denyAll: false,
+      rules: [],
+    },
+  };
+  const [networkPolicy, setNetworkPolicy] = React.useState(emptyPolicy);
```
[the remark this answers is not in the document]
jotak: That's correct (I was just a little bit ahead of time because in a future PR, this component will have more to do with networkPolicy ; but you're totally right if we consider strictly this commit)
I'll change that

frontend/public/components/network-policies/_create-network-policy.scss
```diff
@@ -0,0 +1,56 @@
...
+  display: flex;
+}
+
+.co-create-networkpolicy__rule-header label {
+  margin-top: 8px;
+}
+
+.co-create-networkpolicy__rule-header .pf-m-secondary {
+  margin: 8px;
+  top: -8px;
```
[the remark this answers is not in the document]
jotak: I can't remember, and it doesn't seem necessary anymore, I'm removing it.

frontend/public/components/network-policies/_create-network-policy.scss
```diff
@@ -0,0 +1,56 @@
...
+/* Style override for FormFieldGroupExpandable: bigger font (~h2), move toggle icon a little bit lower */
+.co-create-networkpolicy__expandable-xl > .pf-c-form__field-group-header .pf-c-form__field-group-header-title-text {
+  font-size: var(--pf-global--FontSize--xl);
+  font-family: var(--pf-global--FontFamily--heading--sans-serif);
+  font-weight: var(--pf-global--FontWeight--normal);
+  line-height: var(--pf-global--LineHeight--md);
+}
+
+.co-create-networkpolicy__expandable-xl > .pf-c-form__field-group-toggle {
+  margin-top: 0.3rem;
```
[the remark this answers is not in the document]
jotak: moving all to `rem`

frontend/public/components/network-policies/_create-network-policy.scss
jotak: This comment was already addressed. "Ingress" and "Egress" are similar to Policy type

jotak:
@spadgett @sg00dwin thanks for reviewing. If I didn't forget anything, all comments should have been addressed or answered.

frontend/public/components/network-policies/_create-network-policy.scss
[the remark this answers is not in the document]
spadgett: We've used `px` in most contexts in console. @rhamilto @sg00dwin any opinion on this?

frontend/public/components/network-policies/network-policy-form.tsx
spadgett: I agree it should be an info message for the reasons you state.

We could try to get the configuration for users who have authority, but if the SDN team is working to expose it to normal users, I'm fine leaving it as-is.

jotak:
PR rebased, @spadgett @andrew-ronaldson I removed the namespace field

frontend/packages/console-app/src/components/network-policies/network-policy-add-peer-dropdown.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { t } = useTranslation();
+  const { title, onSelect } = props;
+  const options = [
+    t('public~Allow pods from the same namespace'),
+    t('public~Allow pods from inside the cluster'),
+    t('public~Allow peers by IP block'),
+  ];
+  return (
+    <div className="form-group co-create-networkpolicy__add-peer">
+      <Dropdown
```
spadgett, line 19: Note: For new dropdowns, you should prefer the PatternFly dropdown. We're trying to move away from the internal one.

OK to leave as-is for this PR, though.

spadgett (approved):
/lgtm

PR #8659, opened 2021-04-14
build strategy i18n miss

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5765

**Solution description:**
Add i18n support to  build strategy filter dropdown present in the build configs list page

**Screenshot:**
Screenshot from 2021-04-15 20-32-10


frontend/public/components/build-config.tsx
```diff
@@ -148,11 +148,11 @@ const buildStrategy = (buildConfig: K8sResourceKind): BuildStrategyType =>
...
-  BuildStrategyType.Devfile,
-  BuildStrategyType.JenkinsPipeline,
-  BuildStrategyType.Source,
-  BuildStrategyType.Custom,
+  { id: BuildStrategyType.Docker, title: i18n.t('public~Docker') },
+  { id: BuildStrategyType.Devfile, title: i18n.t('public~Devfile') },
+  { id: BuildStrategyType.JenkinsPipeline, title: i18n.t('public~JenkinsPipeline') },
+  { id: BuildStrategyType.Source, title: i18n.t('public~Source') },
+  { id: BuildStrategyType.Custom, title: i18n.t('public~Custom') },
 ];
```
andrewballantyne, line 156: Need to move this into `BuildConfigsPage` -- don't want to be doing `i18n.t` in the root of a file. If we do that, the i18n logic fires once at the first import and gets cached. The user changing their language would then see the old language here as it was cached.

frontend/public/components/build.tsx
```diff
@@ -450,7 +451,15 @@ BuildsList.displayName = 'BuildsList';
...
-const allPhases = ['New', 'Pending', 'Running', 'Complete', 'Failed', 'Error', 'Cancelled'];
+const allPhases = [
+  { id: BuildPhase.New, title: i18n.t('public~New') },
+  { id: BuildPhase.Pending, title: i18n.t('public~Pending') },
+  { id: BuildPhase.Running, title: i18n.t('public~Running') },
+  { id: BuildPhase.Complete, title: i18n.t('public~Complete') },
+  { id: BuildPhase.Failed, title: i18n.t('public~Failed') },
+  { id: BuildPhase.Error, title: i18n.t('public~Error') },
+  { id: BuildPhase.Cancelled, title: i18n.t('public~Cancelled') },
+];
```
andrewballantyne, line 462: Same issue here. Please move into `BuildsPage`

andrewballantyne (changes requested):
I'm not entirely sure we wanted to translate Build Phases.

/cc @rebeccaalpert could you help out here?

rebeccaalpert:
I'm not sure about the Build phase; Sam would know. I tagged him on Slack here: https://coreos.slack.com/archives/C0176H51F8V/p1618845621055200.

rebeccaalpert (approved):
Looks good to me. Just need to hear back from Sam about the Build phase question.

spadgett:
Hi, @debsmita1. Typically we don't attempt to translate data coming from resources like build phases. I don't think we translate pod phases today as an example. Even if build phases are well defined, it doesn't seem practical to go through and translate all known values for all resources. I also wouldn't translate the dropdown if values in the table are not translated (as in your screenshot). I'd leave the build filter as-is.

andrewballantyne (approved):

andrewballantyne:
/lgtm

You made the change Sam wanted and Rebecca agreed with the rest of it. Adding the other label.

PR #8669, opened 2021-04-15
Add support for showing the dashboard based on components

bipuladh:
Flexible Deployment Support
- Show `Block and File` dashboard only when Ceph Cluster is present.
- Show `Object` dashboard only when MCG is present.

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/features.ts
+++ b/frontend/packages/ceph-storage-plugin/src/features.ts
@@ -10,7 +10,7 @@ import { FeatureDetector } from '@console/plugin-sdk';
 import { getAnnotations, getName } from '@console/shared/src/selectors/common';
 import { fetchK8s } from '@console/internal/graphql/client';
 import { StorageClassModel } from '@console/internal/models';
-import { OCSServiceModel } from './models';
+import { OCSServiceModel, CephClusterModel, NooBaaSystemModel } from './models';
 import {
   CEPH_STORAGE_NAMESPACE,
   OCS_SUPPORT_ANNOTATION,
@@ -23,16 +23,17 @@ import { StorageClusterKind } from './types';
 
 export const OCS_INDEPENDENT_FLAG = 'OCS_INDEPENDENT';
 export const OCS_CONVERGED_FLAG = 'OCS_CONVERGED';
-export const NOOBAA_FLAG = 'NOOBAA';
-
-// Used to activate NooBaa dashboard
-export const OCS_FLAG = 'OCS';
-// Todo(bipuladh): Remove this completely in 4.6
-export const CEPH_FLAG = 'CEPH';
 
 export const LSO_FLAG = 'LSO';
 
 export const RGW_FLAG = 'RGW';
+
+// Based on the existence of NooBaaSystem
+export const MCG_FLAG = 'MCG';
+// Based on the existence of CephCluster
+export const CEPH_FLAG = 'CEPH';
+// Based on the existence of StorageCluster
+export const OCS_FLAG = 'OCS';
 
 export enum GUARDED_FEATURES {
   // Flag names to be prefixed with "OCS_" so as to seperate from console flags
@@ -127,15 +128,44 @@ export const detectOCS: FeatureDetector = async (dispatch) => {
         (sc: StorageClusterKind) => sc.status.phase !== 'Ignored',
       );
       const isInternal = _.isEmpty(storageCluster?.spec?.externalStorage);
-      dispatch(setFlag(OCS_FLAG, true));
       dispatch(setFlag(OCS_CONVERGED_FLAG, isInternal));
       dispatch(setFlag(OCS_INDEPENDENT_FLAG, !isInternal));
+      dispatch(setFlag(OCS_FLAG, true));
     }
   } catch (error) {
+    dispatch(setFlag(OCS_FLAG, false));
     dispatch(setFlag(OCS_CONVERGED_FLAG, false));
     dispatch(setFlag(OCS_INDEPENDENT_FLAG, false));
-    dispatch(setFlag(OCS_FLAG, false));
   }
+};
+
+export const detectComponents: FeatureDetector = async (dispatch) => {
+  let id = null;
+  let noobaaFound = false;
+  let cephFound = false;
+  const detector = async () => {
+    try {
+      if (!cephFound) {
+        const cephClusters = await k8sList(CephClusterModel, { ns: CEPH_STORAGE_NAMESPACE });
+        if (cephClusters?.length > 0) {
+          dispatch(setFlag(CEPH_FLAG, true));
+          cephFound = true;
+        }
+      }
+      if (!noobaaFound) {
+        const noobaaSystems = await k8sList(NooBaaSystemModel, { ns: CEPH_STORAGE_NAMESPACE });
+        if (noobaaSystems?.length > 0) {
+          dispatch(setFlag(MCG_FLAG, true));
+          noobaaFound = true;
+          clearInterval(id);
+        }
+      }
+    } catch {
+      dispatch(setFlag(CEPH_FLAG, false));
+      dispatch(setFlag(MCG_FLAG, false));
+    }
+  };
+  id = setInterval(detector, 15 * SECOND);
 };
 
 const detectFeatures = (dispatch, csv: ClusterServiceVersionKind) => {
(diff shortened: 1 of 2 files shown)


afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -667,6 +672,24 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+  {
+    type: 'NavItem/Href',
+    properties: {
+      id: 'ocsdashboards',
+      section: 'storage',
+      insertBefore: 'persistentvolumes',
+      componentProps: {
+        // t('ceph-storage-plugin~Overview')
+        name: '%ceph-storage-plugin~Overview%',
+        href: '/ocs-dashboards',
```
afreen23, line 685: why we cant keep just overview ? aligning with component visible on UI.
```suggestion
        href: '/overview',
```

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -127,17 +128,48 @@ export const detectOCS: FeatureDetector = async (dispatch) => {
...
   } catch (error) {
+    dispatch(setFlag(OCS_FLAG, false));
     dispatch(setFlag(OCS_CONVERGED_FLAG, false));
     dispatch(setFlag(OCS_INDEPENDENT_FLAG, false));
-    dispatch(setFlag(OCS_FLAG, false));
   }
 };
 
+export const detectComponents: FeatureDetector = async (dispatch) => {
+  let id = null;
```
afreen23, line 143: since its only for noobaa.
```suggestion
  let hasNoobaa = null;
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -127,17 +128,48 @@ export const detectOCS: FeatureDetector = async (dispatch) => {
...
+      if (!cephFound) {
+        const cephCluster = await k8sList(CephClusterModel, { ns: CEPH_STORAGE_NAMESPACE });
+        if (cephCluster?.length > 0) {
+          dispatch(setFlag(CEPH_FLAG, true));
+          cephFound = true;
+        }
+      }
+      // Test for Presence of NooBaa System
+      if (!noobaaFound) {
+        const noobaaSystem = await k8sList(NooBaaSystemModel, { ns: CEPH_STORAGE_NAMESPACE });
```
afreen23, line 158: nit-pick
```suggestion
        const noobaaSystems = await k8sList(NooBaaSystemModel, { ns: CEPH_STORAGE_NAMESPACE });
```

frontend/packages/ceph-storage-plugin/src/features.ts
```diff
@@ -127,17 +128,48 @@ export const detectOCS: FeatureDetector = async (dispatch) => {
...
 
+export const detectComponents: FeatureDetector = async (dispatch) => {
+  let id = null;
+  let noobaaFound = false;
+  let cephFound = false;
+  const detector = async () => {
+    try {
+      // Test for Presence of Ceph Cluster
+      if (!cephFound) {
+        const cephCluster = await k8sList(CephClusterModel, { ns: CEPH_STORAGE_NAMESPACE });
```
afreen23, line 150: nit-pick
```suggestion
        const cephClusters = await k8sList(CephClusterModel, { ns: CEPH_STORAGE_NAMESPACE });
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: will clash with the cluster overivew.

frontend/packages/ceph-storage-plugin/src/features.ts
bipuladh: it's it id of the returned `setInterval` job.

frontend/packages/ceph-storage-plugin/src/plugin.ts
afreen23: I meant `/storage/overview` or `/storage/overview/ocs`. This route seems better IMO.
Also it will be similar to what `monitroing` section does for their nav items - `/monitoring/dashboards/`

frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -667,6 +672,24 @@ const plugin: Plugin<ConsumedExtensions> = [
       required: [OCS_FLAG],
     },
   },
+  // In case Noobaa is deployed without OCS Operator
+  {
+    type: 'NavItem/Href',
+    properties: {
+      id: 'ocsdashboards',
```
afreen23, line 679:
```suggestion
      id: 'ocs-dashboards',
```

frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -667,6 +672,24 @@ const plugin: Plugin<ConsumedExtensions> = [
       required: [OCS_FLAG],
     },
   },
+  // In case Noobaa is deployed without OCS Operator
```
afreen23, line 675: Why do we need to do this ?
Can't the respective tabs extensions be guarded ?

frontend/packages/ceph-storage-plugin/src/features.ts
afreen23: My request was to specify that its for interval job checking for noobaa only.
Since I can see when noobaa is found its getting unset.

frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -98,14 +101,23 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+      required: [OCS_MODEL_FLAG],
     },
   },
   {
     type: 'FeatureFlag/Custom',
     properties: {
       detect: detectRGW,
     },
+    flags: {
+      required: [MCG_FLAG],
```
afreen23, line 113: rgw only present with noobaa ?

frontend/packages/ceph-storage-plugin/src/plugin.ts
```diff
@@ -98,14 +101,23 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+    },
+  },
+  {
+    type: 'FeatureFlag/Custom',
+    properties: {
+      detect: detectComponents,
```
afreen23, line 119:
```suggestion
      detect: detectOCSComponents,
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: Yes.

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: I prefer `detectComponents`. We are detecting components from OCS Plugin.

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: I can live with ocs-dashboards for now.We shoudl revisit this when we want to make the section better by standardizing routes for OBC, PVC, VS, SC etc..

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: The situation is not so simple. 
Just MCG deployment will not have OCS. So in that case we need to show only the MCG dashboard. Without the block and file. Tab extension guard will not be enough.

frontend/packages/ceph-storage-plugin/src/features.ts
bipuladh: I would prefer `id` because that is what it stores. The `id` of the task.

frontend/packages/ceph-storage-plugin/src/plugin.ts
bipuladh: Let's keep it same https://github.com/openshift/console/pull/8669/files#diff-10b74cbaa04568da87a8244d0807997527af2aa4bf767d6ca29133ec4713eb48R662

afreen23 (commented):
/lgtm

PR #8533, opened 2021-03-31
provide code snippet to customize roles in project access

debsmita1:
**Story:**
https://issues.redhat.com/browse/ODC-5449

**Root analysis:**
To support the admin to configure the roles correctly, the developer console should provide a code snippet for the customization YAML resource (Console CRD).

**Solution description:**
Added snippet in the Console resources which provides a default list of Roles. 

**Screenshot:**
Screenshot from 2021-03-31 19-56-06

**Test setup:**
1. CRD URL example: http://localhost:9000/k8s/cluster/customresourcedefinitions/consoles.operator.openshift.io
Search for customProductName, add this yaml between customProductName and documentationBaseURL.
Add :
```
                  projectAccess:
                    description: projectAccess allows customizing the available list
                      of ClusterRoles in the Developer Console Project access page
                      which can be used by a project admin to specify roles to other
                      users and restrict access within the project.
                    type: object
                    properties:
                      availableClusterRoles:
                        description: availableClusterRoles are the list of ClusterRoles
                          that are assignable to users through the project access
                          tab.
                        type: array
                        items:
                          type: string
```

2. Open the cluster resource of the YAML and open the sidebar.
Cluster config example: http://localhost:9000/k8s/cluster/operator.openshift.io~v1~Console/cluster/yaml

3. Insert the default project access roles snippet and save the resource.

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

PR #8502, opened 2021-03-29
Add customized project access list to server flags

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5447

**Description**: 
Based on the enhancement proposal https://github.com/openshift/enhancements/pull/668, get the project access configuration options from `console-config` or through arguments supplied to bridge. And make it available in `window.SERVER_FLAGS`.

Fetching of options from `Console` CRD into the `console-config` ConfigMap is handled in https://github.com/openshift/console-operator/pull/514

**Test setup:**
Pass options as string to bridge with `-project-access-cluster-roles` or pass a `ConsoleConfig` yaml with `-config` option.

Example:
- Specify project access options to bridge with `bin/bridge -project-access-cluster-roles '["View", "Edit", "Admin"]'`.
- Open localhost:9000 and check `SERVER_FLAGS` in console:
  ```js
  > console.log(window.SERVER_FLAGS.projectAccessClusterRoles)
  '["View", "Edit", "Admin"]'
  ```

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

PR #8672, opened 2021-04-15
Show roles in project access page based on customization

vikram-raj:
Fix: https://issues.redhat.com/browse/ODC-5448

**Description**
Show roles in project access page role based on customization

**Screenshots/gif**
image

Kapture 2021-04-16 at 19 58 18

**Tests coverage**
image

**Test setup**
- Build the backend
- Start backend bridge with ```./bin/bridge -project-access-cluster-roles '["monitoring-edit", "edit", "admin"]'```

Note: putting it work in progress as the backend is not ready.

**Review this commit https://github.com/openshift/console/pull/8672/commits/aa3fd975b2ac4e113cc9cfe900c7f4a336e6302a**


serenamarie125:
@vikram-raj this is great!  Curious if you could add a image or animated gif on how the admin customizes this?

vikram-raj:
> @vikram-raj this is great! Curious if you could add a image or animated gif on how the admin customizes this?

Added GIF in PR description.

andrewballantyne:
@serenamarie125 if you're asking how does an Admin ADD the dropdown option? That's being handled over here: https://github.com/openshift/console/pull/8533

frontend/packages/dev-console/src/components/project-access/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type ClusterRoleKind = K8sResourceCommon & {
+  rules: { [key: string]: string[] }[];
+};
+
+export enum accessRoles {
+  admin = 'Admin',
+  edit = 'Edit',
+  view = 'View',
+}
```
jerolimov, line 13: I think this is more a defaultAccessRoles not instead of an enum, or?

```suggestion
export const defaultAccessRoles = {
  admin: 'Admin',
  edit: 'Edit',
  view: 'View',
};

```

frontend/packages/dev-console/src/components/project-access/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+// for testing until backend support
+// eslint-disable-next-line dot-notation
+window.SERVER_FLAGS['availableClusterRoles'] = ['admin', 'view', 'edit', 'cluster-reader'];
+export const useProjectAccessRoles = () => {
+  // eslint-disable-next-line dot-notation
+  const availableClustersRoles = window.SERVER_FLAGS['availableClusterRoles'];
+  const [clusterRoles, clusterRolesLoaded] = useK8sWatchResource<ClusterRoleKind[]>({
+    kind: ClusterRoleModel.kind,
+    isList: true,
+  });
```
jerolimov, line 23: Can we check the `availableClusterRoles` here to fetch the ClusterRoles only if necessary?

```suggestion
  const [clusterRoles, clusterRolesLoaded] = useK8sWatchResource<ClusterRoleKind[]>(availableClustersRoles ? {
    kind: ClusterRoleModel.kind,
    isList: true,
  } : {});
```

frontend/packages/dev-console/src/components/project-access/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+  admin = 'Admin',
+  edit = 'Edit',
+  view = 'View',
+}
+// for testing until backend support
+// eslint-disable-next-line dot-notation
+window.SERVER_FLAGS['availableClusterRoles'] = ['admin', 'view', 'edit', 'cluster-reader'];
+export const useProjectAccessRoles = () => {
+  // eslint-disable-next-line dot-notation
+  const availableClustersRoles = window.SERVER_FLAGS['availableClusterRoles'];
```
jerolimov, line 19: Small typo, let us call this availableClusterRoles similar to the server flag.

```suggestion
  const availableClusterRoles = window.SERVER_FLAGS['availableClusterRoles'];
```

frontend/packages/dev-console/src/components/project-access/hooks.ts
vikram-raj: We can not pass an empty object to `useK8sWatchResource` as kind is the required field.

gajanan-more:
Screenshot from 2021-04-26 19-39-00
Screenshot from 2021-04-26 19-38-34


Working as expected.
/label qe-approved

vikram-raj:
/hold

Wait for backend [PR](https://github.com/openshift/console/pull/8502) to get merged.

rottencandy:
Since the epic acceptance criteria of [ODC-4650](https://issues.redhat.com/browse/ODC-4650) have not changed and the epic has an `px-ack`, we approve this on behalf of PX.

/label px-approved

andrewballantyne:
@vikram-raj Typescript errors are being reported.

```
ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/project-access/__tests__/hooks.spec.ts
ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/project-access/__tests__/hooks.spec.ts(18,25):
TS2339: Property 'projectAccessClusterRoles' does not exist on type '{ alertManagerBaseURL: string; authDisabled: boolean; basePath: string; branding: string; consoleVersion: string; customLogoURL: string; customProductName: string; documentationBaseURL: string; ... 26 more ...; consolePlugins: string[]; }'.
```

vikram-raj:
> @vikram-raj Typescript errors are being reported.
> 
> ```
> ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/project-access/__tests__/hooks.spec.ts
> ERROR in /go/src/github.com/openshift/console/frontend/packages/dev-console/src/components/project-access/__tests__/hooks.spec.ts(18,25):
> TS2339: Property 'projectAccessClusterRoles' does not exist on type '{ alertManagerBaseURL: string; authDisabled: boolean; basePath: string; branding: string; consoleVersion: string; customLogoURL: string; customProductName: string; documentationBaseURL: string; ... 26 more ...; consolePlugins: string[]; }'.
> ```

It is expected as the backend is not yet merged.

frontend/packages/dev-console/locales/en/devconsole.json
```diff
@@ -434,8 +434,8 @@
   "Rate of transmitted packets dropped": "Rate of transmitted packets dropped",
   "Average Container bandwidth by Pod: received": "Average Container bandwidth by Pod: received",
   "Average Container bandwidth by Pod: transmitted": "Average Container bandwidth by Pod: transmitted",
-  "Project access allows you to add or remove a user&apos;s access to the project. More advanced management of role-based access control appear in <2>Roles</2> and <5>Role Bindings</5>. For more information, see the <8>role-based access control documentation</8> .": "Project access allows you to add or remove a user&apos;s access to the project. More advanced management of role-based access control appear in <2>Roles</2> and <5>Role Bindings</5>. For more information, see the <8>role-based access control documentation</8> .",
```
jerolimov, line 437: Are this change requested by someone or should we keep the old version? Also if we don't have a translation for this strings now, we should avoid unnecessary changes. If there is a reason for that I'm absolutely fine with that change!

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -23,21 +23,27 @@ import {
...
 import { validationSchema } from './project-access-form-validation-utils';
 import ProjectAccessForm from './ProjectAccessForm';
-import { Verb, UserRoleBinding, Roles, roleBinding } from './project-access-form-utils-types';
+import { Verb, UserRoleBinding, roleBinding } from './project-access-form-utils-types';
 
 export interface ProjectAccessProps {
   formName: string;
   namespace: string;
   roleBindings?: { data: []; loaded: boolean; loadError: {} };
+  roles: Roles;
```
jerolimov, line 32: Based on my change in the hook, you need to change this to:

```suggestion
  roles: { data: Roles; loaded: boolean };
```

This makes it easier to check if the data are loaded, instead of checking that the data are empty or not.

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -23,21 +23,27 @@ import {
...
-const ProjectAccess: React.FC<ProjectAccessProps> = ({ formName, namespace, roleBindings }) => {
+const ProjectAccess: React.FC<ProjectAccessProps> = ({
+  formName,
+  namespace,
+  roleBindings,
+  roles,
+}) => {
   const { t } = useTranslation();
-  if (!roleBindings.loaded && _.isEmpty(roleBindings.loadError)) {
+  if ((!roleBindings.loaded && _.isEmpty(roleBindings.loadError)) || _.isEmpty(roles)) {
```
jerolimov, line 42: With the changed hook and changed props above, you can check for the loaded state, instead of expecting that the data are not empty.

```suggestion
  if ((!roleBindings.loaded && _.isEmpty(roleBindings.loadError)) || !roles.loaded) {
```

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -23,21 +23,27 @@ import {
...
     return <LoadingBox />;
   }
 
-  const filteredRoleBindings = filterRoleBindings(roleBindings.data, Roles);
+  const filteredRoleBindings = filterRoleBindings(roleBindings.data, roles);
```
jerolimov, line 46: IMHO, the filter function must not need to know the frontend details, so we can just provide the keys here (the ClusterRole names)

```suggestion
  const filteredRoleBindings = filterRoleBindings(roleBindings.data, Object.keys(roles.data));
```

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -108,8 +114,9 @@ const ProjectAccess: React.FC<ProjectAccessProps> = ({ formName, namespace, role
     <>
       <PageHeading>
         <Trans t={t} ns="devconsole">
-          Project access allows you to add or remove a user&apos;s access to the project. More
-          advanced management of role-based access control appear in{' '}
+          {
+            "Project access allows you to add or remove a user's access to the project. More advanced management of role-based access control appear in "
+          }
```
jerolimov, line 119: Is this change requested by someone or can we keep the old version?

Also if we don't have a translation for this strings now, I would personally avoid unnecessary changes to translation.

If there is a reason for that I'm absolutely fine with that change.

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -131,7 +138,7 @@ const ProjectAccess: React.FC<ProjectAccessProps> = ({ formName, namespace, role
             onReset={handleReset}
             validationSchema={validationSchema}
           >
-            {(formikProps) => <ProjectAccessForm {...formikProps} />}
+            {(formikProps) => <ProjectAccessForm {...formikProps} roles={roles} />}
```
jerolimov, line 141: Based on the changed hook signature you need extract the roles here:

```suggestion
            {(formikProps) => <ProjectAccessForm {...formikProps} roles={roles.data} />}
```

frontend/packages/dev-console/src/components/project-access/__tests__/ProjectAccess.spec.tsx
```diff
@@ -25,6 +26,7 @@ describe('Project Access', () => {
         loaded: false,
         loadError: {},
       },
+      roles: defaultAccessRoles,
```
jerolimov, line 29: Again. Because my recommendation is to change the hook return value, you need to change this test as well:

```suggestion
      roles: {
        data: defaultAccessRoles,
        loaded: true,
      },
```

frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
```diff
@@ -5,7 +5,22 @@ import {
...
+  admin: 'Admin',
+  edit: 'Edit',
+  view: 'View',
+};
+
+export type Roles = {
+  [key: string]: string;
+};
+
+export const getAvailableAccessRoles = () => {
```
jerolimov, line 18: Personal preferences is to provide types here, esp. with this JSON.parse

```suggestion
export const getAvailableAccessRoles = (): string[] | undefined => {
```

frontend/packages/dev-console/src/components/project-access/project-access-form-utils.ts
```diff
@@ -5,7 +5,22 @@ import {
...
+  if (!window.SERVER_FLAGS.projectAccessClusterRoles) return undefined;
+  return JSON.parse(window.SERVER_FLAGS.projectAccessClusterRoles);
+};
+
+export const filterRoleBindings = (roleBindings: RoleBinding[], roles: Roles): RoleBinding[] => {
   return _.filter(roleBindings, (user: RoleBinding) => _.keys(roles).includes(user.roleRef.name));
 };
```
jerolimov, line 25: Also a personal preference which you can also ignore.. The Roles type is a "frontend type" which is not required for this utils function. I would simplify this and change the signature to `clusterRoleNames: string[]` and extract this data if required in the calling component.

```suggestion
export const filterRoleBindings = (roleBindings: RoleBinding[], clusterRoleNames: string[]): RoleBinding[] => {
  return _.filter(roleBindings, (user: RoleBinding) => clusterRoleNames.includes(user.roleRef.name));
};
```

frontend/packages/dev-console/src/components/project-access/hooks.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+  const projectAccessRoles = projectClusterRoles.reduce((acc, projectClusterRole) => {
+    const clusterRoleName = projectClusterRole.metadata.name;
+    acc[clusterRoleName] =
+      projectClusterRole.metadata?.annotations?.['console.openshift.io/display-name'] ||
+      clusterRoleName;
+    return acc;
+  }, {});
+  return projectAccessRoles;
+};
```
jerolimov, line 33: As we discussed this on slack and because I had some strange issues on my cluster, and not on a shared cluster I have done many experiments and debugging sessions in your hook. At the end I found that we have an issue with our Dropdown which doesn't support when the data (items) changed after the first rendering. (We should analyse this issue as well...)

I could not fix this with the usage of `useK8sWatchResource`. But with switch to `useK8sWatchResources` the UI works fine for me. The pro and con is that this method requested not all ClusterRoles but all configured and customized ClusterRoles individually. If the cluster admin configures no "availableClusterRoles" no API request is required.

I returned an extra `loaded` parameter so that the UI can show the LoadingBox bases on this information, so that it is not required that the hook returns an empty object for this use case (and the current Dropdown issue).

Please take a look and change whatever you want. At least on my local cluster this works fine. The fallback to the clusterRoleName works also if the API fails for any reason. (For example there is not such a ClusterRole name.)

```tsx
import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
import { ClusterRoleModel } from '@console/internal/models';
import { K8sResourceCommon, WatchK8sResource } from '@console/internal/module/k8s';
import { defaultAccessRoles, getAvailableAccessRoles, Roles } from './project-access-form-utils';

export type ClusterRoleKind = K8sResourceCommon & {
  rules: { [key: string]: string[] }[];
};

export const useProjectAccessRoles = (): { data: Roles; loaded: boolean } => {
  const availableClusterRoles = getAvailableAccessRoles();

  const watchedClusterRoles = React.useMemo<Record<string, WatchK8sResource>>(() => {
    if (!availableClusterRoles || !availableClusterRoles.length) {
      return {};
    }
    return availableClusterRoles?.reduce((acc, role) => {
      acc[role] = {
        kind: ClusterRoleModel.kind,
        name: role,
      };
      return acc;
    }, {} as Record<string, WatchK8sResource>);
  }, [availableClusterRoles]);

  const clusterRoles = useK8sWatchResources<Record<string, ClusterRoleKind>>(watchedClusterRoles);

  if (!availableClusterRoles || !availableClusterRoles.length) {
    return { data: defaultAccessRoles, loaded: true };
  }

  const mappedRoles = availableClusterRoles.reduce((acc, role) => {
    const clusterRole = clusterRoles[role];
    if (clusterRole && clusterRole.loaded && clusterRole.data) {
      const label =
        clusterRole.data.metadata?.annotations?.['console.openshift.io/display-name'] ||
        clusterRole.data.metadata.name ||
        role;
      acc[role] = label;
    } else {
      acc[role] = role;
    }
    return acc;
  }, {} as Roles);

  const allLoadedOrFailed = Object.values(clusterRoles).every(
    (clusterRole) => clusterRole.loaded || clusterRole.loadError,
  );

  return { data: mappedRoles, loaded: allLoadedOrFailed };
};
```

jerolimov (changes requested):
@vikram-raj As written in slack, I understand now why this change doesn't work while testing it locally. My cluster has more then 250 ClusterRoles because it has multiple additional operators installed. And because `useK8sWatchResource` returns all of them, first 250, then 300+ in my case, your empty check doesn't work. As we also already discussed, an update to the Dropdown items removes the selected title from it. It would be cool to fix this in the Dropdown component. But because this is a common used component I would recommend to keep the workaround and render the table after we know which cluster roles are available.

I added some recommendations to your code. Please take a look. I'm sure the change looks bigger then it is. The biggest change is in the `useProjectAccessRoles` hook itself.

And I'm sure that you need to adjust some more unit tests bases on this change.

jerolimov:
Epic has already been approved by docs and noted that documentation is part of the stretch goal for 4.8 release.

/label docs-approved

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
vikram-raj: This change fixes the `missing i18n key` error on the project access page.

jerolimov:
Reviewed code, tested it locally and works fine.
/lgtm

jerolimov:
I verified this change now again on a cluster, together with PR #8502

Started the cluster via clusterbot `launch openshift/console#8502,openshift/console#8672`

Added a ClusterRole with and one without display-name annotation:

```yaml
kind: ClusterRole
metadata:
  name: test1
  annotations:
    console.openshift.io/display-name: Test number one
rules: []
---
kind: ClusterRole
metadata:
  name: test2
rules: []
```

Searched the `cluster` Console resource and configured the following customisation:


Console path to YAML editor: `/k8s/cluster/operator.openshift.io~v1~Console/cluster/yaml`


```diff
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
+ customization:
+   projectAccess:
+     availableClusterRoles:
+       - admin
+       - test1
+       - test2
+       - test3
```

Open the developer perspective project page, “Project access” tab:

Console path to this page: `/project-details/ns/default/access`

Verified that ClusterRoles are RoleBindings could be created, updated and deleted.

Installed OpenShift Pipelines and OpenShift Serverless operator.

Created new ClusterRoles test5 and test6 and verified the customisation with this entries as well as we had issues earlier if a cluster has more then 250 ClusterRoles.

jerolimov (approved):

PR #8676, opened 2021-04-15
Refactor Console plugin init code

vojtechszocs:
- `loadPluginFromURL` & associated logic moved to dynamic SDK `src/runtime/plugin-loader.ts`
- Console plugin initialization logic moved to dynamic SDK `src/runtime/plugin-init.ts`
- `public/plugins.ts` module is now only responsible for creating the `PluginStore` instance

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 
+import * as _ from 'lodash';
 import { PluginStore } from '@console/plugin-sdk/src/store';
 import { overrideSharedModules } from '../shared-modules';
 import { ConsolePluginManifestJSON } from '../schema/plugin-manifest';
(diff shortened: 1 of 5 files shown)


frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-loader.ts
```diff
@@ -105,6 +107,27 @@ export const registerPluginEntryCallback = (pluginStore: PluginStore) => {
   );
 };
 
+export const loadPluginFromURL = async (baseURL: string) => {
+  const manifest = await fetchPluginManifest(baseURL);
+  return loadDynamicPlugin(baseURL, manifest);
```
vojtechszocs, line 112: FYI, there was `return await` here but I've removed the `await` to simply return the promise itself.

Reference: https://eslint.org/docs/rules/no-return-await

rawagner:
```
public/plugins.ts module is now only responsible for creating the PluginStore instance
```

why wont we move the PluginStore creation to dynamic SDK `runtime` too ?

vojtechszocs:
> ```
> public/plugins.ts module is now only responsible for creating the PluginStore instance
> ```
> 
> why wont we move the PluginStore creation to dynamic SDK `runtime` too ?

Creation of the `PluginStore` instance, as well as the Redux `Store` instance and any other app-specific objects, should be the responsibility of Console application.

PR #8677, opened 2021-04-15
Fix dynamic demo plugin build

vojtechszocs:
- [x] reconcile recent changes with Console root `package.json`
- [x] reconcile recent changes with dynamic SDK `tsconfig.json`

For now, demo plugin code does not import from SDK's main index file (`packages/console-dynamic-plugin-sdk/src/index.ts`). This will change once current static & dynamic SDKs are refactored into two new packages:
- **Console plugin SDK** - contains everything needed to build & develop dynamic plugins
- **Console plugin runtime** - code needed to initialize & load plugins, including static plugin support

Dynamic SDK's `webpack` dependency moved to `devDependencies` to avoid clash with demo plugin's `devDependencies`.

### DIFF
--- a/frontend/dynamic-demo-plugin/tsconfig.json
+++ b/frontend/dynamic-demo-plugin/tsconfig.json
@@ -6,8 +6,9 @@
     "target": "es2016",
     "jsx": "react",
     "allowJs": true,
-    "strict": true,
-    "noUnusedLocals": true
+    "experimentalDecorators": true,
+    "noUnusedLocals": true,
+    "skipLibCheck": true
   },
   "include": ["src"]
 }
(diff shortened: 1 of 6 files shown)


frontend/dynamic-demo-plugin/tsconfig.json
```diff
@@ -6,8 +6,9 @@
     "target": "es2016",
     "jsx": "react",
     "allowJs": true,
-    "strict": true,
```
vojtechszocs, line 9: Note: Console plugin SDK code (and any other code it references) is not compiled in TS [strict mode](https://www.typescriptlang.org/tsconfig#strict).

Once we separate between Console provided APIs (utility functions, components, etc.) in terms of type declaration vs. implementation, we can re-enable TS strict mode in the demo plugin.

christianvogt:
/lgtm

demo plugin builds and runs

PR #8678, opened 2021-04-15
Bug 1948788:  always link to metrics viewer even if graph is empty

rhamilto:
After:

https://user-images.githubusercontent.com/895728/114930954-99f51b80-9e03-11eb-8334-3d8821cdb690.mov


rhamilto:
/hold

I just noticed there is a `<StackChart>` that should be updated as well.

spadgett (approved):
/lgtm

PR #8683, opened 2021-04-16
CONSOLE-2351 Add a "Wrap lines" toggle to log viewers

kdoberst:
This allows the user to turn on or off line wrap when viewing logs.

Addresses [CONSOLE-2351](https://issues.redhat.com/browse/CONSOLE-2351)

Added:
 - "Wrap lines" checkbox
 - CSS class to wrap lines
 -  Persist wrap lines settings when moving to different logs

CONSOLE-2351-WrapLines-Screenshot

https://user-images.githubusercontent.com/82059948/115767962-4c326300-a36f-11eb-8f47-9e25c674de2f.mov

NOTE: Tests were not added because there doesn't appear to be text and the changes are almost entirely on the presentation layer.

### DIFF
--- a/frontend/public/components/utils/_log-window.scss
+++ b/frontend/public/components/utils/_log-window.scss
@@ -30,3 +30,8 @@ $color-log-window-divider: $color-log-window-header-bg;
   white-space: pre;
   width: 0;
 }
+
+.log-window__lines--wrap {
+  white-space: pre-wrap;
+  width: inherit;
+}
--- a/frontend/public/components/utils/log-window.jsx
+++ b/frontend/public/components/utils/log-window.jsx
@@ -1,6 +1,7 @@
 import * as _ from 'lodash-es';
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
+import classnames from 'classnames';
 import { STREAM_EOF, STREAM_PAUSED, STREAM_ACTIVE } from './resource-log';
 import { OutlinedPlayCircleIcon } from '@patternfly/react-icons';
 import { Button } from '@patternfly/react-core';
@@ -114,7 +115,7 @@ class LogWindowWithTranslation extends React.PureComponent {
   }
 
   render() {
-    const { bufferFull, lines, linesBehind, status, t } = this.props;
+    const { bufferFull, lines, linesBehind, status, t, wrapLines } = this.props;
     const { content, height } = this.state;
     // TODO maybe move these variables into state so they are only updated on changes
     const headerText = bufferFull
@@ -131,7 +132,13 @@ class LogWindowWithTranslation extends React.PureComponent {
         <div className="log-window__body">
           <div className="log-window__scroll-pane" ref={this._setScrollPane}>
             <div className="log-window__contents" ref={this._setLogContents} style={{ height }}>
-              <div className="log-window__lines">{content}</div>
+              <div
+                className={classnames('log-window__lines', {
+                  'log-window__lines--wrap': wrapLines,
+                })}
+              >
+                {content}
+              </div>
             </div>
           </div>
         </div>
@@ -154,4 +161,9 @@ LogWindow.propTypes = {
   linesBehind: PropTypes.number.isRequired,
   status: PropTypes.string.isRequired,
   updateStatus: PropTypes.func.isRequired,
+  wrapLines: PropTypes.bool,
+};
+
+LogWindow.defaultProps = {
+  wrapLines: false,
 };
--- a/frontend/public/components/utils/resource-log.tsx
+++ b/frontend/public/components/utils/resource-log.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { Base64 } from 'js-base64';
-import { Alert, AlertActionLink, Button } from '@patternfly/react-core';
+import { Alert, AlertActionLink, Button, Checkbox } from '@patternfly/react-core';
 import * as _ from 'lodash-es';
 import { Trans, useTranslation } from 'react-i18next';
 import {
@@ -10,7 +10,8 @@ import {
   OutlinedWindowRestoreIcon,
 } from '@patternfly/react-icons';
 import * as classNames from 'classnames';
-import { FLAGS } from '@console/shared/src/constants';
+import { FLAGS, USERSETTINGS_PREFIX } from '@console/shared/src/constants';
+import { useUserSettings } from '@console/shared';
 import { LoadingInline, LogWindow, TogglePlay, ExternalLink } from './';
 import { modelFor, resourceURL } from '../../module/k8s';
 import { WSFactory } from '../../module/ws-factory';
@@ -89,6 +90,8 @@ export const LogControls: React.FC<LogControlsProps> = ({
   containerName,
   podLogLinks,
   namespaceUID,
+  toggleWrapLines,
+  isWrapLines,
 }) => {
   const { t } = useTranslation();
   return (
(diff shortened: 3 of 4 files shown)


spadgett:
Thanks @kdoberst 👍 

/assign @rhamilto

frontend/public/components/utils/resource-log.tsx
```diff
@@ -89,8 +89,10 @@ export const LogControls: React.FC<LogControlsProps> = ({
   containerName,
   podLogLinks,
   namespaceUID,
+  toggleWrapLines,
 }) => {
   const { t } = useTranslation();
+  const [wrapLines, setWrapLines] = React.useState(false);
```
spadgett, line 95: We might make this a user setting with `useUserSettings` so it's remembered next time you visit the logs tab. (It also makes the default less important.) @andybraren @christianvogt opinion?

Here's an example: https://github.com/openshift/console/blob/master/frontend/public/components/app-contents.tsx#L64-L67

frontend/public/components/utils/log-window.jsx
```diff
@@ -131,7 +131,9 @@ class LogWindowWithTranslation extends React.PureComponent {
         <div className="log-window__body">
           <div className="log-window__scroll-pane" ref={this._setScrollPane}>
             <div className="log-window__contents" ref={this._setLogContents} style={{ height }}>
-              <div className="log-window__lines">{content}</div>
+              <div className={`log-window__lines${this.props.wrapLines ? '--wrap' : ''}`}>
```
rhamilto, line 134: Add `import classnames from 'classnames';` in the imports at the top of the file.
Add `wrapLines` on line 117 since we are already destructuring.
Then:
```suggestion
              <div className={classnames('log-window__lines', { 'log-window__lines--wrap': wrapLines })}>
```

frontend/public/components/utils/_log-window.scss
```diff
@@ -24,9 +24,16 @@ $color-log-window-divider: $color-log-window-header-bg;
   padding-top: 10px;
 }
 
-.log-window__lines {
+.log-window__lines, .log-window__lines--wrap{
```
rhamilto, line 27: These changes on lines 27, 30-32 aren't necessary if you follow the suggested `classnames()` change below.  Apologies for not being more clear on that in my first round of comments.

frontend/public/components/utils/_log-window.scss
```diff
@@ -24,9 +24,16 @@ $color-log-window-divider: $color-log-window-header-bg;
...
   padding-left: 10px;
   padding-right: 10px;
+}
+
+.log-window__lines {
   white-space: pre;
   width: 0;
```
rhamilto, line 34: I don't think this existing declaration is needed, and it's actually causing a problem with my suggestion
```suggestion
```

frontend/public/components/utils/log-window.jsx
```diff
@@ -151,7 +153,12 @@ export const LogWindow = withTranslation()(LogWindowWithTranslation);
 LogWindow.propTypes = {
   bufferFull: PropTypes.bool.isRequired,
   lines: PropTypes.array.isRequired,
+  wrapLines: PropTypes.bool,
```
rhamilto, line 156: Nit:  maintain alphabetization here (move this to line 156).

rhamilto (commented):
Nice first PR, @kdoberst!  A few suggestions.

frontend/public/components/utils/log-window.jsx
```diff
@@ -151,7 +153,12 @@ export const LogWindow = withTranslation()(LogWindowWithTranslation);
...
   linesBehind: PropTypes.number.isRequired,
   status: PropTypes.string.isRequired,
   updateStatus: PropTypes.func.isRequired,
 };
+
+LogWindow.defaultProps = {
+  wrapLines: false,
```
rhamilto, line 163: On the fence about whether this is necessary.  @spadgett, what's your opinion?

spadgett: We've largely switched to functional components and hooks where we use es6 default values. It's probably a good practice for class components... I'd keep it.

frontend/public/components/utils/log-window.jsx
kdoberst: Made suggested changes. Resolving conversation

frontend/public/components/utils/_log-window.scss
kdoberst: Made suggested changes. Resolving conversation

frontend/public/components/utils/log-window.jsx
kdoberst: Made suggested changes. Resolving conversation

frontend/public/components/utils/_log-window.scss
kdoberst: Made suggested changes. Resolving conversation

frontend/public/components/utils/_log-window.scss
```diff
@@ -24,9 +24,15 @@ $color-log-window-divider: $color-log-window-header-bg;
...
 .log-window__lines {
   padding-left: 10px;
   padding-right: 10px;
   white-space: pre;
   width: 0;
 }
+
+.log-window__lines--wrap {
+  white-space: pre-wrap;
+  width:inherit;
```
rhamilto, line 37: Nit
```suggestion
  width: inherit;
```

frontend/public/components/utils/resource-log.tsx
kdoberst: Good point about persisting the choice. In the ticket it indicates that there is possible a log toolbar redesign coming and that this fix should be kept super simple.  Wondering how simple we want to keep this.

andybraren: Agreed about saving this as a local preference. The log toolbar redesign is still in progress, but when it exists we'll still need to persist that preference somewhere.

I think this checkbox + persisting the preference is the level of simplicity we can target for this near-term PR. 🙂

spadgett: @kdoberst I think it's a simple change since we have the `useUserSettings` hook. You should be able to do something like... (untested)

```suggestion
  const [wrapLines, setWrapLines] = useUserSettings<boolean>(
    `${USERSETTINGS_PREFIX}.log.wrapLines`,
    false,
  );
```

kdoberst: Implemented

kdoberst:
Made suggestions from PR including persisting choice.  Requesting for re-review.

frontend/public/components/utils/_log-window.scss
```diff
@@ -24,9 +24,15 @@ $color-log-window-divider: $color-log-window-header-bg;
   padding-top: 10px;
 }
 
+
```
rhamilto, line 27: Nit:  extra line not needed

frontend/public/components/utils/log-window.jsx
```diff
@@ -131,7 +132,13 @@ class LogWindowWithTranslation extends React.PureComponent {
         <div className="log-window__body">
           <div className="log-window__scroll-pane" ref={this._setScrollPane}>
             <div className="log-window__contents" ref={this._setLogContents} style={{ height }}>
-              <div className="log-window__lines">{content}</div>
+              <div
+                className={classnames('log-window__lines', {
+                  'log-window__lines--wrap': this.props.wrapLines,
```
rhamilto, line 137:
```suggestion
                  'log-window__lines--wrap': wrapLines,
```
and change line 118 to
```
    const { bufferFull, lines, linesBehind, status, t, wrapLines } = this.props;
```
since we are already destructuring props.

frontend/public/components/utils/resource-log.tsx
```diff
@@ -207,6 +221,12 @@ export const ResourceLog: React.FC<ResourceLogProps> = ({
   const linkURL = getResourceLogURL(resource, containerName);
   const watchURL = getResourceLogURL(resource, containerName, bufferSize, true);
 
+  const [wrapLines, setWrapLines] = useUserSettings<boolean>(
```
rhamilto, line 224: Nit:  omit the new line above

yapei:
@schituku will provide testing results when finished

schituku:
The wrap lines checkbox is working and also persisting the choice when going to other pod logs
/label qe-approved

PR #8684, opened 2021-04-16
Bug 1949411:  VolumeSnapshot, VolumeSnapshotClass and VolumeSnapshotConent Details tab is not translated - i18n misses

cyril-ui-developer:
/assign @spadgett


spadgett (approved):
/lgtm

PR #8687, opened 2021-04-16
Add Devfile Sample Endpoint

maysunfaisal:
Signed-off-by: Maysun J Faisal <maysunaneek@gmail.com>

This PR introduces a devfile sample endpoint `POST/PUT /api/devfile/samples/` which expects data
```
registry: "sample-placeholder"
```

Since the devfile samples is still a WIP and to circumvent this, I've added a json `pkg/devfile/sample-placeholder.json` that mimics the output to unblock Console development.


rottencandy:
The endpoint didn't work with data `registry: "sample-placeholder.json"`, but worked with `registry: "pkg/devfile/sample-placeholder.json"`.
The registry path seems to be relative to the directory the bridge was started from.

invincibleJai:
@rottencandy pls take a look 

cc @jhadvig @christianvogt

maysunfaisal:
@rottencandy oops, corrected!

rottencandy:
@maysunfaisal Thanks. Reading the json at runtime is fine, just wanted to confirm if that was the intended behavior.

I also see an "icon" field in one of the samples, is the icon data meant to be stored in the frontend? If not, would it be a good idea to supply the base64 encoded image as the field?

pkg/devfile/sample-placeholder.json
```diff
@@ -0,0 +1,37 @@
+[
+    {
+        "name": "nodejs-basic",
+        "displayName": "Basic NodeJS",
+        "description": "A simple Hello world NodeJS application",
+        "type": "sample",
```
christianvogt, line 6: What does this `type` rerpresent?

elsony: The type is more for our main registry support. In our registry, we support stacks (generic project type support) and samples. ODC currently only supports samples so all the items return to you are sample type only. We reuse the same json for registry query and that's why that field is there. You can safely ignore that since when we do the registry query, we'll apply filter to only return samples in the ODC use case.

maysunfaisal:
@rottencandy the icons are meant to be stored in the registry/sample. I've updated the json entry and included the base64 data for the icons.

rottencandy (approved):
Tested the endpoint, looks good.

pkg/server/devfile-handler.go
```diff
@@ -22,6 +23,35 @@ import (
...
+	err := json.NewDecoder(r.Body).Decode(&data)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to decode response: %v", err)
+		klog.Error(errMsg)
+		serverutils.SendResponse(w, http.StatusBadRequest, serverutils.ApiError{Err: errMsg})
+		return
+	}
+
+	if data.Registry != "" {
+		registry = data.Registry
```
jhadvig, line 40: Does this meant that all the registries will need to be part of the backend code ?

pkg/devfile/types.go
```diff
@@ -0,0 +1,20 @@
+package devfile
+
+// Schema is the index file schema for Samples
+type Schema struct {
+	Name        string            `yaml:"name,omitempty" json:"name,omitempty"`
+	DisplayName string            `yaml:"displayName,omitempty" json:"displayName,omitempty"`
+	Description string            `yaml:"description,omitempty" json:"description,omitempty"`
+	Tags        []string          `yaml:"tags,omitempty" json:"tags,omitempty"`
+	Icon        string            `yaml:"icon,omitempty" json:"icon,omitempty"`
+	Type        string            `yaml:"type,omitempty" json:"type,omitempty"`
```
jhadvig, line 10: Do we want to support any type? or should be have an enum here ?

jhadvig (commented):
The change looks good. Added couple of questions.

pkg/server/devfile-handler.go
maysunfaisal: Devfile registries will be in https://github.com/devfile/registry or anywhere else where its decided to be hosted. The index placeholder json here is what will be returned, when the sample is in place at https://github.com/devfile/registry

pkg/devfile/types.go
maysunfaisal: Good point, I'll pass this along to the https://github.com/devfile/registry  folks when they're writing their schema for the index entries but this info is not to be used in ODC. I have just put it here to perform some simple go test, Discussion about type brought up in this PR too - https://github.com/openshift/console/pull/8687#discussion_r620493235

maysunfaisal: FYI the actual schema for the index is here, so I'll get them to switch to enum, good point 👍 
https://github.com/devfile/registry-support/blob/master/index/generator/schema/schema.go#L79

jhadvig: Please do, thanks 👍

jhadvig: @maysunfaisal it partially does, but we need to use this type here as well, that was my main concern. I would suggest to vendor the `schema.go` here.

maysunfaisal: @jhadvig I've added the enum as type to this PR. I am not going to vendor the other repo yet as its still in development phase and they dont have all the work done yet.

pkg/devfile/types.go
```diff
@@ -0,0 +1,31 @@
+package devfile
```
jhadvig, line 1: why are we copying this instead of vendoring ?

maysunfaisal: Sorry, I've had to wait for PR https://github.com/devfile/registry-support/pull/51 to be merged before vendoring as you requested.

I've updated, PTAL!

jhadvig:
@maysunfaisal the change looks good 👍 thanks !
Could you please squash the commits only into two commits?
- One that provides the vendor changes (`go.mod`, `go.sum`, `/vendor`)
- Second that provides changes in the codebase (`/pkg`)
Thanks !

maysunfaisal:
@jhadvig updated!

go.mod
```diff
@@ -5,8 +5,9 @@ go 1.13
 require (
 	github.com/coreos/go-oidc v2.1.0+incompatible
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
-	github.com/devfile/api v0.0.0-20201126204309-ec222215253e
-	github.com/devfile/library v0.0.0-20201203185121-a216646cf21d
+	github.com/devfile/api/v2 v2.0.0-20210211160219-33a78aec06af
+	github.com/devfile/library v1.0.0-alpha.2
+	github.com/devfile/registry-support/index/generator v0.0.0-20210505173027-d06fe2bb3ee8
```
jhadvig, line 10: So I dont see where this pkg is being used.

jhadvig (approved):
/lgtm
/approve

jhadvig:
@maysunfaisal thanks for addressing comments!

rohitkrai03 (approved):
/lgtm

Did a proper QE testing of the PR along with the frontend integration PR https://github.com/openshift/console/pull/8794 using a cluster created through cluster bot. The backend API works as expected to list out all the sample devfiles.

GIF - 

https://user-images.githubusercontent.com/6041994/117885025-ee889b00-b2ca-11eb-9646-05b38bbe0d12.mov



/label qe-approved 
on behalf of devconsole QE team after QE testing the PR as agreed upon.

/label px-approved

/label docs-approved
added on behalf of docs team as the epic has doc-ack label. The feature is dev preview and marked with no-docs label.

PR #8688, opened 2021-04-17
Bug 1949404: Fix label selector translation on pvc creation page

yapei:
@rebeccaalpert  @spadgett  Could any one help review?

### DIFF
--- a/frontend/public/components/storage/create-pvc.tsx
+++ b/frontend/public/components/storage/create-pvc.tsx
@@ -256,9 +256,9 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
         {useSelector && (
           <NameValueEditorComponent
             nameValuePairs={nameValuePairs}
-            valueString="Selector"
-            nameString="Label"
-            addString="Add Value"
+            valueString={t('public~Selector')}
+            nameString={t('public~Label')}
+            addString={t('public~Add value')}
             readOnly={false}
             allowSorting={false}
             updateParentData={handleNameValuePairs}
--- a/frontend/public/locales/en/public.json
+++ b/frontend/public/locales/en/public.json
@@ -562,6 +562,8 @@
   "Access mode": "Access mode",
   "Desired storage capacity": "Desired storage capacity",
   "Use label selectors to request storage": "Use label selectors to request storage",
+  "Selector": "Selector",
+  "Add value": "Add value",
   "PersistentVolume resources that match all label selectors will be considered for binding.": "PersistentVolume resources that match all label selectors will be considered for binding.",
   "Create PersistentVolumeClaim": "Create PersistentVolumeClaim",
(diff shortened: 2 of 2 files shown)


frontend/public/locales/en/persistent-volume-claim.json
```diff
@@ -1,6 +1,8 @@
 {
   "PersistentVolumeClaim details": "PersistentVolumeClaim details",
   "Label selector": "Label selector",
+  "Selector": "Selector",
+  "Add Value": "Add Value",
```
yapei, line 5: I'm not sure where should they go~`public.json` or `persistent-volume-claim.json`. How do we determine where should they reside?

yapei:
@spadgett  I see `frontend` test is failing, is it caused by my changes? I see several conflict messages and `i18n files are not up to date. Commit them to fix.`  Not sure which one caused the failure

frontend/public/locales/en/persistent-volume-claim.json
spadgett: Hi, @yapei. We're moving to having a namespace per package in the console instead of a separate namespace for each module. So I would just put this in `public`. `persistent-volume-claim.json` will eventually go away.

frontend/public/components/storage/create-pvc.tsx
```diff
@@ -256,9 +256,9 @@ export const CreatePVCForm: React.FC<CreatePVCFormProps> = (props) => {
         {useSelector && (
           <NameValueEditorComponent
             nameValuePairs={nameValuePairs}
-            valueString="Selector"
-            nameString="Label"
-            addString="Add Value"
+            valueString={t('persistent-volume-claim~Selector')}
+            nameString={t('public~Label')}
+            addString={t('persistent-volume-claim~Add Value')}
```
spadgett, line 261: Not new, but we're trying to fix the case to conform to PatternFly standards as well internationalize the console.

```suggestion
            addString={t('persistent-volume-claim~Add value')}
```

spadgett (commented):
Thanks @yapei

rebeccaalpert:
Hey @yapei - If you run `yarn i18n` from the `frontend` folder, that'll fix your frontend tests. You don't have to manually update the json files; that command will do it for you. :) 

Everything else looks ok to me (same comments as Sam).

frontend/public/components/storage/create-pvc.tsx
yapei: updated

frontend/public/locales/en/persistent-volume-claim.json
yapei: thank you @spadgett  updated

yapei:
@spadgett  you comments are addressed, could you pay another look?

spadgett (approved):
/lgtm

PR #8691, opened 2021-04-18
Bug 1922063: "Virtual Machine" should be "Templates" in template wizard

avivtur:
In this example, fixing message in the template wizard when creating a customized template from existing one.

before (taken from BZ): 
screencapture-0-0-0-0-9000-k8s-ns-openshift-virtualization-new-wizard-1611893990827 (1) (1)

after: 
Screenshot from 2021-04-18 17-42-11

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1922063


avivtur:
In this example we can see that the string is changed only for the template wizard.

Template wizard:
bz_1922063_after_template_wizard

VM wizard:
bz_1922063_after_vm_wizard

PR #8702, opened 2021-04-19
[WIP] Bug 1897621: Auth test.Login test.logs in as kubeadmin user, Timeout

dtaylor113:
During CI the very first step of the first Protractor test which attempts to login seems to occasionly timeout/fail.  This seems to periodically be an issue when CI is running expecially slow.  We saw this last week for a few hours.  Protractor screenshots from recently failed PRs just shows the three loading dots.  

I can't reproduce this locally, or running `test-protractor.sh` against a remote cluster. Extended wait period to 10 seconds after redirect.

### DIFF
--- a/frontend/integration-tests/tests/login.scenario.ts
+++ b/frontend/integration-tests/tests/login.scenario.ts
@@ -11,8 +11,10 @@ const { BRIDGE_KUBEADMIN_PASSWORD } = process.env;
 
 describe('Auth test', () => {
   beforeAll(async () => {
+    console.log(`\n----> getting ${appHost}`);
     await browser.get(appHost);
-    await browser.sleep(6000); // Wait long enough for the login redirect to complete
+    console.log('----> sleeping 10 seconds');
+    await browser.sleep(10000); // Wait long enough for the login redirect to complete
   });
 
   describe('Login test', async () => {
--- a/frontend/integration-tests/views/login.view.ts
+++ b/frontend/integration-tests/views/login.view.ts
@@ -10,22 +10,33 @@ export const pf3Login = until.presenceOf($('.login-pf'));
 export const pf4Login = until.presenceOf($('[data-test-id="login"]'));
 
 export const selectProvider = async (provider: string) => {
+  console.log('  ----> selectProvider: getting idpLink');
   const idpLink = element(by.cssContainingText('.idp', provider));
+  console.log('  ----> selectProvider: while !(await idpLink.isPresent())');
   while (!(await idpLink.isPresent())) {
+    console.log('    ----> selectProvider: idpLink not Present, get appHost, sleep 10 secs.');
     await browser.get(appHost);
-    await browser.sleep(3000);
+    await browser.sleep(10000);
   }
+  console.log('  ----> selectProvider: idpLink.click()');
   await idpLink.click();
 };
 
 export const login = async (providerName: string, username: string, password: string) => {
+  console.log('----> in login');
   if (providerName) {
+    console.log('----> calling selectProvider(providerName)');
     await selectProvider(providerName);
   }
+  console.log('----> wait for visibility of username input');
   await browser.wait(until.visibilityOf(nameInput));
+  console.log('----> type username');
   await nameInput.sendKeys(username);
+  console.log('----> type password');
   await passwordInput.sendKeys(password);
+  console.log('----> click submit');
   await submitButton.click();
+  console.log('----> wait for userDropdown');
   await browser.wait(until.presenceOf(userDropdown));
 };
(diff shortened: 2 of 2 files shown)


frontend/integration-tests/tests/login.scenario.ts
```diff
@@ -4,15 +4,15 @@ import { appHost } from '../protractor.conf';
 import * as loginView from '../views/login.view';
 
 const JASMINE_DEFAULT_TIMEOUT_INTERVAL = jasmine.DEFAULT_TIMEOUT_INTERVAL;
-const JASMINE_EXTENDED_TIMEOUT_INTERVAL = 1000 * 60 * 3;
+const JASMINE_EXTENDED_TIMEOUT_INTERVAL = 1000 * 60 * 4;
```
spadgett, line 7: I wouldn't change this. If it's taking more than 3 minutes for login to work, that's a problem. I think it is better for the tests to fail in that case.

It might be better if we reduce this value honestly now that the OLM tests have been moved to Cypress. IIRC, we needed a long wait because create the catalog source in the OLM tests took a while.

frontend/integration-tests/tests/login.scenario.ts
```diff
@@ -4,15 +4,15 @@ import { appHost } from '../protractor.conf';
...
 const KUBEADMIN_IDP = 'kube:admin';
 const KUBEADMIN_USERNAME = 'kubeadmin';
 const { BRIDGE_KUBEADMIN_PASSWORD } = process.env;
 
 describe('Auth test', () => {
   beforeAll(async () => {
     await browser.get(appHost);
-    await browser.sleep(6000); // Wait long enough for the login redirect to complete
+    await browser.sleep(8000); // Wait long enough for the login redirect to complete
```
spadgett, line 15: We should determine whether the tests are failing because

1. It takes the login page too long to load
2. It takes too long to redirect back after login

If the problem is (1), changing the browser sleep here makes sense. If it's (2), we probably need to increase the timeout for this call: https://github.com/openshift/console/blob/master/frontend/integration-tests/views/login.view.ts#L29

spadgett: If we do change this sleep, I might give it more cushion like 10s.

dtaylor113: Added some debugging output, but this flake doesn't happen often:
```
Test Suites & Specs:

1) Auth test
----> getting https://console-openshift-console.apps.jephilli-4-8-04-20-0629.devcluster.openshift.com
----> sleeping 5 seconds
   2) Login test
----> in login
----> wait for visibility of username input
----> type username
----> type password
----> click submit
----> wait for userDropdown
----> after login
      ✔ logs in as kubeadmin user
```

The difficult part will be seeing this flake.

spadgett: @dtaylor113 We might have to merge some debugging log messages so we catch it in CI and don't have to keep retesting this PR.

dtaylor113:
Hi @spadgett, was able to provide some debugging output for this flake:
```
Test Suites & Specs:

1) Auth test
----> getting https://console-openshift-console.apps.ci-op-h254wl4p-75d12.origin-ci-int-gce.dev.rhcloud.com
----> sleeping 5 seconds


   2) Login test
----> in login
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
      ✖ logs in as kubeadmin user (2 failures)
```
Next lines are:
```
if (providerName) {
    await selectProvider(providerName);
  }
console.log('----> wait for visibility of username input');
```
Since we don't see the line "----> wait for visibility of username input", it might've timed out in `selectProvider(providerName)`,
I'll add more logging.
However, I'm not sure if this will determine which one of the following is causing the failure:
1. It takes the login page too long to load
2. It takes too long to redirect back after login

frontend/integration-tests/tests/login.scenario.ts
spadgett: @dtaylor113 Let's increase the sleep at this spot to 10s and rollback the other changes in the PR. It sounds like the we're not waiting long enough for the select IDP page to load.

dtaylor113: @spadgett Done :-)

dtaylor113:
`prow/e2e-gcp-console` failed with exact timeout error we're trying to fix.  Bumped wait to 14secs.

dtaylor113:
Wonder if this is related to #8820

dtaylor113:
Closing in favor of #8820

PR #8706, opened 2021-04-19
Bug 1921106: A11y Violation: button name(s) on Utilization Card on Cluster Dashboard

dtaylor113:
This bug was originally thrown in the Cypress `other-routes.spec.ts` test which initially tests URL `/` which results in the Overview Dashboard.  I was able to reproduce this error by reducing our `cy.wait(...)` from 5 seconds to 1 second which resulted in `cy.testA11y()` executing while the dashboard charts were still rendering:
image

Investigating further I restored the original `cy.wait` of 5 seconds and determined that it was long enough to address this particular BZ.  

However, in an effort to remove any `cy.waiting`, I added a `waitFor` method for each url to determine when the url/view/page was truly loaded.  Since `other-routes.spec.ts` loads a variety of one-off routes, each one needed their own `waitFor` criteria.

Fixed this BZ by adding a `waitFor () => cy.byTestID('skeleton-chart').should('not.exist')` check for route `/` (Overview Dashboard).

This fix should:
1. reduce uneccesary waiting and speed up our CI test runs
2. execute `cy.testA11y` exactly after the url/page is loaded

### DIFF
--- a/frontend/packages/integration-tests-cypress/tests/crud/other-routes.spec.ts
+++ b/frontend/packages/integration-tests-cypress/tests/crud/other-routes.spec.ts
@@ -1,4 +1,6 @@
 import { checkErrors } from '../../support';
+import { listPage } from '../../views/list-page';
+import { detailsPage } from '../../views/details-page';
 
 describe('Visiting other routes', () => {
   before(() => {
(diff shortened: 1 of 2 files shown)


spadgett (approved):
/lgtm

Thanks @dtaylor113 !

zherman0 (approved):

frontend/packages/integration-tests-cypress/tests/crud/other-routes.spec.ts
```diff
@@ -13,41 +15,120 @@ describe('Visiting other routes', () => {
...
       : []),
   ];
   otherRoutes.forEach((route) => {
-    it(`successfully displays view for route: ${route}`, () => {
-      cy.visit(route);
-      // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(5000); // wait for page to load
+    it(`successfully displays view for route: ${route.url}`, () => {
+      cy.visit(route.url);
+      cy.url().should('include', `${route.url}`);
```
spadgett, line 126: Unnecessary backticks.

```suggestion
      cy.url().should('include', route.url);
```

Should it be equal to or just include?

dtaylor113: I "think" _include_, whole url would be `http://..`, Route is a relative route `/api-resource/ns/default...`.

spadgett: It would be better to get the path from the URL and then compare it to what we expect rather than doing a substring match. Otherwise testing `/pods/foo` then `/pods` would incorrectly pass the URL check here before it had been changed.

spadgett: Or maybe we can prepend `Cypress.config().baseUrl` to `route.url`.

We should rename `route.url` to `route.path` if it's not a full URL. That tripped me up here :)

dtaylor113:
/retest
failed to acquire lease

frontend/packages/integration-tests-cypress/tests/crud/other-routes.spec.ts
dtaylor113: thanks, fixed all of these

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

PR #8711, opened 2021-04-19
Bug 1951221: remove VMDisksFirehose component in favor of a hook

suomiy:
- so it doesn't load templates on kubernetes


suomiy:
rebased

PR #8713, opened 2021-04-19
[WIP] Bug 1950460 Change Request size input to a number spinner input

kdoberst:
Addresses [Bug 1950460](https://bugzilla.redhat.com/show_bug.cgi?id=1950460)

Changes:
 - Changes free text input to a number spinner input 
     - Adds a "plus" and "minus" buttons to allow the user to easily increment the value.  
     - The "minus" button is disabled when trying to go below the existing value.

Bug-195046

https://user-images.githubusercontent.com/82059948/115305452-f326bd00-a12b-11eb-8854-646a758ccc0e.mov


kdoberst:
This currently is using the existing Number Spinner.  There is a Beta [Number Spinner](https://www.patternfly.org/v4/components/number-input) in PatternFly.  The next step would be to change the existing Number Spinner to call the PatternFly Number Spinner.

rhamilto:
I think we want some margin between the `+` button and the unit dropdown so they don't abut.

frontend/public/components/utils/_number-spinner.scss
```diff
@@ -23,3 +23,6 @@
   top: 4px;
 }
 
+.pf-c-input-group > .co-m-number-spinner {
+  margin-right: 20px;
```
rhamilto, line 27: Since you really only need the spacing when an item follows a number spinner, I recommend adding margin-left to the unit dropdown.

Additionally, as a general rule, we try to avoid descendent (in this case, child) selectors like this one since they are reliant on the markup.

kdoberst:
@rhamilto   @spadgett  @andrewballantyne  Added a second commit to pull in Patternfly's Number Input component.  This was done instead of completely removing the current number spinner is to make updates easier.  The Patternfly Number Input is currently in Beta.

This impacts numerous places in the console including the knative-plugin and a couple of the dev-console components.

andrewballantyne:
/retitle [WIP] Bug 1950460: Change Request size input to a number spinner input

Retitling to pick up the BZ.

frontend/public/components/utils/request-size-input.tsx
```diff
@@ -2,32 +2,37 @@ import * as React from 'react';
...
+import { NumberSpinner } from './number-spinner';
 
 export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   children,
   defaultRequestSizeUnit,
   defaultRequestSizeValue,
   describedBy,
   dropdownUnits,
-  inputClassName,
+  // inputClassName,
```
spadgett, line 13: We should go ahead and remove this instead of just commenting it out if it's not needed anymore. Do you know if any code is using this prop?

spadgett (commented):
Nice work!

frontend/public/components/utils/request-size-input.tsx
kdoberst: Good catch. I'll don't think this prop is being used, but since it has been a little bit, I'll double check.

spadgett: Yeah, I suspect TypeScript would catch it if it was. It should be safe to remove if it builds when commented out.

frontend/public/components/utils/request-size-input.tsx
```diff
@@ -84,7 +86,7 @@ export type RequestSizeInputProps = {
   describedBy?: string;
   step?: number;
   minValue?: number;
-  inputClassName?: string;
+  // inputClassName?: string;
```
spadgett, line 89: Remove commented out code :)

PR #8718, opened 2021-04-20
fix Quick Search CTA in topology

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5760

**Analysis / Root cause**: 
The "Start building your application" and the Quick Search "Add to project" catalog icon (in the top left) don't seem to be working when I click on them.

**Solution Description**: 
stopPropagation is required for the button.

Kapture 2021-04-20 at 20 41 21


rottencandy (approved):
Tested locally.
/lgtm

PR #8723, opened 2021-04-21
Add functionality to edit d/dc via form or yaml interchangeably

divyanshiGupta:
Story: https://issues.redhat.com/browse/ODC-5454

This PR adds the functionality to edit d/dc using form or yaml interchangeably. 
Some code refactor is also done in this PR for making few components reusable for edit-form. 

https://user-images.githubusercontent.com/20724543/117016523-6b86a400-ad10-11eb-9925-db21128f4753.mov


frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentFormEditor.tsx
```diff
@@ -3,6 +3,8 @@ import { K8sResourceKind } from '@console/internal/module/k8s';
 import AdvancedSection from './AdvancedSection';
 import DeploymentStrategySection from './deployment-strategy/DeploymentStrategySection';
 import ImagesSection from './images/ImagesSection';
+import SwitchToYAMLAlert from '@console/shared/src/components/alerts/SwitchToYAMLAlert';
```
karthikjeeyar, line 6: nit: sort import

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentPage.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import Helmet from 'react-helmet';
+import { StatusBox } from '@console/internal/components/utils';
+import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
+import { K8sResourceKind } from '@console/internal/module/k8s';
+import { useTranslation } from 'react-i18next';
+import { RouteComponentProps } from 'react-router-dom';
```
karthikjeeyar, line 7: sort imports

karthikjeeyar:
Not able to uncheck the paused rollouts checkbox
paused-rollouts

frontend/packages/dev-console/src/components/edit-deployment/EditDeployment.tsx
```diff
@@ -0,0 +1,82 @@
...
+        actions.setStatus({ submitSuccess: '' });
+        actions.setStatus({ submitError: e.message });
+      });
+  };
+
+  return (
+    <Formik
+      initialValues={initialValues}
+      onSubmit={handleSubmit}
+      // validationSchema={{}}
```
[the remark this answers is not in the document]
karthikjeeyar: Ah you mentioned it in your PR Description.  sorry I missed it

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/utils/types.ts
```diff
@@ -1,31 +1,33 @@
...
+  [DeploymentStrategyType.recreateParams]: 'Recreate',
+  [DeploymentStrategyType.rollingParams]: 'Rolling',
+  [DeploymentStrategyType.customParams]: 'Custom',
 };
 
 export const KubernetesDeploymentStrategyHelpText = {
-  [DeploymentStrategyType.recreate]: 'Recreate',
-  [DeploymentStrategyType.rolling]: 'Rolling Update',
+  [DeploymentStrategyType.recreateParams]: 'Recreate',
+  [DeploymentStrategyType.rollingUpdate]: 'Rolling Update',
```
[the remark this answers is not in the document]
divyanshiGupta: Its a To-do. Mentioned in description.

frontend/packages/dev-console/src/components/edit-deployment/EditDeployment.tsx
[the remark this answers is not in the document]
divyanshiGupta: Its a To-do. Mentioned in description.

karthikjeeyar:
Based on design command's field help text should be below the label, right now it is below the textColumnField.

command-help-text

divyanshiGupta:
> Based on design command's field help text should be below the label, right now it is below the textColumnField.
> 
> command-help-text

@karthikjeeyar yes because UX for similar usecase is same as above in console for example deploy-image registry options.

divyanshiGupta:
cc: @bmignano

frontend/packages/dev-console/src/plugin.tsx
```diff
@@ -272,6 +272,19 @@ const plugin: Plugin<ConsumedExtensions> = [
...
   },
+  {
+    type: 'Page/Route',
+    properties: {
+      exact: true,
+      path: '/edit-deployment/ns/:ns',
+      loader: async () =>
+        (
+          await import(
+            './components/edit-deployment/EditDeploymentPage' /* webpackChunkName: "dev-console-edit" */
```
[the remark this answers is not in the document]
divyanshiGupta: I thought I have added deployment to that.

frontend/packages/dev-console/src/components/edit-deployment/images/AdvancedImageOptions.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <ExpandCollapse
+      textExpanded={t('devconsole~Hide advanced image options')}
+      textCollapsed={t('devconsole~Show advanced image options')}
+    >
+      <ResourceDropdownField
+        name="formData.imagePullSecret"
+        label={t('devconsole~Pull Secret')}
+        helpText={t(
+          'devconsole~Secret for authentication when pulling image from a secured registry',
```
sahil143, line 38:
```suggestion
          'devconsole~Secret for authentication when pulling image from a secured registry.',
```

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookForm.tsx
```diff
@@ -1,52 +1,61 @@
...
+              ),
+            },
+          ]}
+        />
+        <DropdownField
+          name={`formData.deploymentStrategy.data.${dataAttribute}.${lifecycleHook}.failurePolicy`}
+          label={t('devconsole~Failure Policy')}
+          title={t('devconsole~Select a Failure Policy')}
+          items={FailurePolicyOptions}
+          helpText={t('devconsole~Fail the deployment if the hook fails')}
```
sahil143, line 52:
```suggestion
          helpText={t('devconsole~Fail the deployment if the hook fails.')}
```

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -1,4 +1,24 @@
-export const getContainerNames = (containers: { [key: string]: any }[]) => {
+import { ContainerSpec, K8sResourceKind } from '@console/internal/module/k8s';
+import * as _ from 'lodash';
```
sahil143, line 2: sort import

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +28,415 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+): DeploymentStrategy => {
+  const { strategy } = deployment.spec ?? {};
+  let type: string;
+
+  if (resourceType === Resources.OpenShift) {
+    type = strategy?.type ?? DeploymentStrategyType.rollingParams;
+    return {
+      type,
+      lifecycleAction: LifecycleAction.execNewPod,
+      ...getStrategyData(type, strategy, deployment.metadata?.name, deployment.metadata?.namespace),
```
sahil143, line 172: is there any case where metadata will be undefined?
```suggestion
      ...getStrategyData(type, strategy, deployment.metadata.name, deployment.metadata.namespace),
```

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +28,415 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+    };
+  }
+
+  type = strategy?.type ?? DeploymentStrategyType.rollingUpdate;
+  return {
+    type,
+    ...(type === DeploymentStrategyType.rollingUpdate
+      ? getStrategyData(type, strategy, deployment.metadata?.name, deployment.metadata?.namespace)
```
sahil143, line 180: same as above

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
```diff
@@ -0,0 +1,104 @@
...
+      formData: {
+        name: resName,
+        project: { name: resNamespace },
+        deploymentStrategy,
+      },
+    },
+  } = useFormikContext<FormikValues>();
+  const [{ value: lifecycleHookExist }] = useField<boolean>(
+    `formData.deploymentStrategy.lifecycleHooks.${lifecycleHookName}`,
+  );
```
sahil143, line 32: can't we get these value from formit context? Why do we need this?

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/RollingStrategy.tsx
```diff
@@ -21,20 +23,22 @@ const RollingStrategy: React.FC<StrategyFieldProps> = ({ resourceType }) => {
...
-        name="deploymentStrategy.data.maxSurge"
+        name={`formData.deploymentStrategy.data.${dataAttribute}.maxSurge`}
         label={t('devconsole~Maximum number of surge Pods')}
         helpText={t(
           'devconsole~The maximum number of pods that can be scheduled above the original number of pods while the rolling deployment is in progress. This can be either a percentage (10%) or a whole number (1).',
         )}
       />
-      <AdvancedStrategyOptions />
+      {resourceType === Resources.OpenShift && (
+        <AdvancedStrategyOptions dataAttribute="rollingParams" resourceObj={resourceObj} />
```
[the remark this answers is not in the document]
divyanshiGupta: Yes there are no advanced strategy options for deployment. Refer https://docs.openshift.com/container-platform/4.6/rest_api/workloads_apis/deployment-apps-v1.html

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/ExecNewPodForm.tsx
```diff
@@ -4,52 +4,58 @@ import { useFormikContext, FormikValues } from 'formik';
...
         envPath={['spec', 'template', 'spec', 'containers']}
       />
       <InputField
-        name={`deploymentStrategy.data.${lifecycleHook}.execNewPod.volumes`}
+        name={`formData.deploymentStrategy.data.${dataAttribute}.${lifecycleHook}.execNewPod.volumes`}
         label={t('devconsole~Volumes')}
-        helpText={t('devconsole~List of named volumes to copy to the hook pod.')}
+        helpText={t(
+          'devconsole~List of comma (,) separated named volumes to copy to the hook pod.',
+        )}
```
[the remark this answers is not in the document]
divyanshiGupta: Yes I changed it so that the user knows how they are supposed to enter the values

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
divyanshiGupta: Since its a variable depending upon which lch is being added.

frontend/packages/dev-console/src/components/edit-deployment/ContainerField.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { t } = useTranslation();
+  const {
+    values: {
+      formData: { containers },
+    },
+  } = useFormikContext<FormikValues>();
+  return (
+    <FormGroup fieldId="container-name">
+      <span className="co-section-heading-tertiary">
+        {t('devconsole~Container:')} &nbsp;
```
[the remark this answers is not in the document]
divyanshiGupta: Its getting removed by prettier.

bmignano:
> > Based on design command's field help text should be below the label, right now it is below the textColumnField.
> > command-help-text
> 
> @karthikjeeyar hmm but this is how TextColumnField help-text is shown elsewhere because by default it shows the help-text at bottom of the field.

Thanks for the tag! I've touched base with some other designers and I would recommend following the design and placing the help text below the label. This form field is dynamic and since you can add multiple rows I think the better experience would be to have the help text directly below the label rather than allowing it to potentially be pushed down by multiple rows. In addition, this field can be slightly more complex than others, so I think it makes sense to show the help text before / above the control.

karthikjeeyar:
@divyanshiGupta You can log a ticket to fix the helptext in a separate PR, I ll leave it up to you :)

karthikjeeyar:
@divyanshiGupta I have verified the PR locally, its works fine. Let me know once you are done with the todo items that are mentioned in the PR, will add lgtm

divyanshiGupta:
> > > Based on design command's field help text should be below the label, right now it is below the textColumnField.
> > > command-help-text
> > 
> > @karthikjeeyar hmm but this is how TextColumnField help-text is shown elsewhere because by default it shows the help-text at bottom of the field.
> 
> Thanks for the tag! I've touched base with some other designers and I would recommend following the design and placing the help text below the label. This form field is dynamic and since you can add multiple rows I think the better experience would be to have the help text directly below the label rather than allowing it to potentially be pushed down by multiple rows. In addition, this field can be slightly more complex than others, so I think it makes sense to show the help text before / above the control.

@bmignano Thanks for the clarification. So do you want this changed across console or just for this usecase?

bmignano:
> > > > Based on design command's field help text should be below the label, right now it is below the textColumnField.
> > > > command-help-text
> > > 
> > > 
> > > @karthikjeeyar hmm but this is how TextColumnField help-text is shown elsewhere because by default it shows the help-text at bottom of the field.
> > 
> > 
> > Thanks for the tag! I've touched base with some other designers and I would recommend following the design and placing the help text below the label. This form field is dynamic and since you can add multiple rows I think the better experience would be to have the help text directly below the label rather than allowing it to potentially be pushed down by multiple rows. In addition, this field can be slightly more complex than others, so I think it makes sense to show the help text before / above the control.
> 
> @bmignano Thanks for the clarification. So do you want this changed across console or just for this usecase?

Yes, let's update this wherever this field is used so we are consistent. Thank you!!

sahil143:
@divyanshiGupta Input field does not show  up for `Command`. There should one to start with.
Screenshot 2021-04-30 at 6 59 25 PM

divyanshiGupta:
> @divyanshiGupta I have verified the PR locally, its works fine. Let me know once you are done with the todo items that are mentioned in the PR, will add lgtm
@sahil143 I am not showing the input field here since it's optional. If the user wants they can add using the add button. Do you think otherwise?

sahil143:
> @sahil143 I am not showing the input field here since it's optional. If the user wants they can add using the add button. Do you think otherwise?

Just wanted to confirm if it's intentional because it's showing up in the demo that posted in PR description

sahil143:
In mid lifecycle hook clicking the radio button `Tags the current image as an image stream tag if the deployment succeeds` breaks the UI

bmignano:
> > @sahil143 I am not showing the input field here since it's optional. If the user wants they can add using the add button. Do you think otherwise?
> 
> Just wanted to confirm if it's intentional because it's showing up in the demo that posted in PR description

In the design, I show the input field by default even if it's empty. That's what we do on the Environment tab as well when there are no env variables yet, so I mimicked that behavior. It also may be strange for the button to read "Add another argument" when there isn't one at all. What do you think @divyanshiGupta @sahil143 ?

sahil143:
> > > @sahil143 I am not showing the input field here since it's optional. If the user wants they can add using the add button. Do you think otherwise?
> > 
> > 
> > Just wanted to confirm if it's intentional because it's showing up in the demo that posted in PR description
> 
> In the design, I show the input field by default even if it's empty. That's what we do on the Environment tab as well when there are no env variables yet, so I mimicked that behavior. It also may be strange for the button to read "Add another argument" when there isn't one at all. What do you think @divyanshiGupta @sahil143 ?

@bmignano IMO, We should show the input field since that's the behaviour in other parts(environment variable) as well. If it's optional, we will not show the red astrick mark on the input label.

sahil143:
Also, the command is required in a lifecycle hook? If it is then we should definitely show the input field

Screenshot 2021-04-30 at 7 29 56 PM

divyanshiGupta:
@sahil143 PTAL.

divyanshiGupta:
@bmignano advanced fields for rolling strategy `Update Period` and `Interval` fields are not shown in the doc so I have added them as per the UX in 3.11.

https://user-images.githubusercontent.com/20724543/116713705-ac21ad00-a9f2-11eb-8ec3-b8b6f4b58870.mov

Let me know if that is fine with you or the UX should be something else.

bmignano:
@divyanshiGupta Adding those fields is fine with me to align with the 3.11 UX! However, in your clip it looks like the advanced options toggle moves below the content when it's in the open state. We should make sure the content is below the toggle as shown in PF [here](https://www.patternfly.org/v4/components/expandable-section).

divyanshiGupta:
> @divyanshiGupta Adding those fields is fine with me to align with the 3.11 UX! However, in your clip it looks like the advanced options toggle moves below the content when it's in the open state. We should make sure the content is below the toggle as shown in PF [here](https://www.patternfly.org/v4/components/expandable-section).

@bmignano If you see the lifecycle hooks are below the toggle but the additional fields are added along the other fields as it was in 3.11 but I think I will move the additional fields below the toggle as well. LMKWDYT.

sahil143:
@divayanshi while editing a lifecycle hook the image stream dropdown is not prefilled
lifecycle-hook-edit



missing close bracket on (runtime only)

Screenshot 2021-05-03 at 4 01 23 PM

sahil143:
It seems like when the volume field is cleared in lifecycle hooks it's still present as an empty string value which is causing it to throw an error on save
Screenshot 2021-05-03 at 4 22 31 PM
lifecycle-hook-volume

bmignano:
> > @divyanshiGupta Adding those fields is fine with me to align with the 3.11 UX! However, in your clip it looks like the advanced options toggle moves below the content when it's in the open state. We should make sure the content is below the toggle as shown in PF [here](https://www.patternfly.org/v4/components/expandable-section).
> 
> @bmignano If you see the lifecycle hooks are below the toggle but the additional fields are added along the other fields as it was in 3.11 but I think I will move the additional fields below the toggle as well. LMKWDYT.

@divyanshiGupta I see. I think in 3.11 we used a link to show and hide additional fields so the behavior was kind of different. Since we're using an expandable section here, I'd recommend keeping any additional fields underneath the toggle that reveals and hides them.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
```diff
@@ -21,30 +23,59 @@ const DeploymentStrategySection: React.FC<DeploymentStrategySectionProps> = ({
...
+            resNamespace,
+            resourceType,
+          );
+          const strategyData = {
+            ...(_.omit(resourceObj.spec?.strategy, [
+              'rollingParams',
+              'recreateParams',
+              'customParams',
+              'rollingUpdate',
+            ]) ?? {}),
```
andrewballantyne, line 71: Lodash omit always returns an object.
```suggestion
            ])),
```

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
```diff
@@ -21,30 +23,59 @@ const DeploymentStrategySection: React.FC<DeploymentStrategySectionProps> = ({
...
+            type: value,
+            ...strategyDefaultValues,
+          };
+          initialValues.formData.deploymentStrategy = strategyData;
```
andrewballantyne, line 75: Not sure you ever want to mutate initialValues after the first usage. Formik does support updating to them but I'd reckon that's not something we should ever do.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/ExecNewPodForm.tsx
```diff
@@ -4,52 +4,58 @@ import { useFormikContext, FormikValues } from 'formik';
...
+import { FormSection } from '@patternfly/react-core';
 
-const ExecNewPodForm: React.FC<{ resourceObj: K8sResourceKind; lifecycleHook: string }> = ({
-  resourceObj,
-  lifecycleHook,
-}) => {
+const ExecNewPodForm: React.FC<{
+  resourceObj: K8sResourceKind;
+  lifecycleHook: string;
+  dataAttribute: string;
```
andrewballantyne, line 12: Nit: Separate type when there is more than one or two (in some cases).

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
```diff
@@ -0,0 +1,137 @@
...
+  } = useFormikContext<FormikValues>();
+  const [{ value: lifecycleHookExist }] = useField<boolean>(
+    `formData.deploymentStrategy.${dataAttribute}.${lifecycleHookName}.exists`,
+  );
+  const [showForm, setShowForm] = React.useState(false);
+  const resourceType = getResourcesType(resourceObj);
+
+  const onRemove = () => {
+    const strategy = deploymentStrategy;
+    strategy[dataAttribute][lifecycleHookName] = undefined;
```
andrewballantyne, line 39: This will mutate the object... perhaps a deep clone on line 38?

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
```diff
@@ -0,0 +1,137 @@
...
+      `formData.deploymentStrategy.${dataAttribute}.${lifecycleHookName}`,
+      data[dataAttribute][lifecycleHookName],
+    );
+    setFieldValue(
+      `formData.deploymentStrategy.imageStreamData.${lifecycleHookName}`,
+      data.imageStreamData[lifecycleHookName],
+    );
+  };
+  return (
+    <div>
```
andrewballantyne, line 61: Fragment maybe? There are usages for divs in the case of form/grid usages but otherwise we should avoid unstyled divs 🤔

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookForm.tsx
```diff
@@ -1,52 +1,82 @@
...
-const LifecycleHookForm: React.FC<{ lifecycleHook: string; resourceObj: K8sResourceKind }> = ({
-  lifecycleHook,
-  resourceObj,
-}) => {
+const LifecycleHookForm: React.FC<{
+  lifecycleHook: string;
+  resourceObj: K8sResourceKind;
+  dataAttribute: string;
+  onSubmit: () => void;
+  onClose: () => void;
```
andrewballantyne, line 19: Standalone type please.

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+  };
+};
+
+export const getStrategyData = (
+  type: string,
+  strategy: any,
+  resName: string,
+  resNamespace: string,
+  resourceType: string,
+) => {
```
andrewballantyne, line 102: Is it possible to get a return type for this? Or will that be messy? It'll make it easier to call upon this method in the future if the full signature is noted.

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+          resourceType,
+        )
+      : {}),
+  };
+};
+
+export const getTriggersAndImageStreamValues = (
+  deployment: K8sResourceKind,
+  resourceType: string,
+) => {
```
andrewballantyne, line 207: Same here

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+  };
+};
+
+export const getUpdatedContainers = (
+  containers: ContainerSpec[],
+  fromImageStreamTag: boolean,
+  isi: ImageStreamImageData,
+  imageName?: string,
+  envs?: EnvVar[],
+) => {
```
andrewballantyne, line 297: Same here

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+  };
+  return newContainers;
+};
+
+export const getUpdatedLchData = (
+  lch: LifecycleHookData,
+  lchName: string,
+  lcAction: string,
+  imageStreamData: LifecycleHookImagestreamData,
+) => {
```
andrewballantyne, line 314: Same here

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+        },
+      }),
+      ...(lcAction === LifecycleAction.tagImages && {
+        tagImages: getUpdatedTagImages(),
+      }),
+    },
+  };
+};
+
+export const getUpdatedStrategy = (strategy: DeploymentStrategy, resourceType: string) => {
```
andrewballantyne, line 351: Same here

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -8,3 +29,501 @@ export const getContainerNames = (containers: { [key: string]: any }[]) => {
...
+    }
+    default:
+      return {};
+  }
+};
+
+export const convertEditFormToDeployment = (
+  formValues: EditDeploymentFormData,
+  deployment: K8sResourceKind,
+) => {
```
andrewballantyne, line 450: Same here

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
```diff
@@ -1,4 +1,25 @@
-export const getContainerNames = (containers: { [key: string]: any }[]) => {
+import * as _ from 'lodash';
+import { ContainerSpec, EnvVar, K8sResourceKind } from '@console/internal/module/k8s';
```
andrewballantyne, line 2: Can we go with `DeploymentKind` instead of `K8sResourceKind`?

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-validation-utils.ts
```diff
@@ -0,0 +1,137 @@
+import * as yup from 'yup';
+import { TFunction } from 'i18next';
```
andrewballantyne, line 2: Change to the way things are done for functions. Can you update your functions not to take a `TFunction`.

* Components use `useTranslation` hook
* Functions that use `t` should use `i18n.t` instead as it pulls the global lang state at use
* Don't use `i18n.t` on root-file constants as it only gets executed when the file is imported the first time (cached)

```suggestion
import i18n from '@console/internal/i18n';
```

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
```diff
@@ -21,30 +23,59 @@ const DeploymentStrategySection: React.FC<DeploymentStrategySectionProps> = ({
...
 
   return (
     <FormSection title={t('devconsole~Deployment strategy')}>
       <DropdownField
-        name="deploymentStrategy.type"
+        name="formData.deploymentStrategy.type"
         label={t('devconsole~Strategy type')}
         items={DeploymentStrategyDropdownData[resourceType].items}
         helpText={DeploymentStrategyDropdownData[resourceType].helpText[type]}
+        onChange={(value) => {
```
rohitkrai03, line 57: Pull this out into a callback.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+    initialValues,
+  } = useFormikContext<FormikValues>();
+  const [{ value: lifecycleHookExist }] = useField<boolean>(
+    `formData.deploymentStrategy.${dataAttribute}.${lifecycleHookName}.exists`,
+  );
+  const [showForm, setShowForm] = React.useState(false);
+  const resourceType = getResourcesType(resourceObj);
+
+  const onRemove = () => {
```
rohitkrai03, line 37: Use `useCallback`.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <NamespacedPage disabled variant={NamespacedPageVariants.light}>
+      <Helmet>
+        <title>{heading}</title>
+      </Helmet>
+      <StatusBox
+        loaded={loaded}
+        loadError={loadError}
+        label={`${heading}${t('devconsole~form')}`}
```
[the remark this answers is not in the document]
divyanshiGupta: It seems weird having no space between `Deployment` and `form` since these are two different words.

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+  errors,
+  values: { editorType, formData, yamlData },
+}) => {
+  const { t } = useTranslation();
+  const resourceType = getResourcesType(resource);
+
+  const isStale = React.useMemo(
+    () => resource.metadata.resourceVersion !== formData.resourceVersion,
+    [resource.metadata.resourceVersion, formData.resourceVersion],
+  );
```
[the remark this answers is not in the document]
divyanshiGupta: Not sure what do you mean by cheap but I wanted to calculate `isStale` only when resource/formData resource version changes that is why used `useMemo`.

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { history } from '@console/internal/components/utils';
+import {
+  convertDeploymentToEditForm,
+  convertEditFormToDeployment,
+} from './utils/edit-deployment-utils';
+import { safeJSToYAML } from '@console/shared/src/utils/yaml';
+
+const EditDeploymentForm: React.FC<FormikProps<FormikValues> & {
+  heading: string;
+  resource: K8sResourceKind;
```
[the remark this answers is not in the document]
divyanshiGupta: This is used for both `Deployment` and `DeploymentConfig` and hence `DeploymentKind` cannot be used here as the two have a few different attributes like DeploymentStrategy, Triggers etc.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
divyanshiGupta: In this case its needed to mutate initialValues and we are doing it at other places as well.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
divyanshiGupta: With fragment all form styles were not getting applies properly.

frontend/packages/dev-console/src/components/edit-deployment/utils/edit-deployment-utils.ts
divyanshiGupta: Same as above.

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentPage.tsx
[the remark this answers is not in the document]
andrewballantyne: I'm confused... yes, I agree?

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentForm.tsx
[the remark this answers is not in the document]
andrewballantyne: You're calculating it when it changes... but why? Why not always? Booleans are either true or false. Any use-case I can think of here would be able to diff and understand that `true` from the last loop is the same as `true` from this loop -- with or without memo. 

"Cheap" as in not a computational cost that is worth memo'ing... it's not creating a new object or doing a copy of fields or anything. It's almost as simple of a logical check you can get in coding.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
divyanshiGupta: With fragment all form styles were not getting applies properly.

frontend/packages/dev-console/src/components/edit-deployment/EditDeploymentForm.tsx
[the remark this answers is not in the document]
andrewballantyne: Okies, I guess we can continue with `K8sResourceKind`... it's probably something we should really get away from relying on because it helps us in almost no way other than `has metadata`.

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
andrewballantyne: Hmmm... interesting. Not sure it brings value, if it's not ignored we likely are doing questionable stuff with our `initialValues` as they stop being 'initial' when we can mutate them and change things. What happens if you remove this line?

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/advanced-options/LifecycleHookField.tsx
andrewballantyne: kk, thanks!

frontend/packages/dev-console/src/components/edit-deployment/deployment-strategy/DeploymentStrategySection.tsx
divyanshiGupta: When imageStream component is rendered it takes values from initialValues and populate the dropdowns accordingly. I am updating the initialValues to add image-stream values when the user adds the values in tagImages section of the lifecyclehook form so that when the user closes the form or navigates from yaml-editor to form-editor the previously selected values still show up.

rohitkrai03:
@divyanshiGupta Seems like theres too much gap between the heading and toggle filed.

Screenshot 2021-05-06 at 12 07 03 AM

karthikjeeyar:
In Yaml View-> If the yaml is updated or if it has a new version/content, it doesn't allow the user to `save` until they reload to get new content in place.

image

This error is not shown in `form View` on submit, so it reverts the new changes/yaml content that was caused by elsewhere. I think we should show the error even in form view. @divyanshiGupta  Thoughts?

karthikjeeyar:
In Edit Deployment flow, under pause rollouts it says `pause rollouts for this deployment config` . Instead of **deployment config** it should read **deployment** in case of deployment flow :thinking: . 
image

karthikjeeyar:
life cycle hook form validation is not working as expected when the same strategy type is chose again and enables the submit button.
eg:
In Edit deploymentConfig, choose strategy `Recreate` and click on `Add pre lifecycle hook` form and without filling any required fields, again scroll back up to the strategy type dropdown and choose `Recreate` again. This will enable the form submit even if the required fields in the pre lifecycle hook form is not filled.

divyanshiGupta:
> life cycle hook form validation is not working as expected when the same strategy type is chose again and enables the submit button.
> eg:
> In Edit deploymentConfig, choose strategy `Recreate` and click on `Add pre lifecycle hook` form and without filling any required fields, again scroll back up to the strategy type dropdown and choose `Recreate` again. This will enable the form submit even if the required fields in the pre lifecycle hook form is not filled.

@karthikjeeyar lifecyclehooks are optional. And when you change the strategy type form resets to its default values specific to that strategy type so even on submit no undesired values will be added

karthikjeeyar:
Verified the PR locally, works as expected
/lgtm

rohitkrai03 (approved):
/approve

Code looks good to me. Tested the PR locally and it works as expected.

rohitkrai03:
/label qe-approved

/label px-approved 

/label docs-approved 
added on behalf of docs team as the epic has doc-ack label.

PR #8725, opened 2021-04-21
Bug 1946232: VM Page search is slow with multiple rows

glekner:


glekner:
* Added debounce for the name/label query input. 200ms for the VM Page, default is 0 for all other resources.
* Moved some resource watching from `Fireman` to the `VMTable` component, so there won't be a refetch after every query.

Overall the experience is better when having 600+ rows. typing is fluid and waiting for results is faster.

frontend/public/components/filter-toolbar.tsx
```diff
@@ -420,6 +433,7 @@ type FilterToolbarProps = {
   labelFilterPlaceholder?: string;
   // Used when multiple tables are in the same page
   uniqueFilterName?: string;
+  debounceTimeout?: number;
```
spadgett, line 436: I'm not sure if we want to introduce a debounce here since the table will no longer react right away as you type. We used to be able to handle that well before we rewrote the filter toolbar component. I'd like to profile it to see if we can't fix wihtout debouncing. We have a separate Bugzilla tracking filter performance.

https://bugzilla.redhat.com/show_bug.cgi?id=1934229

Is the filter toolbar only slow for VMs, or is it all resources?

If we do debounce, we should consider making it the behavior for all toolbars instead of using a prop.

cc @TheRealJon

glekner: Its only slow for VMs AFAIK, because the VM's page is watching a lot of resources.
the default debounce is 0, so only the VM table will have a delay.

here is a profiling of a filter change on the pvc view (600 rows)
image

here is a profiling of a filter change on the vm view (600 rows)
image

Again, the VM view watches a lot of resources, so I believe this is the issue, its not about the `flatten`.
I also noticed the fetching chunks are 250 each, so because there are 600 rows, the computation is happening multiple times. IMO we should consider building a paginated GraphQL table from scratch, and I dont mind digging into it but I have plenty of extensions to write. thoughts? @spadgett @TheRealJon

spadgett: Filtering is a little slow for all resources... @TheRealJon has PR #8821 that adds debounce generally to the filter toolbar.

glekner:
/hold

waiting for https://github.com/openshift/console/pull/8821

yaacov:
#8821 merged

PR #8726, opened 2021-04-21
Bug 1949019: - PersistentVolumes page cannot sync project status automatically which will block user to create PV

cyril-ui-developer:
/assign @TheRealJon 
cc: @spadgett 

Since PV and Storage classes are cluster resources and don't need a project creation before creating a resource, I have removed the Start Guide hint from them. So the Create button will be enabled by default for PV.

Screen Shot 2021-04-21 at 9 54 25 AM
Screen Shot 2021-04-21 at 9 54 36 AM
Screen Shot 2021-04-21 at 9 54 50 AM
Screen Shot 2021-04-21 at 9 55 00 AM

### DIFF
--- a/frontend/public/components/start-guide.tsx
+++ b/frontend/public/components/start-guide.tsx
@@ -8,10 +8,8 @@ import { FLAGS } from '@console/shared/src/constants';
 import { createProjectMessageStateToProps } from '../reducers/ui';
 import { Disabled, HintBlock, ExternalLink, openshiftHelpBase, LinkifyExternal } from './utils';
 import { connectToFlags } from '../reducers/features';
-import { ProjectModel, RoleModel, StorageClassModel } from '../models';
+import { ProjectModel } from '../models';
 import { createProjectModal } from './modals/create-namespace-modal';
-
-const WHITELIST = [RoleModel.kind, StorageClassModel.kind];
 
 export const OpenShiftGettingStarted = connect(createProjectMessageStateToProps)(
   ({ canCreateProject = true, createProjectMessage }: OpenShiftGettingStartedProps) => (
@@ -70,9 +68,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
               <OpenShiftGettingStarted canCreateProject={flags[FLAGS.CAN_CREATE_PROJECT]} />
             </HintBlock>
           </div>
-          {// Whitelist certain resource kinds that should not be disabled when no projects are
-          // available. Disabling should also be optional
-          !disable || WHITELIST.includes(kind) ? (
+          {!disable || (rest.kindObj && !rest.kindObj.namespaced) ? (
             <WrappedComponent {...rest} noProjectsAvailable />
           ) : (
             <Disabled>


frontend/public/components/start-guide.tsx
```diff
@@ -8,10 +8,15 @@ import { FLAGS } from '@console/shared/src/constants';
 import { createProjectMessageStateToProps } from '../reducers/ui';
 import { Disabled, HintBlock, ExternalLink, openshiftHelpBase, LinkifyExternal } from './utils';
 import { connectToFlags } from '../reducers/features';
-import { ProjectModel, RoleModel, StorageClassModel } from '../models';
+import { ProjectModel, RoleModel, StorageClassModel, PersistentVolumeModel } from '../models';
 import { createProjectModal } from './modals/create-namespace-modal';
 
-const WHITELIST = [RoleModel.kind, StorageClassModel.kind];
+const WHITELIST = [RoleModel.kind, PersistentVolumeModel.kind];
+const CLUSTER_WIDE_WHITELIST = [
```
TheRealJon, line 15: Not new, but If we still need these globals in the final solution, we should probably rename them to use better language (remove "whitelist" terminology).

frontend/public/components/start-guide.tsx
```diff
@@ -58,7 +63,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
     const kind = _.get(kindObj, 'kind', rest.kind);
 
     // The start guide does not need to be shown on the Projects list page.
-    if (kind === ProjectModel.kind) {
+    if (CLUSTER_WIDE_WHITELIST.includes(kind)) {
```
TheRealJon, line 66: IMO, we still only need to hide the Getting Started message on the projects page. I think [this](https://github.com/openshift/console/blob/c8d960ac111224ccdf3570a0dcc6e56f1b3a5870/frontend/public/components/start-guide.tsx#L75) is where we need to adjust logic so that only namespaced resources get disabled.

```suggestion
    if (kind === ProjectModel.kind) {
```

TheRealJon (changes requested):
I wonder if we should keep the existing condition for showing the getting started message and just disable the page for namespaced resources. We should be able to use `kindObj` prop to check if the resource is namespaced instead of using a hard-coded list. 

We also still need to address the underlying issue where the SHOW_OPENSHIFT_START_GUIDE flag doesn't get updated when it should.

frontend/public/components/start-guide.tsx
```diff
@@ -72,7 +70,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
           </div>
           {// Whitelist certain resource kinds that should not be disabled when no projects are
           // available. Disabling should also be optional
-          !disable || WHITELIST.includes(kind) ? (
+          !disable || !kindObj.namespaced || kind === RoleModel.kind ? (
```
TheRealJon, line 73: It might make sense to include the RoleBindings as well here. We weren't before, but it seems like it's very similar to the Roles list page in that RBs and CRBs are listed together on the same page. An edge case might be possible where a user can create CRBs without having access to any namespaces. We also might need to look into any other possible cases like this where namespaced and non-namespaced resources are listed together on the same page. We probably should only disable pages that exclusively list namespaced resources. @spadgett any thoughts?

TheRealJon (commented):
See comments

frontend/public/components/start-guide.tsx
```diff
@@ -70,9 +68,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
               <OpenShiftGettingStarted canCreateProject={flags[FLAGS.CAN_CREATE_PROJECT]} />
             </HintBlock>
           </div>
-          {// Whitelist certain resource kinds that should not be disabled when no projects are
-          // available. Disabling should also be optional
-          !disable || WHITELIST.includes(kind) ? (
+          {!disable || !kindObj.namespaced || kind === RoleModel.kind ? (
```
TheRealJon, line 71: As per discussion with @spadgett, we should remove the special case for the Roles list page.
```suggestion
          {!disable || !kindObj.namespaced (
```

cyril-ui-developer:
@TheRealJon I have updated the code. Thanks.

cyril-ui-developer:
**Before:**

Screen Shot 2021-04-21 at 9 54 25 AM

**### After**

TheRealJon (approved):
/lgtm

frontend/public/components/start-guide.tsx
```diff
@@ -70,9 +68,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
               <OpenShiftGettingStarted canCreateProject={flags[FLAGS.CAN_CREATE_PROJECT]} />
             </HintBlock>
           </div>
-          {// Whitelist certain resource kinds that should not be disabled when no projects are
-          // available. Disabling should also be optional
-          !disable || WHITELIST.includes(kind) ? (
+          {!disable || !kindObj.namespaced ? (
```
spadgett, line 71: Do we know if `kindObj` is always defined? We use `_.get` above. I'm wondering if there is an edge case where it might be undefined while API discovery is still happening since we don't block console load on discovery.

cyril-ui-developer:
@spadgett @TheRealJon I checked for storage, workloads and a couple of other resources, the `kindObj` is defined.  So I'm not  sure of that.

spadgett:
> @spadgett @TheRealJon I checked for storage, workloads and a couple of other resources, the `kindObj` is defined. So I'm not sure of that.

You might have to check on a private window and hit the URL directly to see if there's a problem. If you navigate from another page, API discovery will probably be complete, and we cache the API resources in `localStorage`.

I'd probably add optional chaining here just to be safe.

cyril-ui-developer:
@spadgett Updated the code. Thanks.

frontend/public/components/start-guide.tsx
```diff
@@ -52,11 +51,10 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
     FLAGS.SHOW_OPENSHIFT_START_GUIDE,
     FLAGS.CAN_CREATE_PROJECT,
   )(({ flags, ...rest }: any) => {
-    const { kindObj } = rest;
-    const kind = _.get(kindObj, 'kind', rest.kind);
+    const { kindObj = {} } = rest;
```
spadgett, line 54: The logic here is different than before. Previously we were defaulting to `rest.kind` if `kindObj` was not set.

I also wouldn't default it to `{}`. That's not a valid `kindObj`. Better to leave it `const { kindObj } = rest;` and always use optional chaining.

cyril-ui-developer:
@spadgett Oh I see.  I wanted to avoid writing something like this: `{!disable || !rest?.kindObj?.namespaced ? (`.

cyril-ui-developer:
@spadgett  Updated the code. Thanks.

frontend/public/components/start-guide.tsx
```diff
@@ -70,9 +68,7 @@ export const withStartGuide = (WrappedComponent, disable: boolean = true) =>
               <OpenShiftGettingStarted canCreateProject={flags[FLAGS.CAN_CREATE_PROJECT]} />
             </HintBlock>
           </div>
-          {// Whitelist certain resource kinds that should not be disabled when no projects are
-          // available. Disabling should also be optional
-          !disable || WHITELIST.includes(kind) ? (
+          {!disable || !rest?.kindObj?.namespaced ? (
```
spadgett, line 71: We only want to test `!namespaced` if `kindObj` is defined. Looking above, I believe it's possible to pass a `kind` as a prop instead of a `kindObj`. So I'd change the logic to

```suggestion
          {!disable || (rest.kindObj && !rest.kindObj.namespaced) ? (
```

With the current change if `kindObj` was undefined, `!rest?.kindObj?.namespaced` would evaluate to `true`, not showing the disabled page.

cyril-ui-developer:
@spadgett Updated. Thanks.

spadgett (approved):
/lgtm

PR #8728, opened 2021-04-21
Monitoring dashboards: Show dashboard's tags in Dashboards dropdown

kyoto:
screenshot

FYI @cshinn


frontend/public/components/monitoring/dashboards/index.tsx
```diff
@@ -247,6 +247,65 @@ const AllVariableDropdowns = () => {
...
+    'red',
+    'purple',
+    'blue',
+    'green',
+    'cyan',
+    'orange',
+  ];
+
+  const allTags = _.flatMap(items, 'tags');
+  const uniqueTags = _.uniq(allTags);
```
spadgett, line 265: Is the order of tags in the array going to be the same every time? It might be odd if the tag colors are different each time you visit the page.

kyoto: Yes, it should be preserved because `_.uniq` preserves the order (see https://lodash.com/docs/4.17.15#uniq).

kyoto:
@sferich888 Could you add `px-approved`?

PR #8738, opened 2021-04-22
Bug 1952310: Added a create override options on useSecret hook

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1952310

**Analysis / Root cause**: 
Secret tried to update instead of creating

**Solution Description**: 
Added override options on the hook function


yaacov:
/hold
tests are failing, need to check why

yaacov:

PR #8740, opened 2021-04-22
Bug 1950219: fixes issue with KnativeServing not shown in list on global config page

invincibleJai:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1950219

**Analysis / Root cause**: 
KnativeServing is not shown in list on global config page as with dynamic plugins we don't models as `k8sKind` get only group, version, kind  which won't help for custom resources as underlying utils expects `apiVersion`, `apiGroup`  alongwith `kind` and optional types`plural`, `namespaced`, `crd` for custom resources.

 

**Solution Description**: 
- updated models type definition and model data
- updated flags

image

image

**Test setup:**
- Install Serverless Operator
- Create an instance of `knativeServing` in `knative-serving` namespace

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/console-extensions.json
+++ b/frontend/packages/knative-plugin/console-extensions.json
@@ -78,7 +78,7 @@
       "namespace": "knative-serving"
     },
     "flags": {
-      "required": ["FLAG_KNATIVE_SERVING"]
+      "required": ["KNATIVE_SERVING"]
     }
   }
 ]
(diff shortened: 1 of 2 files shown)


frontend/packages/console-dynamic-plugin-sdk/src/extensions/cluster-settings.ts
```diff
@@ -9,9 +9,12 @@ export type ClusterGlobalConfig = ExtensionDeclaration<
...
     /** The model which refers to a cluster config resource. */
     model: {
-      group: string;
-      version: string;
+      apiVersion: string;
+      apiGroup: string;
       kind: string;
+      plural?: string;
+      namespaced?: boolean;
+      crd?: boolean;
```
christianvogt, line 17: As discussed in slack, keep the current api but transform the model to appropriate data structure for the util.

invincibleJai: have updated it

PR #8750, opened 2021-04-22
Bug 1952625: Remove model and use Trans

rebeccaalpert:
Removed a deprecated model and updated a phrase to use the Trans component.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1952625.


spadgett (approved):
/lgtm

PR #8752, opened 2021-04-23
Pipeline Builder AutoComplete (Part 2 - Improved UX)

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-4277

Based off the foundation started in https://github.com/openshift/console/pull/8692.

**Analysis / Root cause**: 
Support missing items that were missed from #8692.

**Solution Description**: 

Known missing items from #8692:
- [x] Keyboard support to enter auto complete dropdown (currently only mouse or in some cases, tab)
- [x] Better support for inserting $( at the start of existing text
- [x] Potential support for handling $( in the middle of an existing item / error handling for incomplete $( references
- [x] String Array fields are not currently supported
- [x] Removing `$(` after the dropdown has started doesn't close it
- [ ] Tests

**Screen shots / Gifs for design review**: 
PipelineBuilder-AutoCompleteImproved

- Supported Keyboard Support: `Tab` key and `Down Arrow` will enter the auto complete dropdown. 
- Type Searches work on a startsWith concept for all the sections (split on `.`) as well as the full value. Potential here for us to do also every section within (see gif, `pipelineRun.name` does not match anything, but `context.pipelineRun.name` does).
- Empty option was not in the designs, WIP wording. cc @bgliwa01 

**Unit test coverage report**: 
Coming soon...

**Test setup:**

Easiest way to test is to create a nodejs app from the From Git add flow and add a Pipeline to it. Editing that created Pipeline gives you plenty of parameters as well as workspaces.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #8753, opened 2021-04-23
Add Trigger Param Auto Complete

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5686

**Analysis / Root cause**: 
With the added support of AutoComplete in the Pipeline Builder it is only fitting the Add Trigger modal gets the same treatment.

**Solution Description**: 
Add support for AutoComplete using the mechanism added in https://github.com/openshift/console/pull/8752.

(taken with https://github.com/openshift/console/pull/8752 implemented for ease of use of the keyboard)
AddTriggerAutoComplete

None

**Test setup:**
Have a Pipeline, use Add Trigger kebab. Once the modal is open, select a trigger to get access to more parameters, otherwise `uid` should be the only one currently.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
verified locally, works fine
/lgtm

PR #8754, opened 2021-04-23
Bug 1891696: [LSO] Add capacity UI does not check for node present in selected storageclass

SanjalKatiyar:
Added UI validations for add-capacity modal (for arbiter and normal deployment modes).

Screenshot from 2021-04-23 10-07-09

Screenshot from 2021-04-23 10-09-54

Screenshot from 2021-04-23 10-29-40

### DIFF
--- a/frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
+++ b/frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
@@ -1,10 +1,17 @@
-import { testEbsSC, testNoProvisionerSC } from '../mocks/storageclass';
+import {
+  testEbsSC,
+  testNoProvisionerSC,
+  testPersistentVolume1,
+  testPersistentVolume2,
+  testPersistentVolume3,
+} from '../mocks/storageclass';
 import { commonFlows } from '../views/common';
 import { K8sResourceKind } from '@console/internal/module/k8s';
 import { DeviceSet } from '../../src/types';
 import { OCS_INTERNAL_CR_NAME } from '../../src/constants';
 import { NS } from '../consts';
 import { getCurrentDeviceSetIndex } from '../../src/utils/add-capacity';
+import { getNodeRole } from '../../../console-shared/src/selectors/node';
 
 interface IndexAndDeviceSet {
   index: number;
@@ -28,6 +35,11 @@ const addCapacity = (uid: string, scName: string) => {
   cy.byTestID('dropdown-menu-item-link')
     .contains(scName)
     .click();
+
+  // wait for Pvs & Nodes data to load (used for validations)
+  // to-do (Sanjal): add a more robust way instead of wait by checking the currentCapacity/PVsAvailableCapacity components state
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(5000);
   cy.byTestID('confirm-action').click();
 };
 
@@ -44,12 +56,28 @@ describe('Add capacity using multiple storage classes', () => {
     cy.install();
     cy.exec(`echo '${JSON.stringify(testEbsSC)}' | kubectl apply -f -`);
     cy.exec(`echo '${JSON.stringify(testNoProvisionerSC)}' | kubectl apply -f -`);
+    cy.exec('oc get nodes -o json').then((res) => {
+      const nodes = JSON.parse(res.stdout);
+      const allWorkerNodes = nodes.items.filter((node) => getNodeRole(node) === 'worker');
+      testPersistentVolume1.spec.nodeAffinity.required.nodeSelectorTerms[0].matchExpressions[0].values[0] =
+        allWorkerNodes[0].metadata.name;
+      testPersistentVolume2.spec.nodeAffinity.required.nodeSelectorTerms[0].matchExpressions[0].values[0] =
+        allWorkerNodes[1].metadata.name;
+      testPersistentVolume3.spec.nodeAffinity.required.nodeSelectorTerms[0].matchExpressions[0].values[0] =
+        allWorkerNodes[2].metadata.name;
+      cy.exec(`echo '${JSON.stringify(testPersistentVolume1)}' | kubectl apply -f -`);
+      cy.exec(`echo '${JSON.stringify(testPersistentVolume2)}' | kubectl apply -f -`);
+      cy.exec(`echo '${JSON.stringify(testPersistentVolume3)}' | kubectl apply -f -`);
+    });
     commonFlows.navigateToOCS();
     cy.byLegacyTestID('horizontal-link-Storage Cluster').click();
   });
   after(() => {
     cy.exec(`echo '${JSON.stringify(testEbsSC)}' | kubectl delete -f -`);
     cy.exec(`echo '${JSON.stringify(testNoProvisionerSC)}' | kubectl delete -f -`);
+    cy.exec(`echo '${JSON.stringify(testPersistentVolume1)}' | kubectl delete -f -`);
+    cy.exec(`echo '${JSON.stringify(testPersistentVolume2)}' | kubectl delete -f -`);
+    cy.exec(`echo '${JSON.stringify(testPersistentVolume3)}' | kubectl delete -f -`);
     cy.logout();
   });
 
--- a/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
@@ -10,7 +10,12 @@ import {
   ModalBody,
 } from '@console/internal/components/factory';
 import { useK8sWatchResource } from '@console/internal/components/utils/k8s-watch-hook';
-import { K8sResourceKind, k8sPatch, StorageClassResourceKind } from '@console/internal/module/k8s';
+import {
+  K8sResourceKind,
+  k8sPatch,
+  StorageClassResourceKind,
+  NodeKind,
+} from '@console/internal/module/k8s';
 import { usePrometheusQueries } from '@console/shared/src/components/dashboard/utilization-card/prometheus-hook';
 import { getName, getRequestedPVCSize } from '@console/shared';
 import { FieldLevelHelp } from '@console/internal/components/utils/field-level-help';
@@ -31,10 +36,11 @@ import {
 } from '../../../constants';
 import { OCSStorageClassDropdown } from '../storage-class-dropdown';
 import { PVsAvailableCapacity } from '../../ocs-install/pvs-available-capacity';
-import { pvResource } from '../../../resources';
+import { pvResource, nodeResource } from '../../../resources';
 import { createDeviceSet, getDeviceSetCount } from '../../ocs-install/ocs-request-data';
 import { DeviceSet } from '../../../types';
 import './add-capacity-modal.scss';
+import { isArbiterSC, isValidTopology } from '../../../utils/install';
 import { checkArbiterCluster, checkFlexibleScaling } from '../../../utils/common';
 
 const queries = (() => Object.values(CAPACITY_INFO_QUERIES))();
@@ -48,7 +54,8 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const deviceSets: DeviceSet[] = ocsConfig?.spec.storageDeviceSets || [];
 
   const [values, loading, loadError] = usePrometheusQueries(queries, parser as any);
-
+  const [pvData, pvLoaded, pvLoadError] = useK8sWatchResource<K8sResourceKind[]>(pvResource);
+  const [nodesData] = useK8sWatchResource<NodeKind[]>(nodeResource);
   const [storageClass, setStorageClass] = React.useState<StorageClassResourceKind>(null);
   /* TBD(Afreen): Show installation storage class as preselected
                   Change state metadata
@@ -56,8 +63,6 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
 
   const [inProgress, setProgress] = React.useState(false);
   const [errorMessage, setError] = React.useState('');
-
-  const [pvData, pvLoaded, pvLoadError] = useK8sWatchResource<K8sResourceKind[]>(pvResource);
 
   const osdSizeWithUnit = getRequestedPVCSize(deviceSets[0].dataPVCTemplate);
   const osdSizeWithoutUnit: number = OSD_CAPACITY_SIZES[osdSizeWithUnit];
@@ -73,6 +78,22 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const usedCapacityMetric = values?.[1];
   const usedCapacity = humanizeBinaryBytes(usedCapacityMetric);
   const totalCapacity = humanizeBinaryBytes(totalCapacityMetric);
+
+  const validateSC = React.useCallback(() => {
+    if (!selectedSCName) return t('ceph-storage-plugin~No StorageClass selected');
+    if (!isNoProvionerSC || hasFlexibleScaling) return '';
+    if (isArbiterEnabled && !isArbiterSC(selectedSCName, pvData, nodesData)) {
+      return t(
+        'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different StorageClass or create a new LocalVolumeSet that matches the minimum node requirement.',
+      );
+    }
+    if (!isArbiterEnabled && !isValidTopology(selectedSCName, pvData, nodesData)) {
+      return t(
+        'ceph-storage-plugin~The StorageCluster requires a minimum of 3 nodes. Please choose a different StorageClass or create a new LocalVolumeSet that matches the minimum node requirement.',
+      );
+    }
+    return '';
+  }, [selectedSCName, t, isNoProvionerSC, isArbiterEnabled, hasFlexibleScaling, pvData, nodesData]);
 
   let currentCapacity: React.ReactNode;
   let availablePvsCount: number = 0;
@@ -133,8 +154,9 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
       patch.value = deviceSets[deviceSetIndex].count + deviceSetCount;
     }
 
-    if (!selectedSCName) {
-      setError(t('ceph-storage-plugin~No StorageClass selected'));
+    const validation: string = validateSC();
+    if (validation) {
+      setError(validation);
       setProgress(false);
     } else {
       k8sPatch(OCSServiceModel, ocsConfig, [patch])
--- a/frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
+++ b/frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
@@ -5,6 +5,8 @@ export const ZONE_LABELS = [
   'failure-domain.beta.kubernetes.io/zone', // deprecated
 ];
 
+export const RACK_LABEL = 'topology.rook.io/rack';
+
 export enum MODES {
   INTERNAL = 'Internal',
   ATTACHED_DEVICES = 'Internal - Attached Devices',
--- a/frontend/packages/ceph-storage-plugin/src/utils/install.ts
+++ b/frontend/packages/ceph-storage-plugin/src/utils/install.ts
@@ -18,7 +18,7 @@ import {
 } from '@console/local-storage-operator-plugin/src/constants';
 import { getNodeCPUCapacity, getNodeAllocatableMemory, getName } from '@console/shared';
 import { NodeModel } from '@console/internal/models';
-import { ocsTaint, NO_PROVISIONER, MINIMUM_NODES, ZONE_LABELS } from '../constants';
+import { ocsTaint, NO_PROVISIONER, MINIMUM_NODES, ZONE_LABELS, RACK_LABEL } from '../constants';
 import { getSCAvailablePVs } from '../selectors';
 
 export const hasNoTaints = (node: NodeKind) => {
@@ -58,6 +58,8 @@ export const filterSCWithoutNoProv = (sc: StorageClassResourceKind) =>
 export const getZone = (node: NodeKind) =>
   node.metadata.labels?.[ZONE_LABELS[0]] || node.metadata.labels?.[ZONE_LABELS[1]];
 
+export const getRack = (node: NodeKind) => node.metadata.labels?.[RACK_LABEL];
+
 export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
   const nodes = pvs.reduce((res, pv) => {
     const matchExpressions: MatchExpression[] =
@@ -73,6 +75,21 @@ export const getAssociatedNodes = (pvs: K8sResourceKind[]): string[] => {
   return Array.from(nodes);
 };
 
+export const getTopologyInfo = (nodes: NodeKind[]) =>
+  nodes.reduce(
+    (data, node) => {
+      const zone = getZone(node);
+      const rack = getRack(node);
+      if (zone && (hasOCSTaint(node) || hasNoTaints(node))) data.zones.add(zone);
+      if (rack && (hasOCSTaint(node) || hasNoTaints(node))) data.racks.add(rack);
+      return data;
+    },
+    {
+      zones: new Set<string>(),
+      racks: new Set<string>(),
+    },
+  );
+
 export const getNodeInfo = (nodes: NodeKind[]) =>
   nodes.reduce(
     (data, node) => {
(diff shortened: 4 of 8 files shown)


frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -41,6 +48,8 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const deviceSets: DeviceSet[] = ocsConfig?.spec.storageDeviceSets || [];
 
   const [response, loadError, loading] = usePrometheusPoll(cephCapacityResource);
+  const [pvData] = useK8sWatchResource<K8sResourceKind[]>(pvResource);
+  const [nodesData] = useK8sWatchResource<NodeKind[]>(nodeResource);
```
SanjalKatiyar, line 52: dataLoaded and dataError checks might not be required here... in case of such errors `pvData === [] & nodesData === []` and `isArbiterSC & isNormalSC` functions can handle that...
I do not think there will be any use case where dataLoaded and dataError fields are required here ??

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -60,6 +69,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+      if (isArbiterEnabled && !isArbiterSC(null, pvData, nodesData, selectedSCName)) {
+        setError(
+          t(
+            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement',
+          ),
+        );
+      } else if (
+        !hasFlexibleScaling &&
+        !isArbiterEnabled &&
+        !isNormalSC(null, pvData, nodesData, selectedSCName)
```
afreen23, line 83:
```suggestion
        !hasFlexibleScaling &&
        !isNormalSC(null, pvData, nodesData, selectedSCName)
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -60,6 +69,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+      ) {
+        setError(
+          t(
+            'ceph-storage-plugin~The OCS Storage cluster requires a minimum of 3 nodes. Please choose a different storage class or create a new volume set that matches the minimum node requirement',
```
afreen23, line 87:
```suggestion
            'ceph-storage-plugin~The OCS Storage cluster requires a minimum of 3 nodes. Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -60,6 +69,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const replica = isArbiterEnabled ? OCS_DEVICE_SET_ARBITER_REPLICA : OCS_DEVICE_SET_REPLICA;
   const name = getName(ocsConfig);
 
+  React.useEffect(() => {
+    if (selectedSCName) {
+      if (isArbiterEnabled && !isArbiterSC(null, pvData, nodesData, selectedSCName)) {
+        setError(
+          t(
+            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement',
+          ),
```
afreen23, line 78:
```suggestion
            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
          ),
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -60,6 +69,28 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const replica = isArbiterEnabled ? OCS_DEVICE_SET_ARBITER_REPLICA : OCS_DEVICE_SET_REPLICA;
   const name = getName(ocsConfig);
 
+  React.useEffect(() => {
```
afreen23, line 72: Its okay to show the error on clicking add capacity button for now. 
Since otherwise , we need to disable add capacity button and change the error message text as well.
For future I want to move to use FormSelect in this modal and run validations then 
https://www.patternfly.org/v4/components/form-select#invalid. Hence lets keep

afreen23:
/retitle Bug 1891696: [LSO] Add capacity UI does not check for node present in selected storageclass

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,12 +113,31 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+): boolean => {
+  const pvs: K8sResourceKind[] = getSCAvailablePVs(pvData, scName || getName(sc));
+  const scNodeNames = getAssociatedNodes(pvs);
+  const tableData: NodeKind[] = nodesData.filter(
+    (node: NodeKind) =>
+      scNodeNames.includes(getName(node)) ||
+      scNodeNames.includes(node.metadata.labels?.['kubernetes.io/hostname']),
+  );
+  const uniqSelectedNodesZones: Set<string> = new Set(tableData.map(getZone));
+  if (_.compact([...uniqSelectedNodesZones]).length < 3) return false;
```
afreen23, line 130: There is a helper function `getNodeInfo` I think we can use that instead of later using `_.compact` to remove values.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,12 +113,31 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+  return true;
+};
+
 export const isArbiterSC = (
   sc: StorageClassResourceKind,
   pvData: K8sResourceKind[],
   nodesData: NodeKind[],
+  scName?: string,
```
afreen23, line 138: Its better to update the rest of places to get `scName` instead of whole storageClass data.
I can see only name is getting used in this function

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,12 +113,31 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
 export const nodesWithoutTaints = (nodes: NodeKind[]) =>
   nodes.filter((node: NodeKind) => hasOCSTaint(node) || hasNoTaints(node));
 
+export const isNormalSC = (
+  sc: StorageClassResourceKind,
+  pvData: K8sResourceKind[],
+  nodesData: NodeKind[],
+  scName?: string,
```
afreen23, line 120: Keep only one - scName

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
SanjalKatiyar: this will cause a bug... if `isArbiterEnabled === true && !isArbiterSC === false` then control will fall to L81 and here `!hasFlexibleScaling === true && !isNormalSC === true` (beacuse current storage-class has nodes w.r.t to arbiter requirements, so it is not normal)... so it will start showing the incorrect error: `"The OCS Storage cluster requires a minimum of 3 node....."` even though my nodes/pvs configurations are correct w.r.t arbiter mode...

frontend/packages/ceph-storage-plugin/src/utils/install.ts
SanjalKatiyar: function `getNodeInfo` only considers nodes which `hasNoTaints`... so, if I have created a storage-cluster with say 3 nodes () and now I am adding-capacity with the same storage-class... it will show error and won't allow to expand...

afreen23: okay

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: I would like to see it made simpler - 
if (isArbiterEnabled) return isArbiterSC();
if (!isFlexibleScaling) return isInternalAttachedSC()

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,12 +113,31 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+): boolean => {
+  const pvs: K8sResourceKind[] = getSCAvailablePVs(pvData, scName || getName(sc));
+  const scNodeNames = getAssociatedNodes(pvs);
+  const tableData: NodeKind[] = nodesData.filter(
+    (node: NodeKind) =>
+      scNodeNames.includes(getName(node)) ||
+      scNodeNames.includes(node.metadata.labels?.['kubernetes.io/hostname']),
+  );
```
afreen23, line 128: This logic can be extracted out as one function. Its re-used in `isArbiterSc`

afreen23 (changes requested):

afreen23:
Some frontend tests are failing, please check

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -65,6 +65,26 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+    if (!selectedSCName) return t('ceph-storage-plugin~No StorageClass selected');
+    if (!isNoProvionerSC) return '';
+    if (isArbiterEnabled) {
+      return !isArbiterSC(selectedSCName, pvData, nodesData)
+        ? t(
+            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
+          )
+        : '';
+    }
+    if (!hasFlexibleScaling) {
```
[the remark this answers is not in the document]
SanjalKatiyar: I have already added condition for that on L70.

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,22 +113,40 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+  return tableData;
+};
+
+export const isIntenalAttachedSC = (
+  scName: string,
+  pvData: K8sResourceKind[],
+  nodesData: NodeKind[],
+): boolean => {
+  const tableData: NodeKind[] = getSelectedNodes(scName, pvData, nodesData);
+  return tableData.length >= MINIMUM_NODES;
```
afreen23, line 137: we need to add check for rack labels as well.
This is to make sure that the PVs are exactly from three different racks.
Sorry for the last minute change, but its final.
Thanks.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -74,6 +79,26 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
   const totalCapacity = humanizeBinaryBytes(totalCapacityMetric);
 
+  const validateSC = React.useCallback(() => {
+    if (!selectedSCName) return t('ceph-storage-plugin~No StorageClass selected');
+    if (!isNoProvionerSC) return '';
+    if (isArbiterEnabled) {
+      return !isArbiterSC(selectedSCName, pvData, nodesData)
+        ? t(
+            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
+          )
```
afreen23, line 89:
```suggestion
    if (isArbiterEnabled && !isArbiterSC(selectedSCName, pvData, nodesData) 
        return t(
            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
          )
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -74,6 +79,26 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+        : '';
+    }
+    if (!hasFlexibleScaling) {
+      return !isIntenalAttachedSC(selectedSCName, pvData, nodesData)
+        ? t(
+            'ceph-storage-plugin~The OCS Storage cluster requires a minimum of 3 nodes. Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
+          )
+        : '';
+    }
```
afreen23, line 98:
```suggestion
    if (!hasFlexibleScaling && !isIntenalAttachedSC(selectedSCName, pvData, nodesData)
        return  t(
            'ceph-storage-plugin~The OCS Storage cluster requires a minimum of 3 nodes. Please choose a different storage class or create a new volume set that matches the minimum node requirement.',)
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
SanjalKatiyar: will be buggy this way... if `isArbiterEnabled === true && !isArbiterSC === false` then control will move to L92 and here `!hasFlexibleScaling === true && !isIntenalAttachedSC === true` (beacuse current storage-class has nodes w.r.t to arbiter configuration)... so it will start showing the incorrect error: "The OCS Storage cluster requires a minimum of 3 node....." even though my nodes/pvs configurations are correct w.r.t arbiter mode...

frontend/packages/ceph-storage-plugin/src/constants/ocs-install.ts
```diff
@@ -5,6 +5,8 @@ export const ZONE_LABELS = [
   'failure-domain.beta.kubernetes.io/zone', // deprecated
 ];
 
+export const RACK_LABEL = ['topology.rook.io/rack'];
```
afreen23, line 8: There is no other kind of rack label.
```suggestion
export const RACK_LABEL = 'topology.rook.io/rack';
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23: We dont want it to be getting true for `isInternalAttachedSC`. Its only required to be true/false or checked when `failureDomain === 'rack' .

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -74,6 +80,35 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
...
+    if (!selectedSCName) return t('ceph-storage-plugin~No StorageClass selected');
+    if (!isNoProvionerSC) return '';
+    if (isArbiterEnabled) {
+      return !isArbiterSC(selectedSCName, pvData, nodesData)
+        ? t(
+            'ceph-storage-plugin~The Arbiter stretch cluster requires a minimum of 4 nodes (2 different zones, 2 nodes per zone). Please choose a different storage class or create a new volume set that matches the minimum node requirement.',
+          )
+        : '';
+    }
+    if (!hasFlexibleScaling) {
```
afreen23, line 93: This condition aligns with what we want to do -  only for rack based clusters , we want to check further whether added storage class has min 3 rack labels or not.
```suggestion
    if (failureDomain === 'rack'  && !isIntenalAttachedSC) {
```

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,22 +115,45 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+  nodesData: NodeKind[],
+  failureDomain: string,
+): boolean => {
+  const tableData: NodeKind[] = getSelectedNodes(scName, pvData, nodesData);
+  if (failureDomain === 'rack') {
+    const uniqSelectedNodesRacks: Set<string> = new Set(tableData.map(getRack));
+    return _.compact([...uniqSelectedNodesRacks]).length >= MINIMUM_NODES;
+  }
+  return tableData.length >= MINIMUM_NODES;
+};
```
afreen23, line 145: We will check only for rack labels here and that they 
```suggestion
): boolean => {
  const tableData: NodeKind[] = getSelectedNodes(scName, pvData, nodesData);
    const uniqSelectedNodesRacks: Set<string> = new Set(tableData.map(getRack));
    return _.compact([...uniqSelectedNodesRacks]).length >= MINIMUM_NODES;
  }

};
```

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,22 +115,45 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+export const isIntenalAttachedSC = (
+  scName: string,
+  pvData: K8sResourceKind[],
+  nodesData: NodeKind[],
+  failureDomain: string,
+): boolean => {
+  const tableData: NodeKind[] = getSelectedNodes(scName, pvData, nodesData);
+  if (failureDomain === 'rack') {
+    const uniqSelectedNodesRacks: Set<string> = new Set(tableData.map(getRack));
+    return _.compact([...uniqSelectedNodesRacks]).length >= MINIMUM_NODES;
```
afreen23, line 142: Suggestion: I would prefer to use `reduce` to marshalling/unmarshalling twice.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
SanjalKatiyar: I have few concerns related to this:

1) What if we are adding capacity with `hasFlexibleScaling === true` & `failureDomain === 'rack'` configuration. In this case L93 will be true and it will show wrong error message. So, I will add a check above somewhere that if `hasFlexibleScaling ==true` do not check for `isIntenalAttachedSC` or `isArbiterSC`.

2) "This condition aligns with what we want to do - only for rack based clusters" --> Do we not want to add `isIntenalAttachedSC` validation for AWS provider as well ?? 
Right now, `isIntenalAttachedSC` function handles both situation (Rack and AWS). For AWS provider we are checking if we have 3 nodes with PV on each or not, similar to what we do in 2nd step of wizard flow and if failureDomain is "rack" we check for total racks with PVs attached. So, all situations are handled right now.

afreen23: 1. is invalid, flexible scaling sets a host based failure domain
2.Again its only for LSO based deployments on AWS/Baremetal/Vsphere.
Let's call `isIntenalAttachedSC` - validateRacks or something else if thats confusing.

SanjalKatiyar: 1) thanks, yeah u r right, I missed this point.

2) "its only for LSO based deployments on **AWS**/Baremetal/Vsphere" --> that is what I wanted to confirm. For Baremetal/Vsphere it is fine, but for AWS we do not have rack labels, right ?? So, checking `failureDomain === 'rack'` for AWS infrastructure will be always false. So, we do not want to add `isIntenalAttachedSC` validation for AWS case ??

afreen23: For AWS as well we expect to check. 
The nodes should be in three different racks/topologies - its for all LSO based deployment.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -113,22 +127,55 @@ export const countNodesPerZone = (nodes: NodeKind[]) =>
...
+    const uniqSelectedNodesRacks: Set<string> = getRackCount(tableData).racks;
+    return uniqSelectedNodesRacks.size >= MINIMUM_NODES;
+  }
+
+  /** For AWS case
+   *  Failure Domain will not be "rack"
+   *  Checking if PVs are in 3 different zones or not
+   */
+  const uniqSelectedNodesZones: Set<string> = getNodeInfo(tableData).zones;
+  return uniqSelectedNodesZones.size >= MINIMUM_NODES;
```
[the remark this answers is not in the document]
SanjalKatiyar: okay... so I am adding `compact[...]` back if I am adding `tableData.map(node => getRack(node) || getZone(node))`... 
Or if you suggest to use reduce only, I will make a new function similar to `getNodesInfo` to count unique racks and zones at once ??

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/utils/install.ts
```diff
@@ -80,15 +82,18 @@ export const getNodeInfo = (nodes: NodeKind[]) =>
...
+      if (zone && (hasOCSTaint(node) || hasNoTaints(node))) data.zones.add(zone);
+      if (rack && (hasOCSTaint(node) || hasNoTaints(node))) data.racks.add(rack);
       return data;
     },
     {
       cpu: 0,
       memory: 0,
-      zones: new Set(),
+      zones: new Set<string>(),
+      racks: new Set<string>(),
```
afreen23, line 96: racks only required for add capacity , we should not add that to generic utility.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -44,12 +51,30 @@ describe('Add capacity using multiple storage classes', () => {
     cy.install();
     cy.exec(`echo '${JSON.stringify(testEbsSC)}' | kubectl apply -f -`);
     cy.exec(`echo '${JSON.stringify(testNoProvisionerSC)}' | kubectl apply -f -`);
+
+    cy.exec('oc get nodes -o json').then((res) => {
+      const nodes = JSON.parse(res.stdout);
+      const allWorkerNodesReady = nodes.items.filter((node) => getNodeRole(node) === 'worker');
```
afreen23, line 57:
```suggestion
      const allWorkerNodes = nodes.items.filter((node) => getNodeRole(node) === 'worker');
```

afreen23 (approved):
/lgtm

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
SanjalKatiyar: yeah... I copied it from other file and missed correcting it... thanks...

afreen23:
ceph-storage-plugin are failing @SanjalKatiyar please check

afreen23:
#8764 has merged which has updated some keys, please rebase your PR.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -28,6 +35,10 @@ const addCapacity = (uid: string, scName: string) => {
   cy.byTestID('dropdown-menu-item-link')
     .contains(scName)
     .click();
+
+  // wait for Pvs & Nodes data to load (used for validations)
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(5000);
```
afreen23, line 41: This can be a source of flakiness. We should try to avoid waits, since cypress by itself waits for components to be visible.

SanjalKatiyar: added a to-do task foe later...

afreen23 (approved):
/lgtm

PR #8756, opened 2021-04-23
CONSOLE-2828: Convert Table to React.FC

rawagner:
Converting Table to functional components, using hooks and improving types a little bit

### DIFF
--- a/frontend/packages/container-security/src/components/__tests__/ImageVulnerabilitiesTable.spec.tsx
+++ b/frontend/packages/container-security/src/components/__tests__/ImageVulnerabilitiesTable.spec.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
-import { Table } from '@console/internal/components/factory';
+import { Table, ComponentProps } from '@console/internal/components/factory';
 import { fakeVulnFor } from '../../../integration-tests/bad-pods';
 import { Priority } from '../../const';
 import ImageVulnerabilitiesTable from '../ImageVulnerabilitiesTable';
@@ -37,7 +37,7 @@ describe('ImageVulnerabilitiesTable', () => {
     const headers = wrapper
       .find(Table)
       .props()
-      .Header();
+      .Header({} as ComponentProps);
     expectedImageVulnerabilitiesTableHeader.forEach((header, i) => {
       expect(headers[i].title).toBe(header);
     });
--- a/frontend/public/components/factory/table.tsx
+++ b/frontend/public/components/factory/table.tsx
@@ -1,8 +1,25 @@
 import * as _ from 'lodash-es';
-import * as PropTypes from 'prop-types';
 import * as React from 'react';
-import { connect } from 'react-redux';
-import { withTranslation } from 'react-i18next';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
+import { useDispatch } from 'react-redux';
+import { match as RMatch } from 'react-router-dom';
+import {
+  Table as PfTable,
+  TableHeader,
+  TableBody,
+  TableGridBreakpoint,
+  SortByDirection,
+  OnSelect,
+  TableProps as PfTableProps,
+} from '@patternfly/react-table';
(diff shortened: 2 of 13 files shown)


frontend/public/components/factory/table.tsx
```diff
@@ -552,320 +386,295 @@ const getActiveColumns = (
...
+      data,
+      filters,
+      selected,
+      match,
+      kindObj,
+      activeColumns,
+      columnManagementID,
+      showNamespaceOverride,
+      t,
+    ],
```
spadgett, line 484: In these dependency lists, can you confirm none of these items are changing every render? I just want to make sure we're not passing anonymous functions or similar as props somewhere causing the body to reevaluate each time.

spadgett (commented):
Thank you! 👍 We should make sure we have a JIRA story tracking these changes. We can link it to https://issues.redhat.com/browse/CONSOLE-2512 if you need an epic.

for approvals

@ahardin-rh @sferich888 This only refactors code. There are no user visible changes

yapei:
@yanpzhan will provide test results back when finished verification testing

yanpzhan:
After run a round of regression test on lots of pages, check tables and filter from table, check column management, no regression issue is found.

frontend/public/components/factory/table.tsx
rawagner: good idea.
`Header` is problematic, many times passed as new function.
`filters` can also cause issues I think.
The other dependencies are fine.

There's the same problem in `table-data-hook` where all filter and sort props can cause unnecessary rerenders. 

I solved it by using `useDeepCompareMemoize` but we will want to fix the parent components so we dont have to use the deep compare in the future.

rawagner: FYI there is no need to use the deep memoize in new table proposal https://github.com/openshift/console/pull/8829

spadgett:
We have general px-approval for dynamic plugin SDK changes, so adding the label

/label px-approved
/unassign @sferich888

frontend/packages/container-security/src/components/__tests__/ImageVulnerabilitiesTable.spec.tsx
```diff
@@ -37,7 +37,7 @@ describe('ImageVulnerabilitiesTable', () => {
     const headers = wrapper
       .find(Table)
       .props()
-      .Header();
+      .Header({} as ComponentProps, (str) => str);
```
vojtechszocs, line 40: Why do we need `as` type cast here? Aren't all properties in `ComponentProps` optional, so `{}` should be assignable to `ComponentProps` type? :confused:

vojtechszocs (changes requested):

frontend/packages/container-security/src/components/__tests__/ImageVulnerabilitiesTable.spec.tsx
rawagner: no, all props of `ComponentProps` are required, but for this test, they are not needed

frontend/public/components/factory/table.tsx
```diff
@@ -460,73 +335,31 @@ export type RowFunctionArgs<T = any, C = any> = {
...
-};
-
-type ComponentProps = {
-  data?: any[];
-  filters?: Object;
-  selected?: any;
-  match?: any;
-  kindObj?: K8sResourceKindReference;
-};
+type HeaderFunc = (componentProps: ComponentProps, t: TFunction) => any[];
```
spadgett, line 351: We shouldn't pass `t` as an argument. It's better to simply use `i18next.t` inside the function. (Not all `t`s are the same since `useTranslation` has params that control things like the default namespace.)

This applies here and a few other places it looks like.

frontend/packages/container-security/src/components/__tests__/ImageVulnerabilitiesTable.spec.tsx
vojtechszocs: OK, thanks for clarification.

frontend/public/components/factory/table.tsx
rawagner: ok, thank you @spadgett . I removed every occurence of passing `t`

spadgett (approved):
/lgtm

spadgett (approved):
/lgtm

PR #8758, opened 2021-04-23
Avoid setting JAVA_ARGS if cmd args are not provided

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5781

**Analysis / Root cause**: 
The `JAVA_ARGS` env var was being set even when cmd args were not provided.

**Solution Description**: 
Only set `JAVA_ARGS` if cmd args are given.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug


sahil143 (approved):
/lgtm

verfied changes locally

invincibleJai:
/approve

PR #8759, opened 2021-04-23
fixes issue with ksvc edit for upload jar flow

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5782

**Analysis / Root cause**: 
Edit flow on upload jar flow for KSVC doesn't create active revision if jar is not modified as new ImageStream is created on all edit, should happen only if jar gets updated.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


rottencandy (approved):
/lgtm

PR #8760, opened 2021-04-23
Bug 1952635: Revert "create app-content div with modal-container as s…

rhamilto:
…ibling"

This reverts commit c15cdaed0375db95f1be0a05b27ba36dad7a15cd.

https://bugzilla.redhat.com/show_bug.cgi?id=1926289 needs to be reopened since this PR reverts its fix, but Bug 1952635 is a much more severe bug, so it needs an immediate fix.


rhamilto:
Closing as @christianvogt has agreed to determine a proper fix without reverting.

PR #8761, opened 2021-04-23
Bug 1952958: Gracefully degrade when `Intl.RelativeTimeFormat` unavailable

spadgett:
`Intl.RelativeTimeFormat` is not supported by Safari 13 and some older browsers. Fall back to normal date/time display for those browsers instead of relative time.

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #8763, opened 2021-04-23
CONSOLE-2384: Add metrics tab to select resources

TheRealJon:
Add new metrics tab to Deployment, StatefulSet, and DaemonSet details pages and move Build, Pod, and Route metrics to a separate tab for consistency.

### Deployment metrics
deployment-metrics

### Build metrics
build metrics

### Pod metrics
pod metrics

### Route metrics
route metrics


spadgett:
cc @openshift/openshift-team-monitoring

frontend/packages/console-shared/src/components/dashboard/dashboard-card/card.scss
```diff
@@ -8,6 +8,13 @@ $co-dashboard-text-font-size: 0.875rem;
   margin: 0;
 }
 
+.co-dashboard-card--resource-metrics .query-browser__wrapper {
```
spadgett, line 11: Should this be `co-dashboard-card__resource-metrics`? It feels like an element rather than a modifier.

http://getbem.com/naming/

It would be better if we can add a specific class name to the wrapper element here. (I'm not sure if we are able to pass a `className` through to the nested component).

frontend/packages/console-shared/src/promql/resource-metrics.ts
```diff
@@ -0,0 +1,152 @@
...
+  [ResourceMetricsQuery.RESOURCE_NETWORK_UTILIZATION_OUT]: [
+    podControllerUtilizationQueries[ResourceMetricsQuery.RESOURCE_NETWORK_UTILIZATION_OUT]({
+      name,
+      namespace,
+      type,
+    }),
+  ],
+});
+
+export const getResourceUtilizationQueries = (obj: K8sResourceKind, model: K8sKind) => {
```
spadgett, line 148: nit: We could determine the type from `obj.apiVersion` and `obj.kind` if we wanted to remove one of the function arguments.

spadgett (commented):
Nice work! No major comments. We can follow up with a separate story for other workload types.

for approvals

spadgett:
We should probably move pod, build, and route metrics under a tab as part of this change so we're consistent.

TheRealJon:
@spadgett I've updated the PR to move Build, Pod, and Route metrics into new tabs as well

yapei:
@yanpzhan is testing the PR and will report test results back

yanpzhan:
@TheRealJon  Tested against the pr, and drafted a case which was pasted in https://issues.redhat.com/browse/CONSOLE-2384.
For now the metrics tab page work well, will approve the pr if there is no other checkpoint need to be tested.

spadgett (approved):
/lgtm
/hold for approvals

spadgett:
> @TheRealJon Tested against the pr, and drafted a case which was pasted in https://issues.redhat.com/browse/CONSOLE-2384.
> For now the metrics tab page work well, will approve the pr if there is no other checkpoint need to be tested.

Thanks, @yanpzhan. I've approved the test cases. Based on this comment, I'm adding the qe-approved label.

/label qe-approved

PR #8766, opened 2021-04-25
Bug 1952372: VM stop action should not be there if the VM is not running

glekner:


yaacov:

yaacov:
/hold
tests failing, holding until tests fixed

PR #8767, opened 2021-04-25
Bug 1952367: No VM status on overview page when VM is pending

glekner:
`HealthState.LOADING` seems to be bugged, replaced `Pending` to be `HealthState.Loading`

Screen Shot 2021-04-25 at 16 53 14


yaacov:

PR #8769, opened 2021-04-26
Bug 1953418: fix wizard validate tests

gouyang:
The default disk type of RHEL6 is not `sata` any more, remove this tests.


gouyang:
@yaacov PTAL

PR #8770, opened 2021-04-26
fix: Update token length

wtrocki:
When during development we updated token length based on the current offline token length provided by our current keycloak. This token has a lot of extra fields and data.
For GA we going to have different setup and token length will not be guaranteed to be longer than 500 characters (mostly due less data encoded into the token).
To not risk plugin being nonfunctional with the new architecture, we need to lower significantly token validation


wtrocki:
CC @christianvogt @invincibleJai - minor bugfix (or future disaster prevention). Should be easy to merge as it doesn't affect anything right now (we lovered token length validation

wtrocki:
FYI @christiemolloy

christianvogt:
@wtrocki we should have a jira / bugzilla attached to this

wtrocki:
@christianvogt Sorry for that. Relevant jira issue assigned to me: https://issues.redhat.com/browse/ODC-5785

christiemolloy (approved):
Looks good

PR #8771, opened 2021-04-26
Bug 1952732: Users did not understand the boot source labels

glekner:


yaacov:
/hold

@glekner can you check the failing tests ?

PR #8776, opened 2021-04-26
Bug 1952759: It was not immediately clear what the Star icon meant

glekner:
* Created a new column for pinned status
* When a user hovers on a row that is not pinned, a greyed star is shown, once he hovers on the star he can click to favorite
* Creating a tooltip for each rows caused issues, with multiple tooltips remaining opened when hovering on the rows, so I created the tooltip on the column title, watch the video below

@matthewcarleton @yaacov please review the new flow

https://user-images.githubusercontent.com/24938324/116110732-a0667b80-a6be-11eb-902d-df099d573d18.mov

### DIFF
--- a/frontend/packages/kubevirt-plugin/locales/en/kubevirt-plugin.json
+++ b/frontend/packages/kubevirt-plugin/locales/en/kubevirt-plugin.json
@@ -696,6 +696,8 @@
   "Launch console": "Launch console",
   "Preparing for customization": "Preparing for customization",
   "Boot source is being prepared for customization.": "Boot source is being prepared for customization.",
+  "Favorite templates are pinned to the top of the list": "Favorite templates are pinned to the top of the list",
+  "Template Provider": "Template Provider",
   "Any template with an available source will show in the developer catalog.": "Any template with an available source will show in the developer catalog.",
   "Virtual Machine Templates": "Virtual Machine Templates",
   "Virtualization": "Virtualization",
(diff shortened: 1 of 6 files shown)


yaacov:
WOW amazing, @matthewcarletonI PTAL, I liked the star better when they where after the name, Matt WDYT ?

yaacov:
@matthewcarleton ^^ I misspelled you name in the  last comment PTAL

yaacov:
@glekner in the actions dropdown we have a gray star it has a lighter gray color, can you use the same gray star ?

matthewcarleton:
Oh ya this is way better! Can we have a light grey star always present and turn it yellow with they favourite it? 
I like the placement on the left and this aligns with the PF design too.

glekner:
Screen Shot 2021-04-27 at 13 57 38

updated @yaacov @matthewcarleton

yaacov:
@glekner hi, pressing the top star in the title row does not do anything, was it suppose to clear all / select all ?

glekner:
> @glekner hi, pressing the top star in the title row does not do anything, was it suppose to clear all / select all ?

no, just give the information tooltip.

yaacov:
when customizing an image we get the new template (that is not created yet) without the star:
screenshot-localhost_9000-2021 04 27-16_38_51

yaacov:
@matthewcarleton WDYT about the star as a column header ?

matthewcarleton:
I think we can drop it from the header since it's not interactive

PR #8777, opened 2021-04-26
Bug 1952111: Optimized imports from @patternfly/react-tokens

dtaylor113:
In many files we are importing from @patternfly/react-tokens using a format like:

  `import { something } from '@patternfly/react-tokens' ` 

Whenever we do that, webpack has to parse 12k different token files. This caused a build failure in recent PF latest smoketesting.  It also increases build time.  

It's much better to import a specific file, like: 

  `import something from '@patternfly/react-tokens/dist/esm/something'`

Ex. 

Replace:

  `import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens';`

With

  `import { global_warning_color_100 as warningColor } from '@patternfly/react-tokens/dist/esm/global_warning_color_100'`


spadgett (approved):
/lgtm

PR #8779, opened 2021-04-26
Add when expression in pipeline builder

karthikjeeyar:
**Fixes**: 
https://user-images.githubusercontent.com/9964343/116135381-b55b0300-a6ee-11eb-93eb-a8e4b769776a.mp4

https://issues.redhat.com/browse/ODC-4278

**Analysis / Root cause**: 

As a user, I would like to be able to configure my when statements in the Pipeline Builder as I am crafting my Pipeline.

**Solution Description**: 

Added support for when expression in pipeline builder.

**When Expression:**
image
**Invalid When expression:**

invalid-error

whenexpression-pipeline-builder

**validation-utils.spec.ts**
```
Validate When Expresssions
        ✓ should fail if the when expression is missing input value (7ms)
        ✓ should fail if the when expression is missing operator value (8ms)
        ✓ should fail if the when expression is missing values (12ms)
        ✓ should fail if the when expression has partial missing values (8ms)
        ✓ should pass if the when expression is valid (5ms)
```
**utils-spec.ts**
```
 getFinallyTaskWidth
    ✓ expect to return larger width if any nodes present in finally section
    ✓ expect to return smaller width if nodes are not present in finally section
  taskHasWhenExpression
    ✓ expect to return false if the task does not contain when expressions
    ✓ expect to return true if the task contains when expressions
  nodesHasWhenExpression
    ✓ expect to return false if the nodes does not contain when expressions
    ✓ expect to return true if the node contains when expressions (1ms)
  hasWhenExpression
    ✓ expect to return false if the pipeline does not contain when expression
    ✓ expect to return true if the regular tasks in the pipeline contains when expression
    ✓ expect to return true if the finally tasks in the pipeline contains when expression
```

**Test setup:**

Install pipelines and visit pipeline builder page and click on a task to add when expression
**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind feature


bgliwa01 (approved):
lgtm!!

karthikjeeyar:
As discussed with Brigid on slack, added finally task light background color and updated the screenshots

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/WhenExpressionForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+          name={`${namePrefix}.operator`}
+          label={t('pipelines-plugin~Operator')}
+          items={WhenExpressionOperatorType}
+          title={t('pipelines-plugin~Select Operator')}
+          onChange={(operator: string) => setFieldValue(`${namePrefix}.operator`, operator)}
+          fullWidth
+        />
+      </GridItem>
+      <GridItem>
+        <TextColumnField
```
andrewballantyne, line 51: FWIW, I need to do this for my PR https://github.com/openshift/console/pull/8752

andrewballantyne (changes requested):

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnField.tsx
```diff
@@ -26,6 +26,10 @@ const MultiColumnField: React.FC<MultiColumnFieldProps> = ({
   toolTip,
   spans,
   complexFields,
+  isInlineControl,
+  rowClassName,
+  rowControlClassName,
+  rowControlLabel,
```
andrewballantyne, line 32: (not so short) Summary of a conversation I had with Karthik over slack.

These changes raised a flag for me. 4 new props that became pass-thru props to the Row component had me wondering what was going on.

After digging through the use-case he was doing it for, it started to concern me around the separation of concerns. Inherently, seeing `...ClassName` props is not bad. Sometimes we do want to configure the slight styles of the internal components of these multi-faceted components. Especially so when they are pretty heavy like this one is.

However, when we get into the goals of this change, it was more-or-less unnecessary if we approached it from a custom-use-case point of view. The main issue here is the location and rendering visibility of the remove button. Where we want to maintain and control it via a more tailored experience to this particular PR's use-case. So we can render-prop this and save us from getting into the weeds and trying to make this component bend customly to our particular use-case.

Providing renderProps with default fallbacks is quite useful for existing components and main-use-case components. If you have a component, like this one, that has a primary goal doing a singular goal. If you start getting into props and flags that create a much bigger "flow diagram" to cover its entire set of use-cases, you end up in a maintenance nightmare where the cost of updating a component is several hours because of the use-cases that need to be tested and handled. Naturally tests help with this, but if you take a test failure as a means to investigate what happened, that in itself can prevent you from creating bugs but not reducing the time you spend chasing the change/test failure.

This is all to say, components should aim to do a singular thing. Sometimes that singular thing is "do a complex thing", which makes it very subjective how to address changes. Always aiming to keep your special use-cases outside of a component and just create "escape hatches" (ie. render props) usually serves to make things easier. Naturally, that has a limit too... Too many escape hatches is an indication of component breakdown being needed... but that's a different discussion :)

andrewballantyne:
With the conflict now and the various stuff from my merge, you should be able to complete the last of it. Let me know if you have any questions / concerns.

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnField.tsx
karthikjeeyar: Added a render prop `rowRenderer` to customize the look and feel from the consumer component.

frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnFieldRow.tsx
```diff
@@ -14,29 +14,33 @@ import {
...
-  rowIndex: number;
-  children: React.ReactNode;
-  onDelete: () => void;
+export interface RowRendererProps {
+  fieldName: string;
   isReadOnly?: boolean;
   disableDeleteRow?: boolean;
   spans: gridItemSpanValueShape[];
   complexFields?: boolean[];
+  isInlineControl?: boolean;
```
andrewballantyne, line 23: Unused Prop

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWhenExpression.scss
```diff
@@ -0,0 +1,23 @@
...
+  &__section {
+    margin-bottom: var(--pf-global--spacer--md);
+    border: 1px dashed var(--pf-global--BorderColor--100);
+    padding: var(--pf-global--spacer--md);
+  }
+  &__control_button {
+    &_wrapper {
+      display: flex;
+      justify-content: flex-end;
+    }
```
andrewballantyne, line 13: Small nit, try not to break elements in half with `&` it makes it harder to look them up.

Recommend just retyping the shared wording as it's the "element" name in BEM.


Double small nit, I think BEM wants `block-with-dashes__element-with-dashes`

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWhenExpression.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {field.value?.length > 0 ? (
+          <Trans ns="pipelines-plugin">
+            use this format while referencing the variables in this form: <code>$(</code>
+          </Trans>
+        ) : (
+          t('pipelines-plugin~No When expressions are associated with this task')
+        )}
+      </p>
+      <MultiColumnField
+        name={`${name}`}
```
andrewballantyne, line 40:
```suggestion
        name={name}
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWhenExpression.tsx
```diff
@@ -0,0 +1,70 @@
...
+        addLabel={t('pipelines-plugin~Add When Expression')}
+        headers={[]}
+        emptyValues={{ input: '', operator: '', values: [''] }}
+        rowRenderer={({ onDelete, fieldName }: RowRendererProps) => (
+          <div className="opp-task-sidebar-when-expression__section">
+            <WhenExpressionForm autoCompleteValues={autoCompleteValues} namePrefix={fieldName} />
+            <div className={'opp-task-sidebar-when-expression__control_button_wrapper'}>
```
andrewballantyne, line 47:
```suggestion
            <div className="opp-task-sidebar-when-expression__control_button_wrapper">
```

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/WhenExpressionForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const WhenExpressionForm: React.FC<WhenExpressionFormProps> = ({
+  namePrefix,
+  autoCompleteValues,
+}) => {
+  const { setFieldValue } = useFormikContext<PipelineBuilderFormikValues>();
+  const { t } = useTranslation();
+
+  return (
+    <Grid hasGutter>
```
andrewballantyne, line 23: Hmmm, think this is a weird use of grid... but it works... so maybe it's not an issue. Would think a Flex split and a div (or nothing) for the `TextColumnField`.

Not the end of the world to leave it this way.

andrewballantyne (changes requested):
Slew of nits nothing too crazy.

Only real concern is what is becoming of MultiColumnField. The rowRenderer allows a lot of flexibility but complexFields loses ground a lot here. Also, at the end of the day you don't need empty state, headers, or most of the other props to do your own rendering. So it may boil down to just using the FormikArray component customly instead. There may be some work we can do here, but I don't want to push for it so late. I think this works (let me do some more UI testing to vet it) so we can go with this for 4.8 / Feature Freeze and aim to clean it up / split it into multiple components post this. Please log a Task Jira ticket for that work @karthikjeeyar

andrewballantyne (approved):
Approving because it works as I would think it should. Definitely log that ticket though. I think we are having too much overlap in MultiColumnField usages.

karthikjeeyar:
Created a ticket to refactor the MulticolumnField overlap  https://issues.redhat.com/browse/ODC-5800

frontend/packages/console-shared/src/components/formik-fields/text-column-field/TextColumnField.tsx
```diff
@@ -21,7 +20,7 @@ const TextColumnField: React.FC<TextColumnFieldProps> = (props) => {
       name={name}
       render={(arrayHelpers) => (
         <>
-          <DndProvider backend={HTML5Backend}>
+          <DragAndDrop>
```
karthikjeeyar, line 23: @invincibleJai Request you to take a look at this DnD implementation change.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-topology/WhenExpressionForm.tsx
karthikjeeyar: Will refactor this while creating the custom field array component for when expression form in ODC-5800.

frontend/packages/console-shared/src/components/formik-fields/text-column-field/TextColumnField.tsx
karthikjeeyar: Fixed the `error: cannot-have-two-html5-backends-at-the-same-time` issue, which was observed in other areas like git Import -> healthChecks and containerSourceForm. Now iI am using the useRef to save the HTML5Backend to reuse the backend between the renders to avoid the issue.

invincibleJai:
/lgtm

PR #8787, opened 2021-04-26
Allow focus on all alerts in modals

jcaianirh:
This came up as a screen reader user typed in an upper case char during project creation displaying an error alert in the create project modal. 
The alert does not allow focus, so that was added by this pr as well as info and success alerts.

This needs more discussion with @jessiehuff before merging.

Before:
no-focus

After:
focus

### DIFF
--- a/frontend/public/components/utils/button-bar.jsx
+++ b/frontend/public/components/utils/button-bar.jsx
@@ -25,16 +25,17 @@ const ErrorMessage = ({ message }) => {
       className="co-alert co-alert--scrollable"
       variant="danger"
       title={t('public~An error occurred')}
+      tabIndex={0}
     >
       <div className="co-pre-line">{message}</div>
     </Alert>
   );
 };
 const InfoMessage = ({ message }) => (
-  <Alert isInline className="co-alert" variant="info" title={message} />
+  <Alert isInline className="co-alert" variant="info" title={message} tabIndex={0} />
 );
 const SuccessMessage = ({ message }) => (
-  <Alert isInline className="co-alert" variant="success" title={message} />
+  <Alert isInline className="co-alert" variant="success" title={message} tabIndex={0} />
 );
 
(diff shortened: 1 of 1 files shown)


jcaianirh:
/hold for discussion on:
1.  is this the right fix, and if so
2. should this be a default pf behavior that devs can opt out of

jessiehuff:
Hey @jcaianirh! I've been thinking about this one a lot. I’ve been trying to think of ways to communicate the a11y necessary while also making the least amount of work for consumers possible. Basically consumers need a live region container where the alerts get rendered that notifies users of these alerts and this container has to exist on page load. That makes it difficult to make the alert by itself accessible because unless we significantly change the alert and keep users from conditionally rendering alerts (so that the container can exist on the page before updates), which I’d guess that most consumers using alert are currently doing, we can’t make the alert by itself accessible in all screen readers.

My attempt at only changing [Alert](https://github.com/patternfly/patternfly-react/pull/5773) became far more complicated than expected. By trying to make the alert itself accessible in all screen readers, I found that I was having to almost completely change all of the logic and put us at higher risk of unexpected errors. I can communicate how to do the container approach instead in the documentation, but that also leaves a lot of room for error that consumers would have to read it and understand the details of creating that container.

With this newer approach, I've[ updated the AlertGroup](https://github.com/patternfly/patternfly-react/pull/5946) and the suggestion to put all dynamic alerts inside of AlertGroup. AlertGroup can be there on page load and consumers can continue to conditionally render Alerts as most are likely currently doing. AlertGroup will handle all of the a11y considerations, and based on my testing, works in all three of the main screen readers (JAWS, NVDA, and VO). I'm also currently working on [updating the documentation](https://github.com/patternfly/patternfly-org/pull/2572) so that this will be as clear as possible. There will be a new "Accessibility" tab to communicate exactly what products need to do with this more complicated component and continue handling the heavy lifting with PF. :)

PR #8751, opened 2021-04-22
[WIP] update devfile form for preselected devfile

debsmita1:

PR #8794, opened 2021-04-27
Add devfile samples to samples catalog

rottencandy:
**Story**: 
https://issues.redhat.com/browse/ODC-5727

**Solution Description**: 
Create a catalog item provider for devfile samples using the devfile sample registry API.

Based on top of https://github.com/openshift/console/pull/8687 to use API.
Based on top of https://github.com/openshift/console/pull/8751 to use updated devfile form.

unchanged

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/import/ImportPage.tsx
```diff
@@ -88,6 +88,12 @@ const ImportPage: React.FunctionComponent<ImportPageProps> = ({ match, location
   } else if (importType === ImportTypes.devfile) {
     importData = ImportFlows(t).devfile;
     resources = [
+      {
+        kind: ImageStreamModel.kind,
+        prop: 'imageStreams',
+        isList: true,
+        namespace: 'openshift',
+      },
```
rohitkrai03, line 96: Why do you need builder images for devfile flow?

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -22,9 +27,36 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
   setFieldValue,
 }) => {
   const { t } = useTranslation();
+  const searchParams = new URLSearchParams(window.location.search);
+  const gitRepoUrl = searchParams.get('gitRepo');
+  const devfileName = searchParams.get('devfileName') || detectGitRepoName(gitRepoUrl);
+  const formType = searchParams.get('formType');
+  const language = searchParams.get('language');
+  const selectedBuilderImage = builderImages[language];
```
rohitkrai03, line 35: I think there's been some confusion here. Shouldn't we show selected devfile instead of selectedBuilderImage here?

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -33,11 +65,31 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+              name="git.url"
+              label={t('devconsole~Git repo URL')}
+              data-test-id="git-form-input-url"
+              isDisabled
+            />
+            {selectedBuilderImage && (
+              <ImageStreamInfo
+                displayName={selectedBuilderImage.name}
+                tag={selectedBuilderImage.tags[0]}
+              />
```
rohitkrai03, line 81: This info should be about selected devfile. The API returning sample devfile can be called here to get all devfiles and then select and show information using the metadata returned by the API.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -33,11 +65,31 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+            )}
+          </FormSection>
+        ) : (
+          <GitSection
```
rohitkrai03, line 85: There's one more flow that this form should handle. The dev catalog flow where user comes from dev catalog, in that case then git URL will be filled but the inout field will not be disabled.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Sorry I was confused with builder image samples, updated to use the selected devfile sample.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Updated.

frontend/packages/dev-console/src/components/catalog/providers/useDevfileSamples.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const href = `/import?importType=devfile&formType=sample&devfileName=${uid}&gitRepo=${gitRepoUrl}`;
+
+    const item: CatalogItem = {
+      uid,
+      type: 'Sample',
+      name: displayName,
+      description,
+      tags,
+      cta: {
+        label: '',
```
christianvogt, line 22: This will require a label to work with @invincibleJai 's change to include samples in quick search.

rottencandy: Added label.

frontend/packages/dev-console/src/components/catalog/providers/useDevfileSamples.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [devfileSamples, setDevfileSamples] = React.useState<DevfileSample[]>([]);
+  const [loadedError, setLoadedError] = React.useState<APIError>();
+
+  React.useEffect(() => {
+    const payload = {
+      registry: 'pkg/devfile/sample-placeholder.json',
+    };
+    coFetchJSON
+      .put('/api/devfile/samples', payload)
+      .then(setDevfileSamples)
```
rohitkrai03, line 48: Check if the component is still mounted before setting the state.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -22,9 +28,49 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+      }
+    }
+  }, [gitRepoUrl, devfileName, formType, setFieldValue]);
+
+  React.useEffect(() => {
+    if (devfileName) {
+      const payload = {
+        registry: 'pkg/devfile/sample-placeholder.json',
+      };
+      coFetchJSON
```
rohitkrai03, line 65: This would fetch all the devfile samples every time `devfileName` changes. The list shouldn't change that often so Ideally it should only fetch the list once and search through it.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -22,9 +28,49 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+        .put('/api/devfile/samples', payload)
+        .then((samples: DevfileSample[]) =>
+          setSelectedSample(samples.find((sample) => sample.name === devfileName)),
```
rohitkrai03, line 68: Check to see if the component is still mounted.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Updated.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Removed effect dependencies.

rohitkrai03:
@rottencandy We might need to tweak our devfile logic for catalog flow. In that flow we want to use the sample devfile but user can update git repo with their own which might not have a devfile, in that case we would want to use devfile contents from the sample repo. Let's talk about this more on Monday.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -23,8 +29,59 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+        .catch(() => {
+          setFieldValue('devfile.devfileContent', null);
+          setFieldValue('devfile.devfileHasError', true);
+        });
+
+      if (formType === 'sample') {
+        setFieldValue('name', createComponentName(devfileName));
+        setFieldValue('git.url', gitRepoUrl);
+        setFieldValue('git.type', gitType);
+        const defaultAppGroup = 'sample-app';
```
rohitkrai03, line 60: It would be good to keep this in a const file because the same is being used in other sample form as well.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -23,8 +29,59 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+  const devfileName = searchParams.get('devfileName');
+  const formType = searchParams.get('formType');
+  const [devfileSamples, setDevfileSamples] = React.useState<DevfileSample[]>();
+  const selectedSample = React.useMemo(
+    () => devfileSamples?.find((sample) => sample.name === devfileName),
+    [devfileSamples, devfileName],
+  );
   useDevfileDirectoryWatcher(values, setFieldValue);
 
+  React.useEffect(() => {
```
rohitkrai03, line 43: Would be probably good to move all this logic into a custom hook.

frontend/packages/dev-console/src/components/import/git/GitSection.tsx
```diff
@@ -100,7 +100,7 @@ const GitSection: React.FC<GitSectionProps> = ({
         values.application.selectedKey !== UNASSIGNED_KEY &&
         setFieldValue('application.name', `${gitRepoName}-app`);
 
-      if (buildStrategy === 'Devfile') {
+      if (buildStrategy === 'Devfile' && !values.devfile?.devfileSourceUrl) {
```
rohitkrai03, line 103: Wouldn't this create issues where user adds a git URL and then changes it to something else? This logic would not run the second time and still use the previous devfile contents.

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Moved to custom hook.

frontend/packages/dev-console/src/components/import/git/GitSection.tsx
rottencandy: I don't think this is an issue, the devfile contents from the user provided git repo are not used anyways,
and the devfile source repo would be updated in the custom hook.

rohitkrai03: I am thinking about the normal Devfile flow where user comes in directly from Add page, in that flow we want to rerun the detection every time user updates the git repo URL. But the custom hook only runs for git repo URL coming from query argument in case of sample or catalog flow.

rottencandy: In the normal devfile flow, we aren't using another repo for the devfile source, so `values.devfile.devfileSourceUrl` would not be set.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -23,7 +27,34 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+  const gitRepoUrl = searchParams.get('gitRepo');
+  const devfileName = searchParams.get('devfileName');
+  const formType = searchParams.get('formType');
+  const [devfileSamples, setDevfileSamples] = React.useState<DevfileSample[]>();
+  const selectedSample = React.useMemo(
+    () => devfileSamples?.find((sample) => sample.name === devfileName),
+    [devfileSamples, devfileName],
+  );
   useDevfileDirectoryWatcher(values, setFieldValue);
+  useDevfileSource(gitRepoUrl, devfileName, formType, setFieldValue);
```
rohitkrai03, line 40: You can actually use `useFormikContext` and that you won't have to send `setFieldValue`. Also he query params can be fetched inside the hook.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
```diff
@@ -23,7 +27,34 @@ const DevfileImportForm: React.FC<FormikProps<FormikValues> & DevfileImportFormP
...
+
+  React.useEffect(() => {
```
rohitkrai03, line 42: This can also be inside a custom hook that returns the `selectedDevfile`. Something like - 

`const selectedDevFile = useSelectedDevFile()`

rohitkrai03 (changes requested):

frontend/packages/dev-console/src/components/import/git/GitSection.tsx
rohitkrai03: Ah, so its not being used anywhere else except in catalog flow. Makes sense.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Updated.

frontend/packages/dev-console/src/components/import/devfile/DevfileImportForm.tsx
rottencandy: Updated.

frontend/packages/dev-console/src/components/catalog/providers/useDevfileSamples.tsx
```diff
@@ -0,0 +1,69 @@
...
+import * as _ from 'lodash';
+import { useTranslation } from 'react-i18next';
+import { TFunction } from 'i18next';
+import { APIError } from '@console/shared';
+import { ExtensionHook, CatalogItem } from '@console/dynamic-plugin-sdk';
+import { coFetchJSON } from '@console/internal/co-fetch';
+import { DevfileSample } from '../../import/devfile/devfile-types';
+
+const normalizeDevfileSamples = (devfileSamples: DevfileSample[], t: TFunction): CatalogItem[] => {
+  const normalizedDevfileSamples = _.map(devfileSamples, (sample) => {
```
invincibleJai, line 11: we can  avoid `lodash` here

rottencandy: Updated.

frontend/packages/dev-console/src/const.ts
```diff
@@ -18,3 +18,5 @@ export enum CONNECTOR_INCONTEXT_ACTIONS {
 }
 
 export const SERVICE_BINDING_ENABLED = 'SERVICE_BINDING_ENABLED';
+
+export const DEFAULT_APPLICATION_GROUP = 'sample-app';
```
rohitkrai03, line 22:
```suggestion
export const SAMPLE_APPLICATION_GROUP = 'sample-app';
```

rohitkrai03 (changes requested):
One last small change. Everything else looks good to me.

frontend/packages/dev-console/src/const.ts
rottencandy: Updated.

rohitkrai03 (approved):
/approve

Changes looks good to me. We'll wait for the backend PR to get in.

frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
```diff
@@ -107,3 +112,63 @@ export const useDevfileDirectoryWatcher = (
...
+        });
+
+      if (formType === 'sample') {
+        setFieldValue('name', createComponentName(devfileName));
+        setFieldValue('git.url', devfileSourceUrl);
+        setFieldValue('git.type', gitType);
+        setFieldValue('application.initial', SAMPLE_APPLICATION_GROUP);
+        setFieldValue('application.name', SAMPLE_APPLICATION_GROUP);
+        setFieldValue('application.selectedKey', SAMPLE_APPLICATION_GROUP);
+      }
```
invincibleJai, line 143: shouldn't we be setting name, git url and type otherwise as well here as even for catalog flow we need to update it?

```suggestion
        setFieldValue('name', createComponentName(devfileName));
        setFieldValue('git.url', devfileSourceUrl);
        setFieldValue('git.type', gitType);
      if (formType === 'sample') {
        setFieldValue('application.initial', SAMPLE_APPLICATION_GROUP);
        setFieldValue('application.name', SAMPLE_APPLICATION_GROUP);
        setFieldValue('application.selectedKey', SAMPLE_APPLICATION_GROUP);
      }
```

frontend/packages/dev-console/src/components/import/devfile/DevfileSampleInfo.tsx
```diff
@@ -0,0 +1,48 @@
...
+            className="co-catalog-item-icon__img co-catalog-item-icon__img--large"
+            src={iconUrl}
+            alt={displayName}
+          />
+        )}
+        <div>
+          <h2 className="co-section-heading co-catalog-item-details__name">{displayName}</h2>
+          {tags && (
+            <p className="co-catalog-item-details__tags">
+              {_.map(tags, (tag, i) => (
```
invincibleJai, line 31: nit: can avoid lodash

frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
rottencandy: Updated.

invincibleJai:
/lgtm

verified the changes and works as expected

rottencandy:
> @rottencandy Tested this PR on a cluster bot. Noticed some issue with the catalog flow where the Git type field flickers for a moment when the form is submitted.

This does not seem to be specific to devfile import but happens on all the flows which use `GitSection`, filed a bug [ODC-5826](https://issues.redhat.com/browse/ODC-5826) to track it.

rohitkrai03 (approved):
/lgtm

Did a proper QE testing of the PR using a cluster created through cluster bot. Everything works as expected.

GIF - 

https://user-images.githubusercontent.com/6041994/117885025-ee889b00-b2ca-11eb-9646-05b38bbe0d12.mov



/label qe-approved 
on behalf of devconsole QE team after QE testing the PR as agreed upon.

/label px-approved

/label docs-approved
added on behalf of docs team as the epic has doc-ack label. The feature is dev preview and marked with no-docs label.

PR #8795, opened 2021-04-27
Use latest WebTerminal plugin instead of 4.5.0

sleshchenko:
Initially, it was thought openshift console will use a different terminal plugin version to let the operator know which tooling is needed. But then the stuff changed and we just have 4.5.0 web terminal plugin which is used for every OpenShift plugin, which is a bit confusing.
So, WTO 1.3(Unreleased for OpenShift 4.8), 1.2.1(OpenShift 4.7) and 1.1.1(OpenShift 4.6) embed `latest` terminal plugin version, and OpenShift console should use it instead.

`4.5.0` is not removed as well just for backward compatibility, so, all previously created CRs should work as well.

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/cloud-shell-utils.ts
+++ b/frontend/packages/console-app/src/components/cloud-shell/cloud-shell-utils.ts
@@ -57,7 +57,7 @@ export const newCloudShellWorkSpace = (name: string, namespace: string): CloudSh
         {
           plugin: {
             name: 'web-terminal',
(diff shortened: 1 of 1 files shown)


sahil143 (approved):
/lgtm

tested changes locally. works as expected

christianvogt:
/label qe-approved based on sahil's testing and my testing of the same

/approved
/hold

This change is an implementation detail. No change to docs or user experience.
cc @RickJWagner @Preeticp

christianvogt:
/label docs-approved
no docs required
/hold cancel

PR #8797, opened 2021-04-27
multiline pipeline params: remove max height for multilined fields

abhinandan13jan:
# Addresses
https://issues.redhat.com/browse/ODC-5754

# Screenshot
Screenshot from 2021-04-29 16-52-27

# Test
no changes

# Browser conformance
Chrome

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnField.scss
+++ b/frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnField.scss
@@ -3,10 +3,6 @@
     position: relative;
     margin-right: var(--pf-global--spacer--lg);
     margin-bottom: var(--pf-global--spacer--md);
-
-    &--singleLine {
-      max-height: var(--pf-global--spacer--xl);
-    }
   }
   &__header {
     position: relative;
@@ -22,8 +18,7 @@
     line-height: var(--pf-global--spacer--xl);
     position: absolute;
     right: -30px;
-    top: 50%;
-    transform: translateY(-50%);
+    top: 0;
   }
   &__empty-message {
     margin-bottom: var(--pf-global--spacer--sm);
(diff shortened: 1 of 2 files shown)


frontend/packages/console-shared/src/components/formik-fields/field-types.ts
```diff
@@ -97,6 +97,7 @@ export interface MultiColumnFieldProps extends FieldProps {
   emptyMessage?: string;
   headers: ({ name: string; required: boolean } | string)[];
   complexFields?: boolean[];
+  isMultiline?: boolean;
```
andrewballantyne, line 100: As discussed on slack, but for posterity, this is not a valid solution. Almost all our usages of MultiColumnField are single lines suffering from this height issue. This prop would effectively always need to be on for every usage. 

As noted in the comments (and in Slack DMs) aim to see about removing the offending css and see if everything works fine. It's unclear why it was added but it seems to be an issue now (possible PF update of spacing to cause this issue)

abhinandan13jan: @andrewballantyne I updated and removed the undesired CSS block. I tested many areas like Pipelines, PipelineBuilder, Project Access which makes use of the Component and they all seem to work fine

andrewballantyne (changes requested):
No objection to the changes you made, but I think the height change makes a weird shift in the remove button:

Screen Shot 2021-04-28 at 5 40 18 PM

If we just modify the CSS a tiny bit more, we can achieve a desired experience here. In `frontend/packages/console-shared/src/components/formik-fields/multi-column-field/MultiColumnField.scss`, if we modify the `&__col--button` and set `top` to `0` and remove `transform` we can align it to the top of the cell so it does not shift down. 

Before/After on a row:
Screen Shot 2021-04-28 at 5 40 18 PM
Screen Shot 2021-04-28 at 5 39 57 PM

Before/After on the larger grouping:
Screen Shot 2021-04-28 at 5 40 25 PM
Screen Shot 2021-04-28 at 5 40 34 PM

abhinandan13jan:
@andrewballantyne Thanks for the input, I've updated accordingly

andrewballantyne (approved):

jerolimov (approved):
Tested this locally. Works fine

/lgtm

PR #8799, opened 2021-04-27
Bug 1951212:  User/Group details shows unrelated subjects in role bindings tab

cyril-ui-developer:
/assign @spadgett 

Screen Shot 2021-04-27 at 10 48 09 AM

### DIFF
--- a/frontend/public/components/factory/table-filters.ts
+++ b/frontend/public/components/factory/table-filters.ts
@@ -100,20 +100,10 @@ export const tableFilters: TableFilterMap = {
   'role-binding-roleRef-kind': (kind: string, binding) => binding.roleRef.kind === kind,
 
   // Filter role bindings by user name
-  'role-binding-user': (userName, binding) =>
-    _.some(binding.subjects, {
-      kind: 'User',
-      apiGroup: 'rbac.authorization.k8s.io',
-      name: userName,
-    }),
+  'role-binding-user': (userName, { subject }) => subject.name === userName,
 
   // Filter role bindings by group name
-  'role-binding-group': (groupName, binding) =>
-    _.some(binding.subjects, {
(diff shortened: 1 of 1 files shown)


frontend/public/components/factory/table-filters.ts
```diff
@@ -100,12 +100,7 @@ export const tableFilters: TableFilterMap = {
   'role-binding-roleRef-kind': (kind: string, binding) => binding.roleRef.kind === kind,
 
   // Filter role bindings by user name
-  'role-binding-user': (userName, binding) =>
-    _.some(binding.subjects, {
-      kind: 'User',
-      apiGroup: 'rbac.authorization.k8s.io',
-      name: userName,
-    }),
+  'role-binding-user': (userName, { subject }) => subject.name === userName,
```
florkbr, line 103: @cyril-ui-developer do you know if `subjects` array in the original `binding` object will ever contain more than one array item? Wondering if we should display a binding that contains the `userName` in the array of `subjects` even if other subjects are present? (but not those bindings that do not contain the user name).

image

florkbr:
/lgtm

Note - I cannot speak yet to what the intended design is here. Deferring to @spadgett on if this is the correct/intended behavior. 

## Before:
Screen Shot 2021-04-27 at 11 36 33 AM

## After:
Screen Shot 2021-04-27 at 11 36 47 AM

cyril-ui-developer:
Group RoleBinding
Before:
Screen Shot 2021-04-27 at 10 48 09 AM

After:
Screen Shot 2021-04-27 at 3 04 20 PM

spadgett (approved):
/lgtm

Thanks @cyril-ui-developer. I think the new behavior is what was intended.

PR #8803, opened 2021-04-27
Bug 2009420: Use live regions for alerts in modals

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2009420

This came up as a screen reader user typed in an upper case char during project creation displaying an error alert in the create project modal.
The alert does not get announced by the screen reader. Creating a liveRegion is a way to announce errors for screen readers. 

This is a pr to test with the jaws screen reader.

cc: @jessiehuff 

Before:

https://user-images.githubusercontent.com/35978579/135485341-0c32508b-d6c6-4715-8eeb-4056d77d2843.mov

After:

https://user-images.githubusercontent.com/35978579/135485565-1afa76d3-8b3a-4424-8dc3-f78f21ab23aa.mov

### DIFF
--- a/frontend/public/components/utils/button-bar.jsx
+++ b/frontend/public/components/utils/button-bar.jsx
@@ -25,16 +25,35 @@ const ErrorMessage = ({ message }) => {
       className="co-alert co-alert--scrollable"
       variant="danger"
       title={t('public~An error occurred')}
+      isLiveRegion
+      aria-live="assertive"
+      aria-atomic="true"
     >
       <div className="co-pre-line">{message}</div>
     </Alert>
   );
 };
 const InfoMessage = ({ message }) => (
-  <Alert isInline className="co-alert" variant="info" title={message} />
+  <Alert
+    isInline
+    className="co-alert"
+    variant="info"
+    title={message}
+    isLiveRegion
+    aria-live="polite"
+    aria-atomic="true"
+  />
 );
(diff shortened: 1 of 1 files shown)


dtaylor113 (approved):
/lgtm

zherman0 (approved):

jcaianirh:
/hold
for discussions and testing with jessie

jessiehuff:
Hey @jcaianirh, we updated our alert and alert groups to make this easier on products, and I think all you should need to do is wrap it in an AlertGroup with the same properties that you have in this PR ([example](https://www.patternfly.org/v4/components/alert#dynamic-live-region-alert)). I also created a11y documentation to try to spell this out easily [here](https://www.patternfly.org/v4/components/alert/accessibility).

jcaianirh:
> Hey @jcaianirh, we updated our alert and alert groups to make this easier on products, and I think all you should need to do is wrap it in an AlertGroup with the same properties that you have in this PR ([example](https://www.patternfly.org/v4/components/alert#dynamic-live-region-alert)). I also created a11y documentation to try to spell this out easily [here](https://www.patternfly.org/v4/components/alert/accessibility).

This is great news @jessiehuff. Can you tell me which version of patternfly-react we would need to pick up this change? We currently are on 2021-10.

spadgett (approved):
/lgtm

PR #8804, opened 2021-04-27
GitOps- Convert list card view to table view

reginapizza:
This PR would change the current GitOpsListPage from a card view to a table view. This work is being done as part of [GITOPS-557](https://issues.redhat.com/browse/GITOPS-557)/[ODC-5157](https://issues.redhat.com/browse/ODC-5157) (further broken down into [GITOPS-878](https://issues.redhat.com/browse/GITOPS-878)). This is the first stage of the implementation, as work on the backend to get the rest of the required data is still in progress; this PR just take the data that we were already using with the card view and converts it into a table (separate PR's will follow to continue with this work)

Mocks can be found [here](https://docs.google.com/document/d/1Peh17eBJnaMLDr-BPIGvImsRGOmrzO3Y2vKTXJmsJCY/edit).

https://user-images.githubusercontent.com/50851526/117678419-5e890b00-b17d-11eb-9ddd-59e3065bb059.mp4

@bgliwa01 @serenamarie125 @wtam2018 @keithchong @chetan-rns


bgliwa01 (commented):
Looks good except the table headers should be sentence case so "Application name" "Git repository" "Environment status" and "Last deployment"

reginapizza:
@bgliwa01 thanks for catching that, just fixed it :)

rohitkrai03:
@reginapizza I think you have some CI failures. Seems like `i18n` issue, running `yarn i18n` should fix it.

frontend/packages/gitops-plugin/src/components/GitOpsListPage.tsx
```diff
@@ -43,41 +41,21 @@ const GitOpsListPage: React.FC = () => {
...
-      />
+      <PageHeading title={t('gitops-plugin~Environments')} badge={<DevPreviewBadge />} />
       {!appGroups && !emptyStateMsg ? (
         <LoadingBox />
       ) : (
-        <GitOpsList appGroups={appGroups} emptyStateMsg={emptyStateMsg} argocdLink={argocdLink} />
+        <>
+          <p className="co-catalog-page__description">
+            Select an application to view the environment it&apos;s deployed in.
+          </p>
```
rohitkrai03, line 56: I think you can use `PageHeading` to set this subtitle. Sending this as a child to `PageHeading` would work.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,46 +1,119 @@
...
+  classNames('col-lg-2', 'col-md-2', 'col-sm-4', 'hidden-xs'),
+  classNames('col-lg-2', 'col-md-2', 'hidden-sm', 'hidden-xs'),
+  classNames('col-lg-2', 'hidden-md', 'hidden-sm', 'hidden-xs'),
+];
 interface GitOpsListProps {
-  appGroups: GitOpsAppGroupData[];
+  appGroups: GitOpsAppGroupData;
   emptyStateMsg: string;
-  argocdLink?: K8sResourceKind;
+  props?: any;
```
rohitkrai03, line 23: What is this being used for? Don't think the parent component is sending anything except `appGroups` and `emptyStateMsg`.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,46 +1,119 @@
...
 
-const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg, argocdLink }) => {
+const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg, props }) => {
   const { t } = useTranslation();
+
+  const handleClick = (appGroup) => {
+    history.push(`/environments/${appGroup.name}?url=${appGroup.repo_url}`);
+  };
+
+  const GitOpsTableHeader = () => {
```
rohitkrai03, line 33: Would be better to pull this out into its own file. Keeping this in the render context of component is not a good idea.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,46 +1,119 @@
...
+      },
+      {
+        title: t('gitops-plugin~Last deployment'),
+        sortField: 'lastDeployment',
+        transforms: [sortable],
+        props: { className: tableColumnClasses[3] },
+      },
+    ];
+  };
+  const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = ({
```
rohitkrai03, line 61: Same as above.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,46 +1,119 @@
...
+          {...props}
+        />
       ) : (
         <GitOpsEmptyState emptyStateMsg={emptyStateMsg} />
       )}
     </div>
   );
 };
 
+GitOpsList.displayName = 'GitOpsList';
```
rohitkrai03, line 117: Don't think you need to set this explicitly.

rohitkrai03 (changes requested):

reginapizza:
just a heads up, still working on extracting the header and row functions into their own files and then I'll push changes (just facing an error with it at the moment and need more time to figure it out)- in the meantime though this PR is still a WIP and should not be merged.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,41 +1,51 @@
...
-              </SplitItem>
-            </Split>
-          </StackItem>
-          {_.map(appGroups, (appGroup) => (
-            <StackItem key={`${appGroup.name}-${appGroup.repo_url}`}>
-              <GitOpsListItem appGroup={appGroup} argocdLink={argocdLink} />
-            </StackItem>
-          ))}
-        </Stack>
+      {!emptyStateMsg && !appGroups ? (
```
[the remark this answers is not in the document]
divyanshiGupta: Also it seems there is no need of adding the check for `appGroups` here as there is already a check `!emptyStateMsg && !appGroups` in `GitopsListPage` which when false will render the `LoadingBox` so `GitopsList` will either get `emptyStateMsg` as true or false.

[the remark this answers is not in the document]
reginapizza: ahh thanks for catching that! Maybe I didn't have my file saved after I made the changes... should be okay now though!

reginapizza:
@divyanshiGupta @rohitkrai03 could you please re-review? Thanks!

frontend/packages/gitops-plugin/src/components/GitOpsListPage.tsx
```diff
@@ -43,41 +41,21 @@ const GitOpsListPage: React.FC = () => {
...
-        }
-      />
+      <PageHeading title={t('gitops-plugin~Environments')} badge={<DevPreviewBadge />} />
       {!appGroups && !emptyStateMsg ? (
         <LoadingBox />
       ) : (
-        <GitOpsList appGroups={appGroups} emptyStateMsg={emptyStateMsg} argocdLink={argocdLink} />
+        <>
+          <PageHeading className="co-catalog-page__description">
+            Select an application to view the environment it&apos;s deployed in.
```
andrewballantyne, line 55: Needs to be i18n'ed

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <TableData className={classNames(tableColumnClasses[1])}>
+        <ExternalLink href={appGroup.repo_url} additionalClassName={'co-break-all'}>
+          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>
+            {routeDecoratorIcon(appGroup.repo_url, 12, t)}
+          </span>
+          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>{appGroup.repo_url}</span>
+        </ExternalLink>
+      </TableData>
+      <TableData className={tableColumnClasses[2]}>
+        <span>{appGroup.environments.join(', ')}</span>
```
andrewballantyne, line 45: You shouldn't need unstyled spans. Remove?

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </TableData>
+      <TableData className={tableColumnClasses[3]}>
+        {/* this is just a placeholder until backend changes can go in to get this data */}
+        <span>-</span>
```
andrewballantyne, line 49: Same here

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { routeDecoratorIcon } from '@console/dev-console/src/components/import/render-utils';
+
+const tableColumnClasses = [
+  classNames('pf-m-width-20'), // Application name
+  classNames('pf-m-width-40'), // Git repository
+  classNames('pf-m-hidden', 'pf-m-visible-on-md', 'pf-m-width-20'), // Environments
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-20'), // Last deployment
+];
+
+const handleClick = (appGroup) => {
```
andrewballantyne, line 17: Type standalone functions the best you can...

```suggestion
const handleClick = (appGroup: GitOpsAppGroupData) => {
```

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { obj: appGroup, index, key, style } = props;
+  const t = (tKey) => i18n.t(tKey);
+  return (
+    <TableRow id={index} index={index} trKey={key} style={style}>
+      <TableData className={tableColumnClasses[0]}>
+        <Button
+          onClick={() => handleClick(appGroup)}
+          aria-label={appGroup.name}
+          variant="link"
+          style={{ padding: '0px' }}
```
andrewballantyne, line 31: Would the prop `isInline` get you what you're looking for instead of custom styles?

andrewballantyne (changes requested):
Some small changes, will test it out once I get a cluster to test it on.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -1,41 +1,49 @@
...
 
-const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg, argocdLink }) => {
+const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg }) => {
   const { t } = useTranslation();
+  const [textFilter, setTextFilter] = React.useState('');
+
+  const visibleItems = appGroups?.filter(({ name }) => {
+    return (
+      fuzzyCaseInsensitive(textFilter, name) ||
+      name.toLowerCase().indexOf(textFilter.toLowerCase()) !== -1
```
andrewballantyne, line 23: Fuzzy should cover an indexOf check.

andrewballantyne (changes requested):
One more comment - Otherwise looks good.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
reginapizza: Sorry not sure I understand this one, can you elaborate? Or point to somewhere this is happening? I tried to look at other implementations and I don't think I'm seeing this unless I'm just missing it

andrewballantyne: `fuzzyCaseInsensitive` covers the "indexOf" case… so you’ll never end up with `false || true` in this statement… you’ll either get `true || true`, `true || false` or naturally `false || false` because Fuzzy is a superset calculation of the indexOf string.

reginapizza: ohhh I see, okay. I saw it with the `indexOf` check in other places so I thought it might be necessary. I'll remove it, thanks for the explanation!

frontend/packages/gitops-plugin/src/components/GitOpsListPage.tsx
```diff
@@ -43,41 +41,23 @@ const GitOpsListPage: React.FC = () => {
...
-      />
+      <PageHeading title={t('gitops-plugin~Environments')} badge={<DevPreviewBadge />} />
       {!appGroups && !emptyStateMsg ? (
         <LoadingBox />
       ) : (
-        <GitOpsList appGroups={appGroups} emptyStateMsg={emptyStateMsg} argocdLink={argocdLink} />
+        <>
+          <PageHeading className="co-catalog-page__description">
+            {t(
+              `gitops-plugin~Select an application to view the environment it&apos;s deployed in.`,
```
andrewballantyne, line 56:
```suggestion
              "gitops-plugin~Select an application to view the environment it's deployed in.",
```
Also you'll need to add this to the translation file & run `yarn i18n`

andrewballantyne (changes requested):

andrewballantyne (approved):
/lgtm

Issue #8729, opened 2021-04-21
Console Devfile Import Dev Preview broken


maysunfaisal:
When I use the sample repository from the Devfile Import Form, the pod never comes up:
```
NAME                                  READY   STATUS             RESTARTS   AGE     LABELS
pod/devfile-sample-1-build            0/1     Completed          0          9m20s   openshift.io/build.name=devfile-sample-1
pod/devfile-sample-54794cb986-g5g5g   1/2     ImagePullBackOff   0          8m23s   app=devfile-sample,deploymentconfig=devfile-sample,pod-template-hash=54794cb986
```

When I describe the pod, there are two containers, one of them being invalid:
```
Containers:
  buildguidance:
    Container ID:   
    Image:          buildguidanceimage-placeholder
    Image ID:       
    Port:           3001/TCP
    Host Port:      0/TCP
    State:          Waiting
      Reason:       ImagePullBackOff
    Ready:          False
    Restart Count:  0
    Limits:
      memory:  1Gi
    Requests:
      memory:  1Gi
    Environment:
      PROJECTS_ROOT:   /projects
      PROJECT_SOURCE:  /projects
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-bgqxs (ro)
  devfile-sample:
    Container ID:   cri-o://f4f4eed3105233ac5e2a2b115a5b499496684886145561e9fb3f011a14b99980
    Image:          image-registry.openshift-image-registry.svc:5000/mjf/devfile-sample@sha256:f4c7d9b18328b1b722885c4ddce77f4dd15a9f031d513626e5d5786cbf2c602e
    Image ID:       image-registry.openshift-image-registry.svc:5000/mjf/devfile-sample@sha256:f4c7d9b18328b1b722885c4ddce77f4dd15a9f031d513626e5d5786cbf2c602e
    Port:           <none>
    Host Port:      <none>
    State:          Running
      Started:      Wed, 21 Apr 2021 12:15:55 -0400
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-bgqxs (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  default-token-bgqxs:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-bgqxs
    Optional:    false
[...]

PR #8807, opened 2021-04-28
Bug 1952214: Save additional Devfile container parameters from build guidance container and drop it to fix ImagePullBackOff

jerolimov:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1952214
#8729

**Analysis / Root cause**: 
When importing the sample Devfile from git repo https://github.com/redhat-developer/devfile-sample the new Deployment doens't start successfully. The Pod fails to start because the container image "buildguidanceimage-placeholder" could not be fetched.

The reason is that the [devfile.yaml](https://github.com/redhat-developer/devfile-sample/blob/5ffa4845474ed8321cba9b69f47f02ca16210334/devfile.yaml#L9-L17) contains a "build guidance" container which should be dropped. The console should only keep the configure from this container, like the right ports, environment variables and resource limits.

**Solution Description**: 
When submitting the `DevfileImportForm` it calls `createDevfileResources` which then calls `createOrUpdateDeployment` with the `formData` and the parsed `devfileResourceObjects.deployResource`. See [import-submit-utils.ts](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/import/import-submit-utils.ts#L510-L516)

Because `createOrUpdateDeployment` loads the ports, environment variables and limits from the `formData` it is required to change it before that. It's also required to remove the "placeholder" container from the parsed `deployResource`.

I decided to do this in the form right after parsing the Devfile. This will allow us to validate or maybe provide additional fields for these parameters later.

Created Deployment without this PR:

```yaml
kind: Deployment
spec:
  template:
    spec:
      containers:
        - name: buildguidance
          image: buildguidanceimage-placeholder
          ports:
            - name: http-3001
              containerPort: 3001
              protocol: TCP
          env:
            - name: PROJECTS_ROOT
              value: /projects
            - name: PROJECT_SOURCE
              value: /projects
          resources:
            limits:
              memory: 1Gi
        - name: devfile-sample
          image: >-
            image-registry.openshift-image-registry.svc:5000/christoph/devfile-sample@sha256:3bafb69d0c1757b0e7db139ecf5d12812b56d584de940dbb6ac2065b62097de4
          resources: {}
```

Created Deployment with this PR:

```yaml
kind: Deployment
spec:
  template:
    spec:
      containers:
        - name: devfile-sample9
          image: >-
            image-registry.openshift-image-registry.svc:5000/christoph/devfile-sample9@sha256:25bc781ebdabd2602549e2ce0d1f40ae217889156a6c333be46b7d6c135475ed
          ports:
            - name: http-3001
              containerPort: 3001
              protocol: TCP
          env:
            - name: PROJECTS_ROOT
              value: /projects
            - name: PROJECT_SOURCE
              value: /projects
          resources:
            limits:
              memory: 1Gi
```

UI does not change.

Created deployment without this PR:
image

Created deployment with this PR:
image

Created a new unit test `createDevfileResources` before I decided to move this changes "up" from this function to the form hook. I kept the new tests also if it doesn't verify the build guidance container change here... :-/

```
    createDevfileResources
      ✓ return a Deployment with just one container which includes ports, env from Devfile container  (3ms)
```

**Test setup:**
1. Open the Developer perspective
2. Open Add > "From Devfile"
3. Press "Try sample" to import the Devfile from `https://github.com/redhat-developer/devfile-sample`
4. Create the Deployment and verify if the pods starts successfully.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug
/assign @andrewballantyne 
/cc @andrewballantyne @maysunfaisal @elsony

### DIFF
--- a/frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
+++ b/frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
@@ -1,15 +1,16 @@
 import * as React from 'react';
 import { FormikValues } from 'formik';
 import { useTranslation } from 'react-i18next';
+import { getLimitsDataFromResource } from '@console/shared/src';
 import { parseDevfile } from '@console/internal/module/k8s';
 import { DevfileSuggestedResources } from '../import-types';
 
 const suffixSlash = (val: string) => (val.endsWith('/') ? val : `${val}/`);
 const prefixDotSlash = (val) => (val.startsWith('/') ? `.${val}` : val);
 
-export const useDefileServer = (
+export const useDevfileServer = (
   values: FormikValues,
-  setFieldValue: (name: string, value: any) => any,
+  setFieldValue: (name: string, value: any, shouldValidate?: boolean) => any,
 ): [boolean, string] => {
   const { t } = useTranslation();
   const [devfileParseError, setDevfileParseError] = React.useState<string>(null);
@@ -56,6 +57,17 @@ export const useDefileServer = (
         if (value) {
           clearError();
           const { imageStream, buildResource, deployResource, service, route } = value;
+
+          if (deployResource.spec?.template?.spec?.containers?.length > 0) {
+            const buildGuidanceContainer = deployResource.spec.template.spec.containers[0];
+
+            setFieldValue('image.ports', buildGuidanceContainer.ports || [], false);
+            setFieldValue('deployment.env', buildGuidanceContainer.env || [], false);
+            setFieldValue('limits', getLimitsDataFromResource(deployResource), false);
+
+            delete deployResource.spec.template.spec.containers;
+          }
+
           setFieldValue('devfile.devfileSuggestedResources', {
             imageStream,
             buildResource,
--- a/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils-data.ts
+++ b/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils-data.ts
@@ -492,3 +492,372 @@ export const nodeJsBuilderImage = {
   },
   imageStreamNamespace: 'openshift',
 };
+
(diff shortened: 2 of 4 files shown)


jerolimov:
> AssertionError: Timed out retrying after 30000ms: Expected to find element: `[data-test-id="operator-uninstall-btn"]`, but never found it. 
> at Context.eval (https://console-openshift-console.apps.ci-op-dtzgiww8-75d12.*********************************/__cypress/tests?p=tests/operator-install-global.spec.ts:13649:57)

/retest

frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
```diff
@@ -56,6 +57,17 @@ export const useDefileServer = (
         if (value) {
           clearError();
           const { imageStream, buildResource, deployResource, service, route } = value;
+
+          if (deployResource.spec?.template?.spec?.containers?.length > 0) {
+            const buildGuidanceContainer = deployResource.spec.template.spec.containers[0];
+
+            setFieldValue('image.ports', buildGuidanceContainer.ports || [], false);
+            setFieldValue('deployment.env', buildGuidanceContainer.env || [], false);
```
andrewballantyne, line 65: Today it does not matter (they are the same value)... but I am wondering if we should refactor `getBaseInitialValues` to allow for you to use the static settings for `ports` and `env` and then use those here instead of empty. 

Since it's unlikely to change, I think we can avoid holding this PR up for that change... food for thought though since what you're really doing here is a reset to defaults.

andrewballantyne (approved):
/lgtm

maysunfaisal (commented):
I've tested it with my various node devfile samples and seems good so far 👍

PR #8810, opened 2021-04-28
Bug 1952635: fixes: Web console displays a blank page- white space instead of cluster information

sahil143:
This PR fixes the blank page issue  with safari https://bugzilla.redhat.com/show_bug.cgi?id=1952635

Screenshot 2021-04-28 at 1 37 51 PM

### DIFF
--- a/frontend/public/style/_layout.scss
+++ b/frontend/public/style/_layout.scss
@@ -32,6 +32,13 @@ body,
   flex-direction: column;
 }
 
+.co-m-app__content {
+  display: flex;
+  flex: 1 0 auto;
+  flex-direction: column;
+  height: 100% // doesn't work on safari without height 
+}
+
 .co-p-has-sidebar {
   position: relative;
   display: flex;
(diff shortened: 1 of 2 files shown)


frontend/public/style/_layout.scss
```diff
@@ -32,6 +32,13 @@ body,
   flex-direction: column;
 }
 
+.co-m-app__content {
+  display: flex;
+  flex: 1 0 auto;
+  flex-direction: column;
+  height: 100% // doesn't work on safari without height
```
andrewballantyne, line 39: Definitely a solution... tested and it works on Safari 14.

Alternative would be to just set the child to a `flex-grow: 1` to fill the page instead of this height.

christianvogt: Did you test the solution in both desktop and mobile? I've encountered issues with the PF drawer on mobile.
I don't know which solution is best without further investigating myself.
So @andrewballantyne are you wanting this change to be made or not? Since you didn't give lgtm

PR #8813, opened 2021-04-28
Bug 1941625: Add two missing translations for topology list view

jerolimov:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1941625

**Analysis / Root cause**: 
The BZ issue describes different areas in topology which are not translated yet. Most of them are already prepared for `t(ranslation)` but are not translated yet. We ignore this cases here. From the issue there are just one translations which needs to be prepared for translation:

image

While testing this I found another translation which is missing if the Deployment has some "Alerts"

**Solution Description**: 
* Added translation for the topology list section headlines which shows the resource label.
* Added translation for "Alerts:"

Translated topology list **before**:

image

Translated topology list **with this PR**:

image

The reason that DeploymentConfig is not translated yet, is because we can find Deployment translations but not DeploymentConfig translations:

image

Topology list with this enabled pseudo localization:

image

Not changed

**Test setup:**
1. Create a Deployment and DeploymentConfig
2. Open the topology list with enabled pseudo localization (like http://localhost:9000/topology/ns/christoph?view=list&pseudolocalization=true&lng=en)
3. Select an application to open the sidebar which shows Deployment and DeploymentConfigs again

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


jerolimov:
/kind bug
/cc @jeff-phillips-18 @rebeccaalpert

Jeff, Rebecca, can you check this solution? Thanks :)

frontend/packages/topology/src/components/list-view/list-view-utils.tsx
```diff
@@ -1,16 +1,27 @@
...
+  const model: K8sKind | undefined = modelFor(kindString);
+  if (model) {
+    if (model.labelKey) {
+      return model.labelKey;
+    }
+    return model.label;
+  }
+  return _.startCase(kindString);
+};
+
```
jeff-phillips-18, line 24: I'm surprised nothing like this exists already. Should this be in `console-shared` somewhere?

jeff-phillips-18 (approved):
/lgtm

PR #8815, opened 2021-04-28
Query Browser: Add seconds back to query browser tooltips

kyoto:
These were accidentally removed during the Console wide date format rework (https://github.com/openshift/console/pull/8490).

/cc @spadgett 

FYI @sichvoge, @cshinn 

screenshot
screenshot-1


spadgett (approved):
/lgtm

PR #8816, opened 2021-04-28
Bug 1941669: Translate pinned resources when labelPluralKey is available

jerolimov:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1941669

**Analysis / Root cause**:
From https://bugzilla.redhat.com/show_bug.cgi?id=1941669#c1 by Debsmita

> Except for the pinned `ConfigMaps` nav item in the dev perspective, the other highlighted string is already translated. It will reflect for ja & zh when their respective json files are updated.

So this PR focus on pinned navigation items. The pinned resources in the developer perspective just show the resource label before.

**Solution Description**: 
Instead of just showing the resource label it is required to translate the `labelPluralKey`.

Before with enabled pseudo localization:
Peek 2021-04-28 14-24

After with enabled pseudo localization:
Peek 2021-04-28 14-26

Translated values with this PR:
Peek 2021-04-28 14-28

Unchanged

**Test setup:**
1. Open the Developer perspective
2. Search for resources like ConfigMap or Secret and pin this resource
3. Switch to different languages to see that they are shown in their translated version.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/kind bug
/cc @rebeccaalpert

### DIFF
--- a/frontend/public/components/nav/perspective-nav.tsx
+++ b/frontend/public/components/nav/perspective-nav.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { NavItemSeparator, NavGroup, Button } from '@patternfly/react-core';
 import { MinusCircleIcon } from '@patternfly/react-icons';
 import {
@@ -11,7 +12,7 @@ import {
   isSeparatorNavItem,
 } from '@console/plugin-sdk';
 import { useActivePerspective, usePinnedResources } from '@console/shared';
-import { modelFor, referenceForModel } from '../../module/k8s';
+import { K8sKind, modelFor, referenceForModel } from '../../module/k8s';
 import { getSortedNavItems } from './navSortUtils';
 import confirmNavUnpinModal from './confirmNavUnpinModal';
 import { NavSection } from './section';
@@ -26,12 +27,8 @@ import {
 
 import './_perspective-nav.scss';
 
-const getLabelForResource = (resource: string): string => {
-  const model = modelFor(resource);
-  return model ? model.labelPlural : '';
-};
-
 const PerspectiveNav: React.FC<{}> = () => {
+  const { t } = useTranslation();
   const [perspective] = useActivePerspective();
   const allItems = useExtensions<PluginNavSection | NavItem | SeparatorNavItem>(
     isNavSection,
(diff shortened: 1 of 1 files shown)


frontend/public/components/nav/perspective-nav.tsx
```diff
@@ -64,14 +72,15 @@ const PerspectiveNav: React.FC<{}> = () => {
...
-            const { labelPlural, apiVersion, apiGroup, namespaced, crd, plural } = model;
+            const { apiVersion, apiGroup, namespaced, crd, plural } = model;
+            const label = getLabelForResource(resource);
             const duplicates =
-              pinnedResources.filter((res) => getLabelForResource(res) === labelPlural).length > 1;
+              pinnedResources.filter((res) => getLabelForResource(res) === label).length > 1;
             const props = {
               key: `pinned-${resource}`,
-              name: labelPlural,
+              name: label,
```
andrewballantyne, line 81: Hmmm, so if we can't get a model, but have a model... we need to try and have some default. Can we not use `model.label` as a fallback? We don't ever really want to show an empty sidebar item. Untranslated may not be good, but it's better than nothing.

rebeccaalpert: I think he's using the output of the check (including fallback) from line 76 here.

rebeccaalpert (approved):
This looks good to me.

frontend/public/components/nav/perspective-nav.tsx
andrewballantyne: Ah, looking closely, I missed the `if (!model)` check... seems odd to fetch it twice but I guess it makes the utility useful for the filter of duplicates.

nvm me! Looks good, just odd flow confused me on first glance haha.

andrewballantyne (approved):

frontend/public/components/nav/perspective-nav.tsx
```diff
@@ -56,6 +53,17 @@ const PerspectiveNav: React.FC<{}> = () => {
     return <AdminNav />;
   }
 
+  const getLabelForResource = (resource: string): string => {
+    const model: K8sKind | undefined = modelFor(resource);
+    if (model) {
+      if (model.labelPluralKey) {
+        return t(model.labelPluralKey);
+      }
+      return model.labelPlural || model.plural;
```
abhinandan13jan, line 62: Just a query, maybe in a later Bug/Feature should we not update the `labelPlural` and `plural` and add corresponding 
`// t(label)` entries in our model files just like we do for `labelKey` and `label`

PR #8817, opened 2021-04-28
Split cloud-init component to different logic and view components

metalice:


yaacov:
needs 'yarn i18n'

yaacov:
p.s.
can we add to the form this missing fields, because they are used in common templates ?
user:
password: 
chpasswd:

PR #8818, opened 2021-04-28
adding console.pages extension types to dynamic-plugin-sdk

glekner:


frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,6 +1,25 @@
 import { RouteComponentProps } from 'react-router';
 import { Extension, ExtensionDeclaration, CodeRef } from '../types';
 
+type ResourcePage = {
+  model: {
```
vojtechszocs, line 5: We have multiple declarations of type literal

```ts
model: {
  group: string;
  version: string;
  kind: string;
};
```

as well as this one

```ts
model: {
  group: string;
  version?: string;
  kind?: string;
};
```

I think it's best to create a helper type in `packages/console-dynamic-plugin-sdk/src/api/common-types.ts` to represent a k8s model/resource and use that one for better consistency.

This will also keep the generated JSON schema simpler (references to type definitions, instead of inlining those definitions everywhere they're used).

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
   }
 >;
 
+/** Adds new page to Console router. */
+export type RoutePage = ExtensionDeclaration<
+  'console.page/route',
+  {
+    /** The React component to load for the page. */
+    component: CodeRef<React.Component<RouteComponentProps>>;
```
vojtechszocs, line 41: `React.Component` refers to class components only. `React.FC` refers to function components only.

Please use `React.ComponentType` here, as well as in `StandaloneRoutePage` extension declaration above.

```ts
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```

In general, all extension declarations should use `React.ComponentType` when referring to a React component.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
   }
 >;
 
+/** Adds new page to Console router. */
+export type RoutePage = ExtensionDeclaration<
+  'console.page/route',
+  {
+    /** The React component to load for the page. */
```
vojtechszocs, line 40:
```suggestion
    /** The component to be rendered when the route matches. */
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
...
+    component: CodeRef<React.Component<RouteComponentProps>>;
+    /** The perspective ID to which this item belongs to. If not specified, contributes to all perspectives. */
```
vojtechszocs, line 42:
```suggestion
    /** The perspective to which this page belongs to. If not specified, contributes to all perspectives. */
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
...
+    perspective?: string;
+    /** Valid URL path or array of paths that `path-to-regexp@^1.7.0` understands. */
+    path: string | string[];
+    /** When true, will only match if the path matches the `location.pathname` exactly. */
+    exact?: boolean;
+    /** When true, a path that has a trailing slash will only match a location.pathname with a trailing slash. This has no effect when there are additional URL segments in the location.pathname. */
+    strict?: boolean;
```
vojtechszocs, line 49: Please go over all usages of the current `Page/Route` extension and see if we really need those additional properties, like `strict` and `sensitive`.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
...
+    /** When true, will match if the path is case sensitive. */
+    sensitive?: boolean;
+  }
+>;
+
+/** Adds new resource list to Console router. */
```
vojtechszocs, line 55:
```suggestion
/** Adds new resource list page to Console router. */
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
...
+export type ResourceListPage = ExtensionDeclaration<'console.page/resource/list', ResourcePage>;
+
+/** Adds new details page to Console router. */
```
vojtechszocs, line 58:
```suggestion
/** Adds new resource details page to Console router. */
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -14,7 +33,65 @@ export type StandaloneRoutePage = ExtensionDeclaration<
...
+    /** The name of the tab. */
+    name: string;
+    /** The optional href for the tab link. If not provided, the first `path` is used. */
+    href?: string;
+    /** Valid URL path or array of paths that `path-to-regexp@^1.7.0` understands. */
+    path: string | string[];
+    /** When true, will only match if the path matches the `location.pathname` exactly. */
+    exact?: boolean;
+    /** When true, a path that has a trailing slash will only match a location.pathname with a trailing slash. This has no effect when there are additional URL segments in the location.pathname. */
+    strict?: boolean;
```
vojtechszocs, line 77: Please go over all usages of the current `Page/Resource/Tab` extension and see if we really need those additional properties, like `strict` and `sensitive`.

vojtechszocs (changes requested):

vojtechszocs:
We should avoid duplicating common properties, like

```ts
/** The React component to load for the page. */
component: CodeRef<React.ComponentType<RouteComponentProps>>;
```

by creating meaningful helper types, e.g. in `packages/console-dynamic-plugin-sdk/src/extensions/navigation.ts` we have `NavItemProperties` type which is used in multiple ways, including

```ts
Omit<NavItemProperties, 'foo'> // remove properties
NavItemProperties & { bar: string } // add properties
Omit<NavItemProperties, 'foo'> & { bar: string } // combination of above
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
glekner: could be useful for later usages?

vojtechszocs: We can always add more properties later, I'd recommend to keep the API surface minimal :smiley:

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,20 +1,82 @@
...
+  exact?: boolean;
+};
+
+/** Adds new page to Console router. */
+export type RoutePage = ExtensionDeclaration<'console.page/route', RoutePageProperties>;
+
+/** Adds new resource list page to Console router. */
+export type ResourceListPage = ExtensionDeclaration<
+  'console.page/resource/list',
+  ResourcePageProperties & { model: ExtensionK8sModel }
```
vojtechszocs, line 36:
```suggestion
  ResourcePageProperties & {}
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,20 +1,82 @@
...
+>;
+
+/** Adds new resource details page to Console router. */
+export type ResourceDetailsPage = ExtensionDeclaration<
+  'console.page/resource/details',
+  ResourcePageProperties & { model: ExtensionK8sModel }
```
vojtechszocs, line 42:
```suggestion
  ResourcePageProperties & {}
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,20 +1,82 @@
...
+>;
+
+/** Adds new resource tab page to Console router. */
+export type ResourceTabPage = ExtensionDeclaration<
+  'console.page/resource/tab',
+  ResourcePageProperties & {
```
vojtechszocs, line 48:
```suggestion
  Omit<ResourcePageProperties, 'component'> & {
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,20 +1,82 @@
...
+    href?: string;
     /** When true, will only match if the path matches the `location.pathname` exactly. */
     exact?: boolean;
   }
 >;
 
+/** Adds new standalone page (rendered outside the common page layout) to Console router. */
+export type StandaloneRoutePage = ExtensionDeclaration<
+  'console.page/route/standalone',
+  Omit<RoutePageProperties, 'perspective' | 'strict' | 'sensitive'>
```
vojtechszocs, line 63:
```suggestion
  Omit<RoutePageProperties, 'perspective'>
```

frontend/packages/console-dynamic-plugin-sdk/src/extensions/pages.ts
```diff
@@ -1,20 +1,82 @@
...
+export const isResourceListPage = (e: Extension): e is ResourceListPage =>
+  e.type === 'console.page/resource/list';
+
+export const isResourceDetailsPage = (e: Extension): e is ResourceDetailsPage =>
+  e.type === 'console.page/resource/details';
+
+export const isResourceTabPage = (e: Extension): e is ResourceTabPage =>
+  e.type === 'console.page/resource/tab';
+
+export type ResourcePage = ResourceListPage | ResourceDetailsPage;
```
vojtechszocs, line 82: I'd suggest removing this union type here.

In `public/components/resource-pages.ts` we can add

```ts
type ResourcePage = ResourceListPage | ResourceDetailsPage;
```

since the fact that ^^ code handles both resource page extensions is not really relevant to extension definitions.

vojtechszocs (changes requested):

PR #8821, opened 2021-04-28
Bug 1934229: Improve performance of ToolbarFilter

TheRealJon:
- Debounce name filter
- Use memoization and React lifecycle hooks to minimize re-rendering


frontend/public/components/filter-toolbar.tsx
```diff
@@ -49,31 +52,20 @@ type FilterKeys = {
...
-                  ? _.filter(data, (d) => grp.isMatch(d, item.id)).length
-                  : _.countBy(data, grp.reducer)?.[item.id] ?? '0'}
-              </Badge>
-            </SelectOption>
-          );
-        })}
-      </SelectGroup>
-    );
-  });
+const RowFilterSelectGroup: React.FC<RowFilterDropdownGroupProps> = ({ rowFilter, data }) => (
```
TheRealJon, line 55:
```suggestion
const RowFilterSelectGroup: React.FC<RowFilterSelectGroupProps> = ({ rowFilter, data }) => (
```

frontend/public/components/filter-toolbar.tsx
```diff
@@ -49,31 +52,20 @@ type FilterKeys = {
...
-              </Badge>
-            </SelectOption>
-          );
-        })}
-      </SelectGroup>
-    );
-  });
+const RowFilterSelectGroup: React.FC<RowFilterSelectGroupProps> = ({ rowFilter, data }) => (
+  <SelectGroup key={rowFilter.filterGroupName} label={rowFilter.filterGroupName}>
+    {_.map(rowFilter.items ?? [], (item) => (
```
spadgett, line 57: nit: prefer optional chaining. (I think the `?? []` is unnecessary if using `_.map` as well since it handles null/undefined.)

spadgett (commented):
Jon, thanks. Code looks good overall.

I notice when testing that the first keystroke updates immediately, and then it waits until you pause. That seemed a little strange. It might be better to use the tail end of the debounce instead of updating immediately on the first keystroke then waiting.

We should consider updating immediately when the filter is removed as well. If you select all -> backspace, I'd expect the page to update right away. I'm not sure if that's possible with `useDebounceCallback`, though.

glekner:
any update here @TheRealJon ?

TheRealJon:
@glekner  Just pushed an update. @spadgett I addressed your comment and a few other things I found.

spadgett (approved):
/lgtm

PR #8826, opened 2021-04-29
adding sdk support for dynamic page extensions

glekner:
blocked by https://github.com/openshift/console/pull/8818

This is the first and I suppose the last extension which I don't migrate all consumings to use the new dynamic extensions because there are so many. So we now support both Static and Dynamic page extensions


glekner:
Tested all route extensions. `RoutePage, ResourceListPage, ResourceDetailsPage, ResourceTabPage` with console/app extensions, and they work fine

PR #8792, opened 2021-04-27
CONSOLE-2829: Convert list page to ts

rawagner:

PR #8827, opened 2021-04-29
New ListPage components

rawagner:
This PR revamps a way how to create List pages. Converts pods, machines and bindings pages as example.

Depends on https://github.com/openshift/console/pull/8792

### The old way

```js
<ListPage
  // ...gazillions of props
/>
```

I think I dont need to go into details - we know how painful the ListPage creation was.

## The new way - no more props drilling, use hooks & JSX composition
1. Fetch all resources you desire with `useK8sWatchResource` hook

```js
 const [machines, loaded, loadError] = useK8sWatchResource<K8sResourceCommon[]>({
    kind: referenceForModel(MachineModel),
    isList: true,
    namespace,
  });
```
2. use `useListPageFilter` to filter the resources yourself. The filtering is no longer done via redux and in table.

```js
const [data, filteredData, onFilterChange] = useListPageFilter(machines, rowFilters, staticFilters);
```

3. build List page with JSX, pass filtered data to table.

```js
<>
  <ListPageHeader title={showTitle ? t(MachineModel.labelPluralKey) : undefined}>
    <ListPageCreate groupVersionKind={referenceForModel(MachineModel)}>
      {t('machines~Create machine')}
    </ListPageCreate>
  </ListPageHeader>
  <ListPageBody>
    <ListPageFilter data={data} loaded={loaded} onFilterChange={onFilterChange} />
    <MachineList data={filteredData} loaded={loaded} loadError={loadError} />
  </ListPageBody>
</>
```


frontend/public/actions/k8s.ts
```diff
@@ -71,8 +71,11 @@ export const getResources = () => (dispatch: Dispatch) => {
     });
 };
 
-export const filterList = (id: string, name: string, value: string) =>
-  action(ActionType.FilterList, { id, name, value });
+export const filterList = (
+  id: string,
+  name: string,
+  value: string | string[] | { selected: Set<string>; all: string[] },
```
bipuladh, line 77:
```suggestion
  value: string | string[] | { selected: string[]; all: string[] },
```
We should not put non-serializable objects in the redux store.
https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

frontend/public/components/factory/ListPage/filter-hook.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { getAllTableFilters } from '../table';
+import { RowFilter } from '../../filter-toolbar';
+
+export type OnFilterChange = (type: string, value: string | string[]) => void;
+
+const filterData = <D>(
+  data: D[],
+  filter: { [key: string]: string | string[] } = {},
+  rowFilters?: RowFilter[],
```
bipuladh, line 11:
```suggestion
  rowFilters?: RowFilter[],
```
Is this really `RowFilter` ? 
Because `rowFilters[type]` should give undefined as `type` is `string`?

frontend/public/actions/k8s.ts
rawagner: not new, but yes, you're right. I've updated the code to not pass `Set` to store anymore

frontend/public/components/factory/ListPage/filter-hook.ts
rawagner: good catch!. The type was wrong, fixed

frontend/public/components/RBAC/bindings.jsx
```diff
@@ -280,63 +286,95 @@ export const bindingType = (binding) => {
...
+    !resources.ClusterRoleBinding.loadError;
+
+  const rowFilters = React.useMemo(
+    () => [
+      {
+        filterGroupName: t('public~Kind'),
+        type: 'role-binding-kind',
+        reducer: bindingType,
+        filter: (filter, binding) =>
+          filter.selected?.includes(bindingType(binding)) || filter.selected?.size === 0,
```
cyril-ui-developer, line 331: nit: The `filter.selected?.size === 0` could be written as `!filter.selected?.length`.

frontend/packages/kubevirt-plugin/src/components/vm-disks/table-filters.ts
```diff
@@ -18,7 +18,9 @@ export const diskSourceFilter: RowFilter = {
   filter: (disks, obj) => {
     const diskType = typeReducer(obj);
     return (
-      disks.selected.size === 0 || disks.selected.has(diskType) || !_.includes(disks.all, diskType)
+      !disks.selected.length ||
+      disks.selected.includes(diskType) ||
+      !_.includes(disks.all, diskType)
```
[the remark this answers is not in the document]
rawagner: right, fixed!

frontend/packages/kubevirt-plugin/src/components/vms/vm.tsx
```diff
@@ -274,6 +275,26 @@ const VMList: React.FC<React.ComponentProps<typeof Table> & VMListProps> = (prop
 
 VMList.displayName = 'VMList';
 
+const rowFilters: RowFilter<VMRowObjType>[] = [
+  {
+    filterGroupName: 'Status',
```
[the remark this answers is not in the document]
rawagner: changed to i18n key

frontend/public/components/RBAC/bindings.jsx
rawagner: updated

vojtechszocs:
:+1: looks good to me

rawagner:
/hold
e2e tests are failing

spadgett:
/label px-approved
/label docs-approved

There are no user visible changes, and we have approval for dynamic plugin APIs.

PR #8829, opened 2021-04-30
New virtualized table component

rawagner:
This PR revamps table component. Depends on revamped list pages https://github.com/openshift/console/pull/8827 . Converts PodList and MachineList pages.

New table is virtualized only for now. Non-virtualized table should be a separate component to keep the amount of props to minium IMO.

## Changes in new table
1. filtering - sice new list pages are taking care of filtering https://github.com/openshift/console/pull/8827 there is no need to have the functionality in the new table which simplified things quite a bit
2. sorting - colum (previously called Header, i've changed that to align with PF) provides a sort function - a specific implementation not just some string which loosly references implementation (previously `sortFunc`) or field selector (previously `sortField`) 
3. no more `customData` prop drilling
4. `Row` prop is an actual React.FC not a function which returns a JSX.


spadgett:
This is great to see and looks like a much cleaner API. Thank you @rawagner 👍 

I did want to mention we have a long standing bug with table performance tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1856355 and https://github.com/patternfly/patternfly-react/issues/5501. @jcaianirh and @dtaylor113 have been looking at it. If we're going to be reworking our tables in console anyway, it might be a good time to try to tackle that, particularly if it requires changes that touch all pages. @jcaianirh mentioned that the PatternFly composable table might perform better, but I'm not sure if that works with virtualization or how big a change that is. It's possible the performance issue is just an upstream PatternFly bug, but I wanted to raise it.

Ideally virtualization would just be a single boolean prop to enable or disable, but I'm not sure if that's possible since the PatternFly virtualized and non-virtualized tables have different APIs.

rawagner:
I tried composable table (example with virtualized table is [here](http://patternfly-react.surge.sh/extensions/virtual-scroll-table#using-composable-table-components) ) but I didnt see any performance improvement :/

rawagner:
@spadgett i've made some changes to cell cache and wrapped row rendering functions in useCallback and I think there's is a difference when scrolling but it still isnt completely fluid.

frontend/public/components/machine.tsx
```diff
@@ -162,13 +170,24 @@ const MachineDetails: React.SFC<MachineDetailsProps> = ({ obj }: { obj: MachineK
...
   const { t } = useTranslation();
-  const MachineTableHeader = () => {
-    return [
+
+  const machineTableColumn = React.useMemo<TableColumn<MachineKind>[]>(
+    () => [
       {
         title: t('public~Name'),
-        sortField: 'metadata.name',
+        sort: (data, direction) => {
```
spadgett, line 186: I'm unclear on why this was changed to `sort` whereas the other columns still use `sortField`

frontend/public/components/pod.tsx
```diff
@@ -193,109 +201,181 @@ const podRowStateToProps = ({ UI }) => ({
...
-        props: { className: Kebab.columnClass },
-      },
-    ];
-  };
-};
+const getColumns = (showNodes: boolean, t: TFunction): TableColumn<PodKind>[] => [
+  {
+    title: t(podColumnInfo.name.title),
+    id: podColumnInfo.name.id,
+    sort: (data, direction) => {
```
spadgett, line 208: I like being able to pass a sort function here, but I also like the `sortField` shorthand for the common case (which I'd guess is 80-90% of fields). Is it possible to support both? Even with the `sortResourceByValue` helper, there is still a bit of similar code that gets passed around otherwise.

frontend/public/components/pod.tsx
```diff
@@ -911,28 +1004,55 @@ export const PodsPage = connect<{}, PodPagePropsFromDispatch, PodPageProps>(
...
-            selectedColumns:
-              tableColumns?.[columnManagementID]?.length > 0
-                ? new Set(tableColumns[columnManagementID])
-                : null,
-            showNamespaceOverride: props?.customData?.showNamespaceOverride,
-            type: t('public~Pod'),
-          }}
-        />
+        <>
+          <ListPageHeader title={showTitle ? t('public~Pods') : undefined}>
```
spadgett, line 1023: I think this is a better API in general. It's more flexible and composable. It's a bit more code for the resource pages compared to using `ListPage`, but I think that's a reasonable tradeoff and the common header/filter/list components will help keep consistency. This will also allow dynamic plugins to do something special when needed without having to change `ListPage`.

spadgett (commented):
@rawagner Thank you for your work on this 👍 I know it's a big effort.

I think this is a good change overall. It would be good to get @christianvogt's opinion on the API.

frontend/packages/console-shared/src/hooks/useNamespace.ts
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
+import { useSelector } from 'react-redux';
+
+import { getActiveNamespace } from '@console/internal/actions/ui';
+import { ALL_NAMESPACES_KEY } from '../constants/common';
+
+export const useNamespace = () => {
```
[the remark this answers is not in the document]
rawagner: removed, using `useActiveNamespace` now

frontend/public/locales/en/bindings.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
rawagner: no, my bad. Removed

frontend/public/components/factory/list-page.tsx
```diff
@@ -281,40 +339,28 @@ export const FireMan_ = connect(null, { filterList })(
...
+> = (resources: FirehoseResourcesResult<F>) => R;
+
+type ListPageProps<L = any, C = any> = PageCommonProps<L, C> & {
+  kind: string;
+  helpText?: React.ReactNode;
+  selector?: Selector;
+  fieldSelector?: string;
+  createHandler?: () => void;
+  name?: string;
+  filters?: any;
```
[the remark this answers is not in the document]
rawagner: that change is from https://github.com/openshift/console/pull/8792 . I rebased as 8792 is already merged. I will take a look at improving the type in another PR

frontend/public/components/machine.tsx
```diff
@@ -47,7 +55,7 @@ const tableColumnClasses = [
 const getMachineProviderState = (obj: MachineKind): string =>
   obj?.status?.providerStatus?.instanceState;
 
-const MachineTableRow: RowFunction<MachineKind> = ({ obj, index, key, style }) => {
+const MachineTableRow: React.FC<any> = ({ obj, index, key, style }) => {
```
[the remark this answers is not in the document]
rawagner: Type updated

frontend/public/components/pod.tsx
rawagner: I agree.
`sort` now accepts `string` too.
I've also updated `sortResourceByValue` helper which now returns a function directly consumable by `Array.sort`

frontend/public/components/factory/Table/VirtualizedTable.tsx
```diff
@@ -0,0 +1,290 @@
...
+  );
+
+  return (
+    <div className="co-m-table-grid co-m-table-grid--bordered">
+      <StatusBox
+        skeleton={<div className="loading-skeleton--table" />}
+        data={data}
+        loaded={loaded}
+        loadError={loadError}
+        // unfilteredData={propData} TODO
```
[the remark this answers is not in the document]
rawagner: yes, TODO is now done :)

frontend/public/components/machine.tsx
rawagner: I forgot to do that. All other columns are changed now as well

frontend/public/components/factory/ListPage/ListPageCreate.tsx
```diff
@@ -0,0 +1,129 @@
...
+export const ListPageCreateDropdown: React.FC<ListPageCreateDropdownProps> = ({
+  items,
+  createAccessReview,
+  children,
+  onClick,
+}) => {
+  return (
+    <CreateWithPermissions createAccessReview={createAccessReview}>
+      <div className="co-m-primary-action">
+        <Dropdown
```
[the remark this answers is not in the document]
rawagner: Updated to PF dropdown. Please note that the change is from https://github.com/openshift/console/pull/8827 which this PR is based on.

rawagner:
FYI new table also has scrolling perf improvement from https://github.com/openshift/console/pull/9797

frontend/public/components/factory/Table/VirtualizedTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+  'aria-label': ariaLabel,
+  gridBreakPoint = TableGridBreakpoint.none,
+  onSelect,
+  Row,
+  activeColumns,
+  columnManagementID,
+  showNamespaceOverride,
+  rowData,
+  unfilteredData,
+}: VirtualizedTableProps<D, R>) => {
```
rawagner, line 147: @spadgett WDYT about typing a component like this instead of `React.FC<FooProps>` ? 
The advantage here is that when we use this component we can specify the generics like
```js
<VirtualizedTable<PodKind, PodRowData>
      {...props}
      aria-label={t('public~Machines')}
      columns={getColumns(showNodes, t)}
      Row={PodTableRow} //Row is typed as  React.ComponentType<RowProps<PodKind, PodRowData>>
      rowData={rowData} //rowData is typed as PodRowData 
    />
```

spadgett: I didn't know you could do this... I think it's more important to be able to specify generics, so this looks better to me.

As an aside, some argue [not to use `React.FC`](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/) in general.

spadgett:
/label px-approved
/label docs-approved

There are no user visible changes, and we have approval for dynamic plugin APIs.

rawagner:
Ok, I added one last change - `useActiveColumns` which loads user settings & filters the table columns based on those settings. The table is handling only column visibility based on the actual size.

The table how has 3 less properties and IMO its more transparent to see whats going on.

yapei:
Checked sorting, filtering, column management and responsive of the table on Pods and Machines list page, looks good
/label qe-approved

frontend/public/components/pod.tsx
```diff
@@ -215,109 +223,110 @@ const podColumnInfo = Object.freeze({
...
-        additional: true,
-      },
-      {
-        title: '',
-        props: { className: Kebab.columnClass },
-      },
-    ];
-  };
-};
+const getColumns = (showNodes: boolean, t: TFunction): TableColumn<PodKind>[] => [
```
spadgett, line 226: IMO, it's generally better to use `i18next.t` rather than passing it as a parameter. Is there a reason that needed to change? (Not all `t`s are the same since `useTranslation` takes arguments, and it means one less parameter you have to pass.)

frontend/public/components/pod.tsx
```diff
@@ -831,46 +839,37 @@ export const PodsDetailsPage: React.FC<PodDetailsPageProps> = (props) => {
...
+        rowData={rowData}
+      />
+    )
   );
-});
+};
 PodList.displayName = 'PodList';
 
-export const getFilters = () => [
+export const getFilters = (t: TFunction): RowFilter<PodKind>[] => [
```
spadgett, line 870: Same question about passing `t` here.

frontend/public/components/pod.tsx
```diff
@@ -887,84 +886,109 @@ export const getFilters = () => [
...
+    namespace,
+    selector,
+    fieldSelector,
+  });
+
+  const [data, filteredData, onFilterChange] = useListPageFilter(pods, undefined, {
+    name: { selected: [nameFilter] },
+  });
+
+  const filters = React.useMemo(() => getFilters(t), [t]);
```
spadgett, line 945: Looks like this won't be reevaluated on language changes due to `useMemo`.

frontend/public/components/pod.tsx
rawagner: Im seeing that the `useTranslation` will return a new instance of `t` when the language is changed. And all useMemo-s etc are correctly re-evaluated. And thats exactly what I expect from the `useTranslation` hook otherwise it wouldn't be responsive .. ?

I want to memoize result of `getColumns` otherwise I'd be passing a new instance of array on every re-render.
If i use `i18next.t` in `getColumns` with useMemo then its not responsive.

spadgett (approved):
/lgtm

PR #8832, opened 2021-04-30
Improve form validation and allow optional resources and workspaces

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5804

**Analysis / Root cause**: 
After importing a Deployment from Git with enabled Pipelines option, the automatically copied Pipeline could not be opened and saved in the Pipeline builder without fixing a validation error. This happen because there are optional fields in this copied Pipeline which are not supported by the Pipeline builder.

The Pipeline builder expected that all defined resources and workspaces of a task are mandatory, but for the resource "source" of the task "openshift-client" there resource defined in the Pipeline:

Screenshot from 2021-05-04 19-48-43

**Solution Description**: 
This change removes the form (formik + yup) "required" check for all resources and workspaces. Instead of a static check it now requires such a link to a Pipeline resource or workspace only if the linked task resource or workspace is mandatory. The form also shows the mandatory asterisk depending on this flag. You can see this change in the first two gifs below.

But while testing this I noticed some other "new" issues with the optional fields:

1. When the user changed a resource or workspace name in the pipeline, the task which refers such a resource or workspace need to be updated similar to required fields. This shows the user which tasks are affected by his change. So that's fine. But there is no option to de-select a resource or workspace again. The dropdown shows just the option "Select resource...." which is disabled and the other resources.
2. When the user removes a resource or workspace which is used in a task, and there is no other resource or workspace defined, the task is in a state where the user could not fix it anymore. The dropdown in the sidebar is completely disabled (because there is nothing to select) and it shows one an error that the user should change it ("Resource name has changed, reselect", or "Workspace name has changed, reselect").

To fix this I made an UI change which:

1. If the previously linked resource name could not be found anymore shows an additional option in the Dropdown which could not be selected anymore.
2. Enable the Dropdown if this option is shown or if there is any other resource is available. (It is still disabled if there is no resource is selected and there is no other resource available.)

Finally if this PR removes not linked resources and workspaces before it sends them to the backend. Otherwise the admission controller don't accept these optional fields.

Origin bug BEFORE this change:
odc-5804-before

Origin bug AFTER this change:
odc-5804-after

FULL recording of the new bahaviour and how the user can solve the issue when she / he removes a resource or workspace now:
option-to-unselect-optional-fields

Notice that the previously selected option automatically disappear as soon as the user switches to another option:
Peek 2021-05-04 20-28

Additional tests for PipelineBuilder `pipeline-builder/validation-utils.ts`:

```
      Validate Resources
        ✓ should pass if the task does not contain the optional resources (7ms)
        ✓ should pass if the task contain the optional resources (9ms)
        ✓ should pass if the task contain the optional resources with an empty string (6ms)
        ✓ should pass if the task contains the required but not the optional resources (5ms)
        ✓ should fail if the task contains the optional but not the required resources (7ms)
      Validate Workspaces
        ✓ should pass if the task does not contain an optional workspaces (6ms)
        ✓ should fail if the task has a optional workspace but miss a required workspaces (7ms)
```

New unit test for `pipeline-builder/utils.ts`

```
  removeEmptyFormFields
    ✓ should not fail without any additional data
    ✓ should drop empty parameter values (1ms)
    ✓ should drop unlinked resources
    ✓ should drop unlinked workspaces
    ✓ should drop all optional or unlinked parameters
```

**Test setup:**
In general:
1. Install the OpenShift Pipelines operator (tested with version 1.4)

For the initial bug:
1. Import a Deployment from git. As repo you can use `https://github.com/jerolimov/nodeinfo` but this should fail also with others. Enable the "Add pipeline" option.
2. Open the Pipelines list
3. Select for your new pipeline the action "Edit pipeline"
4. Try to save this pipeline (previously it should an error and could not be saved)

For additional tests:
1. Create a task with optional resources and an optional workspace:
```yaml
kind: Task
metadata:
  name: example-task
spec:
  resources:
    inputs:
      - name: input-resource
        optional: true
        type: git
    outputs:
      - name: output-resource
        optional: true
        type: image
  steps:
    - command:
        - /bin/bash
        - '-c'
        - echo
        - $(inputs.params.appName)
      image: registry.redhat.io/ubi7/ubi-minimal
      name: ''
      resources: {}
  workspaces:
    - name: a-workspace
      optional: true
```

2. Create a new Pipeline with the Pipeline builder
3. Add a resource / workspace, use them in your task, remove the resource / workspace from the Pipeline
4. Mix this with some more edge cases of adding and removing the data and try to save the pipeline when its possible.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

/assign @andrewballantyne 
/cc @andrewballantyne @karthikjeeyar

### DIFF
--- a/frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
+++ b/frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
@@ -166,6 +166,7 @@
   "Output resources": "Output resources",
   "Display Name": "Display Name",
   "Name is already in use": "Name is already in use",
+  "No {{resourceType}} resource": "No {{resourceType}} resource",
   "Select {{resourceType}} resource...": "Select {{resourceType}} resource...",
   "Only showing resources for this type ({{resourceType}}).": "Only showing resources for this type ({{resourceType}}).",
   "View shortcuts": "View shortcuts",
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
@@ -19,7 +19,7 @@ const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
     availableResources,
     hasResource,
     name,
-    resource: { name: resourceName, type: resourceType },
+    resource: { name: resourceName, type: resourceType, optional = false },
   } = props;
 
   const dropdownResources = availableResources.filter(
@@ -30,7 +30,6 @@ const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
       label: t('pipelines-plugin~Select {{resourceType}} resource...', { resourceType }),
       value: '',
       isPlaceholder: true,
-      isDisabled: true,
     },
     ...dropdownResources.map((resource) => ({ label: resource.name, value: resource.name })),
   ];
@@ -49,7 +48,7 @@ const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
           setFieldValue(name, { name: resourceName, resource: selectedResource });
         }
       }}
-      required
+      required={!optional}
     />
   );
 };
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
@@ -12,9 +12,14 @@ type TaskSidebarResourceProps = {
   resource: TektonResource;
 };
 
+interface ResourceLink {
+  name: string;
+  resource: string;
+}
+
 const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
   const { t } = useTranslation();
-  const { setFieldValue } = useFormikContext<PipelineBuilderFormikValues>();
+  const { getFieldMeta, setFieldValue } = useFormikContext<PipelineBuilderFormikValues>();
   const {
     availableResources,
     hasResource,
@@ -25,14 +30,32 @@ const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
   const dropdownResources = availableResources.filter(
     (resource) => resourceType === resource.type && !!resource.name?.trim(),
   );
+
+  const currentLinkedResourceName = getFieldMeta<ResourceLink>(name).value?.resource;
+  const currentLinkedResourceSelectable =
+    currentLinkedResourceName &&
+    dropdownResources.some((resource) => resource.name === currentLinkedResourceName);
+
   const options: FormSelectFieldOption[] = [
     {
-      label: t('pipelines-plugin~Select {{resourceType}} resource...', { resourceType }),
+      label: optional
+        ? t('pipelines-plugin~No {{resourceType}} resource', { resourceType })
+        : t('pipelines-plugin~Select {{resourceType}} resource...', { resourceType }),
       value: '',
       isPlaceholder: true,
+      isDisabled: !optional,
     },
-    ...dropdownResources.map((resource) => ({ label: resource.name, value: resource.name })),
   ];
+  if (currentLinkedResourceName && !currentLinkedResourceSelectable) {
+    options.push({
+      label: currentLinkedResourceName,
+      value: currentLinkedResourceName,
+      isDisabled: true,
+    });
+  }
+  options.push(
+    ...dropdownResources.map((resource) => ({ label: resource.name, value: resource.name })),
+  );
 
   return (
     <FormSelectField
@@ -42,7 +65,7 @@ const TaskSidebarResource: React.FC<TaskSidebarResourceProps> = (props) => {
         resourceType,
       })}
       options={options}
-      isDisabled={dropdownResources.length === 0}
+      isDisabled={options.length === 1}
       onChange={(selectedResource: string) => {
         if (!hasResource) {
           setFieldValue(name, { name: resourceName, resource: selectedResource });
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
@@ -28,7 +28,6 @@ const TaskSidebarWorkspace: React.FC<TaskSidebarWorkspaceProps> = (props) => {
     {
       label: t('pipelines-plugin~Select workspace...'),
       value: '',
-      isDisabled: true,
       isPlaceholder: true,
     },
     ...dropdownWorkspaces.map((workspace) => ({
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
@@ -12,6 +12,11 @@ type TaskSidebarWorkspaceProps = {
   resourceWorkspace: TektonWorkspace;
 };
 
+interface WorkspaceLink {
+  name: string;
+  workspace: string;
+}
+
 const TaskSidebarWorkspace: React.FC<TaskSidebarWorkspaceProps> = (props) => {
   const {
     availableWorkspaces,
@@ -20,27 +25,44 @@ const TaskSidebarWorkspace: React.FC<TaskSidebarWorkspaceProps> = (props) => {
     resourceWorkspace: { name: workspaceName, optional = false },
   } = props;
   const { t } = useTranslation();
-  const { setFieldValue } = useFormikContext<PipelineBuilderFormikValues>();
+  const { getFieldMeta, setFieldValue } = useFormikContext<PipelineBuilderFormikValues>();
 
   const dropdownWorkspaces = availableWorkspaces.filter((workspace) => !!workspace.name?.trim());
 
+  const currentLinkedWorkspaceName = getFieldMeta<WorkspaceLink>(name).value?.workspace;
+  const currentLinkedWorkspaceSelectable =
+    currentLinkedWorkspaceName &&
+    dropdownWorkspaces.some((resource) => resource.name === currentLinkedWorkspaceName);
+
   const options: FormSelectFieldOption[] = [
     {
-      label: t('pipelines-plugin~Select workspace...'),
+      label: optional
+        ? t('pipelines-plugin~No workspace')
+        : t('pipelines-plugin~Select workspace...'),
       value: '',
       isPlaceholder: true,
+      isDisabled: !optional,
     },
+  ];
+  if (currentLinkedWorkspaceName && !currentLinkedWorkspaceSelectable) {
+    options.push({
+      label: currentLinkedWorkspaceName,
+      value: currentLinkedWorkspaceName,
+      isDisabled: true,
+    });
+  }
+  options.push(
     ...dropdownWorkspaces.map((workspace) => ({
       value: workspace.name,
       label: workspace.name,
     })),
-  ];
+  );
 
   return (
     <FormSelectField
       name={`${name}.workspace`}
       label={workspaceName}
-      isDisabled={dropdownWorkspaces.length === 0}
+      isDisabled={options.length === 1}
       options={options}
       onChange={(selectedWorkspace: string) => {
         if (!hasWorkspace) {
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
@@ -262,16 +262,25 @@ const removeListRunAfters = (task: PipelineTask, listIds: string[]): PipelineTas
   return task;
 };
 
-const removeEmptyDefaultParams = (task: PipelineTask): PipelineTask => {
+export const removeEmptyFormFields = (task: PipelineTask): PipelineTask => {
+  let trimmedTask = task;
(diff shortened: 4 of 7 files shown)


frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
```diff
@@ -176,6 +177,7 @@
   "No When expressions are associated with this task": "No When expressions are associated with this task",
   "Add When Expression": "Add When Expression",
   "Select workspace...": "Select workspace...",
+  "{{workspaceName}} (deleted)": "{{workspaceName}} (deleted)",
```
andrewballantyne, line 180: Wonder if we can reduce these down to `{{itemName}} (deleted)` instead... that way we use one for both. Translations are always going to be the same as both of these are nouns (in English) and thus should translate equally to all languages the same way.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
```diff
@@ -12,28 +12,49 @@ type TaskSidebarResourceProps = {
...
-    ...dropdownResources.map((resource) => ({ label: resource.name, value: resource.name })),
   ];
+  if (currentLinkedResourceName && !currentLinkedResourceSelectable) {
+    options.push({
+      label: t('pipelines-plugin~{{resourceName}} (deleted)', {
+        resourceName: currentLinkedResourceName,
+      }),
+      value: currentLinkedResourceName,
+      isDisabled: true,
+    });
```
andrewballantyne, line 53: This may be a QoL but this is a change in the way we manage the dropdown, thus a feature. I think we need to roll back to `Select...` like we do with everything else. The error under the field informs what happened.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarResource.tsx
```diff
@@ -12,28 +12,49 @@ type TaskSidebarResourceProps = {
...
+  const currentLinkedResourceSelectable =
+    currentLinkedResourceName &&
+    dropdownResources.some((resource) => resource.name === currentLinkedResourceName);
+
   const options: FormSelectFieldOption[] = [
     {
       label: t('pipelines-plugin~Select {{resourceType}} resource...', { resourceType }),
       value: '',
       isPlaceholder: true,
-      isDisabled: true,
```
andrewballantyne, line 33: Disabled when required, not disabled when optional? I think that's better than just enabling the placeholder option.

Maybe we also change the label to indicate optional 🤔   Maybe something like `No {{resourceType}} resource` to mimic what we do in the Start/Trigger Modals.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
```diff
@@ -20,28 +25,43 @@ const TaskSidebarWorkspace: React.FC<TaskSidebarWorkspaceProps> = (props) => {
...
     },
+  ];
+  if (currentLinkedWorkspaceName && !currentLinkedWorkspaceSelectable) {
+    options.push({
+      label: t('pipelines-plugin~{{workspaceName}} (deleted)', {
+        workspaceName: currentLinkedWorkspaceName,
+      }),
+      value: currentLinkedWorkspaceName,
+      isDisabled: true,
+    });
```
andrewballantyne, line 51: Same comment here

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
```diff
@@ -20,28 +25,43 @@ const TaskSidebarWorkspace: React.FC<TaskSidebarWorkspaceProps> = (props) => {
...
+  const currentLinkedWorkspaceName = getFieldMeta<WorkspaceLink>(name).value?.workspace;
+  const currentLinkedWorkspaceSelectable =
+    currentLinkedWorkspaceName &&
+    dropdownWorkspaces.some((resource) => resource.name === currentLinkedWorkspaceName);
+
   const options: FormSelectFieldOption[] = [
     {
       label: t('pipelines-plugin~Select workspace...'),
       value: '',
-      isDisabled: true,
```
andrewballantyne, line 31: And same comment here.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -265,16 +265,25 @@ const removeListRunAfters = (task: PipelineTask, listIds: string[]): PipelineTas
...
+    const inputs = task.resources?.inputs?.filter((resource) => resource.resource);
+    const outputs = task.resources?.outputs?.filter((resource) => resource.resource);
+    output = { ...output, resources: { ...output.resources, inputs, outputs } };
+  }
+  // Drop workspaces which are not linked to an (optional) workspace.
+  if (task.workspaces?.length > 0) {
+    const workspaces = task.workspaces?.filter((workspace) => workspace.workspace);
+    output = { ...output, workspaces };
+  }
+  return output;
```
andrewballantyne, line 286: I think you can revert this change based on my other comments.

andrewballantyne (changes requested):

frontend/packages/pipelines-plugin/locales/en/pipelines-plugin.json
andrewballantyne: This will not be needed (if you agree with my other changes), but other translations will be.

jerolimov:
@bgliwa01 @openshift/team-devconsole-ux As this bugfix changes also the UI, this comes more to a feature change to support optional resources and workspaces. So I think it would be cool if this get an review (and feedback and/or approval) from you. Thanks :pray:

jerolimov:
@bgliwa01 @andrewballantyne I changed this now based on Andrews ideas / feedback:

* "No git resource" if the resource is optional
* Don't add a suffix "(deleted")

So it looks now like this, which is much cleaner I think:

Peek 2021-05-04 21-40

Wdyt?

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
```diff
@@ -265,16 +265,25 @@ const removeListRunAfters = (task: PipelineTask, listIds: string[]): PipelineTas
   return task;
 };
 
-const removeEmptyDefaultParams = (task: PipelineTask): PipelineTask => {
+export const removeEmptyFormFields = (task: PipelineTask): PipelineTask => {
+  let output = task;
```
abhinandan13jan, line 269: nit: You could use a more descriptive naming as we are already dealing with `resource.output` in the same function.
One more question, instead of 
`let output = task ( object pointing)` should we make it
`let trimmedTask = { ...task, params }` in line 273 directly?

karthikjeeyar:
@jerolimov  Optional resources and workspaces works as expected. However I have noticed an issue in the below scenario,
When Task definition has required workspace (output  *) and it is mapped to the pipeline's workspace, but when you remove the pipeline's workspace, then the error message in that task bubble says `Missing workspaces` but in the sidebar it still has the deleted workspace selected but no error message. In this case, shall we show some inline error message eg: `No workspace available` , thoughts?

image

jerolimov:
> In this case, shall we show some inline error message eg: `No workspace available` , thoughts?

@karthikjeeyar For an required field and when the removed workspace is still liniked we should show "Workspace name has changed, reselect".

Thanks for finding this. I will check if I can reproduce and fix this.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
jerolimov: Thanks @abhinandan13jan I changed `output` to `trimmedTask`.

In line 273 I could not change to `let trimmedTask` because I need access to this variable also in the other if statements.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/utils.ts
jerolimov: We need this also if the user selects "No image resource" or "No workspace". This is required because the form task data contains a list of all configurable resources and workspaces. If the user does not select an optional field, this entries must be removed before the API was called. Or did I missed another comment?

jerolimov:
> > In this case, shall we show some inline error message eg: `No workspace available` , thoughts?
> 
> @karthikjeeyar For an required field and when the removed workspace is still linked we should show "Workspace name has changed, reselect" like before.

@andrewballantyne @karthikjeeyar I analysed this issue now. This happen already before if you edit a pipeline and do not touch the field (task dropdown) before you remove the resource or workspace. The reason is that the FormSelectField do not show errors before a field is touched. Maybe we can skip this finding for now and create another issue for this? Wdyt? We can also make a short sync call on this.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/task-sidebar/TaskSidebarWorkspace.tsx
jerolimov: Changed this to `isDisabled: !optional` aka `No {{resourceType}} resource`

Force pushed, PTAL

jerolimov:
Applied feedback from Andrew and Abhi.

UI now disable `Select {{resource}} resource` / `Select workspace` again for required resources/workspaces and do not disable `No {{resource}} resource` / `No workspace` for optional resources/workspaces.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/validation-utils.ts
```diff
@@ -146,15 +147,48 @@ const hasRequiredResources = (
...
+  t: TFunction,
+): TektonWorkspace | false => {
+  // Search the taskPath which is parent of the given path.
+  // If an path like formData.finallyTasks[0].workspaces[0].workspace is given
+  // it returns the path formData.finallyTasks[0]
+  const taskPath = path
+    .split('.')
+    .slice(0, 2)
+    .join('.');
+  const pipelineTask: PipelineTask = _.get(formValues, taskPath);
```
andrewballantyne, line 179: Not the way I did the resources... but seems to be a better way 🤔  👍

andrewballantyne (approved):
Once you rebase to get around the removal of "`t` passing" we should be good to merge this.

karthikjeeyar:
you need a rebase

jerolimov:
> you need a rebase

@karthikjeeyar @andrewballantyne I'm still not sure if we want include this change as there are still some issues I wanted to discuss in the ODC/Pipelines meeting. Or with you if you want. If you both give a thumb up that we will merge the latest solution I'm happy to rebase this.

jerolimov:
I found an issue, don't take a look right now. I will ping you when this PR is ready for a review again. Thx.
/hold

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/validation-utils.ts
```diff
@@ -256,8 +306,7 @@ const resourceDefinition = (formValues: PipelineBuilderFormYamlValues, taskType:
               taskType,
             );
           },
-        )
-        .required(i18n.t('pipelines-plugin~Required')),
+        ),
```
jerolimov, line 309: One of the latest issues was that the workspace was always required, also if the resource is optional.

Instead of this "hard coded" required call, I added a `is-resource-is-required` check above.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/validation-utils.ts
```diff
@@ -351,9 +406,8 @@ const taskValidation = (formValues: PipelineBuilderFormYamlValues, taskType: Tas
                   'is-workspace-link-broken',
                   i18n.t('pipelines-plugin~Workspace name has changed, reselect'),
                   (workspaceValue?: string) =>
-                    !!workspaceValue && !!workspaces.find(({ name }) => name === workspaceValue),
-                )
-                .required(i18n.t('pipelines-plugin~Required')),
+                    !workspaceValue || !!workspaces.find(({ name }) => name === workspaceValue),
+                ),
```
jerolimov, line 410: This here was the other latest fix. This checks passes now if there is no value is defined.

The tests `are-workspaces-available` and `is-workspace-link-broken` only handle cases where a workspaceValue is defined. We have the `is-workspace-is-required` to verify that the user has any workspace defined (for required fields).

jerolimov (commented):
@andrewballantyne @karthikjeeyar thank you for your patience, I rebased and updated the PR. I added also this two comments to show the latest changes:

jerolimov:
FYI: There is one last known issue: If the user opens a Pipeline and remove a workspace or resource, we don't show an error in the task sidebar. The reason is that the `FormSelectField` field doesn't show errors for untouched fields.

Karthik found this issue above. This happen also on master and is a general issue for all our forms. I personally expect that we could fix this only if we touch the task resource when the pipeline resource was touched. If you like I can create another ticket for this?!

andrewballantyne:
> FYI: There is one last known issue: If the user opens a Pipeline and remove a workspace or resource, we don't show an error in the task sidebar. The reason is that the `FormSelectField` field doesn't show errors for untouched fields.
> 
> Karthik found this issue above. This happen also on master and is a general issue for all our forms. I personally expect that we could fix this only if we touch the task resource when the pipeline resource was touched. If you like I can create another ticket for this?!

Please log another issue... this PR doesn't need to solve everything about the Pipeline Builder.

jerolimov:
Follow up: [ODC-5894](https://issues.redhat.com/browse/ODC-5894) Pipeline builder sometimes doesn't show errors in task sidebar after pipeline workspace is removed

karthikjeeyar:
Verified locally, works as expected for optional resources.
/lgtm

andrewballantyne (approved):
For posterity, lgtm as well 🙂

jerolimov:
> For posterity, lgtm as well 🙂

/woof

PR #8835, opened 2021-05-03
Add popover to storage in overview status card

afreen23:
-  The storage dashboard are moved under storage navigation.For accessibility added a popover on storage section enabling linking to dashboards .

- Fixed an issue in control plane component

  - incorrect usage of `healthStateMessage` as an object.

Screenshot from 2021-05-03 12-20-23


afreen23:
@rawagner  PTAL for control plane chnages.
Thanks

PR #8841, opened 2021-05-03
Bug 1956313: Save additional parameters from build guidance container and drop it

jerolimov:
Manual cherry pick of #8807

This was required because 4.8 extracted the `getLimitsData` from `edit-application-utils.ts` into `getLimitsDataFromResource` in `console-shared`.

Chrry picked this PR on `release-4.7` and tested it locally. Could verify that the sample Devfile starts successfully with this change.

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1956313

/kind bug
/cc @andrewballantyne
/assign @christianvogt

### DIFF
--- a/frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
+++ b/frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
@@ -4,12 +4,14 @@ import { useTranslation } from 'react-i18next';
 import { parseDevfile } from '@console/internal/module/k8s';
 import { DevfileSuggestedResources } from '../import-types';
 
+import { getLimitsData } from '../../edit-application/edit-application-utils';
+
 const suffixSlash = (val: string) => (val.endsWith('/') ? val : `${val}/`);
 const prefixDotSlash = (val) => (val.startsWith('/') ? `.${val}` : val);
 
-export const useDefileServer = (
+export const useDevfileServer = (
   values: FormikValues,
-  setFieldValue: (name: string, value: any) => any,
+  setFieldValue: (name: string, value: any, shouldValidate?: boolean) => any,
 ): [boolean, string] => {
   const { t } = useTranslation();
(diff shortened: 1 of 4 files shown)


frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
```diff
@@ -4,12 +4,14 @@ import { useTranslation } from 'react-i18next';
 import { parseDevfile } from '@console/internal/module/k8s';
 import { DevfileSuggestedResources } from '../import-types';
 
+import { getLimitsData } from '../../edit-application/edit-application-utils';
```
andrewballantyne, line 7: Utility moved, that's a conflict for sure. I think there was also a conflict in the test data/tests, but that's less concerning.

andrewballantyne (approved):

frontend/packages/dev-console/src/components/import/devfile/devfileHooks.ts
jerolimov: The conflict was just in `DevfileImportForm.tsx`, but it's true that I changed also the test data a little bit.

There are a strange typecheck in 4.7 [1]. That was the reason why the (not changeable) `serverless` form data from 4.8 [2] was not accepted in 4.7. I changed them to state which was the default state in 4.7 [3].

[1] Trouble making typecheck in 4.7: https://github.com/openshift/console/blob/release-4.7/frontend/packages/dev-console/src/components/import/import-types.ts#L219-L224 (fixed in master)
[2] 4.8 PR: https://github.com/openshift/console/pull/8807/files#diff-0df3addd6c88d6b961d96092db9fd058ba4724a995f6b13704839672c3a2dda4R518-R531
[3] 4.7 now: https://github.com/openshift/console/pull/8841/files#diff-0df3addd6c88d6b961d96092db9fd058ba4724a995f6b13704839672c3a2dda4R524-R531

PR #8843, opened 2021-05-03
Bug 1952487: Template filter does not work properly

avivtur:
fixes also:
Bug 1952737: [RFE]Users had difficulty distinguishing between “ Supported” and “Provided”

in this example, changing select options for template provider filter, to avoid further confusion, adding label next to Template provider in templates page

before:

bz_1952487_before_wizard-5

bz_1952487_before_wizard-3

after (U/S):

bz_1952487_after_US-new

after (D/S):

bz_1952487_after_DS-new

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1952487
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1952737


avivtur:
@matthewcarleton @yaacov please review

matthewcarleton:
@yaacov should we pull out the provided vs support here based on our discussion on Monday? This doesn't seem to align with that.

yaacov:
> @yaacov should we pull out the provided vs support here based on our discussion on Monday? This doesn't seem to align with that.

@matthewcarleton yes, this PR will fix the provided vs support issue, @avivtur can you replace the multiple screenshots in the description with a two short clips, one for upstream and one for downstream ?

avivtur:
@yaacov added clips to commit description

yaacov:
@matthewcarleton PTAL at the latest "after" clips at the description

matthewcarleton:
looking very nice!
a few small comments : 
1. Can we have the label just say "Community" and drop supported for spacing. it will look better especially in the list view.
2. Can the Star icon on the card always be on the far right for consistency
3. I see a template with "New templates" in the provider column. Is that intentional? Sounds strange.

avivtur:
> 1. Can we have the label just say "Community" and drop supported for spacing. it will look better especially in the list view.
> 2. Can the Star icon on the card always be on the far right for consistency
@matthewcarleton Thanks for looking :) 
I have applied your suggestions to the code, you can check it in the description clips
> 3. I see a template with "New templates" in the provider column. Is that intentional? Sounds strange.
I guess it is intentional, no way to really check that since it's ready for customization

matthewcarleton:
> > 1. Can we have the label just say "Community" and drop supported for spacing. it will look better especially in the list view.
> > 2. Can the Star icon on the card always be on the far right for consistency
> 
> @matthewcarleton Thanks for looking :)
> I have applied your suggestions to the code, you can check it in the description clips
> 
> > 1. I see a template with "New templates" in the provider column. Is that intentional? Sounds strange.
> 
> I guess it is intentional, no way to really check that since it's ready for customization
I think we could improve this - could we just use whatever was from the template that is being copied? Seems a bit odd to have "new template" as the name too - is that auto generated?

I'm also seeing RHEL 6 as a community template? Is that right? @yaacov is that because it doesn't have Red Hat Support anymore?

avivtur:
> > > 1. Can we have the label just say "Community" and drop supported for spacing. it will look better especially in the list view.
> > > 2. Can the Star icon on the card always be on the far right for consistency
> > 
> > 
> > @matthewcarleton Thanks for looking :)
> > I have applied your suggestions to the code, you can check it in the description clips
> > > 1. I see a template with "New templates" in the provider column. Is that intentional? Sounds strange.
> > 
> > 
> > I guess it is intentional, no way to really check that since it's ready for customization
> 
> I think we could improve this - could we just use whatever was from the template that is being copied? Seems a bit odd to have "new template" as the name too - is that auto generated?

according to the following code, the user who creates the template, select the template name and template provider name:
https://github.com/openshift/console/blob/master/frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/tabs/vm-settings-tab/vm-settings-tab.tsx#L104-L119
so no auto generated input, only user input

> I'm also seeing RHEL 6 as a community template? Is that right? @yaacov is that because it doesn't have Red Hat Support anymore?

yaacov:
> I'm also seeing RHEL 6 as a community template? Is that right? @yaacov is that because it doesn't have Red Hat Support anymore?

Yes, templates that are not supported by redhat are still community supported

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.scss
```diff
@@ -5,6 +5,6 @@
 
 .kv-create-vm-modal__footer {
   // --pf-global--ZIndex--xs == 100
-  z-index: calc(var(--pf-global--ZIndex--xs) - 4); 
+  z-index: calc(var(--pf-global--ZIndex--xs) - 4) !important; // Fixing Regression for Bug https://bugzilla.redhat.com/show_bug.cgi?id=1930015
```
pcbailey, line 8: Is there no way to fix this by giving higher specificity to this element instead of using `!important`?

frontend/packages/kubevirt-plugin/src/selectors/vm-template/basic.ts
```diff
@@ -65,8 +65,9 @@ export const getTemplateProvider = (
   withProviderPrefix = false,
 ): string => {
   let provider = getAnnotation(template, TEMPLATE_PROVIDER_ANNOTATION);
-  if (!provider && isCommonTemplate(template)) {
-    const isUpstream = window.SERVER_FLAGS.branding === 'okd';
+  const isUpstream = window.SERVER_FLAGS.branding === 'okd';
```
pcbailey, line 68: We have this exact same test in multiple places, including two different locations in this PR. I think we should extract it out to a function somewhere like frontend/packages/kubevirt-plugin/src/utils/common.ts so we can just call it wherever we need it.

pcbailey (commented):
I just have a couple of nitpicky items. I'm going to lgtm, but it would be great if we could fix them in a followup PR.

frontend/packages/kubevirt-plugin/src/components/create-vm-wizard/create-vm-wizard-footer.scss
avivtur: https://github.com/openshift/console/pull/8918

frontend/packages/kubevirt-plugin/src/selectors/vm-template/basic.ts
avivtur: https://github.com/openshift/console/pull/8918

PR #8844, opened 2021-05-03
Bug 1955803: OperatorHub - console accepts any value for infraFeatures

florkbr:
Fixes bug where the console UI is showing separate filters for a custom
catalog resource with supported infrastructure features annotation.
IE: `proxy` vs `proxy-aware` (both should map to the same UI filter)

See https://github.com/openshift/openshift-docs/pull/29511/files?short_path=15febe7#diff-15febe7334e281a679b50f39acc4f828f19f7c0efcd3e5033ed2db30c602a1d0 for new supported values versus old.

Also fixes bug where unsupported annotation values are shown in the UI.
IE: `bacon` 🥓

https://bugzilla.redhat.com/show_bug.cgi?id=1955803

### Before
image
image

### After
image
Screen Shot 2021-05-03 at 12 54 47 PM
Screen Shot 2021-05-03 at 12 58 26 PM
Screen Shot 2021-05-03 at 1 04 52 PM

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx
@@ -27,6 +27,7 @@ import {
   OperatorHubCSVAnnotations,
   InstalledState,
   OperatorHubCSVAnnotationKey,
+  InfraFeatures,
 } from './index';
 import { parseJSONAnnotation } from '@console/shared/src/utils/annotations';
 import { Trans, useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-page.tsx
```diff
@@ -71,12 +72,18 @@ export const OperatorHubList: React.FC<OperatorHubListProps> = ({
...
-          });
+          const [parsedInfraFeatures = [], validSubscription] = ANNOTATIONS_WITH_JSON.map(
+            (annotationKey) => {
+              return parseJSONAnnotation(currentCSVAnnotations, annotationKey, () =>
+                // eslint-disable-next-line no-console
+                console.warn(`Error parsing annotation in PackageManifest ${pkg.metadata.name}`),
+              );
+            },
+          );
+          const filteredInfraFeatures = _.chain(parsedInfraFeatures)
```
spadgett, line 83: We should check whether `_.chain` pulls in all of lodash into our dependencies bundle or if tree shaking still works. Unfortunately, webpack-bundle-analyzer just broke in our repo for us to check. @vojtechszocs is working on a fix for that.

spadgett: The analyze script should be fixed in #8845 (thanks @vojtechszocs !)

florkbr: @spadgett would you prefer if I avoided chain for the time being and used `_.compact` and `_.map` directly until we sort out tree shaking? It looks like we already `import * as _ from 'lodash'` at the top of the file.

florkbr: @spadgett fwiw I ran `yarn run analyze` on:
## My branch + @vojtechszocs branch
Screen Shot 2021-05-03 at 3 29 17 PM

## @vojtechszocs branch only

spadgett: Hey, @florkbr. Yeah, I think it would be better to avoid `_.chain` here, although admittedly I'm not sure if it is a problem or not based on the screenshots there. Do you mind pushing an update?

florkbr: @spadgett the screenshots show that only 1 new call to lodash was added (IE `chain`) resulting in 0.21 KB to be added to the bundle.

I went ahead and updated the logic anyway to avoid chain for now.

spadgett (approved):
/lgtm

Thanks @florkbr !

spadgett:
/cherry-pick release-4.7
to make sure 4.7 also honors the recommended values from the docs

PR #8845, opened 2021-05-03
Bug 1956353: Fix analyze script

vojtechszocs:
This PR fixes issues related to writing and reading a huge (600+ MB) `stats.json` file containing webpack build stats.

- `webpack-bundle-analyzer` logic gets invoked as part of webpack build via `BundleAnalyzerPlugin`
- `analyze.sh` script computes webpack bundle size via `du -b` instead of parsing `stats.json` file
- These two changes allow us to skip `stats.json` file generation, speeding up the `analyze` script.

`webpack-bundle-analyzer` version bumped to latest `3.9.x`. Avoid bumping to `4.x` since those versions are built against webpack 5 (Console application still uses webpack 4 so this is to minimize risk potential).

### DIFF
--- a/analyze.sh
+++ b/analyze.sh
@@ -8,14 +8,14 @@ ARTIFACT_DIR=${ARTIFACT_DIR:=/tmp/artifacts}
 
 cd frontend
 echo "Analyzing Webpack bundles..."
-yarn run analyze --no-open
+yarn run analyze
 if [ -d "$ARTIFACT_DIR" ]; then
   echo "Copying the Webpack Bundle Analyzer report to $ARTIFACT_DIR..."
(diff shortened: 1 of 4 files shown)


analyze.sh
```diff
@@ -8,14 +8,14 @@ ARTIFACT_DIR=${ARTIFACT_DIR:=/tmp/artifacts}
 
 cd frontend
 echo "Analyzing Webpack bundles..."
-yarn run analyze --no-open
```
vojtechszocs, line 11: @spadgett FYI, with this PR, running `yarn analyze` won't open the generated HTML report in default browser.

spadgett (approved):
/lgtm

PR #8847, opened 2021-05-04
adding console.topology extension types to dynamic-plugin-sdk

glekner:


christianvogt:
/hold
documented topology extension API hasn't been updated and reviewed in a while

christianvogt:
/hold cancel

We hope to update the topology API but this current solution will allow kubevirt to create their plugin externally in the interim.

christianvogt:
/lgtm
/approve

this change only adds new types for the migration of topology extension which aren't in use at this time

PR #8852, opened 2021-05-04
WIP: Fix FireMan not finding CRD Models

glekner:
This PR allows deletion of static `ModelDefinition` extensions


spadgett:
/hold

Please update the virtual machines page to use the full group/resource/kind in its URLs. We can add a redirect as needed from the old URL.

We should make sure we have the API group in all URLs. I'm hesitant to merge changes that make it easier to add URLs that do not contain the group.

PR #8853, opened 2021-05-04
Bug 1956920: can't open terminal for pods that have more than one co…

### LINKED ISSUE #8762: can't open terminal for pods that have more than one container running
Hello,

We are running Openshift console on native kubernetes.  When we try to open a terminal to a pod that has more than one container, we can get to a terminal for the default container but when we switch containers we receive the following error:

> Oh no! Something went wrong.
> Description:
> This API only accepts integers

The details box has the following stack trace:

```
vendors~main-chunk-08b0abf6231b0b9835e1.min.js:163997 Error: This API only accepts integers
    at e._verifyIntegers (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:100807)
    at e.refresh (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:100807)
    at Object.onConnectionClosed (303-chunk-8193ff095b2a5fbf4a73.min.js:1)
    at Object.connect_ (303-chunk-8193ff095b2a5fbf4a73.min.js:1)
    at Object.<anonymous> (303-chunk-8193ff095b2a5fbf4a73.min.js:1)
    at Ni (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:163997)
    at Li (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:163997)
    at vendors~main-chunk-08b0abf6231b0b9835e1.min.js:163997
    at t.unstable_runWithPriority (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:164291)
    at li (vendors~main-chunk-08b0abf6231b0b9835e1.min.js:163997)
```

Can anyone suggest a solution?  I'm surprised no one has reported this yet, so it makes me wonder if we're missing something in our setup.  Everything else is working great.  The only issue is opening a terminal to a pod and switching containers.

christianaranda:
…ntainer running

This fix uses a valid range for the refresh function of xterm, starting with row 0 and ending at rows-1.

Invalid start and end parameters were being passed to the refresh function of xterm during the switching of container terminals.  This caused an exception when attempting to switch container terminals.

Per the xterm code:

    /**
     * Tells the renderer to refresh terminal content between two rows
     * (inclusive) at the next opportunity.
     * @param start The row to start from (between 0 and this.rows - 1).
     * @param end The row to end at (between start and this.rows - 1).
     */

Fixes #8762

### DIFF
--- a/frontend/public/components/terminal.jsx
+++ b/frontend/public/components/terminal.jsx
@@ -68,7 +68,7 @@ class Terminal_ extends React.Component {
     terminal.write(`\x1b[31m${reason || 'disconnected'}\x1b[m\r\n`);
     terminal.cursorHidden = true;
     terminal.setOption('disableStdin', true);
-    terminal.refresh(terminal.y, terminal.y);
+    terminal.refresh(0, terminal.rows - 1);
   }
 
   componentDidMount() {


spadgett (approved):
/lgtm

Thank you for the fix!

spadgett:
/lgtm canel

Hi, @christianaranda. It looks like you need to run prettier to format the file. You should be able to use `yarn run prettier --write <filename>` and then amend your commit.

spadgett (approved):
/lgtm

I pushed the prettier fix.

PR #8855, opened 2021-05-04
give feedback when modal submit is clicked

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5799

**Root analysis:**
The submit status information is not sent to the modal submit footer

**Solution description:**
The submit status information is sent to the `submitDisabled` & `inProgress` props of the ModalSubmitFooter

**GIF:**
modal-submit

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/components/modals/DeleteResourceModal.tsx
+++ b/frontend/packages/console-shared/src/components/modals/DeleteResourceModal.tsx
@@ -63,7 +63,7 @@ const DeleteResourceForm: React.FC<FormikProps<FormikValues> & DeleteResourceMod
       </ModalBody>
       <ModalSubmitFooter
         submitText={submitLabel}
-        submitDisabled={(status && !!status.submitError) || !isValid}
+        submitDisabled={(status && !!status.submitError) || !isValid || isSubmitting}
         cancel={cancel}
         inProgress={isSubmitting}
         submitDanger
@@ -78,19 +78,18 @@ class DeleteResourceModal extends PromiseComponent<
   DeleteResourceModalState
 > {
   private handleSubmit = (values, actions) => {
-    actions.setSubmitting(true);
     const { onSubmit, close, redirect } = this.props;
-    onSubmit &&
+    return (
+      onSubmit &&
       this.handlePromise(onSubmit(values))
         .then(() => {
-          actions.setSubmitting(false);
           close();
           redirect && history.push(redirect);
         })
         .catch((errorMessage) => {
-          actions.setSubmitting(false);
           actions.setStatus({ submitError: errorMessage });
-        });
+        })
+    );
   };
 
   render() {


frontend/packages/console-shared/src/components/modals/DeleteResourceModal.tsx
```diff
@@ -78,12 +80,12 @@ class DeleteResourceModal extends PromiseComponent<
   DeleteResourceModalState
 > {
   private handleSubmit = (values, actions) => {
-    actions.setSubmitting(true);
+    actions.setStatus({ isSubmitting: true });
```
rohitkrai03, line 83: This is just a hack to make this work and not a good way to set submitting status. Formik handles submitting status for you automatically if the submit handler is async and returns a promise. You can read more about why this doesn't work right now in this [code comment](https://github.com/formium/formik/blob/master/packages/formik/src/Formik.tsx#L748)

frontend/packages/console-shared/src/components/modals/DeleteResourceModal.tsx
```diff
@@ -63,9 +63,11 @@ const DeleteResourceForm: React.FC<FormikProps<FormikValues> & DeleteResourceMod
       </ModalBody>
       <ModalSubmitFooter
         submitText={submitLabel}
-        submitDisabled={(status && !!status.submitError) || !isValid}
+        submitDisabled={
+          (status && !!status.submitError) || !isValid || status?.isSubmitting || isSubmitting
+        }
         cancel={cancel}
-        inProgress={isSubmitting}
+        inProgress={status?.isSubmitting || isSubmitting}
```
rohitkrai03, line 70: This change won't be needed if formik's status handling works correctly.

rohitkrai03 (changes requested):

christianvogt:
gif looks good!

frontend/packages/console-shared/src/components/modals/DeleteResourceModal.tsx
```diff
@@ -80,7 +80,8 @@ class DeleteResourceModal extends PromiseComponent<
   private handleSubmit = (values, actions) => {
     actions.setSubmitting(true);
     const { onSubmit, close, redirect } = this.props;
-    onSubmit &&
+    return (
+      onSubmit &&
       this.handlePromise(onSubmit(values))
         .then(() => {
           actions.setSubmitting(false);
```
rohitkrai03, line 87: Same here.

rohitkrai03 (changes requested):

rohitkrai03 (changes requested):
In progress loading animation is shown but the submit button is not disabled. Can you also add the logic to disable the submit button while form is submitting?

rohitkrai03 (approved):
/approve
/lgtm

PR #8860, opened 2021-05-06
Bug 1954105: Update Taskrun tab to make namespace based calls

abhinandan13jan:
# Issue
https://bugzilla.redhat.com/show_bug.cgi?id=1954105
https://issues.redhat.com/browse/OCPBUGSM-28361

# Problem
The TaskRun tab requests TaskRun list with selector over whole cluster

# Solution
Make the TaskRun call namespace scoped

# Screenshot
Created two PipelineRuns with same name in two different namespaces `abhi` and `abhi2`. As we see in the screenshot, the list only displays the TaskRuns from the desired namespace.

Screenshot from 2021-05-05 19-52-42
Screenshot from 2021-05-05 19-52-50
Screenshot from 2021-05-05 19-52-31
Screenshot from 2021-05-05 19-52-21

# Tests
None Changed

# Browser conformance
Chrome

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsListPage.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsListPage.tsx
@@ -16,10 +16,10 @@ const TaskRunsListPage: React.FC<Omit<
   React.ComponentProps<typeof ListPage>,
   'canCreate' | 'kind' | 'ListComponent' | 'rowFilters'
 > &
-  TaskRunsListPageProps> = ({ hideBadge, showPipelineColumn = true, ...props }) => {
+  TaskRunsListPageProps> = ({ hideBadge, showPipelineColumn = true, namespace, ...props }) => {
   const searchParams = getURLSearchParams();
   const kind = searchParams?.kind;
-  const badge = usePipelineTechPreviewBadge(props.namespace);
+  const badge = usePipelineTechPreviewBadge(namespace);
   return (
     <ListPage
       {...props}
@@ -29,6 +29,7 @@ const TaskRunsListPage: React.FC<Omit<
       ListComponent={TaskRunsList}
       rowFilters={taskRunFilters}
       badge={hideBadge ? null : badge}
+      namespace={namespace}
     />
   );
 };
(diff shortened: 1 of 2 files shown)


frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsListPage.tsx
```diff
@@ -16,19 +16,20 @@ const TaskRunsListPage: React.FC<Omit<
   React.ComponentProps<typeof ListPage>,
   'canCreate' | 'kind' | 'ListComponent' | 'rowFilters'
 > &
-  TaskRunsListPageProps> = ({ hideBadge, showPipelineColumn = true, ...props }) => {
+  TaskRunsListPageProps> = ({ hideBadge, showPipelineColumn = true, namespace, selector }) => {
```
invincibleJai, line 19: Don't we need to accept all other props and pass them down to `ListPage`

abhinandan13jan: Correct me if I got this wrong. As per all the props available to `TaskRunListPage`, `{...props}` only has one prop `selector`. After the changes, I see that there's no use keeping ...props for only one prop `selector` so I destructured it,

abhinandan13jan: updated and brought back ...props
Screenshot from 2021-05-12 21-55-01

frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsListPage.tsx
```diff
@@ -16,19 +16,20 @@ const TaskRunsListPage: React.FC<Omit<
...
     <ListPage
-      {...props}
       customData={{ showPipelineColumn }}
       canCreate={kind?.includes(referenceForModel(TaskRunModel)) ?? false}
       kind={referenceForModel(TaskRunModel)}
       ListComponent={TaskRunsList}
       rowFilters={taskRunFilters}
       badge={hideBadge ? null : badge}
+      namespace={namespace}
+      {...props}
```
andrewballantyne, line 32: Spread should go before, not after, static props. You are "overriding" the spread props, not hoping the spread doesn't override you. It's confusing to see a static value and it got magically overridden because the calling component decided to pass an additional prop you weren't expecting.

Undo the location change.

andrewballantyne (changes requested):

andrewballantyne:
/retitle Bug 1954105: Update Taskrun tab to make namespace based calls

@abhinandan13jan please don't forget to update your PR with the BZ numbers otherwise it won't follow the status of the PR.

andrewballantyne (approved):
/lgtm

frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsListPage.tsx
abhinandan13jan: Thanks for pointing it out as well as updating the title. I've rectified the changes

PR #8861, opened 2021-05-06
Fix Dev perspective nav items

glekner:
This PR fixes the missing nav items in the resources section in the dev perspective


jerolimov (approved):
Thanks, tested it locally and it works fine :+1:

/lgtm

rohitkrai03 (approved):
/approve

PR #8862, opened 2021-05-06
Yaml editor with connected form fields

metalice:

yaml-editor


frontend/packages/kubevirt-plugin/src/components/form-with-editor/FormWithEditor.tsx
```diff
@@ -0,0 +1,115 @@
+import cn from 'classnames';
+import yamlParser from 'js-yaml';
+import { cloneDeep, get, set } from 'lodash';
+import * as React from 'react';
```
yaacov, line 4: are you sure the import order currect ?

PR #8863, opened 2021-05-06
fix(OCS dashboard): s/namesapce/namespace/ prometheus query typos

tumido:
Digging through the OCS queries I've noticed a typo here - `namesapce` is a nonsense. Neither of those queries has such label. I think this should have been a `namespace` which is a valid label. On the other hand  I'm not sure how much is this `on()` property here even relevant or important, because the queries were left like this for long long time.


afreen23:
@anmolsachan  PTAL
I think we recently updated some of them as the bug fix.
Thanks

anmolsachan (approved):

PR #8864, opened 2021-05-06
Bug 1957756: Device Replacemet UI, The status of the disk is "replacement ready" before I clicked on "start replacement"

afreen23:
Added check for same node and same disk name to avoid collisions.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1957756

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
@@ -36,6 +36,7 @@ import {
   NodesDisksListPage,
   NodesDisksListPageProps,
 } from '@console/local-storage-operator-plugin/src/components/disks-list/disks-list-page';
+import { getAnnotations } from '@console/shared/src';
 import { OCSKebabOptions } from './ocs-kebab-options';
 import { OCSStatus } from './ocs-status-column';
 import {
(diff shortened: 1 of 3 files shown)


frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -181,9 +184,8 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
 
     if (tiLoaded && !tiLoadError && tiData.length) {
       const newData: OCSDiskList = tiData.reduce((data, ti) => {
-        const disk = ti.metadata.annotations?.disk;
-        const osd = ti.metadata.annotations?.osd;
-        if (osd && disk) {
+        const { disk, node, osd } = getAnnotations(ti) || {};
+        if (osd && disk && node === nodeName && ocsState.metricsMap?.[disk]?.osd === osd) {
```
GowthamShanmugam, line 188: If we know the nodeName we can watch only that node using node name in https://github.com/openshift/console/pull/8864/files#diff-2fe4b3881950e5603a824223a9a6a00e3955713e1ee8d6ece63aff914590257aR142,

GowthamShanmugam, line 188: oh ok. it showing status from some other nodes TemplateInstance, you need some property to identify TemplateInstance from other nodes to avoid a collision. that is why you nodeName annotation in TemplateInstance.

bipuladh (approved):
/lgtm

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
afreen23: Yes!

PR #8865, opened 2021-05-06
CONSOLE-1338: Add support for multiple documents when importing YAML

jcaianirh:
For: https://issues.redhat.com/browse/CONSOLE-1338

Add support for dragging/editing multiple documents in the yaml import page. For a single yaml file, there is no change in behavior in the import. For multiple yaml files, the following is done:

- client side validation reported to the user: basic validation to make sure the most important yaml content is well formed and ready to go to the server and errors are reported to the user on the import yaml page
- server side validation reported to the user: the resources are sent to the server with the dryRun flag and errors are reported to the user on the import yaml page
- send yaml to server for actual creation
- report status on new import yaml results page with links to new resources
- if any errors occurred, then the try again button will appear and allow the user to retry all failed resources by stepping back to the import page, with the original yaml content displayed for those failed resources for modification 

Example for:
- 1 stateful set
- 2 secrets in ns jcaiani
- 4 secrets in ns: test-namespace
- 1 config map
- 1 deployment

multi-yaml-800

Todo: unit/e2e tests
cc: @bmignano


frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+      <>
+        <GreenCheckCircleIcon size="lg" />
+        <h2>{t('public~Resources successfully created')}</h2>
+      </>
+    );
+  } else {
+    StatusBlock = (
+      <>
+        <YellowExclamationTriangleIcon size="lg" />
+        <h2>{t('public~One or more resources failed to create')}</h2>
```
spadgett, line 55:
```suggestion
        <h2>{t('public~One or more resources failed to be created')}</h2>
```

frontend/public/components/import-yaml.tsx
```diff
@@ -12,10 +13,15 @@ export const ImportYamlPage = () => {
       </Helmet>
       <AsyncComponent
         loader={() => import('./droppable-edit-yaml').then((c) => c.DroppableEditYAML)}
+        allowMultiple={true}
```
spadgett, line 16:
```suggestion
        allowMultiple
```

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+  models,
+}) => {
+  const { t } = useTranslation();
+  const [rows, setRows] = React.useState(
+    location.state.objs.map((obj) => ({
+      obj: { ...obj },
+      status: { creating: true, message: t('public~Creating') },
+    })),
+  );
+  const [sending, setSending] = React.useState(true);
```
spadgett, line 104: What is "sending"? Does this mean the requests are in flight? I'd probably use `inFlight`

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+    const requests = createRequests(location.state.objs, models);
+    Promise.allSettled(requests).then(() => {
+      history.replace({ state: { ...location.state, done: true } });
+      setSending(false);
+    });
+    requests.forEach((resourceRequest, index) => {
+      resourceRequest
+        .then((resourceObj) =>
+          setRows((prevRows) =>
+            // eslint-disable-next-line max-nested-callbacks
```
spadgett, line 125: It would be better to work to flatten the code so we don't need to disable this linter rule.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+            prevRows.map((resourceRow) =>
+              resourceObj?.metadata?.name === resourceRow.obj.metadata.name &&
+              resourceObj?.kind === resourceRow.obj.kind &&
+              resourceObj?.metadata?.namespace === resourceRow.obj.metadata.namespace
```
spadgett, line 129: This is not guaranteed to be unique since we can have resources with the same kind in different API groups.

Instead of storing rows in state and trying to update them, can we simply generate the correct rows on render? I think this would be simpler.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+      }
+      return getTableCell(b, index).localeCompare(getTableCell(a, index));
+    });
+    setRows(updatedRows);
+  };
+
+  const onRetry = () => {
+    const failedResourceObjects = rows
+      .filter((row) => row.status.error)
+      .map((rowObject) => rowObject.obj);
```
spadgett, line 177: This feels backwards. We shouldn't be looking at the table data to find out what's failed. We should track the status of each object in state and generate the table from the status on render.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+    </span>
+  );
+};
+
+const IMPORT_COLS = {
+  NAME: 0,
+  NAMESPACE: 1,
+  STATUS: 2,
+};
+const COLUMN_MAP = ['metadata.name', 'metadata.namespace', 'message'];
```
spadgett, line 84: IMO, it would be better to have something like

```ts
const columns = [{
  label: 'Name',
  path: 'metadata.name',
}, {
  label: 'Namespace',
  path: 'metadata.namespace',
}
// ...
];
```

Then we iterate over the columns when generating the header and the table content. That way the column headers and column map can never get out of sync.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+                    },
+                  }
+                : resourceRow,
+            ),
+          );
+        });
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+  const onSort = (event, index, direction) => {
```
spadgett, line 161: I'm tempted to ask you to remove the sort. It seems to be adding a lot of complexity, and I don't see a lot of value to the user. I'd expect to rarely have more than 10 items.

@bmignano How important do you think sorting is?

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const [rows, setRows] = React.useState(
+    location.state.objs.map((obj) => ({
+      obj: { ...obj },
+      status: { creating: true, message: t('public~Creating') },
+    })),
+  );
+  const [sending, setSending] = React.useState(true);
+  const [errors, setErrors] = React.useState(false);
+  const [activeSortIndex, setActiveSortIndex] = React.useState(-1);
+  const [activeSortDirection, setActiveSortDirection] = React.useState(null);
```
spadgett, line 107: Since we're initializing state with `null`, TypeScript won't be able to infer a type. Let's declare a type here.

frontend/public/components/import-yaml-results.tsx
spadgett: It looks like a lot of functions are missing types. Can you make sure we use types consistently throughout?

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,299 @@
...
+  Tr,
+} from '@patternfly/react-table';
+
+import { history, ResourceLink } from './utils';
+import { createRequests } from './edit-yaml';
+import { K8sKind, K8sResourceKindReference } from '../module/k8s';
+import { formatNamespacedRouteForResource } from '@console/shared/src/utils';
+
+const reactPropFix = {
+  translate: 'no',
```
spadgett, line 30: Can you describe why this is needed?

spadgett (commented):
@jcaianirh We shouldn't be changing the URL or passing data through URL query parameters. Let's see if we can't simplify the table rendering. State could be as simple as a key for each obj mapped to the status. Then we can generate the table by iterating over the list of objects each render. I don't think much else needs to be in state, particularly if we remove sorting.

jcaianirh:
@spadgett thanks for the review, will take a look at these.

frontend/public/components/import-yaml-results.tsx
bmignano: @spadgett Agree sorting is probably not necessary here (I assumed we got it for free with the table). If it's adding lots of complexity, let's remove it!

frontend/public/components/app-contents.tsx
```diff
@@ -682,6 +682,15 @@ const AppContents: React.FC<{}> = () => {
 
             {inactivePluginPageRoutes}
 
+            <LazyRoute
+              path="/importyamlresults"
```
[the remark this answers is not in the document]
jcaianirh: Since we agreed that we don't need to code to design and its ok to have the project selector here, we can reduce this functionality into a component that resides on the existing import page instead of a new page.

frontend/public/components/edit-yaml.jsx
```diff
@@ -215,9 +236,16 @@ export const EditYAML_ = connect(stateToProps)(
...
           }
 
-          const yaml = this.convertObjToYAMLString(obj);
-
+          let yaml = this.convertObjToYAMLString(obj);
           this.displayedVersion = _.get(obj, 'metadata.resourceVersion');
+
+          if (this.props?.location?.state?.yamlObjects) {
+            const yamlTextData = [];
+            this.props.location?.state?.yamlObjects.forEach((yamlObj) => {
```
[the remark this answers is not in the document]
jcaianirh: stale code here, now coming from callbacks and state

frontend/public/components/edit-yaml.jsx
```diff
@@ -264,47 +292,46 @@ export const EditYAML_ = connect(stateToProps)(
...
+            //catch these individually so we can report out all errors
+            dryRunRequests.forEach((request, i) =>
+              request.catch((error) => this.handleErrors(objs[i], error?.message)),
+            );
+            Promise.all(dryRunRequests).then(() => {
+              this.setState({ success: true, error: null, sending: false, displayResults: true });
+              history.push({
+                pathname: '/importyamlresults',
+                state: { objs, done: false },
+              });
```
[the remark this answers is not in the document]
jcaianirh: Switching over to a component within import page, stale code.

frontend/public/components/droppable-edit-yaml.tsx
```diff
@@ -31,56 +32,81 @@ export const DroppableEditYAML = withDragDropContext<DroppableEditYAMLProps>(
...
       };
       this.handleFileDrop = this.handleFileDrop.bind(this);
     }
 
-    handleFileDrop(item, monitor) {
-      if (!monitor) {
-        return;
-      }
-      const [file] = monitor.getItem().files;
+    calculateUploadState(fileUpload: string, input: string) {
```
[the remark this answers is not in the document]
jcaianirh: I think you meant: `addDocument(existingEditorContent: string, newFileContent: string) {`...

frontend/public/components/droppable-edit-yaml.tsx
```diff
@@ -31,56 +32,81 @@ export const DroppableEditYAML = withDragDropContext<DroppableEditYAMLProps>(
...
       if (file.size <= maxFileUploadSize) {
         const reader = new FileReader();
         reader.onload = () => {
           const input = reader.result as string;
           if (containsNonPrintableCharacters(input)) {
-            this.setState({
-              error: fileTypeErrorMsg,
-            });
+            this.setState((previousState) => ({
+              errors: [...previousState.errors, `Ignoring ${file.name}: ${fileTypeErrorMsg}\n`],
```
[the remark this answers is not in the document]
jcaianirh: done. And yes, the errors get cleared in the function handleFileDrop (on file upload)

frontend/public/components/import-yaml.tsx
jcaianirh: done and removing location, no longer needed here

frontend/public/components/import-yaml-results.tsx
jcaianirh: done

frontend/public/components/import-yaml-results.tsx
jcaianirh: removed sorting

frontend/public/components/edit-yaml.jsx
```diff
@@ -264,47 +292,46 @@ export const EditYAML_ = connect(stateToProps)(
...
-            this.handleError(t('yaml~Error parsing YAML: {{e}}', { e }));
-            return;
-          }
+        validateYAML(obj) {
+          const { models, t } = this.props;
 
           if (!obj.apiVersion) {
-            this.handleError(t('yaml~No "apiVersion" field found in YAML.'));
-            return;
+            return t('yaml~No "apiVersion" field found in YAML.');
```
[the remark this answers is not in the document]
jcaianirh: done

frontend/public/components/edit-yaml.jsx
```diff
@@ -264,47 +292,46 @@ export const EditYAML_ = connect(stateToProps)(
...
-        save() {
-          const { onSave, t } = this.props;
-          const { owner } = this.state;
-          let obj;
-
-          if (onSave) {
-            onSave(this.getEditor().getValue());
-            return;
-          }
+        updateAllYAML(objs) {
```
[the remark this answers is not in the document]
jcaianirh: @spadgett  tere are use cases that will be applied to existing single instance or new multi instance functions, and I just coopted the existing names for the new functions (ie updateYAML vs updateAllYAML)...I can change this if you like though.

frontend/public/components/import-yaml-results.tsx
jcaianirh: @spadgett added explanation comment taken from pipelines code.

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +419,51 @@ export const EditYAML_ = connect(stateToProps)(
...
+              const validationError = this.validateYAML(obj);
+              if (validationError) {
+                noErrors = false;
+                this.handleErrors(obj, validationError);
+              }
+            });
+            if (noErrors) {
+              //Check for duplicate name/kinds
+              const uniqueEntries = _.uniqBy(objs, (obj) =>
+                [obj.metadata.name, obj.metadata?.namespace, obj.kind].join(),
```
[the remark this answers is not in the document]
jcaianirh: Added check for API group, and updated the error message to reflect the change.

frontend/public/components/import-yaml-results.tsx
jcaianirh: changed the way to index the data. we are generating the rows on render, but maybe the naming of the state was implying otherwise. updated the state names for better readability as well.

frontend/public/components/import-yaml-results.tsx
jcaianirh: we are actually looking at the status of each object, and generating the table from the status on render, but the naming wasn't indicating that. changed the naming scheme of the state and subsequent methods reacting to state changes

frontend/public/components/import-yaml-results.tsx
jcaianirh: this is no longer needed without sorting.

frontend/public/components/import-yaml-results.tsx
jcaianirh: flattened this and removed the lint disable statement

frontend/public/components/edit-yaml.jsx
```diff
@@ -85,18 +106,20 @@ export const EditYAML_ = connect(stateToProps)(
...
-
         handleError(error) {
           this.setState({ error, success: null });
         }
 
+        handleErrors(resourceObject, error) {
+          const resourceName = resourceObject?.metadata?.name;
+          this.setState((prevState) => ({
+            error: [
+              ...(prevState?.error ? prevState.error.split(', ') : []),
```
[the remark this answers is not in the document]
jcaianirh: stored errors array in state removed split and added a join to display commas between messages

frontend/public/components/edit-yaml.jsx
```diff
@@ -79,24 +102,31 @@ export const EditYAML_ = connect(stateToProps)(
...
+        handleErrors(resourceObject, error) {
+          const resourceName = resourceObject?.metadata?.name;
+          this.setState((prevState) => {
+            const errors = [
+              ...prevState.errors,
+              resourceName ? `${resourceName}: ${error}` : error,
+            ];
+            return {
+              errors,
+              error: errors.join(', '),
```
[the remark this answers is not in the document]
spadgett: We probably want to separate with newlines or blank lines rather than commas.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+  React.useEffect(() => {
+    const requests = createRequests(importResources, models);
+    Promise.allSettled(requests).then(() => {
+      setInFlight(false);
+    });
+    requests.forEach((resourceRequest, index) => {
+      resourceRequest
+        .then((resourceObj) =>
+          setResources((prevResources) => [
+            ...prevResources.slice(0, index),
```
spadgett, line 106: I would just copy the array and set the one value in the new array.

```ts
const nextResources = [...prevResources];
nextResources[index] = { ... };
return nextResources;
```

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+          )}
+        </div>
+      </Bullseye>
+    </div>
+  );
+};
+
+type ImportYAMLStatusProps = {
+  errors?: boolean;
+  sending: boolean;
```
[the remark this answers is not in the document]
jcaianirh: I actually did change this name, but didn't update the type. Thanks for catching this.

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +444,58 @@ export const EditYAML_ = connect(stateToProps)(
...
+          let noErrors = true;
+          this.setState({ error: null, errors: [] }, () => {
+            try {
+              objs = safeLoadAll(this.getEditor().getValue());
+            } catch (e) {
+              this.handleError(t('public~Error parsing YAML: {{e}}', { e }));
+              return;
+            }
+            if (objs.length === 1) {
+              return this.save();
```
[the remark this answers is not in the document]
jcaianirh: done

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +444,58 @@ export const EditYAML_ = connect(stateToProps)(
...
+            });
+            if (noErrors) {
+              //Check for duplicate name/kinds
+              const uniqueEntries = _.uniqBy(objs, (obj) =>
+                [
+                  obj.metadata.name,
+                  obj.metadata.namespace,
+                  obj.kind,
+                  groupVersionFor(obj.apiVersion).group,
+                ].join(),
```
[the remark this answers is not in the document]
jcaianirh: thanks, adding a '-' to separate them, foobar-baz !== foo-barbaz

frontend/public/components/_import-yaml-results.scss
```diff
@@ -0,0 +1,25 @@
+// Bootstrap Core variables and mixins
+@import '~bootstrap-sass/assets/stylesheets/bootstrap/variables';
+
+.co-import-yaml-status {
```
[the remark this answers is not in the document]
jcaianirh: @spadgett i think this block class is valid bem, but i'll alpha the classes and fix the ID issue you mentioned below as well.

frontend/public/components/_import-yaml-results.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin-bottom: 20px;
+}
+
+#import-yaml-results-page {
```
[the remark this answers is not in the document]
jcaianirh: restructured the classes and removed rules.

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +444,58 @@ export const EditYAML_ = connect(stateToProps)(
...
+              this.handleError(t('public~Error parsing YAML: {{e}}', { e }));
+              return;
+            }
+            if (objs.length === 1) {
+              return this.save();
+            } else if (objs.length === 0) {
+              return;
+            }
+            //Run through client side validation for all resources
+            _.each(objs, (obj) => {
```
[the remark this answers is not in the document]
jcaianirh: used foreach instead.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+ * This is the best that I can find relates to this value:
+ * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3423b4fc3e3da09f8acc386bc2fee6fb8f5e0880/types/react/index.d.ts#L1763
+ */
+const reactPropFix = {
+  translate: 'no',
+};
+
+const ImportYAMLPageStatus: React.FC<ImportYAMLStatusProps> = ({ errors, sending }) => {
+  const { t } = useTranslation();
+  let StatusBlock;
```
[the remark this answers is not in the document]
jcaianirh: added type React.ReactNode here

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+        <YellowExclamationTriangleIcon size="lg" />
+        <h2>{t('public~One or more resources failed to be created')}</h2>
+      </>
+    );
+  }
+  return <div className="co-import-yaml-status text-center">{StatusBlock}</div>;
+};
+
+const ImportYAMLResourceStatus: React.FC<ImportYAMLResourceStatusProps> = ({ status }) => {
+  let StatusIcon;
```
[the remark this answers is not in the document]
jcaianirh: added type React.ReactNode here

frontend/public/components/edit-yaml.jsx
```diff
@@ -264,47 +292,71 @@ export const EditYAML_ = connect(stateToProps)(
...
+                  error: null,
+                  errors: [],
+                  sending: false,
+                  resourceObjects: objs,
+                });
+                this.showResults(true);
+              })
+              .catch((error) => {
+                // eslint-disable-next-line no-console
+                console.error(error);
```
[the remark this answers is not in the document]
jcaianirh: Remove the log, and add a comment here that we've already published the error to the user in the forEach catch, but keep the empty catch so we're not firing uncaught message to the console.

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +444,58 @@ export const EditYAML_ = connect(stateToProps)(
...
             }
           }
+          this.updateYAML(obj, getModel(models, obj), newNamespace, newName);
+        }
 
-          this.updateYAML(obj, model, newNamespace, newName);
+        saveAll() {
+          const { t } = this.props;
+          let objs;
+          let noErrors = true;
```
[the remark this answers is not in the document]
jcaianirh: removed the double negative, flipped logic.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+export const ImportYAMLResults: React.FC<ImportYAMLResultsProps> = ({
+  importResources,
+  models,
+  retryFailed,
+  showResults,
+}) => {
+  const { t } = useTranslation();
+  const [resources, setResources] = React.useState(
+    importResources.map((obj) => ({
+      obj: { ...obj },
```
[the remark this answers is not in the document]
jcaianirh: yeah i think you're right...we don't really need the obj....we could use the prop for all that content, and then just keep the state...i'll rewrite this.

frontend/public/components/import-yaml-results.tsx
jcaianirh: yes this is a bit simpler. thanks for the tip

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+                  <Td {...reactPropFix}>
+                    <ResourceLink
+                      kind={resource.obj.kind}
+                      name={resource.obj.metadata.name}
+                      namespace={resource.obj.metadata.namespace}
+                      linkTo={!inFlight && !resource.status.error}
+                    />
+                  </Td>
+                  <Td {...reactPropFix}>
+                    <ResourceLink kind="Namespace" name={resource.obj.metadata.namespace} />
```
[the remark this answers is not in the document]
jcaianirh: Yeah good point. I updated this to show the '-' in the namespace column for non namespaced resources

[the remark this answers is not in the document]
jcaianirh: Screen Shot 2021-05-21 at 3 37 08 PM

frontend/public/components/edit-yaml.jsx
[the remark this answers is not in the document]
jcaianirh: removed original error, and used only errors with a blank line join instead of comma.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { ResourceLink } from './utils';
+import { createRequests } from './edit-yaml';
+import { K8sKind } from '../module/k8s';
+
+/**
+ * Without this prop our current TS types fail to match and require a `translate` prop to be added. PF suggests we
+ * update our types, but that causes other issues. This will have to do as a workaround for now.
+ *
+ * This is the best that I can find relates to this value:
+ * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3423b4fc3e3da09f8acc386bc2fee6fb8f5e0880/types/react/index.d.ts#L1763
```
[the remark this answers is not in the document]
jcaianirh: @spadget this was an issue with updating one of our types packages, i may be wrong, but i thought we pushed that out in our tech debt story about keeping packages up to date.

frontend/public/components/edit-yaml.jsx
```diff
@@ -50,19 +51,40 @@ const stateToProps = ({ k8s, UI }) => ({
...
 });
 
+const getModel = (models, obj) => {
+  if (_.isEmpty(obj) || !models) {
+    return null;
+  }
+  return models.get(referenceFor(obj)) || models.get(obj.kind);
+};
+
+export const createRequests = (objs, models, isDryRun) => {
```
spadgett, line 61: nit: The name here is a bit confusing since this actually makes the HTTP requests. It kind of sounds like it's creating the request body, but not firing the request. Maybe

```suggestion
export const createResources = (objs, models, isDryRun) => {
```

frontend/public/components/edit-yaml.jsx
```diff
@@ -79,22 +101,28 @@ export const EditYAML_ = connect(stateToProps)(
...
         }
 
-        handleError(error) {
-          this.setState({ error, success: null });
+        handleErrors(resourceObject, error) {
+          const resourceName = resourceObject?.metadata?.name;
+          this.setState((prevState) => {
+            const errors = [
+              ...(prevState.errors || []),
+              resourceName ? `${resourceName}: ${error}` : error,
```
spadgett, line 120: Do we need kind+name here since resource name is not unique? Or is this normally in the error message (in which case, maybe we don't need the prefix)?

frontend/public/components/edit-yaml.jsx
```diff
@@ -264,47 +289,69 @@ export const EditYAML_ = connect(stateToProps)(
           });
         }
 
-        save() {
-          const { onSave, t } = this.props;
-          const { owner } = this.state;
-          let obj;
+        createResources(objs) {
```
spadgett, line 292: OK I see why you used `createRequests` above, but it's sort of confusing that we have two functions with slightly different names that do slightly different things.  I would try to make it clear in the name somehow the difference between the two functions.

frontend/public/components/edit-yaml.jsx
```diff
@@ -369,14 +439,59 @@ export const EditYAML_ = connect(stateToProps)(
...
+            });
+            if (!hasErrors) {
+              //Check for duplicate name/kinds
+              const uniqueEntries = _.uniqBy(objs, (obj) =>
+                [
+                  obj.metadata.name,
+                  obj.metadata.namespace,
+                  obj.kind,
+                  groupVersionFor(obj.apiVersion).group,
+                ].join('-'),
```
spadgett, line 484: `-` isn't a good separate here since it's allowed in names. You would still have false positives for

name: foo-bar
namespace: baz

and

name: foo
namespace: bar-baz

Use something like `~` which isn't allowed in names.

frontend/public/components/import-yaml-results.tsx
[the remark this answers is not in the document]
spadgett: > added type React.ReactNode here

I don't see the type :)

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,236 @@
...
+    Promise.allSettled(requests).then(() => {
+      setInFlight(false);
+    });
+    requests.forEach((resourceRequest, index) => {
+      resourceRequest
+        .then(() =>
+          setImportStatus((prevResources) => {
+            const nextResources = [...prevResources];
+            nextResources[index] = {
+              id: index,
```
spadgett, line 113: If `id` is just the index in the array, why do we need `id` at all? Can't we just use the index, which we should already know?

spadgett (commented):
Thanks, @jcaianirh. This is getting close.

frontend/public/components/edit-yaml.jsx
jcaianirh: The error references the part of the yaml that has the issue, so if the ns is missing, the ns is called out, but not the kind. I'll add the kind here as well.
kind name: error ie 
Secret example 1: namespace "joesnamespace" doesn't exist

frontend/public/components/edit-yaml.jsx
jcaianirh: used createResources above and performDryRun here

frontend/public/components/import-yaml-results.tsx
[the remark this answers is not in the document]
jcaianirh: Where did my type gooooo.. ok, readded.

frontend/public/components/import-yaml-results.tsx
jcaianirh: @spadgett i was using the index in the retry method to easily index the filtered list of error objects, but maybe i can adjust that to remove the filter , and just use map so I have an index.

spadgett: We should be able to use `Array.reduce`, something like

```ts
const failedResourceObjects = importStatus.reduce((resource, arr, i) => {
  if (!resource.error) {
    return arr;
  }
  return [...arr, importResources[i]];
}, []);
```

frontend/public/components/edit-yaml.jsx
```diff
@@ -278,47 +305,69 @@ export const EditYAML_ = connect(stateToProps)(
           });
         }
 
-        save() {
-          const { onSave, t } = this.props;
-          const { owner } = this.state;
-          let obj;
+        performDryRun(objs) {
+          const { models } = this.props;
+          let dryRunResources = [];
```
spadgett, line 310: This name is a little misleading since they're not resources, they're promises. Since it's only used in `setState`, can we initialize it as a `const` there?

```ts
const promises = createResources(objs, models, true);
```

frontend/public/components/edit-yaml.jsx
```diff
@@ -278,47 +305,69 @@ export const EditYAML_ = connect(stateToProps)(
...
+                //catch this error but do nothing since we show individual errors above
+              });
+          });
+        }
 
-          if (onSave) {
-            onSave(this.getEditor().getValue());
-            return;
-          }
+        showResults(value) {
```
spadgett, line 332: Naming nit: I'd call this `setDisplayResults` so it's clear the boolean parameter is setting that value. Otherwise it looks like `showResults` always shows the results and the parameter is a flag that controls some other behavior. Also it's good to use a consistent name with the state value.

frontend/public/components/edit-yaml.jsx
```diff
@@ -278,47 +305,69 @@ export const EditYAML_ = connect(stateToProps)(
...
-          } catch (e) {
-            this.handleError(t('public~Error parsing YAML: {{e}}', { e }));
-            return;
+        onRetry(retryObjs) {
+          this.showResults(false);
+          if (retryObjs) {
+            const yamlTextData = [];
+            retryObjs.forEach((yamlObj) => {
+              yamlTextData.push(this.convertObjToYAMLString(yamlObj).trim());
+            });
```
spadgett, line 343: Better to use `Array.map` here. Also naming nit: they aren't "YAML objects," they're objects. `yamlObj` makes it sound like it's already the YAML string. I would call them `yamlDocuments` as well.

```suggestion
            const yamlDocuments = retryObjs.map((obj) => {
              return this.convertObjToYAMLString(obj).trim();
            });
```

frontend/public/components/edit-yaml.jsx
```diff
@@ -278,47 +305,69 @@ export const EditYAML_ = connect(stateToProps)(
...
+            this.setState({ displayResults: false }, () => {
+              this.getEditor().setValue(yamlTextData.join('\n---\n'));
+            });
           }
+        }
+
+        validateYAML(obj) {
```
spadgett, line 350: This name is confusing since the parameter isn't YAML, it's an object. (YAML is the string representation of the object.) Maybe just call it `validate`.

frontend/public/components/edit-yaml.jsx
spadgett: Or you could call them `requests` if that's more clear

jcaianirh: I think the code block changed, const makes more sense now, updated the name too

frontend/public/components/edit-yaml.jsx
jcaianirh: sounds good...or even:
`const yamlDocuments = retryObjs.map((obj) => this.convertObjToYAMLString(obj).trim());`

frontend/public/components/edit-yaml.jsx
jcaianirh: Its validating yaml syntax? I'll change the name

jcaianirh:
@spadgett addressed review comments

frontend/public/components/edit-yaml.jsx
```diff
@@ -64,19 +66,40 @@ const WithYamlTemplates = (Component) =>
     );
   };
 
+const getModel = (models, obj) => {
```
spadgett, line 69: If you make this part of the component class, you wouldn't need to pass `models` as an argument since it would be available in props.

frontend/public/components/edit-yaml.jsx
```diff
@@ -64,19 +66,40 @@ const WithYamlTemplates = (Component) =>
...
+  if (_.isEmpty(obj) || !models) {
+    return null;
+  }
+  return models.get(referenceFor(obj)) || models.get(obj.kind);
+};
+
+export const createResources = (objs, models, isDryRun) => {
```
spadgett, line 76: Same comment here. If this were a method on the class, you wouldn't need to pass `models`.

frontend/public/components/import-yaml-results.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const [importStatus, setImportStatus] = React.useState<ImportYAMLResourceStatusProps[]>(
+    importResources.map(() => ({
+      creating: true,
+      message: t('public~Creating'),
+    })),
+  );
+  const [inFlight, setInFlight] = React.useState(true);
+  const [errors, setErrors] = React.useState(false);
+  React.useEffect(() => {
+    const resources = createResources(importResources, models);
```
spadgett, line 102: These aren't resources... The name is misleading since they are promises for the HTTP requests. I would call this either `requests` or `promises`.

spadgett (commented):
For approvals. This is ready for QE review.

Thanks @jcaianirh 👍

frontend/public/components/edit-yaml.jsx
```diff
@@ -278,47 +305,65 @@ export const EditYAML_ = connect(stateToProps)(
...
 
-          try {
-            obj = safeLoad(this.getEditor().getValue());
-          } catch (e) {
-            this.handleError(t('public~Error parsing YAML: {{e}}', { e }));
-            return;
+        onRetry(retryObjs) {
+          this.setDisplayResults(false);
+          if (retryObjs) {
+            const yamlDocuments = retryObjs.map((obj) => this.convertObjToYAMLString(obj).trim());
```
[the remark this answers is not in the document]
jcaianirh: @spadgett yes, this is to trim the final newline when joining docs...do you want me to try for a more specific remove for that?

frontend/public/components/edit-yaml.jsx
jcaianirh: Yeah, I changed this such that they are both class methods where the createResources is passed to the new results component as a props, and I don't have to shuffle models around anymore.

frontend/public/components/edit-yaml.jsx
jcaianirh: Yeah, I changed this such that they are both class methods where the createResources is passed to the new results component as a props, and I don't have to shuffle models around anymore.

frontend/public/components/import-yaml-results.tsx
jcaianirh: changed this to requests in stead of resources.

frontend/public/components/edit-yaml.jsx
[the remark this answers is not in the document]
spadgett: Can we instead just leave off the first `\n` when joining? If we get a file without a final newline, we could add it.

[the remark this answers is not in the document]
jcaianirh: @spadgett yeah if we leave off the trim and join with '---\n' we get the desired result.

jcaianirh:
@spadgett Pushed latest changes

yapei:
@yanpzhan is doing verification testing and will report test results back when finished

yanpzhan:
> * if any errors occurred, then the try again button will appear and allow the user to retry all failed resources by stepping back to the import page, with the original yaml content displayed for those failed resources for modification

@jcaianirh  I didn't see the try again button when errors occurred(eg, errors about wrong api version, resource has been existing, etc), still stayed on the yaml editor page. Which scenario would trigger the try again button's appearance?

jcaianirh:
> > * if any errors occurred, then the try again button will appear and allow the user to retry all failed resources by stepping back to the import page, with the original yaml content displayed for those failed resources for modification
> 
> @jcaianirh I didn't see the try again button when errors occurred(eg, errors about wrong api version, resource has been existing, etc), still stayed on the yaml editor page. Which scenario would trigger the try again button's appearance?

Hi, The retry button would appear if there is a collision creating a resource, for example. IE in between the time the back end dry run is complete and when the actual resource create request is received, if a user created a duplicate resource, the error retry button would appear. The way I tested was to pause the code in the debugger, after hitting create, then logging into the cluster and creating the identical resource, and then resuming the code to replicate that collision.

yanpzhan:
After another round of test, the scenario about "Retry" button works and there is no issue.

frontend/packages/integration-tests-cypress/views/yaml-editor.ts
```diff
@@ -13,5 +13,13 @@ export const setEditorContent = (text: string) => {
 // initially yamlEditor loads with all grey text, finished loading when editor is color coded
 // class='mtk26' is the light blue color of property such as 'apiVersion'
 export const isLoaded = () => cy.get("[class='mtk26']").should('exist');
-export const clickSaveCreateButton = () => cy.byTestID('save-changes').click();
+// since yaml editor class mtk26 is a font class it doesn't work on an import page with no text
+// adding a check for the 1st line number, AND providing a wait allowed the load of the full component
+export const isImportLoaded = () => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000);
+  cy.get("[class='line-numbers']").should('exist');
```
spadgett, line 21: I would think the blocking `cy.get` would be sufficient here without the wait. Does it timeout?

frontend/packages/integration-tests-cypress/views/yaml-editor.ts
```diff
@@ -13,5 +13,13 @@ export const setEditorContent = (text: string) => {
...
+};
+export const clickSaveCreateButton = () => cy.byTestID('save-changes').click({ force: true });
```
spadgett, line 23: Why is `force: true` required?

spadgett (commented):
@jcaianirh Thank you for adding the tests!

frontend/packages/integration-tests-cypress/views/yaml-editor.ts
andrewballantyne: My understanding the use of `force: true` in these calls is for when the UI viewport doesn't have the item directly visible but it's still in the DOM; ie scrolled down further or via some hidden aspect/overlay -- this part concerns me a bit as I wonder if background renderings (other tabs for instance) would be caught by this.

dtaylor113: We may need to dig into this, I've noticed in the past the [Create] button for the YAML editor sometimes runs into this.  I thought it might have something to do with the hidden "View [schema] sidebar", but Import YAML editor doesn't have that sidebar.

frontend/packages/integration-tests-cypress/views/yaml-editor.ts
dtaylor113: The monaco editor has a [const editorDidMount = React.useCallback](https://github.com/openshift/console/blob/3e2224c0f281cf3163fd8c81ea778b04b746741e/frontend/packages/console-shared/src/components/editor/YAMLEditor.tsx#L32) and I've noticed the editor initially 'loads' and then 'reloads', so cypress initially sees it's loaded, enters contents, then editor is reloaded.  I've tried to set a `window.yamlEditorLoaded=true` in this callback like cypress does [here](https://docs.cypress.io/api/commands/window#Start-tests-when-app-is-ready).  But callback happens more than once so still and issue.

spadgett (approved):
/lgtm

frontend/packages/integration-tests-cypress/views/yaml-editor.ts
jcaianirh: @spadgett @andrewballantyne @dtaylor113 i was able to remove the force click for these tests.

PR #8866, opened 2021-05-06
Bug 1957967: Increase specificity of listPage.row.clickRowByName

TheRealJon:
The selector used in the Cypress listPage.row.clickRowByName helper is not specific enough and could lead to flakes in some cases. Updating to use data attribute to specifically click the resource name link, since this should be consistent across all list pages.


dtaylor113 (approved):
/lgtm

PR #8872, opened 2021-05-10
Bug 1959916: Make console works against api proxy

wanghaoran1988:
Changing some of the hard coded api endpoint, so it works when launching a console locally against an api proxy.


spadgett (approved):
/lgtm

PR #8873, opened 2021-05-10
Minor fixes to add-capacity-modal

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-1850

**Earlier:**

**After:**


frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -37,7 +37,6 @@ const addCapacity = (uid: string, scName: string) => {
     .click();
 
   // wait for Pvs & Nodes data to load (used for validations)
-  // to-do (Sanjal): add a more robust way instead of wait by checking the currentCapacity/PVsAvailableCapacity components state
   // eslint-disable-next-line cypress/no-unnecessary-waiting
   cy.wait(5000);
```
SanjalKatiyar, line 41: I think that we have to wait for a few seconds here, so that "pvData" and "nodesData" get loaded (used for validation in `isValidTopology` function in `install.ts` file), then we can click on the “add” button. Otherwise, the test will fail because of validation errors.
I tried this here:
```
cy.byTestID('ceph-add-capacity-element')
  .should(‘be.visible’)
  .byTestID('confirm-action')
  .click();
```
But, this might introduce flakes, cause element with ID 'ceph-add-capacity-element' (i.e., `currentCapacity/PVsAvailableCapacity` components in `add-capacity-modal`) is only dependent on "pvData" not on "nodesData", and we have to wait till both pvData and nodesData gets loaded or else validation will fail and clicking on “add” button will not allow addition of new capacity, hence test will fail.

bipuladh:
/hold 
To be merged on 4.9.

frontend/packages/ceph-storage-plugin/integration-tests-cypress/tests/multiple-storageclass-selection.spec.ts
SanjalKatiyar: so, in short...
currentCapacity/PVsAvailableCapacity --> these components' state is only dependent on "pvData", thus, we cannot check for their visibility...
cause we only want to click "add" button once both "pvData" and "nodesData" gets loaded...

afreen23: This can be a problem in production too then, lets disable the add capacity button if there are any errors or data is not completely loaded.

bipuladh:
We need to look into whether we should show a total loading skeleton until all the data that is required is present. And not just show some bits as loading.

afreen23:
> We need to look into whether we should show a total loading skeleton until all the data that is required is present. And not just show some bits as loading.

For 4.8 we definitely don't want to add any loading states, it will require some breaking changes.
For 4.8, we do want to stabilize tests and code both by disabling the button.

SanjalKatiyar:
Closing this PR as this issue is now obsolete.

PR #8877, opened 2021-05-10
Bug 1930007: Allow multiple selection on resources drop down

kdoberst:
Addresses [Bug 1930007](https://bugzilla.redhat.com/show_bug.cgi?id=1930007)

This allows the user to select multiple options without closing the drop down for the Resources drop down on the Events and Search pages.

This swapped out the custom Dropdown component with the PatternFly Select component.

NOTE:  The Workloads -> Pods page, click 'Filter' dropdown was resolved as part of another PR.  This PR only applies to the Events and Search pages.

https://user-images.githubusercontent.com/82059948/117692759-f2a9a100-b182-11eb-9862-f221cabda1a5.mov

### DIFF
--- a/frontend/public/components/_resource-dropdown.scss
+++ b/frontend/public/components/_resource-dropdown.scss
@@ -55,6 +55,7 @@
 
 .co-search-group__resource {
   margin: 0 15px 15px 0;
+  width: auto !important;
 }
 
 .co-search-input.pf-c-form-control {
--- a/frontend/public/components/resource-dropdown.tsx
+++ b/frontend/public/components/resource-dropdown.tsx
@@ -3,10 +3,9 @@ import * as _ from 'lodash-es';
 import { connect } from 'react-redux';
 import { Map as ImmutableMap, Set as ImmutableSet } from 'immutable';
 import * as classNames from 'classnames';
-import * as fuzzy from 'fuzzysearch';
 import { useTranslation } from 'react-i18next';
 
-import { Dropdown, ResourceIcon } from './utils';
+import { ResourceIcon } from './utils';
 import {
   K8sKind,
   K8sResourceKindReference,
@@ -14,7 +13,7 @@ import {
   referenceForModel,
   DiscoveryResources,
 } from '../module/k8s';
-import { Badge, Checkbox } from '@patternfly/react-core';
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
 
 // Blacklist known duplicate resources.
 const blacklistGroups = ImmutableSet([
@@ -27,40 +26,14 @@ const blacklistResources = ImmutableSet([
   'events.k8s.io/v1beta1.Event',
 ]);
 
-const DropdownItem: React.SFC<DropdownItemProps> = ({ model, showGroup, checked }) => {
-  const { t } = useTranslation();
-  return (
-    <>
-      <span className={'co-resource-item'}>
-        <Checkbox
-          tabIndex={-1}
-          id={`${model.apiGroup}:${model.apiVersion}:${model.kind}`}
-          isChecked={checked}
-        />
-        <span className="co-resource-icon--fixed-width">
-          <ResourceIcon kind={referenceForModel(model)} />
-        </span>
-        <span className="co-resource-item__resource-name">
-          <span>
-            {model.labelKey ? t(model.labelKey) : model.kind}
-            {model.badge && model.badge === 'Tech Preview' && (
-              <span className="co-resource-item__tech-dev-preview">{t('public~Tech Preview')}</span>
-            )}
-          </span>
-          {showGroup && (
-            <span className="co-resource-item__resource-api text-muted co-truncate show co-nowrap small">
-              {model.apiGroup || 'core'}/{model.apiVersion}
-            </span>
-          )}
-        </span>
-      </span>
-    </>
-  );
-};
-
 const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
-  const { selected, onChange, allModels, className, groupToVersionMap } = props;
+  const { selected, onChange, allModels, groupToVersionMap, className } = props;
   const { t } = useTranslation();
+
+  const [isOpen, setOpen] = React.useState(false);
+  const [selectedOptions, setSelectedOptions] = React.useState(selected);
+  const [filterText, setFilterText] = React.useState<string>(null);
+
   const resources = allModels
     .filter(({ apiGroup, apiVersion, kind, verbs }) => {
       // Remove blacklisted items.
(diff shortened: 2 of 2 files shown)


frontend/public/components/_resource-dropdown.scss
```diff
@@ -55,6 +55,7 @@
 
 .co-search-group__resource {
   margin: 0 15px 15px 0;
+  width: auto !important;
```
kdoberst, line 58: This is less than ideal, but there wasn't another way to overwrite the patternfly width on just the events page.  `.co-search-group__resource` is only used on that page.

frontend/public/components/resource-dropdown.tsx
```diff
@@ -27,40 +26,14 @@ const blacklistResources = ImmutableSet([
...
-    </>
-  );
-};
-
 const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
-  const { selected, onChange, allModels, className, groupToVersionMap } = props;
+  const { selected, onChange, allModels, groupToVersionMap, className } = props;
   const { t } = useTranslation();
+
+  const [isOpen, setIsOpen] = React.useState(false);
```
spadgett, line 33: nit: Maybe it's my background as a Java developer, but I prefer `isOpen` and `setOpen`

frontend/public/components/resource-dropdown.tsx
```diff
@@ -27,40 +26,14 @@ const blacklistResources = ImmutableSet([
...
+  const [selectedOptions, setSelectedOptions] = React.useState(selected);
+  const [filterText, setFilterText] = React.useState(null);
```
spadgett, line 35: TypeScript won't be able to infer a type here since we're initializing the value to `null`, so it's better to declare it explicitly.

```suggestion
  const [filterText, setFilterText] = React.useState<string>(null);
```

frontend/public/components/resource-dropdown.tsx
```diff
@@ -92,51 +65,79 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+    setSelectedOptions(selected);
+  }, [selected]);
 
-    return fuzzy(
-      _.toLower(text),
-      model.labelKey ? _.toLower(t(model.labelKey)) : _.toLower(model.kind),
+  const items = [];
+  resources.forEach((model) => {
+    items.push(
+      <SelectOption
```
spadgett, line 75: Prefer `Array.map`.

```ts
const items = resources.map((model: K8sKind) => <SelectOption>...</SelectOption>;
```

frontend/public/components/resource-dropdown.tsx
```diff
@@ -92,51 +65,79 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+        key={referenceForModel(model)}
+        value={`${referenceForModel(model)}`}
```
spadgett, line 77:
```suggestion
        value={referenceForModel(model)}
```

frontend/public/components/resource-dropdown.tsx
```diff
@@ -92,51 +65,79 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+        data-filter-text={`${model.abbr}${model.labelKey ? t(model.labelKey) : model.kind}`}
+      >
+        <span className={'co-resource-item'}>
```
spadgett, line 80:
```suggestion
        <span className="co-resource-item">
```

frontend/public/components/resource-dropdown.tsx
```diff
@@ -92,51 +65,79 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+                {model.apiGroup || 'core'}/{model.apiVersion}
+              </span>
+            )}
+          </span>
+        </span>
+      </SelectOption>,
     );
+  });
+
+  const onCustomFilter = (event) => {
```
spadgett, line 104: We should add a type for `event`. (I'm not sure the right type offhand... We should be able to check the PatternFly doc.)

frontend/public/components/resource-dropdown.tsx
```diff
@@ -92,51 +65,79 @@ const ResourceListDropdown_: React.SFC<ResourceListDropdownProps> = (props) => {
...
+      return item.props['data-filter-text'].toLowerCase().includes(filterText);
+    });
   };
 
-  const handleSelected = (value: string) => {
+  const handleSelected = (event, value: string) => {
     onChange(referenceForModel(modelFor(value)));
   };
 
+  const onToggle = (newOpenState) => {
```
spadgett, line 120:
```suggestion
  const onToggle = (newOpenState: boolean) => {
```

spadgett (commented):
Thanks @kdoberst

frontend/public/components/resource-dropdown.tsx
kdoberst: Changed

frontend/public/components/resource-dropdown.tsx
kdoberst: Added

frontend/public/components/resource-dropdown.tsx
kdoberst: Changed

frontend/public/components/resource-dropdown.tsx
kdoberst: Added type using PatternFly's docs:

onFilter: event: React.ChangeEvent<HTMLInputElement>
onSelect:  React.MouseEvent | React.ChangeEvent

frontend/public/components/resource-dropdown.tsx
spadgett: The result of the map should be the same... except TypeScript is able to infer a more specific type for the array. When you initialize the array as `const items = []`, TypeScript is treating it as an array of `any`, so we're skipping some type checking.

Can you share the code you tried?

kdoberst: Here is the code I tried:
```
const items = resources
    .map((model: K8sKind) => (
      <SelectOption
        key={referenceForModel(model)}
        value={referenceForModel(model)}
        data-filter-text={`${model.abbr}${model.labelKey ? t(model.labelKey) : model.kind}`}
      >
          <span className="co-resource-icon--fixed-width">
          </span>
          <span className="co-resource-item__resource-name">
            <span>
              {model.labelKey ? t(model.labelKey) : model.kind}
              {model.badge && model.badge === 'Tech Preview' && (
                <span className="co-resource-item__tech-dev-preview">
                  {t('public~Tech Preview')}
                </span>
              )}
            </span>
            {isDup(model.kind) && (
              <span className="co-resource-item__resource-api text-muted co-truncate show co-nowrap small">
                {model.apiGroup || 'core'}/{model.apiVersion}
              </span>
            )}
          </span>
        </span>
      </SelectOption>
    ))
```
Pretty straight forward, but It seems like PF really wants an actual array.

With that said, I think just changing it to an array using `.toArray()` might be  a better option than the approach what I was taking.  Updated commit coming soon.

spadgett: @kdoberst Ah, `resources` is not an array. It uses Immutable.js! I see now. Yes, you'll need `toArray`!

spadgett (approved):
/lgtm

PR #8878, opened 2021-05-10
Bug 1957895: Cypress helper projectDropdown.shouldContain is not an assertion

dtaylor113:
The test would still fail with `.contains()` since the element would not be found, but we should probably use an assertion when the helper name starts with "should".

Added `contains().should('exist')` for readability

### DIFF
--- a/frontend/packages/integration-tests-cypress/views/common.ts
+++ b/frontend/packages/integration-tests-cypress/views/common.ts
@@ -19,7 +19,11 @@ export const projectDropdown = {
       .contains(projectName)
       .click();
   },
-  shouldContain: (name: string) => cy.byLegacyTestID('namespace-bar-dropdown').contains(name),
+  shouldContain: (name: string) =>
+    cy
+      .byLegacyTestID('namespace-bar-dropdown')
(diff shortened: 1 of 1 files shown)


frontend/packages/integration-tests-cypress/views/common.ts
```diff
@@ -19,7 +19,11 @@ export const projectDropdown = {
       .contains(projectName)
       .click();
   },
-  shouldContain: (name: string) => cy.byLegacyTestID('namespace-bar-dropdown').contains(name),
+  shouldContain: (name: string) =>
+    cy
+      .byLegacyTestID('namespace-bar-dropdown')
+      .contains(name)
+      .should('exist'),
```
TheRealJon, line 26: nit, this should probably be the inverse of `shouldNotContain`:

```suggestion
      .should('contain', name),
```

TheRealJon (approved):
/lgtm

Not going to block the PR for this, but IMO the `shouldContain` and `shouldNotContain` helpers should probably be inverses of each other for consistency.

PR #8880, opened 2021-05-10
Bug 1957756: Fix key value in replacement object 

afreen23:
- key is `node` but we are fetching `nodeName`.
- removed the check for metrics map since it's redundant.
- updated check for the replaced device based on  path and deviceID to differentiate similar paths on same node


bipuladh (approved):
/lgtm

afreen23:
/hold cancel
tested with actual disks based environment. Working fine

PR #8882, opened 2021-05-10
Bug 1943238: Display olm conditions descriptors in their own page section on operand details pages

TheRealJon:
Update condition descriptors to be rendered in separate, full width tables like the standard status.conditions table.

### Before
before

### After
after

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/descriptors/utils.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/descriptors/utils.tsx
@@ -11,6 +11,23 @@ import {
   REGEXP_NESTED_ARRAY_PATH,
 } from './const';
 import { Descriptor, SpecCapability, StatusCapability } from './types';
+import { getSchemaAtPath } from '@console/shared';
+
+export const useCalculatedDescriptorProperties = (descriptorType, descriptor, schema, obj) => {
+  const propertySchema = getSchemaAtPath(schema, `${descriptorType}.${descriptor.path}`);
+  const fullPath = [descriptorType, ..._.toPath(descriptor.path)];
+  const displayName =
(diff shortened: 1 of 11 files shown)


frontend/packages/operator-lifecycle-manager/src/components/descriptors/utils.tsx
```diff
@@ -11,6 +11,23 @@ import {
   REGEXP_NESTED_ARRAY_PATH,
 } from './const';
 import { Descriptor, SpecCapability, StatusCapability } from './types';
+import { getSchemaAtPath } from '@console/shared';
+
+export const withCalculatedDescriptorProperties = (descriptorType, descriptor, schema, obj) => {
```
spadgett, line 16: The `with` prefix here makes this sound like it's a higher order component like `withFallback` or `withTranslation`, but it's not. I might just call this `getDescriptorProperties`

frontend/packages/operator-lifecycle-manager/src/components/operand/index.tsx
```diff
@@ -582,6 +593,10 @@ export const OperandDetails = connectToModel(({ crd, csv, kindObj, obj }: Operan
           <Conditions conditions={status.conditions} />
         </div>
       )}
+      {conditionsStatusDescriptors?.length &&
```
spadgett, line 596: Will this output a `0` to the page if it's an empty array?

spadgett (approved):
/lgtm

alexbern65:
Hello,
I am from the team that had submitted the original defect. Thanks for taking care of it.
I could not figure out how this placement is going to be controlled, are there going to be any controls to manage this, or is everything going to be transparent, let's say when I upgrade my cluster tables will just be displayed differently?

Thank you

PR #8883, opened 2021-05-10
Bug 1957561: lack of pseudolocalization for some text on Cluster Setting

florkbr:
Updates hard coded strings for "ClusterOperators" tab on the cluster
settings page and the 3 progress indicators when the cluster is
upgrading for "ClusterOperators", "Master Nodes", and "Worker Nodes".

Screen Shot 2021-05-10 at 4 06 06 PM

https://bugzilla.redhat.com/show_bug.cgi?id=1957561


frontend/public/components/cluster-settings/cluster-settings.tsx
```diff
@@ -528,7 +528,7 @@ export const NodesUpdatesGroup: React.FC<NodesUpdatesGroupProps> = ({
     <UpdatesGroup divided={divided}>
       <UpdatesType>
         <Link to={`/k8s/cluster/nodes?rowFilter-node-role=${machineConfigPool.metadata.name}`}>
-          {name} {NodeModel.labelPlural}
+          {t('public~{{name}} {{resource}}', { name, resource: NodeModel.labelPlural })}
```
spadgett, line 531: I would just use `Nodes` inline here so that the resource kind gets translated.

```suggestion
          {t('public~{{name}} Nodes', { name })}
```

florkbr: @spadgett done!

PR #8885, opened 2021-05-11
Fix that boolean values (esp. false) are not loaded and synced correctly

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5820

**Analysis / Root cause**: 
This issue happen only when user settings are saved in local storage. Which is commonly used in local development.

The value is not loaded and synced because of two truthy checks.

**Solution Description**: 
While loading the data from local storage, accept also `false` values and replace the truthy check with an not undefined check.

Do this as well when a sync event was received to update `false` values between browser tabs and windows.

Before:
odc-5820-before

After:
odc-5820-after

Not changed as there are no breaking test.

**Test setup:**
Add the following code to a functional component:

```tsx
import { Switch } from '@patternfly/react-core';
import { useUserSettings } from '@console/shared/src';

...

const [value, setValue] = useUserSettings('test', true, true);

...

<Switch
  isChecked={value || false}
  onChange={(changedValue) => setValue(changedValue)}
  value={JSON.stringify(value)}
/>
```

JS snippet to remove the test user settings:

```js
us = JSON.parse(window.localStorage.getItem('console-user-settings')); delete us.test; window.localStorage.setItem('console-user-settings', JSON.stringify(us))
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettingsLocalStorage.ts
@@ -14,7 +14,8 @@ export const useUserSettingsLocalStorage = <T>(
   const [data, setData] = React.useState(() => {
     const valueInStorage =
       storage.getItem(storageKey) !== null && deseralizeData(storage.getItem(storageKey));
-    return valueInStorage?.hasOwnProperty(keyRef.current) && valueInStorage[keyRef.current]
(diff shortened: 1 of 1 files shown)


jerolimov:
/cc @nemesis09
This change fixes your issue with using user settings on the add page redesign.

invincibleJai:
/lgtm

PR #8888, opened 2021-05-11
Bug 1961393: Use K8sModel as kindObj if available

rawagner:
there's already a code which uses props.kindObj or k8sModel

```js
const [model] = useK8sModel(props.kind);
const kindObj = props.kindObj ?? model;
```

but in some places we still work only with `props.kindObj` , not `kindOjb`


spadgett (approved):
/lgtm
/hold

@rawagner Do you mind opening a Bugzilla so we can merge?

rawagner:
@spadgett linked to bugzilla

PR #8892, opened 2021-05-11
Fix node-v15 pipeline test failures

andrewballantyne:
https://issues.redhat.com/browse/ODC-5824

Does two things:

1. Cleans up logs so the pipeline tests run without spewing run-time console logs
2. Fixes up a couple tests so they can work with node-v15 -- `jest.spyOn` invokes the method with no arguments and one of our methods was incorrectly trying to test that a try/catch caught errors and mocked a situation where an undefined Pipeline object was being invoked for `.metadata.name`

Best guess why this is happening is this is somehow in the Node code on how to swallow some errors that jest is creating (maybe in a worker thread?) and thus when upgraded to v15 it exposes this issue instead of hiding it. There is a slack thread of others on the team testing on v14 of node and it passes as it always has -- see the ticket for that thread.

### DIFF
--- a/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
+++ b/frontend/packages/dev-console/src/components/import/__tests__/import-submit-utils.spec.ts
@@ -9,6 +9,7 @@ import {
   BuildConfigModel,
   SecretModel,
 } from '@console/internal/models';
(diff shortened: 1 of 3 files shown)


karthikjeeyar:
Verified on latest node, works as expected without any errors in the output.
/lgtm

PR #8895, opened 2021-05-11
CONSOLE-2769: Remove console plugin on operator uninstall

rebeccaalpert:
Added logic to the deletion modal to disable enabled console plugins on operator deletion. I added the text Peter Kreuser suggested in Jira. I asked how he wanted to handle situations when the user doesn't have patch authority, but he was still waiting to hear back when I last spoke to him.

Screen Shot 2021-05-11 at 1 40 09 PM

Fixes https://issues.redhat.com/browse/CONSOLE-2769.


rebeccaalpert:
CC @openshift/team-ux-review.

rebeccaalpert:
Looks like a test flake. /test e2e-gcp-console

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,6 +34,19 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+  });
+
+  const canPatchConsoleOperatorConfig = useAccessReview({
+    group: ConsoleOperatorConfigModel.apiGroup,
+    resource: ConsoleOperatorConfigModel.plural,
+    verb: 'patch',
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+
   const submit = React.useCallback(
```
spadgett, line 50: Not new, but I don't think we want to use `useCallback` here. I'd remove that.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -85,7 +125,8 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
           <strong>{{ namespace }}</strong>. Removing the Operator will not remove any of its custom
           resource definitions or managed resources. If your Operator has deployed applications on
           the cluster or configured off-cluster resources, these will continue to run and need to be
-          cleaned up manually.
+          cleaned up manually. Any console plugins provided by this operator will be disabled and
+          removed.
```
spadgett, line 129: We shouldn't show this in the message to users who don't actually have authority to patch the config and remove the plugin.

cc @itsptk for design input

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -50,6 +80,8 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
                 ).catch(() => Promise.resolve()),
               ]
             : []),
+          canPatchConsoleOperatorConfig &&
```
spadgett, line 83: Won't this sometimes pass `false` to `Promise.all`? What will that do? I think we want to fix it so the array only contains promises.

spadgett (changes requested):

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
rebeccaalpert: [It looks like Promise.all is smart enough to handle it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#using_promise.all), but I can change it.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,6 +34,19 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
 }) => {
   const { t } = useTranslation();
   const [handlePromise, inProgress, errorMessage] = usePromiseHandler();
+  const [consoleOperatorConfig] = useK8sWatchResource<K8sResourceKind>({
```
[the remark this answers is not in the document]
rebeccaalpert: How do you test this? I asked Robb and he wasn't sure. It also didn't come up in his earlier PR.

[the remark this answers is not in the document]
spadgett: We should check if it's not a problem in the other PR.

You'll need to run these commands to create a test/test user:

```
$ oc apply -f ./frontend/integration-tests/data/htpasswd-secret.yaml
$ oc patch oauths cluster --patch "$(cat ./frontend/integration-tests/data/patch-htpasswd.yaml)" --type=merge
```

Then you'll need to grant them authority to uninstall operators. I'll have to dig into exactly what is required for that. I suspect granting the `olm-operators-admin` role would work. @TheRealJon might be able to help.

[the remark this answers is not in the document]
rebeccaalpert: I talked to @TheRealJon and he walked me through this. I impersonated the user and opened the modal/deleted the operator and it seemed ok.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
rebeccaalpert: @itsptk will be on paternity leave starting tomorrow. This was his last message to me about this. 

> I mentioned at our weekly stakeholder call with Sam and Ali that we were trying to sort out this messaging. Sam mentioned he thought it was possible but wasn't sure how common it would be, but did want to follow up at an architects meeting to discuss this scenario more in general. He said he would follow back up with us.

rebeccaalpert:
I addressed the feedback I have the resources to address at this point.

rebeccaalpert:
Megan provided a new design. Her design changed the text and included a link to the cluster plugin. I didn't see any easy way to do that, so I left that out and bolded it instead.

Screen Shot 2021-05-18 at 5 06 27 PM

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,69 +34,119 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
-        .catch(() => {});
-    },
-    [close, handlePromise, k8sKill, subscription],
+  const [consoleOperatorConfig] = useK8sWatchResource<K8sResourceKind>({
+    kind: referenceForModel(ConsoleOperatorConfigModel),
+    isList: false,
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+
+  const canPatchConsoleOperatorConfig = useAccessReview({
```
spadgett, line 43: It would be better to move this above `useK8sWatchResource` and pass `null` to `useK8sWatchResource` if the user doesn't have permission to patch. Then we don't try to watch something that's not needed.

```ts
const [consoleOperatorConfig] = useK8sWatchResource<K8sResourceKind>(canPatchConsoleOperatorConfig
  ? { kind: referenceForModel(ConsoleOperatorConfigModel), ... }
  : null);

File: frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx [hunk 6]
```diff
@@ -24,69 +34,119 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+    group: ConsoleOperatorConfigModel.apiGroup,
+    resource: ConsoleOperatorConfigModel.plural,
+    verb: 'patch',
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+
+  const csvPlugins = getClusterServiceVersionPlugins(csv?.metadata?.annotations);
```
spadgett, line 50: I would also check that `cvsPlugins` is not empty before watching the operator config resource.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,69 +34,119 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+  const enabledPlugins = csvPlugins.filter((plugin) =>
+    isPluginEnabled(consoleOperatorConfig, plugin),
   );
 
+  const getPlugins = () => {
```
spadgett, line 55: We should make it clear this isn't returning an array of plugin names but text we show in the UI.

```suggestion
  const getPluginsLabel = () => {
```

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,69 +34,119 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+            Operator <strong>{{ name }}</strong> will be removed from{' '}
+            <strong>{{ namespace }}</strong>, but any of its custom resource definitions or managed
+            resources will remain. If your Operator deployed applications on the cluster or
+            configured off-cluster resources, these will continue to run and require manual cleanup.
+          </Trans>
+        </p>
+        {canPatchConsoleOperatorConfig && enabledPlugins.length === 1 && (
+          <p>
+            <Trans t={t} ns="olm">
+              The <strong>{{ pluginName }}</strong> console plugin provided by this operator will be
```
spadgett, line 137: This is going to bold the `and` in the plugin label.

spadgett: We might just skip showing the plugin name. It's generally not too important and could cause problems for translation.

rebeccaalpert: More than happy to drop that.

rebeccaalpert:
All set!

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,50 +34,76 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+  const canPatchConsoleOperatorConfig = useAccessReview({
+    group: ConsoleOperatorConfigModel.apiGroup,
+    resource: ConsoleOperatorConfigModel.plural,
+    verb: 'patch',
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+
+  const csvPlugins = getClusterServiceVersionPlugins(csv?.metadata?.annotations);
+
+  const [consoleOperatorConfig] = useK8sWatchResource<K8sResourceKind>(
```
spadgett, line 47: Sorry, missed this on the last review, but we need to be handling the `loadError` return value or check that `consoleOperatorConfig` is defined. We shouldn't try to patch the operator config if it didn't load.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -80,13 +116,29 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+            Operator <strong>{{ name }}</strong> will be removed from{' '}
+            <strong>{{ namespace }}</strong>, but any of its custom resource definitions or managed
+            resources will remain. If your Operator deployed applications on the cluster or
+            configured off-cluster resources, these will continue to run and require manual cleanup.
+          </Trans>
+        </p>
+        {canPatchConsoleOperatorConfig && enabledPlugins.length === 1 && (
+          <p>
+            <Trans t={t} ns="olm">
+              The <strong>{{ pluginName }}</strong> console plugin provided by this operator will be
```
spadgett, line 130: It looks like you removed the logic to get the plugin name but left the variable in the message.

rebeccaalpert: My bad; I only removed one.

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -80,13 +117,24 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+        {canPatchConsoleOperatorConfig && enabledPlugins.length === 1 && (
+          <p>
+            {t('olm~The console plugin provided by this operator will be disabled and removed.')}
+          </p>
+        )}
+        {canPatchConsoleOperatorConfig && enabledPlugins.length > 1 && (
+          <p>
+            {t('olm~The console plugins provided by this operator will be disabled and removed.')}
+          </p>
+        )}
```
spadgett, line 137: Can we use i18next plurals here instead of two different sentences?

rebeccaalpert:
Changes are all set!

frontend/packages/operator-lifecycle-manager/src/components/modals/uninstall-operator-modal.tsx
```diff
@@ -24,50 +34,80 @@ export const UninstallOperatorModal: React.FC<UninstallOperatorModalProps> = ({
...
+    event.preventDefault();
+    const deleteOptions = {
+      kind: 'DeleteOptions',
+      apiVersion: 'v1',
+      propagationPolicy: 'Foreground',
+    };
+
+    const patch =
+      removePlugins &&
+      enabledPlugins.map((plugin) => getPluginPatch(consoleOperatorConfig, plugin, false));
```
spadgett, line 74: We should make sure `patch` is only ever assigned an object, not a boolean.

```suggestion
      removePlugins ?
      enabledPlugins.map((plugin) => getPluginPatch(consoleOperatorConfig, plugin, false)) : null;
```

spadgett:
for approvals

rebeccaalpert:
Changes all set!

spadgett (approved):
/lgtm
/hold

yanpzhan:
@rebeccaalpert , Hi, I see in your pictures, the console plugin name is printed, but in my test, the plugin name is not shown. Is it the new approaching?
Screenshot from 2021-05-26 16-51-09

spadgett:
> @rebeccaalpert , Hi, I see in your pictures, the console plugin name is printed, but in my test, the plugin name is not shown. Is it the new approaching?

@yanpzhan Yes, we've decided not to show the name. That's expected.

yanpzhan:
Thanks for your confirmation, Sam.

frontend/packages/operator-lifecycle-manager/locales/en/olm.json
```diff
@@ -167,7 +167,9 @@
...
+  "Operator <1>{{name}}</1> will be removed from <4>{{namespace}}</4>, but any of its custom resource definitions or managed resources will remain. If your Operator deployed applications on the cluster or configured off-cluster resources, these will continue to run and require manual cleanup.": "Operator <1>{{name}}</1> will be removed from <4>{{namespace}}</4>, but any of its custom resource definitions or managed resources will remain. If your Operator deployed applications on the cluster or configured off-cluster resources, these will continue to run and require manual cleanup.",
```
sferich888, line 170: Do we need to make it more clear that this is being removed from a namespace? 
Based on the PR it seems like we are setup to convey this to the user as best we can; but it unclear how clear this messaging is.

megan-hall: We do call out "Operator" before we list the name of the Operator so we could probably do the same for "Project" if you think that level of information would make it more clear.

PR #8896, opened 2021-05-11
CONSOLE-2223: Past logs for selected container on pod logs tab.

kdoberst:
This address [Console 223](https://issues.redhat.com/browse/CONSOLE-2223)

Additions:
  - Add a select item allowing a user to view either the current log or the previous log for a container.
  - The select will be disabled if a previous log is not available

Notes:
 - The code determines if a previous log is available if the container restart count is greater than 0
 - The select is not available for build logs
 - The tooltip for a disabled select is not available in full screen mode.
 - When a user changes containers from the container select, the log type changes back to current log.
- There is a known issue where if the container only has a current long, and the log is rather long, the log type tooltip is not available until number of log lines equal to the buffer (~1000 log lines) are loaded.  This is usually a couple of seconds.

https://user-images.githubusercontent.com/82059948/118874050-14f09c80-b8b0-11eb-946f-6d721819057d.mov


kdoberst:
@spadgett @rhamilto Started a PR to ask a couple of questions before I finish coding:

 1.  `resource-log.tsx ` is starting to get rather long.  I know that this code will be touched again when the new PatterFly log viewer is released.  Wasn't sure if this was worth the effort of refactoring and possibly breaking up into multiple components/files.   Thoughts?
 
 2. In order to check if the "previous log" option is enabled is based on if a container has restarted.  It is a quick api call to get the data (vs a streaming type call).  I've made a quick axios get call.  I noticed that axios isn't used a lot in the app.  Is there a different approach I should be using?
 
 3.  Tests.  There doesn't appear to be any unit tests (aka Jest) tests for the ResourceLog (defined in `resource-log.tsx`).  Again, knowing that this may touched in a future phase, was unsure about writing test to cover the functionality.  If I do write unit tests, should I write tests for other functionality while I in there?  I'm not entirely sure, but I don't think there are functional tests either.  Thoughts?

frontend/public/components/utils/resource-log.tsx
```diff
@@ -239,6 +324,27 @@ export const ResourceLog: React.FC<ResourceLogProps> = ({
...
+      .then((response) => {
+        const container = response.data.status.containerStatuses?.find(
+          (pod) => pod.name === containerName,
+        );
+        setPreviousLogs(container?.restartCount > 0);
+      })
+      .catch(() => {
+        setPreviousLogs(false);
+      });
+  }, [containerName, resource]);
```
spadgett, line 346: Hey, @kdoberst. We have a `k8sGet` utility to get a Kubernetes resource. In most places, however, we watch the resource instead using WebSockets with `useK8sWatchResource` or `Firehose` so that the UI updates when the resource updates.

The `resource` prop itself should be the `Pod` on the Pod page, so in this case, you shouldn't need to fetch it again. I believe that `resource` can be a `Build`, however, so we'll need to handle that case. @TheRealJon is the most familiar with the log code and can probably help.

spadgett:
@kdoberst I agree that this code could use some refactoring. Given that we're pretty close to freeze for 4.8, we should probably wait... I think it makes sense to look at that and improving tests when we rework the logs to use the PatternFly component.

frontend/public/components/utils/resource-log.tsx
kdoberst: Thanks for pointing out that the content is already available.  Removing axios get call.

kdoberst:
Sounds good.  Well hold off on refactoring and testing enhancement when we implement the new PatternFly items.

kdoberst:
Assigning Sam, Robb, and Jon because they've seen parts of this.

/assign @spadgett 
/assign @rhamilto 
/assign @TheRealJon

frontend/public/components/utils/resource-log.tsx
```diff
@@ -239,6 +341,17 @@ export const ResourceLog: React.FC<ResourceLogProps> = ({
     }
   }, [status, totalLineCount, previousTotalLineCount, bufferSize]);
 
+  //Check to see if previous log exists
+  React.useEffect(() => {
+    if (resource.kind === 'Pod') {
+      setLogType(LOG_TYPE_CURRENT); // Set back to viewing current log when switching resources or containers
```
spadgett, line 347: I believe this will switch the log type back to current when the pod status updates since we're watching the pod. Even if `resource` changes, it might be a status update rather than a totally different resource.

frontend/public/components/utils/resource-log.tsx
```diff
@@ -412,6 +524,10 @@ export const ResourceLog: React.FC<ResourceLogProps> = ({
           namespaceUID={namespaceUID}
           toggleWrapLines={setWrapLines}
           isWrapLines={wrapLines}
+          hasPreviousLog={hasPreviousLogs}
+          changeLogType={setLogType}
+          logType={logType}
+          isContainer={resource.kind === 'Pod'}
```
spadgett, line 530: Maybe we should call this something like `showLogTypeSelect` if that's how it's used? Builds run pods which also have containers, so this might be confusing otherwise.

spadgett (commented):
Nice work! Thanks @kdoberst

frontend/public/components/utils/resource-log.tsx
kdoberst: Very good point. I agree that name, wasn't quite right.  Changed it to `showLogTypeSelect`

frontend/public/components/utils/resource-log.tsx
kdoberst: Good catch, you are right.  Changed the code so the log type is reset back to `current` only if the container name changes (and if the resource.kind is `Pod`).

spadgett (approved):
/lgtm
/hold for approvals

frontend/public/components/utils/resource-log.tsx
```diff
@@ -108,8 +145,69 @@ export const LogControls: React.FC<LogControlsProps> = ({
...
           {t(streamStatusMessages[status])}
-        </div>
+        </>
+      );
+    }
+    return (
+      <>
+        {t(streamStatusMessages[STREAM_EOF])}{' '}
+        {isFullscreen ? null : (
+          <Popover bodyContent={t('public~Play and pause are not available for historical logs.')}>
```
spadgett, line 155: Hey @megan-hall, I was wondering if we should consider removing this popover. The icon makes it look like a question ("Log stream ended?"), and I think "Log stream ended" on its own explains why there are no play or pause buttons. It's also what we show for current logs when the pod is complete. It's not specific to previous logs.

megan-hall: Good point. Sounds good to me!

kdoberst: @spadgett @megan-hall I removed the popover next to "Log stream ended".  Updated the PR description.

spadgett (approved):
/lgtm

yapei:
@schituku  Could you help test and report test result back?

schituku:
Test case: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-41926
tested the PR and looks good. 
/label qa-approved

PR #8900, opened 2021-05-12
Bug 1960059: Monitoring dashboards: Remove the "Grafana UI" link

kyoto:
The Grafana UI is still accessible via its route, but some dashboards
are only available on Console's Monitoring > Dashboards page, so
generally people should go there instead.

FYI @sichvoge, @lbarbeevargas, @cshinn


kyoto:
Thanks @spadgett. I created a Bugzilla bug for easier tracking.

PR #8901, opened 2021-05-12
Bug 1959550: Overly generic CSS rules for dd and dt elements breaks styling elsewhere in console

avivtur:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1959550

**Analysis / Root cause**: 
root cause is explained in BZ above

**Solution Description**: 
inserted generic CSS rules to it's specific intended class, so it affects only the correct CSS class.

Before:

bz_1959550_before

After:

bz_1959550_after


avivtur:
@yaacov Please review

PR #8906, opened 2021-05-12
Fix Uncaught TypeError for OCS Internal mode 

SanjalKatiyar:
Following error occurred while moving from step-1 to step-2 for Internal-mode wizard flow:

Screenshot from 2021-05-11 22-50-45


bipuladh:
/approve
/lgtm
Thank you

PR #8909, opened 2021-05-12
fixes: WebTerminal widget should send resize events

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5773

Screenshots: while pasting a command that exceeds 80 cols
Before:
Screenshot 2021-05-17 at 8 04 44 PM
After:
Screenshot 2021-05-17 at 8 04 55 PM

### DIFF
--- a/frontend/packages/console-app/src/components/cloud-shell/Terminal.tsx
+++ b/frontend/packages/console-app/src/components/cloud-shell/Terminal.tsx
@@ -14,6 +14,7 @@ const terminalOptions: ITerminalOptions = {
 
 type TerminalProps = {
   onData: (data: string) => void;
+  onResize: (cols: number, rows: number) => void;
 };
 
 export type ImperativeTerminalType = {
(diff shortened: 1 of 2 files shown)


JPinkney (approved):
Seems to work well for me! I did notice a couple things:
1. If you continuously type you'll eventually get the horizontal scroll bar at the bottom of the terminal. This happens about 15 characters before the end of the line in the terminal
2. When you re-size from big terminal -> small terminal -> big terminal the words have this weird text overlap thing. Though I'm not sure if we can control that from our side.

Both of those issues should be shown in the gif below. Their fixes (if any) should probably be brought in a different PR. I just wanted to let you know!

openshift-console-test

invincibleJai:
Thanks @sahil143 , have verified the changes and it works as expected with copy and paste of long strings/commands not seeing any wrapping.

i too observe the horizontal scrollbar for a moment as mentioned above and another scenario is on resize cursor position is changed

sahil143:
@invincibleJai @JPinkney I am not able to reproduce the horizontal scrollbar issue.

scrollbar


As for the second issue 

> When you re-size from big terminal -> small terminal -> big terminal the words have this weird text overlap thing. Though I'm not sure if we can control that from our side.

this seems to be a problem with xtermjs. there is already an open issue for this https://github.com/xtermjs/xterm.js/issues/3065

invincibleJai:
/approve

verified the changes on the cluster  and work as expected

frontend/packages/console-app/src/components/cloud-shell/Terminal.tsx
```diff
@@ -24,64 +25,74 @@ export type ImperativeTerminalType = {
...
-    loadAttachAddon: (addOn: ITerminalAddon) => {
-      if (!terminal.current) return;
-      terminal.current.loadAddon(addOn);
-    },
-  }));
+    React.useEffect(() => {
+      const term = terminal.current;
+      term.onData(onData);
+      term.onResize(handleResize);
+    }, [onData, handleResize]);
```
christianvogt, line 68: Dispose of old listeners.

sahil143:
/hold

failing ci/prow/e2e-gcp-console

PR #8918, opened 2021-05-13
Followup to PR #8843

avivtur:
Minor adjustments according to pcbailey comments:
https://github.com/openshift/console/pull/8843

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vm-templates/VMTemplateSupport.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vm-templates/VMTemplateSupport.tsx
@@ -4,12 +4,12 @@ import { useTranslation } from 'react-i18next';
 import { ExternalLink } from '@console/internal/components/utils';
 
 import { SUPPORT_URL } from '../../constants/vm-templates';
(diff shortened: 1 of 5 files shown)


avivtur:
@pcbailey please review

frontend/packages/kubevirt-plugin/src/components/vm-templates/VMTemplateSupport.tsx
```diff
@@ -4,12 +4,13 @@ import { useTranslation } from 'react-i18next';
 import { ExternalLink } from '@console/internal/components/utils';
 
 import { SUPPORT_URL } from '../../constants/vm-templates';
+import { isUpstream } from '../../utils/common';
 
 const VMTemplateSupport: React.FC<VMTemplateSupportProps> = ({ details }) => {
   const { t } = useTranslation();
-  const isUpstream = window.SERVER_FLAGS.branding === 'okd';
+  const upstream = isUpstream();
```
pcbailey, line 11: nitpick: Couldn't we just call it inline in all of the places it's currently used since none of them use a loop? I think it would read a little better this way in addition to declaring another variable.

PR #8919, opened 2021-05-13
CONSOLE-2403: Add a UI for enabling and disabling plugins under Cluster Settings

rhamilto:
Includes https://github.com/openshift/console/pull/8966

Setup for testing:
1.  Deploy `console-demo-plugin` to a cluster via `oc apply -f https://raw.githubusercontent.com/openshift/console/master/frontend/dynamic-demo-plugin/oc-manifest.yaml`
2.  Start `bridge` with the `plugin` option specified as follows:  `./bin/bridge -plugins console-demo-plugin=http://localhost:9001/`
3.  In a new terminal, run `yarn build-dev` and then `yarn http-server` from `console/frontend/dynamic-plugin-demo` 
4.  In a new terminal, run `yarn dev` from `console/frontend` as you normally would

https://user-images.githubusercontent.com/895728/119027544-75e1a880-b974-11eb-9b04-6fe08585c856.mov


frontend/public/components/cluster-settings/cluster-settings.tsx
```diff
@@ -996,7 +996,7 @@ export const ClusterSettingsPage: React.FC<ClusterSettingsPageProps> = ({ match
     },
     {
       href: 'globalconfig',
```
rhamilto, line 998: Wondering if we should change the route to `config` and add a redirect from `globalconfig`?  Related, do we rename `GlobalConfigPage`?  @spadgett, opinions?

rhamilto: I have an optional commit including the rename (went with route: `configuration` so it matches the tab label), but discovered a redirect isn't easily achievable with the way tabs work.  What do you think, @spadgett?

spadgett: Sorry for the late response. If redirects don't work well, I'd probably just keep the existing URL. Better not to break links / bookmarks.

rhamilto: Sounds good.  I've removed the rename commit.  That means the file name, component names, and route all retain their original "global" designations.

frontend/packages/operator-lifecycle-manager/src/plugin.tsx
```diff
@@ -281,6 +282,18 @@ const plugin: Plugin<ConsumedExtensions> = [
...
+    properties: {
+      model: ConsoleOperatorConfigModel,
+      loader: async () =>
+        (
+          await import(
+            './components/console-operator-config' /* webpackChunkName: "console-operator-config" */
+          )
+        ).ConsoleOperatorConfigDetailsPage,
+    },
+  },
```
spadgett, line 296: The console operator details page should really be part of the core console, although I understand why you put it here since the dialog is part of the OLM package. It would be better if we can move the details page into console-app and the modal into console-shared, although I don't know how much of a ripple effect that has :/

frontend/public/components/cluster-settings/global-config.tsx
```diff
@@ -41,13 +42,14 @@ export const breadcrumbsForGlobalConfig = (detailsPageKind: string, detailsPageP
   },
 ];
 
-const ItemRow = ({ item }) => {
+const ItemRow = ({ item, isDup }) => {
```
spadgett, line 45: Naming nit: I would probably rename `isDup` to `showAPIGroup` since that's really the behavior it controls. I think it makes the meaning more obvious below when we use `ItemRow`.

frontend/public/components/cluster-settings/global-config.tsx
```diff
@@ -150,13 +155,15 @@ const GlobalConfigPage_: React.FC<GlobalConfigPageProps & GlobalConfigPageExtens
...
-  }, [configResources, globalConfigs, t]);
+  }, [clusterOperatorConfigResources, configResources, globalConfigs, t]);
   const visibleItems = items.filter(({ label, description = '' }) => {
     return (
       fuzzyCaseInsensitive(textFilter, label) ||
       description.toLowerCase().indexOf(textFilter.toLowerCase()) !== -1
     );
   });
+  const groupedItems = _.groupBy(visibleItems, _.property('label'));
+  const isDup = (item) => _.get(groupedItems, item)?.length > 1;
```
spadgett, line 166: nit: Prefer optional chaining

```suggestion
  const isDup = (item) => groupedItems?.item?.length > 1;
```

frontend/public/locales/en/public.json
```diff
@@ -136,8 +136,18 @@
   "Completed": "Completed",
   "Release notes": "Release notes",
   "Cluster Settings": "Cluster Settings",
-  "Global configuration": "Global configuration",
+  "Configuration": "Configuration",
   "ClusterVersion details": "ClusterVersion details",
+  "{{kind}} details": "{{kind}} details",
+  "{{label}}": "{{label}}",
```
spadgett, line 142: What is the context for this message? Usually if the entire message is a variable it doesn't make sense to externalize it.

spadgett (commented):
@rhamilto, I've reviewed your changes, but skipped @vojtechszocs' since there is a separate PR for that. Nice work!

Thank you @vojtechszocs for the assist!

frontend/public/locales/en/public.json
rhamilto: This is [an existing entry](https://github.com/openshift/console/blob/85086217feb35e64c36d68a8d315fe59275e510b/frontend/public/locales/en/public.json#L160) that is just getting reordered.

frontend/packages/operator-lifecycle-manager/src/components/console-operator-config.tsx
```diff
@@ -0,0 +1,215 @@
...
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  };
+  const [consoleOperatorConfig] = useK8sWatchResource<K8sResourceKind>(console);
+  const canPatchConsoleOperatorConfig = useAccessReview({
+    group: ConsoleOperatorConfigModel.apiGroup,
+    resource: ConsoleOperatorConfigModel.plural,
+    verb: 'patch',
+    name: CONSOLE_OPERATOR_CONFIG_NAME,
+  });
+  const trusted = false;
```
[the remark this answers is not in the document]
rhamilto: Done.

frontend/packages/operator-lifecycle-manager/src/plugin.tsx
rhamilto: Done.  Wasn't too bad.  https://github.com/openshift/console/pull/8919/commits/582fed15426e4d7a27912b762e314f141af8aee6

frontend/public/components/cluster-settings/global-config.tsx
rhamilto: Done.

frontend/public/components/cluster-settings/global-config.tsx
rhamilto: Done.

yapei:
1. Deploy console-demo-plugin manifests
```
$ oc apply -f -f https://raw.githubusercontent.com/openshift/console/master/frontend/dynamic-demo-plugin/oc-manifest.yaml
```
2. Start http-server locally 
```
$ cd frontend/dynamic-demo-plugin; yarn install; yarn build; yarn http-server
```
3. Run` bridge` with loading plugins like this: 
```
./bin/bridge -plugins console-demo-plugin=http://localhost:9001/
``` 
4. Check Cluster Settings -> Click on console.operator.openshift.io -> Console plugins, on 'Console plugins' page user can Enable & Disable plugins

spadgett:
@rhamilto Looks like there are some linter errors.

rhamilto:
> @rhamilto Looks like there are some linter errors.

Thanks.  These should be fixed.  Rebasing long-running PRs is always a challenge.

frontend/public/components/cluster-settings/global-config.tsx
spadgett: Oops I meant `groupedItems?.[item]?.length > 1`, but looks like you caught my bug

frontend/packages/operator-lifecycle-manager/src/components/console-operator-config.tsx
```diff
@@ -0,0 +1,214 @@
+import * as React from 'react';
```
spadgett, line 1: It looks like we have this in the PR twice now?

rhamilto: Good catch! I orphaned this when I was moving files around.  This copy is now deleted.

spadgett (approved):
/lgtm

spadgett:
/lgtm cancel
Needs `yarn i18n` after removing the duplicate file

rhamilto:
> /lgtm cancel
> Needs `yarn i18n` after removing the duplicate file

Fixed.  🤞

rhamilto:
Rebased.  @spadgett please re-tag.

spadgett (approved):
/lgtm

rhamilto:
/hold cancel

all approvals added

PR #8921, opened 2021-05-13
Bug 1940318: Monitoring dashboards: Support colored text for single values

kyoto:
Support the `thresholds` object, which specifies colors for single stats
depending on their value.

Color names are defined by the Grafana dashboard definition format. They
are mapped to PatternFly chart colors to be consistent with the graph
colors on the same page.

Also adds support for `gauge` panels, but just renders them as single
values (with support for threshold colors).

PatternFly colors are from https://www.patternfly.org/v4/guidelines/colors-for-charts

FYI @cshinn 

screenshot


spadgett (approved):
/lgtm

👍

PR #8923, opened 2021-05-13
CONSOLE-2855: Cypress: port protractor Storage tests

dtaylor113:
Ported Protractor's `storage.scenario.ts` to Cypress:
image


frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        detailsPage.clickPageActionFromDropdown('Add storage');
+        cy.contains('Select claim').should('be.visible');
+        cy.byTestID('Create new claim-radio-input').click();
+        cy.byTestID('pvc-name').type(pvcName);
+        cy.byTestID('pvc-size').type(pvcSize);
+        cy.byTestID('mount-path').type(mountPath);
+        cy.get(submitButton).click();
+        cy.get(errorMessage).should('not.exist');
+        cy.contains(pvcName).should('exist');
+        cy.contains(mountPath).should('exist');
```
spadgett, line 59: This just checks if the text is anywhere on the page? I don't think the check is specific enough since we could still be on the form page in theory, which would have both of these strings. Can we scope the check to test specifically that it's part of the volumes table on the workload details page?

spadgett (commented):
Thank you @dtaylor113

frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
dtaylor113: Good point, added a `.parent` and `.within` to limit scope:
image

frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    const pvcName = `${resourceType}-pvc`;
+    const pvcSize = '1';
+    const mountPath = '/data';
+    describe(resourceType, () => {
+      it(`create a ${resourceType} resource and adds storage to it`, () => {
+        listPage.createNamespacedResourceWithDefaultYAML(resourceType, testName);
+        cy.get(errorMessage).should('not.exist');
+
+        detailsPage.clickPageActionFromDropdown('Add storage');
+        cy.contains('Select claim').should('be.visible');
```
florkbr, line 51: @dtaylor113 do we ever intend to run E2E with translation enabled? I'm just curious why we are checking/using strings over IDs, or HTML `data-` attributes.

dtaylor113: In general our e2e tests are run in default us English.   We do run a `pseudo-translation` [test suite](https://github.com/openshift/console/blob/1867306615a027a309ce101d5878b61cfb81b660/frontend/packages/integration-tests-cypress/tests/i18n/pseudolocalization.spec.ts#L4), as well as execute multiple [cy.testI18n(...) calls](https://github.com/openshift/console/blob/f1d9bea02c3ed94fb0f6fcb9a21a67dd41728b4e/frontend/packages/integration-tests-cypress/tests/crud/k8-openshift-cruds.spec.ts#L160) for 27 resource we run thru for our k8.crud testing -this tests various areas of the list view and detail pages.

frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
spadgett: Is there a reason we can't use the existing data-test attributes on the volume table columns?

frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
dtaylor113: irt. `cy.contains('Select claim')`, yeah that is probably better served as an id, so I added a `data-test` attribute and switched to:

`cy.byTestID('claim-name').should('be.visible');`
- thanks

frontend/packages/integration-tests-cypress/tests/crud/add-storage-crud.spec.ts
dtaylor113: Wow!  Didn't realize we pushed `data-test-...` id's to each table cell, that does make things easier :-)
Updated to:
```
cy.get(`[data-test-volume-name-for="${pvcName}"]`).should('have.text', pvcName);
cy.get(`[data-test-mount-path-for="${pvcName}"]`).should('have.text', mountPath);
```

spadgett (approved):
/lgtm

TheRealJon (approved):
/lgtm

PR #8931, opened 2021-05-14
GitOps: Add Sync Status and Latest Deployment Time to List Page (751)

keithchong:
This PR adds enhancements to that table provided by https://github.com/openshift/console/pull/8804.  Please see this for more background information.

First, it adds sync status information of the application environments to the the third column of the table. See https://issues.redhat.com/browse/GITOPS-751.  Tooltips are added to non-zero status counts to show totals.
Second, it adds the latest deployment time of each application in their respective environments, to the last column of the table.  See https://issues.redhat.com/browse/GITOPS-879

This change calls a new `/application` API provided by the GitOps backend.  In cases where the version of Console is dealing with an older version of the GitOps backend without this new API, it will revert to calling the `/pipelines` API to get the list of applications, and the resulting list page will show the 'base' information that is shown in https://github.com/openshift/console/pull/8804.

Screen Shot 2021-05-14 at 9 17 30 AM

Signed-off-by: Keith Chong <kykchong@redhat.com>

### DIFF
--- a/frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
+++ b/frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
@@ -21,6 +21,10 @@ const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg }) => 
     return fuzzyCaseInsensitive(textFilter, name);
   });
 
+  const hasSyncStatus: boolean =
+    appGroups?.some(
+      ({ sync_status }) => sync_status /* eslint-disable-line @typescript-eslint/camelcase */,
+    ) || false;
   return (
     <div className="odc-gitops-list">
       {!emptyStateMsg && appGroups ? (
@@ -35,7 +39,7 @@ const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg }) => 
           <Table
             data={visibleItems}
             aria-label={t('gitops-plugin~Environments table')}
-            Header={GitOpsTableHeader}
+            Header={GitOpsTableHeader(hasSyncStatus)}
             Row={GitOpsTableRow}
             loaded={!emptyStateMsg}
             virtualize
--- a/frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.scss
+++ b/frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.scss
@@ -1,0 +1,9 @@
+.odc-gitops-syncStatus {
+    max-width: '100px'
+}
+.odc-gitops-lastDeploymentTime {
+    color: var(--pf-global--palette--black-600);
+}
+.odc-gitops-tooltip-text {
+    color: white;
+}
--- a/frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
+++ b/frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
@@ -3,47 +3,130 @@ import i18n from '@console/internal/i18n';
 import * as classNames from 'classnames';
 import { GitOpsAppGroupData } from '../utils/gitops-types';
 import { RowFunction, TableData, TableRow } from '@console/internal/components/factory';
-import { Button } from '@patternfly/react-core';
-import { history, ExternalLink } from '@console/internal/components/utils';
+import { Flex, FlexItem, Split, SplitItem } from '@patternfly/react-core';
+import { ExternalLink, Timestamp } from '@console/internal/components/utils';
 import { routeDecoratorIcon } from '@console/dev-console/src/components/import/render-utils';
+import GitOpsSyncFragment from './GitOpsSyncFragment';
+import {
+  GreenCheckCircleIcon,
+  YellowExclamationTriangleIcon,
+  GrayUnknownIcon,
+} from '@console/shared';
+import './GitOpsTableRow.scss';
+import { Link } from 'react-router-dom';
 
 const tableColumnClasses = [
   classNames('pf-m-width-20'), // Application name
-  classNames('pf-m-width-40'), // Git repository
+  classNames('pf-m-width-30'), // Git repository
   classNames('pf-m-hidden', 'pf-m-visible-on-md', 'pf-m-width-20'), // Environments
-  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-20'), // Last deployment
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-30'), // Last deployment
 ];
 
-const handleClick = (appGroup: GitOpsAppGroupData) => {
-  history.push(`/environments/${appGroup.name}?url=${appGroup.repo_url}`);
-};
+const getMatchingEnvs = (envs: string[], desiredStatus: string) => (
+  acc: string[],
+  status: string,
+  idx: number,
+): string[] =>
+  desiredStatus === status
+    ? [...acc, envs[idx]] // 1:1 between a status and an env
+    : acc;
 
 const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = (props) => {
   const { obj: appGroup, index, key, style } = props;
+  const {
+    name,
+    sync_status: syncStatuses = [],
+    environments: envs,
+    last_deployed: lastDeployed = [],
+    repo_url: repoUrl,
+  } = appGroup;
   const t = (tKey) => i18n.t(tKey);
+  const syncedEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'Synced'), []);
+  const outOfSyncEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'OutOfSync'), []);
+  const unknownEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'Unknown'), []);
+  const latestDeployedTime = lastDeployed.reduce(
+    (leadingDeployedTime, deployedTime) =>
+      leadingDeployedTime < deployedTime ? deployedTime : leadingDeployedTime,
+    '',
+  );
+  const latestDeployedEnv = latestDeployedTime
+    ? envs[lastDeployed.indexOf(latestDeployedTime)]
+    : '';
   return (
     <TableRow id={index} index={index} trKey={key} style={style}>
       <TableData className={tableColumnClasses[0]}>
-        <Button
-          onClick={() => handleClick(appGroup)}
-          aria-label={appGroup.name}
-          variant="link"
-          isInline
-        >
-          {appGroup.name}
-        </Button>
+        <Link to={`/environments/${appGroup.name}?url=${appGroup.repo_url}`} title={name}>
+          {name}
+        </Link>
       </TableData>
       <TableData className={classNames(tableColumnClasses[1])}>
-        <ExternalLink href={appGroup.repo_url} additionalClassName={'co-break-all'}>
+        <ExternalLink href={repoUrl} additionalClassName={'co-break-all'}>
           <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>
-            {routeDecoratorIcon(appGroup.repo_url, 12, t)}
+            {routeDecoratorIcon(repoUrl, 12, t)}
           </span>
-          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>{appGroup.repo_url}</span>
+          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>{repoUrl}</span>
         </ExternalLink>
       </TableData>
-      <TableData className={tableColumnClasses[2]}>{appGroup.environments.join(', ')}</TableData>
+      <TableData className={classNames(tableColumnClasses[2], 'pf-u-text-nowrap')}>
+        {syncStatuses.length > 0 ? (
+          <Flex className="odc-gitops-syncStatus">
+            <GitOpsSyncFragment
+              tooltip={syncedEnvs.map((env) => (
+                <Split className="odc-gitops-tooltip-text" hasGutter key={`${name}-${env}`}>
+                  <SplitItem>
+                    <GreenCheckCircleIcon />
+                  </SplitItem>
+                  <SplitItem isFilled>{env}</SplitItem>
+                  <SplitItem>{t('gitops-plugin~Synced')}</SplitItem>
+                </Split>
+              ))}
+              count={syncedEnvs.length}
+              icon="check"
+            />
+            <GitOpsSyncFragment
+              tooltip={outOfSyncEnvs.map((env) => (
+                <Split className="odc-gitops-tooltip-text" hasGutter key={`${name}-${env}`}>
+                  <SplitItem>
+                    <YellowExclamationTriangleIcon />
+                  </SplitItem>
+                  <SplitItem isFilled>{env}</SplitItem>
+                  <SplitItem>{t('gitops-plugin~OutOfSync')}</SplitItem>
+                </Split>
+              ))}
+              count={outOfSyncEnvs.length}
+              icon="exclamation"
+            />
+            <GitOpsSyncFragment
+              tooltip={unknownEnvs.map((env) => (
+                <Split className="odc-gitops-tooltip-text" hasGutter key={`${name}-${env}`}>
+                  <SplitItem>
+                    <GrayUnknownIcon />
+                  </SplitItem>
+                  <SplitItem isFilled>{env}</SplitItem>
+                  <SplitItem>{t('gitops-plugin~Unknown')}</SplitItem>
+                </Split>
+              ))}
+              count={unknownEnvs.length}
+              icon="unknown"
+            />
+          </Flex>
+        ) : (
+          <span>{envs.join(', ')}</span>
+        )}
+      </TableData>
       <TableData className={tableColumnClasses[3]}>
-        {/* this is just a placeholder until backend changes can go in to get this data */}-
+        {latestDeployedTime !== '' ? (
+          <Flex>
+            <FlexItem className="odc-gitops-lastDeploymentTime" spacer={{ default: 'spacerXs' }}>
+              <span>
+                <Timestamp timestamp={latestDeployedTime} />
+              </span>
+            </FlexItem>
+            <FlexItem>{latestDeployedEnv}</FlexItem>
+          </Flex>
+        ) : (
+          <span>-</span>
+        )}
       </TableData>
     </TableRow>
   );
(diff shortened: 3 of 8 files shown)


frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
       <TableData className={tableColumnClasses[0]}>
         <Button
           onClick={() => handleClick(appGroup)}
           aria-label={appGroup.name}
           variant="link"
-          isInline
+          style={{ padding: '0px' }}
         >
           {appGroup.name}
         </Button>
```
andrewballantyne, line 112: Ah I should have noted this on the last PR... I think you want the `Link` component 
```
import { Link } from 'react-router-dom';
```

Example usage from [Helm](https://github.com/openshift/console/blob/master/frontend/packages/helm-plugin/src/components/list-page/HelmReleaseListRow.tsx#L21-L27):
```
        <Link
          to={`/helm-releases/ns/${obj.namespace}/release/${obj.name}`}
          title={obj.name}
          className="co-resource-item__resource-name"
        >
          {obj.name}
        </Link>
```

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.scss
```diff
@@ -0,0 +1,6 @@
+.syncStatus {
+    max-width: '100px'
+}
+.lastDeploymentTime {
+    color: var(--pf-global--palette--black-600);
+}
```
andrewballantyne, line 6: CSS needs to follow the [BEM naming structure](http://getbem.com/naming/) + have a package prefix... ie `.gop-` or something; we don't currently have one now that GitOps is in its own package.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
   classNames('pf-m-width-20'), // Application name
-  classNames('pf-m-width-40'), // Git repository
+  classNames('pf-m-width-30'), // Git repository
   classNames('pf-m-hidden', 'pf-m-visible-on-md', 'pf-m-width-20'), // Environments
-  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-20'), // Last deployment
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-30'), // Last deployment
 ];
 
-const handleClick = (appGroup: GitOpsAppGroupData) => {
+const handleClick = (appGroup) => {
```
andrewballantyne, line 23: Avoid removing types. Always swap them out for another one. If this is not the right type, we need to swap it out for another one or update the type. 

This may be moot for the `Link` implementation -- but if you keep a stand alone method for crafting the link url, it needs types. Stand-alone methods always need types as there is no TS smart typing possible.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
 const tableColumnClasses = [
   classNames('pf-m-width-20'), // Application name
-  classNames('pf-m-width-40'), // Git repository
+  classNames('pf-m-width-30'), // Git repository
   classNames('pf-m-hidden', 'pf-m-visible-on-md', 'pf-m-width-20'), // Environments
-  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-20'), // Last deployment
+  classNames('pf-m-hidden', 'pf-m-visible-on-lg', 'pf-m-width-30'), // Last deployment
 ];
 
-const handleClick = (appGroup: GitOpsAppGroupData) => {
```
reginapizza, line 17: adding `: GitOpsAppGroupData` was a change that @andrewballantyne had suggested on my PR yesterday, what's the reason for removing it?

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
+      if (latestDeployedTime < deployedTime) {
+        latestDeployedTime = deployedTime;
+        latestDeployedEnv = env;
+      }
+    });
+  }
+
+  if (latestDeployedTime === '') {
+    latestDeployedTime = null;
+  }
```
andrewballantyne, line 101: Both of these are falsy... any reason you want `null` specifically?

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -41,12 +119,65 @@ const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = (props) => {
...
+        ) : (
+          <span>-</span>
+        )}
       </TableData>
     </TableRow>
   );
 };
 
+const GetSyncFragment: React.FC<SyncProps> = ({ tooltip, count, icon }) => {
+  let targetIcon: any;
```
andrewballantyne, line 152:
```suggestion
  let targetIcon: React.ReactNode;
```

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -41,12 +119,65 @@ const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = (props) => {
...
+          </Flex>
+        ) : (
+          <span>-</span>
+        )}
       </TableData>
     </TableRow>
   );
 };
 
+const GetSyncFragment: React.FC<SyncProps> = ({ tooltip, count, icon }) => {
```
andrewballantyne, line 151: One Component per file please. Move this to a different file and import it.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
```diff
@@ -21,6 +21,13 @@ const GitOpsList: React.FC<GitOpsListProps> = ({ appGroups, emptyStateMsg }) =>
     return fuzzyCaseInsensitive(textFilter, name);
   });
 
+  let hasSyncStatus: boolean = false;
+  if (appGroups.length > 0) {
+    const appGroup = appGroups[0];
+    if (appGroup.sync_status) {
+      hasSyncStatus = true;
+    }
+  }
```
andrewballantyne, line 30: Any reason you don't want to loop all? You could simplify this if you look for the first one:

```suggestion
  let hasSyncStatus: boolean = appGroups?.some(({ sync_status }) => sync_status) || false;
```

Also `appGroups` can be null, best use optional chaining: `?.`

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: @reginapizza , it was not my intention to change this. Perhaps, when I applied my patch on top of your latest changes did this...I'll add it back.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
   const t = (tKey) => i18n.t(tKey);
+  let numOfSynced = 0;
+  let numOfOutOfSync = 0;
+  let numOfUnknown = 0;
+  const ttSynced = [];
+  const ttOutOfSync = [];
+  const ttUnknown = [];
+  let latestDeployedTime = '';
+  let latestDeployedEnv = '';
+  if (appGroup && appGroup.sync_status && appGroup.last_deployed) {
```
andrewballantyne, line 44: I think you want to do this differently... being more data-driven is more functional. This feels bulky. I'll do up a changeset and see if I can clean this up some.

andrewballantyne (changes requested):
Also run `yarn i18n` to fixup the test failure in the Frontend CI.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
andrewballantyne: Ah, already requested by Regina.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.scss
keithchong: I've made the changes, but haven't committed them yet.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Done.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -3,32 +3,110 @@ import i18n from '@console/internal/i18n';
...
+      } else if (status === 'Unknown') {
+        ttUnknown.push(
+          <Split hasGutter key={`${appGroup.name}-${env}`}>
+            <SplitItem>
+              <GrayUnknownIcon />
+            </SplitItem>
+            <SplitItem isFilled style={{ color: 'white' }}>
+              {env}
+            </SplitItem>
+            <SplitItem style={{ color: 'white' }}>{i18n.t('gitops-plugin~Unknown')}</SplitItem>
```
andrewballantyne, line 83: A lot of uses of `color: white`... could you just do up into a className and assign it to the surrounding div; ie `syncStatus` in the case of your current classNames.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
andrewballantyne: For posterity, I chopped this up and treated it as data-first... the rendering happens later in the return statement. I've provided the code to Keith in Slack DMs... but I figured I should add it here as well so others can learn from it too. 

<details>
  <summary>The Replacement Component Rendering</summary>
  
  The diff is hard to read, so here is the final state:
```
const getMatchingEnvs = (envs: string[], desiredStatus: string) => (
  acc: string[],
  status: string,
  idx: number,
): string[] =>
  desiredStatus === status
    ? [...acc, envs[idx]] // 1:1 between a status and an env
    : acc;

const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = (props) => {
  const { obj: appGroup, index, key, style } = props;
  const {
    name,
    sync_status: syncStatuses = [],
    environments: envs,
    last_deployed: lastDeployed = [],
    repo_url: repoUrl,
  } = appGroup;
  const t = (tKey) => i18n.t(tKey);

  const syncedEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'Synced'), []);
  const outOfSyncEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'OutOfSync'), []);
  const unknownEnvs: string[] = syncStatuses.reduce(getMatchingEnvs(envs, 'Unknown'), []);

  const latestDeployedTime = lastDeployed.reduce(
    (leadingDeployedTime, deployedTime) =>
      latestDeployedTime < deployedTime ? deployedTime : leadingDeployedTime,
    '',
  );
  const latestDeployedEnv = latestDeployedTime
    ? envs[lastDeployed.indexOf(latestDeployedTime)]
    : '';

  return (
    <TableRow id={index} index={index} trKey={key} style={style}>
      <TableData className={tableColumnClasses[0]}>
        <Button
          onClick={() => handleClick(appGroup)}
          aria-label={name}
          variant="link"
          style={{ padding: '0px' }}
        >
          {name}
        </Button>
      </TableData>
      <TableData className={classNames(tableColumnClasses[1])}>
        <ExternalLink href={repoUrl} additionalClassName={'co-break-all'}>
          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>
            {routeDecoratorIcon(repoUrl, 12, t)}
          </span>
          <span style={{ marginRight: 'var(--pf-global--spacer--xs)' }}>{repoUrl}</span>
        </ExternalLink>
      </TableData>
      <TableData className={classNames(tableColumnClasses[2], 'pf-u-text-nowrap')}>
        {syncStatuses ? (
          <Flex className="syncStatus">
            <GetSyncFragment
              tooltip={syncedEnvs.map((env) => (
                <Split hasGutter key={`${name}-${env}`}>
                  <SplitItem>
                    <GreenCheckCircleIcon />
                  </SplitItem>
                  <SplitItem isFilled style={{ color: 'white' }}>
                    {env}
                  </SplitItem>
                  <SplitItem style={{ color: 'white' }}>{t('gitops-plugin~Synced')}</SplitItem>
                </Split>
              ))}
              count={syncedEnvs.length}
              icon="check"
            />
            <GetSyncFragment
              tooltip={outOfSyncEnvs.map((env) => (
                <Split hasGutter key={`${name}-${env}`}>
                  <SplitItem>
                    <YellowExclamationTriangleIcon />
                  </SplitItem>
                  <SplitItem isFilled style={{ color: 'white' }}>
                    {env}
                  </SplitItem>
                  <SplitItem style={{ color: 'white' }}>{t('gitops-plugin~OutOfSync')}</SplitItem>
                </Split>
              ))}
              count={outOfSyncEnvs.length}
              icon="exclamation"
            />
            <GetSyncFragment
              tooltip={unknownEnvs.map((env) => (
                <Split hasGutter key={`${name}-${env}`}>
                  <SplitItem>
                    <GrayUnknownIcon />
                  </SplitItem>
                  <SplitItem isFilled style={{ color: 'white' }}>
                    {env}
                  </SplitItem>
                  <SplitItem style={{ color: 'white' }}>{t('gitops-plugin~Unknown')}</SplitItem>
                </Split>
              ))}
              count={unknownEnvs.length}
              icon="unknown"
            />
          </Flex>
        ) : (
          <span>{envs.join(', ')}</span>
        )}
      </TableData>
        {latestDeployedTime ? (
          <Flex>
            <FlexItem className="lastDeploymentTime" spacer={{ default: 'spacerXs' }}>
              <span>
                <Timestamp timestamp={latestDeployedTime} />
              </span>
            </FlexItem>
            <FlexItem>{latestDeployedEnv}</FlexItem>
          </Flex>
        ) : (
          <span>-</span>
        )}
      </TableData>
    </TableRow>
  );
};
```
</details>

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
```diff
@@ -41,12 +119,65 @@ const GitOpsTableRow: RowFunction<GitOpsAppGroupData> = (props) => {
...
+          <Flex>
+            <FlexItem>
+              {targetIcon} {count}
+            </FlexItem>
+          </Flex>
+        )}
+      </FlexItem>
+    </Flex>
+  );
+};
```
andrewballantyne, line 181: I just noticed this... this is a lot of Flex for zero(?) Flex usage. I think you can achieve this with just divs. 

```suggestion
    <div> // may need to do a display: inline-block class here
      {count > 0 ? (
        <Tooltip isContentLeftAligned content={<div>{tooltip}</div>}>
          {targetIcon} {count}
        </Tooltip>
      ) : (
        {targetIcon} {count}
      )}
    </div>
  );
};
```

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: I removed this. Done.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Done.

frontend/packages/gitops-plugin/src/components/list/GitOpsList.tsx
keithchong: Done

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Done. I made it on the parent `<Split>`.  That works.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: I was able to make all 'three columns' (consisting of the icon and number) to align vertically across rows.  I had tried removing the outer Flex, but that didn't work.  I can try <div>

andrewballantyne: You are likely suffering from the `display: block` problem. Flex likely go around that with the use of the flex styles. `display: inline-block;` should work... but if it does not, we can probably keep the outer Flex. The inner ones were unneeded.

keithchong: Hey Andrew, it didn't work.   I need the flex.  Also, I need an inner div or else there is an error.  eg. replace the flex/flexitem with a div

```
    <Flex flex={{ default: 'flex_1' }}>
      <FlexItem>
        {count > 0 ? (
          <Tooltip isContentLeftAligned content={<div>{tooltip}</div>}>
            <div>
              {targetIcon} {count}
            </div>
          </Tooltip>
        ) : (
          <div>
            {targetIcon} {count}
          </div>
        )}
      </FlexItem>
    </Flex>
```

andrewballantyne: Ah fair, that works then.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Done

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: I had to make one slight tweak.  Had to check that `syncStatus.length > 0` since it is an empty array initially, whereas before the sync status could be null.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Done. With this change, the handleClick method is removed too.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: The handleClick method is no longer need because of the above suggestion.

frontend/packages/gitops-plugin/src/components/list/GitOpsTableRow.tsx
keithchong: Thanks Andrew for the suggestion. And I thought that pushing the elements was elegant.

andrewballantyne (approved):
/lgtm

PR #8932, opened 2021-05-14
Bug 1942117: i18n Add Storage page

rebeccaalpert:
Some parts of the Add Storage page were missed during the first round of i18n.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1942117.


frontend/public/components/storage/attach-storage.tsx
```diff
@@ -53,25 +54,26 @@ const AttachStorage: React.FC<AttachStorageFormProps> = (props) => {
...
+          <h1 className="co-m-pane__heading">Add Storage </h1>
+          <div className="co-m-pane__explanation co-storage-heading__subtitle">
+            to{' '}
+            <ResourceLink
+              inline
+              kind={props?.kindObj?.kind}
+              name={match.params.name}
+              namespace={match.params.ns}
+            />
+          </div>
```
spadgett, line 75: It's probably better to move the space outside the `h1` element for the translators, although I'm not sure if that will change the margin

```suggestion
          <h1 className="co-m-pane__heading">Add Storage</h1>
          <div className="co-m-pane__explanation co-storage-heading__subtitle">
            {' '}to{' '}
            <ResourceLink
              inline
              kind={props?.kindObj?.kind}
              name={match.params.name}
              namespace={match.params.ns}
            />
          </div>
```

rebeccaalpert: All set!

spadgett (approved):
/lgtm

PR #8934, opened 2021-05-14
Bug 1960683: Fix hot loop in global config page

spadgett:
`oauthMenuItems` was initialized to a different value on every render,
and it was included in the dependency list of the `useEffect` hook. This
caused the hook to hot loop. Initialize `oauthMenuItems` inside
`useEffect` instead since it's only used in that block. Also remove
`errors` from the dependency list since it wasn't referenced inside
the hook.

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #8935, opened 2021-05-14
 Bug 1960686: Fix hot loop in global config page

spadgett:
Manual 4.7 backport of #8934

`oauthMenuItems` was initialized to a different value on every render,
and it was included in the dependency list of the `useEffect` hook. This
caused the hook to hot loop. Initialize `oauthMenuItems` inside
`useEffect` instead since it's only used in that block. Also remove
`errors` from the dependency list since it wasn't referenced inside
the hook.

/assign @rhamilto


dtaylor113 (approved):
/lgtm

rhamilto (approved):
/lgtm

zherman0 (approved):

yapei (approved):
/lgtm

yapei:
confirmed that global config resources are not requested in infinite loop

derekwaynecarr:
[patch-manager] :rocket: Approved for z-stream by score: 0.70

approved

PR #8940, opened 2021-05-14
Fix project access reload and form footer

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5815

**Root analysis:**
Currently, the reload button simply resets the form and doesn't fetch the latest role bindings. Also, it doesn't work with conflicts.

**Solution description:**
- Showing an info alert if the displayed list is stale
- fixed the form footer alignment

**GIF:**
pa

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -16,39 +20,58 @@ const ProjectAccessForm: React.FC<ProjectAccessFormProps> = ({
...
+      setIsStaleInfo(true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [roleBindings]);
+
+  const onReload = () => {
+    handleReset();
+    setIsStaleInfo(false);
+  };
+
```
divyanshiGupta, line 41:
```suggestion
 const disableSubmit = !dirty || !_.isEmpty(errors);
 const isStaleInfo = !_.isEqual(roleBindings, values);
```

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -16,39 +20,58 @@ const ProjectAccessForm: React.FC<ProjectAccessFormProps> = ({
...
-          errorMessage={status && status.submitError}
-          successMessage={status && !dirty && status.success}
-          disableSubmit={disableSubmit}
-          showAlert={!disableSubmit}
-          submitLabel={t('devconsole~Save')}
-          resetLabel={t('devconsole~Reload')}
-        />
       </div>
+      <FormFooter
+        handleReset={onReload}
```
divyanshiGupta, line 65:
```suggestion
        handleReset={handleReset}
```

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -16,39 +20,58 @@ const ProjectAccessForm: React.FC<ProjectAccessFormProps> = ({
...
+        isSubmitting={isSubmitting}
+        errorMessage={status && status.submitError}
+        successMessage={status && !dirty && status.success}
+        disableSubmit={disableSubmit}
+        showAlert={!disableSubmit}
+        submitLabel={t('devconsole~Save')}
+        resetLabel={t('devconsole~Reload')}
+        showStaleAlert={isStaleInfo}
```
divyanshiGupta, line 73: I dont think you need to add a new alert element in FormFooter for displaying this alert as it is an information alert which we already have in FormFooter. If the rolebindings have updated in the background we need to show only the  stale-info alert.
```suggestion
        showAlert={isStaleInfo || !disableSubmit}
        infoTitle={isStaleInfo && <title>}
        infoMessage={isStaleInfo && <message>}
```

Also if you want to have a separate alert for this, though there is no need at the moment, you can add a new prop to FormFooter that accepts an array of custom alert(s) and render it below the generic info-alert rather than adding a specific prop for each alert as there might be other custom alerts we want to show later in some use-case. FormFooter shouldn't be bothered to handle the custom alerts.

frontend/packages/console-shared/src/components/form-utils/FormFooter.tsx
```diff
@@ -57,6 +58,16 @@ const FormFooter: React.FC<FormFooterProps> = ({
...
+        {showStaleAlert && (
+          <Alert
+            isInline
+            className="co-alert"
+            variant="info"
+            title={t('console-shared~This object has been updated.')}
+          >
+            {t('console-shared~Click reload to see the new version.')}
+          </Alert>
+        )}
```
divyanshiGupta, line 70: you can remove this.
```suggestion
      
```

frontend/packages/console-shared/src/components/form-utils/form-utils-types.ts
```diff
@@ -12,6 +12,7 @@ export interface FormFooterProps {
   successMessage?: string;
   disableSubmit: boolean;
   showAlert?: boolean;
+  showStaleAlert?: boolean;
```
divyanshiGupta, line 15:
```suggestion

```

divyanshiGupta (changes requested):

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
debsmita1: I tried to follow what we do for the yaml editor. I wanted to show both the info alerts (`You made changes to this page.` & `This object has been updated.`) which is why I went ahead with a separate prop for showing a stale alert. But I could try out what you suggested i.e creating an array of custom alerts

Few questions here:
- do you expect the `Save` button to be disabled if the displayed list is stale?
- right now when we make changes in the form it shows an info alert saying changes have been made to the form. So, when the stale info alert is shown should we hide the above alert?

cc @serenamarie125 @christianvogt @jerolimov

divyanshiGupta: Yes, I think it makes sense to disable save when the currently shown data becomes stale as the user should not update it because it might cause conflicts. Also if we show stale-info alert, as I have mentioned above also, the user should not update the stale data so we shouldn't show obj-updated alert along stale-info alert. Since we shouldn't have both alerts showing at the same time I think we go can go with the first approach I suggested for this usecase.

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
debsmita1: @divyanshiGupta  The issue that I am trying to address here is, if the user has 2 tabs open and updates the role binding in tab 1 then the stale info alert should be shown in tab 2. The user should reload and fetch the latest list. The stale info alert is shown only when the project access list is updated other wise if the user edits the form it should still show `you-have-made-changes..`

I have updated this PR. Now, if the list is stale, the Save button is disabled. The user must reload the page to be able to add new role bindings

divyanshiGupta: I am aware about that but I didn't understand what it has to do with the changes that I suggested. The above changes will do the same thing with a single line addition. Have you tried it out? Let me know if it doesn't work for you.

debsmita1: @divyanshiGupta Yes, I have tried it out. with the above changes, when I try to do `add access` it shows the stale info alert instead of `you-have-made-changes` alert

divyanshiGupta: Okay in that you instead of using useEffect and also mainting a state you can simply use useMemo.

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
```diff
@@ -16,39 +20,59 @@ const ProjectAccessForm: React.FC<ProjectAccessFormProps> = ({
...
+    if (!_.isEqual(roleBindings, values)) {
+      setIsStaleInfo(true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [roleBindings]);
+
+  const onReload = () => {
+    handleReset();
+    setIsStaleInfo(false);
+  };
```
divyanshiGupta, line 40:
```suggestion
// eslint-disable-next-line react-hooks exhaustive-deps
const isStaleInfo = React.useMemo(() => !_.isEqual(roleBindings, values), [roleBindings]);
```

jerolimov:
@debsmita1 I don't know if this was already found in the comments above

* When changing a username and pressing save, everything works as expected.
* But when adding a new role binding the save it, I got an "successfully saved" message and then "This list has been updated." (on the current tab)

https://user-images.githubusercontent.com/139310/119061174-07ccce00-b9d4-11eb-9367-c13396cae13b.mov

frontend/packages/dev-console/src/components/project-access/ProjectAccessForm.tsx
debsmita1: With this change, if the list is updated then on reload the stale info alert does not get removed. This happens because on reload the `values` is updated and the useMemo here is triggered only when `roleBindings` is updated

frontend/packages/dev-console/src/components/project-access/__tests__/project-access-form-utils.spec.ts
```diff
@@ -18,4 +23,23 @@ describe('Fetch required roles', () => {
...
+
+  it('should ignore the role binding name from the role binding obj', async () => {
+    const roleBinding: UserRoleBinding[] = [
+      {
+        user: 'ab',
+        role: 'edit',
+      },
+      { user: 'kubeadmin', role: 'admin' },
+      { user: 'de', role: 'view' },
+    ];
```
divyanshiGupta, line 35: @debsmita1 here in the mock roleBindings you havent added roleBindingName in any obj so `ignoreRoleBindingName` is not getting tested fully IMO. The array is simply getting sorted which is not the only thing this function does.

debsmita1: Fixed

divyanshiGupta:
Screenshot 2021-06-01 at 6 35 36 PM

@debsmita1 Please remove the extra space being added by this PR as it is looking really weird.

debsmita1:
> Screenshot 2021-06-01 at 6 35 36 PM
> 
> @debsmita1 Please remove the extra space being added by this PR as it is looking really weird.

Fixed it !
Screenshot from 2021-06-01 21-11-26

wondering if we should keep the `<hr/>` , this form is similar to Pipeline Parameters & Resources forms and those forms do not have this horizontal line. Should we remove this ? cc @beaumorley 
Screenshot from 2021-06-01 21-12-17

divyanshiGupta:
> > Screenshot 2021-06-01 at 6 35 36 PM
> > @debsmita1 Please remove the extra space being added by this PR as it is looking really weird.
> 
> Fixed it !
> Screenshot from 2021-06-01 21-11-26
> 
> wondering if we should keep the `<hr/>` , this form is similar to Pipeline Parameters & Resources forms and those forms do not have this horizontal line. Should we remove this ? cc @beaumorley
> Screenshot from 2021-06-01 21-12-17

@beaumorley WDYT about this? Should we remove the horizontal separator from this form?

christianvogt:
@divyanshiGupta @beaumorley we can remove the line because we don't have this on our other forms

debsmita1:
> @divyanshiGupta @beaumorley we can remove the line because we don't have this on our other forms

Done
Screenshot from 2021-06-02 23-19-52

divyanshiGupta (approved):
/lgtm

frontend/packages/dev-console/src/components/project-access/ProjectAccess.tsx
```diff
@@ -95,7 +95,7 @@ const ProjectAccess: React.FC<ProjectAccessProps> = ({
           values: {
             projectAccess: values.projectAccess,
           },
-          status: { success: `Successfully updated the ${formName}.` },
+          status: { success: t('devconsole~Successfully updated the {{formName}}.', { formName }) },
```
jerolimov, line 98: I would remove the variable here and drop the prop `formName` from ProjectAccess. ProjectAccessPage sets this always to "project access".

This would also make the translation much easier.

jerolimov (approved):
Tested this locally. Works as expected. Also the issue when removing an entry is solved now.

The space mentioned above is already removed.

/approved

PR #8937, opened 2021-05-14
PF Update to 2021-6

jcaianirh:
Update pf packages to the latest release: 2021-6.

PR #8802, opened 2021-04-27
Bug 1930007: Filter dropdown doesn’t support multi selection

kdoberst:
This addresses [bug 1930007](https://bugzilla.redhat.com/show_bug.cgi?id=1930007)

The filter dropdown/select will remain open so the user can make multiple selections.  The filter/select dropdown can be closed by clicking on the filter button again or by clicking an item outside the filter dropdown/select.

https://user-images.githubusercontent.com/82059948/116313777-d0238b00-a773-11eb-9d95-e9cec776ea0f.mov

PR #8941, opened 2021-05-14
[WIP] Fix Cypress role binding test flake

dtaylor113:
Seems like clicking on Filter button opens filter dropdown, check's an item correctly, but dropdown stays open.  The attempt to type into filter input box closes open filter dropdown, but doesn't return focus for typing.

Added open and close filter dropdown commands before typing.

### DIFF
--- a/frontend/packages/integration-tests-cypress/views/list-page.ts
+++ b/frontend/packages/integration-tests-cypress/views/list-page.ts
@@ -27,10 +27,11 @@ export const listPage = {
       });
     },
     by: (rowFilter: string) => {
+      cy.byLegacyTestID('filter-dropdown-toggle')
+        .find('button')
+        .as('filterDropdownBtn');
       cy.get('.pf-c-toolbar__content-section').within(() => {
-        cy.byLegacyTestID('filter-dropdown-toggle')
-          .find('button')
-          .click();
+        cy.get('@filterDropdownBtn').click(); // open filter dropdown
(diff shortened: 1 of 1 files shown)


dtaylor113:
@jcaianirh brought this possible new flake to my attention, possibly brought on by new pf versions -he has contacted them.

dtaylor113:
...testing on clean master to verify if issue with new PF versions

jcaianirh (commented):
I think the tests correctly flagged a change in behavior...

spadgett:
@dtaylor113 The behavior probably changed with #8802

dtaylor113:
I think the issue with #8802 and the role binding test is that clicking in the filter input box causes the filter dropdown to close, but focus returns to filter dropdown button, not the filter input box:

https://www.awesomescreenshot.com/video/3748683

jcaianirh:
@dtaylor113 the test in question failed due to a valid issue with the [pf update ](https://github.com/openshift/console/pull/8937). Its being addressed by the pf folks, and can be updated in the that pr with new packages when the fix is ready. I don't think we need to take any action in the cypress tests.

dtaylor113:
Thanks @jcaianirh , closing due to "the test in question failed due to a valid issue with the pf update . Its being addressed by the pf folks, and can be updated in the that pr with new packages when the fix is ready. I don't think we need to take any action in the cypress tests."

PR #8943, opened 2021-05-17
Bug 1960531: Monitoring dashboards: Fix bug where panels could be duplicated

kyoto:
We should clone the panel so that the `row.panels.push` action isn't
altering the original panel's rows.


spadgett (approved):
/lgtm

PR #8949, opened 2021-05-17
CONSOLE-2458: Add endpoint for listing operators operands

jhadvig:
Adding new `/api/operands/<operatorName>.<operatorNamespace>`  endpoint that will return list of all operator's operands.

Screenshot 2021-05-17 at 22 30 43

/assign @spadgett 

ccing @dtaylor113


pkg/server/operands_handler.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to marshal the list operands response: %v", err)
+		klog.Error(errMsg)
+		serverutils.SendResponse(w, http.StatusInternalServerError, serverutils.ApiError{Err: errMsg})
+	}
+	w.Write(resp)
+}
+
+func parseOperatorNameAndNamespace(urlPath string) *OperatorMeta {
+	nameAndNamespace := strings.SplitN(urlPath, ".", 2)
```
spadgett, line 62: Can you share what the URL will look like? The name/namespace is split on a `.` character?

We could consider just making these values query parameters if it's easier to parse.

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,73 @@
...
+func operandsListHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		w.Header().Set("Allow", "GET")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Method unsupported, the only supported methods is GET"})
+		return
+	}
+	operatorMeta := parseOperatorNameAndNamespace(r.URL.Path)
+
+	cfg := &action.Configuration{}
+	err := cfg.Load()
```
spadgett, line 29: I'm not familiar with how this works. Does it discover the right configuration from the environment since it's running in a pod? Does this work for a development environment?

We'll want to support this for multicluster as well when we have the ACM integration.

pkg/server/operands_handler.go
jhadvig: the format is currently:
`<cluster-url>/api/list-operands/<operator-name>.<operator-namespace>`

the `<operator-name>.<operator-namespace>` is basically format what the `kubectl get operators` returns, since all operators are installed in a particualr NS, even if cluster operator.
eg.
`<cluster-url>/api/list-operands/etcd.openshift-etcd`

pkg/server/operands_handler.go
jhadvig: So from what I understand its getting the right configuration from the CLI, since its `kubectl-operator`  is a kubectl plugin. The `Load()` will use the oc/kubectl configuration context (namespace, ...) and also create needed clients. Since we are using it diretcly its best to call the `Load()` go get those clients, then set the namespace and then make the API call.

jhadvig: > We'll want to support this for multicluster as well when we have the ACM integration.

in that case the kubectl-operator has to be installed on all the managed clusters in order to list the opernds.

pkg/server/operands_handler.go
jhadvig: > We could consider just making these values query parameters if it's easier to parse.

yeah was considering that as well, but not sure if there is any benefit. I could imagine that if we would want to list operands for multiple operators by single call. Is that something we should  think about ?

spadgett: I'm concerned a period in a namespace name would cause problems. (I'm not sure offhand if that's allowed.)

The other benefit is that we don't have to parse the path. We just read the query parameters. So it's simpler. Since this is an internal console API, what we use doesn't matter too much, so I would do the simplest thing.

pkg/server/operands_handler.go
spadgett: OK, this sounds like it will be an issue in 4.9 since we won't have any kubectl configuration for spoke clusters when we implement multi-cluster. We need to work with the OLM team to understand if we can initialize the client another way (API server URL + CA file + access token).

Is this using the console service account token to make the request?

Does this work in a console development environment?

jhadvig: CC'ing @joelanford for more detailed answer. Thanks

joelanford: If `cfg.Load()` is run from a pod in a cluster and KUBECONFIG is unset, it _should_ load the in-cluster configuration from the mounted service account directory. If it doesn't, that's a bug that we would fix.

If you have more specific logic for where to pull the configuration from, the `action.Configuration` struct's fields are exported and can be manually configured as needed.

spadgett: @jhadvig Have you tested this with an in-cluster console? I don't see a `$HOME/.kube/config` or `$KUBECONFIG` set in the console pod, so I wouldn't expect it to work.

jhadvig: @spadgett right, but `Load()` is also setting the clients for the `Configuration` object.

jhadvig: > Does this work in a console development environment?

when running locally you have to have [installed the kubectl plugin ](https://github.com/operator-framework/kubectl-operator#install) and logged in into cluster via oc/kubectl

jhadvig: > Is this using the console service account token to make the request?

I think it should, but would rather double check with @joelanford

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,73 @@
...
+	// We need to manually set the namespace here since the kubectl-operator
+	// is setting the namespace to "default" upon calling the `cfg.Load()` method
+	if len(operatorMeta.Namespace) != 0 {
+		cfg.Namespace = operatorMeta.Namespace
+	}
+
+	operatorListOperands := action.NewOperatorListOperands(cfg)
+
+	operandsList, err := operatorListOperands.Run(context.TODO(), operatorMeta.Name)
+	if err != nil {
```
exdx, line 44: Looks like if any error is returned from the command it is treated the same. In https://github.com/operator-framework/kubectl-operator/pull/48 we introduce a new `OperandListError` that helps differentiate between different error types. It should help external callers.

jhadvig: @exdx thanks for the link. Question is if it make sense to surface these error msgs to the user?

Could be confusing since the RC for the request is `400`, at least for:
`CSV underlying operator is not in a succeeded state: custom resource list may not be accurate`

jhadvig:
@spadgett I’ve updated the https://github.com/openshift/console/pull/8949 …now all the configs are in play, meaning:
- KUBECONFIG
- in-cluster config
- $HOME/.kube/config

Also now the request should look like
/api/list-operands?name={operator-name}&namespace={operator-namespace}

pkg/server/operands_handler.go
jhadvig: Scratch that, I see we already are doing that.

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,100 @@
...
+func operandsListHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		w.Header().Set("Allow", "GET")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Method unsupported, the only supported methods is GET"})
+		return
+	}
+	query := r.URL.Query()
+	operatorName := query.Get("name")
+	operatorNamespace := query.Get("namespace")
+	if len(operatorName) < 1 || len(operatorNamespace) < 1 {
```
TheRealJon, line 50: Should there be a `return` at the end of this if block (and the other error handling blocks) so that we don't execute the rest of the logic in this func? I assume once we write the error response, we don't need to continue execution.

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,100 @@
...
+		errMsg := fmt.Sprintf("Failed to get operator name or namespace from the request URL: %q", r.URL.String())
+		klog.Error(errMsg)
+		serverutils.SendResponse(w, http.StatusInternalServerError, serverutils.ApiError{Err: errMsg})
+	}
+
```
TheRealJon, line 55: nit, extra newline.

```suggestion
```

TheRealJon (commented):
Just one question about making sure we halt execution of `operandsListHandler` after an error occurs. I might be misunderstanding something, so take it with a grain of salt.

pkg/server/operands_handler.go
jhadvig: Thanks @TheRealJon 👍  nice catch

jhadvig:
Thanks @TheRealJon ..comments addressed

jhadvig:
`operator-install-single-namespace.spe.ts` is failing, but it looks like a flake.

/retest

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,102 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetConfig() (*rest.Config, error) {
+	// If an env variable is specified with the config location, use that
+	if len(os.Getenv("KUBECONFIG")) > 0 {
+		return clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
+	}
+	// If no explicit location, try the in-cluster config
+	if c, err := rest.InClusterConfig(); err == nil {
```
spadgett, line 27: For multi-cluster, we're not going to have a kubeconfig file, and `InClusterConfig` won't work except for the hub cluster. Is there another way to initialize the client with only the API server URL and CA file? It would be good to determine that before merging or it will cause us problems down the road.

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,102 @@
...
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Method unsupported, the only supported methods is GET"})
+		return
+	}
+	query := r.URL.Query()
+	operatorName := query.Get("name")
+	operatorNamespace := query.Get("namespace")
+	if len(operatorName) < 1 || len(operatorNamespace) < 1 {
+		errMsg := fmt.Sprintf("Failed to get operator name or namespace from the request URL: %q", r.URL.String())
+		klog.Error(errMsg)
+		serverutils.SendResponse(w, http.StatusInternalServerError, serverutils.ApiError{Err: errMsg})
```
spadgett, line 53: This should be a 400 bad request rather than an internal server error.

pkg/server/operands_handler.go
```diff
@@ -0,0 +1,102 @@
...
+		Client:    client,
+		Scheme:    scheme,
+		Namespace: operatorNamespace,
+	}
+	operatorListOperands := action.NewOperatorListOperands(cfg)
+	operandsList, err := operatorListOperands.Run(context.TODO(), operatorName)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to list operands: %v", err)
+		klog.Error(errMsg)
+		serverutils.SendResponse(w, http.StatusBadRequest, serverutils.ApiError{Err: errMsg})
```
spadgett, line 89: Bad request doesn't seem accurate for all cases here. If we need a general error and don't have a better status code, maybe 502 Bad Gateway?

jhadvig:
@spadgett comments addressed

pkg/server/operands_handler.go
spadgett: OK, do you know if we can pass a CA certificate as well?

jhadvig:
Closing this PR in favour of https://github.com/openshift/console/pull/9142

PR #8952, opened 2021-05-18
Bug 1961452: Monitoring: Remove "Alertmanager UI" link

kyoto:
The Alertmanager UI is still accessible via its route, but all necessary
information should be available through the Console admin UI.

FYI @sichvoge, @lbarbeevargas, @cshinn


spadgett (approved):
/lgtm

PR #8959, opened 2021-05-18
Re enable customize boot source tests

yaacov:
Re enable customize common template source E2E tests


yaacov:
@glekner @gouyang please review

PR #8960, opened 2021-05-18
Bug 1961675: Hide TaskRun edit actions for rows in Pipelinerun's TaskRun tab

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5736

**Analysis / Root cause**: 
User should only be able to Delete a taskrun, but the taskrun kebab menu includes options to Edit lable, annotation, and taskrun.

**Solution Description**: 
Hide taskrun edit actions for rows in pipelinerun's taskrun tab.

(Unchanged)

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import { TableRow, TableData, RowFunction } from '@console/internal/components/factory';
-import { ResourceLink, Timestamp, Kebab, ResourceKebab } from '@console/internal/components/utils';
+import { ResourceLink, Timestamp, ResourceKebab } from '@console/internal/components/utils';
 import { referenceForModel } from '@console/internal/module/k8s';
 import { TaskRunKind } from '../../../types';
 import { getModelReferenceFromTaskKind } from '../../../utils/pipeline-augment';
+import { getTaskRunKebabActions } from '../../../utils/pipeline-actions';
 import { TaskRunModel, PipelineModel } from '../../../models';
 import { tableColumnClasses } from './taskruns-table';
 import { taskRunFilterReducer } from '../../../utils/pipeline-filter-reducer';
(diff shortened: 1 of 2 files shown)


frontend/packages/pipelines-plugin/src/components/taskruns/list-page/TaskRunsRow.tsx
```diff
@@ -66,7 +66,11 @@ const TaskRunsRow: RowFunction<TaskRunKind> = ({ obj, index, key, style, ...prop
       <Timestamp timestamp={obj?.status?.startTime} />
     </TableData>
     <TableData className={tableColumnClasses[7]}>
-      <ResourceKebab actions={Kebab.factory.common} kind={taskRunsReference} resource={obj} />
+      <ResourceKebab
+        actions={props.customData?.hideEditActions ? [Kebab.factory.Delete] : Kebab.factory.common}
```
karthikjeeyar, line 70: In every instance of `TaskRun`'s kebab action,  we only need to have `Delete TaskRun` option. I don't see a place where we might need to enable edit action. So It could be a simple change `kebab.factory.common` to `kebab.factory.Delete`. @andrewballantyne  any thoughts on this ?

karthikjeeyar: I see that you are showing the Edit options on admin side, but we shouldn't show it even there because it redirects to a yaml page that doesn't have `save`  button. 
image

andrewballantyne (changes requested):
As Karthik noted, we don't want Edit actions EVER for TaskRuns... they are end-stage, execution objects. Editing them serves no purpose as they are merely a result of an execution.

PipelineRuns don't have this issue because we had to build in the custom "rerun" option and that allowed us to address this concern then. This is just to catch up for TaskRuns. Hard update the kebabOptions to only say Delete. 

If we want to super pure here, create a function in `frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx` called `getTaskRunKebabActions` and just put Delete in there. But honestly that's optional from my point of view.

andrewballantyne (approved):
Adding lgtm as well as Karthik is off on PTO now until next week.

/lgtm

Looks good, checked Search, PipelineRun's TaskRun tab, Admin TaskRun list, and the TaskRun list pages. All are just delete 👍

PR #8962, opened 2021-05-18
Bug 1926029: Disable snapshot save button when save isn't possible

pcbailey:
Currently, the user is able to click the "Save" button inside of the snapshot modal when there are no disks from which a snapshot can be made. This PR displays a warning alert when no snapshot-compatible disks are found and disables the save button.

Current modal when no viable disks are found:
Selection_743

New modal when no viable disks are found:
Selection_744


yaacov:
/retest

p.s.
the convention we use now is to name files that hold one react component in camel case e.g. `HelloWorld.tsx` instead of `hello-world.tsx` since we are adding new files, maybe we can move the modal to use the new convention ?

yaacov:
@yfrimanm PTAL

yfrimanm:
Thank you @pcbailey I think the new modal is a great improvement. Just 2 nits: in the second paragraph maybe we can add the words 'either' and 'or' to make it a bit clearer, so it will be like this:
"You can either edit an existing disk to add a snapshot-supported storage class, or add a new disk with a compatible storage class defined, or contact your cluster admin for further details."

pcbailey:
@yfrimanm Thanks! Having two "or"' options after "either" reads oddly to me. Typically, "either" is used with a binary choice, either one thing or the other. While it may be grammatically correct in the strictest sense, it's likely that it will add additional cognitive overhead for users that aren't used to seeing it used in this way. Could we add clarity in a different way? Maybe something like a bulleted list?

@yaacov I'll absolutely fix it here. I must have forgotten that we were moving to the new convention. I apologize!

yfrimanm:
Thanks @pcbailey https://github.com/pcbailey

I think that a separation in this long sentence could make it easier for the user to understand what the options are, so having the 3 options in a bulleted list could work.
How about something like this:

To take snapshot you can either 
1. Edit an existing disk to add a snapshot-supported storage class
2. Add a new disk with a compatible storage class defined
3. Contact your cluster admin for further details

WDYT?

pcbailey:
@yfrimanm Now that I see it in list form, I think it may be best to break it into two separate sentences. What do you think about the following?

"To take a snapshot you can either edit an existing disk to add a snapshot-supported storage class or add a new disk with a compatible storage class defined. For further details, please contact your cluster admin."

yfrimanm:
Thank you @pcbailey LGTM!

pcbailey:
@yfrimanm Thanks! I have updated the commit and the screenshot.

@yaacov I update the files names to meet convention.

yaacov:
/hold

@pcbailey can you check the failing test, it looks like it's actually failing because of the change ?

PR #8965, opened 2021-05-18
fix dev-catalog type tooltip

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5819

**Root analysis:**
The tooltip displays off the screen so text cannot be read. There is no way to scroll to see the unreadable text.

**Solution description:**
- passed `inline` property to `SyncMarkdownView`
- showing the description of the catalog item only if it's available

**GIF:**
dev-cat

**Browser conformance:**
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogTypeSelector.tsx
+++ b/frontend/packages/dev-console/src/components/catalog/catalog-view/CatalogTypeSelector.tsx
@@ -21,7 +21,12 @@ const CatalogTypeSelector: React.FC<CatalogTypeSelectorProps> = ({
 
   const typeDescriptions = React.useMemo(
     () =>
-      catalogTypes.map((type) => <SyncMarkdownView key={type.value} content={type.description} />),
+      catalogTypes.map(
+        (type) =>
+          type.description && (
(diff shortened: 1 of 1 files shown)


beaumorley:
lgtm. thanks @debsmita1

debsmita1:
> @beaumorley I disagree with the lgtm. I don't believe that any popover should ever look this way with a scrollbar in the middle of the view and text cut off at the at top and bottom. The scrollbar is out of place.
> 
> @debsmita1 the reason for the original bug because the content of the popover is loading asynchronously. The popover requires the content to be present at render time in order to correctly determine the size and position. Why is it asynchronous? Because the markdown view was setup to render in an iframe (async) instead of inline.
> 
> The fix here is to set `inline` property on `SyncMarkdownView` in `CatalogTypeSelector.tsx`.
> Along with this change you will need to remove `popoverHasAutoWidth` from the `FieldLevelHelp`. The result:
> 
> image

Fixed !

beaumorley:
Hi @christianvogt thanks for your input on this and for pushing for the better design. Duly noted!

invincibleJai:
/lgtm

PR #8966, opened 2021-05-18
Add useDynamicPluginInfo hook

vojtechszocs:
This PR adds new React hook `useDynamicPluginInfo` which allows consuming Console [dynamic plugin](https://github.com/openshift/console/tree/master/frontend/packages/console-dynamic-plugin-sdk) runtime information.

This hook re-renders the given component in any of the following cases:
- dynamic plugin(s) finished loading - status change `Pending` :arrow_forward: `Loaded`
- dynamic plugin(s) failed to load - status change `Pending` :arrow_forward: `Failed`

Hook return type is `DynamicPluginInfo[]` with `status: 'Pending' | 'Loaded' | 'Failed'` element property as the common denominator. Plugin metadata & enabled flag is available only for loaded plugins.

@spadgett FYI, this modifies `PluginStore` to allow loading only dynamic plugins listed via `SERVER_FLAGS.consolePlugins`.

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/.gitignore
+++ b/frontend/packages/console-dynamic-plugin-sdk/.gitignore
@@ -1,1 +1,1 @@
-schema
+/schema
--- a/frontend/packages/console-plugin-sdk/src/api/useDynamicPluginInfo.ts
+++ b/frontend/packages/console-plugin-sdk/src/api/useDynamicPluginInfo.ts
@@ -1,0 +1,59 @@
+import * as React from 'react';
+import { useForceRender } from '@console/shared/src/hooks/useForceRender';
+import { DynamicPluginInfo } from '../store';
+import { subscribeToDynamicPlugins } from './pluginSubscriptionService';
+
+/**
+ * React hook for consuming Console dynamic plugin runtime information.
+ *
+ * When the runtime status of a dynamic plugin changes, the React component
+ * is re-rendered with the hook returning an up-to-date plugin information.
+ *
+ * Example usage:
+ *
+ * ```ts
+ * const Example = () => {
+ *   const pluginEntries = useDynamicPluginInfo();
+ *   // process plugin entries and render your component
+ * };
+ * ```
+ *
+ * The hook's result is guaranteed to be referentially stable across re-renders.
+ *
+ * @returns Console dynamic plugin runtime information.
+ */
+export const useDynamicPluginInfo = (): DynamicPluginInfo[] => {
+  const forceRender = useForceRender();
+
+  const isMountedRef = React.useRef(true);
+  const unsubscribeRef = React.useRef<VoidFunction>(null);
+  const pluginEntriesRef = React.useRef<DynamicPluginInfo[]>([]);
+
+  const trySubscribe = React.useCallback(() => {
+    if (unsubscribeRef.current === null) {
+      unsubscribeRef.current = subscribeToDynamicPlugins((pluginEntries) => {
+        pluginEntriesRef.current = pluginEntries;
+        isMountedRef.current && forceRender();
+      });
+    }
+  }, [forceRender]);
+
+  const tryUnsubscribe = React.useCallback(() => {
+    if (unsubscribeRef.current !== null) {
+      unsubscribeRef.current();
+      unsubscribeRef.current = null;
+    }
+  }, []);
+
+  trySubscribe();
+
+  React.useEffect(
+    () => () => {
+      isMountedRef.current = false;
+      tryUnsubscribe();
+    },
+    [tryUnsubscribe],
+  );
+
+  return pluginEntriesRef.current;
+};
--- a/frontend/packages/console-plugin-sdk/src/store.ts
+++ b/frontend/packages/console-plugin-sdk/src/store.ts
@@ -34,7 +34,14 @@ export const getGatingFlagNames = (extensions: Extension[]): string[] =>
   ]);
 
 /**
- * Provides access to Console plugin data.
+ * Provides access to Console plugins and their extensions.
+ *
+ * Only plugins listed via `allowedDynamicPluginNames` can be added dynamically at runtime.
+ *
+ * Subscribed `listeners` are invoked upon any of the following events:
+ *
+ * - when the list of extensions which are currently in use changes
+ * - when the runtime status of a dynamic plugin changes
  *
  * In development, this object is exposed as `window.pluginStore` for easier debugging.
  */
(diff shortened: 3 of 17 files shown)


frontend/packages/console-dynamic-plugin-sdk/.gitignore
```diff
@@ -1 +1 @@
-schema
+/schema
```
vojtechszocs, line 1: @rawagner FYI - this fixes an issue where Git would ignore `schema` directory at any level under the SDK package.

frontend/packages/console-plugin-sdk/src/store.ts
```diff
@@ -155,8 +233,21 @@ type DynamicPluginManifest = Readonly<ConsolePluginManifestJSON>;
...
+  pluginID: string;
+  enabled: boolean;
+};
+
+type NotLoadedDynamicPluginInfo = {
+  status: 'Pending' | 'Failed';
+  pluginName: string;
+};
+
+export type DynamicPluginInfo = LoadedDynamicPluginInfo | NotLoadedDynamicPluginInfo;
```
rhamilto, line 253: Seeing the following TypeScript errors:
```
ERROR in /Users/rhamilto/Git/console/frontend/packages/operator-lifecycle-manager/src/components/console-operator-config.tsx(51,54):
TS2339: Property 'name' does not exist on type 'DynamicPluginInfo'.
  Property 'name' does not exist on type 'NotLoadedDynamicPluginInfo'.
ERROR in /Users/rhamilto/Git/console/frontend/packages/operator-lifecycle-manager/src/components/console-operator-config.tsx(54,24):
TS2339: Property 'version' does not exist on type 'DynamicPluginInfo'.
  Property 'version' does not exist on type 'NotLoadedDynamicPluginInfo'.
ERROR in /Users/rhamilto/Git/console/frontend/packages/operator-lifecycle-manager/src/components/console-operator-config.tsx(55,28):
TS2339: Property 'description' does not exist on type 'DynamicPluginInfo'.
  Property 'description' does not exist on type 'NotLoadedDynamicPluginInfo'.
```
with the following code:
```
const ConsolePluginsList: React.FC<ConsolePluginsListType> = ({ obj }) => {
  const [consolePlugins, consolePluginsLoaded] = useK8sWatchResource<ConsolePluginKind[]>({
    isList: true,
    kind: referenceForModel(ConsolePluginModel),
  });
  const dynamicPluginInfo: DynamicPluginInfo[] = useDynamicPluginInfo();
  const [rows, setRows] = React.useState([]);
  const [sortBy, setSortBy] = React.useState<ISortBy>({});
  React.useEffect(() => {
    const data = consolePlugins.map((plugin) => {
      const info = dynamicPluginInfo?.find((i) => i?.name === plugin?.metadata?.name);
      return {
        name: plugin?.metadata?.name,
        version: info?.version,
        description: info?.description,
        status: info?.status,
      };
    });
    setRows(
      data?.map((item) => {
        const placeholder = '-';
        return [
          item.name,
          item.version || placeholder,
          item.description || placeholder,
          item?.status || placeholder,
        ];
      }),
    );
  }, [consolePlugins, dynamicPluginInfo]);
...
```
Something is incorrect with the `type` or I am using it incorrectly?

vojtechszocs: @rhamilto This is because the overlap of type union `LoadedDynamicPluginInfo | NotLoadedDynamicPluginInfo` is just the `status` field.

I've updated the PR and added `isLoadedDynamicPluginInfo` type guard which narrows `DynamicPluginInfo` down to `LoadedDynamicPluginInfo` for easier consumption.

```ts
const dynamicPluginInfo = useDynamicPluginInfo();

// inside consolePlugins.map() transformer
const loadedPluginInfo = dynamicPluginInfo
  .filter(isLoadedDynamicPluginInfo)
  .find((i) => i.name === plugin.metadata.name);

return {
  name: plugin.metadata.name,
  version: loadedPluginInfo?.version || '<not available>',
  description: loadedPluginInfo?.description || '<not available>',
};
```

frontend/packages/console-plugin-sdk/src/api/useDynamicPluginInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+ * The hook's result is guaranteed to be referentially stable across re-renders.
+ *
+ * @returns Console dynamic plugin runtime information.
+ */
+export const useDynamicPluginInfo = (): DynamicPluginInfo[] => {
+  const forceRender = useForceRender();
+
+  const isMountedRef = React.useRef(true);
+  const unsubscribeRef = React.useRef<VoidFunction>(null);
+  const pluginInfoEntriesRef = React.useRef<DynamicPluginInfo[]>([]);
```
christianvogt, line 21: Couldn't we use `useState` and avoid the need for `useForceRender` ?
Looks like this is a similar pattern to `useExtenions`.

frontend/packages/console-plugin-sdk/src/store.ts
```diff
@@ -155,8 +233,25 @@ type DynamicPluginManifest = Readonly<ConsolePluginManifestJSON>;
...
+  status: 'Loaded';
+  pluginID: string;
+  metadata: DynamicPluginMetadata;
+  enabled: boolean;
+};
+
+export type NotLoadedDynamicPluginInfo = {
+  status: 'Pending' | 'Failed';
+  pluginName: string;
+};
```
christianvogt, line 252: Why don't they both use `pluginID` or `pluginName`; why the difference?

rhamilto:
@vojtechszocs, now that 4.9 is open, can you update so I can move forward with https://github.com/openshift/console/pull/8919?

vojtechszocs:
> @vojtechszocs, now that 4.9 is open, can you update so I can move forward with #8919?

Sorry for the late response! I'll rebase this PR and address Christian's comments. (Almost done with #9310.)

vojtechszocs:
/label px-approved
/label docs-approved

We have a general px- and docs-approval for Console dynamic plugin changes.

vojtechszocs:
@christianvogt Switched back to `useRef` since `useState` will always trigger a render with provided initial value (or `undefined` if called without an argument). We want our `useDynamicPluginInfo` hook's initial value to reflect what `subscribeToDynamicPlugins` returns (it calls the listener immediately, so the user will never get `[]` as the hook's result, assuming that `window.SERVER_FLAGS.consolePlugins` is not an empty array).

Here's the change diff
```diff
@@ -22,17 +23,20 @@ import { subscribeToDynamicPlugins } from './pluginSubscriptionService';
  * @returns Console dynamic plugin runtime information.
  */
 export const useDynamicPluginInfo = (): DynamicPluginInfo[] => {
+  const forceRender = useForceRender();
+
   const isMountedRef = React.useRef(true);
   const unsubscribeRef = React.useRef<VoidFunction>(null);
-  const [pluginEntries, setPluginEntries] = React.useState<DynamicPluginInfo[]>([]);
+  const pluginEntriesRef = React.useRef<DynamicPluginInfo[]>([]);
 
   const trySubscribe = React.useCallback(() => {
     if (unsubscribeRef.current === null) {
-      unsubscribeRef.current = subscribeToDynamicPlugins((entries) => {
-        isMountedRef.current && setPluginEntries(entries);
+      unsubscribeRef.current = subscribeToDynamicPlugins((pluginEntries) => {
+        pluginEntriesRef.current = pluginEntries;
+        isMountedRef.current && forceRender();
       });
     }
-  }, [setPluginEntries]);
+  }, [forceRender]);
 
   const tryUnsubscribe = React.useCallback(() => {
     if (unsubscribeRef.current !== null) {
@@ -51,5 +55,5 @@ export const useDynamicPluginInfo = (): DynamicPluginInfo[] => {
     [tryUnsubscribe],
   );
 
-  return pluginEntries;
+  return pluginEntriesRef.current;
 };
```

frontend/packages/console-plugin-sdk/src/store.ts
christianvogt: vojtech already addressed this in another comment.

frontend/packages/console-plugin-sdk/src/api/useDynamicPluginInfo.ts
christianvogt: this was only a suggestion but it seems based on the requirement to compute an initial state through other means, a ref is needed.

vojtechszocs:
This PR is now included in & tested as part of #8919.

PR #8950, opened 2021-05-18
Add getting started section to the add page in developer perspective

jerolimov:
**Fixes**: 
Epic: https://issues.redhat.com/browse/ODC-5013
Story: https://issues.redhat.com/browse/ODC-5560

**Analysis / Root cause**: 
As a user, I want to see the new getting started section with areas to "Set up your cluster", Quick Starts, Guided Tours and more Learning resources.

**Solution Description**: 
This PR adds a new section to "+Add" page in the developer perspective above all options to import different types of resources. (This grid of add actions are redesigned as part of issue #8867.)

This new "Getting started" section adds three content cards with different links:

1. "Create applications using samples" - The first column is based on the developer catalog and shows two samples. The code contains two hardcoded "featured" samples (Quarkus and Sprint Boot) which should be preferred. When this samples are not found, some other samples will be displayed.
2. "Build with guided documentation" - This column shows up to two quick starts. It also prefers two hardcoded quickstarts (Quarkus and Sprint Boot again). It but will automatically hide them when the user completed them. When all quick started are completed, this column disappears.
3. "Explore new developer features" shows some static links. Currently it shows a link to `https://developers.redhat.com/products/openshift/getting-started` and two options to open the topology. The first link opens also opens the "Add to project" search bar.

It contains an option to hide this getting started section.

*Note:* The button too restore this section is not correct aligned, as this need some changes from #8867. Depending on which PR is merged first, this PR here or #8867 need to address this.

**Screen shots / Gifs for design review**: 
@openshift/team-devconsole-ux

image

Peek 2021-05-18 02-26

**Unit test coverage report**: 
image

image

**Test setup:**
* Open the developer perspective, Add Page.
* Check the different options.
* Complete a shown quick start and check that another quick start is shown.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #8867, opened 2021-05-07
Redesign add page to follow masonry layout

nemesis09:
**Jira issues:**
Epic: https://issues.redhat.com/browse/ODC-5013
Story: https://issues.redhat.com/browse/ODC-5561, https://issues.redhat.com/browse/ODC-5417
Bug: https://issues.redhat.com/browse/ODC-5788

**Problem/Description:**
Redesign Add page to support masonry layout

**Solution:**
- Categorize add extensions into groups.
- Add a card for each group in a masonry layout.
- Add data-test-ids wherever required.

**Screens:**
__Add Page redesigned__
add-page-redesign

__Add Page redesigned: responsive__
add-page-redesign-responsive

**Test Coverage:**
Screenshot from 2021-05-10 21-34-05
Screenshot from 2021-05-10 21-35-24
Screenshot from 2021-05-10 21-36-00

**Browser Conformance:**

- [x] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

Note: ~~This PR skips an integration test in kubevirt plugin which checks for Quick starts. It will be reinstated with the PR https://github.com/openshift/console/pull/8950 for Getting Started Section.
https://github.com/openshift/console/blob/1ff04963549964e5f467516622b1e4baf3183597/frontend/packages/kubevirt-plugin/integration-tests-cypress/tests/devconsole/devconsole.spec.ts#L69~~
No change to integration tests. The concerned test has been updated in [PR 8970](https://github.com/openshift/console/pull/8970)

PR #8970, opened 2021-05-19
remove quick starts check on Add page

gouyang:


yaacov:
/lgtm

Temporary removing dependency on QS for -
https://github.com/openshift/console/pull/8867

Will re test QS after
https://github.com/openshift/console/pull/8950

cc:// @jerolimov @nemesis09

jerolimov:
@gouyang @yaacov Also #8950 is now merged. So you can revert this change now again (or we can provide a PR). But is it really required that the "kubevirt" tests checks for quick starts???

The new add page looks like this, so you could check for example for "View all quick starts" instead of "Quick Starts".

image

PR #8973, opened 2021-05-19
Bug 1957457: Support secondary text in ResourceDropdown

bipuladh:
Add namespace in Multus Dropdown
Before:
Screenshot from 2021-05-19 13-13-38
After:
Screenshot from 2021-05-22 01-28-09

### DIFF
--- a/frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
+++ b/frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
@@ -16,18 +16,27 @@ import {
   modelFor,
   referenceFor,
 } from '@console/internal/module/k8s';
+import { getNamespace } from '../../selectors';
 
 type DropdownItemProps = {
   model: K8sKind;
   name: string;
+  namespace?: string;
 };
 
-const DropdownItem: React.FC<DropdownItemProps> = ({ model, name }) => (
+const DropdownItem: React.FC<DropdownItemProps> = ({ model, name, namespace }) => (
   <span className="co-resource-item">
-    <span>
+    <span className="co-resource-icon--fixed-width">
       <ResourceIcon kind={referenceForModel(model)} />
     </span>
-    <span className="co-truncate show co-nowrap small">{name}</span>
+    <span className="co-resource-item__resource-name">
+      <span>{name}</span>
+      {namespace && (
+        <span className="text-muted co-truncate show co-nowrap small co-resource-item__resource-namespace">
+          {namespace}
+        </span>
+      )}
+    </span>
   </span>
 );
 
@@ -200,6 +209,8 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
     } = props;
 
     const unsortedList = { ...appendItems };
+    const namespaces = new Set(_.flatten(_.map(resources, ({ data }) => data?.map(getNamespace))));
+    const containsMultipleNs = namespaces.size > 1;
     _.each(resources, ({ data, kind }) => {
       _.reduce(
         data,
@@ -209,8 +220,14 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
           if (dataValue) {
             if (showBadge) {
               const model = modelFor(referenceFor(resource)) || (kind && modelFor(kind));
+              const namespace = containsMultipleNs ? getNamespace(resource) : null;
               acc[dataValue] = model ? (
-                <DropdownItem key={resource.metadata.uid} model={model} name={name} />
+                <DropdownItem
+                  key={resource.metadata.uid}
+                  model={model}
+                  name={name}
+                  namespace={namespace}
+                />
               ) : (
                 name
               );
--- a/frontend/public/components/_dropdown.scss
+++ b/frontend/public/components/_dropdown.scss
@@ -91,7 +91,7 @@ $dropdown-background--hover: $color-pf-black-200; // pf-c-dropdown__menu-item--h
       color: inherit;
       padding: 4px 0 2px !important;
 
-      @media(min-width: $grid-float-breakpoint) {
+      @media (min-width: $grid-float-breakpoint) {
         padding-bottom: 10px !important;
         padding-top: 11px !important;
       }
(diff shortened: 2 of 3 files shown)


yuvalgalanti:
@bipuladh  is the namespace information is needed for the user for taking a decision on which network to choose?

bipuladh:
> @bipuladh is the namespace information is needed for the user for taking a decision on which network to choose?

According to @rohan47 and @ebondare yes.

yuvalgalanti:
usually, we use this pattern to give more information on a specific resource, so I think we can add that. WDYT on "the network is part of **openshift-storage** namespace"?

bipuladh:
> usually, we use this pattern to give more information on a specific resource, so I think we can add that. WDYT on "part of **openshift-storage** namespace"?

Well we show from two different namsepaces openshift-storage and default.

spadgett:
I wonder if the dropdown should always show the namespace when querying namespaced resources across all namespaces, rather than adding the secondary text prop.

To be more consistent with how we show API groups as secondary text in some dropdown, we might just show the namespace without the prefix. I think you would probably recognize the namespaces you're working in and repeating "Namespace:" for every item adds clutter.

Screen Shot 2021-05-21 at 3 36 46 PM

spadgett:
We should use the same margins as the other dropdown as well where the secondary text is aligned with the primary label (rather than below the icon).

bipuladh:
> I wonder if the dropdown should always show the namespace when querying namespaced resources across all namespaces, rather than adding the secondary text prop.
IMO it would be better to allow users to specify what they want to show. For example: In the future we want to show the subnet of the NAD along with the namespace. So making it configurable would make things easier.

spadgett:
> IMO it would be better to allow users to specify what they want to show. For example: In the future we want to show the subnet of the NAD along with the namespace. So making it configurable would make things easier.

I think we should add that feature when we need it. The fix you have should apply to any resource dropdown that shows resources across namespaces. The current PR would only fix one specific case instead of all of them.

bipuladh:
Added CSS to disable namespace from showing in the button.
Before:
Screenshot from 2021-05-25 01-22-12
After:
Screenshot from 2021-05-25 01-32-49

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -200,6 +209,9 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
     } = props;
 
     const unsortedList = { ...appendItems };
+    const containsMultipleNs =
+      _.uniq(_.flatten(Object.values(resources || {})?.map(({ data }) => data?.map(getNamespace))))
+        ?.length > 1;
```
spadgett, line 214: I think something like this is simpler (not tested)

```ts
const namespaces = new Set(resources?.map(getNamespace));
const showNamespace = namespaces.size > 1;
```

spadgett: Or it's `resources` is a object you could use `_.map`

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -200,6 +209,8 @@ class ResourceDropdown extends React.Component<ResourceDropdownProps, State> {
     } = props;
 
     const unsortedList = { ...appendItems };
+    const namespaces = new Set(_.flatten(_.map(resources, ({ data }) => data?.map(getNamespace))));
```
spadgett, line 212: What does the data look like? I'm unsure why we need `flatten` and multiple calls to `map`.

frontend/packages/console-shared/src/components/dropdown/ResourceDropdown.tsx
```diff
@@ -16,18 +16,27 @@ import {
...
   <span className="co-resource-item">
-    <span>
+    <span className="co-resource-icon--fixed-width">
       <ResourceIcon kind={referenceForModel(model)} />
     </span>
-    <span className="co-truncate show co-nowrap small">{name}</span>
+    <span className="co-resource-item__resource-name">
+      <span>{name}</span>
+      {namespace && (
+        <span className="co-resource-item__resource-api text-muted co-truncate show co-nowrap small">
```
spadgett, line 35: We should generalize the `co-resource-item__resource-api` class name if we're using it in a different context.

bipuladh: This class is not required at all removing the usage of it from here.

spadgett (approved):
/lgtm

Thanks @bipuladh 👍

frontend/public/components/_dropdown.scss
```diff
@@ -260,5 +258,8 @@ $dropdown-background--hover: $color-pf-black-200; // pf-c-dropdown__menu-item--h
   .co-resource-item__resource-name {
     overflow: hidden;
     text-overflow: ellipsis;
+    .text-muted {
```
spadgett, line 261: I think we want to target this one specific case instead of hiding any muted text.

bipuladh: This targets the muted text inside a button inside of a resource-name item. The outer class is a class applied to Dropdown button.

spadgett: Right, I think that's still too broad. We want to only target this one specific case. We use `co-resource-item__resource-name` in many places.

spadgett: For instance, here's a place we use it in a different dropdown:

https://github.com/TheRealJon/console/blob/62b5fec283068b665c3ade8195150762bfa84cf3/frontend/public/components/resource-dropdown.tsx#L43

bipuladh: The section you mentioned wouldn't get applied this style because the parent class in not `pf-c-dropdown__toggle-text`

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/_configure.scss
```diff
@@ -13,4 +13,7 @@
 
 .ceph__multus-dropdown-button {
   min-width: 18em;
+  .text-muted {
+    display: none !important;
+  }
```
spadgett, line 18: This will only hide for ceph.

We want a BEM class name specific to the resource dropdown like `co-resource-dropdown__namespace` and then hide that text in the button. We shouldn't be keying on a generic class like `text-muted`, but we shouldn't scope it just to ceph either.

spadgett (approved):
/lgtm

PR #8974, opened 2021-05-19
Add page loading indicator

nemesis09:
Jira issues:
Epic: https://issues.redhat.com/browse/ODC-5013
Story: https://issues.redhat.com/browse/ODC-2529

Problem/Description:
Add page does not load all items at a time due to pending fetches and access reviews. This causes flickering in the Add page, when an add action item loads with a delay.

Solution:
- Show a loading indicator until all the add page items are fetched and checked for access.
- Solve the issue with rbac not re-assessing on namespace change
- We do not need a timeout here since the `checkAccess` for rbac check already handles that.

Screens:
_Add Page loading_
add-page-loading

_RBAC re-assessment on namespace change_
extension-on-namespace-change

_Empty State(unable to load)_
Screenshot from 2021-05-26 01-04-28
Screenshot from 2021-05-26 01-05-49

_Empty State(Not enough permissions)_
Screenshot from 2021-05-27 01-58-52
Screenshot from 2021-05-27 01-59-00

Test Coverage:
Screenshot from 2021-05-20 03-38-09

Browser Conformance:
- [x] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

Dependent on PR: https://github.com/openshift/console/pull/8867


nemesis09:
putting on hold until [PR 8867](https://github.com/openshift/console/pull/8867) merges
/hold

nemesis09:
/kind feature

Since the epic acceptance criteria of [ODC-5013](https://issues.redhat.com/browse/ODC-5013) have not changed and the epic has doc-ack and px-ack, putting approve labels on behalf of them.
/label px-approved
/label docs-approved

nemesis09:
cc @openshift/team-devconsole-ux @openshift/team-devconsole-qe

nemesis09:
[PR 8867](https://github.com/openshift/console/pull/8867) merged. removing hold.
/hold cancel

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
```diff
@@ -0,0 +1,35 @@
...
+  &__background {
+    z-index: 1;
+    background-color: white;
+  }
+}
+
+.odc-add-section-skeleton-placeholder {
+  &__container {
+    position: absolute;
+    top: 0px;
```
invincibleJai, line 13: nit: can avoid px with `0` , here and other places
```suggestion
    top: 0;
```

invincibleJai:
@nemesis09 thanks, it looks good and I observed one issue with tiles not updating based on RBAC checks if I change the namespace

May-20-2021 13-47-42-accessissue

frontend/packages/dev-console/src/hooks/useAddActionsAccessReviews.ts
```diff
@@ -53,7 +60,7 @@ export const useAddActionsAccessReviews = (
           });
       }
     });
-  }, [accessReviewResults, addActionExtensions, namespace]);
+  }, [accessReviewResults, addActionExtensions, initialState, namespace]);
```
invincibleJai, line 63: do we need to `initialState` here in deps?

nemesis09: no, it doesn't have to be a dependency. removed it.

invincibleJai:
and getting warnings for duplicate keys on load

image

nemesis09:
> @nemesis09 thanks, it looks good and I observed one issue with tiles not updating based on RBAC checks if I change the namespace
> 
> May-20-2021 13-47-42-accessissue

This was caused due to stale data in the `useAddActionsAccessReviews`. Fixed it.
Thank you @invincibleJai !

nemesis09:
> and getting warnings for duplicate keys on load
> 
> image

fixed

invincibleJai:
/approve

verified the changes

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import './AddCardSectionSkeleton.scss';
+
+const AddCardSectionSkeleton: React.FC = () => {
+  const { t } = useTranslation();
+  return (
+    <div className="odc-add-section-skeleton">
+      <Skeleton
+        height="600px"
+        className="odc-add-section-skeleton__background"
+        screenreaderText={t('devconsole~Loading add actions')}
```
jerolimov, line 13: This is a nice idea, but with a 4 (or n) column layout, this will let the screen reader speak up 4 (or n) times "Loading add actions". This doesn't help here.

Can the MansonryLayout adds a `aria-label` (only) if its loading and shows the LoadingComponent?

frontend/packages/dev-console/src/components/add/AddPageLayout.tsx
```diff
@@ -20,49 +19,46 @@ const AddPageLayout: React.FC<AddPageLayoutProps> = ({ title, hintBlock }) => {
...
+        </div>
       </div>
-    </div>
-  );
+    );
+  };
 
   return (
-    <PageLayout title={title} hint={hintBlock || defaultHintBlock} isDark>
+    <PageLayout title={title} hint={resolved && (hintBlock || getHintBlock())} isDark>
```
jerolimov, line 55: Why did you show the hint only when the extensions are resolved?

At least the title subline "Select a way to create an Application, component or service from one of the options." should be shown before that. At the moment the page jumps slightly down after the extensions are loaded. See gif:

Peek 2021-05-20 16-30

frontend/packages/dev-console/src/components/add/AddCardSection.tsx
```diff
@@ -23,22 +25,38 @@ const AddCardsSection: React.FC<AddCardsSectionProps> = ({
...
 }) => {
   const filteredAddActionExtensions: ResolvedExtension<AddAction>[] = filterNamespaceScopedUrl(
     namespace,
     useAccessFilterExtensions(namespace, addActionExtensions),
   );
 
-  const sortedActionGroup: LoadedExtension<AddActionGroup>[] = getSortedExtensionItems<
-    LoadedExtension<AddActionGroup>
-  >(addActionGroupExtensions);
+  const isLoading: boolean = loading || filteredAddActionExtensions.length === 0;
```
jerolimov, line 35: You check for both values here. loading or filtered extensions is empty.

Maybe it's possible that `useAccessFilterExtensions` returns is loading status as well so that you check for a clean 
access-review loading-status here as well instead of the fact that there is no action available.

frontend/packages/dev-console/src/components/add/AddCardSection.tsx
```diff
@@ -23,22 +25,38 @@ const AddCardsSection: React.FC<AddCardsSectionProps> = ({
...
+    >(addActionGroupExtensions);
 
-  const addCards: React.ReactElement[] = addGroups.map(({ id, name, items }) => (
-    <AddCard key={id} title={name} items={items} namespace={namespace} />
-  ));
+    const addGroups: AddGroup[] = getAddGroups(filteredAddActionExtensions, sortedActionGroup);
 
-  return <MasonryLayout columnWidth={COLUMN_WIDTH}>{addCards}</MasonryLayout>;
+    return addGroups.map(({ id, name, items }) => (
+      <AddCard key={id} title={name} items={items} namespace={namespace} />
```
jerolimov, line 48: I thought we added a `data-test-id` to each AddCard already in the other PR. Maybe we lost this. Can you add this `data-test-id` as part of this PR?

jerolimov (commented):
Add some small ideas. Let me know what you think.

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.tsx
nemesis09: not sure how we would add the aria label in the masonry component. removed the screenReader instead.

serenamarie125:
I like the addition of the skeleton/loading pattern.  Will leave testing up to others, but nice job getting this in!

jerolimov:
@nemesis09 Works fine in Chrome and Firefox, but unfortunately there is a layout issue on Safari (the skeleton boxes overlaps the column cards):

Screenshot 2021-05-22 at 22 31 49

jerolimov:
I liked that you are care about the edge-case that there is no add action available or the user has no permissions. I personally would recommend to reuse "options" (from the subtitle) instead of "results" and "actions" (which we use only internal).

image

Instead of:

> ### Add
> Select a way to create an Application, component or service from one of the options.
> 
> #### No results found
> The add actions failed to load or you may not have sufficient permissions. Please try again or contact cluster administrator

Maybe:

> ### Add
> Select a way to create an Application, component or service from one of the options.
> 
> ####  No options available
> No option available for which you have sufficient permissions. Please try again or contact cluster administrator.

@nemesis09 @serenamarie125 @beaumorley what did you think? Any other preference?

nemesis09:
> I liked that you are care about the edge-case that there is no add action available or the user has no permissions. I personally would recommend to reuse "options" (from the subtitle) instead of "results" and "actions" (which we use only internal).
> 
> image
> 
> Instead of:
> 
> > ### Add
> > Select a way to create an Application, component or service from one of the options.
> > #### No results found
> > The add actions failed to load or you may not have sufficient permissions. Please try again or contact cluster administrator
> 
> Maybe:
> 
> > ### Add
> > Select a way to create an Application, component or service from one of the options.
> > #### No options available
> > No option available for which you have sufficient permissions. Please try again or contact cluster administrator.
> 
> @nemesis09 @beaumorley what did you think? Any other preference?

Thanks @jerolimov !
I agree that it's better to avoid usage of internal terms. But I wanted to capture the scenario of failure to fetch the extensions as well in this message.

Maybe we could have something like this:

### Add
Select a way to create an Application, component or service from one of the options.
#### No options available
Add options failed to load or you may not have sufficient permissions. Please try again or contact cluster administrator.

Please let me know your thoughts on this. @jerolimov @beaumorley

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
```diff
@@ -0,0 +1,35 @@
+.odc-add-section-skeleton {
+  position: relative;
+  width: 100%;
```
jerolimov, line 3: `width` is not required here, or?

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
```diff
@@ -0,0 +1,35 @@
...
+  &__background {
+    z-index: 1;
```
jerolimov, line 5: `z-index` is not required if you remove both `z-index`, or?

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+.odc-add-section-skeleton-placeholder {
+  &__container {
+    position: absolute;
+    top: 0;
+    z-index: 2;
+    width: 100%;
+    height: 100%;
```
jerolimov, line 16: `height` not required, or?

frontend/packages/dev-console/src/components/add/AddPageLayout.tsx
```diff
@@ -25,58 +28,75 @@ const AddPageLayout: React.FC<AddPageLayoutProps> = ({ title, hintBlock: additio
...
+          <div className="odc-add-page-layout__additional-hint-block">{additionalHint}</div>
+        )}
+      </>
+    );
+  };
 
   return (
     <div className="odc-add-page-layout">
-      <PageLayout title={title} hint={hint} isDark>
+      <PageLayout title={title} hint={getHint()} isDark>
```
jerolimov, line 93:
```suggestion
      <PageLayout title={title} hint={getHint()}>
```

Please drop the `isDark` here. It only adds a light background to the page content which doesn't match the page itself. See:

image

frontend/packages/dev-console/src/components/add/layout/MasonryLayout.tsx
```diff
@@ -6,20 +6,33 @@ import './MasonryLayout.scss';
...
+    : [];
+
   return (
     <Measure bounds onResize={(contentRect) => setWidth(contentRect.bounds?.width)}>
       {({ measureRef }) => (
         <div className="odc-masonry-container" ref={measureRef}>
-          {columnCount ? <Masonry columnsCount={columnCount}>{children}</Masonry> : null}
+          {columnCount ? (
+            <Masonry columnsCount={columnCount}>{loading ? loadingElements : children}</Masonry>
+          ) : null}
```
jerolimov, line 35:
```suggestion
          {columnCount ? <Masonry columnsCount={columnCount}>{coumns}</Masonry> : null}
```

frontend/packages/dev-console/src/components/add/layout/MasonryLayout.tsx
```diff
@@ -6,20 +6,33 @@ import './MasonryLayout.scss';
...
   const [width, setWidth] = React.useState<number>(0);
   const columnCount = React.useMemo(() => (width ? Math.floor(width / columnWidth) || 1 : null), [
     columnWidth,
     width,
   ]);
 
+  const loadingElements: React.ReactElement[] = LoadingComponent
+    ? Array.from({ length: columnCount }, (_, i) => <LoadingComponent key={i.toString()} />)
+    : [];
+
```
jerolimov, line 28: It's not required to create the `LoadingComponent` when the `loading` prop is false. Maybe you can change it like this?

```suggestion
  const coumns: React.ReactElement[] =
    loading && LoadingComponent
      ? Array.from({ length: columnCount }, (_, i) => <LoadingComponent key={i.toString()} />)
      : children;
```

jerolimov (changes requested):

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
nemesis09: removed redundant property

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
nemesis09: z-index is not required. I added z-index so as to not rely on default.
but i think we can remove it.

frontend/packages/dev-console/src/components/add/AddCardSectionSkeleton.scss
nemesis09: removed

gdoyle1:
@nemesis09 @beaumorley asked me to a design review so going to leave a few comments!

The skeleton loader animation should only be applied to the bars rather than the whole page - see this demo https://www.patternfly.org/v4/components/skeleton/html-demos/skeleton-card/. Can we update that? PF also recommends using the skeleton loader when you know the shape/format of the data - it seems weird that the cards change height after loading completes - is there a way we can fix the height of the cards during loading to match the actual card height?

For the empty states, is it possible to detect between the 2 cases?:
1. Problem with user permissions - this would use the no access icon https://www.patternfly.org/v4/components/empty-state/design-guidelines#no-access
Title: Access permissions needed.
Description: You do not have sufficient permissions to access these add options. Contact your cluster administrator.
2. Failed to load data - this would use the back-end failure icon https://www.patternfly.org/v4/components/empty-state/design-guidelines#back-end-failure
Title: Unable to load.
Description: Add options failed to load. Check your connection and reload the page.

nemesis09:
> @nemesis09 @beaumorley asked me to a design review so going to leave a few comments!
> 
> The skeleton loader animation should only be applied to the bars rather than the whole page - see this demo https://www.patternfly.org/v4/components/skeleton/html-demos/skeleton-card/. Can we update that? PF also recommends using the skeleton loader when you know the shape/format of the data - it seems weird that the cards change height after loading completes - is there a way we can fix the height of the cards during loading to match the actual card height?
> 
> For the empty states, is it possible to detect between the 2 cases?:
> 
>     1. Problem with user permissions - this would use the no access icon https://www.patternfly.org/v4/components/empty-state/design-guidelines#no-access
>        Title: Access permissions needed.
>        Description: You do not have sufficient permissions to access these add options.
> 
>     2. Failed to load data - this would use the back-end failure icon https://www.patternfly.org/v4/components/empty-state/design-guidelines#back-end-failure
>        Title: Unable to load.
>        Description: Add options failed to load. Check your connection and reload the page.

Thanks @gdoyle1 for the review.
1. Regarding the heights of the skeleton cards. The loading state renders for two cases-
-  waiting for add actions to load after checking for operators installed and other dependent flags.
- waiting for rbac checks on add actions
in both cases, we wouldn't know how many add actions will finally go through to the UI and the height of the cards depend on the number of add actions in it. In case no add actions are allowed for a add group, the card for that group wouldn't be rendered.

So the number of cards and the height of the cards is decided dynamically based on how many add actions are rendered.
This makes fixing the skeleton card heights to match the actual cards quite difficult. We could choose a "most commonly occurring scenario" and make the skeleton matching it, to minimize the change in shapes.
WDYT? 
cc @jerolimov 

2. We could differentiate between loading error and access error in the `Empty State`. I'll update the `Empty State` accordingly.

jerolimov:
@gdoyle1 @nemesis09 Yeah we could not know how many items we will show. If you want we can reduce the height of some columns or so. But let's not make this too complicated as we could not guess the correct height (or correct number of columns).

gdoyle1:
@nemesis09 @jerolimov Ok not a big deal to not have the correct height/number of columns. We can leave it as is then but the loading effect still needs to be updated so it only applies to the bars rather than the whole card!

jerolimov (approved):
Tested this locally, works fine. :+1: 

/lgtm

sanketpathak:
Tested pr using cluster-bot, it is working fine
/label qe-approved

gdoyle1:
@nemesis09 last thing for this is to make the lock icon `--pf-global--icon--Color--light` (#6A6E73)

nemesis09:
putting it on hold due to discovery of a related  bug
/hold

nemesis09:
> --pf-global--icon--Color--light

Thanks @gdoyle1  This was already changed to a lighter shade, But i have changed the variable to the one you suggested.

nemesis09:
fixed edge case bug for infinite rendering on disabling add actions.
/hold cancel

jerolimov:
Tested it locally. The disabled actions works now, but unfortunately the add page without any disabled action doesn't load anymore. Sorry. :confused: 

/hold

jerolimov:
Works fine now with and without filtered (customizeed) add actions. :+1:

/lgtm
/hold cancel

gdoyle1:
LGTM!! @nemesis09

PR #8977, opened 2021-05-19
Bug 1961363: remove extra space in string

yaacov:
Remove extra space in string


yaacov:
@glekner please review

PR #8979, opened 2021-05-19
Bug 1962153: fix VolumeSnapshot routes

jotak:
- Fix routes so that they're not ambiguous
- Fix Edit YAML link in case we're coming from a PVC page
- Simplify create-volume-snapshot:
    - async loading of kind data shouldn't be necessary
    - rename "resourceName" to "pvcName", more explicit
    - absence/presence of pvcName props is sufficient to decide conditional
      display of the PVC dropdown

### DIFF
--- a/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
+++ b/frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
@@ -16,6 +16,7 @@ import {
   withHandlePromise,
   convertToBaseValue,
   humanizeBinaryBytes,
+  getURLSearchParams,
 } from '@console/internal/components/utils';
 import {
   referenceForModel,
@@ -130,16 +131,14 @@ const isDefaultSnapshotClass = (volumeSnapshotClass: VolumeSnapshotClassKind) =>
   ] === 'true';
 
 const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
-  const { pvcName, namespace, handlePromise, inProgress, errorMessage } = props;
+  const { namespace, pvcName, handlePromise, inProgress, errorMessage } = props;
 
   const { t } = useTranslation();
-  const [statePvcName, setPVCName] = React.useState(pvcName);
+  const [selectedPVCName, setSelectedPVCName] = React.useState(pvcName);
   const [pvcObj, setPVCObj] = React.useState<PersistentVolumeClaimKind>(null);
   const [snapshotName, setSnapshotName] = React.useState(`${pvcName || 'pvc'}-snapshot`);
   const [snapshotClassName, setSnapshotClassName] = React.useState('');
-  const [vscObj, vscLoaded, vscErr] = useK8sGet<ListKind<VolumeSnapshotClassKind>>(
-    VolumeSnapshotClassModel,
-  );
+  const [vscObj, , vscErr] = useK8sGet<ListKind<VolumeSnapshotClassKind>>(VolumeSnapshotClassModel);
   const [scObjList, scObjListLoaded, scObjListErr] = useK8sGet<ListKind<StorageClassResourceKind>>(
     StorageClassModel,
   );
@@ -151,9 +150,9 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
         namespace,
         isList: true,
       },
-      statePvcName ? { name: statePvcName } : null,
+      selectedPVCName ? { name: selectedPVCName } : null,
     );
-  }, [namespace, statePvcName]);
+  }, [namespace, selectedPVCName]);
 
   const [data, loaded, loadError] = useK8sWatchResource<PersistentVolumeClaimKind[]>(resourceWatch);
   const scList = scObjListLoaded ? scObjList.items : [];
@@ -163,7 +162,7 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
     (snapshotClass: VolumeSnapshotClassKind) => provisioner?.includes(snapshotClass?.driver),
     [provisioner],
   );
-  const vscList = vscLoaded && vscObj ? vscObj.items : [];
+  const vscList = vscObj?.items || [];
   const getDefaultItem = React.useCallback(
     (snapFilter) => {
       const filteredVSC = vscList.filter(snapFilter);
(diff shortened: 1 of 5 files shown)


frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -133,18 +130,10 @@ const isDefaultSnapshotClass = (volumeSnapshotClass: VolumeSnapshotClassKind) =>
...
-  const {
-    resourceName,
-    plural,
-    namespace,
-    kindObj,
-    handlePromise,
-    inProgress,
-    errorMessage,
-  } = props;
+  const { pvcName, namespace, handlePromise, inProgress, errorMessage } = props;
```
jotak, line 133: ping @spadgett 
I've simplified this file by removing async loading of Kind data, because it seems to me that it's not necessary. As far as I can tell, it works as I would expect. However I'm not sure if I'm missing something about the reason why we would load it asynchronously rather than using the existing constants.

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -133,18 +130,10 @@ const isDefaultSnapshotClass = (volumeSnapshotClass: VolumeSnapshotClassKind) =>
...
 
   const { t } = useTranslation();
-  const [pvcName, setPVCName] = React.useState(resourceName);
+  const [statePvcName, setPVCName] = React.useState(pvcName);
```
spadgett, line 136: The setter function name should match the variable name. Also note that acronyms should be capitalized like `statePVCName` by the conventions in the console repo.

Can we give something more descriptive than `statePvcName`? Something that describes how this is different than the `pvcName` which is passed in through props?

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -174,7 +163,7 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
     (snapshotClass: VolumeSnapshotClassKind) => provisioner?.includes(snapshotClass?.driver),
     [provisioner],
   );
-  const vscList = vscLoaded ? vscObj.items : [];
+  const vscList = vscLoaded && vscObj ? vscObj.items : [];
```
spadgett, line 166: This probably can be

```suggestion
  const vscList = vscObj?.items || : [];
```

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -230,6 +219,7 @@ const CreateSnapshotForm = withHandlePromise<SnapshotResourceProps>((props) => {
   };
 
   const isBound = (pvc: PersistentVolumeClaimKind) => pvc?.status?.phase === 'Bound';
+  const kindId = `${VolumeSnapshotModel.apiGroup}~${VolumeSnapshotModel.apiVersion}~${VolumeSnapshotModel.kind}`;
```
spadgett, line 222: Use `referenceForModel`.

```suggestion
  const kindReference = referenceForModel(VolumeSnapshotModel);
```

frontend/public/components/app-contents.tsx
```diff
@@ -449,7 +449,7 @@ const AppContents: React.FC<{}> = () => {
             />
 
             <LazyRoute
-              path={`/k8s/ns/:ns/:plural/:name/${VolumeSnapshotModel.plural}/~new/form`}
+              path={`/k8s/ns/:ns/${VolumeSnapshotModel.plural}/pvc/:pvcName/~new/form`}
```
spadgett, line 452: I would recommend using a query parameter instead like

```
/k8s/ns/:ns/${VolumeSnapshotModel.plural}/~new/form?pvc=my-pvc
```

Then we don't need a separate route. This follows the existing URL pattern we have better as well.

spadgett (commented):
Thanks @jotak 👍

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
jotak: Well, it's quite hard to answer, it's really the same thing except one is props (provided by parent) and one is state (mutated inside component). If you want I can rename the state one `finalPvcName`, but don't hesitate to tell me if you got a better idea

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
jotak: Cool, I wanted to ask, was pretty sure this had to exist :)

frontend/public/components/app-contents.tsx
jotak: Sure, makes sense. I guess the query param doesn't have to be declare in the route definition.

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
jotak: done

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
jotak: done

frontend/public/components/app-contents.tsx
jotak: done

frontend/packages/console-app/src/components/volume-snapshot/create-volume-snapshot/create-volume-snapshot.tsx
```diff
@@ -133,24 +131,14 @@ const isDefaultSnapshotClass = (volumeSnapshotClass: VolumeSnapshotClassKind) =>
...
-    kindObj,
-    handlePromise,
-    inProgress,
-    errorMessage,
-  } = props;
+  const { namespace, pvcName, handlePromise, inProgress, errorMessage } = props;
 
   const { t } = useTranslation();
-  const [pvcName, setPVCName] = React.useState(resourceName);
+  const [selectedPVCName, setSelectedPVCName] = React.useState(pvcName);
```
spadgett, line 137: 👍 Thank you, this name looks better

spadgett (approved):
/lgtm

PR #8980, opened 2021-05-19
Bug 1962256: virt: use the new rhel8 image as an example

dankenigsberg:
For downtream it would be better than Fedora

Signed-off-by: Dan Kenigsberg <danken@redhat.com>


yaacov:
/retitle Bug 1962256: virt: use the new rhel8 image as an example

PR #8986, opened 2021-05-20
Bug 1960534: Monitoring dashboards: Always allow custom format for tooltip entries

kyoto:
The `legendFormat` function is used for formatting the legend entries
both below the graph and in the graph tooltip. However, we were ignoring
it when `panel.legend.show` isn't `true`. This changes the tooltips to
always use `legendFormat` when it is provided.

| Before | After |
|-|-|
| before | after


lihongyan1:
/label qe-approved
Test with pr, tooltips display as expected.

spadgett (approved):
/lgtm

PR #8990, opened 2021-05-20
Rename 'Insights' to 'Insights Advisor'

gkarat:
- [x] Replace `Insights` occurrences with `Insights Advisor` in the Insights widget (but some spots should be left as `Insights` after the request from PMs).
- [x] Add missing paragraph to the content of Insights widget.

Before:
Screenshot from 2021-05-20 16-28-57
After:
Screenshot from 2021-05-20 16-26-50

PR #8991, opened 2021-05-20
Insights Advisor widget: refactor

gkarat:
- [x] Make a refactor of the Insights Advisor widget. Its look should be aligned as much as possible to the existing mockup https://marvelapp.com/prototype/852jaj9/screen/70579115. Also, the code for the widget from OCM can be helpful (copy/pasted): https://gitlab.cee.redhat.com/service/uhc-portal/-/tree/master/src/components/clusters/ClusterDetails/components/Overview/InsightsAdvisor.
- [x] Cover the state when `clusterId` is unavailable for some reason (put correct OCM URL).
- [x]  Replace Insights occurrences with Insights Advisor in the Insights widget (but some spots should be left as Insights after the request from PMs).
- [x] Add missing paragraph to the content of Insights widget.

See commit history, I tried to dedicate a single commit for a single change.

Before:
Screenshot from 2021-05-20 17-12-35

After:
Screenshot from 2021-05-20 17-08-27


gkarat:
@spadgett, @zherman0, @Fewwy, please, take a look :star2:

frontend/packages/insights-plugin/locales/ja/insights-plugin.json
```diff
@@ -1,15 +1,16 @@
...
+  "Important": "重要 (Important)",
+  "Critical": "重大 (Critical)",
   "Temporary unavailable.": "一時的に利用不可です。",
   "Disabled or waiting for results.": "無効にされているか、または結果を待機しています。",
   "Total issue_0": "全体の問題",
   "Total issue_1": "全体の問題",
   "Total Risk": "全体リスク",
   "Fixable issues": "修正可能な問題",
   "View all in OpenShift Cluster Manager": "OpenShift Cluster Manager ですべてを表示する",
+  "Go to OpenShift Cluster Manager": "Go to OpenShift Cluster Manager",
```
gkarat, line 13: @rebeccaalpert, @spadgett, could you please help me with that and explain the algorithm of updating/adding i18n keys? Who we need to ask to translate this? How much time will it take? I prefer to finish translation in the scope of a separate patch ideally.

Fewwy:
Tried it on my local env, LGTM

frontend/packages/insights-plugin/locales/ja/insights-plugin.json
rebeccaalpert: Hi @gkarat - Thanks for tagging us here. You shouldn't ever update these files in the ja, zh, or ko folders. Sam and I take care of arranging translation for the whole console (including packages) right now, so there's nothing you need to do here beyond reverting the changes you've made to these files. We send over files about once a week now for translation.

rebeccaalpert (changes requested):
Hi there - I left some other i18n-related comments. Thanks for tagging us!

gkarat:
@rebeccaalpert, I've edited the last commit and applied your suggestions, PTAL :)

frontend/packages/insights-plugin/src/components/AdvisorChart/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      data={issues.map(([k, v]) => ({
+        label: `${v} ${t(riskLabels[k])}`,
+        x: k,
+        y: v,
+      }))}
+      title={`${issuesCount}`}
+      subTitle={
+        issuesCount === 1
+          ? t('insights-plugin~Total issue', { count: issuesCount })
+          : t('insights-plugin~Total issues', { count: issuesCount })
```
rebeccaalpert, line 33: Hey - There's no need to add a check here; if you just use the `count` keyword in `t` and run `yarn i18n` it will update the strings correctly.

gkarat:
Added the changes from https://github.com/openshift/console/pull/8990 to avoid merge conflicts later.

rebeccaalpert (changes requested):
One comment; the rest looks good to me! Thanks for making those changes.

frontend/packages/insights-plugin/src/components/AdvisorChart/index.tsx
gkarat: @rebeccaalpert, wow, that works like a charm. Added one more commit :)

rebeccaalpert (approved):
Perfect; thank you for making those updates!

PR #8994, opened 2021-05-20
Bug 1961829: Fix quick start prerequisites getting truncated

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/OCPBUGSM-29337

**Analysis / Root cause**: Catalog tiles truncate the description if it iss more than 3 lines. We send the prerequisites along with description to the tile and that is why it was not being shown for long description.

**Solution Description**: `CatalogTile` has an option to send children for the body which doesn't get truncated. So sending the description as children solves the truncation issue.

**Screen shots / Gifs for design review**: @openshift/team-devconsole-ux @bmignano 

https://user-images.githubusercontent.com/6041994/119015191-523f5200-b9b6-11eb-8405-ba668ee5334b.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTile.tsx
+++ b/frontend/packages/console-app/src/components/quick-starts/catalog/QuickStartTile.tsx
@@ -40,10 +40,9 @@ const QuickStartTile: React.FC<QuickStartTileProps> = ({
       featured={isActive}
       title={<QuickStartTileHeader name={displayName} status={status} duration={durationMinutes} />}
       onClick={onClick}
-      description={
-        <QuickStartTileDescription description={description} prerequisites={prerequisites} />
-      }
-    />
+    >
+      <QuickStartTileDescription description={description} prerequisites={prerequisites} />
+    </CatalogTile>
   );
 };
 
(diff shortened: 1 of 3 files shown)


rohitkrai03:
@bmignano I've updated the quick start catalog tile to show the whole description along with prerequisites. This will lead to some cards getting bigger than others but the whole row will have the same height. Can you confirm that this behaviour is okay from the UX side?

bmignano:
@rohitkrai03 I think we should set a max of 3 lines for the description, and always show the prerequisites. If we need to cut a description off, let's do the same thing we do in OperatorHub and use "..." to cut it off. The goal is for the descriptions to be concise, so we may want to make sure the docs reiterate this and discourage people from adding too long descriptions. This will also help set boundaries and avoid any potential custom quick starts from creating very tall cards with long descriptions.
FYI @ahardin-rh

rohitkrai03:
@bmignano I've updated the description section to truncate after three lines and show `...` just like OperatorHub. Prerequisites will always be shown at the bottom of the tile. Can you approve the PR if this looks good to you?

bmignano (approved):
Thanks @rohitkrai03 , LGTM

jerolimov (approved):
Tested this in Chrome and in Firefox and both works fine and clamps the description after 3 lines. And if the style is not supported "Prerequisites" is just moved done.

/lgtm

PR #8995, opened 2021-05-20
Change the cancel label to close in delete resource modal

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5850

**Analysis / Root cause**: When user clicks the `Delete` button on a delete resource modal, the `Cancel` button stays enabled and sends a wrong message that the delete action can be cancelled. 

**Solution Description**: Change the `Cancel` text to say `Close` which means the button will close the modal and not cancel the delet action.

**Screen shots / Gifs for design review**: @openshift/team-devconsole-ux 

Screenshot 2021-05-20 at 10 30 22 PM

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge


rohitkrai03:
/hold

This change needs an approval from @openshift/team-devconsole-ux.

andrewballantyne (approved):
I think this is the right action, disabling the button feels weird and we don't (as far as I know) stop any actions that we have started - creating or deleting.

Naturally it makes sense to wait for UX to confirm, but lgtm (once you run `yarn i18n` -- the invocation of `Close` must be earlier in the import tree now)

serenamarie125:
Let's hold this PR for a bit -- we should actively include UX on this, to be sure that things are consistent throughout the console.  The intent of the confirmation is to confirm the destructive action, not to allow the user to cancel the confirmation of the destructive action.

beaumorley:
Agree with @serenamarie125. It looks like this interaction needs some reworking. Once the user hits Uninstall we cannot Cancel. I'll add a DTUX story to the issue to indicate we need a design. Thanks.

rohitkrai03:
Closing this PR as we don't have clarity on UX at the moment.

PR #8996, opened 2021-05-20
Auto focus first input field of import forms

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-3252

**Analysis / Root cause**: Import forms don't focus the first fields which can lead to undesired behaviours like going back to a focus nav item when clicking enter.

**Solution Description**: Focus first field of the import forms. Adding auto focus to git section and image search input fields cover all of the import forms for us as these are re-used.

**Screen shots / Gifs for design review**: @openshift/team-devconsole-ux 

https://user-images.githubusercontent.com/6041994/119029699-963a5300-b9c6-11eb-9840-7f3836142464.mov

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
+++ b/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
@@ -38,6 +38,7 @@ const GitSection: React.FC<GitSectionProps> = ({
   builderImages,
 }) => {
   const { t } = useTranslation();
+  const inputRef = React.useRef<HTMLInputElement>();
   const { values, setFieldValue, setFieldTouched, touched, dirty } = useFormikContext<
     FormikValues
   >();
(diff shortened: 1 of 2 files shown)


jerolimov (approved):
Tested this locally, works fine.
/lgtm

Do you want wait for an UX approval? Add a hold for that, feel free to remove it when you don't want to wait or there is a LGTM.
/hold

serenamarie125:
As a user, I think this would be expected behavior.  This looks like a great change to me, having that first field in focus is also more efficient for the user.  
@beaumorley do you have any objections to this behavior?

beaumorley:
thanks @rohitkrai03 UX approves as well. to @serenamarie125 's point we'll take on looking at this more broadly through openshift.

rohitkrai03:
/hold cancel

Have UX and PM approval on the change.

PR #8997, opened 2021-05-21
add vm creation to smoke tests

gouyang:
Add smoke tests for vm creation from URL, CLONE_PVC and Registry, also covers rhel/fedora/windows.


gouyang:
@yaacov @pcbailey PTAL.

frontend/packages/kubevirt-plugin/integration-tests-cypress/tests/devconsole/devconsole.spec.ts
```diff
@@ -76,7 +76,7 @@ describe('test dev console', () => {
   describe('create vm in dev console', () => {
     it('ID(CNV-5699) create virtual machine', () => {
       cy.byLegacyTestID(addHeader).click();
-      cy.get('.odc-namespaced-page__content').scrollTo('bottom');
+      cy.get('#content-scrollable').scrollTo('bottom');
```
yaacov, line 79: :+1:

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/actions.ts
```diff
@@ -6,10 +6,10 @@ export const confirmButton = '[data-test="confirm-action"]';
 export const confirmCloneButton = 'button[id="confirm-action"]';
 export const kebabButton = 'kebab-button';
 
-export const selectItemFromDropdown = (action: string, selector: string) => {
+export const selectItemFromDropdown = (item: string, selector: string) => {
```
pcbailey, line 9: nitpick: I'm not sure if the `TestActionID` is used on all dropdowns. If it isn't, then this function is actions-specific and I think we should use names to indicate that, such as `selectActionFromDropdown` and leave `item` as `action`. If `TestActionID` is used on all dropdowns, then I think we should move this function to a more common location or use an existing function if one is available.

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/wizard.ts
```diff
@@ -18,18 +18,46 @@ export const wizard = {
         }
       });
     },
-    processBootSource: (source: ProvisionSource, cdrom: boolean, size: string) => {
+    processBootSource: (
```
pcbailey, line 21: I don't know that using an object literal is the most maintainable way of creating our views. I think we need to look at eventually moving them to individual functions/variables or some other format. If we use individual functions/variables, things could be better organized while still calling the whole file like `import * as vmWizard from './view/wizard'`.

pcbailey (commented):
Just a couple of small comments. Nothing worth holding it up over, but may be worth looking at in follow up PRs.

frontend/packages/kubevirt-plugin/integration-tests-cypress/view/actions.ts
gouyang: I agree with you to change the function name to `selectActionFromDropdown` and use the parameter `action`.
For now, let it sits here until we have a clear requirement to put it into `utils/utils.ts`.

PR #9003, opened 2021-05-21
Bug 1945312: Fix project deletion issue and reset active namespace

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/OCPBUGSM-27202

**Analysis / Root cause**: When `activeNamespace`was moved to `userSettings` in https://github.com/openshift/console/pull/7433, it was refactored to use a global namespace context. Since modals created by `createModalLauncher` uses `ReactDOM.render` to render the modal in a new context, `delete-namespace-modal` was unable to get the namespace context.

**Solution Description**: There's no way to get the actual global `Namespace` context unless we refactor the way modal are being created by `createModalLauncher`. As a workaround, directly fetch `activeNamespace` from redux store and set a new namespace after deletion in redux and user settings.

https://user-images.githubusercontent.com/6041994/119167965-0ce65780-ba7e-11eb-926e-063b2bac8d98.mov

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
+++ b/frontend/packages/console-app/src/components/detect-namespace/namespace.ts
@@ -8,9 +8,10 @@ import { useUserSettingsCompatibility } from '@console/shared/src/hooks/useUserS
 import { setActiveNamespace } from '@console/internal/actions/ui';
 import {
   ALL_NAMESPACES_KEY,
-  USERSETTINGS_PREFIX,
   NAMESPACE_USERSETTINGS_PREFIX,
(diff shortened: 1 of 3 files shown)


christianvogt:
I think it's about time we have an issue to clean up the modal launcher. It should be rendered in the same react root as the rest of the app and have access to the same react contexts.

This however would require a large refactoring which expect to be able to launch a model from any static call.

@spadgett is this a tech debt item that we can add to console backlog if you agree?

PR #9008, opened 2021-05-21
adjust styles of add page and getting started to align with UX

christianvogt:
**Fixes**: 
Addresses UX changes related to story: https://issues.redhat.com/browse/ODC-5561
and PF issue: https://github.com/patternfly/patternfly/issues/4085

**Analysis / Root cause**: 
UX change request to include details in add page cards on hover.
Also addresses styling issue in PF related to https://github.com/patternfly/patternfly/issues/4085

**Solution Description**: 
Override styles.

Focus and hover states:
image
image

**Browser conformance**: 

- [ ] Chrome
- [ ] Firefox
- [x] Safari
- [x] Edge

cc @nemesis09 @jerolimov 

@gdoyle1 please approve UX

### DIFF
--- a/frontend/packages/console-shared/src/components/getting-started/GettingStartedCard.scss
+++ b/frontend/packages/console-shared/src/components/getting-started/GettingStartedCard.scss
@@ -1,4 +1,10 @@
 .ocs-getting-started-card {
+  .pf-c-simple-list__item-link {
+    --pf-c-simple-list__item-link--focus--FontWeight: normal;
+    --pf-c-simple-list__item-link--active--FontWeight: normal;
+    --pf-c-simple-list__item-link--hover--FontWeight: normal;
+  }
+
   &__title-icon {
     margin-right: var(--pf-global--spacer--sm);
   }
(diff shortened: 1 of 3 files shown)


jerolimov (approved):
After this change, we could/should also remove the wrapping `div` in `AddCardItem`. But this goes into a good direction, so let us merge this.

Verified with a local version on Chrome, Firefox and Safari.

/lgtm

gdoyle1:
@christianvogt @jerolimov I know this has already been merged but I just got the chance to look at it now - seeing it next to the getting started card, I think we should change the description text and the icon to blue on hover/select/focus as well. It would help make it look more consistent (those are the only 2 places we are using the simple list today).

jerolimov:
See also https://github.com/patternfly/patternfly/issues/4085#issuecomment-847275771

I would create a small PR to fix this when we now if we want change the color on hover or not.

gdoyle1:
> See also [patternfly/patternfly#4085 (comment)](https://github.com/patternfly/patternfly/issues/4085#issuecomment-847275771)
> 
> I would create a small PR to fix this when we now if we want change the color on hover or not.

@jerolimov that's a great idea to wait until the PF fix happens too before we make changes to the hover color - let's just do that

gdoyle1:
@jerolimov it looks like @christianvogt already fixed the problem brought up about the text getting bolder. So I think we should just create a small PR to make the icons on the add page also blue upon hover. Then whenever patternfly/patternfly#4085 is fixed, we can update based on that... what do you think @jerolimov

PR #9009, opened 2021-05-23
Bug 1963211: Kubevirt i18n typo

yaacov:
"This template is provided by Red Hat, But is not supported " should be "This template is provided by Red Hat, but is not supported"

Signed-off-by: yaacov <kobi.zamir@gmail.com>


yaacov:
@avivtur please review

PR #9012, opened 2021-05-23
Bug 1963676: Vm wizard choose os not template

yaacov:
Problem:
In current flow, user choose the template + flavor in the first step
but are only given one flavor for each template family, users don't
know they can choose the flavor in the next step.

Solution:
Let user only select the template in the first step and choose the flavor in the next step.

Screenshots:
Before
screenshot-localhost_9000-2021 05 23-15_30_10

After
screenshot-localhost_9000-2021 05 23-15_29_33

Signed-off-by: yaacov <kobi.zamir@gmail.com>


yaacov:
@avivtur @glekner @metalice @pcbailey please review

PR #9013, opened 2021-05-23
add dynamic sdk support for model metadata extensions

glekner:
JIRA: https://issues.redhat.com/browse/CONSOLE-2540

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
@@ -3,7 +3,7 @@ import { Extension, ExtensionDeclaration } from '../types';
 
 /** Customize the display of models by overriding values retrieved and generated through API discovery. */
 export type ModelMetadata = ExtensionDeclaration<
-  'console.resource-metadata',
+  'console.model-metadata',
   {
     /** The model to customize. May specify only a group, or optional version and kind. */
     model: ExtensionK8sGroupModel;
@@ -23,4 +23,4 @@ export type ModelMetadata = ExtensionDeclaration<
 // Type guards
 
 export const isModelMetadata = (e: Extension): e is ModelMetadata =>
-  e.type === 'console.resource-metadata';
+  e.type === 'console.model-metadata';
--- a/frontend/public/module/k8s/get-resources.ts
+++ b/frontend/public/module/k8s/get-resources.ts
@@ -2,8 +2,11 @@ import * as _ from 'lodash-es';
 import { plural } from 'pluralize';
 
 import { K8sKind, K8sVerb } from '../../module/k8s';
+import { isModelMetadata, ModelMetadata } from '@console/dynamic-plugin-sdk';
+import { LoadedExtension } from '@console/dynamic-plugin-sdk/src/types';
 import { API_DISCOVERY_RESOURCES_LOCAL_STORAGE_KEY } from '@console/shared/src/constants';
 import { fetchURL } from '../../graphql/client';
+import { pluginStore } from '../../plugins';
 
 const ADMIN_RESOURCES = new Set([
   'roles',
@@ -102,6 +105,43 @@ export const pluralizeKind = (kind: string): string => {
   return pluralized;
 };
 
+export const getModelExtensionMetadata = (
+  extensions: LoadedExtension<ModelMetadata>[],
+  group: string,
+  version?: string,
+  kind?: string,
+) => {
+  const groupVersionKindMetadata = extensions
+    .filter(
+      ({ properties }) =>
+        properties.model.group === group &&
+        properties.model.kind === kind &&
+        properties.model.version === version,
+    )
+    .map((e) => e.properties);
+  const groupKindMetadata = extensions
+    .filter(
+      ({ properties }) =>
+        properties.model.version == null &&
+        properties.model.group === group &&
+        properties.model.kind === kind,
+    )
+    .map((e) => e.properties);
+  const groupMetadata = extensions
+    .filter(
+      ({ properties }) =>
+        properties.model.kind == null &&
+        properties.model.version == null &&
+        properties.model.group === group,
+    )
+    .map((e) => e.properties);
+
+  return _.omit(
+    Object.assign({}, ...groupMetadata, ...groupKindMetadata, ...groupVersionKindMetadata),
+    ['model'],
+  );
+};
+
 export const getResources = () =>
   fetchURL('/apis').then((res) => {
     const groupVersionMap = res.groups.reduce(
@@ -137,6 +177,9 @@ export const getResources = () =>
       const adminResources = [];
 
       const defineModels = (list: APIResourceList): K8sKind[] => {
+        const metadataExtensions = pluginStore
+          .getExtensionsInUse()
+          .filter(isModelMetadata) as LoadedExtension<ModelMetadata>[];
         const groupVersionParts = list.groupVersion.split('/');
         const apiGroup = groupVersionParts.length > 1 ? groupVersionParts[0] : null;
         const apiVersion = groupVersionParts.length > 1 ? groupVersionParts[1] : list.groupVersion;
@@ -157,6 +200,7 @@ export const getResources = () =>
               path: name,
               id: singularName,
               crd: true,
+              ...getModelExtensionMetadata(metadataExtensions, apiGroup, apiVersion, kind),
             };
           });
       };
--- a/frontend/public/module/k8s/k8s-models.ts
+++ b/frontend/public/module/k8s/k8s-models.ts
@@ -4,7 +4,12 @@ import * as _ from 'lodash-es';
 // @ts-ignore
 import { useSelector } from 'react-redux';
 
-import { K8sResourceKindReference, K8sKind, DiscoveryResources } from './index';
+import {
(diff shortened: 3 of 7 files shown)


glekner:
@christianvogt @vojtechszocs 
please review

frontend/public/module/k8s/k8s-models.ts
```diff
@@ -57,27 +59,53 @@ const getK8sModels = () => {
...
  */
 export const modelFor = (ref: K8sResourceKindReference) => {
+  const dynamicModelMetadata = pluginStore
+    .getAllExtensions()
+    .filter(isModelMetadata)
+    ?.find(
+      (e) =>
+        ref === referenceForExtensionModel(e?.properties?.model) ||
+        ref === e?.properties?.model?.kind,
+    )?.properties;
```
christianvogt, line 69: It's possible that there may be multiple matches.
eg. An extension could contribute the tech preview badge to all resources with a given api group. A second extension could then provide specific labels for the same resource.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
```diff
@@ -13,14 +13,18 @@ export type ModelMetadata = ExtensionDeclaration<
...
     /** Override the label. Requires `kind` be provided. */
     label?: string;
+    /** Override plural. Requires `kind` be provided. */
+    plural?: string;
     /** Override the plural label. Requires `kind` be provided. */
     labelPlural?: string;
     /** Customize the abbreviation. Defaults to All uppercase chars in the kind up to 4 characters long. Requires `kind` be provided. */
     abbr?: string;
+    /** Flag if the model is a Custom Resource Definition */
+    crd?: boolean;
```
christianvogt, line 23: Why would `crd` be part of the metadata?
It should be known at discovery time whether the model is a CRD or not and therefore it isn't required to be augmented by an extension.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
```diff
@@ -13,14 +13,18 @@ export type ModelMetadata = ExtensionDeclaration<
     color?: string;
     /** Override the label. Requires `kind` be provided. */
     label?: string;
+    /** Override plural. Requires `kind` be provided. */
+    plural?: string;
```
christianvogt, line 17: Isn't the `plural` value part of the CRD definition? 
Why would you ever have a need to override the this value?

frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
spadgett: `crd` is a misnomer. IIRC, it just determines whether we use the fully-qualified kind/group/version in the URL instead of the short plural name (for legacy paths we don't want to break). IMO all models contributed by plugins should have `crd: true`, and there shouldn't be a way to override.

christianvogt: Ok so i think we can agree that we can omit this value in the extension and internally set it to `true` if needed.

frontend/public/module/k8s/k8s-models.ts
glekner: how should we deal with this then?

christianvogt: You'll need to combine all valid matches of metadata together.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
```diff
@@ -3,7 +3,7 @@ import { Extension, ExtensionDeclaration } from '../types';
 
 /** Customize the display of models by overriding values retrieved and generated through API discovery. */
 export type ModelMetadata = ExtensionDeclaration<
-  'console.resource-metadata',
+  'console.model-metadata',
```
rawagner, line 6: renaming an extension id is API breaking change

glekner: @christianvogt asked for me IIRC, wdyt?

spadgett: If we make the change, we'll need to deprecate the old value and support both for a while. (Normal policy is 12 months / 3 releases.)

spadgett: Assuming it's an API we've already shipped.

christianvogt: My original extension had identifier `console.resource-metadata` but I think I referred to it by name as 'model metadata'.

Since there is no implementation which consumes this extension I think it's fair to make any change now.

Let's just make sure we agree on the name.
I'm indifferent. Both are suitable to me.

rawagner: The API was shipped in 4.8 https://github.com/openshift/console/blob/release-4.8/frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts

frontend/public/module/k8s/get-resources.ts
```diff
@@ -136,7 +139,41 @@ export const getResources = () =>
...
+          return Object.assign({}, ...groupKindMetadata.map((e) => e.properties));
+        }
+        const groupMetadata = extensions.filter(
+          ({ properties }) => properties?.model?.group === group,
+        );
+        if (groupMetadata.length > 0) {
+          return Object.assign({}, ...groupMetadata.map((e) => e.properties));
+        }
+        return {};
+      };
```
christianvogt, line 171: You need to merge all matching permutations into a single object.
First check least specific to most specific.
Match only group first then match group + kind and finally group + kind + version.

frontend/public/module/k8s/k8s-models.ts
```diff
@@ -57,28 +59,44 @@ const getK8sModels = () => {
...
+  const metadataExtensions = pluginStore
+    .getExtensionsInUse()
+    .filter(
+      (e) =>
+        isModelMetadata(e) &&
+        (ref === referenceForExtensionModel(e?.properties?.model) ||
+          ref === e?.properties?.model?.kind),
+    )
+    .map((e) => e.properties);
+  const combinedModelMetadata = Object.assign({}, ...metadataExtensions);
```
christianvogt, line 71: metadata should be combined in the same order as with api discovery.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/resource-metadata.ts
christianvogt: while it was shipped, there was no support for the extension

spadgett: If the extension was there, but did nothing, we can rename it. It shouldn't break any existing plugins.

frontend/public/module/k8s/get-resources.ts
```diff
@@ -102,6 +105,32 @@ export const pluralizeKind = (kind: string): string => {
...
+  const groupKindMetadata = extensions
+    .filter(
+      ({ properties }) => properties?.model?.group === group && properties?.model?.kind === kind,
+    )
+    .map((e) => e?.properties);
+  const groupMetadata = extensions
+    .filter(({ properties }) => properties?.model?.group === group)
+    .map((e) => e?.properties);
+
+  return Object.assign({}, ...groupMetadata, ...groupKindMetadata, ...groupVersionKindMetadata);
```
christianvogt, line 131: The result should omit the `model` property.

christianvogt:
fyi @karthikjeeyar @rohitkrai03 since this PR implements support for the model metadata which some plugins already added extensions for, can you give it a test to ensure these areas, such as pipelines, work as expected.

christianvogt:
@karthikjeeyar Here are some differences I noticed with this change being on the left:

image
image

@glekner Notice how `ClusterInterceptor` and `Trigger` both have the abbreviation of `EL`. I don't see this assignment in the extensions.

frontend/public/module/k8s/get-resources.ts
```diff
@@ -102,6 +105,35 @@ export const pluralizeKind = (kind: string): string => {
...
+        properties?.model?.version === version,
+    )
+    .map((e) => e?.properties);
+  const groupKindMetadata = extensions
+    .filter(
+      ({ properties }) => properties?.model?.group === group && properties?.model?.kind === kind,
+    )
+    .map((e) => e?.properties);
+  const groupMetadata = extensions
+    .filter(({ properties }) => properties?.model?.group === group)
```
christianvogt, line 128: Need to ensure that the extension model only has a group:
```suggestion
    .filter(({ properties }) => properties.model.kind == null && properties.model.version == null && properties.model.group === group)
```

christianvogt:
Tested on cluster bot, the correct metadata was selected.
The issue previously seen for tekton resources has been fixed.

christianvogt:
No docs updates with this change required as documenting the SDK is not yet complete.

/label docs-approved
/label px-approved

PR #9018, opened 2021-05-24
Bug 1957982: Make disabled action items not clickable and focusable

rohitkrai03:
**Fixes**: 
- https://issues.redhat.com/browse/ODC-3045
- https://issues.redhat.com/browse/OCPBUGSM-28886

**Analysis / Root cause**: The disabled action items in action menus were still clickable and focusable which gave a false sense that clicking on the action would work. 

**Solution Description**: 
- Disable the action item button so that it is not clickable or focusable.
- Send a fallback element to the focus trap when there are no tabbable elements. 

Before - 

https://user-images.githubusercontent.com/6041994/119333445-7d23f180-bca7-11eb-92ff-c652a32c2f85.mov

After - 

https://user-images.githubusercontent.com/6041994/119333460-82813c00-bca7-11eb-80e9-55fad2094224.mov

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


invincibleJai:
/lgtm

verified the changes, works as expected

PR #9022, opened 2021-05-24
Fix the alignment of success ratio and pipeline run count chart

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5627

**Analysis / Root cause**: 

The number of pipeline runs graph shrinks so it isn't aligned with pipeline success ratio graph once there are data points.

**Solution Description**: 

Added a missing negative margin in donut chart which is present in the adjacent chart (success percentage chart) in the first card.
 
image

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @bgliwa01 @andrewballantyne

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
@@ -99,10 +99,16 @@ const PipelineSuccessRatioDonut: React.FC<PipelineMetricsGraphProps> = ({
   }
   const successValue = _.find(finalArray, { x: 'success' })?.['count'] ?? 0;
   const successData = _.sortBy(finalArray, 'sortOrder');
+  const OVERLAP = 20;
   return (
     <Grid hasGutter>
       <GridItem xl2={3} xl={3} lg={3} md={3} sm={3}>
-        <div style={{ height: DEFAULT_CHART_HEIGHT }}>
+        <div
+          style={{
+            height: DEFAULT_CHART_HEIGHT,
+            marginTop: OVERLAP * -1,
+          }}
+        >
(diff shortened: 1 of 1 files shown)


frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/PipelineSuccessRatioDonut.tsx
```diff
@@ -99,10 +99,16 @@ const PipelineSuccessRatioDonut: React.FC<PipelineMetricsGraphProps> = ({
...
   const successData = _.sortBy(finalArray, 'sortOrder');
+  const OVERLAP = 20;
   return (
     <Grid hasGutter>
       <GridItem xl2={3} xl={3} lg={3} md={3} sm={3}>
-        <div style={{ height: DEFAULT_CHART_HEIGHT }}>
+        <div
+          style={{
+            height: DEFAULT_CHART_HEIGHT,
+            marginTop: OVERLAP * -1,
```
karthikjeeyar, line 109: Added here based on the implementation to solve the [tooltip issue](  https://github.com/openshift/console/pull/8062)

andrewballantyne: Interesting. Well that's bad on my part. I wonder if the solution is actually to add padding to inverse the margin rather than to include more margin offsets 🤔 

We needed space for the tooltip due to poor portal-ing by PF (or Victory).

andrewballantyne (approved):
I'm going to approve this as it continues the hack I did to try to get the tooltips to show properly. I'm wondering if that original (hack) solution is holding water though. Maybe we need to investigate this behind the scenes but at least apply this fix for the time being.

jerolimov (approved):
Looks better then before :)

/lgtm

PR #9023, opened 2021-05-24
Fix number spinner in  NumberSizeInput without default values

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5852

**Analysis / Root cause**: 
- `NumberSpinner` of `RequestSizeInput` doesn't work correctly when there is no default value.
- https://github.com/openshift/console/pull/8713 updated `RequestSizeInput` to use `NumberSpinner` instead of simple `input` field. 
- When no default value is passed to `RequestSizeInput` it sets the value to `NaN` which is why the spinner doesn't work.

**Solution Description**: Check for `NaN` value before changing it and default to `0`.

https://user-images.githubusercontent.com/6041994/119369695-15cf6700-bcd2-11eb-81f1-9a01ff8a10e3.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/public/components/utils/request-size-input.tsx
+++ b/frontend/public/components/utils/request-size-input.tsx
@@ -28,8 +28,10 @@ export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   };
 
   const changeValueBy = (changeBy: number) => {
-    setValue(value + changeBy);
-    onChange({ value: value + changeBy, unit });
+    // When default defaultRequestSizeValue is not set, value becomes NaN and increment decrement buttons of NumberSpinner don't work.
+    const newValue = Number.isFinite(value) ? value + changeBy : 0 + changeBy;
+    setValue(newValue);
+    onChange({ value: newValue, unit });
   };
 
   const onUnitChange = (newUnit) => {


frontend/public/components/utils/request-size-input.tsx
```diff
@@ -28,8 +28,10 @@ export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   };
 
   const changeValueBy = (changeBy: number) => {
-    setValue(value + changeBy);
-    onChange({ value: value + changeBy, unit });
+    // When default defaultRequestSizeValue is not set, value becomes NaN and increment decrement buttons of NumberSpinner doesn't work.
+    const newValue = !isNaN(value) ? value + changeBy : 0 + changeBy;
```
spadgett, line 32: `Number.isFinite` is a better check here since it handles some other cases.

```suggestion
    const newValue = !Number.isFinite(value) ? value + changeBy : 0 + changeBy;
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite

frontend/public/components/utils/request-size-input.tsx
```diff
@@ -28,8 +28,10 @@ export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   };
 
   const changeValueBy = (changeBy: number) => {
-    setValue(value + changeBy);
-    onChange({ value: value + changeBy, unit });
+    // When default defaultRequestSizeValue is not set, value becomes NaN and increment decrement buttons of NumberSpinner doesn't work.
```
spadgett, line 31:
```suggestion
    // When default defaultRequestSizeValue is not set, value becomes NaN and increment decrement buttons of NumberSpinner don't work.
```

spadgett (commented):
Thanks @rohitkrai03

frontend/public/components/utils/request-size-input.tsx
rohitkrai03: Updated.

frontend/public/components/utils/request-size-input.tsx
rohitkrai03: Updated.

spadgett (approved):
/lgtm

PR #9024, opened 2021-05-24
Bug 1963213: Fixed bug in Memsource download script

rebeccaalpert:
Quotes prevented the Memsource CLI tool from correctly reading the job information, leading to an API error. I also made a tiny docs update.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1963213.

### DIFF
--- a/frontend/i18n-scripts/memsource-download.sh
+++ b/frontend/i18n-scripts/memsource-download.sh
@@ -30,12 +30,12 @@ DOWNLOAD_PATH="$(mktemp -d)" || { echo "Failed to create temp folder"; exit 1; }
 for i in "${LANGUAGES[@]}"
 do
   COUNTER=0
-  CURRENT_PAGE=$(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number 0 -c uid | tr '\n' ' ')
+  CURRENT_PAGE=( $(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number 0 -c uid) )
   until [ -z "$CURRENT_PAGE" ]
   do
     ((COUNTER++))
     echo Downloading page "$COUNTER"
-    memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id "$CURRENT_PAGE"
+    memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id "${CURRENT_PAGE[@]}"
     CURRENT_PAGE=$(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number "$COUNTER" -c uid | tr '\n' ' ')
   done
 done
@@ -57,4 +57,4 @@ git add public/locales
 git add packages/**/locales
 git commit -m "chore(i18n): update translations
 
-Adding latest translations from Memsource project #$PROJECT_ID"
+Adding latest translations from Memsource project https://cloud.memsource.com/web/project2/show/$PROJECT_ID"
--- a/frontend/i18n-scripts/README.md
+++ b/frontend/i18n-scripts/README.md
@@ -15,7 +15,7 @@ Memsource, the tool the Red Hat Globalization team uses for translation jobs.
 Before running either tool, you must first install the [unofficial Memsource CLI client](https://github.com/unofficial-memsource/memsource-cli-client#pip-install).
 You also have to [configure it with your Memsource login info](https://github.com/unofficial-memsource/memsource-cli-client#configuration-red-hat-enterprise-linux-derivatives).
 
-Example CLI usage for upload script: `yarn memsource-upload -v 4.8 -s "200"`
+Example CLI usage for upload script: `yarn memsource-upload -v 4.8 -s 200`
 * -v is the current OpenShift version
 * -s is the current sprint number
 


spadgett:
/hold

@rebeccaalpert Are you sure the quotes were the issue? It looked like the shell had an unterminated quote when you were demoing, maybe because the notes app converted the quotes to a unicode character.

I'd like to keep the scripts so that they pass shellcheck.

rebeccaalpert:
@spadgett - That was the upload script. I had noticed the download script wasn't working when we got the last batch of files and I narrowed it down to the quotes there.

I did update the README for the upload script.

frontend/i18n-scripts/memsource-download.sh
```diff
@@ -35,7 +35,7 @@ do
   do
     ((COUNTER++))
     echo Downloading page "$COUNTER"
-    memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id "$CURRENT_PAGE"
+    memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id $CURRENT_PAGE
```
spadgett, line 38: I don't understand the issue here. Can you share an example value for `$CURRENT_PAGE`?

rebeccaalpert: Example `$CURRENT_PAGE` call:

```
ralpert$ memsource job list --project-id uKowQDyNXO9LfdYTyhcmK0 --target-lang ja -f value --page-number 0 -c uid | tr '\n' ' '

1cGI1WC1sFfLUE8AWvEp21 zJPMVi9qc7ET35FhBHrsD1 8aAVOfXFfjWI5xdROEgt70 qiLf64McobUX3l44AgG4Y1 l0G9A40uZZxKM4ckVXW3df kbJD17Z69rkqhi0B4V8uqe ieKPp7rKxlb42laqY2zLe3 t7ewcgNYmkpge90APPyMk5 NIF35GJ7938v5JBvoOV1B4 dhGOF5oSIf9f8B1pnnGF26 MuOXVuBNAiEr4JPBuxH1Q4 zVLHl8mHfcW26lDtJ67HU3 n1AQVsBF8aQLei9DhxVEq3 7Z7TtDwsMBaA2J3QNLEk60 FAtAXemjwdVX1ak4sQqLK4 pRoM00PPARy2ZQoU7BlKd5 rrOgA9Vf7o8CdpWaRHIuD3 eEzc3EQZf98c5xMaKHvad1 XHov9I7HQtpg8BFHi8t023 JVAkrLWz4vFs3plSEMjjq1 vTdB1d9b6lSObG1g3NKA56 SqczWrvEewaT7llcQksTM0 hP5TAqoL0rmXvq6bQmPk56 EmM2IE7oiRGX0We1Vn6gR0 vIqMxdlfZiGo1yny3YDZQ3 vHWEiiSGFPr7dJkOZdcF82 eaPUO9XbrKW84JeRyOXKs2 1E1bC16uBY5nZFv4kIse88 0dK0CtEEAFHcsQQTaQTVZ8 X5yGuOOaVArU75SAmCfyr1 i9k1ZJiPGtSleGTGIdjw61 
```

spadgett: Ah, OK. You want to be using arrays here when initializing `$CURRENT_PAGE`. I believe the syntax is

```
CURRENT_PAGE=( $(command) )
```

then it should be

```suggestion
    memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id "${CURRENT_PAGE[@]}"
```

See https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions

rebeccaalpert: I get the same "invalid parameters" error message with this syntax.

spadgett: @rebeccaalpert I think you need to remove ` | tr '\n' ' '` from the command above when you initialize `CURRENT_PAGE`.

We should avoid the unquoted expression and handle this as an array. We don't want it to glob if there are ever special characters.

rebeccaalpert: The issue is that the quotes make it pass in `'A B C'` when the API expects `A B C`.

spadgett: Right, that's why you use an array and `"${CURRENT_PAGE[@]}"`. It means the output of the command where you initialize `CURRENT_PAGE` needs to read it as an array of values.

If you just don't quote, special characters in the values like `*` will be treated as shell globs.

Did you try something like this, removing the `tr` stuff. This is actually simpler if it works (and avoids the problem with globbing).

```bash
CURRENT_PAGE=( $(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number 0 -c uid) )
# ...
memsource job download --project-id "$PROJECT_ID" --output-dir "$DOWNLOAD_PATH/$i" --job-id "${CURRENT_PAGE[@]}"
```

I don't have an environment to try it myself, so I can't test that.

spadgett: In other words, I think your script is initializing `CURRENT_PAGE` to an array with a single string like `["foo bar baz"]` instead of `["foo", "bar", "baz"]`. `"${CURRENT_PAGE[@]}"` should work if it's the latter. If you remove `tr` from the job list command and surround the expression with `(` and `)`, I _think_ it will initialize the array correctly. But I don't have an environment to test.

rebeccaalpert: Thanks, the second suggestion (extra parentheses) seems to actually work.

frontend/i18n-scripts/memsource-download.sh
```diff
@@ -30,12 +30,12 @@ DOWNLOAD_PATH="$(mktemp -d)" || { echo "Failed to create temp folder"; exit 1; }
 for i in "${LANGUAGES[@]}"
 do
   COUNTER=0
-  CURRENT_PAGE=$(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number 0 -c uid | tr '\n' ' ')
+  CURRENT_PAGE=( $(memsource job list --project-id "$PROJECT_ID" --target-lang "$i" -f value --page-number 0 -c uid) )
```
spadgett, line 33: Thanks, this is simpler and more correct since it handle special characters 👍

spadgett (approved):
/lgtm

zherman0 (approved):

PR #9025, opened 2021-05-24
Bug 1957756: Missing already replaced condition in disk modal

afreen23:
The PR #8880 added checking for a replaced disk in list page but we missed out adding in disk replacement modal as well.
This PR adds that  same condition only.
Edit:

While fixing this I found on the test setup we can have more templates on same node, hence changed existing data structure from map (`replacementMap`) to array (`replacedDiskList`)

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
@@ -48,7 +48,7 @@ import {
   OCSColumnStateAction,
   Status,
   OCSDiskStatus,
-  ReplacementMap,
+  ReplacedDisk,
 } from './state-reducer';
 import { CEPH_STORAGE_NAMESPACE, OSD_DOWN_ALERT, OSD_DOWN_AND_OUT_ALERT } from '../../constants';
 
@@ -106,6 +106,7 @@ const diskRow: RowFunction<DiskMetadata, OCSMetadata> = ({
       <TableData className={tableColumnClasses[1]}>{obj.status.state}</TableData>
       <OCSStatus
         ocsState={ocsState}
+        nodeName={nodeName}
         diskName={obj.path}
         diskID={obj.deviceID}
         diskSerial={obj.serial}
@@ -119,14 +120,7 @@ const diskRow: RowFunction<DiskMetadata, OCSMetadata> = ({
         {humanizeBinaryBytes(obj.size).string || '-'}
       </TableData>
       <TableData className={tableColumnClasses[5]}>{obj.fstype || '-'}</TableData>
-      <OCSKebabOptions
-        disk={obj}
-        nodeName={nodeName}
-        alertsMap={ocsState.alertsMap}
-        replacementMap={ocsState.replacementMap}
-        isRebalancing={ocsState.isRebalancing}
-        dispatch={dispatch}
-      />
+      <OCSKebabOptions disk={obj} nodeName={nodeName} ocsState={ocsState} dispatch={dispatch} />
     </TableRow>
   );
 };
@@ -163,6 +157,7 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
       ocsDiskList[metric.device] = {
         osd: metric.ceph_daemon,
         status: Status.Online,
+        node: metric.exported_instance,
       };
       return ocsDiskList;
     }, {});
@@ -170,7 +165,7 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
       (ocsDiskList: OCSDiskList, alert: Alert) => {
         const { rule, labels } = alert;
         const status = getAlertsBasedStatus(rule.name);
-        if (status) ocsDiskList[labels.device] = { osd: labels.disk, status };
+        if (status) ocsDiskList[labels.device] = { osd: labels.disk, status, node: labels.host };
         return ocsDiskList;
       },
       {},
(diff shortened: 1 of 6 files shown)


frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -193,11 +193,13 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
         const { devicePath, deviceID, deviceOsd, deviceNode, deviceSerial } =
           getAnnotations(ti) || {};
         if (devicePath && deviceOsd && deviceNode === nodeName) {
-          data[devicePath] = {
+          data[ti.metadata.name] = {
```
afreen23, line 196: When the same name disk is replaced then we can run into bugs where the same disk is updated.
Hence using template name which will be different.

frontend/packages/ceph-storage-plugin/src/components/modals/disk-replacement-modal.tsx
```diff
@@ -108,8 +109,21 @@ const DiskReplacementAction = (props: DiskReplacementActionProps) => {
     event.preventDefault();
     setProgress(true);
     try {
-      const { osd } = alertsMap[diskName];
-      const replacementStatus = replacementMap[diskName]?.status;
+      const { osd } = alertsMap?.[diskName];
+      const {
+        disk: { id, serial, path, node },
+        status: replacementStatus,
```
GowthamShanmugam, line 115: is this replacementStatus came from the index? i can't find declaration here

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -189,24 +189,26 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
...
             },
             status: getTiBasedStatus(ti.status.conditions?.[0].type),
-          };
+          });
         }
         return data;
-      }, {});
-      if (!_.isEqual(newData, ocsState.replacementMap)) {
+      }, []);
+      if (!_.isEqual(newData, ocsState.replacedDiskList)) {
```
GowthamShanmugam, line 209: Even one disk is not matching for a node then we are considering everything as replaced disk?

GowthamShanmugam: is one have only one disk at a time, that why we are keeping condition like this?

afreen23: This list builds the details of replaced disk on the given node.
We collate all replaced disks data and then pass it down to `OCSStatus` component where per disk is rendered.
If a disk is not for a specific node, then yes we dont want that in this list.

I hope I understand the concern and answered correctly.

frontend/packages/ceph-storage-plugin/src/components/modals/disk-replacement-modal.tsx
afreen23: I have updated this. Please check now.

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -119,14 +120,7 @@ const diskRow: RowFunction<DiskMetadata, OCSMetadata> = ({
...
       <TableData className={tableColumnClasses[5]}>{obj.fstype || '-'}</TableData>
-      <OCSKebabOptions
-        disk={obj}
-        nodeName={nodeName}
-        alertsMap={ocsState.alertsMap}
-        replacementMap={ocsState.replacementMap}
-        isRebalancing={ocsState.isRebalancing}
-        dispatch={dispatch}
-      />
+      <OCSKebabOptions disk={obj} nodeName={nodeName} ocsState={ocsState} dispatch={dispatch} />
```
afreen23, line 123: just passing full state object instead of separate states

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -163,14 +157,15 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
       ocsDiskList[metric.device] = {
         osd: metric.ceph_daemon,
         status: Status.Online,
+        node: metric.exported_instance,
```
afreen23, line 160: aligning metadata for all states.

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
```diff
@@ -163,14 +157,15 @@ const OCSDisksList: React.FC<TableProps> = React.memo((props) => {
...
       };
       return ocsDiskList;
     }, {});
     const newAlertsMap: OCSDiskList = alertsData.reduce(
       (ocsDiskList: OCSDiskList, alert: Alert) => {
         const { rule, labels } = alert;
         const status = getAlertsBasedStatus(rule.name);
-        if (status) ocsDiskList[labels.device] = { osd: labels.disk, status };
+        if (status) ocsDiskList[labels.device] = { osd: labels.disk, status, node: labels.host };
```
afreen23, line 168: aligning metadata for all states.

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
GowthamShanmugam: ack

frontend/packages/ceph-storage-plugin/src/components/disk-inventory/ocs-disks-list.tsx
GowthamShanmugam: ack

GowthamShanmugam:
lgtm

PR #9048, opened 2021-05-26
show menuaction label over model label on multiTabListPage create options

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5893

**Analysis / Root cause**: 
menuaction label was not shown on multiTabListPage create options and always model label where shown.

**Solution Description**: 
Preferred menuaction label over the model label

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


andrewballantyne (approved):
/lgtm

Thanks for the revert + fix of the logic @invincibleJai -- I think this is the better structure for pulling that label out.

PR #9049, opened 2021-05-26
remove visual fix that was breaking patternfly table

wtrocki:
I'm removing CCS override added by @christiemolloy to fix some visual problems with Patternfly Table.
This fix was working fine on chrome but was causing problems on firefox (checkbox was under the text).
Looks like firefox doesn't particularly like setting fixed width on table columns

image

After discussion with @christianvogt I went to:

- Remove override that was causing issue on firefox
- Logged issue in patternfly https://github.com/patternfly/patternfly-react/issues/5835

The drawback of this is that on both chrome and firefox radio column will be extremely long right now. 
I have spent some time trying to figure out some nice hack but that was above level of my basic understanding of PF CSS

### DIFF
--- a/frontend/packages/rhoas-plugin/src/components/service-table/ServiceInstanceTable.scss
+++ b/frontend/packages/rhoas-plugin/src/components/service-table/ServiceInstanceTable.scss
@@ -1,3 +1,0 @@
-.rhoas-service-instance-table tr>*:empty {
-  width: 45px; /* Match width of PF React Table until solution is figured out */
-}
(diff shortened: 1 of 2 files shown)


christianvogt:
/kind bug

@wtrocki can you please post a screenshot of what it looks like now

christianvogt:
/retitle remove visual fix that was breaking patternfly table

christianvogt:
@wtrocki you are missing the removal of the import of the removed scss file (see ServiceInstanceTable.tsx)

wtrocki:
Screenshot 2021-05-26 at 15 44 19

christianvogt:
cc @rohitkrai03 @invincibleJai

invincibleJai:
> Screenshot 2021-05-26 at 15 37 14
> 
> It has 8% fixed size of the screen so it scales differently than text.
> It is very difficult to change width of that on firefox as any width selector doesn't seem to work.
> 
> @christianvogt I have been experimenting a lot and could not find anything that would work on both browsers.
> Do we know anyone from patternfly who could shed some light why this is happening.

tested on safari, chrome and firefox. The feature works as expected I just see some spacing after the radio and next columns.

PR #9052, opened 2021-05-26
Force en_US locale for jest tests

jotak:
Jest could fail when a locale different from en_US was used (e.g.
fr_FR.UTF-8)

In particular there was failures in `packages/operator-lifecycle-manager/src/components/descriptors/status/conditions.spec.tsx`
and `__tests__/units.spec.js`

E.g.
```
    Expected value to equal:
      "10.23 KiB"
    Received:
      "10,23 KiB"
```

(notice the dot notation versus coma)


jotak:
cc @spadgett

spadgett (approved):
/lgtm

PR #9054, opened 2021-05-26
Fix to disable submit button when yaml editor is empty

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5629

**Root cause/Analysis:**
Even when YAML editor is empty create/save buttons are clickable.

**Solution:**
Disable create/save buttons when YAML editor is empty.

**ScreenRecording:**

https://user-images.githubusercontent.com/20724543/119669545-c4e87b80-be55-11eb-8fdc-386d765c73d9.mov

/kind bug

### DIFF
--- a/frontend/public/components/edit-yaml.jsx
+++ b/frontend/public/components/edit-yaml.jsx
@@ -85,6 +85,7 @@ export const EditYAML_ = connect(stateToProps)(
             fileUpload: props.fileUpload,
             showSidebar: !!props.create,
             owner: null,
+            isSaveOrCreateDisabled: true,
           };
           this.monacoRef = React.createRef();
           // k8s uses strings for resource versions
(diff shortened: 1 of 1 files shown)


spadgett:
This is actually working as intended. The admin perspective intentionally leaves the buttons enabled on forms. Unfortunately, we handle form validation differently in the different perspectives :(

I suspect this will conflict with #8865, which is making significant changes to import YAML. Can we hold until after that feature merges?

cc @jcaianirh 

/hold

rohitkrai03:
@spadgett If this is the intended behaviour from admin perspective then should we just mark this bug obsolete and close the PR?

divyanshiGupta:
> This is actually working as intended. The admin perspective intentionally leaves the buttons enabled on forms. Unfortunately, we handle form validation differently in the different perspectives :(
> 
> I suspect this will conflict with #8865, which is making significant changes to import YAML. Can we hold until after that feature merges?
> 
> cc @jcaianirh
> 
> /hold

@spadgett should we close the PR if this is the intended behaviour?

divyanshiGupta:
Closing this PR and marking the associated bug as obsolete. If needed later the bug can be opened again.

PR #9055, opened 2021-05-26
add dynamic sdk support for topology extensions

glekner:


frontend/packages/topology/src/components/graph-view/Topology.tsx
```diff
@@ -138,10 +143,14 @@ const Topology: React.FC<TopologyProps &
   const componentFactoryExtensions = useExtensions<TopologyComponentFactory>(
     isTopologyComponentFactory,
   );
+  const [dynamicComponentFactoryExtensions, isFactoryComponentsResolved] = useResolvedExtensions<
+    DynamicTopologyComponentFactory
+  >(isDynamicTopologyComponentFactory);
```
christianvogt, line 148: Can we move away from `useExtensions` for the old extension and make use of `useResolvedExtensions`?
Then combine the results two variables: one for the resolved state and one for an array of factories.

This way it'll be two code paths for setup but consumption will be simple.

frontend/packages/topology/src/components/graph-view/Topology.tsx
```diff
@@ -230,12 +239,18 @@ const Topology: React.FC<TopologyProps &
...
+            ...dynamicComponentFactoryExtensions.map((f) => f.properties.getFactory),
+          ]);
+        })
+        .catch(() => {});
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [componentFactoriesPromises, isFactoryComponentsResolved]);
 
   React.useEffect(() => {
     if (componentFactoriesPromises.length && !componentFactories.length) {
```
christianvogt, line 256: Since you've separated the static and dynamic extensions into different variables, this check is no longer valid as it only applies to one.

See other comment about using `useResolvedExtensions` for both static and dynamic plugins.

frontend/packages/topology/src/data-transforms/DataModelProvider.tsx
```diff
@@ -18,13 +23,23 @@ const DataModelProvider: React.FC<DataModelProviderProps> = ({ namespace, childr
   }, [namespace]);
 
   const modelFactories = useExtensions<TopologyDataModelFactory>(isTopologyDataModelFactory);
+  const [dynamicModelFactories, isModelResolved] = useResolvedExtensions<
+    DynamicTopologyDataModelFactory
+  >(isDynamicTopologyDataModelFactory);
```
christianvogt, line 28: Can both paths use `useResolvedExtensions` so that we don't have one path doing manual resolution.

frontend/packages/topology/src/components/dropdowns/ApplicationDropdown.tsx
```diff
@@ -40,10 +45,13 @@ interface ApplicationDropdownProps {
 const ApplicationDropdown: React.FC<ApplicationDropdownProps> = ({ namespace, ...props }) => {
   const { t } = useTranslation();
   const modelFactories = useExtensions<TopologyDataModelFactory>(isTopologyDataModelFactory);
+  const [dynamicModelFactories] = useResolvedExtensions<DynamicTopologyDataModelFactory>(
+    isDynamicTopologyDataModelFactory,
+  );
```
christianvogt, line 50: Can we update the static extension to make use of `CodeRef` then simplified the resolution with `useResolvedExtensions` 

We probably don't want to render the firehose component until all extensions have been resolved.

glekner:
@christianvogt modified extensions, now consuming all with `useResolvedExtensions`

frontend/packages/helm-plugin/src/topology/helmFilters.ts
```diff
@@ -35,3 +36,14 @@ export const applyHelmDisplayOptions = (model: Model, filters: DisplayFilters):
 };
 
 export const applyDisplayOptions = () => applyHelmDisplayOptions;
+
+export const getHelmWatchedResources = (namespace: string): WatchK8sResources<any> => {
```
christianvogt, line 40: Why was this moved to a filters module?

frontend/packages/rhoas-plugin/src/topology/createConnector.ts
```diff
@@ -19,3 +22,16 @@ export const getCreateConnector = (createHints: string[], source: Node, target:
   }
   return null;
 };
+
+export const getRhoasWatchedResources = (
```
christianvogt, line 26: Moving to `createConnector` doesn't seem like a logical home.

frontend/packages/topology/src/operators/operatorFilters.ts
```diff
@@ -35,3 +39,25 @@ export const applyOperatorDisplayOptions = (model: Model, filters: DisplayFilter
 };
 
 export const applyDisplayOptions = () => applyOperatorDisplayOptions;
+
+export const getOperatorWatchedResources = (namespace: string): WatchK8sResources<any> => {
```
christianvogt, line 43: Same comment about location.

christianvogt (commented):
Operator backed services aren't showing up correctly. Use the dev catalog
Before:
image

After:
image

There's also flashing of the old value when switching from one namespace to another.

christianvogt:
Display filters are missing:

Before:
image

After:
image

christianvogt:
@glekner the latest changes fixed the missing visuals.

One outstanding issue remains. Flashing of the old value when switching from one namespace to another.
In the following gif, the visual of the operator backed service disappears when we change namespaces (expected), then then reappears (unexpected) before the proper topology for the new namespace is displayed.

topology-flash

glekner:
hey again,
the flashing is fixed by reverting `ModelFactory` to use `useExtensions`.
since you can’t pass a function without a CodeRef for the dynamic extensions, instead we are passing the resources that need to be watched, and we build a function for them in the consuming of the extensions. please review @christianvogt

glekner:
/test e2e-gcp-console



<br class="Apple-interchange-newline">

rohitkrai03 (changes requested):

frontend/packages/console-dynamic-plugin-sdk/src/extensions/topology.ts
```diff
@@ -88,7 +88,7 @@ export const isTopologyCreateConnector = (e: Extension): e is TopologyCreateConn
 export const isTopologyDataModelFactory = (e: Extension): e is TopologyDataModelFactory =>
   e.type === 'console.topology/data/factory';
 
-export const isTopologyDisplayFilters = (e: Extension): e is TopologyDisplayFilters =>
+export const isTopologyDisplayFilter = (e: Extension): e is TopologyDisplayFilters =>
```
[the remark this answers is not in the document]
glekner: to match the static extension

frontend/packages/console-dynamic-plugin-sdk/src/extensions/topology.ts
```diff
@@ -39,7 +39,7 @@ export type TopologyDataModelFactory = ExtensionDeclaration<
     /** Priority for the factory */
     priority: number;
     /** Resources to be fetched from useK8sWatchResources hook. */
-    resources?: CodeRef<(namespace: string) => WatchK8sResources<any>>;
+    resources?: WatchK8sResources<any>;
```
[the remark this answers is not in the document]
glekner: since you cant set a function without a CodeRef, i'm building the namespaced routes here
https://github.com/openshift/console/pull/9055/files#diff-281f5462e39247e7a7894ee5ad688563cf18b46d97e7707da0b660b79c236f41R17

frontend/packages/topology/src/components/dropdowns/ApplicationDropdown.tsx
```diff
@@ -56,7 +69,9 @@ const ApplicationDropdown: React.FC<ApplicationDropdownProps> = ({ namespace, ..
       ...watchedBaseResources[key],
       prop: key,
     }));
-  }, [modelFactories, namespace]);
+
+    return [];
```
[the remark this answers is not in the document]
glekner: thanks, removed

frontend/packages/console-dynamic-plugin-sdk/src/extensions/topology.ts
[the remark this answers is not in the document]
rohitkrai03: So why not make it a coderef?

frontend/packages/console-dynamic-plugin-sdk/src/extensions/topology.ts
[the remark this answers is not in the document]
rohitkrai03: Okay, I guess it was a typo in the static extensions. But we can keep it like this for now.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/topology.ts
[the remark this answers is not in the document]
christianvogt: I talked to @glekner  about this too. I don't see an issue with this change unless the extension wants to hard code a namespace. In which case we can make sure to not set the namespace if one is already provided.

I don't think we have such a use case at this time that will be negatively affected by this.

Note that we still need to review the topology extensions for their exposed API and at that time may change things.

frontend/packages/topology/src/data-transforms/DataModelProvider.tsx
```diff
@@ -10,6 +14,23 @@ interface DataModelProviderProps {
...
+) =>
+  factories.map(({ properties, ...ext }) => ({
+    ...ext,
+    properties: {
+      ...properties,
+      resources: (ns: string) =>
+        Object.assign(
+          {},
+          ...Object.entries(properties.resources).map(([k, v]) => ({
+            [k]: { ...v, namespace: ns },
```
christianvogt, line 28: We should probably accept the namespace provided before overriding? Not sure we have a use case for this yet but if an extension wanted to pull resources from a specific namespace we'd have to do:

```suggestion
            [k]: { namespace: ns, ...v },
```

rohitkrai03 (approved):
/lgtm
/approve

PR #9056, opened 2021-05-26
Fix git type dropdown reset issue

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5889

**Analysis / Root cause**: The git type dropdown uses `gitTypeTouched` to show the dropdown and use user entered `git.type` value. We also set `gitTypeTouched` to `false` when user selects it. This caused the dropdown to reset back to `Others` every time user selected an option from the dropdown.

**Solution Description**: 
- Only reset the touched state of the dropdown if there is an error with the git repo url. Which means every time user enters a new URL.
- Also do not do any detection or validation of repo when the repo URL is not valid.

Before -

https://user-images.githubusercontent.com/6041994/119678832-7c34c080-be5d-11eb-95ce-3f999acf0051.mov

After -

https://user-images.githubusercontent.com/6041994/119678851-81920b00-be5d-11eb-979e-d2024f16b670.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
+++ b/frontend/packages/dev-console/src/components/import/git/GitSection.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { isEmpty } from 'lodash';
-import { useFormikContext, FormikValues, FormikTouched } from 'formik';
+import { useFormikContext, FormikValues, FormikTouched, FormikErrors } from 'formik';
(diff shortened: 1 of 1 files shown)


rottencandy:
Verified changes locally.
/lgtm

rottencandy (approved):
/lgtm

PR #9059, opened 2021-05-26
fix terminal new tab redirect

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5885

This PR fixes the flag check for DevWorkspace. `useFlag` returns undefined which means the flag check is pending and we redirect to the `/`. Check for `false` while doing flag check.

new-tab


rottencandy:
Tested locally
/lgtm

invincibleJai:
/approve

verified the changes and works as expected

PR #9061, opened 2021-05-26
Bug 1965116: Add max height to resource and filter drop downs

kdoberst:
The CSS for the newly added PatternFly Select does not have a default max-height.   Add the max-height setting from the removed custom drop down component CSS to the the PatternFly select.

Fixes an issue introduced as part of the following PRs:
 - https://github.com/openshift/console/pull/8877
 - https://github.com/openshift/console/pull/8802


spadgett (approved):
/lgtm

PR #9064, opened 2021-05-27
[WIP] Bug 1965212: Query Browser: Fix aria title being visible

kyoto:
The title was visible, which partially obscured the tooltip.

Using `ariaDesc` instead of `ariaTitle` prevents it being displayed.

### Before
screenshot


lihongyan1:
Test with PR

kyoto:
It seems this is now also a problem in the latest Chrome version.

kyoto:
This change causes an accessibility test failure, so I'm closing this PR.

PR #9065, opened 2021-05-27
Bug 1962429: Images are now saved in redux and can be accessed earlier in vm creation

metalice:

**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1962429

**Analysis / Root cause**: 
Promise resolves too late when config map is needed before

**Solution Description**: 
Making a call earlier and saving the result in redux


yaacov:
/hold
@metalice can you check the failing tests ?

PR #9069, opened 2021-05-27
use quickstarts library

jschuler:
Our goal is to have a single code base related to quick starts that can be used across our products, like OpenShift console, RHODS, Managed Services, etc.

To that purpose, quick start related code was copied out and modularized at https://github.com/patternfly/patternfly-quickstarts.

A sample app is built based on the package:
https://quickstarts.netlify.app/quickstarts

This PR:
- Adds dependency `"@patternfly/quickstarts"`
- Changing language in Console will load the quickstart resource bundle and update the quick start context
- Removes components and code that was pulled out into the package
- Adds the `@patternfly/quickstarts/dist/quickstarts.css` that contains the quick start related styles
- Quick start related type imports come from the `@patternfly/quickstarts` package
- Updates quick-start-context for use with the module, which has some extra context values to handle:
  - OpenShift adding more custom markdown extensions
  - Hiding the catalog card footers (OS removed the footers, but other products still rely on them so it is an option now)
- Removes quick start related jest tests. they are included with the module.


andrewballantyne:
Now that `master` is 4.9, did you want to give this a rebase @jschuler and we can look into getting it merged?

frontend/public/components/modals/language-preferences-modal.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import i18next, { TFunction } from 'i18next';
-
+// eslint-disable-next-line import/no-unresolved
+import { QuickStartContextValues } from '@patternfly/quickstarts';
```
sahil143, line 5: why is eslint-disable used here?

sahil143, line 5: why is eslint-disable needed here?

sahil143 (commented):
Styling for markdown code snippets  is a bit weird 
Screenshot 2021-06-29 at 11 07 37 AM
Screenshot 2021-06-29 at 11 06 26 AM

and for the resource dropdown under search section 
Screenshot 2021-06-29 at 11 19 43 AM

frontend/public/components/modals/language-preferences-modal.tsx
jschuler: I was getting this eslint error, only when I was pulling in a type from the lib as I am doing here. I can take another look tho to see if I can fix it without disabling for this line.

jschuler:
@sahil143 Thanks for taking a look, I will take a look soon to reproduce and check out the CSS issues you saw

jschuler:
@sahil143 style bugs should be resolved now. I removed all global styles from the `quickstarts.css`.

sahil143 (changes requested):
@jschuler Thanks for fixing CSS issues.
I found one more issue. quick starts on getting started card in both Admin and dev perspective are not opening even though id is being added in query params.
Kapture 2021-07-01 at 21 06 49

jschuler:
@sahil143 Thanks, I fixed a bug in the library, this should work now

frontend/packages/console-app/src/components/quick-starts/QuickStartCatalogPage.tsx
```diff
@@ -1,31 +1,31 @@
...
-import QuickStartCatalog from './catalog/QuickStartCatalog';
 import QuickStartsLoader from './loader/QuickStartsLoader';
 
 const QuickStartCatalogPage: React.FC = () => {
   const { t } = useTranslation();
   return (
     <>
       <Helmet>
-        <title>{t('quickstart~Quick Starts')}</title>
+        <title>{t('console-app~Quick Starts')}</title>
```
sahil143, line 13: this should be 
```suggestion
        <title>{t('quickstart~Quick Starts')}</title>
```

since we have separate locales for quick starts

frontend/packages/console-app/src/components/quick-starts/QuickStartCatalogPage.tsx
```diff
@@ -1,31 +1,31 @@
...
-        </div>
-      </div>
       <QuickStartsLoader>
         {(quickStarts, loaded) =>
-          loaded ? <QuickStartCatalog quickStarts={quickStarts} /> : <LoadingBox />
+          loaded ? (
+            <PfQuickStartCatalogPage
+              quickStarts={quickStarts}
+              showFilter
+              title={t('console-app~Quick Starts')}
```
sahil143, line 21: same as above

frontend/packages/console-app/src/components/quick-starts/QuickStartCatalogPage.tsx
```diff
@@ -1,31 +1,31 @@
...
+              hint={t(
+                'console-app~Learn how to create, import, and run applications on OpenShift with step-by-step instructions and tasks.',
```
sahil143, line 23:
```suggestion
              hint={t(
                'quickstart~Learn how to create, import, and run applications on OpenShift with step-by-step instructions and tasks.',
```

sahil143:
This PR also fixes bug: https://issues.redhat.com/browse/ODC-6090

frontend/packages/console-app/src/components/quick-starts/QuickStartCatalogPage.tsx
jschuler: ok, so there are 2 reasons why I moved those 2 translations into console
- Those keys are not used in the quick start library itself, they are only used by OpenShift to add title and subtitle to the catalog
- Because there are no more `quickstart~` keys used in the app (I just load the whole quickstart resource bundle and feed it to the library via context value), if there is a `quickstart~` key, the i18n script will actually try to update the `quickstart.json` file with only the keys it found, which there would only be 2 then, so the CI job would fail.

jschuler:
@sahil143 all quickstart keys were consolidated recently into the console-app.json file https://github.com/openshift/console/pull/9421

jschuler:
This PR also fixes bug https://issues.redhat.com/browse/ODC-6091

jschuler:
Looks like kubevirt tests are broken at the moment
https://github.com/openshift/console/pull/9096#issuecomment-875670647

sahil143:
@jschuler most of the quick starts text is not properly localized
Screenshot 2021-07-09 at 10 40 50 PM
Screenshot 2021-07-09 at 10 41 11 PM

jschuler:
@sahil143 that is due to recent changes to master where the quick start translations were removed
https://github.com/openshift/console/pull/9421/files
I am told that they will come back in future when the translations come back from the Red Hat Globalization team

sahil143 (approved):
/lgtm

sahil143:
Psuedolocalization is not working properly as well. I'm not sure if it should since the library itself doesn't use i18n, just takes the translated keys and values passed to it
Screenshot 2021-07-10 at 12 21 18 AM

frontend/public/components/modals/language-preferences-modal.tsx
```diff
@@ -13,6 +14,12 @@ import {
 
 const LanguagePreferencesModal = (props: LanguagePreferencesModalProps) => {
   const { i18n, t } = useTranslation();
+  const { setResourceBundle } = props.quickStartContext;
+  i18n.on('languageChanged', (lng) => {
+    // Update language resource of quick starts components
+    const resourceBundle = i18n.getResourceBundle(lng, 'console-app');
+    setResourceBundle(resourceBundle, lng);
+  });
```
christianvogt, line 22: This needs to be inside an effect and the listener disposed when unmounted.

analyze.sh
```diff
@@ -14,7 +14,7 @@ if [ -d "$ARTIFACT_DIR" ]; then
   cp public/dist/report.html "${ARTIFACT_DIR}"
 fi
 
-MAX_BYTES=3145728 # ~3 MiB
+MAX_BYTES=3250586 # ~3.1 MiB
```
christianvogt, line 17: fyi @spadgett

frontend/packages/console-app/src/components/quick-starts/QuickStartCatalogPageAsync.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import { AsyncComponent } from '@console/internal/components/utils';
+
+const QuickStartCatalogPageAsync: React.FC = () => (
+  <AsyncComponent loader={() => import('./QuickStartCatalogPage').then((c) => c.default)} />
```
christianvogt, line 5:
```suggestion
  <AsyncComponent loader={() => import('./QuickStartCatalogPage' /* webpackChunkName: "quick-start" */).then((c) => c.default)} />
```

frontend/packages/console-app/src/components/quick-starts/QuickStartDrawerAsync.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { AsyncComponent } from '@console/internal/components/utils';
+
+const QuickStartDrawerAsync: React.FC = ({ children, ...props }) => (
+  <AsyncComponent loader={() => import('./QuickStartDrawer').then((c) => c.default)} {...props}>
```
christianvogt, line 5: Can we use the same chunk name for all quick start async components? I think that will bundle the chunks together.
@vojtechszocs @spadgett do you know?
```suggestion
  <AsyncComponent loader={() => import('./QuickStartDrawer' /* webpackChunkName: "quick-start" */).then((c) => c.default)} {...props}>
```

frontend/packages/console-app/src/plugin.tsx
```diff
@@ -202,7 +202,7 @@ const plugin: Plugin<ConsumedExtensions> = [
       loader: async () =>
         (
           await import(
-            './components/quick-starts/QuickStartCatalogPage' /* webpackChunkName: "co-quick-start" */
+            './components/quick-starts/QuickStartCatalogPageAsync' /* webpackChunkName: "co-quick-start" */
```
christianvogt, line 205: Again, I think we can use the same chunk name for all things quick starts?
```suggestion
            './components/quick-starts/QuickStartCatalogPageAsync' /* webpackChunkName: "quick-start" */
```

frontend/public/components/modals/language-preferences-modal.tsx
christianvogt: @jschuler is this still required?

analyze.sh
jschuler: When I analyzed the bundles, I noticed that some of the modules were shifted around, and some into the vendors bundle. Overall, the webpack bundle actually decreased in size, but this specific vendors bundle went up a bit.

jschuler:
@sahil143 the pseudo localization plugin works on text that was rendered with i18next `t` function. Since I am passing the key/value pairs to the quick start library and I don't want them to be interpolated, I fetch them via i18next.getResourceBundle.
But I can load the pseudo plugin in that case myself and apply the transformation to the resource bundle before passing to the library, this way you can still see in OpenShift that the text is captured. Next commit will have the code, but you can see it already working in this screenshot.
Screen Shot 2021-07-09 at 3 45 24 PM

christianvogt:
Would we not be better off to put the `i18n` `t` function into a callback on the context so that we don't need to put the whole resource bundle in it?
I have some questions about the method in which the quick-starts repo does localization but I don't want to stop the merging because of this as any change would need to be first done in the quick-starts repo.

Maybe we can discuss how the quick starts do localization again.

jschuler:
@christianvogt we can definitely revisit how the text is passed to the lib. Up until very recently, the quick start keys were localized in their own quickstart bundle but a recent PR has merged these all together into the console-app bundle. Perhaps we should restore the quickstart bundle back to its own bundle since it's handled a bit diffferently. The `i18n`, `t` callback idea sounds interesting, can you expand a bit on this idea? I think as long as it doesn't complicate the library more and let other apps pass in their own text overrides without needing i18n it'll be good.

analyze.sh
jschuler: To expand further, the vendor bundle gained around 190 KB, which mostly comes from the addition of the quickstart lib (+93.2 KB), and the showdown package (+93.64 KB) which webpack automatically moved into this bundle (it used to live in a a separate chunk before).
Comparing the whole parsed bundle size at the time of writing, master is at 25.03 MB, and this branch is at 25.02 MB

frontend/packages/console-app/src/components/quick-starts/utils/quick-start-context.tsx
```diff
@@ -1,51 +1,50 @@
...
+  const language = lng || localStorage.getItem('bridge/language') || 'en';
+  const consoleBundle = resourceBundle;
+  if (pseudolocalizationEnabled && language === 'en') {
+    const pseudo = new Pseudo({ enabled: true, wrapped: true });
+    Object.keys(consoleBundle).forEach((key) => {
+      consoleBundle[key] = pseudo.process(consoleBundle[key], '', {}, { language });
+    });
+  }
+
+  return consoleBundle;
```
christianvogt, line 38: Do not mutate the `resourceBundle` because it's causing the pseudo localization to stack.
```suggestion
  let consoleBundle = resourceBundle;
  if (pseudolocalizationEnabled && language === 'en') {
    consoleBundle = {};
    const pseudo = new Pseudo({ enabled: true, wrapped: true });
    Object.keys(resourceBundle).forEach((key) => {
      consoleBundle[key] = pseudo.process(resourceBundle[key], '', {}, { language });
    });
  }

  return consoleBundle;
```

christianvogt:
Seems to all work well.
/approve

analyze.sh
spadgett: @jschuler Sorry, I missed this comment. If we're going to increase the limit, we need more room. If you set it at just above what we're currently using, minor changes in size in our dependencies will cause failures (as seems to have happened in #9419). This job is meant to catch large dependencies that are accidentally added to the main vendor bundle.

The important bundle is the main vendor bundle as that's what's loaded on console load. The other bundles are lazy-loaded. It's better to have a smaller main vendor bundle even if the total size is a little larger.

Is there a reason we couldn't lazy-load the quick starts library?

If we're pulling showdown into the main vendor bundle, we should remove the code that lazy loads it in other places. But it would be nice to lazy-load libraries of significant size when it makes sense.

I notice also that xterm and victory somehow have been pulled into the main vendor bundle likely in an unrelated change. We should look at these as well.

PR #9070, opened 2021-05-27
Support code references at any level within extension's properties

vojtechszocs:
With this PR, `CodeRef<T>` values can appear at any level within extension's properties object.

For example, using the following Console extension declaration

```ts
export type Foo = ExtensionDeclaration<'console.foo', {
  bar: CodeRef<VoidFunction>;
  baz: { qux: CodeRef<VoidFunction> };
}>;
```

both `e.properties.bar` and `e.properties.baz.qux` values will be
1. parsed (`EncodedCodeRef` to `CodeRef`) when loading the given plugin
2. resolved (`CodeRef<T>` to `T`) when consumed via `useResolvedExtensions` hook

Detection of code references (both encoded & executable) relies on a recursive equivalent of [`_.forOwn`](https://docs-lodash.com/v4/for-own/) function that traverses plain objects and arrays. This means that e.g. `CodeRef<VoidFunction>[]` (array of code references) properties are supported as well.

cc @bipuladh @christianvogt


frontend/packages/console-dynamic-plugin-sdk/tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "experimentalDecorators": true,
     "sourceMap": true,
     "noUnusedLocals": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "lib": ["es2020", "dom"]
```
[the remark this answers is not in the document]
vojtechszocs: Another alternative is to use existing `"target": "es2016"` with `"lib": ["es2016", "es2020.promise", "dom"]`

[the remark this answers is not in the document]
vojtechszocs: > Another alternative is to use `"target": "es2016"` with `"lib": ["es2016", "es2020.promise", "dom"]` to add specific types we need on top of the existing target spec.

I've updated the PR to use this approach.

yapei:
Tested with following steps
```
in one terminal,
$ git clone git@github.com:vojtechszocs/console.git
$ cd console; git checkout nested-coderefs
$ git apply nested-coderefs-plugin-test.diff
$ cd frontend; yarn install
$ cd dynamic-demo-plugin; yarn install; yarn build; yarn http-server

in another terminal,
$ cd /back/to/console/root; ./build.sh
$ ./bin/bridge -plugins console-demo-plugin=http://localhost:9001/
```
then check browser console logs, we can see
```
Dynamic plugins: [console-demo-plugin]
Loading plugin manifest from http://localhost:9000/api/plugins/console-demo-plugin/plugin-manifest.json
Loading entry script for plugin console-demo-plugin@0.0.0 from http://localhost:9000/api/plugins/console-demo-plugin/plugin-entry.js
Added plugin console-demo-plugin@0.0.0
main-chunk-7d14ad1bc734ef26df43.min.js:1 Plugin console-demo-plugin@0.0.0 is now enabled
bar.ts:4 testHandler called (e,t)=>{f.b.dispatch(y(e,t))}
bar.ts:8 testing 1
bar.ts:8 testing 2
```

bipuladh (approved):
/lgtm

bipuladh:
```
Unable to connect to the server: dial tcp: lookup api.ci-op-hk8n5pm2-75d12.********************************* on 172.30.0.10:53: no such host
```
Looks like the test-cluster is unable to come up.

vojtechszocs:
/label px-approved
/label docs-approved

We have a general px- and docs-approval for Console dynamic plugin changes.

vojtechszocs:
/hold

`e2e-gcp-console` is failing consistently, need to investigate.

jerolimov:
Hey @vojtechszocs, I investigate why the E2E tests fails. They missed the guided tour when the user switches the first time to the developer perspective.

And thanks these tests I can confirm with some manual tests now that this alert / modal doesn't appear anymore. The guided tour is also not shown anymore in the dropdown menu in the top right corner.

Menu for kubeadmin before:

before-kubeadmin-menu

Menu for kubeadmin with this change:

after-kubeadmin-menu

Similar for a developer before:

before-developer-menu

Similar for a developer with this change:

after-developer-menu

frontend/packages/console-dynamic-plugin-sdk/src/utils/__tests__/url.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { resolveURL } from '../url';
+
+describe('resolveURL', () => {
+  const getDocumentOrigin = () => 'https://example:1234';
+
+  it('uses the base URL as-is if it has the protocol', () => {
+    expect(resolveURL('http://test', 'foo', getDocumentOrigin)).toBe('http://test/foo');
+    expect(resolveURL('http://test/', 'foo', getDocumentOrigin)).toBe('http://test/foo');
+    expect(resolveURL('http://test/foo/', 'bar', getDocumentOrigin)).toBe('http://test/foo/bar');
```
jerolimov, line 9:
```suggestion
    expect(resolveURL('http://test/foo', 'bar', getDocumentOrigin)).toBe('http://test/foo/bar');
    expect(resolveURL('http://test/foo/', 'bar', getDocumentOrigin)).toBe('http://test/foo/bar');
```

frontend/packages/console-dynamic-plugin-sdk/src/utils/__tests__/url.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  });
+
+  it("makes the base URL relative to document origin if it's missing the protocol", () => {
+    expect(resolveURL('/', 'foo', getDocumentOrigin)).toBe('https://example:1234/foo');
+    expect(resolveURL('/foo/', 'bar', getDocumentOrigin)).toBe('https://example:1234/foo/bar');
```
jerolimov, line 14: A second case would be useful here as well:

```suggestion
    expect(resolveURL('/foo', 'bar', getDocumentOrigin)).toBe('https://example:1234/foo/bar');
    expect(resolveURL('/foo/', 'bar', getDocumentOrigin)).toBe('https://example:1234/foo/bar');
```

frontend/packages/console-dynamic-plugin-sdk/src/utils/store.ts
```diff
@@ -1,7 +1,11 @@
...
 import { Extension } from '../types';
 
+/**
+ * Recursively merges the given object into extension's properties.
+ */
 export const mergeExtensionProperties = <E extends Extension>(e: E, properties: {}): E =>
   Object.freeze({
     ...e,
-    properties: Object.assign({}, e.properties, properties),
+    properties: _.merge({}, e.properties, properties),
```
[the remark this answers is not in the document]
jerolimov: @vojtechszocs This is the change which breaks the guided tour! See above.

jerolimov:
The last comment above shows that this PR breaks the devconsole Guided Tour. As mentioned, the change from `Object.assign` to `_.merge` in `store.ts` breaks this. See https://github.com/openshift/console/pull/9070

/lgtm cancel

vojtechszocs:
@jerolimov Thanks so much for looking into this! I'll rebase & update PR accordingly.

frontend/packages/console-plugin-sdk/src/utils/extension-i18n.ts
```diff
@@ -0,0 +1,15 @@
+import { TFunction } from 'i18next';
+import { deepForOwn } from '@console/dynamic-plugin-sdk/src/utils/object';
+import { Extension } from '../typings';
+
+export const isTranslatableString = (value): value is string => {
+  return typeof value === 'string' && /^%.+%$/.test(value);
```
jerolimov, line 6: I don't like micro optimizations, but I also think its not necessary to create/compile/match a regex pattern here.

```suggestion
  return typeof value === 'string' && value.startsWith('%') && value.endsWith('%');
```

jerolimov:
Thanks for the update. The Guided tour is back again. I also tested a code ref on a second level by changing the ContextProvider extension locally. Works fine. :+1:

/lgtm

jerolimov (approved):

frontend/packages/console-plugin-sdk/src/utils/extension-i18n.ts
vojtechszocs: Agreed, we can use `startsWith '%' and endsWith '%' and length > 2` condition instead of RegExp here.

jerolimov:
> ✖  addFlow/create-from-devfile.feature
> ✖  addFlow/create-from-docker-file.feature
> ✖  addFlow/create-from-git.feature

/retest

vojtechszocs:
Rebased, updated `isTranslatableString` function & added a unit test.

jerolimov (approved):
Tested nested code refs again and works. Guided tour also works.

/lgtm

PR #9072, opened 2021-05-27
ODC-5898-Configuring gherkin-lintrc file

makambalaji:
**Description**:
`.gherkin-lintrc` is used by QE team. so moving it to `dev-console/integration-tests` folder and updating the `gherkin-lint` script in package.json file

Including all required tags and modifying below rules due to recently observed scenarios

> While uploading scenarios to polarion, switching off the "no-homogenous-tags"
> Updated allowed tags w.r.to new feature files

**Jira Id**: https://issues.redhat.com/browse/ODC-5898


andrewballantyne (approved):
I don't think you should be moving the global lint file under dev-console, but I think you're trying to get around the approval process for the future.

It's your file, if you want to put it in an odd spot, who are we to argue.

makambalaji:
> I don't think you should be moving the global lint file under dev-console, but I think you're trying to get around the approval process for the future.
> 
> It's your file, if you want to put it in an odd spot, who are we to argue.

gherkin-linter file, doesn't have any impact on frontend code and its mainly used to fix the lint issues on gherkin feature files. So I thought it's better to keep in QE owned folder

PR #9074, opened 2021-05-27
Bug 1951043: Fix Pipeline Parameters in Modals accept empty string defaults

andrewballantyne:
**Analysis / Root cause**:
All parameters were always required. Empty string parameter defaults were not recognized as a valid default by our validation schema.

**Solution Description**:
Allow any parameter that has a default property to accept an empty string.

Caveat is empty string support for required (aka no `default` property) params is not supported. This would require a fundamental UX change in the way the console inputs work today. (no way to determine if you want empty string or an empty field).

Changes to the UI are the omission of the `*` on any parameter with a `default` property -- effectively no change.

No change, tests moved around a bit though.

**Test setup:**

I used my [parameter echo example](https://github.com/andrewballantyne/pipeline-scratch-pad/tree/master/examples/param-echo) -- but any Pipeline that accepts at least 1 param can be modified to have or not have a default for that parameter.

- Test with `default == ‘test’`
- Test with `default == ‘’`
- Test with no `default`

**Browser conformance**:
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified on start pipeline & add trigger modal, works like charm :tada: 
/lgtm

PR #9075, opened 2021-05-27
Missing parens in the Last deployment column (#5899)

keithchong:
This is a bug fix.  The brackets are missing around the environment.

See Jira 5899.


christianvogt:
fixes: https://issues.redhat.com/browse/ODC-5899
/lgtm
/approve

PR #9077, opened 2021-05-27
Bug 1965391: Skip to 3rd step when SC present

rexagod:
Jump to the 3rd step in attached-devices mode when a no-provisioner
storage class has already been created. Also, keep the URL and wizard in
sync while doing so, for an overall better user experience (for
instance, if the user wants to bookmark the page they are currently on,
hence I felt that synchronization was necessary).

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

https://user-images.githubusercontent.com/33557095/119907592-3f0c2380-bf6e-11eb-8087-9c4480118453.mp4

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
@@ -23,6 +23,8 @@ import { OCS_ATTACHED_DEVICES_FLAG } from '@console/local-storage-operator-plugi
 import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager';
 import { resourcePathFromModel } from '@console/internal/components/utils';
 import { getName } from '@console/shared';
+import { useK8sGet } from '@console/internal/components/utils/k8s-get-hook';
+import { StorageClassModel } from '@console/internal/models';
 import { initialState, reducer } from './reducer';
 import {
   DiscoverDisks,
@@ -32,7 +34,7 @@ import { CreateStorageClass } from './install-wizard-steps/create-storage-class/
 import { StorageAndNodes } from './install-wizard-steps/storage-and-nodes-step';
 import { ReviewAndCreate } from './install-wizard-steps/review-and-create-step';
 import { Configure } from './install-wizard-steps/configure-step';
-import { taintNodes } from '../../../utils/install';
+import { taintNodes, filterSCWithNoProv } from '../../../utils/install';
 import {
   CreateStepsSC,
   MINIMUM_NODES,
@@ -121,11 +123,25 @@ const CreateStorageClusterWizard: React.FC<CreateStorageClusterWizardProps> = ({
   const [showInfoAlert, setShowInfoAlert] = React.useState(true);
   const [inProgress, setInProgress] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState('');
+  const [isForceNavigated, setForceNavigated] = React.useState(false);
   const flagDispatcher = useDispatch();
 
   const discoveryNodes = state.lvdIsSelectNodes ? state.lvdSelectNodes : state.lvdAllNodes;
 
-  const { getStep, getParamString, getIndex, getAnchor } = navUtils;
+  const { getStep, getIndex, getAnchor } = navUtils;
+
+  const [hasNoProvSC, setHasNoProvSC] = React.useState(false);
+
+  const [storageClasses, isLoaded, err] = useK8sGet(StorageClassModel);
+
+  React.useEffect(() => {
+    if (!err && isLoaded) {
+      const filteredSCData = storageClasses['items'].filter(filterSCWithNoProv);
+      if (filteredSCData.length) {
+        setHasNoProvSC(true);
+      }
+    }
+  }, [appName, err, isLoaded, ns, storageClasses]);
 
   /**
    * This custom footer for wizard provides a control over the movement to next step.
@@ -285,16 +301,12 @@ const CreateStorageClusterWizard: React.FC<CreateStorageClusterWizardProps> = ({
         <Wizard
           className="ocs-install-wizard"
           steps={steps}
-          startAtStep={getStep()}
+          startAtStep={getStep(0, hasNoProvSC, isForceNavigated, setForceNavigated)}
           onBack={() => {
-            history.push(
-              `~new?${getParamString(getStep() - 1, getIndex(MODES, MODES.ATTACHED_DEVICES))}`,
-            );
+            history.push(getAnchor(getStep() - 1, getIndex(MODES, MODES.ATTACHED_DEVICES)));
           }}
           onNext={() => {
-            history.push(
-              `~new?${getParamString(getStep() + 1, getIndex(MODES, MODES.ATTACHED_DEVICES))}`,
-            );
+            history.push(getAnchor(getStep() + 1, getIndex(MODES, MODES.ATTACHED_DEVICES)));
           }}
           onClose={() =>
             history.push(resourcePathFromModel(ClusterServiceVersionModel, appName, ns))
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
@@ -23,8 +23,6 @@ import { OCS_ATTACHED_DEVICES_FLAG } from '@console/local-storage-operator-plugi
 import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager';
 import { resourcePathFromModel } from '@console/internal/components/utils';
 import { getName } from '@console/shared';
-import { useK8sGet } from '@console/internal/components/utils/k8s-get-hook';
-import { StorageClassModel } from '@console/internal/models';
 import { initialState, reducer } from './reducer';
 import {
   DiscoverDisks,
@@ -34,7 +32,7 @@ import { CreateStorageClass } from './install-wizard-steps/create-storage-class/
 import { StorageAndNodes } from './install-wizard-steps/storage-and-nodes-step';
 import { ReviewAndCreate } from './install-wizard-steps/review-and-create-step';
 import { Configure } from './install-wizard-steps/configure-step';
-import { taintNodes, filterSCWithNoProv } from '../../../utils/install';
+import { taintNodes } from '../../../utils/install';
 import {
   CreateStepsSC,
   MINIMUM_NODES,
@@ -123,25 +121,11 @@ const CreateStorageClusterWizard: React.FC<CreateStorageClusterWizardProps> = ({
   const [showInfoAlert, setShowInfoAlert] = React.useState(true);
   const [inProgress, setInProgress] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState('');
-  const [isForceNavigated, setForceNavigated] = React.useState(false);
   const flagDispatcher = useDispatch();
 
   const discoveryNodes = state.lvdIsSelectNodes ? state.lvdSelectNodes : state.lvdAllNodes;
 
   const { getStep, getIndex, getAnchor } = navUtils;
-
-  const [hasNoProvSC, setHasNoProvSC] = React.useState(false);
-
-  const [storageClasses, isLoaded, err] = useK8sGet(StorageClassModel);
-
-  React.useEffect(() => {
-    if (!err && isLoaded) {
-      const filteredSCData = storageClasses['items'].filter(filterSCWithNoProv);
-      if (filteredSCData.length) {
-        setHasNoProvSC(true);
-      }
-    }
-  }, [appName, err, isLoaded, ns, storageClasses]);
 
   /**
    * This custom footer for wizard provides a control over the movement to next step.
@@ -301,7 +285,7 @@ const CreateStorageClusterWizard: React.FC<CreateStorageClusterWizardProps> = ({
         <Wizard
           className="ocs-install-wizard"
           steps={steps}
-          startAtStep={getStep(0, hasNoProvSC, isForceNavigated, setForceNavigated)}
+          startAtStep={getStep()}
           onBack={() => {
             history.push(getAnchor(getStep() - 1, getIndex(MODES, MODES.ATTACHED_DEVICES)));
           }}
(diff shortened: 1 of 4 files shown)


frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
```diff
@@ -114,18 +114,20 @@ const CreateStorageClusterWizard: React.FC<CreateStorageClusterWizardProps> = ({
...
   navUtils,
+  hasNoProvSC,
 }) => {
   const { t } = useTranslation();
   const { appName, ns } = match.params;
   const [state, dispatch] = React.useReducer(reducer, initialState);
   const [showInfoAlert, setShowInfoAlert] = React.useState(true);
   const [inProgress, setInProgress] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState('');
+  const [isForceNavigated, setForceNavigated] = React.useState(false);
```
afreen23, line 125: Why do we need to maintain two states? May be I am missing something but I am thinking `noProv` would  be suffice.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install.tsx
```diff
@@ -59,6 +68,17 @@ export const CreateAttachedDevicesCluster: React.FC<CreateAttachedDevicesCluster
...
+    k8sList(StorageClassModel)
+      .then((storageClasses: StorageClassResourceKind[]) => {
+        const filteredSCData = storageClasses.filter(filterSCWithNoProv);
+        if (filteredSCData.length) {
+          setHasNoProvSC(true);
+        }
+      })
+      .catch(() => setHasNoProvSC(false));
+  }, [appName, ns]);
+
```
afreen23, line 81: Use `useK8sGet` and move this to the wizard component. We can directly use this in child component.

afreen23 (changes requested):

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx
rexagod: The page keeps refreshing when I only worked with `noProv`, and there was a need to record whether a page has already rendered or not in order to stop that behavior, hence this variable.

rexagod:
@afreen23 @bipuladh  I used `useK8sGet` in this way. The problem here is,

1. `items` does not exist on `k8sResourceCommon` (`storageClasses`) and on adding it to that type I'm getting a lot of conversion errors (even when adding it as an option).
2. `filterSCWithNoProv` expects `items` to be a `StorageClassResourceKind` type, that leads again to the aforementioned error. 

```ts
  const [storageClasses, isLoaded, err] = useK8sGet(StorageClassModel);

  React.useEffect(() => {
    if (!err && isLoaded) {
      const filteredSCData = storageClasses['items'].filter(filterSCWithNoProv);
      if (filteredSCData.length) {
        setHasNoProvSC(true);
      }
    }
  }, [appName, err, isLoaded, ns, storageClasses]);
```

afreen23:
>     1. `items` does not exist on `k8sResourceCommon` (`storageClasses`) and on adding it to that type I'm getting a lot of conversion errors (even when adding it as an option).
```js
const [storageClasses, isLoaded, err] = useK8sGet<ListKind<StorageClassKind>>(StorageClassModel);
```
>     2. `filterSCWithNoProv` expects `items` elements to be of `StorageClassResourceKind` type.
```js
 const scWithNoProv = sc?.items?.some(filterSCWithNoProv);
  setNoProvSc(scWithNoProv);
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,27 +90,37 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
+  };
+
   const getIndex = (searchSpace: any, search: string, offset: number = 0) => {
     const index = Object.values(searchSpace).findIndex((el) => el === search);
     return index - offset + 1;
   };
 
-  const getAnchor = (step: number, mode: number) => `~new?${getParamString(step, mode)}`;
-
   const handleModeChange = (event: React.FormEvent<HTMLInputElement>) => {
```
bipuladh, line 124: IMO this is the function that should take care of setting up the step correctly for Internal Attached Mode. Why can't we make changes here? In cases where `modeIndex` is 2 based on the presence of a no-prov sc change the parameters you pass to `getIndex` at line 118.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
rexagod: Actually ,`handleModeChange` is executed before `getStep`, since `getStep` is linked with the [Wizard component render](https://github.com/rexagod/console/blob/5a7193a28523fc10468e711a3c3bd25f07e70ff1/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard.tsx#L304), hence I used the `getStep` one.

bipuladh: My suggestion is that you setup the step and mode in the URL in the above function. And let `getStep` parse the URL and set the correct step.

bipuladh:
```
Actually ,handleModeChange is executed before getStep
```
This actually makes this a better approach as you can use `handleModeChange` to contain logic related to setting up unique logic for each mode.

rexagod: Seconded, that's a way better approach. On it.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -57,9 +62,17 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
+  const [sc, isLoaded, err] = useK8sGet<ListKind<StorageClassResourceKind>>(StorageClassModel);
   const memoizedCSV = useDeepCompareMemoize(csv, true);
 
+  React.useEffect(() => {
+    if (!err && isLoaded) {
+      const isSCWithNoProv = sc?.items?.some(filterSCWithNoProv);
+      setHasNoProvSC(isSCWithNoProv);
+    }
+  }, [appName, err, isLoaded, ns, sc]);
+
```
bipuladh, line 75: Don't see the need for `useEffect` nor the `hasNoProvSC` state.
Wouldn't this work? 
```
const hasNoProvSC = sc?.items?.some(filterSCWithNoProv);
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -124,9 +127,15 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
     const modeIndex = getIndex(MODES, value);
-    history.push(
-      `~new?${getParamString(getIndex(CreateStepsSC, CreateStepsSC.DISCOVER), modeIndex)}`,
-    );
+    switch (modeIndex) {
+      case 2:
+        if (hasNoProvSC) {
+          history.push(getAnchor(3, 2));
+        }
+        break;
```
bipuladh, line 135: Shouldn't this fall back to default?  Maybe the break should be part of the above `if` block?

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
rexagod: Yeah, I suppose this is extraneous now. Removing.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,40 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
     const modeIndex = getIndex(MODES, value);
-    history.push(
-      `~new?${getParamString(getIndex(CreateStepsSC, CreateStepsSC.DISCOVER), modeIndex)}`,
-    );
+    switch (modeIndex) {
+      case 2:
+        if (hasNoProvSC) {
+          history.push(getAnchor(3, 2));
+          break;
+        }
```
afreen23, line 129:
```suggestion
        if (hasNoProvSC) history.push(getAnchor(3, 2));
        break;
```

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,40 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
+      // eslint-disable-next-line no-fallthrough
```
afreen23, line 130:
```suggestion
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,40 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
 
-  const getAnchor = (step: number, mode: number) => `~new?${getParamString(step, mode)}`;
-
   const handleModeChange = (event: React.FormEvent<HTMLInputElement>) => {
     const { value } = event.currentTarget;
     const modeIndex = getIndex(MODES, value);
-    history.push(
-      `~new?${getParamString(getIndex(CreateStepsSC, CreateStepsSC.DISCOVER), modeIndex)}`,
-    );
+    switch (modeIndex) {
```
afreen23, line 124: Add a comment, since ts only for attached device mode.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
rexagod: In cases where `modeIndex` is 2 and `hasNoProvSC` is false, the control flow will encounter `break` and exit thus not falling back to the default condition, hence the basic functionality will be affected, no? Also, seeing that we need to fallback to a default condition, using `if`s here is a better choice since we can have as many cases as we want that fallback to a default condition. Making changes.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,39 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
   const handleModeChange = (event: React.FormEvent<HTMLInputElement>) => {
     const { value } = event.currentTarget;
     const modeIndex = getIndex(MODES, value);
-    history.push(
-      `~new?${getParamString(getIndex(CreateStepsSC, CreateStepsSC.DISCOVER), modeIndex)}`,
-    );
+    // `modeIndex` can have one of the following numerical values:
+    // 1: INTERNAL
+    // 2: ATTACHED_DEVICES
+    // 3: EXTERNAL
```
afreen23, line 127: I expect the comment should indicate that we are setting this for attached devices installation where the starting step has to be different with the presence of a no-prov storage class,

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,39 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
   const getParamString = (step: number, mode: number) => {
     const searchParams = new URLSearchParams(window.location.search.slice(1));
     searchParams.set('step', step.toString());
     searchParams.set('mode', mode.toString());
     return searchParams.toString();
   };
 
+  const getAnchor = (step: number, mode: number) => `~new?${getParamString(step, mode)}`;
+
+  const getStep = (offset: number = 0) => {
```
bipuladh, line 109:
```suggestion
  const getStep = (maxSteps: number = 5) => {
```
This should make usage of the function easier.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,39 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
+    const searchParams = new URLSearchParams(window.location.search.slice(1));
```
bipuladh, line 110:
```suggestion
    const searchParams = new URLSearchParams(window.location.search);
```
Do we really need to slice it? Are you trying to remove `?` ?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
```diff
@@ -90,33 +97,39 @@ const InstallCluster: React.FC<InstallClusterProps> = ({ match }) => {
...
+    const step = parseInt(searchParams.get('step'), 10) || 1;
+    const sanitizedStep = step && step <= 5 - offset && step >= 1 ? step : 1;
```
bipuladh, line 112:
```suggestion
    const sanitizedStep = step <= maxSteps && step >= 1 ? step : 1;
```
Step is guaranteed to be a integer.

bipuladh (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
rexagod: I was, but I see it works fine without this as well. Thank you for spotting this!

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-page.tsx
rexagod: Noted. Making changes.

bipuladh (approved):

afreen23 (approved):
/lgtm

PR #9081, opened 2021-05-31
show helptext for workspaces in start pipeline form

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5861

**Root analysis:**
Description of workspaces is not shown in the form

**Solution description:**
passed the value of workspace description in the `helptext` prop

**Screenshot:**
Screenshot from 2021-05-31 16-36-10
start-pipeline

**Browser conformance:**
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineWorkspacesSection.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineWorkspacesSection.tsx
@@ -108,6 +108,7 @@ const PipelineWorkspacesSection: React.FC = () => {
                 }
                 fullWidth
                 required={!workspace.optional}
+                helpText={workspace.description}
               />
               {getVolumeTypeFields(workspace.type, index)}
             </div>


frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineWorkspacesSection.tsx
```diff
@@ -108,6 +108,7 @@ const PipelineWorkspacesSection: React.FC = () => {
                 }
                 fullWidth
                 required={!workspace.optional}
+                helpText={workspace.description}
```
karthikjeeyar, line 111: Translation is missing here.

andrewballantyne: Dynamic Translations are not a thing we are currently are supporting. No way to take user-data and translate it on the fly without a 3rd party system which we do not have.

andrewballantyne (approved):
Looks good 👍

PR #9084, opened 2021-05-31
Allow installation storageClass as default in the add capacity modal

SanjalKatiyar:
URL: https://issues.redhat.com/browse/RHSTOR-1406
**After the fix:**
1) The storage class used during the installation is pre-selected in add-capacity modal.
2) Dropped the usage of OCSStorageClassDropdown component.
3) Added the use of StorageClassDropdown instead.


frontend/packages/ceph-storage-plugin/src/components/modals/storage-class-dropdown.tsx
```diff
@@ -11,16 +11,16 @@ const ocsProvisioners = [
   'rbd.csi.ceph.com',
   'noobaa.io/obc',
 ];
+export const filterSC = (sc: StorageClassResourceKind) =>
+  ocsProvisioners.every((provisioner: string) => !sc?.provisioner?.includes(provisioner));
```
afreen23, line 15: move this to utils.
We can get rid of this file itself.

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -46,6 +47,13 @@ import { checkArbiterCluster, checkFlexibleScaling } from '../../../utils/common
 const queries = (() => Object.values(CAPACITY_INFO_QUERIES))();
 const parser = compose((val) => val?.[0]?.y, getInstantVectorStats);
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
+const getSelectedKey = (
+  storageClass: StorageClassResourceKind,
+  deviceSets: DeviceSet[],
+): string => {
+  if (storageClass) return getName(storageClass);
+  return deviceSets?.[0]?.dataPVCTemplate?.spec?.storageClassName;
```
afreen23, line 55: If there are more than one device sets, then we should show the last device set's storage class.

frontend/packages/ceph-storage-plugin/src/components/modals/storage-class-dropdown.tsx
afreen23, line 15:
```suggestion
  !ocsProvisioners.some((ocsProvisioner: string) => sc?.provisioner?.includes(ocsProvisioner));
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
afreen23, line 55:
```suggestion
  return deviceSets?.[deviceSets.length-1]?.dataPVCTemplate?.spec?.storageClassName;
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -34,18 +35,25 @@ import {
...
 
 const queries = (() => Object.values(CAPACITY_INFO_QUERIES))();
 const parser = compose((val) => val?.[0]?.y, getInstantVectorStats);
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
+const getSelectedKey = (
+  storageClass: StorageClassResourceKind,
+  deviceSets: DeviceSet[],
+): string => {
+  if (storageClass) return getName(storageClass);
+  return deviceSets?.[deviceSets.length - 1]?.dataPVCTemplate?.spec?.storageClassName;
```
afreen23, line 55: Please add a comment for this behaviour.

afreen23 (changes requested):
nits:

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
SanjalKatiyar: ack...
but, there is one thing though:
If I am using same storageClass for adding capacity twice, then, that storageClass will not necessarily be always at  `deviceSets[deviceSets.length - 1]` index...
Example:
Initial deployment using `gp2`
Expand for the first time using `gp2-csi`
Expand for the second time using `gp2`
In the above case `gp2` will still be at 0th index of the deviceSets and when we will Expand for the third time `gp2-csi` will be the pre-selected value as it is present at (length-1)th index...

afreen23: Correct, we can only show the last device set index or first device set index.
In practice, the use cases have been around using another storage class when a user wants to use new storage devices. 
But the possibilities and use-cases are beyond our control and let's keep it close to generic so let's revert this to the first storage device set only and add a proper comment.
That is, always show the installation storage class.

afreen23 (changes requested):
remove TBD comment on L73

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -34,18 +35,30 @@ import {
...
+ * By default, returns the storageClass used for initial installation.
+ * In case of multiple deviceSets, returns the last deviceSets's storageClass.
+ * Otherwise, returns the name of the storageClass selected from the dropdown.
+ */
+const getSelectedKey = (
+  storageClass: StorageClassResourceKind,
+  deviceSets: DeviceSet[],
+): string => {
+  if (storageClass) return getName(storageClass);
+  return deviceSets?.[deviceSets.length - 1]?.dataPVCTemplate?.spec?.storageClassName;
```
afreen23, line 60: There is already - `selectedSCName` we are calculating above, lets reuse that.
```suggestion
  return selectedSCName || deviceSets?.[0]?.dataPVCTemplate?.spec?.storageClassName;
```

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -57,9 +64,6 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
   const [pvData, pvLoaded, pvLoadError] = useK8sWatchResource<K8sResourceKind[]>(pvResource);
   const [nodesData] = useK8sWatchResource<NodeKind[]>(nodeResource);
   const [storageClass, setStorageClass] = React.useState<StorageClassResourceKind>(null);
-  /* TBD(Afreen): Show installation storage class as preselected
-                  Change state metadata
-  */
 
   const [inProgress, setProgress] = React.useState(false);
   const [errorMessage, setError] = React.useState('');
```
afreen23, line 69: nit:
```suggestion
  const [inProgress, setProgress] = React.useState(false);
  const [errorMessage, setError] = React.useState('');
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -34,18 +35,24 @@ import {
...
 
 const queries = (() => Object.values(CAPACITY_INFO_QUERIES))();
 const parser = compose((val) => val?.[0]?.y, getInstantVectorStats);
 const getProvisionedCapacity = (value: number) => (value % 1 ? (value * 3).toFixed(2) : value * 3);
+/**
+ * Returns the name of the storageClass selected from the items of the dropdown.
+ * Otherwise, returns the installation storageClass which will be the pre-selected value for the dropdown.
+ */
+const getSelectedKey = (selectedSCName: string, deviceSets: DeviceSet[]): string =>
+  selectedSCName || deviceSets?.[0]?.dataPVCTemplate?.spec?.storageClassName;
```
afreen23, line 55:
```suggestion
const installStorageClass = deviceSets?.[0]?.dataPVCTemplate?.spec?.storageClassName;
```

frontend/packages/ceph-storage-plugin/src/components/modals/add-capacity-modal/add-capacity-modal.tsx
```diff
@@ -190,7 +192,14 @@ export const AddCapacityModal = (props: AddCapacityModalProps) => {
           isRequired
         >
           <div id="add-capacity-dropdown" className="ceph-add-capacity__sc-dropdown">
-            <OCSStorageClassDropdown onChange={onChange} data-test="add-cap-sc-dropdown" />
+            <StorageClassDropdown
+              onChange={(sc: StorageClassResourceKind) => setStorageClass(sc)}
+              noSelection
+              selectedKey={getSelectedKey(selectedSCName, deviceSets)}
```
afreen23, line 198:
```suggestion
              selectedKey={selectedScName || installStorageClass}
```

afreen23 (changes requested):

afreen23 (approved):
/lgtm

PR #9085, opened 2021-05-31
Bug 1966275: Fix Pipeline Parameters in Modals accept empty string defaults

andrewballantyne:
Slimmed down version of: https://github.com/openshift/console/pull/9074
(the functionality without the proper moving of the utility & type differences between the two releases)

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineParameterSection.tsx
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineParameterSection.tsx
@@ -2,16 +2,18 @@ import * as React from 'react';
 import { FieldArray } from 'formik';
 import { useTranslation } from 'react-i18next';
 import { TextInputTypes } from '@patternfly/react-core';
-import { InputField } from '@console/shared';
-import { PipelineParam } from '../../../../utils/pipeline-augment';
+import { InputField, useFormikValidationFix } from '@console/shared';
 import FormSection from '@console/dev-console/src/components/import/section/FormSection';
+import { taskParamIsRequired } from '../../pipeline-builder/utils';
+import { ModalParameter } from './types';
 
 type ParametersSectionProps = {
-  parameters: PipelineParam[];
+  parameters: ModalParameter[];
 };
 
 const PipelineParameterSection: React.FC<ParametersSectionProps> = ({ parameters }) => {
   const { t } = useTranslation();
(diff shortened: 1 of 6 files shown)


karthikjeeyar:
In 4.7, While starting the pipeline with no default param, it opens the start modal with start button enabled even if the user haven't entered any value. If user hasn't entered anything in required field, then start button should be disabled.

andrewballantyne:
> In 4.7, While starting the pipeline with no default param, it opens the start modal with start button enabled even if the user haven't entered any value. If user hasn't entered anything in required field, then start button should be disabled.

It seems to do nothing with my code (happens on 4.7 without my PR + with a more mirrored version of the 4.8 code), but I am unsure what actually is fixing this in 4.8. The `yup` errors are not catching `value` of `undefined` until you touch the param. 🤔 

I think we are going to have to go with what this is. The issue predates my work here and isn't exposed anymore or less by my changes so I don't think it's worth anymore time to try and figure this out. If you have a better idea of what could be causing it @karthikjeeyar - I'd be happy to investigate more.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineParameterSection.tsx
```diff
@@ -3,11 +3,12 @@ import { FieldArray } from 'formik';
...
 import FormSection from '@console/dev-console/src/components/import/section/FormSection';
+import { taskParamIsRequired } from '../../pipeline-builder/utils';
+import { ModalParameter } from './types';
 
 type ParametersSectionProps = {
-  parameters: PipelineParam[];
+  parameters: ModalParameter[];
 };
 
 const PipelineParameterSection: React.FC<ParametersSectionProps> = ({ parameters }) => {
```
karthikjeeyar, line 14: Initial form validation issue is resolved If we introduce this validation hook here.
```suggestion
const PipelineParameterSection: React.FC<ParametersSectionProps> = ({ parameters }) => {
useFormikValidationFix(parameters);

```

andrewballantyne: Ah forgot about that hook! Thanks! Will do that.

andrewballantyne:
Thanks for the suggestion @karthikjeeyar - Completely forgot about that hook for Formik. It probably got implemented somewhere in this form for another reason in 4.8 and thus was solving the issues for the whole modal.

PR #9095, opened 2021-06-01
Bug 1904155: Make utilization chart domain match duration

TheRealJon:
Update all utilization cards so that the chart x domains will match the currently selected duration.

### Before
screencast-localhost_9000-2021 06 01-10_27_19

### After
screencast-localhost_9000-2021 06 01-10_09_08


TheRealJon:
Closing. Opened from the wrong branch.

PR #9097, opened 2021-06-01
Bug 1966520: Add button from ocs add capacity should not be enabled if there are no PV's

SanjalKatiyar:
Disabled the "add" button if no PVs are available.
Screenshot from 2021-06-01 20-12-33

Other add-capacity-modal validations were incorporated in https://github.com/openshift/console/pull/8754


afreen23 (approved):
/lgtm

PR #9102, opened 2021-06-01
Bug 1964902: rework network policy details page

jotak:
- Fix messed up selectors with respect to peers per rule, by showing
  each line as a peer and not as a rule
- Fix selector links for cluster-wide rules
- Show IP Block
- Fix text displayed when ingress/egress are non empty, but contain
  empty peer


jotak:
Here's a screenshot:

Capture d’écran de 2021-06-01 17-20-23

Corresponding policy:

```yaml
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: complex-policy
spec:
  podSelector:
    matchLabels:
      app: httpd
  ingress:
    - ports:
        - protocol: TCP
          port: 8080
      from:
        - podSelector:
            matchLabels:
              aa: bb
        - podSelector:
            matchLabels:
              aa: cc
          namespaceSelector:
            matchLabels:
              project: foo
        - podSelector: {}
    - from:
        - ipBlock:
            cidr: 0.0.0.0/0
            except:
              - 10.0.0.0/16
    - ports:
        - protocol: UDP
          port: 9000
  policyTypes:
    - Ingress
```

jotak:
cc @yapei 
Some comments: I didn't want to change too many things in that page, but tried to limit to fixing spotted issues strictly speaking.
However, perhaps the layout can be improved. The first column "Target pods" shows redundant information since it's the policy's pod selector, it's always going to be repeated ; and is also displayed higher in the page.

So we could remove that column, and instead, split the peer's "podSelector", "namespaceSelector" and IP block in three different columns.

Or alternatively, we could have some kind of quick description as a first column, something like "Some pods in the same namespace", "All pods in some namespaces", etc. Perhaps that would help users to quickly scope the rule.

frontend/public/components/network-policy.tsx
```diff
@@ -135,60 +135,82 @@ const EgressHeader = () => {
...
+                        selector={peer.podSelector}
+                        namespace={peer.namespaceSelector ? undefined : namespace}
+                      />
+                    )}
+                  </div>
+                </div>
+              )}
+              {peer.ipBlock && (
+                <div>
+                  <span className="text-muted">{t('public~IP Block')}:</span>
```
spadgett, line 192: We should use sentence case for labels

```suggestion
                  <span className="text-muted">{t('public~IP block')}:</span>
```

spadgett (commented):
lgtm, just one small comment on the label case

PR #9107, opened 2021-06-01
Bug 1942125: add i18n for serverless breadcrumbs

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/OCPBUGSM-26858

**Root analysis:**
i18n translations missing in the details pages breadcrumbs of Serverless resources

**Solution description:**
- added translations for the breadcrumbs
- removed the spaces in CRD Kind labels

**GIF:**
Peek 2021-06-01 22-54

**Browser conformance:**
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


andrewballantyne (changes requested):
We'll want to drop the spaces in CRD Kind labels (both singular and plural).

Examples:
- [See a previous conversation with Admin](https://github.com/openshift/console/pull/8628#issuecomment-818833110)
- [Pipelines version of this PR](https://github.com/openshift/console/pull/8644)
- [Related Slack Thread](https://coreos.slack.com/archives/C0176H51F8V/p1618412328043000)

frontend/packages/knative-plugin/src/models.ts
```diff
@@ -32,8 +32,8 @@ export const KnativeServingModel: K8sKind = {
   apiGroup: 'operator.knative.dev',
   apiVersion: 'v1alpha1',
   kind: 'KnativeServing',
-  label: 'Knative Serving',
-  labelPlural: 'Knative Servings',
+  label: 'KnativeServing',
+  labelPlural: 'KnativeServings',
```
andrewballantyne, line 36: No `label...Key`?

frontend/packages/knative-plugin/src/models.ts
```diff
@@ -46,8 +46,8 @@ export const KnativeEventingModel: K8sKind = {
   apiGroup: 'operator.knative.dev',
   apiVersion: 'v1alpha1',
   kind: 'KnativeEventing',
-  label: 'Knative Eventing',
-  labelPlural: 'Knative Eventings',
+  label: 'KnativeEventing',
+  labelPlural: 'KnativeEventings',
```
andrewballantyne, line 50: Same here

frontend/packages/knative-plugin/src/utils/fetch-dynamic-eventsources-utils.ts
```diff
@@ -52,8 +51,8 @@ export const fetchEventSourcesCrd = async () => {
             kind,
             plural,
             id: singular,
-            label,
-            labelPlural: `${label}s`,
+            label: kind,
+            labelPlural: `${kind}s`,
```
andrewballantyne, line 55: Hmm... there isn't much we can do here I don't think. But I also don't think slapping `s` on the end of words is good English syntax. For instance `...Class` cannot get `s` as a plural state.

I know this isn't this PR work @debsmita1 -- just making the comment because I see it @invincibleJai FYI.

andrewballantyne (changes requested):
Hmmmm... We have 25 models (cheap search for `label` in this file) and you only have 17 instances of `labelKey`. Is this a problem? Any reason we don't do all that are available?

frontend/packages/knative-plugin/src/utils/fetch-dynamic-eventsources-utils.ts
invincibleJai: @debsmita1 I think for translation we should follow this pattern i.e add `labelKey` and `labelPluralKey` and use the same where applicable for translations. For plural, I don't think we are saving much let's type the whole string.

https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/models/pipelines.ts#L7-L23

@andrewballantyne let us know if the above make sense

frontend/packages/knative-plugin/src/models.ts
invincibleJai: @debsmita1 I think for translation we should follow this pattern i.e add `labelKey` and `labelPluralKey` and use the same where applicable for translations. 

https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/models/pipelines.ts#L7-L23

andrewballantyne (approved):

invincibleJai:
Thanks @debsmita1 , have verified in all scenarios. In kamelets flow am seeing `[[Ḱααṃḛḛḽḛḛţβḭḭṇḍḭḭṇḡ] ḍḛḛţααḭḭḽṡ]` in pseudo translations, is this expected?

image

debsmita1:
> Thanks @debsmita1 , have verified in all scenarios. In kamelets flow am seeing `[[Ḱααṃḛḛḽḛḛţβḭḭṇḍḭḭṇḡ] ḍḛḛţααḭḭḽṡ]` in pseudo translations, is this expected?
> 
> image

Yes, this is expected because we are now translating the lables in the static models. I see a similar thing for pipelines

Screenshot from 2021-06-04 14-33-43


> and from `topology -> source` flow should `PingSources` be as well under pseudo translations
> 
> image

Here, the `PIngSources` is not translated because right now there are now 3/4 versions of PingSources available, but we have a static model defined for only one of the versions. So the translation could be seen only for that particular supported version which is `v1beta1`

PR #9116, opened 2021-06-02
handles check for kafkasource update with managed kafka

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5902

**Analysis / Root cause**: 
on updating `kafkaSource` with drag and drop to an invalid managed kafka connection resource (by invalid mean does not have `bootstrapServerHost` or secrets) form topology the connectors were not updating neither any err/info was shown to the user.

**Solution Description**: 
while updating `kafkaSource` with drag and drop to a managed kafka connection resource then added a check to validate if `KC` has `bootstrapServerHost` and `secrets`. In case of invalid show appropriate message to the user and don't update `kafkaSource` otherwise update it.

image

Kapture 2021-06-02 at 18 02 29

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/locales/en/knative-plugin.json
+++ b/frontend/packages/knative-plugin/locales/en/knative-plugin.json
@@ -194,5 +194,6 @@
   "Move sink to {{resourceObjKind}}": "Move sink to {{resourceObjKind}}",
   "Move sink to URI": "Move sink to URI",
   "Open URI": "Open URI",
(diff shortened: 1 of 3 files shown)


bgliwa01:
If we are using secrets as plural I think the verb needs to be are

invincibleJai:
> If we are using secrets as plural I think the verb needs to be are

thanks, it'll be singular i.e secret. Have updated it.

vikram-raj (commented):
verified the changes and it works as expected.

PR #9118, opened 2021-06-02
Bug 1967108: Use isEqual when comparing async loaders

rawagner:
Temporary workaround.
For proper solution we will want to use React.lazy or just compare referential equality but we need to go over all `AsyncComponent` usages and make sure the references remain the same during rerenders

### DIFF
--- a/frontend/public/components/utils/async.tsx
+++ b/frontend/public/components/utils/async.tsx
@@ -3,13 +3,6 @@ import * as _ from 'lodash-es';
 
 import { LoadingBox } from './status-box';
 
-/**
- * FIXME: Comparing two functions is not the *best* solution, but we can handle false negatives.
- */
-const sameLoader = (a: () => Promise<React.ComponentType>) => (
-  b: () => Promise<React.ComponentType>,
-) => (a || 'a').toString() === (b || 'b').toString();
-
 enum AsyncComponentError {
   ComponentNotFound = 'COMPONENT_NOT_FOUND',
 }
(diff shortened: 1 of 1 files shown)


frontend/public/components/utils/async.tsx
```diff
@@ -23,7 +16,7 @@ export class AsyncComponent extends React.Component<AsyncComponentProps, AsyncCo
   private isAsyncMounted: boolean = false;
 
   static getDerivedStateFromProps(props, state) {
-    if (!sameLoader(props.loader)(state.loader)) {
+    if (!_.isEqual(props.loader, state.loader)) {
```
spadgett, line 19: It looks like this uses strict equality for functions if I'm reading the lodash docs right. Is this the same as using `===` in that case?

https://lodash.com/docs/4.17.15#isEqual

jhadvig: @spadgett yes, but in this case you are comparing the functions not strings, right ?

rawagner: ooh, right :( 

Well in that case I dont have a simple solution for the issue

rawagner:
Well, Im closing this for now :)

frontend/public/components/utils/async.tsx
spadgett: > @spadgett yes, but in this case you are comparing the functions not strings, right ?

Yeah, but if it's an inline function, it will always evaluate as changed and reload the component.

PR #9119, opened 2021-06-02
Bug 1967197: Fix 404 errors loading i18n namespaces

spadgett:
The `k8s` and `oauth` namespaces have been removed, so don't try to load
them. The `git-service` namespace was not being copied to the locales
directory.

/assign @rebeccaalpert @jhadvig


rebeccaalpert (approved):
Looks good to me!

PR #9121, opened 2021-06-02
Bug 1967207: Update "Getting started resources" links

spadgett:
Show "OperatorHub" in place of "Console customizations" in the Administrator perspective "Getting started resources" card.

### DIFF
--- a/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/getting-started/__tests__/explore-admin-features-getting-started-card.spec.tsx
+++ b/frontend/public/components/dashboard/dashboards-page/cluster-dashboard/getting-started/__tests__/explore-admin-features-getting-started-card.spec.tsx
@@ -48,9 +48,9 @@ describe('ExploreAdminFeaturesGettingStartedCard', () => {
         href: '/api-explorer',
       },
       {
-        id: 'console-customizations',
-        title: 'Console customizations',
-        href: '/k8s/cluster/customresourcedefinitions?name=console',
+        id: 'operatorhub',
+        title: 'OperatorHub',
+        href: '/operatorhub',
       },
     ]);
     expect(wrapper.find(GettingStartedCard).props().moreLink).toEqual({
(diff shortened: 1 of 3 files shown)


jerolimov:
@spadgett Switching to API explorer is cool :+1:

But the filter doesn't include the Console / operator.openshift.io resource anymore. And this contains the customization options for the devconsole. Hmm

As alternative we could link to `/settings/cluster/globalconfig`. I'm not sure what's most useful for a cluster admin.

spadgett:
Ideally we'd have a UI for console customizations. But we're not there yet.

I don't think we should link to a page with several unrelated resources. That's pretty confusing to me. And the number of bad matches will grow over time using a fuzzy name filter.

If we're primarily concerned with the console operator config, it might be better to link to that resource specifically, or have two links (one for the console CRDs and one to the operator config).

gdoyle1:
> Ideally we'd have a UI for console customizations. But we're not there yet.
> 
> I don't think we should link to a page with several unrelated resources. That's pretty confusing to me. And the number of bad matches will grow over time using a fuzzy name filter.
> 
> If we're primarily concerned with the console operator config, it might be better to link to that resource specifically, or have two links (one for the console CRDs and one to the operator config).

to your point @spadgett - can we add a hint at the top of the page **(only if they get here from the getting started card)** giving users a little more info? Really crappy wireframe but this is where we could place it

On cluster overview page 1

jerolimov:
Then we could also "hack" that the list shows Console / operator.openshift.io CRD when filtering for console.openshift.io. But I believe we don't want such hacks here, or Sam? :)

Ginas first idea was to filter all CRDs with a label `customization=true`, but we don't have any CRD which has this label - yet. Maybe we should rethink if this is a good idea and add this labels to some of (our) resources?

(But the api explorer also doesn't support a label filter and for me its really nice that it shows the Description column. Compared to the CRD search.)

jerolimov:
Let's integrate this switch to API Explorer and check offline (in slack) if we can improve this with another PR.

/lgtm
/retest

jerolimov (approved):
Works fine. Tests pass locally.

/lgtm

gdoyle1 (approved):
LGTM

PR #9129, opened 2021-06-03
test: test pseudolocalization of all pages in CRUD tests

spadgett:
In the CRUD tests, we had a number of pages that were skipping
pseudolocalization checks since they weren't translated yet. This is no
longer the case, so test i18n on all pages.

Rename k8s-openshift-cruds.spec.ts -> resource-crud.spec.ts

Remove references to the service catalog pages since service catalog has
been removed from OpenShift.

/assign @dtaylor113


dtaylor113:
What about pages under Compute such as Nodes and MachineSets?

dtaylor113 (approved):

PR #9130, opened 2021-06-03
Explicitly touch pipeline builder task fields when workspaces are changed

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5894

**Analysis / Root cause**: 
Formik fields do not echo errors unless they are touched.

**Solution Description**: 
Use a hook to explicitly touch all task fields so they echo errors when dependent fields are changed.

https://user-images.githubusercontent.com/20013884/120647396-30ef6300-c498-11eb-92c5-ed0190c8963a.mp4

(unchanged)

**Test setup:**
- Edit a pipeline that has tasks which use workspaces or resources.
- Update the list of workspaces/resources.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
+++ b/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import { getRandomChars } from '@console/shared';
-import { useFormikContext } from 'formik';
+import { useFormikContext, FormikTouched } from 'formik';
 import { referenceForModel } from '@console/internal/module/k8s';
 import { useK8sWatchResources } from '@console/internal/components/utils/k8s-watch-hook';
 import { ClusterTaskModel, TaskModel } from '../../../models';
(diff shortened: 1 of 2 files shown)


karthikjeeyar:
@rottencandy  Even without this change  I see the errors showing up correctly in the task sidbebar, I have verified it in the cluster latest build 4.8.0-0.nightly-2021-06-06-164529. cc: @jerolimov @andrewballantyne

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
```diff
@@ -262,3 +262,29 @@ export const useNodes = (
...
+
+const touchTaskFields = (task: PipelineTask): FormikTouched<PipelineTask> => ({
+  workspaces: task.workspaces?.map(() => ({ workspace: true })),
+  resources: {
+    inputs: task.resources?.inputs?.map(() => ({ resource: true })),
+    outputs: task.resources?.outputs?.map(() => ({ resource: true })),
+  },
+});
+
+export const useExplicitFormikTouch = () => {
```
karthikjeeyar, line 274: nit: This name `useExplicitFormikTouch` is generic but the implementation seems its tied to `PipelineTask`'s workspace and resources, so I am thinking if we could change it to something that is builder specific.

frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/hooks.ts
```diff
@@ -262,3 +262,29 @@ export const useNodes = (
...
+  React.useEffect(() => {
+    if (workspacesTouched || resourcesTouched) {
+      setTouched({
+        formData: {
+          tasks: values.formData?.tasks?.map(touchTaskFields),
+          finallyTasks: values.formData?.finallyTasks?.map(touchTaskFields),
+        },
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jerolimov, line 288: First of all: I tested it locally and it works fine to get an error message when removing a used (untouched) workspace :+1:

Reducing code is welcome in most cases. But in this case I don't see a reason to touch all resources when a workspace is touched and vice versa. Wdyt?

jerolimov: I'm also fine with this change if Andrew approve it. It fixes the issue and have no "big disadvantage" :)

rottencandy: Thanks, updated to only touch task workspaces if workspaces list is changed, same for resources.

andrewballantyne (approved):

jerolimov (approved):
Tested the changed version again. Works fine.
/lgtm

PR #9138, opened 2021-06-03
Bug 1967713: Show link to the OCM landing page when cluster ID is missing

gkarat:
The Insights widget referred to the wrong Cluster Details page when the external cluster-ID was unavailable for some reason.

The patch replaces that link to redirect the OCM landing page.

### DIFF
--- a/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
+++ b/frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
@@ -98,7 +98,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
         )}
       </div>
       <div className="co-status-popup__section">
-        {!isWaitingOrDisabled && !isError && (
+        {!isWaitingOrDisabled && !isError && clusterID && (
           <>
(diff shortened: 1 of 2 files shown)


frontend/packages/insights-plugin/src/components/InsightsPopup/index.tsx
```diff
@@ -98,7 +98,7 @@ export const InsightsPopup: React.FC<PrometheusHealthPopupProps> = ({ responses,
         )}
       </div>
       <div className="co-status-popup__section">
-        {!isWaitingOrDisabled && !isError && (
+        {!isWaitingOrDisabled && !isError && !clusterID && (
```
spadgett, line 101: Is the logic backwards here? It looks like you're checking `!clusterID` and then using it in the content.

gkarat: @spadgett, sorry, end of the working day, left it out. Fixed :)

spadgett (approved):
/lgtm

gkarat:
@spadgett, something is certainly wrong with the e2e test. Any chance you could take a look at them? The problem seems to be not connected to my PR. I've marked this PR with WIP label to stop automatic tests.

PR #9139, opened 2021-06-03
Bug 1967717: Rename Insights to Insights Advisor and add missing paragraph

gkarat:
The patch changes naming from Insights to Insights Advisor in several places of the related widget. Furthermore, it appeared we are missing one important paragraph there. The patch also adds it.

Before:
Screenshot from 2021-06-03 18-51-34
After:
Screenshot from 2021-06-03 18-50-57

### DIFF
--- a/frontend/packages/insights-plugin/locales/en/insights-plugin.json
+++ b/frontend/packages/insights-plugin/locales/en/insights-plugin.json
@@ -3,6 +3,7 @@
   "moderate": "moderate",
   "important": "important",
   "critical": "critical",
+  "Insights Advisor identifies and prioritizes risks to security, performance, availability, and stability of your clusters.": "Insights Advisor identifies and prioritizes risks to security, performance, availability, and stability of your clusters.",
   "Temporary unavailable.": "Temporary unavailable.",
   "Disabled or waiting for results.": "Disabled or waiting for results.",
   "Total issue": "Total issue",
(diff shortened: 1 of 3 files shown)


spadgett (approved):
/lgtm

spadgett:
/lgtm cancel

> Error: Failed: Missing i18n key "Insights Advisor status" in namespace "insights-plugin" and language "en."

gkarat:
@spadgett, I've added the missing key to locals. Thanks for point it out

frontend/packages/insights-plugin/src/plugin.tsx
```diff
@@ -28,7 +28,7 @@ const plugin: Plugin<ConsumedExtensions> = [
           (m) => m.InsightsPopup,
         ),
       // t('insights-plugin~Insights status')
```
spadgett, line 30: This comment needs to be updated as well so that the i18next parser uses the right string.

frontend/packages/insights-plugin/locales/en/insights-plugin.json
```diff
@@ -15,5 +16,5 @@
   "{{issuesNumber}} issue found": "{{issuesNumber}} issue found",
   "{{issuesNumber}} issues found": "{{issuesNumber}} issues found",
   "Insights": "Insights",
-  "Insights status": "Insights status"
-}
\ No newline at end of file
+  "Insights Advisor status": "Insights Advisor status"
```
spadgett, line 19: You shouldn't modify this file directly. Let the i18next parser do it using the right comment format

gkarat: Thanks, fixed it with the last push :+1:

spadgett (approved):
/lgtm

PR #9068, opened 2021-05-27
Fix keyboard scroll issue of namespaced pages

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5514

**Analysis / Root cause**: The `NamespacedPage` component that is used by all dev console pages had `height: 100%` which caused issues with scroll.

**Solution Description**: Removed `height: 100%` and add `flex: 1`.

**Screen shots / Gifs for design review**: 

https://user-images.githubusercontent.com/6041994/119853432-d3a26180-bf2d-11eb-832b-080d0132aeb9.mov

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

PR #9140, opened 2021-06-03
Remove tabIndex from scrollable containers to fix keyboard scroll issue

rohitkrai03:
**Analysis / Root cause**:
- To fix the keyboard scroll issue https://github.com/openshift/console/pull/9068 changed the layout to flex and removed `max-height` from the container. 
- But this caused another issue where cypress test started failing because `scrollTo` wouldn't work anymore.
- Turns out it was an old issue because of PF setting `tabIndex=-1` on `Page` component mentioned in https://github.com/patternfly/patternfly-react/issues/4180. This caused all the child containers to explicitly set `tabIndex={-1}` in order to enable keyboard scrolling. 

**Solution Description**: 
- Revert back the changes from https://github.com/openshift/console/pull/9068.
- Remove explicit `tabIndex={-1}` from scrollable container.
- Pass explicit `mainTabIndex={null}` to `Page` component.

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/public/components/app.jsx
+++ b/frontend/public/components/app.jsx
@@ -163,6 +163,8 @@ class App_ extends React.PureComponent {
           <div id="app-content" className="co-m-app__content">
             <ConsoleNotifier location="BannerTop" />
             <Page
+              // Need to pass mainTabIndex=null to enable keyboard scrolling as default tabIndex is set to -1 by patternfly
+              mainTabIndex={null}
               header={<Masthead onNavToggle={this._onNavToggle} />}
               sidebar={
                 <Navigation
--- a/frontend/packages/dev-console/src/components/NamespacedPage.scss
+++ b/frontend/packages/dev-console/src/components/NamespacedPage.scss
@@ -1,7 +1,8 @@
 .odc-namespaced-page {
   display: flex;
-  flex: 1;
   flex-direction: column;
+  height: 100%;
+  max-height: 100%;
 
   &__content {
     display: flex;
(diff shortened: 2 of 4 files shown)


christianvogt:
I debugged this issue with Rohit and agree that the tabindex should be removed. Having a tabindex on a container essentially breaks focus and keyboard scroll for all non-focusable containers below it. eg topology details side panel.

Any contributions to the main content should not have to deal with adding their own tabindex attribute to their scroll containers.

frontend/public/components/app.jsx
```diff
@@ -163,6 +163,8 @@ class App_ extends React.PureComponent {
           <div id="app-content" className="co-m-app__content">
             <ConsoleNotifier location="BannerTop" />
             <Page
+              // Need to pass mainTabIndex=null to enable keyboard scrolling as default tabIndex is set to -1 by patternfly
```
spadgett, line 166: @jcaianirh Do you know why PatternFly is defaulting to tabIndex -1?

frontend/public/components/app.jsx
```diff
@@ -163,6 +163,8 @@ class App_ extends React.PureComponent {
...
+              mainTabIndex={null}
```
rhamilto, line 167: With this addition, we should also be able to remove the `tabIndex` on the sidebar. 
 See https://github.com/openshift/console/blob/master/frontend/public/components/sidebars/resource-sidebar.tsx#L25-L26

rohitkrai03: Done.

PR #8720, opened 2021-04-20
CONSOLE-2527: Sprint 198 translations

rebeccaalpert:
Added latest round of translation files. I went through and updated some issues, detailed below by file.

I'll bug the translation team about the translation stuff. I fixed the easy buggy ones locally. We need the team to update Memsource or our changes will keep getting overwritten.

#### Japanese:
console-app.json 
* "VolumeSnapshot": "VolumeSnapshot", but "VolumeSnapshots": "ボリュームスナップショット",

volume-snapshot-class-details.json
* "VolumeSnapshotClass details": "ボリュームスナップショットの詳細", - VolumeSnapshotClass is left untranslated in console-app.json; this appears to be the VolumeSnapshot translation rather than VolumeSnapshotClass

frontend/packages/console-app/locales/ja/volume-snapshot-content.json
* "VolumeSnapshotContent details": "ボリュームスナップショットコンテンツの詳細", but it’s "VolumeSnapshotContent": "VolumeSnapshotContent", in console-app.json and  "VolumeSnapshotContent": "VolumeSnapshotContent", in same file

olm.json
* "InstallPlan failed": "InstalPlan に失敗しました", (We have brought this up at least twice now and it's still not fixed)

#### Korean:
console-app.json
* "VolumeSnapshot details": "볼륨 스냅샷 세부 정보", but  "VolumeSnapshot": "VolumeSnapshot",

volume-snapshot-class-details.json
*  "VolumeSnapshotClass details": "볼륨 스냅샷 클래스 세부 정보", but "VolumeSnapshotClass": "VolumeSnapshotClass", in console-app.json and volume-snapshot-details.json

volume-snapshot.json
* "VolumeSnapshotClass": "볼륨 스냅샷 클래스", but  "VolumeSnapshotClass": "VolumeSnapshotClass" in volume-snapshot-class-details.json

devconsole.json
* "Build Configs": "빌드 설정", but "Select a Project to view the list of BuildConfigs or <2>create a Project</2>.": "프로젝트를 선택하여 BuildConfigs 목록을 보거나 <2>프로젝트를 생성</2>합니다.", and "Select a Project to view the list of BuildConfigs": "빌드 설정 목록을 표시할 프로젝트를 선택",

#### Chinese:
console-app.json
* "Create VolumeSnapshot": "创建卷快照", but "VolumeSnapshot": "VolumeSnapshot",
[...]

PR #9145, opened 2021-06-04
Revert unintended changes during translations

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-5935

**Analysis/Root cause:**
With this translations PR https://github.com/openshift/console/pull/8720 `Finally` text was modified to `Final`.

**Solution:**
Change `Final` back to `Finally` and also fix the translations.

**Screenshots:**
Screenshot 2021-06-04 at 5 25 23 PM

/kind bug

To Do
Update translations for other supported languages.


rohitkrai03 (approved):
/approve

PR #9146, opened 2021-06-04
add unique name to the radio buttons to avoid name collision

karthikjeeyar:
**Fixes**: 

https://issues.redhat.com/browse/ODC-5929

**Analysis / Root cause**: 

Same name to the RadioInput components caused the collision in the start modal, so select an option in the second workspace;s radio input removes the selection from the first workspace radio fields.

**Solution Description**: 

Add a unique name to the RadioInput components in the modal.

image

**Test setup:**

1. Create a pipeline and add two workspaces.
2. Start the pipeline and select `volumeClaimeTemplates` option for both the workspaces.

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

cc: @andrewballantyne


rohitkrai03 (approved):
/lgtm
/approve

PR #9153, opened 2021-06-07
Bug 1967639: fixes: console whitescreens if user settings fails to load

sahil143:
fixes: https://issues.redhat.com/browse/OCPBUGSM-29992

Whenever user settings fail to load console breaks.
This PR fixes:

- Make sure a new config map is created only when K8sWatchResource fails with 404, else fallback to local storage
- If create config map request fails, fallback to local storage.

### DIFF
--- a/frontend/packages/console-shared/src/hooks/useUserSettings.ts
+++ b/frontend/packages/console-shared/src/hooks/useUserSettings.ts
@@ -85,16 +85,14 @@ export const useUserSettings = <T>(
     if (isLocalStorage) {
       return;
     }
-    if (cfLoadError || (!cfData && cfLoaded)) {
+    if (cfLoadError?.response?.status === 404 || (!cfData && cfLoaded)) {
       (async () => {
         try {
           await createConfigMap();
         } catch (err) {
-          if (err?.response?.status === 403 || err?.response?.status === 404) {
-            setFallbackLocalStorage(true);
-          } else {
-            throw err;
-          }
+          // eslint-disable-next-line no-console
+          console.error(err);
+          setFallbackLocalStorage(true);
         }
       })();
     } else if (
@@ -120,9 +118,11 @@ export const useUserSettings = <T>(
       updateConfigMap(cfData, keyRef.current, seralizeData(defaultValueRef.current));
       setSettings(defaultValueRef.current);
       setLoaded(true);
-    } else if (cfLoaded) {
+    } else if (cfLoaded && !cfLoadError) {
       setSettings(defaultValueRef.current);
       setLoaded(true);
+    } else if (cfLoadError && cfLoadError.response?.status !== 404) {
+      setFallbackLocalStorage(true);
     }
     // This effect should only be run on change of configmap data, status.
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/frontend/packages/console-shared/src/hooks/__tests__/useUserSettings.spec.ts
+++ b/frontend/packages/console-shared/src/hooks/__tests__/useUserSettings.spec.ts
@@ -82,7 +82,9 @@ describe('useUserSettings', () => {
 
     // Mock ConfigMap not found
     await act(async () => {
-      useK8sWatchResourceMock.mockReturnValue([null, false, new Error('ConfigMap not found')]);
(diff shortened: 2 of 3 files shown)


frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -85,16 +85,12 @@ export const useUserSettings = <T>(
...
       (async () => {
         try {
           await createConfigMap();
         } catch (err) {
-          if (err?.response?.status === 403 || err?.response?.status === 404) {
-            setFallbackLocalStorage(true);
-          } else {
-            throw err;
-          }
+          setFallbackLocalStorage(true);
```
spadgett, line 93: I'd add a `console.error` message as well for debugging.

frontend/packages/console-shared/src/utils/user-settings.ts
```diff
@@ -44,7 +44,7 @@ export const updateConfigMap = async (
   } catch (err) {
     // eslint-disable-next-line no-console
     console.error(err);
-    throw err;
+    return Promise.reject();
```
spadgett, line 47: I'm unclear on the intent of this change. Won't this be treated as a rejected promise either way?

frontend/packages/console-shared/src/hooks/useUserSettings.ts
```diff
@@ -85,16 +85,12 @@ export const useUserSettings = <T>(
     if (isLocalStorage) {
       return;
     }
-    if (cfLoadError || (!cfData && cfLoaded)) {
+    if (cfLoadError?.response?.status === 404 || (!cfData && cfLoaded)) {
```
spadgett, line 88: I would expect the condition after the `||` to evaluate to `true` on other load errors, so I'm not sure if checking for 404 will change the behavior.

spadgett (changes requested):
We should try to set up a environment with multi-cluster to confirm this fixes the whitescreen. cc @TheRealJon 

We should also catch errors trying to set values in `localStorage`. That can fail for various reasons and shouldn't break console.

cc @christianvogt

frontend/packages/console-shared/src/utils/user-settings.ts
christianvogt: Yes they both return a rejected promise in an async function. Why the change?

jerolimov:
@invincibleJai @sahil143 I think the problem is that `useUserSettings` returns always loaded=false and `useValuesForPerspectiveContext` and `DetectPerspective` doesn't render anything until loaded is true.

Other hooks like `useK8sWatchResource` returns loaded also if there is an, right? I think we should check if we can do this in `useUserSettings` as well.

And maybe we can return the error also? So that the caller can check if useUserSettings has failed (if the code can handle this)

```diff
export const useUserSettings = <T>(
  key: string,
  defaultValue?: T,
  sync: boolean = false,
-): [T, React.Dispatch<React.SetStateAction<T>>, boolean] => {
+): [T, React.Dispatch<React.SetStateAction<T>>, boolean, Error or any] => {
```

Wdyt?

frontend/packages/console-shared/src/utils/user-settings.ts
sahil143: My bad, confused myself here. fixed it.

invincibleJai:
@sahil143  thanks for the fix, I verified below and looks fine to me

- If get configMap fails with 404, attempts to create configMap
- If get configMap fails with any error other than 404,  fallback to local storage
- if create configMap fails with any err, fallback to local storage
- if get configMap succeed, config map is used

there are below scenarios which is considered
- if get configMap errors with 404 but create configMap success but even then get configMap errors with 404 (this is an invalid case)
- if update ConfigMap fails no fallback and will try updating with next iterations

invincibleJai:
/approve

have verified and posted my observation above

cc @christianvogt @spadgett

christianvogt:
> - if get configMap errors with 404 but create configMap success but even then get configMap errors with 404 (this is an invalid case)

I agree this is invalid because we create the role bindings so that shouldn't happen.

> - if update ConfigMap fails no fallback and will try updating with next iterations

This is fine because an update should be working if the creation succeeds.

/lgtm

PR #9156, opened 2021-06-07
Fix issue with selected capacity donut chart

SanjalKatiyar:
URL: https://issues.redhat.com/browse/RHSTOR-1878

**Before:**
Screenshot from 2021-06-07 16-24-27
Screenshot from 2021-06-07 16-24-35

**After:**
Screenshot from 2021-06-07 17-04-53
Screenshot from 2021-06-07 17-05-04

### DIFF
--- a/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard-steps/create-storage-class/selected-capacity.tsx
+++ b/frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard-steps/create-storage-class/selected-capacity.tsx
@@ -123,7 +123,7 @@ export const SelectedCapacity: React.FC<SelectedCapacityProps> = ({ ns, state, d
     );
   }
 
-  const chartDisks = selectedLvsNodes.length
+  const chartDisks = state.lvsIsSelectNodes
     ? filteredDisks.filter((disk: DiscoveredDisk) => selectedLvsNodes.includes(disk.node))
     : filteredDisks;
   const chartNodes: Set<string> = chartDisks.reduce(


frontend/packages/ceph-storage-plugin/src/components/ocs-install/attached-devices-mode/install-wizard-steps/create-storage-class/selected-capacity.tsx
```diff
@@ -123,7 +123,7 @@ export const SelectedCapacity: React.FC<SelectedCapacityProps> = ({ ns, state, d
     );
   }
 
-  const chartDisks = selectedLvsNodes.length
+  const chartDisks = state.lvsIsSelectNodes
```
GowthamShanmugam, line 126:
```
const selectedLvsNodes = state.lvsSelectNodes.map(getName);
```

Both the fix are looking similar only, how this fix works?

SanjalKatiyar: `state.lvsIsSelectNodes` --> `true` if "Select nodes" radio button is clicked else `false`;
`selectedLvsNodes` --> list of nodes selected after "Select nodes" radio button is clicked.
`filteredDisks` --> all the disks which satisfies the applied filters, **not dependent** upon `selectedLvsNodes` list.

Earlier:
If `selectedLvsNodes` is empty (ternary condition `false`) we are displaying `filteredDisks`, thus donut chart will still show disks on all the nodes.
Similary,
Let's suppose I have selected 2 nodes from the nodes list and now I click "All nodes" radio button again. `selectedLvsNodes` list will have 2 elements, and even though I have selected "All nodes" option, donut chart will still show the status of 2 `selectedLvsNodes` disks, instead of disks on all nodes.

GowthamShanmugam: Oh ok lvsIsSelectNodes is a flag, got it

GowthamShanmugam (approved):

GowthamShanmugam:
@afreen23 Changes looks good to me

afreen23 (commented):
/lgtm

PR #9158, opened 2021-06-07
Bug 2005783: Fix accessibility issues in Wizard Modes

rexagod:
Fixes accessibility issues in the "Internal" and "Internal - Attached Mode" Installation Flow, using AXE Devtools. A majority of these issues were found inside some Patternfly components in upstream, and the ones that were not have been fixed in this PR.

Fixes: https://issues.redhat.com/browse/RHSTOR-1868
Fixes: https://issues.redhat.com/browse/RHSTOR-1869


rexagod:
@afreen23 @bipuladh I can take this up with PF devs if that's the way to go here. WDYT?

afreen23 (changes requested):
We need to externalize aria-labels as well.

dtaylor113 (approved):
/lgtm

rexagod:
> We need to externalize aria-labels as well.

Just to clarify, this means that I should introduce `aria-labels` in the Installation Flow wherever possible, right?

afreen23:
> > We need to externalize aria-labels as well.
> 
> Just to clarify, this means that I should introduce `aria-labels` in the Installation Flow wherever possible, right?

externalize for i18n - `t(ceph-storage-plugin~)`

bipuladh:
I can see a bunch of places that are missing aria tags. I think we need to add them in a lot more places. Is the AXE tool not reporting them?

rexagod:
@bipuladh AXE does not report missing `aria-label`s, but it does report if any element has an `aria-*` attribute that does not satisfy that element's `role` attribute.

ss

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
```diff
@@ -125,7 +125,7 @@ export const EncryptionFormGroup: React.FC<EncryptionFormGroupProps> = ({
   };
 
   return (
-    <FormGroup fieldId="configure-encryption" label="Encryption">
```
rexagod, line 128: Removed this since AXE complained about the child elements having nested labels. (Jessie's suggestion)

afreen23: @rexagod this will impact the usual. On this, we step we have two groups - 
1) encryption
2) network
We can pass a custom component in label prop for that so it does not get <label>. 
And its not only one place, if you see below, we have a lot of those.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
```diff
@@ -84,41 +83,23 @@ export const CreateInternalCluster: React.FC<CreateInternalClusterProps> = ({
...
-  const { getStep, getParamString, getIndex, getAnchor } = navUtils;
-
-  const toLink = (steps: any, currentStep: string, modes: any) =>
-    getAnchor(getIndex(steps, currentStep, 2), getIndex(modes, modes.INTERNAL));
+  const { getStep, getParamString, getIndex } = navUtils;
 
   const steps: WizardStep[] = [
     {
-      name: (
-        <Link to={toLink(CreateStepsSC, CreateStepsSC.STORAGEANDNODES, MODES)}>
```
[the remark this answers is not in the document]
afreen23: Will this impact the routes?
I feel you need to add one prop in the wizard to navigate across nav items.

afreen23 (changes requested):

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
rexagod: Okay, so I think it'll be better to revert this in that case, so as to avoid a potential regression?

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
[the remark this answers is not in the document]
rexagod: @afreen23 `onGoToStep` handles the routing part.

https://github.com/openshift/console/pull/9158/files#diff-9cf7ec8e4919d5934a0cf4c2d79d774374f36c1f20ea1a6759edfd3ebf256e9fR171

frontend/packages/ceph-storage-plugin/src/components/ocs-install/install-wizard/configure.tsx
afreen23: Nope, we must fix all of them.
You can pass a custom label.
label=<div></div> or <span></span> whatever suites and does not affect current css.
May be we can use <TextContext>

frontend/packages/ceph-storage-plugin/src/types.ts
```diff
@@ -290,7 +290,7 @@ export type DiscoveredDisk = {
 export type NavUtils = {
   getStep: (maxSteps?: number) => number;
   getParamString: (step: number, mode: number) => string;
-  getIndex: (searchSpace: any, search: string, offset?: number) => number;
+  getIndex: (searchSpace: any, search: string | React.ReactText, offset?: number) => number;
```
bipuladh, line 293:
```suggestion
  getIndex: (searchSpace: any, search: React.ReactText, offset?: number) => number;
```

rexagod:
/test ceph-storage-plugin
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_console/9158/pull-ci-openshift-console-master-ceph-storage-plugin/1427619450982502400#1:build-log.txt%3A49

bipuladh:
/assign @afreen23 
/hold 
wait for 4.10

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
```diff
@@ -84,41 +83,23 @@ export const CreateInternalCluster: React.FC<CreateInternalClusterProps> = ({
...
-          {' '}
-          {t('ceph-storage-plugin~Capacity and nodes')}{' '}
-        </Link>
-      ),
+      name: t('ceph-storage-plugin~Select capacity and nodes'),
```
SanjalKatiyar, line 90: It is `Capacity and nodes` and IIRC `Link` was added because it gave navItems clickable link like UI (otherwise these items will just display as black text). So, if we are removing this we should confirm with UX once ??

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
```diff
@@ -84,41 +83,23 @@ export const CreateInternalCluster: React.FC<CreateInternalClusterProps> = ({
...
       enableNext: !!(state.nodes.length >= MINIMUM_NODES && scName),
     },
     {
-      name: (
-        <Link to={toLink(CreateStepsSC, CreateStepsSC.CONFIGURE, MODES)}>
-          {' '}
-          {t('ceph-storage-plugin~Security and network')}{' '}
-        </Link>
-      ),
+      name: t('ceph-storage-plugin~Configure'),
```
SanjalKatiyar, line 96: Similar, it is `Security and network` not `Configure`... it was changed around 4.8 (https://issues.redhat.com/browse/RHSTOR-1670)

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
SanjalKatiyar: IIRC `Link` also provided clickable link like UI to these navItems (items were displayed blue in color, otherwise they will be displayed as a black text).

rexagod:
@SanjalKatiyar Fixed the naming, but for the steps I think the latter one makes more sense. WDYT?

2021-08-31-222857_188x118_scrot
2021-08-31-221059_191x112_scrot

SanjalKatiyar:
> Fixed the naming, but for the steps I think the latter makes more sense. @SanjalKatiyar WDYT?
> 
> PS. We can always get that link UI back with a little css, so no worries there.
> 
> #### master
> 2021-08-31-222857_188x118_scrot
> 
> #### HEAD (this PR)
> 2021-08-31-221059_191x112_scrot

Both looks good to me (lol)... but which to choose might need UX involvement. Also, I can see that we will be merging this PR in 4.10 and in 4.10 mostly probably this UI will be outdated as we will be using ODF flow for system creation, right ??

rexagod:
> Steps naming was also changed by UX in 4.8
Yep, I've fixed that in the PR during the rebase.

> Both looks good to me (lol)... but which to choose might need UX involvement. Also, I can see that we will be merging this PR in 4.10 and in 4.10 mostly probably this UI will be outdated as we will be using ODF flow for system creation, right ??
I guess we can backport it if need be. I should check with Hadas in this. Thanks for the insight, Sanjal!

SanjalKatiyar:
> > Steps naming was also changed by UX in 4.8
> 
> Yep, I've fixed that in the PR during the rebase.
> 
> > Both looks good to me (lol)... but which to choose might need UX involvement. Also, I can see that we will be merging this PR in 4.10 and in 4.10 mostly probably this UI will be outdated as we will be using ODF flow for system creation, right ??
> 
> I guess we can backport it if need be. I should check with Hadas in this. Thanks for the insight, Sanjal!

no worries...
most probably we will not backport axe related issue, so, just a suggestion that confirming with the team might be better before asking UX guys.

bipuladh (approved):

afreen23 (commented):
This PR makes some UX changes as well. 
The internal mode looks and feel will be different from attached devices mode.

With that said, it might be the case we do not want to support storage cluster at all in 4.10.
I would hold this PR to do any more changes (mentioned above). 
Lets discuss offline.

frontend/packages/ceph-storage-plugin/src/components/ocs-install/internal-mode/install-wizard.tsx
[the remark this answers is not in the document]
rexagod: Not sure if I'm getting this right but don't we already do that in this PR?

Internal mode: https://github.com/openshift/console/pull/9158/files#diff-9cf7ec8e4919d5934a0cf4c2d79d774374f36c1f20ea1a6759edfd3ebf256e9fL94
Attached Devices mode: https://github.com/openshift/console/pull/9158/files#diff-36110363774ffaff83e8bcf48416cdbb872102c6ea085bae2d123b1c97d1fbccL261

afreen23 (approved):

afreen23 (commented):
/lgtm

All changes are consistent can be merged

PR #9161, opened 2021-06-07
Bug 1933101: Add chart URl to annotations during helm install/upgrade

zonggen:
Resolve helm upgrade failure when dependencies are present, caused by
missing chart URL. This commit adds chart URL under chart annotations
during installation. The URL is then picked up and maintained during an
upgrade.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1933101
Signed-off-by: Allen Bai <abai@redhat.com>


pkg/helm/actions/upgrade_release.go
```diff
@@ -23,6 +23,12 @@ func UpgradeRelease(ns, name, url string, vals map[string]interface{}, conf *act
 		}
 		return nil, err
 	}
+
+	// Before proceeding, check if chart URL is present as an annotation
+	if chart_url, ok := rel.Chart.Metadata.Annotations["chart_url"]; url == "" && ok {
```
dperaza4dustbit, line 28: Do we need to check that Annotations is not nil. Since this could be a release that was there prior the fix or it comes from helm cli? Notice that by default Annotation is nil.

pkg/helm/actions/upgrade_release.go
```diff
@@ -45,5 +51,11 @@ func UpgradeRelease(ns, name, url string, vals map[string]interface{}, conf *act
 		}
 	}
 
+	// Ensure chart URL is properly set in the upgrade chart
+	if _, ok := ch.Metadata.Annotations["chart_url"]; url != "" && !ok {
```
dperaza4dustbit, line 55: Same comment on checking for nil.

dperaza4dustbit (changes requested):

pkg/helm/actions/upgrade_release.go
zonggen: Agreed. Changing to check nil condition for Annotations.

pkg/helm/actions/install_chart.go
```diff
@@ -25,6 +25,10 @@ func InstallChart(ns, name, url string, vals map[string]interface{}, conf *actio
 		return nil, err
 	}
 
+	// Add chart URL as an annotation before installation
+	ch.Metadata.Annotations = make(map[string]string)
```
zonggen, line 29: Should I safety check that Metadata is not nil here?

dperaza4dustbit: Not sure... I think since ch was loaded and there was no error I would expect Metadata to be there. However you brought up another possibility which is the charts do come with some annotation. In that case we should not create a new instance but re-use the exiting one there.

pedjak: I think you should check and use the structure if available.

pkg/helm/actions/upgrade_release.go
pedjak: Do we have now a chance to have a situation when `url` is still empty?

pkg/helm/actions/upgrade_release.go
```diff
@@ -23,6 +23,12 @@ func UpgradeRelease(ns, name, url string, vals map[string]interface{}, conf *act
...
 	}
+
+	// Before proceeding, check if chart URL is present as an annotation
+	if chart_url, ok := rel.Chart.Metadata.Annotations["chart_url"]; url == "" && ok {
+		url = chart_url
+	}
+
 	// if url is not provided then we expect user trying to upgrade release with the same
 	// version of chart but with the different values
 	if url == "" {
```
pedjak, line 34: see above, can this happen now?

pkg/helm/actions/install_chart.go
zonggen: Yea we don't want to flush the existing annotations. Will check for existing Annotations here.

pkg/helm/actions/upgrade_release.go
zonggen: Yes it seems that url will be still empty, and this workaround is to pick up url from annotation.

pkg/helm/actions/upgrade_release.go
dperaza4dustbit: I think it can if the release came from cli, for example

pkg/helm/actions/upgrade_release.go
zonggen: Updated.

pkg/helm/actions/upgrade_release.go
zonggen: Updated.

pkg/helm/actions/install_chart.go
zonggen: Updated.

pkg/helm/actions/upgrade_release.go
```diff
@@ -45,5 +53,15 @@ func UpgradeRelease(ns, name, url string, vals map[string]interface{}, conf *act
 		}
 	}
 
+	// Ensure chart URL is properly set in the upgrade chart
+	if url != "" {
+		if ch.Metadata.Annotations == nil {
+			ch.Metadata.Annotations = make(map[string]string)
+		}
+		if _, ok := ch.Metadata.Annotations["chart_url"]; !ok {
```
dperaza4dustbit, line 61: hum... in this case you do want to override what is there. This will cover the case where you are upgrading to a different version which will have a diff url. So... I would not check if there is not value to set url. Set it every time.

dperaza4dustbit (changes requested):

pkg/helm/actions/upgrade_release.go
zonggen: Thanks for pointing this out. Was wondering about that case but went with safe approach, which is sitcking with the first set URL. Will update.

zonggen: Updated.

dperaza4dustbit (approved):
This looks complete to me

PR #9163, opened 2021-06-07
Bug 1966684: Fix creating snapshot for claim <name> string

florkbr:
This string is missing an additional bracket pair around the variable being substituted. While this appeared to work for English, when I set my browser to use Chinese I still see the English string.

## Before
Screen Shot 2021-06-07 at 10 48 32 AM
Screen Shot 2021-06-07 at 11 36 59 AM

## After (with mocked translation change to zh/console.json)
Screen Shot 2021-06-07 at 11 39 17 AM

https://bugzilla.redhat.com/show_bug.cgi?id=1966684


spadgett (approved):
/lgtm

👍

PR #9167, opened 2021-06-07
Add error boundary to topology page

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5809

**Analysis / Root cause**: 
We had different errors in topology when data are missed or new operator versions provide unexpected data to the topology. The current error boundary is part of each individual content page as `ListPage` and `DetailPage` and was not part of `TopologyPage`.

**Solution Description**: 
Added `withFallback` which is used in [ListPage](https://github.com/openshift/console/blob/master/frontend/public/components/factory/list-page.jsx#L317), DetailPage and other components to topology page.

Add this on two different levels to catch errors in topology graph visualization or list and topology at all. The second error boundary on the graph visualization and list allows the user to switch between both topology view types if only the other one iis broken. See screenshots below.

Without error boundary:
White screen and complete console is crashing.

Without the additional error boundary the content block shows an error, but the user can not switch between the graph and list view:
Screenshot from 2021-06-07 18-56-29
Screenshot from 2021-06-07 19-02-30

With the additional error boundary the user can still switch between the graph and list view:
Screenshot from 2021-06-07 19-04-43
Screenshot from 2021-06-07 19-05-47

Only change `TopologyPage` unit tests to find the right view after adding new component hierarchy level.

**Test setup:**
Manually added a crash in `UrlDecorator` by throwing an error. At the moment I don't see an option to test this on a cluster.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/topology/src/__tests__/TopologyPage.spec.tsx
+++ b/frontend/packages/topology/src/__tests__/TopologyPage.spec.tsx
@@ -58,38 +58,45 @@ describe('Topology page tests', () => {
   });
 
   it('should render topology page', () => {
-    const wrapper = shallow(<TopologyPage match={match} hideProjects={false} />);
+    const wrapperWithFallback = shallow(<TopologyPage match={match} hideProjects={false} />);
+    const wrapper = wrapperWithFallback.find('TopologyPage').shallow();
     expect(wrapper.find(NamespacedPage).exists()).toBe(true);
   });
 
   it('should default to graph view', () => {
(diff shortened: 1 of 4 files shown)


jerolimov:
/label px-approved
added on behalf of px as being a tech debt epic it's not needed

/label docs-approved
added on behalf of docs team as being a tech debt epic it's not needed.

invincibleJai:
/lgtm

/label qe-approved
verified the changes and works as expected.

PR #9168, opened 2021-06-07
Bug 1966077: Don't show empty hidden fields on operand details page

TheRealJon:
If a hidden descriptor is applied to an object or all of it's children, hide it on the details page.


spadgett (approved):
/lgtm

PR #9170, opened 2021-06-07
Bug 1963209: Some strings in public.json have errors

florkbr:
✅ ”Version {{version}} not found in channel {{channel}}.  To request update recommendations, configure a channel that supports your version." - There should not be two spaces in front of the period; there should only be one.

✅ "Checking this box will write the url to the global section of the configuration file where it will become default url for future PagerDuty receivers." - "url" is inconsistent with other places we write URL. We should also have a "the" before "default url."

❌ (fixed in master for EN) "Add new Users to Group {name}." - this needs to be reviewed. Usually you need to use double brackets to pass in data

✅ "Remove User {user} from Group {name}?" - this needs to be reviewed. Usually you need to use double brackets to pass in data

✅ "PostStart: {label('postStart')} <3>{postStart}</3>" - this needs to be reviewed. Usually you need to use double brackets to pass in data

✅ "PreStop: {label('preStop')} <3>{preStop}</3>" - this needs to be reviewed. Usually you need to use double brackets to pass in data

https://bugzilla.redhat.com/show_bug.cgi?id=1963209


spadgett (approved):
/lgtm

PR #9179, opened 2021-06-08
Bug 1963027: Upload qcow2 too PVC to small

glekner:
Screen Shot 2021-06-08 at 14 35 28

Added check of pvc size against the image size, if the image is larger this error is shown

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
@@ -88,7 +88,8 @@ import {
   CDI_UPLOAD_URL_BUILDER,
 } from '../consts';
 import { uploadErrorType, UploadPVCFormStatus } from './upload-pvc-form-status';
-
+import { BinaryUnit, convertToBytes } from '../../form/size-unit-utils';
+import { getDataVolumeStorageSize } from '../../../selectors/dv/selectors';
 import './upload-pvc-form.scss';
 
 const templatesResource: WatchK8sResource = {
@@ -114,6 +115,13 @@ export const uploadErrorMessage = (t: TFunction) => ({
     </Trans>
   ),
 });
+
+export const getGiBUploadPVCSizeByImage = (sizeInBytes: number) => {
+  const sizeGi = sizeInBytes / 1024 / 1024 / 1024;
+
+  if (sizeGi < 0.5) return 1;
+  return Math.ceil(sizeGi) * 2;
+};
 
 export const UploadPVCForm: React.FC<UploadPVCFormProps> = ({
   onChange,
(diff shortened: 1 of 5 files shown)


frontend/packages/kubevirt-plugin/src/components/cdi-upload-provider/upload-pvc-form/upload-pvc-form.tsx
```diff
@@ -610,6 +614,8 @@ export const UploadPVCPage: React.FC<UploadPVCPageProps> = (props) => {
     e.preventDefault();
     if (!fileName) {
       setError(uploadErrorType.MISSING);
+    } else if (fileValue?.size > convertToBytes(getDataVolumeStorageSize(dvObj))) {
```
yaacov, line 617: the check should be for double the size of the file, see discussion in BZ

kobi86:
Hey @glekner, 
I think we should consider setting the size of the PVC automatically.
During the wizard of uploading disk, we can just create the PVC automatically twice the size of the image the user uploaded (we can ask to upload the file early as possible) OR we can just create the PVC automatically in the right size  (we can give the user the option to change) with the name of the image-PVC. 
This will very much help help the users and add the feeling of Out Of The Box Experience

glekner:
@yaacov @kobi86
PVC Size is now automatically set to x2 File size on picking a file.

if the pvc size is lower than that, the warning is shown with link to https://docs.openshift.com/container-platform/4.7/virt/virtual_machines/virtual_disks/virt-uploading-local-disk-images-block.html#:~:text=when%20uploading%20a%20disk%20image%20to%20a%20pvc%2C%20the%20pvc%20size%20must%20be%20larger%20than%20the%20size%20of%20the%20uncompressed%20virtual%20disk

Screen Shot 2021-06-09 at 14 29 24

kobi86:
@glekner @yaacov 
Not sure we are going the right way here, if we are wrong about the size (the file the user uploads is 5G but the actual size is 15G) then we are misleading to an error.. so I think that maybe we should show this message all the time with the info: 
"Please insure that you PVC size .... Learn more"

This way if the user doesn't know he will have to check, if he gets an error he will go to the docs.
In this case, we are telling the user something that we are not sure about because we don't know the real size of the disk. 

What do you say?

yaacov:
> @glekner @yaacov
> Not sure we are going the right way here, if we are wrong about the size (the file the user uploads is 5G but the actual size is 15G) then we are misleading to an error.. so I think that maybe we should show this message all the time with the info:
> "Please insure that you PVC size .... Learn more"
> 
> This way if the user doesn't know he will have to check, if he gets an error he will go to the docs.
> In this case, we are telling the user something that we are not sure about because we don't know the real size of the disk.
> 
> What do you say?

I''m ok with any wording that will make the user notice that they need to have a PVC the size of the uncomplresed file, and not 1:1 like I thought before we started talking about this issue.

PR #9184, opened 2021-06-08
Bug 1967275: Prevent awkward wrap of icons with getting started links and buttons

sg00dwin:
Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1967275

Note: I had to switch the icons to display as appended Pseudo-class `:after` elements in order to keep them from wrapping within their parent elements that are displayed as flex containers.

Screen Shot 2021-06-08 at 2 53 48 PM

### DIFF
--- a/frontend/public/style/_common.scss
+++ b/frontend/public/style/_common.scss
@@ -253,7 +253,7 @@ dl.co-inline {
 
 // append external-link icon to <a> so that it doesn't wrap without text
 // there must be no white space between the text and the closing </a> tag holding the pseudo element
-.co-external-link:after {
+.co-external-link::after {
   @include font-awesome-free-solid;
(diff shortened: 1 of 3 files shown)


spadgett (approved):
/lgtm

frontend/public/style/_common.scss
```diff
@@ -281,6 +281,18 @@ dl.co-inline {
   display: block;
 }
 
+.co-goto-arrow:after {
```
rhamilto, line 284: Nit:
```suggestion
.co-goto-arrow::after {
```

sg00dwin: updated

spadgett (approved):
/lgtm

PR #9199, opened 2021-06-09
[release-4.6] Bug 1969878: fix resources fetched by firehose to show the correct Pod donut

sahil143:
This PR fixes the issue with how we fetch the resource for pod donuts through firehose.

Issue: when we show the pod donut on a deployment config page it tries to fetch `deploymentconfigs`, `Pods`, `replicasets` and `replicationcontrollers`. If a user doesn't have access to deployments and replicasets firehose throws an error and pod donut doesn't show up on the deployment config page

Solution: only fetch resources related to them i.e. for a DC only fetch ReplicationControllers and Pods. We don't need to fetch replicasets when we need to show the pod donut for deployment config

Screenshot 2021-06-09 at 8 40 19 PM

### DIFF
--- a/frontend/packages/console-shared/src/components/pod/PodRingDataController.tsx
+++ b/frontend/packages/console-shared/src/components/pod/PodRingDataController.tsx
@@ -3,13 +3,11 @@ import { Firehose, FirehoseResource } from '@console/internal/components/utils';
 import { PodRingResources, PodRingData } from '../../types';
 import { transformPodRingData, podRingFirehoseProps } from '../../utils';
 import {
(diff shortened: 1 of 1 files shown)


invincibleJai:
/lgtm

Verified the changes with cluster-bot and works as expected

cc @vikram-raj

rohitkrai03 (approved):
/approve

vikram-raj (commented):
verified the changes and it works as expected.

/lgtm

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

approved

PR #9203, opened 2021-06-10
Remove &apos; from the text

SanjalKatiyar:
Remove the `&apos;` . Since this is not understood by the translator and causing errors in console.


afreen23 (commented):
/lgtm

PR #9204, opened 2021-06-10
Bug 1970381: Monitoring dashboards: Custom time range inputs should retain values

kyoto:
After setting a custom time range once and then opening the custom time
range modal again, the inputs should show the previously selected
values.


spadgett (approved):
/lgtm

PR #9123, opened 2021-06-02
Refactor multiple-storage-class-selection tests

SanjalKatiyar:
URL: https://issues.redhat.com/browse/RHSTOR-1873

PR #9208, opened 2021-06-10
Remove cy.wait from multiple-storageclass-selection tests

SanjalKatiyar:
`multiple-storageclass-selection` tests --> `Add capacity with a new storage class having NO-PROVISIONER as provisioner` --> We only want to click the add capacity button once both "pvData" and "nodesData" gets loaded, hence cy.wait was added earlier.

Now, removed cy.wait condition. Add capacity button will be disabled if there are any errors or data is not completely loaded.


SanjalKatiyar:
Closed this PR as the fixes for this PR are incorporated in https://github.com/openshift/console/pull/9123

PR #9210, opened 2021-06-10
Filter disabled add actions for topology menus

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-5949

**Analysis / Root cause**: Topology add actions menu that are disabled still show up in the menu.

**Solution Description**: Filter out add actions that are disabled.

https://user-images.githubusercontent.com/6041994/121564968-9f549800-ca39-11eb-8003-f37b038a42dc.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/dev-console/src/actions/add-resources.tsx
+++ b/frontend/packages/dev-console/src/actions/add-resources.tsx
@@ -13,84 +13,121 @@ import {
 import * as helmIcon from '@console/internal/imgs/logos/helm.svg';
 import { ImportOptions } from '../components/import/import-types';
 import { KebabAction, createKebabAction } from '../utils/add-resources-menu-utils';
+import { getDisabledAddActions } from '../utils/useAddActionExtensions';
 
 export const allImportResourceAccess = 'allImportResourceAccess';
 export const allCatalogImageResourceAccess = 'allCatalogImageResourceAccess';
 
-export const fromGit = createKebabAction(
-  // t('devconsole~From Git')
-  'devconsole~From Git',
-  <GitAltIcon />,
-  ImportOptions.GIT,
-  allImportResourceAccess,
-);
+type AddActionItem = { id: string; action: KebabAction };
 
-export const containerImage = createKebabAction(
-  // t('devconsole~Container Image')
-  'devconsole~Container Image',
-  <OsImageIcon />,
-  ImportOptions.CONTAINER,
-  allCatalogImageResourceAccess,
-);
+export const fromGit: AddActionItem = {
+  id: 'import-from-git',
+  action: createKebabAction(
+    // t('devconsole~From Git')
+    'devconsole~From Git',
+    <GitAltIcon />,
+    ImportOptions.GIT,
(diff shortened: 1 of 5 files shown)


frontend/packages/dev-console/src/actions/add-resources.tsx
```diff
@@ -13,84 +13,121 @@ import {
...
+  action: createKebabAction(
+    // t('devconsole~Upload JAR file')
+    'devconsole~Upload JAR file',
+    <FileUploadIcon />,
+    ImportOptions.UPLOADJAR,
+    allCatalogImageResourceAccess,
+  ),
+};
+
+const disabledAddActions = getDisabledAddActions();
```
andrewballantyne, line 128: This will invoke once on first import and then not watch / ever get updated SERVER_FLAGS data, right? Wouldn't this need a browser refresh to be updated?

andrewballantyne (commented):
Unless there are other factors that prevent more dynamic reloading -- this is a problem, no?

frontend/packages/dev-console/src/actions/add-resources.tsx
rohitkrai03: Any update to console config rolls out a new deployment of console which updates the SERVER_FLAGS. That again requires a browser refresh to get the new flags. So I think this is not going to be an issue.

andrewballantyne: Interesting, we have live updates of SERVER_FLAGS, no? I'm only concerned about the response-ability of the app. But if you feel this isn't a loss of features (uniquely), I'll let it go. Some things like the Terminal not live updating into the masthead feels anti-modern app. I'm just trying to state against more of "that" kinda features.

rohitkrai03: All of our features that rely on the console config CR actually have the same issue. The server flags are set using environment variables of the pod by our backend here https://github.com/openshift/console/blob/master/pkg/server/server.go#L498. It just updates the `index.html` with all the flag values. So it cannot have live updates. AFAIK, we do not have a way to get dynamic `SERVER_FLAGS` updates.
@christianvogt correct me if I am wrong.

christianvogt: `SERVER_FLAGS` never change. A change to the config will roll out a new deployment of console. 
We don't need to watch this.

This has been brought up before but the server flags approach is legacy and best to bring it up at the console arch meeting.

andrewballantyne: My bad 👍  Thanks guys.

andrewballantyne (approved):
Tested via the Chrome dev tools console (like the gif)... it works. I however don't seem to be able to make the console config do this. 🤔

andrewballantyne:
/lgtm

Talked with Christian and it's not testable in localhost.

PR #9214, opened 2021-06-10
translate the page heading in the edit flow

debsmita1:
**Fixes:**
https://bugzilla.redhat.com/show_bug.cgi?id=1942125

**Root analysis:**
In the Edit flow, the page heading is not translated

**Solution description:**
Added translation for Page title in the Edit flow

**Screenshot:**
Screenshot from 2021-06-11 00-32-45


frontend/packages/dev-console/src/components/edit-application/EditApplication.tsx
```diff
@@ -53,7 +55,7 @@ const EditApplication: React.FC<EditApplicationProps> = ({
     if (values.build.strategy) {
       const imageStream =
         values.image.selected && builderImages ? builderImages[values.image.selected].obj : null;
-      if (pageHeading === CreateApplicationFlow.JarUpload) {
+      if (pageHeading === i18n.t(EditFlowPageTitle[CreateApplicationFlow.JarUpload])) {
```
andrewballantyne, line 58: This is within the scope of a Component, please use the `t` from `useTranslation`. The main goal is not to pass `t` to functions. Hooks are defined in the component, anything within "it's window" should (for code simplicity) just use the hook `t`.

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -50,6 +51,17 @@ export enum CreateApplicationFlow {
...
+export const EditFlowPageTitle = {
+  // t('devconsole~Import from Git')
+  [CreateApplicationFlow.Git]: 'devconsole~Import from Git',
+  // t('devconsole~Import from Dockerfile')
+  [CreateApplicationFlow.Dockerfile]: 'devconsole~Import from Dockerfile',
+  // t('devconsole~Deploy Image')
+  [CreateApplicationFlow.Container]: 'devconsole~Deploy Image',
+  // t('devconsole~Upload JAR file')
+  [CreateApplicationFlow.JarUpload]: 'devconsole~Upload JAR file',
+};
```
andrewballantyne, line 63: I'm not opposed to this model, but this puts the onus on the caller to understand these are `t` values. Seems odd to impose a dependency on a caller of a map 🤔 

Only alternative I can think of is making this a function call... `getEditFlowTitleWording(CreateApplicationFlow)` => `t` string.

andrewballantyne: No change required... just voicing a thought.

frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
```diff
@@ -43,36 +44,43 @@ const EditApplicationForm: React.FC<FormikProps<FormikValues> & EditApplicationF
       <PageHeading title={createFlowType} />
       <FlexForm onSubmit={handleSubmit}>
         <FormBody flexLayout>
-          {createFlowType !== CreateApplicationFlow.Container &&
-            createFlowType !== CreateApplicationFlow.JarUpload && (
+          {createFlowType !== i18n.t(EditFlowPageTitle[CreateApplicationFlow.Container]) &&
+            createFlowType !== i18n.t(EditFlowPageTitle[CreateApplicationFlow.JarUpload]) && (
```
andrewballantyne, line 48: Again, this is a component, favour the hook `t`

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -77,12 +89,12 @@ export const getPageHeading = (buildStrategy: string, buildType?: string): strin
...
-        ? CreateApplicationFlow.JarUpload
-        : CreateApplicationFlow.Git;
+        ? i18n.t(EditFlowPageTitle[CreateApplicationFlow.JarUpload])
+        : i18n.t(EditFlowPageTitle[CreateApplicationFlow.Git]);
     case BuildStrategyType.Docker:
-      return CreateApplicationFlow.Dockerfile;
+      return i18n.t(EditFlowPageTitle[CreateApplicationFlow.Dockerfile]);
     default:
-      return CreateApplicationFlow.Container;
+      return i18n.t(EditFlowPageTitle[CreateApplicationFlow.Container]);
```
jerolimov, line 97: I think the problem starts here with the name `getPageHeading`. This util method should still return an enum.

The translation should happen when rendering this data. That the value of the enum contains already the heading doesn't make this straight forward, but we should keep this for the moment.

(btw: IMHO the return type of this function should be `CreateApplicationFlow` instead of a string.)

```tsx
export const getApplicationFlow = (
  buildStrategy: string,
  buildType?: string,
): CreateApplicationFlow => {
  switch (buildStrategy) {
    case BuildStrategyType.Source:
      return buildType === BuildSourceType.Binary
        ? CreateApplicationFlow.JarUpload
        : CreateApplicationFlow.Git;
    case BuildStrategyType.Docker:
      return CreateApplicationFlow.Dockerfile;
    default:
      return CreateApplicationFlow.Container;
  }
};
```

frontend/packages/dev-console/src/components/edit-application/EditApplication.tsx
jerolimov: We should avoid comparing titles to enum values and translate the enum values here already. We should translate only values which we render to the user.

If your getPageHeading (think of it as "getApplicationFlow" util function) returns a typed enum it's more clear that this checks should stay at they are before.

(IMHO: To make this more clear you can check `pageHeading` here to `applicationFlow` and remove the i18n.t calls here completly.)

```tsx
const applicationFlow = getApplicationFlow(buildStrategy, buildSourceType);
```

frontend/packages/dev-console/src/components/edit-application/EditApplication.tsx
```diff
@@ -124,8 +126,8 @@ const EditApplication: React.FC<EditApplicationProps> = ({
     };
 
     if (
-      pageHeading === CreateApplicationFlow.Git ||
-      pageHeading === CreateApplicationFlow.JarUpload
+      pageHeading === i18n.t(EditFlowPageTitle[CreateApplicationFlow.Git]) ||
+      pageHeading === i18n.t(EditFlowPageTitle[CreateApplicationFlow.JarUpload])
```
jerolimov, line 130: See above. Imho not necessary.

frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
jerolimov: +1. `useTranslation` is already used here.

But IMHO we should also not check this enum type against translated values. Also if the enum values are the actual english strings we should consider them as "keys". I would recommend to reverse this changes here as well.

If also worth to change the Props definition from `createFlowType: string` to `createFlowType: CreateApplicationFlow` to make this more clear.

frontend/packages/dev-console/src/components/edit-application/EditApplicationForm.tsx
```diff
@@ -43,36 +44,43 @@ const EditApplicationForm: React.FC<FormikProps<FormikValues> & EditApplicationF
       <PageHeading title={createFlowType} />
```
jerolimov, line 44: It should be enough to translate the title here? Wdyt?
```suggestion
      <PageHeading title={t(EditFlowPageTitle[createFlowType])} />
```

jerolimov (changes requested):
Please take a look into my recommendation above. I would like if we don't call the translation so often and just translate the value before rendering `<PageHeading ..`

Wdyt?

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
debsmita1: changed it to a util

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
debsmita1: Fixed it !

frontend/packages/dev-console/src/components/edit-application/EditApplication.tsx
```diff
@@ -137,15 +137,15 @@ const EditApplication: React.FC<EditApplicationProps> = ({
...
+    editFlowType,
   ]);
 
   const renderForm = (formikProps: FormikProps<any>) => (
     <EditApplicationForm
       {...formikProps}
       appResources={appResources}
       enableReinitialize
-      createFlowType={pageHeading}
+      createFlowType={editFlowType}
```
jerolimov, line 148: Maybe we should remove the create and edit prefix? createFlowType=editFlowType looks strange or? I don't know how big the change is. The actual version is also fine for me.

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -50,6 +50,25 @@ export enum CreateApplicationFlow {
   JarUpload = 'Upload JAR file',
 }
 
+export const getEditFlowPageTitle = (editflowtype: CreateApplicationFlow): string => {
```
jerolimov, line 53: If you want change this, getFlowTypeTitle here. Up to you

frontend/packages/dev-console/src/components/edit-application/edit-application-utils.ts
```diff
@@ -50,6 +50,25 @@ export enum CreateApplicationFlow {
...
+      // t('devconsole~Import from Dockerfile')
+      return 'devconsole~Import from Dockerfile';
+    case CreateApplicationFlow.Container:
+      // t('devconsole~Deploy Image')
+      return 'devconsole~Deploy Image';
+    case CreateApplicationFlow.JarUpload:
+      // t('devconsole~Upload JAR file')
+      return 'devconsole~Upload JAR file';
+    default:
+      return '';
```
jerolimov, line 68: We should return the flowType here as fallback. If we add an enum value and don't add the translation we should at least show the not translated version.

jerolimov: For all values it's also possible to return "devconsole~" + flowType.
You can return this always this way or what's fine, just return the default value this way. Both is fine for me. If you keep the switch I see the benefit that this code doesn't expect anymore that the enum value is also the title. 👍

jerolimov (commented):
Small nits / ideas.

jerolimov (approved):
Thanks for cleaning this up. :+1:

Tested it locally and works fine. Tested create and edit flows.

/lgtm
/approved

PR #9215, opened 2021-06-11
Bug 1908772: Fix a11y violation: Dev Console Nav Menu UL contains non-LI elements

jerolimov:
**Fixes**: 
https://bugzilla.redhat.com/show_bug.cgi?id=1908772

**Analysis / Root cause**: 
Dev Console Nav Menu UL contains non-LI elements

**Solution Description**: 
Add li elements.

UI looks like before!

Lists on both levels have now only li elements as child of the ul elements.

https://user-images.githubusercontent.com/139310/121612069-f8ccbf00-ca59-11eb-966a-768c2d89019c.mp4

Untouched

**Test setup:**
* Open developer perspective
* Check rendered elements in browser
* run `yarn test-cypress-devconsole` `create-from-git.feature` successfully

image

* run `yarn test-cypress-console` and run `namespace-crud.spec.ts` successfully

image

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/nav/_perspective-nav.scss
+++ b/frontend/public/components/nav/_perspective-nav.scss
@@ -1,16 +1,15 @@
-.oc-nav-group.pf-c-nav__section {
-  margin-top: 0;
-  .pf-c-nav__section-title {
-    padding-bottom: 0;
-  }
-  &:first-of-type {
+.oc-perspective-nav {
+  .pf-c-nav__section:first-child {
     .pf-c-nav__section-title {
       padding-top: 0;
       border: none;
     }
   }
-  > ul {
-    padding-left: 0;
+  .pf-c-nav__section {
+    margin-top: 0;
+    .pf-c-nav__section-title {
+      padding-bottom: 0;
+    }
   }
 }
 
--- a/frontend/public/components/nav/perspective-nav.tsx
+++ b/frontend/public/components/nav/perspective-nav.tsx
@@ -61,7 +61,7 @@ const PerspectiveNav: React.FC<{}> = () => {
     return '';
   };
 
-  const getPinnedItems = (rootNavLink: boolean = false): React.ReactElement[] =>
+  const getPinnedItems = (): React.ReactElement[] =>
     pinnedResourcesLoaded
       ? pinnedResources
           .map((resource) => {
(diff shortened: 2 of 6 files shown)


frontend/public/components/nav/_perspective-nav.scss
```diff
@@ -1,17 +1,21 @@
-.oc-nav-group.pf-c-nav__section {
-  margin-top: 0;
-  .pf-c-nav__section-title {
-    padding-bottom: 0;
+.pf-c-nav__list {
+  > li {
```
rhamilto, line 2: While you're fixing the markup semantics, it would also be great if we could move away from descendent selectors like these and do a better job of sticking with [BEM conventions](http://getbem.com/introduction/) since descendent selectors like these are explicitly tied to the markup structure, and these changes are making the existing rules even *more* explicit.

It would be better to add classes to the individual elements and target those new classes with overrides.  The existing `oc-nav-group` class and rule is one such example (although ideally it would *not* be paired with the PatternFly class because that makes the rule is dependent on the PatternFly class being present).

@sg00dwin or I can assist if needed.

jerolimov: I tried to be carefully here because of the two levels of navigation items here and tried a minimal-save-change.

The actual tree is something like `ul > li > section > ul > li`

I can take a second look into the dom structure and try to apply your mentioned BEM conventions.

frontend/public/components/nav/perspective-nav.tsx
```diff
@@ -101,17 +101,13 @@ const PerspectiveNav: React.FC<{}> = () => {
               >
                 {removeButton}
               </RootNavLink>
-            ) : (
-              <Component key={resource} className="oc-nav-pinned-item" {...props}>
-                {removeButton}
-              </Component>
```
jerolimov, line 107: rootNavLink is always true!

frontend/public/components/nav/_perspective-nav.scss
jerolimov: @rhamilto Replaced the `>li` with PatternFly components. Cleaned up the code a little bit. If you find the time, please take a new look and let me know what you think.

rhamilto: Nice.  Definitely a step forward instead of a step back.  Good work.

PR #9217, opened 2021-06-11
Fix a11y issue in virtualized tables with sortable headers and redesigned add page

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5119

**Analysis / Root cause**: 
1. When rendering a virtualized table with sortable headers, these generates some a11y issues. Could verify this with the axe devtools in Firefox as well with the accessibility debug tools in Chrome.
2. The popover icon in the title of the redesigned add page should have a role when defined an `aria-label` for the icon.

Firefox Axe DevTools before:
image

Chrome DevTools before:
image

**Solution Description**: 
1. Force `rowgroup` `role` for `thead`for virtualized tables. This is the default for a `thead` when the table was not forced to `presentation`. This allows the aria-sort attribute on the th cell.
2. Added role="button" to the popover icon on the add page.

Firefox Axe DevTools after:
image

Chrome DevTools after:
image

Other issues are not addressed in this issue. Some navigation issues are addressed in #9215

UI doesn't changed.

Untouched

**Test setup:**
* Open add page and select "App projects"
* Run Axe DevTools

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/factory/table.tsx
+++ b/frontend/public/components/factory/table.tsx
@@ -792,7 +792,7 @@ export const Table = connect<
               role={virtualize ? 'presentation' : 'grid'}
               aria-label={virtualize ? null : ariaLabel}
             >
-              <TableHeader />
+              <TableHeader role="rowgroup" />
               {!virtualize && <TableBody />}
             </PfTable>
             {virtualize &&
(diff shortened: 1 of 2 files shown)


frontend/public/components/factory/table.tsx
```diff
@@ -792,7 +792,7 @@ export const Table = connect<
               role={virtualize ? 'presentation' : 'grid'}
```
jerolimov, line 792: Because of this presentation we should force rowgroup for the header.

See also https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role

jerolimov:
> should deploy the image and display it in the topology 

/retest

rohitkrai03 (approved):
/lgtm

dtaylor113 (approved):

PR #9219, opened 2021-06-11
adds feature flag for revision decorator plugin

invincibleJai:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5956

**Solution Description**: 
adds feature flag for revision decorator plugin and check if the resource is a revision in decorator provider

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified locally, works fine
/lgtm

PR #9225, opened 2021-06-11
(fix): Hide RGW storage class in add capacity modal

afreen23:
Added rook bucket provisioner in the blacklist


bipuladh (approved):
/lgtm

PR #9226, opened 2021-06-11
Bug 1970980: Remove usage of i18nKey in Trans components

rebeccaalpert:
Some Trans components were set up to take an i18nKey prop. This key overrides the generated key for the component and prevents us from passing the correct messages on for translation. I removed the i18nKey prop and regenerated the keys so we can get them translated correctly.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1970980.


spadgett:
@rebeccaalpert We should make sure this is in our i18n doc and eventually the README when we add it to the repo.

spadgett:
/lgtm cancel

> Error: Failed: Missing i18n key "Optionally list teams. If specified, only GitHub users that are members of at least one of the listed teams will be allowed to log in. Cannot be used in combination with <strong>organizations</strong>." in namespace "public" and language "en."

rebeccaalpert:
I looked into this and the parser isn't correctly handling all special keys in `Trans` components. See https://github.com/i18next/i18next-parser/issues/264. For the time being, I kept one of the `i18nKeys`, but changed it to match the text value and turned on `transSupportBasicHtmlNodes` in `react-i18next` so we can use our existing toolchain. Longer term, we should invest in contributing back to the parser if we want the root cause fixed.

frontend/public/components/cluster-settings/github-idp-form.tsx
```diff
@@ -221,7 +221,11 @@ class AddGitHubPageWithTranslation extends PromiseComponent<
           <div className="co-form-section__separator" />
           <h3>{t('public~Organizations')}</h3>
           <p className="co-help-text">
-            <Trans i18nKey="teams-help" ns="public">
+            <Trans
+              t={t}
+              ns="public"
+              i18nKey="Optionally list organizations. If specified, only GitHub users that are members of at least one of the listed organizations will be allowed to log in. Cannot be used in combination with <strong>teams</strong>."
```
jhadvig, line 227: Why this needed to stay? Whats the special kay that its not able to handle?

jhadvig (commented):
Change looks good 👍 
Added one question to clear out the special cases.

frontend/public/components/cluster-settings/github-idp-form.tsx
rebeccaalpert: The parser wasn't correctly generating the <strong> key. This does an end run around that issue since we're not using the parser to generate the keys in this case. i18next can handle strong just fine, but apparently sometimes it makes the parser choke. I used the react-i18next default Codepen to check to make sure i18next was ok and narrow down the issue to the parser.

jhadvig: And what about other appearances of `<strong>` tag?, eg. `frontend/public/components/image-stream.tsx` ? is that safe to use there

jhadvig (approved):
/lgtm

PR #9230, opened 2021-06-14
Expose Internal APIs

bipuladh:
Exposes the following Components:

 **Hooks**
 - `usePrometheusPoll`

**Dashboard Related Components**
 - `DashboardGrid`
 - `HealthItem`
 - `AlertItem`
 - `AlertsBody`
 - `PrometheusUtilizationItem`
 - `ResourceInventoryItem`
 - `UtilizationBody`
 - `RecentEventsBody`
 - `OngoingActivityBody`
 - `DashboardCard`
 - `DashboardCardBody`
 - `DashboardCardHeader`
 - `DashboardCardTitle`
 - `DetailsItem`
 - `DetailsBody`
 - `UtilizationItem`
 - `ActivityItem`
 - `ActivityBody`

### DIFF
--- a/frontend/packages/console-dynamic-plugin-sdk/src/extensions/console-types.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/extensions/console-types.ts
@@ -1,4 +1,4 @@
-import { ExtensionK8sGroupKindModel } from '../api/common-types';
+import { ExtensionK8sGroupKindModel, PrometheusLabels, PrometheusValue } from '../api/common-types';
 
 export type OwnerReference = {
   name: string;
@@ -121,11 +121,9 @@ export enum HealthState {
   UNKNOWN = 'UNKNOWN',
   UPDATING = 'UPDATING',
   PROGRESS = 'PROGRESS',
+  UPGRADABLE = 'UPGRADABLE',
   NOT_AVAILABLE = 'NOT_AVAILABLE',
 }
-
-export type PrometheusLabels = { [key: string]: string };
-export type PrometheusValue = [number, string];
 
 // Only covers range and instant vector responses for now.
 export type PrometheusResult = {
--- a/frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
@@ -7,4 +7,48 @@ export const exposePluginAPI = () => {
     useK8sWatchResources: require('@console/internal/components/utils/k8s-watch-hook')
       .useK8sWatchResources,
   };
+  window.internalAPI = {
+    AcitivityItem: require('@console/shared/src/components/dashboard/activity-card/ActivityItem')
+      .default,
+    ActivityBody: require('@console/shared/src/components/dashboard/activity-card/ActivityBody')
+      .default,
+    RecentEventsBody: require('@console/shared/src/components/dashboard/activity-card/ActivityBody')
+      .RecentEventsBody,
+    OngoingActivityBody: require('@console/shared/src/components/dashboard/activity-card/ActivityBody')
+      .OngoingActivityBody,
+
+    AlertsBody: require('@console/shared/src/components/dashboard/status-card/AlertsBody').default,
+    AlertItem: require('@console/shared/src/components/dashboard/status-card/AlertItem').default,
+    HealthItem: require('@console/shared/src/components/dashboard/status-card/HealthItem').default,
+    HealthBody: require('@console/shared/src/components/dashboard/status-card/HealthBody').default,
+
+    DashboardCard: require('@console/shared/src/components/dashboard/dashboard-card/DashboardCard')
+      .default,
+    DashboardCardBody: require('@console/shared/src/components/dashboard/dashboard-card/DashboardCardBody')
+      .default,
+    DashboardCardHeader: require('@console/shared/src/components/dashboard/dashboard-card/DashboardCardHeader')
+      .default,
+    DashboardCardTitle: require('@console/shared/src/components/dashboard/dashboard-card/DashboardCardTitle')
+      .default,
+    DashboardGrid: require('@console/shared/src/components/dashboard//DashboardGrid').default,
+
+    ResourceInventoryItem: require('@console/shared/src/components/dashboard/inventory-card/InventoryItem')
+      .ResourceInventoryItem,
+
+    DetailItem: require('@console/shared/src/components/dashboard/details-card/DetailItem').default,
+    DetailsBody: require('@console/shared/src/components/dashboard/details-card/DetailsBody')
+      .default,
+
+    UtilizationItem: require('@console/shared/src/components/dashboard/utilization-card/UtilizationItem')
+      .default,
+    UtilizationBody: require('@console/shared/src/components/dashboard/utilization-card/UtilizationBody')
+      .default,
+    useUtilizationDuration: require('@console/shared/src/hooks/useUtilizationDuration')
+      .useUtilizationDuration,
+    UtilizationDurationDropdown: require('@console/shared/src/components/dashboard/utilization-card/UtilizationDurationDropdown')
+      .UtilizationDurationDropdown,
+
+    usePrometheusPoll: require('@console/internal/components/graphs/prometheus-poll-hook')
+      .usePrometheusPoll,
+  };
 };
--- a/frontend/@types/console/index.d.ts
+++ b/frontend/@types/console/index.d.ts
@@ -55,6 +55,7 @@ declare interface Window {
   loadPluginEntry?: Function;
   Cypress?: {};
   api: {}; // Console dynamic plugin APIs
+  internalAPI: {};
 }
 
 // TODO: Remove when upgrading to TypeScript 4.1.2+, which has a type for RelativeTimeFormat.
--- a/frontend/packages/console-dynamic-plugin-sdk/scripts/generate-pkg-assets.ts
+++ b/frontend/packages/console-dynamic-plugin-sdk/scripts/generate-pkg-assets.ts
@@ -14,6 +14,7 @@ const createPackageJson = (packagePath: string) => {
     '.': './lib/extensions/index.js',
     './webpack': './lib/webpack/ConsoleRemotePlugin.js',
     './api': './lib/api/api.js',
+    './internalAPI': './lib/api/internal-api.js',
   };
   packageJson.readme = 'README.md';
   packageJson.peerDependencies = _.pick(packageJson.devDependencies, 'webpack');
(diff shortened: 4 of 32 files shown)


rawagner:
I have a PRs which are adding types to ListPage and Table and also another ones which are updating the APIs of the components so its easier for consumer to use them.

types for list page
https://github.com/openshift/console/pull/8792
new list page API
https://github.com/openshift/console/pull/8827

types for Table
https://github.com/openshift/console/pull/8756
new Table API
https://github.com/openshift/console/pull/8829

frontend/packages/console-dynamic-plugin-sdk/src/api/api.ts
```diff
@@ -8,3 +8,8 @@ const MockImpl = () => {
 
 export const useK8sWatchResource: UseK8sWatchResource = MockImpl;
 export const useK8sWatchResources: UseK8sWatchResources = MockImpl;
+export const usePrometheusQuery: UsePrometheusQuery = MockImpl;
```
rawagner, line 11: The name should reflect that it is for dashboards only

frontend/packages/console-dynamic-plugin-sdk/src/api/dashboard/dashboard-api.ts
```diff
@@ -0,0 +1,19 @@
...
+  dashboardGrid,
+  recentEventsBody,
+  prometheusUtilizationItem,
+} from './dashboard-types';
+
+const MockImpl = () => {
+  throw new Error('You need to configure webpack externals to use this function at runtime.');
+};
+
+export const HealthItem: healthItem = MockImpl;
```
rawagner, line 14: the types should start with uppercase

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
```diff
@@ -7,4 +8,5 @@ export const exposePluginAPI = () => {
     useK8sWatchResources: require('@console/internal/components/utils/k8s-watch-hook')
       .useK8sWatchResources,
   };
+  Object.assign(window.api, exposedAPIs);
```
bipuladh, line 11: I din't make a new object for it. Should I add it to `window.provisional` instead?

frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -174,3 +174,42 @@ via `SERVER_FLAGS` object.
   be enabled or disabled separately.
 - Failure to resolve a code reference (unable to load module, missing module export etc.) will disable
   the plugin.
+
+### Component APIs
+
+There are currently two API packages exposed as part of the dynamic plugin SDK. The Stable API package and Internal API packages.
+Stable APIs have been vetted and are unlikely to change in terms of signature.
```
christianvogt, line 181: Stable APIs cannot change without proper version update

frontend/packages/console-dynamic-plugin-sdk/README.md
```diff
@@ -174,3 +174,42 @@ via `SERVER_FLAGS` object.
...
+- `DashboardGrid`
+- `ResourceInventoryItem`
+- `DetailsItem`
+- `DetailsBody`
+- `PageHeading`
+- `HorizontalNav`
+- `FieldLevelHelp`
+- `ListPage`
+- `useMetricDuration`
+- `Duration`
```
christianvogt, line 215: I really don't want to have to keep in sync such an API list with what we expose. 
We can move to generating documentation.

frontend/packages/console-dynamic-plugin-sdk/src/extensions/console-types.ts
```diff
@@ -208,3 +208,56 @@ export type FirehoseResult<
 export type FirehoseResourcesResult = {
   [key: string]: FirehoseResult<K8sResourceCommon | K8sResourceCommon[]>;
 };
+
+export type PrometheusAlert = {
```
christianvogt, line 212: All these new types in `console-types` don't seem to be used? Remove them if part of another PR.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
```diff
@@ -7,4 +7,53 @@ export const exposePluginAPI = () => {
     useK8sWatchResources: require('@console/internal/components/utils/k8s-watch-hook')
       .useK8sWatchResources,
   };
+  window.internalAPI = {
```
christianvogt, line 10: Oh so ugly. @rawagner @vojtechszocs are we planning to move away from global variables in 4.9?

frontend/packages/console-dynamic-plugin-sdk/src/extensions/console-types.ts
bipuladh: This is being used in a different file. All these are introduced to support one type `Alert`. I am moving them to the file where it's being used and also removing `export` from it.

frontend/packages/console-dynamic-plugin-sdk/README.md
bipuladh: Removing doc for now. Will send a different PR which will generate docs based on the exposed APIs.

rawagner:
LGTM

spadgett:
I'm not sure how much putting these under `internalAPI` buys us. If we have plugins using them, changing the API will still break the plugin. Have we thought through upgrade strategies for plugins using internal APIs? How will be promote internal APIs to public APIs? It feels like we should update the dynamic plugins enhancement if we want to introduce an internal API concept.

I don't think we want to put `PageHeading` and `HorizontalNav` under `internalAPI` since we'll need these for kubevirt in 4.9 as part of the SDK. I feel like we should look at cleaning up and simplifying these APIs. I'm not sure we want to expose them as-is.

Can we consider having only one hook for polling Prometheus (and always storing the data in Redux if that's needed)?

spadgett:
It looks like there are more components exposed in the PR than listed in the descriptions. Were these dependencies needed?

@bipuladh @rawagner How much of the basic dashboard layout like `DashboardGrid` and `DashboardCard` can be accomplished with PatternFly components? `DashboardCard` for instance appears to wrap the PatternFly `Card` and only adds one or two CSS classes.

christianvogt:
My original idea behind internals was to expose extensions and their types as internal API  while the extension API got fleshed out. I wasn't expecting the need to expose other components as part of internal that weren't explicitly part of the internal extension itself. This I thought would be a good way to help migrate the extension for use while the extension API got evaluated.

spadgett:
Yeah, I understand the need. We should work out the process of promoting these to public APIs. We'd need at least one release supporting both the old and new interfaces to avoid breaking plugins during upgrades, and we'd need to use an adapter or perhaps write some of the new components greenfield like we're doing for tables.

bipuladh:
> It looks like there are more components exposed in the PR than listed in the descriptions. Were these needed because they are dependencies of the components we're exposing?
> 
> @bipuladh @rawagner How much of the basic dashboard layout like `DashboardGrid` and `DashboardCard` can be accomplished with PatternFly components? `DashboardCard` for instance appears to wrap the PatternFly `Card` and only adds one or two CSS classes.

IMO all of it can be accomplished from PF. However, exposing these simple components should allow us to maintain similarities across dashboards and changing CSS in the core console would allow us to revamp all the dashboards(wouldn't have to wait for two releases) without changing plugin code.

bipuladh:
> I don't think we want to put `PageHeading` and `HorizontalNav` under `internalAPI` since we'll need these for kubevirt in 4.9 as part of the SDK. I feel like we should look at cleaning up and simplifying these APIs. I'm not sure we want to expose them as-is.
> 
I have created a wrapper for PageHeading https://github.com/openshift/console/pull/9465 based on the commonly used props. Apparently only a few of the defined are commonly used. 
I thought we would first expose components as part of the InternalAPI first clean them up and then move them out to the Stable SDK. 
> Can we consider having only one hook for polling Prometheus (and always storing the data in Redux if that's needed)?
`usePrometheusQuery` is tied close to the dashboard. Maybe we could add a prop `isDashboard` which would trigger Dashboard related Actions else it could work exactly like `usePrometheusPoll` or just use it internally....?

rawagner:
> > Can we consider having only one hook for polling Prometheus (and always storing the data in Redux if that's needed)?
> 
> `usePrometheusQuery` is tied close to the dashboard. Maybe we could add a prop `isDashboard` which would trigger Dashboard related Actions else it could work exactly like `usePrometheusPoll` or just use it internally....?

+1 for having one prometheus hook and storing data to redux. I wanted to do it for some time but cannot find the time

rawagner:
> > It looks like there are more components exposed in the PR than listed in the descriptions. Were these needed because they are dependencies of the components we're exposing?
> > @bipuladh @rawagner How much of the basic dashboard layout like `DashboardGrid` and `DashboardCard` can be accomplished with PatternFly components? `DashboardCard` for instance appears to wrap the PatternFly `Card` and only adds one or two CSS classes.
> 
> IMO all of it can be accomplished from PF. However, exposing these simple components should allow us to maintain similarities across dashboards and changing CSS in the core console would allow us to revamp all the dashboards(wouldn't have to wait for two releases) without changing plugin code.

As @bipuladh said. That is the idea behind these simple components.

frontend/packages/console-dynamic-plugin-sdk/src/runtime/plugin-api.ts
vojtechszocs: @christianvogt Yes, this should be done as part of #9432

christianvogt:
Changes lgtm

We should be working towards promoting these extensions to proper APIs.
I don't want to get into having to maintain backwards compatibility for these APIs. Something we touched on in our meetings but need to further discuss.

SanjalKatiyar:
# **Have tested the PR for the following cases:**

- ## **Components used by ODF Dashboard:**

### `usePrometheusPoll`
Screenshot from 2021-07-28 15-09-28
Screenshot from 2021-07-28 15-09-50
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
### `HealthItem`
Screenshot from 2021-07-28 15-08-53
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
### `RecentEventsBody`
Screenshot from 2021-07-28 15-09-06
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
### `DashboardCard, DashboardCardBody, DashboardCardHeader, DashboardCardTitle`
Screenshot from 2021-07-28 18-26-05
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------

- ## **Rest of console dashboards working fine with these changes:**

Screenshot from 2021-07-28 17-49-06
Screenshot from 2021-07-28 18-01-21
Screenshot from 2021-07-28 18-05-16

spadgett:
@yapei has given us the OK for development to test the PR.

/label qe-approved

This PR has no user visible changes and has no doc impact. These APIs are internal and can change at any time.

/label px-approved
/label docs-approved

@ahardin-rh @sferich888 FYI

/hold cancel

PR #9231, opened 2021-06-14
Bug 1970332: Fix page disappearing issue while creating sc for rbd provisioner

SanjalKatiyar:


frontend/packages/ceph-storage-plugin/src/components/kms-config/kms-config.tsx
```diff
@@ -63,7 +63,8 @@ export const KMSConfigure: React.FC<KMSConfigureProps> = ({ state, dispatch, mod
 
   const setAddressPort = (port: string) => {
     kmsObj.port.value = port;
-    kmsObj.port.valid = port !== '' && !_.isNaN(Number(port)) && Number(port) > 0;
+    kmsObj.port.valid =
```
GowthamShanmugam, line 66: Is patternfly is providing any port component?

SanjalKatiyar: I checked patternfly only supports following types for `<TextInput>`:
```
"number" | "time" | "text" | "url" | "date" | "datetime-local" | "email" | "month" | "password" | "search" | "tel"
```

frontend/public/components/storage-class-form.tsx
```diff
@@ -708,7 +708,7 @@ class StorageClassFormWithTranslation extends React.Component<
     const dataParameters = _.pickBy(
       _.mapValues(this.state.newStorageClass.parameters, (value, key) => {
         let finalValue = value.value;
-        if (this.storageTypes[type].parameters[key].format) {
+        if (this.storageTypes[type].parameters[key]?.format) {
```
SanjalKatiyar, line 711: Exact steps are little harder to reproduce, but, this line was causing `Cannot read property 'format' of undefined` error for certain cases.
Anyways, it is safer to use `?.` than not adding it at all.

bipuladh: @rawagner PTAL

PR #9233, opened 2021-06-14
Bug 1967658: improve failure alert for copied CSV

rhamilto:
After:
If CSV is copied:
Screen Shot 2021-06-14 at 10 28 17 AM
Otherwise:
Screen Shot 2021-06-14 at 11 45 32 AM

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
@@ -87,6 +87,7 @@ import {
   ClusterServiceVersionKind,
   ClusterServiceVersionPhase,
   CRDDescription,
+  CSVConditionReason,
   InstallPlanKind,
   PackageManifestKind,
   SubscriptionKind,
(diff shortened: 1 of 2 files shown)


frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
```diff
@@ -987,8 +988,27 @@ export const ClusterServiceVersionDetails: React.FC<ClusterServiceVersionDetails
                   isInline
                   className="co-alert"
                   variant="danger"
-                  title={`${status.phase}: ${status.message}`}
-                />
+                  title={
+                    status.reason === CSVConditionReason.CSVReasonCopied
+                      ? t('olm~Operator failed')
+                      : `${status.phase}: ${status.message}`
```
spadgett, line 994: I would always show "Operator failed" here and then show `status.message` in the alert body below when reason isn't copied.

spadgett (commented):
Thanks @rhamilto

frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx
rhamilto: Thanks, @spadgett.  Updated.

spadgett (approved):
/lgtm

👍

PR #9234, opened 2021-06-14
tablecolumnclass updates public plugin PR 4

abhinandan13jan:
# Addresses 
https://issues.redhat.com/browse/ODC-5816

# Screenshot
tableclass4

# Test
No changes

# Browser conformance
Chrome

### DIFF
--- a/frontend/public/components/machine-autoscaler.tsx
+++ b/frontend/public/components/machine-autoscaler.tsx
@@ -41,11 +41,11 @@ const MachineAutoscalerTargetLink: React.FC<MachineAutoscalerTargetLinkProps> = 
 };
 
 const tableColumnClasses = [
-  classNames('col-md-4', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
+  'pf-u-w-33-on-lg',
+  '',
+  'pf-m-hidden pf-m-visible-on-md',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
   Kebab.columnClass,
 ];
 
--- a/frontend/public/components/machine-config-pool.tsx
+++ b/frontend/public/components/machine-config-pool.tsx
@@ -246,12 +246,12 @@ export const MachineConfigPoolDetailsPage: React.SFC<any> = (props) => (
 );
 
 const tableColumnClasses = [
-  classNames('col-lg-3', 'col-xs-6', 'col-sm-4'),
-  classNames('col-lg-5', 'hidden-md', 'hidden-sm', 'hidden-xs'),
-  classNames('col-lg-1', 'col-sm-2', 'col-xs-3', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'col-sm-2', 'hidden-xs', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'hidden-sm', 'hidden-xs', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'col-sm-2', 'col-xs-3', 'pf-u-w-10-on-lg'),
+  'pf-u-w-25-on-lg pf-u-w-33-on-md pf-u-w-50-on-sm',
+  'pf-m-hidden pf-m-visible-on-xl pf-u-w-50-on-xl',
+  'pf-u-w-25-on-sm pf-u-w-16-on-md pf-u-w-8-on-lg',
+  'pf-m-hidden pf-m-visible-on-md pf-u-w-16-on-md pf-u-w-8-on-lg',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-8-on-lg',
+  'pf-u-w-25-on-lg pf-u-w-33-on-sm pf-u-w-8-on-lg',
   Kebab.columnClass,
 ];
 
--- a/frontend/public/components/resource-quota.jsx
+++ b/frontend/public/components/resource-quota.jsx
@@ -72,11 +72,7 @@ const getResourceUsage = (quota, resourceType) => {
   };
 };
 
-const tableColumnClasses = [
-  classNames('col-md-5', 'col-xs-6'),
-  classNames('col-md-7', 'col-xs-6'),
-  Kebab.columnClass,
-];
+const tableColumnClasses = ['', '', Kebab.columnClass];
 
 export const UsageIcon = ({ percent }) => {
   let usageIcon = <UnknownIcon />;
--- a/frontend/public/components/service-account.jsx
+++ b/frontend/public/components/service-account.jsx
@@ -110,10 +110,10 @@ const menuActions = [
 const kind = 'ServiceAccount';
 
 const tableColumnClasses = [
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-md-2', 'hidden-sm', 'hidden-xs'),
-  classNames('col-sm-2', 'hidden-xs'),
+  '',
+  '',
+  'pf-m-hidden pf-m-visible-on-lg',
+  'pf-m-hidden pf-m-visible-on-md',
   Kebab.columnClass,
 ];
 
--- a/frontend/public/components/RBAC/bindings.jsx
+++ b/frontend/public/components/RBAC/bindings.jsx
@@ -132,11 +132,11 @@ const menuActions = ({ subjectIndex, subjects }, startImpersonate) => {
 };
 
 const tableColumnClasses = [
-  classNames('col-md-3', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-lg-2', 'col-md-3', 'hidden-sm', 'hidden-xs'),
-  classNames('col-lg-2', 'hidden-md', 'hidden-sm', 'hidden-xs'),
-  classNames('col-lg-2', 'col-md-3', 'col-sm-4', 'col-xs-6'),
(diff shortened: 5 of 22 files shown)


frontend/public/components/machine-autoscaler.tsx
```diff
@@ -41,11 +41,11 @@ const MachineAutoscalerTargetLink: React.FC<MachineAutoscalerTargetLinkProps> =
...
-  classNames('col-md-4', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
+  'pf-u-w-33-on-lg',
+  '',
+  'pf-m-hidden pf-m-visible-on-md',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-8-on-lg',
+  'pf-m-hidden pf-m-visible-on-md pf-u-w-8-on-lg',
```
karthikjeeyar, line 48:
```suggestion
  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
  'pf-m-hidden pf-m-visible-on-md pf-u-w-16-on-lg',
```
with `pf-u-w-8-on-lg` :
image

after changing to `pf-u-w-16-on-lg`:
image

frontend/public/components/machine-config-pool.tsx
```diff
@@ -246,12 +246,12 @@ export const MachineConfigPoolDetailsPage: React.SFC<any> = (props) => (
...
 
 const tableColumnClasses = [
-  classNames('col-lg-3', 'col-xs-6', 'col-sm-4'),
-  classNames('col-lg-5', 'hidden-md', 'hidden-sm', 'hidden-xs'),
-  classNames('col-lg-1', 'col-sm-2', 'col-xs-3', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'col-sm-2', 'hidden-xs', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'hidden-sm', 'hidden-xs', 'pf-u-w-10-on-lg'),
-  classNames('col-lg-1', 'col-sm-2', 'col-xs-3', 'pf-u-w-10-on-lg'),
+  classNames('pf-u-w-25-on-lg pf-u-w-33-on-md pf-u-w-50-on-sm'),
+  classNames('pf-m-hidden pf-m-visible-on-xl pf-u-w-40-on-xl'),
```
karthikjeeyar, line 250:
```suggestion
  classNames('pf-m-hidden pf-m-visible-on-xl pf-u-w-50-on-xl'),
```
with `pf-u-w-40-on-xl`

image

with `pf-u-w-50-on-xl`:
image

karthikjeeyar: You could also remove `classNames` from this file.

frontend/public/components/resource-quota.jsx
```diff
@@ -72,11 +72,7 @@ const getResourceUsage = (quota, resourceType) => {
   };
 };
 
-const tableColumnClasses = [
-  classNames('col-md-5', 'col-xs-6'),
-  classNames('col-md-7', 'col-xs-6'),
-  Kebab.columnClass,
-];
+const tableColumnClasses = ['pf-u-w-42-on-lg', 'pf-u-w-58-on-lg', Kebab.columnClass];
```
karthikjeeyar, line 75: above change works well for `All Projects`, but not if you are in a namespace

image

karthikjeeyar (commented):
@abhinandan13jan  The PF classes that are added works well with `All Projects` being selected in the project dropdown, which will add `Namespace` column as the second row.  when you select a project then `Namespace` column will be hidden, then the classes doesn't work in some pages.

frontend/public/components/machine-autoscaler.tsx
abhinandan13jan: updated

frontend/public/components/machine-config-pool.tsx
abhinandan13jan: updated

frontend/public/components/resource-quota.jsx
abhinandan13jan: yes, I changed accordingly

Screenshot from 2021-06-21 16-31-51
Screenshot from 2021-06-21 16-32-02

frontend/public/components/machine-autoscaler.tsx
```diff
@@ -41,11 +41,11 @@ const MachineAutoscalerTargetLink: React.FC<MachineAutoscalerTargetLinkProps> =
...
-  classNames('col-md-4', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'col-xs-6'),
-  classNames('col-md-3', 'col-sm-4', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
-  classNames('col-md-1', 'hidden-sm', 'hidden-xs'),
+  'pf-u-w-33-on-lg',
+  '',
+  'pf-m-hidden pf-m-visible-on-md',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
+  'pf-m-hidden pf-m-visible-on-md pf-u-w-16-on-lg',
```
karthikjeeyar, line 48:
```suggestion
  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
```

frontend/public/components/service-account.jsx
```diff
@@ -110,10 +110,10 @@ const menuActions = [
...
 
 const tableColumnClasses = [
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-md-2', 'hidden-sm', 'hidden-xs'),
-  classNames('col-sm-2', 'hidden-xs'),
+  '',
+  '',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-16-on-lg',
+  'pf-m-hidden pf-m-visible-on-md pf-u-w-16-on-md',
```
karthikjeeyar, line 116: These classnames doesn't map with the existing classnames

abhinandan13jan: Well, it does I believe. breakpoint `sm` in bootstrap is `md` in patternfly and`md` is `lg`
So, first two r default width columns and the second two have corresponding classes
Screenshot from 2021-06-21 20-01-05

karthikjeeyar: If you are in a namespaced  page, then `secret` and `created` column is pushed to its right (originally it was evenly spaced)
image

 so if you are in 980px then the column names are getting truncated even though there is a room to show two column names 
 
image

abhinandan13jan: Thanks for pointing it out...Well, I tried a bit and a direct conversion of the existing classes dont quite work well, so I fine tuned keeping in mind various viewport sizes

Screenshot from 2021-06-22 15-25-04
Screenshot from 2021-06-22 15-25-18
Screenshot from 2021-06-22 15-25-37

karthikjeeyar:
Thanks @abhinandan13jan for updating the classnames, looks good to me
/lgtm

frontend/public/components/service-account.jsx
```diff
@@ -110,11 +110,11 @@ const menuActions = [
...
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-sm-4', 'col-xs-6'),
-  classNames('col-md-2', 'hidden-sm', 'hidden-xs'),
-  classNames('col-sm-2', 'hidden-xs'),
-  Kebab.columnClass,
+  'pf-u-w-50-on-sm pf-u-w-33-on-md pf-u-w-25-on-lg',
+  'pf-u-w-25-on-md',
+  'pf-m-hidden pf-m-visible-on-lg pf-u-w-25-on-lg',
+  'pf-m-hidden pf-m-visible-on-md pf-u-w-25-on-md',
+  `${Kebab.columnClass} pull-right`,
```
jerolimov, line 117: @abhinandan13jan I think pull-right is an indicator that we waste space for the kebab/action icon column. I like that the most other changes has at least one column without a width, which takes over "all leftover" space.

Service accounts on a big screen before:
image

On a small screen before:
image

With this PR on a big screen:
image

With this PR on a small screen (you can see the 25% space for the action column):
image

I would suggest a css without styling for the name and without `pull-right`.

```suggestion
  '',
  '',
  'pf-m-hidden pf-m-visible-on-lg pf-u-w-25-on-lg',
  'pf-m-hidden pf-m-visible-on-md pf-u-w-25-on-md',
  Kebab.columnClass,
```

It looks like this then (yes, also not similar to the original design) on a big screen:
image

And like this on a small screen:
image

Wdyt?

frontend/public/components/resource-quota.jsx
```diff
@@ -73,9 +73,9 @@ const getResourceUsage = (quota, resourceType) => {
 };
 
 const tableColumnClasses = [
-  classNames('col-md-5', 'col-xs-6'),
-  classNames('col-md-7', 'col-xs-6'),
-  Kebab.columnClass,
+  'pf-u-w-42-on-lg',
+  'pf-u-w-58-on-lg',
+  `${Kebab.columnClass} pull-right`,
```
jerolimov, line 78: Similar here. `pull-right` is only required because the namespace column is not shown when a namespace is selected. And the other one has a fix width. So the kebab column will take over the rest of the space.

This means on a ~ 1100px browser window, most space is wasted for the action menu and we have less then 300px for the name column. Which is many cases more then enough. But I personally would prefer a 'dynamic column width'  instead of a hardcoded 5:7 (name to namespace) column width, which makes problem when the namespace is now shown. Wdyt?

image

```suggestion
  '',
  '',
  Kebab.columnClass,
```

abhinandan13jan: updated

frontend/public/components/service-account.jsx
abhinandan13jan: updated

jerolimov:
Re-adding lgtm from @karthikjeeyar since the latest changes only fixes my comments above.
/lgtm

Thanks for fixing the service account and resource quota `pull-right` issues mentioned above. Tested them locally again and lgtm.
/approve

PR #9235, opened 2021-06-14
Capitalize help text for when expressions in the task sidebar

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-5943

**Screenshot:**
Screenshot from 2021-06-14 21-19-24
Screenshot from 2021-06-15 00-59-19
Screenshot from 2021-06-15 00-59-02


andrewballantyne:
This covers the situation... I want to now go on an easter egg hunt to see if anything else is missing. So Brigid doesn't need to log another issue in a few days 😄

bgliwa01:
Help text looks good! The dropdown for operator should be "Select operator" in order to comply with sentence case

andrewballantyne:
For posterity, I've inquired with @bgliwa01 about the two different wordings we have for using `$(`...

Ah, interesting, we have two wordings
```Use this format when referencing variables in this form: <code>$(</code>``` (params)
vs
```Use this format while referencing the variables in this form: <code>$(</code>``` (when)

**Comment by andrewballantyne** [comment]
@debsmita1 Would you mind updating some of the error text to have proper sentence case; missing period in messages found in `frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-builder/validation-utils.ts`:

- `Resource type has changed, reselect`
- `Workspace name has changed, reselect`
- `TaskSpec or TaskRef must be provided`
- `Must define at least one Task`

Last two may not appear today, but it's best to just align it.

**Comment by bgliwa01** [comment]
> For posterity, I've inquired with @bgliwa01 about the two different wordings we have for using `$(`...
> 
> > Ah, interesting, we have two wordings
> > `Use this format when referencing variables in this form: <code>$(</code>` (params)
> > vs
> > `Use this format while referencing the variables in this form: <code>$(</code>` (when)

After talking to content, it should be "Use this format when you reference variables in this form: <code>$(</code>"

**Comment by debsmita1** [comment]
@andrewballantyne Should I change `Must define at least one Task` to `Must define at least one task` ? or should I simply add period to this ?

**Comment by andrewballantyne** [comment]
> @andrewballantyne Should I change `Must define at least one Task` to `Must define at least one task` ? or should I simply add period to this ?

Oooh, yeah `task` please. Task is not the resource, it's a concept in the Pipeline. Good catch.

**Comment by debsmita1** [comment]
> > @andrewballantyne Should I change `Must define at least one Task` to `Must define at least one task` ? or should I simply add period to this ?
> 
> Oooh, yeah `task` please. Task is not the resource, it's a concept in the Pipeline. Good catch.

Aslo ,
Screenshot from 2021-06-15 00-05-00
and 
Screenshot from 2021-06-15 00-06-31

 to `Remove task` and `Select task` respectively ?

**Comment by andrewballantyne** [comment]
> > > @andrewballantyne Should I change `Must define at least one Task` to `Must define at least one task` ? or should I simply add period to this ?
> > 
> > 
> > Oooh, yeah `task` please. Task is not the resource, it's a concept in the Pipeline. Good catch.
> 
> Aslo ,
> Screenshot from 2021-06-15 00-05-00
> and
> Screenshot from 2021-06-15 00-06-31
> 
> to `Remove task` and `Select task` respectively ?

Ah yeah, same reason. I read that and went "yeah, no that's correct" but in reality it is not. Also please change the modal.

File: frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineSecretSection.tsx [hunk 1]
```diff
@@ -98,7 +98,7 @@ const PipelineSecretSection: React.FC = () => {
               className="odc-pipeline-secret-section__secret-action"
               icon={<PlusCircleIcon />}
             >
-              {t('pipelines-plugin~Add Secret')}
+              {t('pipelines-plugin~Add secret')}
```
andrewballantyne, line 101: Secret is a k8s resource. Capital is correct I believe.

cc @bgliwa01

andrewballantyne (changes requested):

andrewballantyne:
There are some test failures too, that vet text you modified.

debsmita1:
> There are some test failures too, that vet text you modified.

Fixed the failing tests

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineSecretSection.tsx
jerolimov: Just fyi:

image

But I also know that Secret is here "exact a Secret resource" and the values in the dropdown are talking about the concept "secret" ... Anyway a little bit confusing.

andrewballantyne (approved):

bgliwa01 (approved):
lgtm

jerolimov (approved):
/lgtm

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineSecretSection.tsx
bgliwa01: When secret is referred to as a resource it should be capital

andrewballantyne:
/hold

@debsmita1 can we address the `Secret` issue I mentioned above before merge to avoid another bug to address it.

frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/PipelineSecretSection.tsx
debsmita1: Done

debsmita1:
> /hold
> 
> @debsmita1 can we address the `Secret` issue I mentioned above before merge to avoid another bug to address it.

Done

andrewballantyne (approved):
/lgtm
/unhold

Thanks @debsmita1 !

PR #9237, opened 2021-06-14
Unable to edit the default Health check probe values

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-6015

**Root analysis:**
The field input type was changed from number to text to bypass the browser validation (https://github.com/openshift/console/pull/8198), because of this the probe data is received as a string instead of number

**Solution description:**
- The field values received is converted to a number before updating the resource
- Fixed the Form footer alignment

**GIF:**
Peek 2021-06-15 13-06

### DIFF
--- a/frontend/packages/dev-console/src/components/health-checks/AddHealthChecks.tsx
+++ b/frontend/packages/dev-console/src/components/health-checks/AddHealthChecks.tsx
@@ -4,13 +4,12 @@ import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { FormikProps, FormikValues } from 'formik';
 import * as _ from 'lodash';
 import Helmet from 'react-helmet';
-import { useTranslation } from 'react-i18next';
+import { useTranslation, Trans } from 'react-i18next';
 import {
   ContainerDropdown,
   history,
   PageHeading,
   ResourceLink,
-  ResourceIcon,
   openshiftHelpBase,
 } from '@console/internal/components/utils';
 import { ContainerModel } from '@console/internal/models';
@@ -93,19 +92,21 @@ const AddHealthChecks: React.FC<FormikProps<FormikValues> & AddHealthChecksProps
           </>
         }
       />
-      <div className="odc-add-health-checks__body">
-        <p>
-          {t('devconsole~Health checks for')} &nbsp;
-          <ResourceLink
-            kind={referenceFor(resource)}
-            name={name}
-            namespace={namespace}
-            title={name}
-            inline
-          />
-        </p>
-        <Form onSubmit={!viewOnly ? handleSubmit : undefined}>
-          <div>
+      <Form onSubmit={!viewOnly ? handleSubmit : undefined}>
+        <div className="odc-add-health-checks__body">
+          <p>
+            <Trans t={t} ns="devconsole">
+              Health checks for &nbsp;
+              <ResourceLink
+                kind={referenceFor(resource)}
+                name={name}
+                namespace={namespace}
+                title={name}
+                inline
+              />
+            </Trans>
+          </p>
+          <p>
             {t('devconsole~Container')} &nbsp;
             {_.size(containers) > 1 ? (
               <ContainerDropdown
@@ -114,24 +115,27 @@ const AddHealthChecks: React.FC<FormikProps<FormikValues> & AddHealthChecksProps
                 onChange={handleSelectContainer}
               />
             ) : (
-              <>
-                <ResourceIcon kind={ContainerModel.kind} />
-                {containers[0].name}
-              </>
+              <ResourceLink
+                kind={ContainerModel.kind}
+                name={containers[0].name}
+                linkTo={false}
+                inline
+              />
             )}
-          </div>
+          </p>
+          <br />
           <HealthChecks resourceType={getResourcesType(resource)} />
-          <FormFooter
-            handleReset={handleReset}
-            errorMessage={status && status?.errors?.json?.message}
-            isSubmitting={isSubmitting}
-            submitLabel={healthCheckAdded ? t('devconsole~Save') : t('devconsole~Add')}
-            disableSubmit={isFormClean || !dirty || !_.isEmpty(errors) || isSubmitting}
-            resetLabel={t('devconsole~Cancel')}
-            hideSubmit={viewOnly}
-          />
-        </Form>
-      </div>
+        </div>
+        <FormFooter
+          handleReset={handleReset}
+          errorMessage={status && status?.errors?.json?.message}
+          isSubmitting={isSubmitting}
+          submitLabel={healthCheckAdded ? t('devconsole~Save') : t('devconsole~Add')}
+          disableSubmit={isFormClean || !dirty || !_.isEmpty(errors) || isSubmitting}
+          resetLabel={t('devconsole~Cancel')}
+          hideSubmit={viewOnly}
+        />
+      </Form>
     </HealthCheckContext.Provider>
   );
 };
--- a/frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
+++ b/frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
@@ -6,8 +6,8 @@ import { HealthCheckProbeData, RequestType, HealthChecksProbeType } from './heal
 
 export const constructProbeData = (data: HealthCheckProbeData, resourceType?: Resources) => {
   const probeData = {
-    ...(data.failureThreshold && { failureThreshold: data.failureThreshold }),
-    ...(data.successThreshold && { successThreshold: data.successThreshold }),
+    ...(data.failureThreshold && { failureThreshold: _.toInteger(data.failureThreshold) }),
+    ...(data.successThreshold && { successThreshold: _.toInteger(data.successThreshold) }),
     ...(data.requestType === RequestType.ContainerCommand && {
       exec: data.exec,
     }),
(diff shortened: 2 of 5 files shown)


frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
```diff
@@ -6,8 +6,8 @@ import { healthChecksDefaultValues } from './health-checks-probe-utils';
 
 export const constructProbeData = (data: HealthCheckProbeData, resourceType?: Resources) => {
   const probeData = {
-    ...(data.failureThreshold && { failureThreshold: data.failureThreshold }),
-    ...(data.successThreshold && { successThreshold: data.successThreshold }),
+    ...(data.failureThreshold && { failureThreshold: _.toNumber(data.failureThreshold) }),
+    ...(data.successThreshold && { successThreshold: _.toNumber(data.successThreshold) }),
```
jerolimov, line 10: Can we use `_.toInteger` here? This was already used in this utils function and we know that the data value only accepts positive integers here.

frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
```diff
@@ -26,10 +26,10 @@ export const constructProbeData = (data: HealthCheckProbeData, resourceType?: Re
       },
     }),
     ...(data.initialDelaySeconds && {
-      initialDelaySeconds: data.initialDelaySeconds,
+      initialDelaySeconds: _.toNumber(data.initialDelaySeconds),
     }),
-    ...(data.periodSeconds && { periodSeconds: data.periodSeconds }),
-    ...(data.timeoutSeconds && { timeoutSeconds: data.timeoutSeconds }),
+    ...(data.periodSeconds && { periodSeconds: _.toNumber(data.periodSeconds) }),
+    ...(data.timeoutSeconds && { timeoutSeconds: _.toNumber(data.timeoutSeconds) }),
```
jerolimov, line 32: Same here, `_.toInteger`

jerolimov (changes requested):
Tested this locally, works fine. But please let us use toInteger instead of toNumber.

jerolimov:
@debsmita1 Can you also update the existing test in `create-health-check-probe-utils.spec.ts`? Maybe its enough to convert the numbers at the top of this file to strings?

debsmita1:
> @debsmita1 Can you also update the existing test in `create-health-check-probe-utils.spec.ts`? Maybe its enough to convert the numbers at the top of this file to strings?

Done

frontend/packages/dev-console/src/components/health-checks/AddHealthChecks.tsx
```diff
@@ -92,19 +92,19 @@ const AddHealthChecks: React.FC<FormikProps<FormikValues> & AddHealthChecksProps
...
+          <p>
+            {t('devconsole~Health checks for')} &nbsp;
+            <ResourceLink
+              kind={referenceFor(resource)}
+              name={name}
+              namespace={namespace}
+              title={name}
+              inline
+            />
+          </p>
```
andrewballantyne, line 106: Nit: This should probably be using `<Trans>...</Trans>`. Translators should always be given full context as not all languages keep the same sentence structure. Sometimes the noun is first, sometimes it's last.

frontend/packages/dev-console/src/components/health-checks/health-checks-types.ts
```diff
@@ -25,10 +25,10 @@ export interface HealthCheckProbeData {
...
   };
   exec?: { command?: string[] };
-  initialDelaySeconds: number;
-  periodSeconds: number;
-  timeoutSeconds: number;
-  successThreshold: number;
+  initialDelaySeconds: number | string;
+  periodSeconds: number | string;
+  timeoutSeconds: number | string;
+  successThreshold: number | string;
```
andrewballantyne, line 31: This muddiest the water of our TypeScript types. Either we need to convert at Input level to a number, and always treat it as number. Or we need to have two types... one for the formik data structure which are input value driven data. And then you want pure representation data that we talk about in "HealthCheck" terms.

Merging these two things together just allows us to not ever know by just looking at the type which state it is in.

andrewballantyne (changes requested):

frontend/packages/dev-console/src/components/health-checks/health-checks-types.ts
debsmita1: I did this for sake of adding a unit test where the input data is string, the utils are all strictly typed, and won't allow testing string data

andrewballantyne: If you need to mangle types for Tests, please extend the type in the test file and expand it there. Best not impact our application types for tests.

jerolimov: Hmmm, maybe its a good idea to split this into two kinds of data.

HealthCheckProbe vs HealthCheckForm and then constructProbeData gets form data and return "kubernetes data".

But I don't know how big this change is. But it sounds not sooo bad, or?

andrewballantyne: We are "just" inside a new release. This is not the time to concern ourselves with it. If this fix/bug is something we need to backport, then perhaps we go with what we have.

If it's not intended for backport or @debsmita1 you're okay with making a slimmed version for backport manually, we should 100% be looking towards the horizons when making type changes and not do something that lowers our quality and ease of use.

frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
```diff
@@ -6,8 +6,8 @@ import { healthChecksDefaultValues } from './health-checks-probe-utils';
 
 export const constructProbeData = (data: HealthCheckProbeData, resourceType?: Resources) => {
   const probeData = {
-    ...(data.failureThreshold && { failureThreshold: data.failureThreshold }),
-    ...(data.successThreshold && { successThreshold: data.successThreshold }),
+    ...(data.failureThreshold && { failureThreshold: _.toInteger(data.failureThreshold) }),
```
nemesis09, line 9: could we use `parseInt` here instead of lodash?

jerolimov: There are some other `_.toInteger` calls in this function before. I think its good to use a similar way in one function.

frontend/packages/dev-console/src/components/health-checks/health-checks-types.ts
jerolimov: I think we should backport this asap as this breaks the health checks, at least if the user changes one of this five numeric fields. So I hope that we pick this up for the first z-stream release.

I checked the code now again. Andrew is in general right that we should distinguish between k8s resource data and internal form data. From my understand the `HealthChecksData` is currently used as form data type, not as resource definition.

The convert from form data to resource data happen in `updateHealthChecksProbe` and `getProbesData`.

And the convert from resource data to form data in `getHealthChecksData`.

At the moment this last method doesn't convert the resource integers to form strings. But when the user changes the values in the form we currently save strings in the form data. So at the moment its right that we have number|string values in the form. (Status quo. Which doesn't mean that I liked this.)

So one option is that we create two clean types. One for the FormData and one for the ResourceData and update all methods above to ensure that they convert between strings and numbers. So that the type doesn't change anymore when the user touches a field.

The other one is to create a follow up issue (tech debt or bug) and accept that the form mixes two types at the moment. This makes at least the verification and backport of this fix easy. When I understand Andrew correctly this is also acceptable for him.

andrewballantyne: > I think we should backport this asap as this breaks the health checks, at least if the user changes one of this five numeric fields. So I hope that we pick this up for the first z-stream release.

No z-streams for 4.8 for at least 3 weeks, if not more. We need to go GA, then 1 or 2 z-streams (at roughly 1 week cadences) before we can get changes into it.

It's fine doing it a tech debt, but I am concerned about our approach to code. We have always had TypeScript lax feelings and this is a good case when we are opting for more lax types to multi-purpose them. I'm personally against this style of coding. We can take this as-is and port it back, wait 3ish weeks to merge it, and in the meantime work to clean this up in 4.9 asap (and it cannot be "on the backlog" imo).

andrewballantyne (approved):
Please log a backlog ticket with my two comments and complete it this sprint (separate of this PR).

frontend/packages/dev-console/src/components/health-checks/health-checks-types.ts
debsmita1: logged ticket https://issues.redhat.com/browse/ODC-6037

frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
nemesis09: lodash is only being used for `toInteger` here which can be done using `parseInt`.
shouldn't we prefer using standard built in functions to external library?

jerolimov (approved):
Verified import flow with an health check and adding/modifing a health check later on an existing deployment.

/lgtm

frontend/packages/dev-console/src/components/health-checks/create-health-checks-probe-utils.ts
andrewballantyne: @nemesis09 You're waaay too late to that party lol. We have corrected it in a few places, but for the most part it is a mixed bag. There isn't too much loss to using lodash, and there is often a lot of code safety to gain in the utilities.

My preference is definitely to go with native functions when they do the same thing, to reduce our reliance on "magic" 3rd party utility functions. But in the end it doesn't really matter.

PR #9238, opened 2021-06-14
Bug 1934604: Add RetryOnConflict wrapper around helm chart install

zonggen:
- [x] Adds a go test that simulates a situation where resourcequota is frequently modified to increase the chance of race condition.    
 - [x] Adds a retry wrapper around console helm chart installtion to workaround the situation where resourcequota has been modified and causes installation failure.
 - [x] Limits to 2 retry attempts.
 - [x] Tidy up vendor directory according to dependencies with `go mod vendor`
    
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1934604
Signed-off-by: Allen Bai <abai@redhat.com>

Upstream issue & PR:
 - https://github.com/helm/helm/issues/9710
 - https://github.com/helm/helm/pull/9713


pkg/helm/actions/install_chart.go
```diff
@@ -5,38 +5,50 @@ import (
...
+			ch.Metadata.Annotations = make(map[string]string)
+		}
+		ch.Metadata.Annotations["chart_url"] = url
+
+		cmd.Namespace = ns
+		release, err = cmd.Run(ch, vals)
+
+		// Cleanup failed release before re-Run
+		if err != nil {
+			UninstallRelease(name, conf)
```
zonggen, line 47: Can the return value of this line be discarded? I think we can because the error will be reflected on the next retry cycle anyway, and will cause next retry to fail. In addition, I don't want the error of this line obscure the error returned by `cmd.Run`.

zonggen: Might need opinions from other folks on this.

pkg/helm/actions/install_chart.go
```diff
@@ -5,38 +5,50 @@ import (
...
-	cmd := action.NewInstall(conf)
+	var release *release.Release
 
-	releaseName, chartName, err := cmd.NameAndChart([]string{name, url})
-	if err != nil {
-		return nil, err
-	}
-	cmd.ReleaseName = releaseName
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		cmd := action.NewInstall(conf)
```
dperaza4dustbit, line 15: Can we just retry around the run call instead of the prep calls?

pkg/helm/actions/resource_stress_test.go
```diff
@@ -0,0 +1,179 @@
+package actions
```
dperaza4dustbit, line 1: Where you able to reproduce from here?

dperaza4dustbit (commented):
See some comments and do we have to include the whole gitlabs binary? Can't we just download from the net on the flight?

pkg/helm/actions/install_chart.go
zonggen: Originally was concerning the setup step plays a role in `"cannot re-use a name that is still in use"` error. Just tested wrapping around only `cmd.Run` and it worked. Will update.

pkg/helm/actions/resource_stress_test.go
zonggen: Can you clarify on this?

pkg/helm/actions/install_chart.go
zonggen: Updated.

zonggen:
Updated according to review, lifted WIP flag.

zonggen:
> expected the bug to target the "4.9.0" release, but it targets "4.3" instead

Should we change the target release to 4.9? @dperaza4dustbit

zonggen:
`ci/prow/e2e-gcp-console` failed with 
```
build github.com/openshift/console/cmd/bridge: cannot load k8s.io/client-go/util/retry: open /go/src/github.com/openshift/console/vendor/k8s.io/client-go/util/retry: no such file or directory
```
even though I've added the dependency to the vendor by `go mod vendor`.

dperaza4dustbit (changes requested):
Lets wait to merge this until we figure out the helm async install story

zonggen:
Blocked because of browser timeout is ~30s, which will cause the console error out before the retry has returned valid result.

Need to revisit again for another solution, closing.

PR #9245, opened 2021-06-15
Fix monitoring error box alignment

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5997

**Analysis / Root cause**: 
Error message is aligned with a flex layout to the center.

**Solution Description**: 
Stretch the content of the flex layout.

I tested also the admin monitoring area but didn't find any difference.

Before (in case of an error, for example with a local CRC):
monitoring-error-before

With this PR and an error (local CRC):
with-pr-without-monitoring

With this PR and installed monitoring (shared cluster):
with-pr-with-monitoring

Unchanged. CSS only change.

**Test setup:**
* To get the error use a CRC
* **Alternative:** switch to monitoring tab (to load the JS), switch back, disconnect from internet or stop a local bridge, and switch back to the monitoring tab

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/monitoring/_monitoring.scss
+++ b/frontend/public/components/monitoring/_monitoring.scss
@@ -20,7 +20,7 @@ $tooltip-background-color: var(--pf-global--BackgroundColor--dark-100);
 .graph-empty-state {
   min-height: 310px;
   &__loaded {
-    align-items: center;
+    align-items: stretch;
     display: flex;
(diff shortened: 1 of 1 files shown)


vikram-raj (commented):
verified the changes and its looks good to me. 

/lgtm

PR #9246, opened 2021-06-15
Bug 1972016: Fix time range issue for devconsole monitoring dashboard

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/OCPBUGSM-30921

**Analysis / Root cause**: 
Set a specific time range, but Dashboards display data with a different time range. `fixedEndTime` prop is missing from the `QueryBrowser` component and it is the cause of this issue.

**Solution Description**: 
 Passed the `endTime` prop to `QueryBrowser` component.

Kapture 2021-06-15 at 14 53 55


jerolimov (approved):
Tested this with a local console against a shared cluster and this change works fine.

/approve

PR #9256, opened 2021-06-15
Bug 1970466: Console's OperatorHub leads users to unrelated install plan, if subscription does not have its own

kdoberst:
Related to [Bug 1970466](https://bugzilla.redhat.com/show_bug.cgi?id=1970466)

When viewing an operator's detail information, a message is shown if the installed CSV(startingCSV) is not found instead of displaying that the operator has been installed.

Bug 1970466

As coded, if a yaml contains an unknown starting CSV, when calling useK8sWatchResource, the name key is set to undefined.  This causes usek8sWatchResource to return a perviously installed operator to be returned causing incorrect information to be shown.

### DIFF
--- a/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
+++ b/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
@@ -75,6 +75,15 @@ const InstalledHintBlock: React.FC<OperatorHubItemDetailsHintBlockProps> = ({
     isList: false,
     namespaced: true,
   });
+
+  if (!subscription?.status?.installedCSV) {
+    return (
+      <HintBlock className="co-catalog-page__hint" title="Operator Not Yet Installed">
+        <p>{t('olm~A suitable CSV is not found.')}</p>
+      </HintBlock>
+    );
+  }
+
   const nsPath = `/k8s/${namespace ? `ns/${namespace}` : 'all-namespaces'}`;
   const to = installedCSV
     ? `${nsPath}/clusterserviceversions/${installedCSV?.metadata?.name ?? ''}`
--- a/frontend/packages/operator-lifecycle-manager/locales/en/olm.json
+++ b/frontend/packages/operator-lifecycle-manager/locales/en/olm.json
@@ -207,6 +207,7 @@
   "Deep Insights": "Deep Insights",
   "Auto Pilot": "Auto Pilot",
   "Checked": "Checked",
+  "A suitable CSV is not found.": "A suitable CSV is not found.",
   "Version <1>{{installedVersion}}</1> of this Operator has been installed on the cluster.": "Version <1>{{installedVersion}}</1> of this Operator has been installed on the cluster.",
   "This Operator has been installed on the cluster.": "This Operator has been installed on the cluster.",
   "View it here.": "View it here.",


kdoberst:
@openshift/team-ux-leads. Can you please take a look at this PR?  There is a new message that is introduced as part of this PR and looking for feedback.

/assign @openshift/team-ux-leads

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
```diff
@@ -75,6 +75,15 @@ const InstalledHintBlock: React.FC<OperatorHubItemDetailsHintBlockProps> = ({
     isList: false,
     namespaced: true,
   });
+
+  if (!subscription?.status?.installedCSV) {
+    return (
+      <HintBlock className="co-catalog-page__hint" title="Operator Not Yet Installed">
```
spadgett, line 81: Hey, Kim. We might look at the messaging here. It's odd to see an "Uninstall" button on the same page with "Operator Not Yet Installed." Is the operator is still installing or installation failed? Is there something in the Subscription resource that will let us give a more specific status to the user?

cc @tlwu2013

kdoberst: I didn't see anything in the subscription resource that would indicate that the CSV hasn't been found vs a failure, but I can look again.

I would love some suggestions on what the UI should respond with installedCSV is undefined or missing.  I just went with what was suggested in the bug.

spadgett: Looking below, I see we try to fallback to the Subscription link when there's no CSV. I think that's the intended behavior, and I'd do that for this case as well.

kdoberst: Here is where this bug gets kind of weird.  On line 71, installedCSV is created by passing information. 
```
 const [installedCSV] = useK8sWatchResource<ClusterServiceVersionKind>({
    kind: referenceForModel(ClusterServiceVersionModel),
    name: subscription?.status?.installedCSV,
    namespace: subscription?.metadata?.namespace,
    isList: false,
    namespaced: true,
  });
```
In the case described in the bug, subscription.staus.insatlledCSV is actually undefined, but a item is returned from useK8sWatchResource.  So the logic used below on line 88:
```
 const to = installedCSV
    ? `${nsPath}/clusterserviceversions/${installedCSV?.metadata?.name ?? ''}`
    : `${nsPath}/subscriptions/${subscription.metadata.name ?? ''}`;
```

Would always return true.  What is actually happening is that the results from the useK8sWatchResource appears to return information on another installed operator where there was a valid CSV.  

The only piece of data that I could find to indicate this situation where need information is known except a valid CSV is checking for `subscription?.status?.installedCSV`.

spadgett:
I think this bug is superseded by story https://issues.redhat.com/browse/CONSOLE-2805, which will add an "Installing" status to OperatorHub.

cc @rebeccaalpert

frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-item-details.tsx
rebeccaalpert: Oh weird. I'm seeing this too. We probably need to beef that logic up a bit more.

rebeccaalpert: I think Sam is right and https://github.com/openshift/console/pull/9346 does cover this case (since it overrides the install hint block), but I'll double check.

kdoberst: Thanks @rebeccaalpert for taking a look.

TheRealJon: Hi @kdoberst! I think I might have written that, good catch. We probably need to add a condition there to make sure that nothing gets passed into `useK8sWatchResource` if `subscription.status.installedCSV` is empty. Something like:
```tsx
 const [installedCSV] = useK8sWatchResource<ClusterServiceVersionKind>( subscription?.status?.installedCSV ? {
    kind: referenceForModel(ClusterServiceVersionModel),
    name: subscription?.status?.installedCSV,
    namespace: subscription?.metadata?.namespace,
    isList: false,
    namespaced: true,
  } : null);
 ```

rebeccaalpert: @kdoberst - I tried the subscription YAML update in https://bugzilla.redhat.com/show_bug.cgi?id=1970466 against my PR and it doesn't trigger the new "isInstalling" hint block from what I can tell. I also wasn't seeing the error though. Did you use a different YAML to test it?

spadgett:
Closing in favor of #9346

PR #9257, opened 2021-06-15
Fix catalog tile links and default behaviour with ctrl+click

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-6032

**Analysis / Root cause**: 
- Catalog tiles were only setting href property if there was explicit href available for that item. This resulted in `/#` as href.
- Catalog tiles add an `onClick` handler for each tile which prevents default. This disables the `Ctrl+Click` behaviour for links.

**Solution Description**: 
- Set correct hrefs for every item in catalog tile.
- Add a check for modified clicks and do no prevent default in that case. This uses `isModified` check from `react-router` `Link` component.

https://user-images.githubusercontent.com/6041994/122077269-36926480-ce19-11eb-9809-031fa8498c22.mov

**Browser conformance**: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
+++ b/frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
@@ -51,6 +51,20 @@ export const updateURLParams = (paramName: string, value: string | string[]) => 
   setURLParams(params);
 };
 
+export const getURLWithParams = (paramName: string, value: string | string[]): string => {
+  const params = new URLSearchParams(window.location.search);
+  const url = new URL(window.location.href);
+
+  if (value) {
+    params.set(paramName, Array.isArray(value) ? JSON.stringify(value) : value);
+  } else {
+    params.delete(paramName);
+  }
+
+  const searchParams = `?${params.toString()}${url.hash}`;
+  return `${url.pathname}${searchParams}`;
+};
+
 export const getCatalogTypeCounts = (
   items: CatalogItem[],
   catalogTypes: CatalogType[],
@@ -63,4 +77,8 @@ export const getCatalogTypeCounts = (
   });
 
   return catalogTypeCounts;
+};
+
+export const isModifiedEvent = (event: React.MouseEvent<HTMLElement>) => {
+  return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 };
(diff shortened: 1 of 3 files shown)


frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
```diff
@@ -51,6 +51,20 @@ export const updateURLParams = (paramName: string, value: string | string[]) =>
   setURLParams(params);
 };
 
+export const getURLWithParams = (paramName: string, value: string | string[]): string => {
+  const params = new URLSearchParams(window.location.search);
+  const url = new URL(window.location.href);
+
+  if (value) {
```
divyanshiGupta, line 58: In this case if it is ever possible to get an empty array then it is better to add a check for that.
```suggestion
  if (!_.isEmpty(value)) {
```

divyanshiGupta (changes requested):

frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
```diff
@@ -64,3 +78,7 @@ export const getCatalogTypeCounts = (
 
   return catalogTypeCounts;
 };
+
+export const isModifiedEvent = (event) => {
```
divyanshiGupta, line 82: Can you add a type for the event?

divyanshiGupta (changes requested):

frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
rohitkrai03: Done.

frontend/packages/dev-console/src/components/catalog/utils/catalog-utils.tsx
rohitkrai03: Empty array is not going to be a problem as we are doing `JSON.stringify` after an `isArray` check.

jerolimov:
@rohitkrai03 The tiles work great, thanks. Is it possible that you fix this also for the developer catalog navigation on the left hand side this PR or should I create another issue for this?

* The catalog switch uses # as well.
* The type switch below has some URLs (starting with `?..`), but they doesn't work if you open them in a new tab. (Opens topology instead again.)

It's fine for me if you want track this in another issue. Please let me know.

rohitkrai03:
> @rohitkrai03 The tiles work great, thanks. Is it possible that you fix this also for the developer catalog navigation on the left hand side:
> 
> * The catalog switch uses # as well.
> * The type switch below has some URLs (starting with `?..`), but they doesn't work if you open them in a new tab. (Opens topology instead again.)
> 
> Do you want fix this in this PR as well or should I create a follow up issue. Just let me know.

@jerolimov I think its better to create a follow up issue for this. Keeps the PR small and targeted.

invincibleJai (approved):

PR #9258, opened 2021-06-15
Bug 1967885: Creating a VM from the UI on OKD 4.7 fails

glekner:
This PR also removes the `ModelDefinition` static extension from Kubevirt

### DIFF
--- a/frontend/packages/kubevirt-plugin/console-extensions.json
+++ b/frontend/packages/kubevirt-plugin/console-extensions.json
@@ -4,6 +4,17 @@
     "properties": {
       "model": {
         "group": "kubevirt.io",
+        "version": "v1alpha3",
+        "kind": "VirtualMachine"
+      },
+      "flag": "KUBEVIRT"
+    }
+  },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "kubevirt.io",
         "version": "v1",
         "kind": "VirtualMachine"
       },
@@ -224,7 +235,11 @@
   {
     "type": "console.page/route",
     "properties": {
-      "path": ["/k8s/ns/:ns/virtualmachines/:name"],
+      "path": [
+        "/k8s/ns/:ns/virtualmachines/:name",
+        "/k8s/ns/:ns/kubevirt.io~v1~VirtualMachine/:name",
+        "/k8s/ns/:ns/kubevirt.io~v1alpha3~VirtualMachine/:name"
+      ],
       "component": {
         "$codeRef": "VirtualMachinesDetailsPage.VirtualMachinesDetailsPage"
       }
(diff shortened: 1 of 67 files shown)


frontend/packages/kubevirt-plugin/src/components/dashboards-page/vm-dashboard/vm-inventory-card.tsx
```diff
@@ -47,7 +48,7 @@ export const VMInventoryCard: React.FC<VMInventoryCardProps> = () => {
   const snapshotResource: WatchK8sResource = React.useMemo(
     () => ({
       isList: true,
-      kind: VirtualMachineSnapshotModel.kind,
+      kind: referenceForModel(VirtualMachineSnapshotModel),
```
vojtechszocs, line 51: Shouldn't we use `kvReferenceForModel` here as well?

frontend/packages/kubevirt-plugin/console-extensions.json
```diff
@@ -1,4 +1,15 @@
 [
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "kubevirt.io",
+        "version": "v1alpha3",
+        "kind": "VirtualMachine"
+      },
+      "flag": "KUBEVIRT"
```
vojtechszocs, line 10: Note that in general, having the same feature flag determined by different models might lead to race condition, but in this particular case it's OK because we don't expect two different `VirtualMachine` CRDs (`v1` alongside `v1alpha3`) to be present on the cluster.

frontend/packages/kubevirt-plugin/src/models/index.ts
```diff
@@ -10,6 +10,7 @@ export const VirtualMachineModel: K8sKind = {
   namespaced: true,
   kind: 'VirtualMachine',
   id: 'virtualmachine',
+  crd: true,
```
vojtechszocs, line 13: :+1: on marking KubeVirt models as CRDs.

frontend/packages/kubevirt-plugin/src/models/kvReferenceForModel.ts
```diff
@@ -0,0 +1,37 @@
+import * as _ from 'lodash';
+import { modelFor, referenceForGroupVersionKind } from '@console/internal/module/k8s';
+import { K8sKind } from '@console/internal/module/k8s/types';
+
+const VERSIONS = ['v1', 'v1alpha3', 'v1beta1', 'v1alpha1'];
+
+export const kvReferenceForModel = (model: K8sKind): string => {
```
vojtechszocs, line 7: Nit: I'd suggest naming this `kubevirtReferenceForModel` but I'm also OK if you want to keep it this way.

frontend/packages/kubevirt-plugin/src/models/kvReferenceForModel.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (availableVersion) {
+    return modelFor(referenceForGroupVersionKind(model.apiGroup)(availableVersion)(model.kind));
+  }
+
+  return model;
+};
+
+export const getKubevirtModelAvailableAPIVersion = (model: K8sKind): string => {
+  const availableVersion = _.uniq([model.apiVersion, ...VERSIONS]).find(
```
vojtechszocs, line 32: Nit: you could extract the `availableVersion` computation into separate function to cut down code duplicity.

```ts
const getKubevirtAvailableModelVersion = (model: K8sKind): string =>
  _.uniq([model.apiVersion, ...VERSIONS]).find(
    (v) => !!modelFor(referenceForGroupVersionKind(model.apiGroup)(v)(model.kind)),
  );
```

and then replace `availableVersion` references with `getKubevirtAvailableModelVersion(model)` calls.

PR #9260, opened 2021-06-15
Use different suffix for pipeline secret if annotation already exists

rottencandy:
**Fixes**: 
https://issues.redhat.com/browse/ODC-6003

**Analysis / Root cause**: 
In pipelines created by git import flow, annotations of a shared secret will get overridden when the git host is different.
For example: `tekton.dev/git-0: https://github.com` will get replaced by `tekton.dev/git-0: https://gitlab.com`, causing the previous pipeline to fail.

**Solution Description**: 
Use a different suffix and don't change existing annotations, ex. `tekton.dev/git-1: https://gitlab.com`.

**Test setup:**
Requires two private repositories, hosted on different git services (example using github & gitlab each)
And sharing the same deploy key (ssh preferred) which can be used as secret.
1. Navigate to git import flow and create a deployment using the first repository, add the deploy key as secret and check the "add pipeline" option.
2. Create another deployment using the second repository, select the above created secret and check "add pipeline" option.
3. After both the pipelines have finished running, rerun the pipeline from the first deployment.

**Browser conformance**: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

/kind bug

### DIFF
--- a/frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-utils.spec.ts
+++ b/frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-utils.spec.ts
@@ -144,6 +144,55 @@ describe('pipeline-utils ', () => {
     });
   });
 
+  it('should have correct annotation key prefix when there are existing annotations', () => {
+    const existingAnnotations = {
+      'tekton.dev/git-0': 'gitlab.com',
+    };
+    const annotations = getSecretAnnotations(
+      {
+        key: SecretAnnotationId.Git,
+        value: 'github.com',
+      },
+      existingAnnotations,
+    );
+    expect(annotations).toEqual({
+      'tekton.dev/git-0': 'gitlab.com',
+      'tekton.dev/git-1': 'github.com',
+    });
+  });
+
+  it('should avoid adding duplicate annotations to secret', () => {
+    const existingAnnotations = {
+      'tekton.dev/git-0': 'github.com',
+    };
+    const annotations = getSecretAnnotations(
+      {
+        key: SecretAnnotationId.Git,
+        value: 'github.com',
+      },
+      existingAnnotations,
+    );
+    expect(annotations).toEqual({
+      'tekton.dev/git-0': 'github.com',
+    });
+  });
+
+  it('should return unmodified annotations if invalid key is provided', () => {
+    const existingAnnotations = {
+      'tekton.dev/git-0': 'gitlab.com',
+    };
+    const annotations = getSecretAnnotations(
+      {
+        key: 'invalid-type',
+        value: 'github.com',
+      },
+      existingAnnotations,
+    );
+    expect(annotations).toEqual({
+      'tekton.dev/git-0': 'gitlab.com',
+    });
+  });
+
   it('expected relative time should be "a few seconds"', () => {
     const relativeTime = calculateRelativeTime('2020-05-22T11:57:53Z', '2020-05-22T11:57:57Z');
     expect(relativeTime).toBe('a few seconds');
--- a/frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
+++ b/frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
@@ -383,15 +383,35 @@ type KeyValuePair = {
   key: string;
   value: string;
 };
-export const getSecretAnnotations = (annotation: KeyValuePair) => {
-  const annotations = {};
+
+const getAnnotationKey = (secretType: string, suffix: number) => {
   const annotationPrefix = 'tekton.dev';
-  if (annotation?.key === SecretAnnotationId.Git) {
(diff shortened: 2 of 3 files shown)


frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
```diff
@@ -383,15 +383,29 @@ type KeyValuePair = {
...
+  annotation: KeyValuePair,
+  existingAnnotations: { [key: string]: string } = {},
+) => {
+  let count = 0;
+  let annotationKey = getAnnotationKey(annotation?.key, count);
+  while (existingAnnotations[annotationKey]) {
+    annotationKey = getAnnotationKey(annotation?.key, ++count);
+  }
+
+  return { ...existingAnnotations, [annotationKey]: annotation?.value };
```
invincibleJai, line 408: do we need to maintain count or can use a random number to generate a new unique key and `getAnnotationKey` even return null as key if  `annotation?.key` is not a match. Does below make sense can avoid function `getAnnotationKey`  as well

```suggestion
export const getSecretAnnotations = (
  annotation: KeyValuePair,
  existingAnnotations: { [key: string]: string } = {},
) => {
 if(!annotation?.key || ![SecretAnnotationId.Git, SecretAnnotationId.Image].includes(annotation?.key )) return existingAnnotations;
  const annotationPrefix = 'tekton.dev';
  const suffix = Math.round(Math.random() * 10000);
  const annotationKey = `tekton.dev/${secretType}-${suffix}`;
  return { ...existingAnnotations, [annotationKey]: annotation?.value };
}
```

karthikjeeyar: In Tekton documentation, says the credential annotation it must begin with  `tekton.dev/git-` or `tekton.dev/docker-`,  technically I guess we can add any random number, however adding the count keeps us consistent with how they outlined the usage in docs and the yaml users might get confused on seeing random number at the suffix. I would say lets go with count.

https://github.com/tektoncd/pipeline/blob/main/docs/auth.md#understanding-credential-selection

invincibleJai: makes sense, let's handle the null scenario for this `getAnnotationKey` and add / update specs for utils  cc @rottencandy

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
```diff
@@ -383,15 +383,29 @@ type KeyValuePair = {
...
+
+export const getSecretAnnotations = (
+  annotation: KeyValuePair,
+  existingAnnotations: { [key: string]: string } = {},
+) => {
+  let count = 0;
+  let annotationKey = getAnnotationKey(annotation?.key, count);
+  while (existingAnnotations[annotationKey]) {
+    annotationKey = getAnnotationKey(annotation?.key, ++count);
+  }
```
karthikjeeyar, line 406: Avoid duplicate entries

```suggestion
  while (existingAnnotations[annotationKey] && existingAnnotations[annotationKey] !== annotation?.value) {
    annotationKey = getAnnotationKey(annotation?.key, ++count);
  }
```

```  annotations:
    tekton.dev/git-0: 'https://github.com'
    tekton.dev/git-1: 'https://gitlab.com'
    tekton.dev/git-2: 'https://github.com'  // We should avoid this duplicate entries for the value
```

frontend/packages/pipelines-plugin/src/utils/__tests__/pipeline-utils.spec.ts
```diff
@@ -144,6 +144,23 @@ describe('pipeline-utils ', () => {
...
+        key: SecretAnnotationId.Git,
+        value: 'github.com',
+      },
+      existingAnnotations,
+    );
+    expect(annotations).toEqual({
+      'tekton.dev/git-0': 'gitlab.com',
+      'tekton.dev/git-1': 'github.com',
+    });
+  });
```
karthikjeeyar, line 162: Could you add one test case for avoiding duplicates and also one for incorrect `annotation.key`  
eg: 
```
getSecretAnnotations(
   {
     key: 'not-a-real-key', //  now it is returning {"null" : 'github.com'} but it should be {}
     value: 'github.com',
     },
    existingAnnotations,
 );
 ```

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
rottencandy: Thanks. Updated to handle invalid key scenario.

frontend/packages/pipelines-plugin/src/utils/pipeline-utils.ts
rottencandy: Updated.

karthikjeeyar:
Thanks @rottencandy, Works as expected
/lgtm

rohitkrai03 (approved):
/approve

PR #9261, opened 2021-06-15
CONSOLE-2810: Update react-jsonschema-form package

cyril-ui-developer:
/assign @TheRealJon 
cc: @rhamilto 

The warning error - `Warning: validateDOMNesting(...): <form> cannot appear as a descendant of <form>.`  that I mentioned in scrum today is gone after I changed the `import` statement of `SchemaField` and  `SchemaFieldProps ` from `'react-jsonschema-form/lib/components/fields/SchemaField'` to  `'@rjsf/core/dist/cjs/components/fields/SchemaField';`


frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
```diff
@@ -362,15 +360,15 @@ export default {
   NullField,
   PodAffinityField,
   ResourceRequirementsField,
-  SchemaField: CustomSchemaField,
```
cyril-ui-developer, line 365: @TheRealJon I removed the `CustomSchemaField`  based on the below comment in code:

`export const CustomSchemaField: React.FC<SchemaFieldProps> = (props) => {
  // TODO Remove this workaround when the issue has been fixed upstream in react-jsonschema-form and
  // we bump our version to include that fix.
  `
Is the below comment still valid?
` // Provide a fallback formData value for objects and arrays to prevent undefined
  // references. This can occur if formData is malformed (for example, almExamples annotation
  // explicitly sets an array value to null). `

TheRealJon (commented):
See comment

cyril-ui-developer:
@TheRealJon Updated! Thanks.

frontend/packages/console-shared/src/components/dynamic-form/fields.tsx
```diff
@@ -324,31 +321,7 @@ export const DropdownField: React.FC<FieldProps> = ({
...
-        return {};
-      default:
-        return undefined;
-    }
-  }, [type]);
-
-  // If a the provided schema will not generate any form field elements, return null.
-  if (hasNoFields(props.schema, props.uiSchema)) {
-    return null;
-  }
```
TheRealJon, line 350: Don't remove these lines. We still want to return null if there are no form fields generated by the schema.

cyril-ui-developer: > Don't remove these lines. We still want to return null if there are no form fields generated by the schema.

@TheRealJon Updated. Thanks.

TheRealJon (approved):
/lgtm

spadgett:
/assign @yapei 
for QE approval

@yapei There are no user visible changes. I think we just need a regression test of the operand form.

yapei:
build a local console with `./build.sh` and did regression tests of operand form, no issues found
/label qe-approved

spadgett:
No user-visible changes. Adding the docs-approved label. 
cc @ahardin-rh 

/label docs-approved
/approve
/hold until we are under our SLO

cyril-ui-developer:
@spadgett I rebased, so it needs lgtm label again.

PR #9262, opened 2021-06-15
Update ProjectsLink to use link instead of button

rohitkrai03:
**Fixes**: https://issues.redhat.com/browse/ODC-6001

**Analysis / Root cause**: 
- Project links in the project list page for all projects doesn't work as proper links.
- `ProjectLink` in `ProjectsTable` uses `button` instead of a link. 

**Solution Description**: 
- Use `Link` instead of button for project links and simulate the same `onClick` behaviour.

https://user-images.githubusercontent.com/6041994/122255403-b1c44b00-ceeb-11eb-9769-f417b1e474a0.mov

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/frontend/public/components/namespace.jsx
+++ b/frontend/public/components/namespace.jsx
@@ -27,8 +27,11 @@ import {
   withUserSettingsCompatibility,
   COLUMN_MANAGEMENT_CONFIGMAP_KEY,
   COLUMN_MANAGEMENT_LOCAL_STORAGE_KEY,
-  useActiveNamespace,
   withLastNamespace,
+  LAST_NAMESPACE_NAME_LOCAL_STORAGE_KEY,
+  LAST_NAMESPACE_NAME_USER_SETTINGS_KEY,
+  useUserSettingsCompatibility,
+  isModifiedEvent,
 } from '@console/shared';
 import { ByteDataTypes } from '@console/shared/src/graph-helper/data-utils';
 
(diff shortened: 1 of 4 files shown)


TheRealJon (changes requested):
Tested this and it looks like this introduces a regression in dev console where active namespace is not properly retained on some pages:
screencast-github com-2021 06 16-12_27_20

frontend/public/components/namespace.jsx
```diff
@@ -530,25 +529,33 @@ const getProjectSelectedColumns = ({ showMetrics, showActions }) => {
...
+  const params = new URLSearchParams(url.search);
+  const basePath = url.pathname;
+  if (params.has('project-name')) {
+    // clear project-name query param from the url
+    params.delete('project-name');
+  }
+  const newUrl = {
+    ...url,
+    search: `?${params.toString()}`,
+    hash: url.hash,
```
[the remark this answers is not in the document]
rohitkrai03: For some reason, spreading of url doesn't work correctly. The `formatNamespaceRoute` expects hash property of location object so setting it explicitly.

rohitkrai03:
> Tested this and it looks like this introduces a regression in dev console where active namespace is not properly retained on some pages:
> screencast-github com-2021 06 16-12_27_20

@TheRealJon I am not seeing that issue anymore. It was the problem when i raised the PR as WIP but fixed it in my latest push. Are you sure you are testing with latest code?

frontend/public/components/namespace.jsx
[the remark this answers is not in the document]
rohitkrai03: It turns out that `URL` object is not enumerable which is why we cannot spread it or copy properties. Removed the spread and now setting just the properties we need.

TheRealJon:
@rohitkrai03 Yes, I just pulled down your latest changes and it does appear to be fixed now.

frontend/public/components/namespace.jsx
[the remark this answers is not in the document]
rohitkrai03: We need to remove `project-name` from query params, that is why I am creating the location object like this.

frontend/public/components/namespace.jsx
```diff
@@ -530,25 +532,44 @@ const getProjectSelectedColumns = ({ showMetrics, showActions }) => {
...
+    params.delete('project-name');
+  }
+  const newUrl = {
+    search: `?${params.toString()}`,
+    hash: url.hash,
+  };
+  const namespacedPath = UIActions.formatNamespaceRoute(project.metadata.name, basePath, newUrl);
+
+  const handleClick = () => {
+    setLastNamespace(project.metadata.name);
```
jerolimov, line 553: This (and maybe the complete handleClick should be canceled similar to #9257 when it is a ctrl+click.

```suggestion
    if (isModifiedEvent(e)) {
      return;
    }
    setLastNamespace(project.metadata.name);
```

Scenario:

When the user ctrl+clicks multiple namespaces to check their (topology) content.

Open / or /add or /topology again should not bring up the latest opened namespace, or?

frontend/public/components/namespace.jsx
```diff
@@ -530,25 +532,44 @@ const getProjectSelectedColumns = ({ showMetrics, showActions }) => {
...
+  };
+
   return (
     <span className="co-resource-item co-resource-item--truncate">
       <ResourceIcon kind="Project" />
-      <Button
-        isInline
+      <Link
+        to={namespacedPath}
         title={project.metadata.name}
```
jerolimov, line 567: Can we remove the `title` prop? Don't see a reason for a tooltip with the exact same content. Also mentioned that in the ticket.

jerolimov (commented):
Tested it locally. Works fine, but I think we should not set the namespace if the user ctrl+clicks a link. Wdyt?

frontend/public/components/namespace.jsx
rohitkrai03: Good call on that. Just updated the code to skip setting namespace on modified clicks.

frontend/public/components/namespace.jsx
rohitkrai03: Done.

jerolimov (approved):
Tested updated version, works fine.
/lgtm

PR #9263, opened 2021-06-15
Fix sample list flickering in the getting started card on the add page

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-6024

**Analysis / Root cause**: 
`SampleGettingStartedCard` uses `CatalogServiceProvider` to load all samples. At the moment this loads samples with two extensions via `useDevfileSamples` and `useBuilderImageSamples`.

The sample card shows a loading indicator until the provider calls the inner children prop with `loaded=true`. And the `CatalogServiceProvider` returns loaded=true only if **all extensions** has fetched their data. See [CatalogServiceProvider.tsx#L47-L50](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/catalog/service/CatalogServiceProvider.tsx#L47-L50)

While debugging the `onValueResolved` ([CatalogServiceProvider.tsx#L71-L73](https://github.com/openshift/console/blob/master/frontend/packages/dev-console/src/components/catalog/service/CatalogServiceProvider.tsx#L71-L73)) I noticed that this method was called for devfile already before the actual data are fetched.

And this happen finally because `useDevfileSamples` returns loaded true already before the API call was resolved.

**Solution Description**: 
`const loaded = !!devfileSamples;` returns true if the initial state is `[]`. Removed this state to return loaded=false until the API call was resolved.

Flickering on the getting started card disappear.

Added new tests:
```
 PASS  packages/dev-console/src/components/catalog/providers/__tests__/useDevfileSamples.spec.ts
  useDevfileSamples:
    ✓ should return loaded false until data are loaded (22ms)
    ✓ should return loaded false until fetch fails (4ms)
```

**Test setup:**
Just refresh the add page again and again to check if the sample list in the getting started card flickers.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/dev-console/src/components/catalog/providers/__tests__/useDevfileSamples.data.ts
+++ b/frontend/packages/dev-console/src/components/catalog/providers/__tests__/useDevfileSamples.data.ts
@@ -1,0 +1,116 @@
+import { CatalogItem } from '@console/dynamic-plugin-sdk/src';
+import { DevfileSample } from '../../../import/devfile/devfile-types';
+
+export const devfileSamples: DevfileSample[] = [
+  {
+    name: 'nodejs-basic',
+    displayName: 'Basic NodeJS',
+    description: 'A simple Hello world NodeJS application',
+    icon: 'trimmed',
+    tags: ['NodeJS', 'Express'],
+    projectType: 'nodejs',
+    language: 'nodejs',
+    git: {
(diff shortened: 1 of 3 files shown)


rohitkrai03 (approved):
/lgtm

PR #9264, opened 2021-06-15
Bug 1967047: Console overview operators shown upgrading when still waiting on approval

jcaianirh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1967047

Check if upgrade requires approval before changing the state to Upgrading from Upgrade available


TheRealJon (approved):
/lgtm
/approved

PR #9265, opened 2021-06-15
Console 2537: Prompt the user to refresh on plugin change (API changes)

florkbr:
These changes are adding the new endpoint that the UI can poll
in order to check for plugin changes.

https://issues.redhat.com/browse/console-2537


pkg/server/server.go
```diff
@@ -439,6 +440,13 @@ func (s *Server) HTTPHandler() http.Handler {
 				pluginsHandler.HandlePlugins(w, r)
 			}),
 		))
+
```
jhadvig, line 443: We need to remove the `if len(s.EnabledConsolePlugins) > 0 {` block and register both handlers all the time, since we also should empty array of plugins.

jhadvig (changes requested):

pkg/server/server.go
florkbr: @jhadvig done!

pkg/server/server.go
```diff
@@ -422,24 +423,28 @@ func (s *Server) HTTPHandler() http.Handler {
...
-		))
-	}
+	handle(pluginsEndpoint, http.StripPrefix(
+		proxy.SingleJoiningSlash(s.BaseURL.Path, pluginsEndpoint),
+		authHandler(func(w http.ResponseWriter, r *http.Request) {
+			pluginsHandler.HandlePlugins(w, r)
+		}),
+	))
+
+	handle(pluginsListEndpoint, http.StripPrefix(
```
jhadvig, line 442: Since we are not doing anything with the request URL Path we dont need to strip the endpoint prefix.
So just
```suggestion
	handle(pluginsListEndpoint, authHandler(pluginsHandler.HandleListPlugins))
```

jhadvig (changes requested):
Going through the story description I see:
> "The console backend will expose an endpoint that lists the available plugins and the current plugin version from each manifest"

So currently we are not passing the plugin version just the name. Also the [ConsolePlugin API ](https://github.com/openshift/api/blob/master/console/v1alpha1/types_console_plugin.go) does not contain any `version` field. Is this something we really want?

Also wondering how it would work if there would be two versions of a plugin?

spadgett:
> So currently we are not passing the plugin version just the name. Also the ConsolePlugin API does not contain any version field. Is this something we really want?

Right, it looks like the version only comes from the plugin manifest, which is read by the frontend. I'd leave that out of the backend endpoint for now. We probably need the frontend to poll if we want to detect updates.

The more important use case IMO is when plugins are added and removed.

cc @vojtechszocs

pkg/plugins/handlers.go
```diff
@@ -76,6 +76,19 @@ func (p *PluginsHandler) HandlePlugins(w http.ResponseWriter, r *http.Request) {
...
+		w.Header().Set("Allow", "GET")
+		serverutils.SendResponse(w, http.StatusMethodNotAllowed, serverutils.ApiError{Err: "Method unsupported, the only supported methods is GET"})
+		return
+	}
+	pluginsList := make([]string, 0, len(p.PluginsEndpointMap))
+	for k := range p.PluginsEndpointMap {
+		pluginsList = append(pluginsList, k)
+	}
+	serverutils.SendResponse(w, http.StatusOK, pluginsList)
+}
```
spadgett, line 90: Can you make this more general and also include the console version in the response? That way we can also prompt to refresh when console itself rolls out a new version.

I believe you can get the version from env var `SOURCE_GIT_COMMIT` when console is running in-cluster.

florkbr: @spadgett can you be more specific around what you'd like to see WRT making this more general? I've added the console version to the response in the form of a struct:

```go
struct {
    ConsoleVersion string   `json:"consoleVersion"`
    Plugins        []string `json:"plugins"`
}{
    ConsoleVersion: os.Getenv("SOURCE_GIT_COMMIT"),
    Plugins:        pluginsList,
}
```

spadgett: Looks good. We might rename the endpoint since it does more than just list plugins now, but that's not a critical issue.

I've confirmed that `SOURCE_GIT_COMMIT` does in fact track the openshift/console commit sha in product builds, so this looks like the right env var to use.

florkbr: Chatted with Sam this morning. We are going to rename `consoleVersion` to `consoleCommit` for clarity. So our new struct will look like:
```go
struct {
    ConsoleCommit string   `json:"consoleCommit"`
    Plugins        []string `json:"plugins"`
}{
    ConsoleCommit: os.Getenv("SOURCE_GIT_COMMIT"),
    Plugins:        pluginsList,
}
```

pkg/server/server.go
```diff
@@ -55,6 +55,7 @@ const (
 	devfileEndpoint                  = "/api/devfile/"
 	devfileSamplesEndpoint           = "/api/devfile/samples/"
 	pluginsEndpoint                  = "/api/plugins/"
+	pluginsListEndpoint              = "/api/plugins"
```
spadgett, line 58: It looks like we have two endpoints at this path now... We might be overriding an existing endpoint.

spadgett: Or maybe the one just above is a prefix? We should confirm that there's no conflict.

florkbr: @spadgett from what I could tell it seems `/api/plugins/` is a prefix used to request information about individual plugins. The plugin name would come after the last forward slash: https://github.com/openshift/console/blob/master/pkg/plugins/handlers.go#L35

florkbr:
@spadgett I've added the UI changes with this PR

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
```diff
@@ -80,6 +83,43 @@ const ConsolePluginStatus: React.FC<ConsolePluginStatusType> = ({ enabled, plugi
 
 const ConsolePluginsList: React.FC<ConsolePluginsListType> = ({ obj }) => {
   const { t } = useTranslation();
+  const toastContext = useToast();
+  const [pluginsData, pluginsError] = useURLPoll<ConsolePluginsListResponseType>(
```
spadgett, line 87: We'd want the toast notification to show up on any page of console, so we'd need to move the logic elsewhere (maybe to the masthead?)

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
```diff
@@ -80,6 +83,43 @@ const ConsolePluginStatus: React.FC<ConsolePluginStatusType> = ({ enabled, plugi
...
+  );
+
+  const prevPluginsDataRef = React.useRef<ConsolePluginsListResponseType>();
+  React.useEffect(() => {
+    prevPluginsDataRef.current = pluginsData;
+  });
+  const prevPluginsData = prevPluginsDataRef.current;
+
+  const pluginsStateInitialized = _.isEmpty(pluginsError) && !_.isEmpty(prevPluginsData);
+  const pluginsChanged = !_.isEqual(prevPluginsData?.plugins, pluginsData?.plugins);
```
spadgett, line 98: I'd suggest `_.xor` so that the order of the plugins doesn't matter. Maybe

```suggestion
  const pluginsChanged = !_.isEmpty(_.xor(prevPluginsData?.plugins, pluginsData?.plugins));
```

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
```diff
@@ -80,6 +83,43 @@ const ConsolePluginStatus: React.FC<ConsolePluginStatusType> = ({ enabled, plugi
...
+  const consoleCommitChanged = !_.isEqual(
+    prevPluginsData?.consoleCommit,
+    pluginsData?.consoleCommit,
+  );
```
spadgett, line 102: For strings, I'd suggest simply

```suggestion
  const consoleCommitChanged = prevPluginsData?.consoleCommit !== pluginsData?.consoleCommit;
```

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
```diff
@@ -212,3 +252,8 @@ type ConsolePluginStatusType = {
 type ConsolePluginsListType = {
   obj: K8sResourceKind;
 };
+
+type ConsolePluginsListResponseType = {
+  consoleCommit?: string;
+  plugins?: any;
```
spadgett, line 258: This is an array of strings, right?

```suggestion
  plugins?: string[];
```

pkg/server/server.go
spadgett: I might generalize the endpoint anyway since it's not just plugins. Maybe something like `/api/check-updates` and `updatesEndpoint`?

spadgett (commented):
Thanks @florkbr

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
florkbr: @spadgett good point... FYI `masthead.jsx` and `masthead-toolbar.jsx` are both JSX and not TSX. I'm assuming we don't want to convert those and drop the typescript functionality for now?

spadgett: Thinking more on this, the top-level `App` component might be better, but that's also JSX.

masthead-toolbar is large enough that we probably want a separate tech debt story to update it.

pkg/server/server.go
florkbr: @spadgett done!

frontend/packages/console-app/src/components/console-operator/ConsoleOperatorConfig.tsx
florkbr: @spadgett done. I have moved the logic to the `app.jsx` container.

florkbr:
@spadgett code review comments addressed! Retested this on a dev cluster by putting console in unmanaged state and added a plugin

Screen Shot 2021-07-20 at 5 41 21 PM

Note 1: this now works on any page (overview page)
Note 2: this logic will poll the new endpoint every 15 seconds indefinitely. Does that sound ok?
Note 3: I noticed the commit the service is returning is blank in the response. Is that because the console is in an unmanaged state?

spadgett:
> Note 2: this logic will poll the new endpoint every 15 seconds indefinitely. Does that sound ok?

Seems OK. We might consider dropping back to 30 seconds? Either is probably fine, though.

> Note 3: I noticed the commit the service is returning is blank in the response. Is that because the console is in an unmanaged state?

Right, the commit will only be set in product builds, not builds you run locally. (That's something we could fix with a Dockerfile update, but not urgent.)

spadgett (approved):
/lgtm
/hold

spadgett:
for approvals

yapei:
in progress of testing

yapei:
@florkbr Shall we keep the prompt message until user close it? The message disappears quickly, user may not be aware of that and didn't have enough time to real all text there(it is short for me to read all text and click refresh button)

florkbr:
@yapei FYI I can also increase the timeout before it is auto-dismissed, but for now I will make the notification remain on the page until it is explicitly dismissed.

florkbr:
@yapei please try again when you have the chance! Thank you

spadgett (approved):
/lgtm

yapei:
@florkbr today I tested the changes on a production build created via cluster-bot, everything works well, one minor comment : shall we update the message to be more specific? 
```
Web console update is available
There has been an update to the web console. Ensure any changes have been saved and refresh your browser to access the latest version.
```
This message only appears when there is some changes to dynamic plugins, however current text seems too generic, user may don't understand what's happening, am I getting this message because of editing some resource YAML(actually I'm not doing any changes)?  For example cluster-admin enable/disable dynamic plugin while other developer are not aware of these changes, they see a prompt message in their browser given these text, that might be confusing ~ HDYT?

spadgett:
@yapei For (1), we've purposely kept it generic. This message can also be presented when there is a cluster update that updates console, not only plugin changes.

>  For example cluster-admin enable/disable dynamic plugin while other developer are not aware of these changes, they see a prompt message in their browser given these text, that might be confusing

That's another reason we want to keep it generic. We want the prompt to show for all users. They might not know or care about specific plugins changing, just that there is an update.

For (2), nothing will force the console to rollout after editing the config map in unmanaged state. You'd have to manually delete the pods.

I wouldn't put the console in unmanaged state to test. If you use cluster-bot, you should simply be able to add a `ConsolePlugin` and update the console config to enable the plugin.

spadgett (approved):
/lgtm

yapei:
Thank you for your comments @spadgett then I'm ok with all changes
/label qe-approved

PR #9268, opened 2021-06-16
Bug 1972514: KSVC - adds check for status in ksvc in util logic

invincibleJai:
**Fixes**: 

**Analysis / Root cause**: 
NPE is thrown in topology and err is shown in the error boundary and this happened on one of case where after creating knative service, the status field was not there in response.

**Solution Description**: 
- Status is usually there when any k8s resource is created, the same is the case for ksvc. Though status is an optional field and good to have a check added as part of this PR
- added specs

image

image

**Test setup:**

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/packages/knative-plugin/src/utils/get-knative-resources.ts
+++ b/frontend/packages/knative-plugin/src/utils/get-knative-resources.ts
@@ -17,6 +17,7 @@ import {
   CamelIntegrationModel,
   CamelKameletBindingModel,
 } from '../models';
+import { Traffic } from '../types';
 
 export type KnativeItem = {
   revisions?: K8sResourceKind[];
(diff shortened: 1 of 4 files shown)


frontend/packages/knative-plugin/src/utils/get-knative-resources.ts
```diff
@@ -306,3 +307,25 @@ export const knativeCamelKameletBindingResourceWatchers = (
     },
   };
 };
+
+export const getTrafficByRevision = (revName: string, service: K8sResourceKind) => {
+  if (!(service.status?.traffic?.length > 0)) {
```
divyanshiGupta, line 312: Just some more readable suggestions. Upto you if you want to change.
```suggestion
  if (!service.status?.traffic?.length)) {
```
or
```suggestion
  if (_.isEmpty(service.status?.traffic)) {
```

invincibleJai: Thanks @divyanshiGupta, make sense. Have updated PTAL.

PR #9269, opened 2021-06-16
Update example placeholder for labels section of odc add flows

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/ODC-6027

This PR updates example placeholder for labels section as `app` label can be misleading for the users since it is a default label and hence is added always when an app is created using add flows and also since only `user-labels` are editable via edit flows.

**Screenshot:**
Screenshot 2021-06-16 at 5 54 22 PM

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/components/import/advanced/LabelSection.tsx
+++ b/frontend/packages/dev-console/src/components/import/advanced/LabelSection.tsx
@@ -17,6 +17,7 @@ const LabelSection: React.FC = () => {
       <SelectorInput
         onChange={(val) => setFieldValue('labels', SelectorInput.objectify(val))}
         tags={labels}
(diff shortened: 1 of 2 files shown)


divyanshiGupta:
@beaumorley PTAL at this PR. I have updated the label placeholder to be `app.io/type=frontend` if that looks fine to you can add your approval.

rohitkrai03 (approved):
/approve

Code changes looks good to me. Will wait for UX to approve the placeholder text itself.

beaumorley:
I am not an expert in the syntax. If the syntax is correct, from UX POV it looks good. Thanks.

nemesis09:
verified locally
has approval from UX for the changes.
/lgtm

PR #9274, opened 2021-06-16
fixes: Serverless deployments are not shown in topology Consumption mode

sahil143:
fixes: https://issues.redhat.com/browse/ODC-6009

Problem: Filter provider is not wrapping the DataModelProvider which uses the `useDisplayFilters` that why filter in DataModelProvider are undefined and connectivity/consumption mode are not updated properly

Solution: Wrap DataModelProvider with FilterProvider

Screenshots:
Screenshot 2021-06-16 at 7 52 42 PM
Screenshot 2021-06-16 at 7 53 07 PM


invincibleJai:
/approve 

Thanks for the fix. Verified the changes work as expected.

PR #9276, opened 2021-06-16
Remove upstream UI URLs from backend

PhilipGough:
See https://issues.redhat.com/browse/MON-1591 for context

### DIFF
--- a/cmd/bridge/main.go
+++ b/cmd/bridge/main.go
@@ -111,11 +111,6 @@ func main() {
 	fStatuspageID := fs.String("statuspage-id", "", "Unique ID assigned by statuspage.io page that provides status info.")
 	fDocumentationBaseURL := fs.String("documentation-base-url", "", "The base URL for documentation links.")
 
-	fAlermanagerPublicURL := fs.String("alermanager-public-url", "", "Public URL of the cluster's AlertManager server.")
-	fGrafanaPublicURL := fs.String("grafana-public-url", "", "Public URL of the cluster's Grafana server.")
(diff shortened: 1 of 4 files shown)


kyoto:
Related frontend changes have now merged.
This PR LGTM, but maybe @jhadvig could take a look?
Related comment: https://github.com/openshift/console/pull/9283#issuecomment-867582415

/cc @jhadvig

jhadvig (approved):
@PhilipGough sorry for the delay, was on PTO.
Went though the story and the changes are looking good 👍 

/approve
/lgtm

jhadvig:
@spadgett please add the `approve` label.

PhilipGough:
@jhadvig hey, do you know what we need to do to move this on from it's current state? (stalled) :)

spadgett:
@simonpasquier Is the plan to remove these routes entirely? For the multi-cluster tech preview, we're looking at reaching out to the Thanos public endpoint on each cluster to request metrics at least until we are able to use ACM observability.

I think this PR specifically won't cause a problem since we talk to the services directly on the hub cluster, but I wanted to check if the routes are going away as well. We don't need Grafana, but we do potentially need to talk to Thanos and Alertmanager.

simonpasquier:
> Is the plan to remove these routes entirely? 

Eventually yes but of course, it's subject to your new requirements for the muti-cluster UI. I'll update the JIRA ticket (+ ping you on it) so it doesn't fall through the cracks.

PhilipGough:
Closing as obsolete - "the current plan is to keep the routes (because of the multi-cluster console effort which is going to be dev preview in 4.10)"

PR #9283, opened 2021-06-17
Monitoring: Remove Prometheus UI link and SERVER_FLAGS.*PublicURL entries

kyoto:
Backend changes for removing the SERVER_FLAGS.*PublicURL entries are covered by https://github.com/openshift/console/pull/9276


spadgett (approved):
/lgtm

spadgett:
We should look at removing these from the console operator as well since they're no longer used. Then we can remove the command line flags from the backend.

cc @jhadvig

PR #9286, opened 2021-06-17
add KindResourcePage dynamic extension

glekner:
This allows the plugin UI to be dynamic and work with any version of an available CRD without hardcoding versions

e.g.
before
```JSON
  {
    "type": "console.page/route",
    "properties": {
      "path": [
        "/k8s/ns/:ns/kubevirt.io~v1~VirtualMachine/:name",
        "/k8s/ns/:ns/kubevirt.io~v1alpha3~VirtualMachine/:name"
      ],
      "component": {
        "$codeRef": "VirtualMachinesDetailsPage"
      }
    },
    "flags": {
      "required": ["KUBEVIRT"]
    }
  }
```
after
```JSON
{
    "type": "console.page/resource/kind/details",
    "properties": {
      "kind": "VirtualMachine",
      "preferredVersion": "v1",
      "component": {
        "$codeRef": "VirtualMachinesDetailsPage"
      }
    },
    "flags": {
      "required": ["KUBEVIRT"]
    }
  }
```

@rawagner @christianvogt @vojtechszocs please review

### DIFF
--- a/frontend/public/components/factory/details.tsx
+++ b/frontend/public/components/factory/details.tsx
@@ -88,11 +88,18 @@ export const DetailsPage = withFallback<DetailsPageProps>(({ pages = [], ...prop
           component: (cProps) => renderAsyncComponent(p, cProps),
         })),
       ...dynamicResourcePageExtensions
-        .filter(
-          (p) =>
-            referenceForExtensionModel(p.properties.model) ===
-            (kindObj ? referenceFor(kindObj) : props.kind),
-        )
+        .filter((p) => {
+          if (p.properties.model.version) {
+            return (
+              referenceForExtensionModel(p.properties.model) ===
+              (kindObj ? referenceFor(kindObj) : props.kind)
+            );
+          }
+          return (
+            p.properties.model.group === kindObj.apiGroup &&
+            p.properties.model.kind === kindObj.kind
+          );
+        })
         .map(({ properties: { href, name, component: Component } }) => ({
           href,
           name,
--- a/frontend/public/components/resource-list.tsx
+++ b/frontend/public/components/resource-list.tsx
@@ -15,6 +15,7 @@ import {
   K8sKind,
   K8sResourceKindReference,
   kindForReference,
+  referenceForExtensionModel,
   referenceForModel,
 } from '../module/k8s';
 import {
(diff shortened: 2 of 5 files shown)


christianvogt:
@glekner please update your description to reflect the latest changes

frontend/public/components/factory/details.tsx
```diff
@@ -88,11 +90,19 @@ export const DetailsPage = withFallback<DetailsPageProps>(({ pages = [], ...prop
...
+            return (
+              referenceForExtensionModel(p.properties.model as ExtensionK8sModel) ===
+              (kindObj ? referenceFor(kindObj) : props.kind)
+            );
+          }
+          return (
+            referenceForModel(
+              findDiscoveryModelForGroupKind(p.properties.model.group, p.properties.model.kind),
+            ) === (kindObj ? referenceFor(kindObj) : props.kind)
+          );
```
christianvogt, line 104: Using `findDiscoveryModelForGroupKind` is incorrect here because a lookup by `group` and `kind` can result in picking up the wrong modal with the preferred version instead of the one which matches `kindObj`.

What you really need to do is simply check if the `kindObj` has the same `group` and `kind` of the extension.

frontend/public/components/resource-pages.ts
```diff
@@ -83,7 +84,7 @@ const addDynamicResourcePage = (
   map: ImmutableMap<ResourceMapKey, ResourceMapValue>,
   page: DynamicResourcePage,
 ) => {
-  const key = referenceForExtensionModel(page.properties.model);
+  const key = referenceForExtensionModel(page.properties.model as ExtensionK8sModel);
```
christianvogt, line 87: I don't like this cast, can we fix the accepted type to be: `{ group?: ..., version?: ..., kind?: ...} ` ?

frontend/public/components/resource-list.tsx
```diff
@@ -99,18 +99,20 @@ export const ResourceDetailsPage = connectToPlural((props: ResourceDetailsPagePr
...
       : null;
-  const componentLoader = getResourceDetailsPages(
-    detailsPageExtensions,
-    dynamicResourceListPageExtensions,
-  ).get(ref, () => Promise.resolve(DefaultDetailsPage));
+  const componentLoader =
+    getResourceDetailsPages(detailsPageExtensions, dynamicResourceListPageExtensions).get(ref) ||
+    getResourceDetailsPages(detailsPageExtensions, dynamicResourceListPageExtensions).get(
+      `${kindObj.apiGroup}~~${kindObj.kind}`,
+    );
```
christianvogt, line 106: I'd prefer that you construct an object with `apiGroup` and `kind` and use one of the `referenceFor*` functions instead of constructing the reference here like this. By doing this, the details on how to construct a reference is owned by a single code location.

frontend/public/components/factory/details.tsx
```diff
@@ -90,8 +89,8 @@ export const DetailsPage = withFallback<DetailsPageProps>(({ pages = [], ...prop
       ...dynamicResourcePageExtensions
         .filter(
           (p) =>
-            referenceForExtensionModel(p.properties.model) ===
-            (kindObj ? referenceFor(kindObj) : props.kind),
+            p.properties.model.group === kindObj.apiGroup &&
+            p.properties.model.kind === kindObj.kind,
```
christianvogt, line 93: when the model contains a version, you should be checking it for a match as well.

PR #9288, opened 2021-06-17
Bug 1972003: Monitoring dashboards: Custom time range fixes

kyoto:
Fix handling of manually entered dates (rather than those selected with
the date picker). Changed the date display format to YYYY-MM-DD so that
the DatePicker component can parse the dates.

Fix submit action to not submit unless dates and times could be parsed.

screenshot


spadgett (approved):
/lgtm

PR #9291, opened 2021-06-17
Pipeline Builder recognizes 'display name' errors as form errors

andrewballantyne:
**Fixes**: 
https://issues.redhat.com/browse/ODC-6033

**Analysis / Root cause**: 
`Display name` for tasks is unique as it's currently tied to the runAfter / edge construction of the visualization. So the error was not tracked in the form data, thus not part of the error block that disabled the submit button when an error was in play.

**Solution Description**: 
Use the `status` to hold onto the error state and error value when the `Display name` errors out. Tie this into the submit button so it disables the button when the error appears.

cc @openshift/team-devconsole-ux / @bgliwa01 

displayNameFix

Added tests for getBuilderTasksErrorGroup which merges in the `Display name` errors with the regular `yup` / form errors.

```
 PASS  packages/pipelines-plugin/src/components/pipelines/pipeline-builder/__tests__/utils.spec.ts (23.462s)
  ...
  getBuilderTasksErrorGroup
    ✓ should not error out if given nothing (1ms)
    ✓ should return a standard structure if there are no errors or status states
    ✓ should not capture any non task/finally errors
    ✓ should ignore other status values
    ✓ should merge in status name errors (1ms)
    ✓ should merge in status name errors at the index of the error
    ✓ should merge in status name errors for finally tasks (1ms)
    ✓ should merge both normal errors and status errors together
```

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


karthikjeeyar:
Verified locally, the change of name using pipeline builder view works as expected :+1: .

I have noticed an issue while authoring the pipeline from the yaml view (with empty space between a task name) and switching back to pipeline builder view, the create button is enabled and in the task sidebar the error is not shown under display name.
image

andrewballantyne:
> Verified locally, the change of name using pipeline builder view works as expected 👍 .
> 
> I have noticed an issue while authoring the pipeline from the yaml view (with empty space between a task name) and switching back to pipeline builder view, the create button is enabled and in the task sidebar the error is not shown under display name.
> image

Dang... the problem is with the way we do error checking for this field 🤦 

I'll see what I can do here, it won't be as simple as touching the field. But this may have exposed the problem with my fix. Let me see if I can work around this.

andrewballantyne:
> /lgtm

For posterity, we are going to go with another approach to handle this that is already being done in https://github.com/openshift/console/pull/9282 (needs rework at this point) and https://github.com/openshift/console/pull/9293 (will base itself off of 9282 & this PR work)

PR #9294, opened 2021-06-18
Bug 1973565: Evaluate dynamic route pages before static ones

rawagner:


jhadvig (approved):
/approve
/lgtm

spadgett:
/hold

Can we hold this for now and discuss next week? We should be careful about letting dynamic plugin replace core console behavior

rawagner:
Sure, i added it to agenta for next week

spadgett:
Closing in favor of #9395

PR #9295, opened 2021-06-18
Bug 1973338: Fix punctuation in pvc upload size warning

yaacov:
When reviewing the files from the translators, we noticed "PVC size is smaller than double the provided image, Please ensure your PVC size covers the requirements of the uncompressed image and any other space requirements" in kubevirt-plugin.json needs a period instead of the comma.

It should be "PVC size is smaller than double the provided image. Please ensure your PVC size covers the requirements of the uncompressed image and any other space requirements."


yaacov:
@glekner please review

PR #9297, opened 2021-06-18
Fix full screen loading indicator when routes are loaded lazy (or components uses React.Suspense)

jerolimov:
**Fixes**: 
https://issues.redhat.com/browse/ODC-5633

**Analysis / Root cause**: 
When switching on the Pod detail page to the "Terminal" tab a lazy route / AsyncComponent was loaded and the user saw for a moment (depending on your internet speed) a strange full screen loading indicator.

This flicking happen because this loading uses [react suspense](https://reactjs.org/docs/concurrent-mode-suspense.html) and we have only one `<React.Suspense />` at the application root level (app.jsx). This catch components works similar to ErrorBoundaries and could be used on different application levels.

**Solution Description**: 
Wrap the page content and horizontal navigation tab content also with a `<React.Suspense />` component.

This ensures that AsyncComponent loading on this levels shows a better loading indicator only in the content area. This would work also for other suspense usages.

I tested this by adding temporary suspense component to `PodExecLoader` (which us used in the terminal tab):

```tsx
// Use global variables and delays only once (when rendering Delay the first time)!
let resolved: boolean = false;
let promise: Promise<null>;

const Delay = () => {
  if (!promise) {
    promise = new Promise((resolve) => {
      setTimeout(() => {
        resolved = true;
        resolve(null);
      }, 2000);
    });
  }
  if (!resolved) {
    throw promise;
  }
  return <div>Delay done!</div>;
};

// Add <Delay /> to PodExecLoader
```

The original issue force with throttled internet connection:

original-issue

Added the `Delay` component above to `PodExecLoader` without any other change:

with-2s-suspense-call-no-other-change

When wrapping the page content (react router Switch) just in `app-contents.tsx`:

with-2s-suspense-call-and-catch-in-app-contents

This helps for all pages without horizontal nav. But the navigation goes away and come back after the `suspense` promise is resolved.

When wrapping also the tab content (react router Switch again) in `horizontal-nav.tsx`:

with-2s-suspense-call-and-catch-in-horizontal-nav

This fixes the issue that the tab navigation goes way and come back after the  `suspense` promise is resolved.

Unchanged

**Test setup:**
* Deploy and application
* Open the Pod detail page
* Reload your browser tab so that the Terminal component is not available!
* Switch to the Terminal tab

Better testable on a cluster or with throttled internet connection.

**Browser conformance**: 
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/frontend/public/components/app-contents.tsx
+++ b/frontend/public/components/app-contents.tsx
@@ -171,521 +171,524 @@ const AppContents: React.FC<{}> = () => {
       <div id="content">
         <GlobalNotifications />
         <Route path={namespacedRoutes} component={NamespaceBar} />
+
         <div id="content-scrollable">
-          <Switch>
-            {pluginPageRoutes}
+          <React.Suspense fallback={<LoadingBox />}>
+            <Switch>
+              {pluginPageRoutes}
 
-            <Route path={['/all-namespaces', '/ns/:ns']} component={RedirectComponent} />
-            <LazyRoute
-              path="/dashboards"
-              loader={() =>
-                import(
-                  './dashboard/dashboards-page/dashboards' /* webpackChunkName: "dashboards" */
-                ).then((m) => m.DashboardsPage)
-              }
-            />
+              <Route path={['/all-namespaces', '/ns/:ns']} component={RedirectComponent} />
+              <LazyRoute
+                path="/dashboards"
+                loader={() =>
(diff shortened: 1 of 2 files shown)


invincibleJai:
/lgtm 

verified the changes works as expected

PR #9300, opened 2021-06-18
Fix HealthCheckProbeData type

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-6037

**Solution description:**
Created types:
- `HealthCheckFormProbeData` : which contains properties required to populate the Health checks form
- `HealthCheckProbeData` : which contains properties required to update the k8s resource

**GIF:**
No UI changes

**Unit Tests:**
Update the unit tests


andrewballantyne (changes requested):
Looking good. It's a pity we don't have Deployment and Deployment Config defined more strictly so we can use your type within' 🤔 

One change to do because we really need to be typing both ends of a utility (I think there are other utilities in your files that are not double-ended typed, but we can fix those each time we touch the area rather than make this PR a hunt-and-fix PR -- because that may not have an easy end).

andrewballantyne (approved):

vikram-raj (commented):
/lgtm

PR #9301, opened 2021-06-20
Add plugin paths to tsconfig.json

glekner:
This fixes auto alias imports in VSCode.

before 
Screen Shot 2021-06-20 at 12 31 26

This would result in a runtime error

after
Screen Shot 2021-06-20 at 12 30 52


glekner:
@spadgett @christianvogt @vojtechszocs

glekner:
rebase

PR #9302, opened 2021-06-20
Bug 1974083: When branding is not redhat, no need to explicitly mark community support.

yaacov:
When branding is not redhat, no need to explicitly mark community support.

Screenshots:
before
Firefox_Screenshot_2021-06-20T13-00-29 902Z
after:
Firefox_Screenshot_2021-06-20T13-00-47 719Z
Firefox_Screenshot_2021-06-20T13-01-56 532Z


yaacov:
@avivtur @glekner @pcbailey @metalice please review

cc:// @matthewcarleton

PR #9306, opened 2021-06-21
create topology side panel using dynamic plugin extensions

sahil143:
fixes: https://issues.redhat.com/browse/ODC-5845

This PR creates the topology side panel to use the dynamic plugin extensions

Screenshot 2021-06-21 at 6 38 14 PM


frontend/packages/console-shared/src/utils/order-extensions.ts
```diff
@@ -0,0 +1,84 @@
+interface ItemsToSort {
```
sahil143, line 1: These utilities are copied from add-page-utils

[TODO]: refactor utils to create common utils which can be used to order the extensions

frontend/packages/helm-plugin/src/topology/sidebar/release-panel/tab-sections.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { GraphElement, Node } from '@patternfly/react-topology';
+import { useTranslation } from 'react-i18next';
+import { StatusBox } from '@console/internal/components/utils/status-box';
+import TopologyGroupResourcesPanel from '@console/topology/src/components/side-bar/TopologyGroupResourcesPanel';
+import { getResource } from '@console/topology/src/utils';
+import HelmReleaseOverview from '../../../components/details-page/overview/HelmReleaseOverview';
+import { TYPE_HELM_RELEASE } from '../../components/const';
+import TopologyHelmReleaseNotesPanel from '../../TopologyHelmReleaseNotesPanel';
+
+export const useHelmReleasePanelDetailsTabSection = (element: GraphElement) => {
```
christianvogt, line 11: We're going to have a lot of sections.
The first check must be made efficiently to filter out all the sections which do not apply to the given element.
These should not be rendered as a hook.

After successfully passing the check, they should return a component which itself can use hooks within.

frontend/packages/topology/src/components/side-bar/components/SideBarHeading.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useDetailsResourceLink } from '../providers/useDetailsResourceLink';
+import SideBarAlerts from './SideBarAlerts';
+import './SideBarHeading.scss';
+
+const SideBarHeading: React.FC<{ element: GraphElement }> = ({ element }) => {
+  const resourceLink = useDetailsResourceLink(element);
+  return (
+    <div className="overview__sidebar-pane-head resource-overview__heading">
+      <h1 className="co-m-pane__heading">
+        <div className="co-m-pane__name co-resource-item">{resourceLink}</div>
```
christianvogt, line 12: There needs to be a fallback resource link if one is not supplied via extension.
1) extract the resource and create a ResourceLink
2) render a non-linkable text based on the element's label

frontend/packages/helm-plugin/console-extensions.json
```diff
@@ -74,5 +74,86 @@
...
+    }
+  },
+  {
+    "type": "topology.details/tab-section",
+    "properties": {
+      "id": "helm-release-panel-tab-section-resource",
+      "tab": "helm-release-panel-tab-resource",
+      "section": { "$codeRef": "helmTopology.getHelmReleasePanelResourceTabSection" },
+      "insertAfter": "helm-release-panel-tab-section-details",
+      "insertBefore": "helm-release-panel-tab-section-releaseNotes"
```
christianvogt, line 141: Why both before and after properties?

Note that the order in which they are defined should remain in tact so long as they are defined by the same plugin. 
Therefore if the helm plugin is contributing all help sections, we should not even need to define the before or after properties.

frontend/packages/topology/src/components/side-bar/providers/useDetailsResourceLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+): React.Component | null | undefined => {
+  const [resurceLinkExtension, resolved] = useResolvedExtensions<DetailsResourceLink>(
+    isDetailsResourceLink,
+  );
+  const resourceLink = React.useMemo(() => {
+    return resolved
+      ? resurceLinkExtension
+          .sort((a, b) => a.properties.priority - b.properties.priority)
+          .find(({ properties: { link } }) => !!link(element))
+          .properties?.link?.(element)
```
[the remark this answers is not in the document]
sahil143: yes, this is intentional as we need to find the callback that doesn't return null and undefined and get result from that callback.
Couldn't find a lodash utility that does this.

rohitkrai03 (approved):
/approve

Changes looks good to me. Tested locally. Works as expected.

frontend/packages/helm-plugin/src/actions/providers.ts
```diff
@@ -16,3 +19,27 @@ export const useHelmActionProvider = (scope: HelmActionsScope) => {
...
+  }, [element]);
+  const { t } = useTranslation();
+  const actions = React.useMemo(
+    () => [
+      getHelmUpgradeAction(scope, t),
+      getHelmRollbackAction(scope, t),
+      getHelmDeleteAction(scope, t),
+    ],
+    [scope, t],
+  );
```
rohitkrai03, line 42: We can actually re use the original provider.
```suggestion
  const actions = useHelmActionProvider(scope);
```

rohitkrai03 (changes requested):

frontend/packages/helm-plugin/src/actions/providers.ts
sahil143: @rohitkrai03 fixed!

rohitkrai03 (approved):
/lgtm
/approve

PR #9307, opened 2021-06-21
Bug 1971532: remove title attribute when value is uid or same as name

rhamilto:
As noted in the bug, we were not consistent with the application of the `title` attribute value on resource links.  So I am removing any `title` props that set the value to `*.metadata.uid` or are simply duplicating the value of the `name` prop as the use of a `title` that is the same is unnecessary.


jhadvig (approved):
/approve
/lgtm

PR #9308, opened 2021-06-21
Bug 1970604:  Add IDP menu items are not translated

cyril-ui-developer:
/assign @rhamilto 
cc: @rebeccaalpert 

Screen Shot 2021-06-21 at 11 20 32 AM

### DIFF
--- a/frontend/public/components/cluster-settings/oauth.tsx
+++ b/frontend/public/components/cluster-settings/oauth.tsx
@@ -47,15 +47,6 @@ const IdentityProviders: React.FC<IdentityProvidersProps> = ({ identityProviders
   );
 };
 
-// t('public~Basic Authentication')
-// t('public~GitHub')
-// t('public~GitLab')
-// t('public~Google')
-// t('public~HTPasswd')
-// t('public~Keystone')
-// t('public~LDAP')
-// t('public~OpenID Connect')
-// t('public~Request Header')
 export const addIDPItems = Object.freeze({
   basicauth: 'Basic Authentication',
   github: 'GitHub',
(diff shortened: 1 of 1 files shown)


frontend/public/components/cluster-settings/oauth.tsx
```diff
@@ -73,17 +64,43 @@ const OAuthDetails: React.FC<OAuthDetailsProps> = ({ obj }: { obj: OAuthKind })
...
+        return type;
+    }
+  };
+
+  const IDPDropdownItems = Object.entries(addIDPItems).map((idp) => {
+    const [key, value] = idp;
+
     return (
       <DropdownItem
-        key={`idp-${addIDPItems[idp]}`}
```
cyril-ui-developer, line 80: @rhamilto @rebeccaalpert I am not sure why we were assigning `addIDPItems` object value here. I think it should be the key.

frontend/public/components/cluster-settings/oauth.tsx
```diff
@@ -47,15 +47,6 @@ const IdentityProviders: React.FC<IdentityProvidersProps> = ({ identityProviders
...
 
-// t('public~Basic Authentication')
-// t('public~GitHub')
-// t('public~GitLab')
-// t('public~Google')
-// t('public~HTPasswd')
-// t('public~Keystone')
-// t('public~LDAP')
-// t('public~OpenID Connect')
-// t('public~Request Header')
```
cyril-ui-developer, line 58: @rhamilto @rebeccaalpert i don't think we need the above any more.

rhamilto:
LGTM but will wait for @rebeccaalpert to give the final approval as the resident i18n expert.  :)

frontend/public/components/cluster-settings/oauth.tsx
rebeccaalpert: You're totally right; we don't need these anymore.

rebeccaalpert (approved):
Looks great to me. :) Thanks for fixing this.

PR #9313, opened 2021-06-22
Migrate Helm actions on topology to use the new extensions

debsmita1:
**JIRA ticket:**
https://issues.redhat.com/browse/ODC-5693

**Solution description:**
- Updated the topology context menu actions for helm release to use the new extension

**GIF/Screenshot:**
Screenshot from 2021-07-08 01-25-23
topology-context-menu
Peek 2021-07-08 01-37

### DIFF
--- a/frontend/packages/topology/src/actions/contextMenuActions.tsx
+++ b/frontend/packages/topology/src/actions/contextMenuActions.tsx
@@ -1,0 +1,12 @@
+import * as React from 'react';
+import { Node } from '@patternfly/react-topology';
+import { ActionsLoader } from '@console/shared';
+import { createContextMenuItems } from '../components/graph-view';
+
+export const contextMenuActions = (element: Node): React.ReactElement[] => {
+  return [
+    <ActionsLoader key="topology" contextId="topology-actions" scope={element}>
+      {(loader) => loader.loaded && createContextMenuItems(loader.options)}
+    </ActionsLoader>,
+  ];
+};
--- a/frontend/packages/helm-plugin/src/topology/components/helmComponentFactory.ts
+++ b/frontend/packages/helm-plugin/src/topology/components/helmComponentFactory.ts
@@ -1,17 +1,15 @@
 import * as React from 'react';
 import {
-  Node,
   GraphElement,
   withDragNode,
   withSelection,
   withDndDrop,
   withCreateConnector,
 } from '@patternfly/react-topology';
-import { kebabOptionsToMenu } from '@console/internal/components/utils';
+import { contextMenuActions } from '@console/topology/src/actions/contextMenuActions';
 import {
   WorkloadNode,
   noRegroupWorkloadContextMenu,
-  createMenuItems,
   createConnectorCallback,
   NodeComponentProps,
   nodeDragSourceSpec,
@@ -21,12 +19,8 @@ import {
   CreateConnector,
 } from '@console/topology/src/components/graph-view';
 import { withEditReviewAccess } from '@console/topology/src/utils';
-import { helmReleaseActions } from '../actions/helmReleaseActions';
 import { TYPE_HELM_RELEASE, TYPE_HELM_WORKLOAD } from './const';
 import HelmRelease from './HelmRelease';
-
-export const helmReleaseContextMenu = (element: Node) =>
-  createMenuItems(kebabOptionsToMenu(helmReleaseActions(element)));
 
 export const getHelmComponentFactory = (
   kind,
@@ -35,7 +29,7 @@ export const getHelmComponentFactory = (
   switch (type) {
     case TYPE_HELM_RELEASE:
       return withSelection({ controlled: true })(
-        withContextMenu(helmReleaseContextMenu)(withNoDrop()(HelmRelease)),
+        withContextMenu(contextMenuActions)(withNoDrop()(HelmRelease)),
       );
     case TYPE_HELM_WORKLOAD:
       return withCreateConnector(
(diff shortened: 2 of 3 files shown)


rohitkrai03 (commented):
/hold

@debsmita1 This PR seems to be just using `ActionLoader` to get extension based action but still using the old `KebabMenu` components. The goal of the story was to use new `KebabMenu` components potentially after https://issues.redhat.com/browse/ODC-5749 which would lead to some more refactoring in how topology uses actions.

rohitkrai03 (changes requested):

frontend/packages/console-shared/src/components/actions/menu/ActionMenuContent.tsx
```diff
@@ -151,12 +151,14 @@ const ActionMenuContent: React.FC<ActionMenuContentProps> = ({ options, onClick,
               );
             default:
               return (
-                <ActionMenuItem
-                  key={option.id}
-                  action={option as Action}
-                  onClick={onClick}
-                  autoFocus={focusItem ? option === focusItem : undefined}
-                />
+                <div className="pf-c-dropdown__menu-item">
```
debsmita1, line 154: Added this so that the menu items in the context menu are rendered with proper spacing and are focused when the cursor hovers on them

debsmita1:
Fixing the alignment of the Sub Menu and grouped menu items in the Context Menu

rohitkrai03:
@debsmita1 https://github.com/openshift/console/pull/9306 just merged. So you can remove the duplicate changes from this PR now.

frontend/packages/console-shared/src/components/actions/menu/ActionMenuContent.tsx
rohitkrai03: This might add some unnecessary styles to normal menu items.

frontend/packages/topology/src/actions/contextMenuActions.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { Node } from '@patternfly/react-topology';
+import { ActionsLoader } from '@console/shared';
+import ActionMenuContent from '@console/shared/src/components/actions/menu/ActionMenuContent';
+
+export const contextMenuActions = (element: Node): React.ReactElement[] => {
+  return [
+    <ActionsLoader contextId="topology-actions" scope={element}>
+      {(loader) => loader.loaded && <ActionMenuContent options={loader.options} />}
```
rohitkrai03, line 9: Not sending an onClick handler keeps the context menu open even after the callback has been called.
Screenshot 2021-07-07 at 5 41 50 PM

rohitkrai03 (changes requested):

frontend/packages/topology/src/actions/contextMenuActions.tsx
debsmita1: Changed the implementation here. PTAL

debsmita1:
> Fixing the alignment of the Sub Menu and grouped menu items in the Context Menu

Created util `createContextMenuItems` which returns the context menu items in a way similar to the menu rendering logic in `ActionMenuContent`

Screenshot from 2021-07-08 01-25-23

frontend/packages/console-dynamic-plugin-sdk/src/extensions/actions.ts
```diff
@@ -95,8 +97,10 @@ export const isActionFilter = (e: Extension): e is ActionFilter => {
 export type Action = {
   /** A unique identifier for this action. */
   id: string;
-  /** The label to display in the UI. */
+  /** The translated label to display in the UI. */
   label: string;
+  /**  To send translation key for label */
+  labelKey?: string;
```
sahil143, line 103: I don't think we need this prop. Since all new extensions are hook based, strings will be translated in the hook itself

sahil143 (changes requested):

frontend/packages/topology/src/actions/contextMenuActions.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import { Node } from '@patternfly/react-topology';
+import { ActionsLoader } from '@console/shared';
+import { createContextMenuItems } from '../components/graph-view';
+
+export const contextMenuActions = (element: Node): React.ReactElement[] => {
+  return [
+    <ActionsLoader
+      key={`topology-actions-${element.getType()}`}
```
christianvogt, line 9: `key` here can be a constant because there's no conflicts
`topology` is sufficient

frontend/packages/topology/src/actions/contextMenuActions.tsx
```diff
@@ -0,0 +1,16 @@
...
+      contextId="topology-actions"
```
christianvogt, line 10: We also need `resource` `contextId` for this util to be reused for other nodes.

@rohitkrai03 we need a way to avoid doing it this way and have a single loader

rohitkrai03: Yeah, I am working on adding support for resource actions in action loader. So we can provide the resource model along with contextId to the action loader and get back all the actions for a specific resource as well as a context.

debsmita1: do we put this PR on hold till then ?

rohitkrai03: I think we can get this merged now. The follow up story that I am working on will take care of resource actions.

rohitkrai03 (approved):
/lgtm
/approve

PR #9314, opened 2021-06-22
Bug 1972011: Fix time range for drag & drop in devconsole monitoring

vikram-raj:
**Fixes**: 
https://issues.redhat.com/browse/OCPBUGSM-30920

**Analysis / Root cause**: 
Dashboards display different time range when drag&drop on the first chart of the dashboard

**Solution Description**: 
Dispatch the onZoom event so that it will reflect all the charts in the dashboard.

Kapture 2021-06-22 at 20 16 51

### DIFF
--- a/frontend/packages/dev-console/src/components/monitoring/dashboard/MonitoringDashboardGraph.tsx
+++ b/frontend/packages/dev-console/src/components/monitoring/dashboard/MonitoringDashboardGraph.tsx
@@ -1,5 +1,12 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import { useDispatch } from 'react-redux';
+import {
+  monitoringDashboardsSetEndTime,
+  monitoringDashboardsSetTimespan,
+} from '@console/internal/actions/ui';
 import { PrometheusGraphLink } from '@console/internal/components/graphs/prometheus-graph';
 import { QueryBrowser } from '@console/internal/components/monitoring/query-browser';
 import { Humanize } from '@console/internal/components/utils';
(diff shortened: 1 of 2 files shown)


invincibleJai:
@vikram-raj do we need to preserve custom time range? i.e 
- go to devconsole
- select valid project
- go to monitoring
- select time range with drag and deop
- go back to topology and move back to monitoring 
- can see still time range is persisted.

I don't see harm but to be aware if this is expected as on the admin side we don't preserve

kyoto:
Thanks @invincibleJai. I think it would be better if the change were not persisted.

Also, since this is using the same redux entries as the admin Console, this will cause the zoom settings to persist when switching from Dev to Admin Console.

It might be cleaner to use `useReducer` / `useContext` instead of Redux in this case (see https://reactjs.org/docs/hooks-faq.html#how-to-avoid-passing-callbacks-down). (Maybe the admin perspective should too.)

andrewballantyne (approved):

andrewballantyne:
> Thanks @invincibleJai. I think it would be better if the change were not persisted.

If this is the state we want to go with, then I've approved the current code.


> Also, since this is using the same redux entries as the admin Console, this will cause the zoom settings to persist when switching from Dev to Admin Console.
> 
> It might be cleaner to use `useReducer` / `useContext` instead of Redux in this case (see https://reactjs.org/docs/hooks-faq.html#how-to-avoid-passing-callbacks-down). (Maybe the admin perspective should too.)

Not entirely sure what the follow up point was here @kyoto -- if we wanted to persist it?

kyoto:
> Not entirely sure what the follow up point was here @kyoto -- if we wanted to persist it?

I think we should not persist it, especially since the change would persist between Dev and Admin Consoles. It would probably be confusing UX.

To be clear, there is no need to change the Admin Console code as part of this PR. I'm just thinking that this Dev Console change should not persist in redux after you leave the page.

jerolimov (approved):
Tested with Firefox (which has this image drag and drop issue but two clicks works fine) and Chrome. Sets the custom time range and works as expected :+1: 

/lgtm

PR #9315, opened 2021-06-22
Bug 1973065: Preserve user annotations while editing an app

divyanshiGupta:
**Fixes:** https://issues.redhat.com/browse/OCPBUGSM-31048

**Analysis/Root Cause:**
Annotations cannot be edited directly via edit flows like labels but the default annotations values are updated if one or more form field values are modified. In order to keep the default annotations values up-to date old annotations were not being added and in turn the user annotations were also being lost.

**Solution:**
Only update the default annotations and keep the user annotations as is.

**Screen-recording:**

https://user-images.githubusercontent.com/20724543/122948165-546f4480-d398-11eb-8ed6-7e49580c7bce.mov

**Test coverage:**
Screenshot 2021-06-22 at 8 29 17 PM

/kind bug

### DIFF
--- a/frontend/packages/dev-console/src/utils/resource-label-utils.ts
+++ b/frontend/packages/dev-console/src/utils/resource-label-utils.ts
@@ -72,6 +72,17 @@ export const getTriggerAnnotation = (
   ]),
 });
 
+export const getUserAnnotations = (allAnnotations: { [key: string]: string }) => {
+  const defaultAnnotations = [
+    'app.openshift.io/vcs-uri',
+    'app.openshift.io/vcs-ref',
+    'openshift.io/generated-by',
+    'image.openshift.io/triggers',
+    'alpha.image.policy.openshift.io/resolve-names',
+  ];
+  return _.omit(allAnnotations, defaultAnnotations);
+};
+
 export const getPodLabels = (name: string) => {
   return {
     app: name,
@@ -90,8 +101,10 @@ export const mergeData = (originalResource: K8sResourceKind, newResource: K8sRes
   };
   if (mergedData.metadata.annotations) {
     mergedData.metadata.annotations = {
+      ...(isDevfileResource
+        ? originalResource?.metadata?.annotations
+        : getUserAnnotations(originalResource?.metadata?.annotations)),
       ...newResource.metadata.annotations,
-      ...(isDevfileResource ? originalResource?.metadata?.annotations : {}),
     };
   }
   if (mergedData.spec?.template?.metadata?.labels) {
--- a/frontend/packages/dev-console/src/utils/resource-label-utils.ts
+++ b/frontend/packages/dev-console/src/utils/resource-label-utils.ts
@@ -79,6 +79,8 @@ export const getUserAnnotations = (allAnnotations: { [key: string]: string }) =>
     'openshift.io/generated-by',
     'image.openshift.io/triggers',
     'alpha.image.policy.openshift.io/resolve-names',
+    'jarFileName',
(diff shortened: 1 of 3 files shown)


frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -72,6 +72,18 @@ export const getTriggerAnnotation = (
...
 
+export const getUserAnnotations = (allAnnotations: { [key: string]: string }) => {
+  const defaultAnnotations = [
+    'app.openshift.io/vcs-uri',
+    'app.openshift.io/vcs-ref',
+    'openshift.io/generated-by',
+    'image.openshift.io/triggers',
+    'alpha.image.policy.openshift.io/resolve-names',
+    'jarFileName',
+  ];
```
andrewballantyne, line 83: So we dropped `'kubectl.kubernetes.io/last-applied-configuration'` as well in the [Pipeline world](https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/components/pipelines/modals/common/utils.ts#L84-L95).

frontend/packages/dev-console/src/utils/resource-label-utils.ts
```diff
@@ -90,8 +102,10 @@ export const mergeData = (originalResource: K8sResourceKind, newResource: K8sRes
   };
   if (mergedData.metadata.annotations) {
     mergedData.metadata.annotations = {
+      ...(isDevfileResource
+        ? originalResource?.metadata?.annotations
+        : getUserAnnotations(originalResource?.metadata?.annotations)),
       ...newResource.metadata.annotations,
-      ...(isDevfileResource ? originalResource?.metadata?.annotations : {}),
```
andrewballantyne, line 94: Interesting how we went at this... but I think the movement to before is a good idea. Original sets it up, New overrides what it needs.

andrewballantyne (commented):
@rohitkrai03 / @christianvogt Did you guys wanna chime in here?

frontend/packages/dev-console/src/utils/resource-label-utils.ts
divyanshiGupta: You meant that this annotation should be added here too?

divyanshiGupta: Hmm I think we might want to remove it as I saw this annotation in other resources as well though we dont add this specifically from our add flows but it might interfere with updating the resources like it probably did with `PipelineRuns` though I am not 100% sure.

andrewballantyne: I think updates are different than creates. Updates I wanna say are good with this annotation being echoed because it's an update. But I don't think you can create a resource with it... I should try to test once I get a moment today to make sure my representation of the world is actually correct lol.

divyanshiGupta: We are not creating any resource with this annotation from add flows and I dont think that we have a similar case as `PipelineRuns` here or I am not aware of it.

andrewballantyne: Hmmm, okay, let's ignore this thread -- I'm having issues reproducing it but my cluster went down as I was playing with various K8s / CRD kinds. Clearly this doesn't seem to fit the issue at hand, at least not directly.

andrewballantyne (approved):
Tested, lgtm. Definitely would help seeing if there are any other annotations that we want to be weary of... But this should be a good one to back port.

rohitkrai03 (approved):
/lgtm

PR #9318, opened 2021-06-22
refactor sorting utils from add-page-utils

sahil143:
This PR extracts sorting utils from add-page-utils and move them to console-shared. So other extensions can use the same utils.


rohitkrai03 (approved):
/lgtm
/approve

PR #9319, opened 2021-06-22
Bug 1975913: fix optional workspace checkbox check/uncheck

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/ODC-6039

**Root analysis:**
In the YAML editor of the pipeline builder form, when the workspace is added like 
```
workspaces:
    - name: workspace
```

the field value of the optional checkbox is received as undefined. As a result, when the Checkbox is checked in the form view it sets the field value to `[""]` instead of  `true`

**Solution description:**
- set value in `ToggableFieldBase.tsx` as false when field value is undefined

**GIF:**
Peek 2021-06-22 23-30

### DIFF
--- a/frontend/packages/console-shared/src/components/formik-fields/ToggleableFieldBase.tsx
+++ b/frontend/packages/console-shared/src/components/formik-fields/ToggleableFieldBase.tsx
@@ -35,7 +35,7 @@ const ToggleableFieldBase: React.FC<ToggleableFieldBaseProps> = ({
       {children({
         ...field,
         ...props,
-        value: field.value,
+        value: field.value ?? false,
         id: fieldId,
         label,
         isChecked: field.checked,


andrewballantyne (changes requested):
I think your solution works, but it is reactionary. I think we need to fix this at the source of the problem.

What is happening under the hood is we are providing `undefined` as the value in this case to Formik. Formik interprets this as an unknown structure for a boolean and thus treats it as you want an array of values:

```
function getValueForCheckbox(currentValue, checked, valueProp) {
  if (valueProp == 'true' || valueProp == 'false') {
    return !!checked;
  }

  if (checked) {
    return Array.isArray(currentValue) ? currentValue.concat(valueProp) : [valueProp];
  }
  ...
```
Not entirely sure their thought pattern of a `type` of `checkbox` equals `array`... but it is what it is. 

FWIW, the call is `function getValueForCheckbox(undefined, false, "")`. The above logic shows this will get you `[""]` and thus make the boolean value break.

We have a bug in our `ToggableFieldBase.tsx` where we set `value` equal to `field.value` (`undefined` in this case).

Solution is to [modify it](https://github.com/openshift/console/blob/master/frontend/packages/console-shared/src/components/formik-fields/ToggleableFieldBase.tsx#L38) to flip to boolean when unset (ie `value: field.value ?? false`). I think this is a fairly safe operation as checkbox is a boolean construct.

debsmita1:
> I think your solution works, but it is reactionary. I think we need to fix this at the source of the problem.
> 
> What is happening under the hood is we are providing `undefined` as the value in this case to Formik. Formik interprets this as an unknown structure for a boolean and thus treats it as you want an array of values:
> 
> ```
> function getValueForCheckbox(currentValue, checked, valueProp) {
>   if (valueProp == 'true' || valueProp == 'false') {
>     return !!checked;
>   }
> 
>   if (checked) {
>     return Array.isArray(currentValue) ? currentValue.concat(valueProp) : [valueProp];
>   }
>   ...
> ```
> 
> Not entirely sure their thought pattern of a `type` of `checkbox` equals `array`... but it is what it is.
> 
> FWIW, the call is `function getValueForCheckbox(undefined, true, "")`. The above logic shows this will get you `[""]` and thus make the boolean value break.
> 
> We have a bug in our `ToggableFieldBase.tsx` where we set `value` equal to `field.value` (`undefined` in this case).
> 
> Solution is to [modify it](https://github.com/openshift/console/blob/master/frontend/packages/console-shared/src/components/formik-fields/ToggleableFieldBase.tsx#L38) to flip to boolean when unset (ie `value: field.value ?? false`). I think this is a fairly safe operation as checkbox is a boolean construct.

@andrewballantyne  I did try doing `value: field.value ?? field.checked` before putting up this PR but was unsure which one would be the more appropriate solution

andrewballantyne:
> @andrewballantyne I did try doing `value: field.value ?? field.checked` before putting up this PR but was unsure which one would be the more appropriate solution

🤔 
We could definitely use `field.checked` (not sure this is a valid prop today for CheckboxField though) but that's still not going to handle unidentified values/checked state. The problem is `undefined` isn't a boolean but in reality, `undefined` is `false` for checkbox values. Formik is bonkers (imo) for converting it to an array... but I am sure they have their reasons. 

Reacting in the field usage (those that use CheckboxField) itself is the only real problem I had which is why I investigated it into Formik to find out what was happening. The solution in our wrapper `ToggleableFieldBase` is up for grabs... if you think a `checked` prop from the field is useful, we can add that too... but I think we just need to convert the value to a boolean (or at least handle `undefined`).

andrewballantyne (approved):
Ah, took my simple solution. 👍 

@karthikjeeyar please give this a look over if you got a moment.

karthikjeeyar:
Verified locally, works fine
/lgtm

PR #9323, opened 2021-06-22
Bug 1970011: Fix edge case for "managed by" links

TheRealJon:
In the 'managed by' link components and helpers, account for the edge case where a resource has a CSV as an ownerReference.

- Update `isOwnedByCSV` helper to only return true when the CSV `group~version~kind` and name match the owner reference. This was the source of the bug. Previously, the match criteria was too broad and was causing `matchOwnerAndCSV` to return a false match.
- Update `ManagedByOperatorResourceLink` to only add the ownerReference `group~version~kind` and name to the link URL if it is not a CSV.

### DIFF
--- a/frontend/public/components/utils/managed-by.tsx
+++ b/frontend/public/components/utils/managed-by.tsx
@@ -5,9 +5,7 @@ import { useTranslation } from 'react-i18next';
 
 import { ResourceIcon } from './resource-icon';
 import {
-  groupVersionFor,
   K8sResourceCommon,
-  referenceForGroupVersionKind,
   referenceForModel,
   referenceForOwnerRef,
   OwnerReference,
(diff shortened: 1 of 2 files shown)


spadgett:
> * Update `isOwnedByOperator` helper to only return true when a CRD has a full `group~version~kind` match with the owner (this is not strictly a fix for this bug, but could have caused other issues where an ownerReference might have matched on an incorrect API version)

Version shouldn't matter. It's the same resource if the group and kind match regardless of API version.

TheRealJon:
> Version shouldn't matter. It's the same resource if the group and kind match regardless of API version.

I'll just remove this change since it doesn't specifically address this bug.

TheRealJon:
Opened https://github.com/openshift/console/pull/9331 instead (target branch in this PR was not named correctly)

PR #9326, opened 2021-06-23
[release-4.8]Bug 1971188: Creating a VM from the UI on OKD 4.7 fails

glekner:
manual cherry pick of 1967885 fix. #9258

### DIFF
--- a/frontend/packages/kubevirt-plugin/src/components/vms/vm-create-yaml.tsx
+++ b/frontend/packages/kubevirt-plugin/src/components/vms/vm-create-yaml.tsx
@@ -1,13 +1,11 @@
-import { safeLoad } from 'js-yaml';
 import * as React from 'react';
-
+import { safeLoad } from 'js-yaml';
 import { CreateYAMLProps } from '@console/internal/components/create-yaml';
 import { ErrorPage404 } from '@console/internal/components/error';
 import { AsyncComponent, LoadingBox } from '@console/internal/components/utils';
 import { connectToPlural } from '@console/internal/kinds';
 import { TemplateModel } from '@console/internal/models';
 import { k8sList } from '@console/internal/module/k8s';
-
 import {
   TEMPLATE_FLAVOR_LABEL,
   TEMPLATE_TYPE_BASE,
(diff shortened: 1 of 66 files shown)


yaacov:
/retest

@vojtechszocs @rawagner hi, this PR should fix https://bugzilla.redhat.com/show_bug.cgi?id=1971188 

Without this PR if you have CNV 2.6.5 (VM apiVersion v1alpha3) you will not see the VMs, so it's very important to get this into 4.8.

frontend/packages/console-shared/src/utils/owner-references.ts
```diff
@@ -16,7 +16,7 @@ export const compareOwnerReference = (
 
   return (
     obj.apiVersion === otherObj.apiVersion &&
-    obj.kind === otherObj.kind &&
+    kindForReference(obj.kind) === kindForReference(otherObj.kind) &&
```
spadgett, line 19: I'm confused why this change was needed. `obj.kind` should always be the k8s kind and not the internal console reference using `~`.

glekner: from what I tested there are usages where the kind does contain `~`
it is a needed safety net that doesn't hurt

yaacov:
@vojtechszocs @rawagner PTAL

frontend/packages/console-shared/src/utils/owner-references.ts
spadgett: What were the usages? That sounds like a bug.

frontend/packages/kubevirt-plugin/src/components/vms/vm-create-yaml.tsx
```diff
@@ -94,10 +92,6 @@ export const VMCreateYAML = (props: any) => {
   return userMode === 'template' ? (
     <CreateVMTemplateYAML {...props} />
   ) : (
-    <VMCreateYAMLLConnected
-      {...(props as any)}
-      kindObj={VirtualMachineModel.kind}
-      plural={VirtualMachineModel.plural}
-    />
+    <VMCreateYAMLConnected {...(props as any)} plural={VirtualMachineModel.plural} />
```
vojtechszocs, line 95: Compared to #9258 this is missing prop

```
kindObj={kubevirtReferenceForModel(VirtualMachineModel)}
```

so please double-check, otherwise looks good.

spadgett:
/hold
I'd like to better understand the change for the owner references check before merging.

yaacov:
@spadgett hi, I see that @glekner removed the changes in `owner-references.ts` can you take a look ?

spadgett (approved):
/lgtm

PR #9328, opened 2021-06-23
Monitoring: Clean up logic for whether to show Monitoring in nav

kyoto:
Since `NavSection` doesn't render unless it has `children`, this change
does not alter the behavior, just simplifies the code.


spadgett (approved):
/lgtm

PR #9329, opened 2021-06-23
CONSOLE-2152: Improve upgrade messaging when ClusterVersion Upgradeab…

rhamilto:
…le=False

Intentional omissions from design:

* Warning not added to new channels when available
* Warning not added to the about modal
* Link to installed Operators list does not include a filter as the list does not support filtering on status (@megan-hall to add a story to cover this omission)

https://user-images.githubusercontent.com/895728/123328740-f19dba80-d509-11eb-96bb-b30edd631e43.mov

Screen Shot 2021-06-25 at 8 49 36 AM

TODO:

- [ ] test notUpgradeableCSVs once they are available in a payload

To test:

- Create a dummy ClusterOperator with `Upgradeable=False`
```
$ cat <<EOF >co.yaml 
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  name: testing
spec: {}
EOF
$ oc apply -f co.yaml
$ oc proxy &
$ curl -k -XPATCH -H "Accept: application/json" -H "Content-Type: application/json-patch+json" 'http://127.0.0.1:8001/apis/config.openshift.io/v1/clusteroperators/testing/status' -d '[{"op": "add", "path": "/status", "value": {"conditions": [{"lastTransitionTime": "2021-06-01T01:01:01Z", "type": "Upgradeable", "status": "False", "reason": "Testing", "message": "The whatsits are broken."}]}}]'
$ killall oc
```


frontend/public/module/k8s/cluster-settings.ts
```diff
@@ -39,6 +39,29 @@ export const getSortedUpdates = (cv: ClusterVersionKind): ClusterUpdate[] => {
...
+      const updateParsed = semver.parse(update.version);
+      return (
+        `${updateParsed.major}.${updateParsed.minor}` >
+        `${currentVersionParsed.major}.${currentVersionParsed.minor}`
+      );
+    },
+  );
+};
+
+export const getMinorVersionIsNewer = (currentVersion, otherVersion) => {
```
spadgett, line 56: Naming nit:
```suggestion
export const isMinorVersionNewer = (currentVersion, otherVersion) => {
```

frontend/public/module/k8s/cluster-settings.ts
```diff
@@ -39,6 +39,29 @@ export const getSortedUpdates = (cv: ClusterVersionKind): ClusterUpdate[] => {
...
+  const currentVersionParsed = semver.parse(currentVersion);
+  const otherVersionParsed = semver.parse(otherVersion);
+  return (
+    `${otherVersionParsed.major}.${otherVersionParsed.minor}` >
+    `${currentVersionParsed.major}.${currentVersionParsed.minor}`
+  );
```
spadgett, line 62: It's not safe to compare these as strings

```
>> "4.10" > "4.9"
<- false
```

The semver library has `lt` and `gt` functions for comparing versions.

spadgett (commented):
Nice work on this

rhamilto:
Thanks for the review, @spadgett.  Updated.

rhamilto:
> Thanks, @rhamilto. I see you still have this as WIP. Is that to get sign off from @megan-hall on the two omissions from the design?

No, it's so that I can test with a CSV that is `Upgradeable=False`, but I am still waiting for the necessary code to land in OLM (@kevinrizza is to alert me when it is ready).  In theory, I shouldn't need to test, but...

rhamilto:
Moved `ReleaseNotesLink` to its own file to fix analyze.

dtaylor113 (approved):
/lgtm
Had to lookup [<Flex ../>](https://www.patternfly.org/2020.04/documentation/core/layouts/flex) :-)

rhamilto:
Drat.  Moving `ReleaseNotesLink` resulted in a reordering of i18n values.  Fixed.

spadgett:
> No, it's so that I can test with a CSV that is `Upgradeable=False`, but I am still waiting for the necessary code to land in OLM (@kevinrizza is to alert me when it is ready). In theory, I shouldn't need to test, but...

I don't know that we should hold the PR unless the OLM changes are about to land... We're only looking at ClusterVersion / ClusterOperator conditions in this PR, correct? I think we can circle back and test the OLM use case when it's ready.

spadgett (approved):
/lgtm

rhamilto:
> I don't know that we should hold the PR unless the OLM changes are about to land... We're only looking at ClusterVersion / ClusterOperator conditions in this PR, correct? I think we can circle back and test the OLM use case when it's ready.

No, I added the relevant bits for CSVs, too.  See https://github.com/openshift/console/pull/9329/files#diff-e77aeff283ead04563609d584e37f08eb6dbb8cf46b88c6624f24954c8196587R820-R827  However, this should never display without the necessary OLM changes.

yapei:
@yanpzhan is going to test and will report test results back

spadgett:
> No, I added the relevant bits for CSVs, too. See https://github.com/openshift/console/pull/9329/files#diff-e77aeff283ead04563609d584e37f08eb6dbb8cf46b88c6624f24954c8196587R820-R827 However, this should never display without the necessary OLM changes.

OK, this seems small enough and safe enough to me to merge, and we can test when the OLM changes are there.

yanpzhan:
After tested against the pr, the upgrade message is shown as designed. New case is drafted and added in jira issue.

rhamilto:
> After tested against the pr, the upgrade message is shown as designed. New case is drafted and added in jira issue.

Hmm.  I am not seeing the test case in https://issues.redhat.com/browse/CONSOLE-2152.

rhamilto:
/hold cancel

all approvals added

yanpzhan:
> > After tested against the pr, the upgrade message is shown as designed. New case is drafted and added in jira issue.
> 
> Hmm. I am not seeing the test case in https://issues.redhat.com/browse/CONSOLE-2152.

Sorry, my comment didn't send out successfully in jira issue, have added comment again.

PR #9331, opened 2021-06-23
Bug 1970011: Fix edge case for "managed by" links

TheRealJon:
In the 'managed by' link components and helpers, account for the edge case where a resource has a CSV as an ownerReference.

- Update isOwnedByCSV helper to only return true when the CSV group~version~kind and name match the owner reference. This was the source of the bug. Previously, the match criteria was too broad and was causing matchOwnerAndCSV to return a false match.
- Update ManagedByOperatorResourceLink to only add the ownerReference group~version~kind and name to the link URL if it is not a CSV.

### DIFF
--- a/frontend/public/components/utils/managed-by.tsx
+++ b/frontend/public/components/utils/managed-by.tsx
@@ -4,11 +4,9 @@ import { Link } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { ResourceIcon } from './resource-icon';
+import { resourcePathFromModel } from './resource-link';
 import {
-  groupVersionFor,
   K8sResourceCommon,
-  referenceForGroupVersionKind,
-  referenceForModel,
   referenceForOwnerRef,
   OwnerReference,
   modelFor,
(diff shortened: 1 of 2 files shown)


frontend/public/components/utils/managed-by.tsx
```diff
@@ -19,38 +17,39 @@ import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager/
...
-
+export const ManagedByOperatorResourceLink: React.SFC<ManagerLinkProps> = ({
+  csvName,
+  namespace,
+  owner,
+  className,
+}) => {
+  const csvGroupVersionKind = referenceForModel(ClusterServiceVersionModel);
+  const ownerGroupVersionKind = referenceForOwnerRef(owner);
+  const ownerIsCSV = ownerGroupVersionKind === csvGroupVersionKind;
```
spadgett, line 28: I think we just want to check group and kind here. We don't care about the version... It's the same resource regardless of the API version referenced.

frontend/public/module/k8s/managed-by.ts
```diff
@@ -9,15 +16,22 @@ const isOwnedByOperator = (csv: ClusterServiceVersionKind, owner: OwnerReference
...
 };
 
-const isOwnedByCSV = (csv: ClusterServiceVersionKind, owner: OwnerReference) =>
-  csv.kind === owner.kind && csv.apiVersion === owner.apiVersion;
+// Check if csv === owner (there is no need to check namespace because ownerReferences must
+// be in the same namespace as the owned resource, or be cluster-scoped).
+const isOwnedByCSV = (csv: ClusterServiceVersionKind, owner: OwnerReference) => {
+  const csvGroupVersionKind = referenceFor(csv);
+  const ownerGroupVersionKind = referenceForOwnerRef(owner);
+  return csvGroupVersionKind === ownerGroupVersionKind && csv.metadata.name === owner.name;
```
spadgett, line 24: Again we don't want to check for any specific version, just group and kind.

TheRealJon:
@spadgett I've addressed your comments.

frontend/public/components/utils/managed-by.tsx
```diff
@@ -19,38 +17,40 @@ import { ClusterServiceVersionModel } from '@console/operator-lifecycle-manager/
...
+  namespace,
+  owner,
+  className,
+}) => {
+  const csvGroupVersionKind = referenceForModel(ClusterServiceVersionModel);
+  const ownerGroupVersionKind = referenceForOwnerRef(owner);
+  const { apiGroup, kind, namespaced } = modelFor(ownerGroupVersionKind) ?? {};
+  const ownerIsCSV =
+    apiGroup === ClusterServiceVersionModel.apiGroup && kind === ClusterServiceVersionModel.kind;
+  const link = `/k8s/ns/${namespace}/${csvGroupVersionKind}/${csvName}`;
```
spadgett, line 31: Not new, but use `resourcePathFromModel` instead of building the URL yourself. This will become important when we start putting the cluster name in URLs.

TheRealJon: @spadgett  updated

spadgett (approved):
/lgtm

PR #9333, opened 2021-06-23
Bug 1961391: Textual Updates 

rexagod:
Fixed the following bugs as reported here: https://bugzilla.redhat.com/show_bug.cgi?id=1961391#c10

* "Caching is a policy that creates local copies of the data. It saves the copies locally to improve performance for frequently accessed data. Each cached copy has a TTL and is verified against the hub. Each non-read operation(upload, overwrite, delete) is performed on the hub" - There should be a space before the parenthesis
* "And"/comma separated list - The concern here is that we're not making the full context available. Not all languages use comma-separated lists (or that they may use different styles of commas). The translators have been having trouble with the out-of-context "and" also.

### Modal now

ss


rexagod:
_Fixing tests._

afreen23:
tests are failing due to -
```
AssertionError: Missing i18n key "<0><0>{{poolName}}</0> cannot be deleted. When a pool is bounded to PVC it cannot be deleted. Please detach all the resources from StorageClass(es):</0><1><0><0>testing-sc</0></0></1>" in namespace "ceph-storage-plugin" and language "en.": expected false to be true
```
/lgtm cancel

afreen23:
squash commits.

afreen23:
/lgtm
Thanks

PR #9336, opened 2021-06-24
Bug 1975696: compareOwnerReference should not accept a reference

glekner:
cc @spadgett @vojtechszocs


frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/combined-disk.ts
```diff
@@ -343,7 +340,7 @@ export class CombinedDiskFactory {
               (getOwnerReferences(p) || []).some((ownerReference) =>
                 compareOwnerReference(ownerReference, {
                   name: dataVolumeName,
-                  kind: kubevirtReferenceForModel(DataVolumeModel),
+                  kind: DataVolumeModel,
```
spadgett, line 343: Shouldn't this be...?

```suggestion
                  kind: DataVolumeModel.kind,
```

spadgett (changes requested):

frontend/packages/kubevirt-plugin/src/k8s/wrapper/vm/combined-disk.ts
glekner: oops

spadgett (approved):
/lgtm

PR #9340, opened 2021-06-24
Feature flag MultiClusterHub and disable Web Terminal

abhinandan13jan:
# Addresses
Commit 1 -> https://issues.redhat.com/browse/ODC-5924 // Create feature flag for MultiClusterHub CR with Running status
Commit 2 -> https://issues.redhat.com/browse/ODC-5927 // Disable Web terminal based on MCH feature flag

# Description
Assumes structure of `SERVER_FLAGS` to be `SERVER_FLAGS.clusters: string [ ]` where presence of more than one entry defines `multiCluster` to be enabled
Creates a function `isMultiClusterEnabled` which reads data from `SERVER_FLAGS` for the presence of multiple clusters and returns accordingly. 
Updates Web terminal to read the presence of multiple clusters and render accordindly. 

# PR Testing
- Install `Web-terminal` operator and check for the `Cloudshell masthead` icon. `/terminal` route should be available.
Sometimes it doesn't work in localhost. You may need to manually set `useCloudShellAvailable` to `true` for testing purpose.
- Manually set `window.SERVER_FLAGS.clusters = [ 'clustera', 'clusterb']` . This data must exist before render.
- Reload, `CloudShell masthead` icon should disappear.
- `/terminal` route should be unavailable/redirected.

# Screenshot
Screencast from 07-01-2021 03_13_02 PM (1)

# Test
Added test-suite for utility `useMCHDetection` in console-app
Updated test-suite for `CloudShellTab`

# Browser Conformance
Chrome


frontend/packages/console-app/src/utils/useMCHDetection.ts
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore: FIXME missing exports due to out-of-sync @types/react-redux version
+import { useSelector } from 'react-redux';
+import { getFlagsObject, flagPending } from '@console/internal/reducers/features';
+import { RootState } from '@console/internal/redux';
+import { MCH_AVAILABILITY_FLAG } from '../features';
+
+const useMCHDetection = () => {
+  const flags = useSelector((state: RootState) => getFlagsObject(state));
+  const isMCHAvailable = flags[MCH_AVAILABILITY_FLAG];
```
christianvogt, line 10: `useFlag` instead?

frontend/packages/console-app/src/components/cloud-shell/CloudShellMastheadButton.tsx
```diff
@@ -20,9 +21,11 @@ type Props = StateProps & DispatchProps;
 
 const ClouldShellMastheadButton: React.FC<Props> = ({ onClick, open }) => {
   const terminalAvailable = useCloudShellAvailable();
+  const [isMCHAvailable, flagLoading] = useMCHDetection();
+
   const { t } = useTranslation();
 
-  if (!terminalAvailable) {
+  if (!terminalAvailable || (!flagLoading && isMCHAvailable)) {
```
christianvogt, line 28:
```suggestion
  if (!terminalAvailable || flagLoading || isMCHAvailable) {
```

frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
```diff
@@ -9,8 +10,9 @@ import './CloudShellTab.scss';
 const CloudShellTab: React.FC = () => {
   const { t } = useTranslation();
   const devWorkspaceFlag = useFlag(FLAG_DEVWORKSPACE);
+  const mchFLag = useFlag(MCH_AVAILABILITY_FLAG);
 
-  if (devWorkspaceFlag === false) return <Redirect to="/" />;
+  if (devWorkspaceFlag === false || mchFLag === true) return <Redirect to="/" />;
```
christianvogt, line 15: mch flag could take some time to be available but we shouldn't be showing the console while console in the meantime. We need to wait for the flag to be ready.

frontend/packages/console-app/src/plugin.tsx
```diff
@@ -193,6 +196,12 @@ const plugin: Plugin<ConsumedExtensions> = [
       required: [FLAGS.CLUSTER_VERSION],
     },
   },
+  {
+    type: 'FeatureFlag/Custom',
+    properties: {
+      detect: detectMCHAvailability,
+    },
+  },
```
christianvogt, line 204: Use dynamic plugin SDK because static plugins are being migrated.

frontend/packages/console-app/src/features.ts
```diff
@@ -0,0 +1,26 @@
+import { setFlag, handleError } from '@console/internal/actions/features';
+import { k8sList, K8sResourceKind } from '@console/internal/module/k8s';
+import { FeatureDetector } from '@console/plugin-sdk';
+import { MultiClusterHubModel } from './models/index';
+
+export const MCH_AVAILABILITY_FLAG = 'MULTICLUSTERHUB_AVAILABLITY';
+
+export const detectMCHAvailability: FeatureDetector = async (dispatch) => {
+  try {
+    const multiclusterHubs = await k8sList(MultiClusterHubModel);
```
spadgett, line 10: We probably want to key on the feature gate we plan to add in https://issues.redhat.com/browse/CONSOLE-2845. We'd only want to disable web terminal if the multicluster is enabled for console, which won't always be the case when ACM is installed.

@christianvogt

frontend/packages/console-app/src/features.ts
```diff
@@ -0,0 +1,26 @@
...
+      );
+      if (multiclusterHub) {
+        isMCH = true;
+      }
+    }
+    dispatch(setFlag(MCH_AVAILABILITY_FLAG, isMCH));
+  } catch (err) {
+    err?.response?.status === 404
+      ? dispatch(setFlag(MCH_AVAILABILITY_FLAG, false))
+      : handleError(err, MCH_AVAILABILITY_FLAG, dispatch, detectMCHAvailability);
```
spadgett, line 24: What is the access control for this resource? What happens for users who don't have authority to list MultiClusterHubs?

christianvogt:
cc @TheRealJon @jhadvig 
We need to ensure the correct mechanism is used to raise the flag when multi cluster is enabled.

spadgett:
> We need to ensure the correct mechanism is used to raise the flag when multi cluster is enabled.

The console operator will pass the backend a list of clusters, so the backend will know. Easiest might be to add something to `window.SERVER_FLAGS`.

TheRealJon:
@spadgett Good point about the RBAC issues of using MultiClusterHub as an indication of multicluster capability. That was by my suggestion, but I also suggested we look for the TechPreviewNoUpgrade config on the cluster, which will also have the same RBAC problems. I agree that window.SERVER_FLAGS might be the best way to go, in which case this story is probably blocked by https://issues.redhat.com/browse/CONSOLE-2845

frontend/packages/console-app/src/plugin.tsx
abhinandan13jan: noted

frontend/packages/console-app/src/components/cloud-shell/CloudShellTab.tsx
abhinandan13jan: noted

abhinandan13jan:
@spadgett @christianvogt @TheRealJon Keeping on hold and will update the implementation post https://issues.redhat.com/browse/CONSOLE-2845

abhinandan13jan:
Updated to mock implementation using `window.SERVER_FLAGS.isMultiClusterEnabled` to unblock the dependant stories. Will update to reading actual data when https://issues.redhat.com/browse/CONSOLE-2845 is merged.

cc: @andrewballantyne @christianvogt @spadgett @TheRealJon

christianvogt:
@abhinandan13jan please ensure your description is up to date with the latest implementation. Including the PR testing section.

abhinandan13jan:
> @abhinandan13jan please ensure your description is up to date with the latest implementation. Including the PR testing section.

Srry for the delay. Updated

frontend/packages/console-app/src/utils/isMultiClusterEnabled.ts
```diff
@@ -0,0 +1,11 @@
+const isMultiClusterEnabled = (): boolean => {
+  let isAvailable = false;
+  // Mock implementation. needs update
+  const mchFlagData = window.SERVER_FLAGS.clusters;
+  if (mchFlagData) {
+    isAvailable = mchFlagData.length > 1; // to-do: update logic/flag when finalised
+  }
+  return isAvailable;
+};
```
TheRealJon, line 9: nit, this could be simplified with optional chaining:
```suggestion
const isMultiClusterEnabled = (): boolean => window.SERVER_FLAGS.clusters?.length > 1;
```

abhinandan13jan: updated

frontend/packages/console-app/src/utils/isMultiClusterEnabled.ts
```diff
@@ -0,0 +1,4 @@
+const isMultiClusterEnabled = (): boolean => window.SERVER_FLAGS.clusters?.length > 1;
```
christianvogt, line 1: Is this valid?
What if the hub cluster only has a one managed cluster? Won't this array contain a single item?

Should we be checking for undefined or null instead?

TheRealJon: The cluster list will always include the hub cluster

christianvogt: Thanks for the clarification.

Ok so if there's one cluster it would mean that we are always targeting the hub cluster in which case the check is correct.

While maybe not correct for the web terminal which can work to access the cluster the console runs on. But that's another discussion.

christianvogt:
@abhinandan13jan @spadgett why the hold? Can this be merged now, it seems like it'll have no impact whether the server flag is present or not.

PR #9344, opened 2021-06-25
Fix typo in const pedingPhases

rhamilto:
Noticed this while investigating adding filters to Installed Operators.


spadgett (approved):
/lgtm
/kind cleanup

